PR #1, opened 2015-09-06
Updated theme

gentlecat:


mayhem:
\ø/ Yay! Thanks!

gentlecat:
There are no proper logos and icons, so it still needs some work.

PR #3, opened 2015-09-07
Cleanup

gentlecat:
:sparkles: :sparkles: :sparkles:

cc @mayhem, @alastair


webserver/views/user.py
```diff
@@ -14,43 +15,44 @@ def lastfmscraper(user_id):
...
-        user = db.user.get_by_mb_id(user_id)
-        if user is None:
-            raise NotFound("Can't find this user.")
 
     if lastfm_username:
-        params = {"base_url": url_for("user.lastfmscraper", user_id=user_id, _external=True),
-                "user_token": user.auth_token,
-                "lastfm_username": lastfm_username}
-        loader = render_template("user/loader.js", **params)
+        loader = render_template("user/loader.js", **{
```
alastair: if you're going to do this inline you may as well use kwargs

webserver/views/user.py
```diff
@@ -14,43 +15,44 @@ def lastfmscraper(user_id):
...
+
+
+def _get_user(user_id):
+    if current_user.is_authenticated() and \
+       current_user.musicbrainz_id == user_id:
+        return current_user
+    else:
+        user = db.user.get_by_mb_id(user_id)
+        if user is None:
+            raise NotFound("Can't find this user.")
```
alastair: I'm not sure about returning view/flask-specific exceptions from helper methods. However, if this was a generic exception then we'd need a try/except in the view. What do you think?

webserver/views/user.py
```diff
@@ -7,6 +7,7 @@
 
 user_bp = Blueprint("user", __name__)
 
+
 @user_bp.route("/lastfmscraper/<user_id>.js")
```
alastair: Not part of this commit, but we thought that this may not be a good place for this url, as it is in the `/user/` namespace (effectively preventing us having a user called `lastfmscraper`, however rare it may be)

webserver/views/api.py
```diff
@@ -208,9 +96,132 @@ def get_listens(user_id):
...
+        messybrainz_id = messy_response['messybrainz_id']
+    except KeyError:
+        current_app.logger.error("MessyBrainz did not return a proper id")
+        raise InternalServerError
+
+    recording_id = messy_response.get('recording_id', None)
+    artist_id = messy_response.get('artist_id', None)
+    release_id = messy_response.get('release_id', None)
+
+    return messybrainz_id, recording_id, artist_id, release_id
```
alastair: The content in here may change in the future. It may be better to just return the `messy_response` directly

webserver/views/api.py
```diff
@@ -20,140 +19,30 @@
...
 
-        if data['listen_type'] not in ('playing_now', 'single', 'import'):
-            log_and_raise_bad_request("JSON document requires a valid listen_type key", payload)
-
-        if (data['listen_type'] == "single" or data['listen_type'] == 'playing_now') and len(payload) > 1:
-            log_and_raise_bad_request("JSON document contains more than listen for a single/playing_now. "
-                             "It should contain only one. ", payload)
-    except KeyError:
-        log_and_raise_bad_request("Invalid JSON document submitted.", raw_data)
+    payload = _get_listen_payload(data, raw_data)
```
alastair: `get` or `validate`?

webserver/views/api.py
```diff
@@ -20,140 +19,30 @@
...
 
     producer = SimpleProducer(_kafka)
```
alastair: can you move this into a kafla module (like we have database stuff in db). The module won't get very complex, so perhaps just a single file. How about `data/kafka.py` and `data/cassandra.py` ?

webserver/views/api.py
```diff
@@ -208,9 +96,132 @@ def get_listens(user_id):
         del temp['user_id']
         listen_data.append(temp)
 
-    payload = {}
-    payload['user_id'] = user_id
-    payload['count'] = count
-    payload['listens'] = listen_data
+    return json.dumps({'payload': {
```
alastair: is this an api endpoint? is it better to use `jsonify` in order to send correct headers

webserver/views/user.py
gentlecat: Yeah, that's how we do in other places too. Done.

webserver/views/user.py
gentlecat: I think it's fine as long as this function is used within views in the same module.

webserver/views/user.py
gentlecat: Yep, this view probably should be in a separate blueprint.

webserver/views/api.py
gentlecat: It's both actually.

webserver/views/api.py
gentlecat: Yep.

webserver/views/api.py
gentlecat: Can we do it in another PR?

alastair: no problem

gentlecat:
:ship::anchor:?

webserver/views/api.py
alastair: i don't like the name of the method if it also validates

webserver/views/user.py
alastair: ok, can you make a note somewhere? we have LB in jira, right?

webserver/views/user.py
alastair: ok, fine

webserver/views/api.py
gentlecat: I can extract validation part into `validate_listen_payload` and put getting where it was. Would that work?

alastair:
yes :rocket:

webserver/views/api.py
alastair: yeah, better

PR #9, opened 2015-09-14
Use MAX_ITEMS_PER_GET as the upper bound of items per get

mineo:
Calling min here makes sure that MAX_ITEMS_PER_GET is used instead of
DEFAULT_ITEMS_PER_GET or request.args.get('count') when either of them
is larger than MAX_ITEMS_PER_GET.

Fixes LB-12.

This commit is untested because I don't want to fight VirtualBox right now.


mayhem:
Heh, yes. I ran out of time before ever getting the argument support working on this page. Thanks!

PR #10, opened 2015-09-14
Use the length of listen_data as count

mineo:
Nothing guarantees the user has actually listened to `count` (which is
defined by either count in the URL, DEFAULT_ITEMS_PER_GET or
MAX_ITEMS_PER_GET) tracks, so use the number of listens fetched earlier
as the count in the payload.

Fixes LB-7.

This commit is untested.


mayhem:
Thanks!

PR #12, opened 2015-09-15
[WIP] Pagination on user page

gentlecat:
I still need to add a URL parameter for keeping current page (timestamp).


mayhem:
So far so good! :+1:

webserver/views/api.py
```diff
@@ -129,6 +131,17 @@ def get_listens(user_id):
     }})
 
 
+def _parse_int_arg(name, default=None):
+        value = request.args.get(name)
+        if value:
+            try:
+                return int(value)
+            except ValueError:
+                raise BadRequest("Invalid %s argument." % name)
```
mineo: I think the error message should include `value` as well because not everyone reading the message is a developer or wants to capture network requests to find out what value some scrobbler is sending. MB itself also includes invalid values in error messages.

webserver/static/js/listens-viewer.js
```diff
@@ -0,0 +1,185 @@
...
+                    timestamp={listen.listened_at}
+                    />);
+            }.bind(this));
+            if (items.length > 0) {
+                return (
+                    <div>
+                        <table className="recordings table table-condensed table-hover">
+                            <thead>
+                            <tr>
+                                <th>track</th>
```
alastair: https://github.com/metabrainz/listenbrainz-server/commit/fb14df355b8353097d85a9f57939a1768abd8b22

webserver/views/api.py
gentlecat: Sounds like a good idea. Fixed.

webserver/static/js/listens-viewer.js
gentlecat: Fixed. :zap:

webserver/templates/user/profile.html
```diff
@@ -3,53 +3,18 @@
...
-                             + "<td>" + listens[i]['track_metadata']['track_name'] + "</td>" 
-                             + '<td><abbr class="timeago" title="' + s + '"></abbr></td></tr>') 
-    }
-    jQuery("abbr.timeago").timeago();
-}
-</script>
+{% block scripts %}
+  {{ super() }}
+  <script src="{{ url_for('static', filename='js/lib/react.js') }}"></script>
+  <script src="{{ url_for('static', filename='js/lib/JSXTransformer.js') }}"></script>
```
mwiencek: This shouldn't be used in production, so you might want to set up babel in a later PR. https://facebook.github.io/react/docs/tooling-integration.html#in-browser-jsx-transform

As of React 0.14 the JSXTransformer will be deprecated, too, and babel will be the only supported method: https://facebook.github.io/react/blog/#deprecation-of-react-tools

But, it'll work for now. :)

gentlecat: Oh yeah, I still need to figure out how to build all the JSX/JavaScript. In AcousticBrainz too, which is where I copied that from.

webserver/templates/user/profile.html
```diff
@@ -3,53 +3,18 @@
...
-  <h2 class="page-title">Most recent listens for {{ user.musicbrainz_id }}</h2>
-  <table id="listens" class="table table-striped table-bordered">
-    <tr><th>track</th><th>artist</th><th>time</th></tr>
-    <tr id="placeholder"><td>Loading your listens...</td></tr>
-  </table>
-
+  <div id="user-profile">
+    <h2 class="page-title">Most recent listens for {{ user.musicbrainz_id }}</h2>
+    <div id="listens-viewer-container" data-user-id="{{ user.musicbrainz_id }}">
+      <span class="loading-placeholder">Loading you listens...</span>
```
alastair: your?

gentlecat: :construction: :shipit: :construction:

PR #14, opened 2015-09-15
UI improvements

mayhem:
Use new large logo.
Cleanup the import page and make steps explicit.
Add a Import lists menu item.


webserver/templates/index/index.html
```diff
@@ -4,7 +4,7 @@
 
     <div class="col-md-7 col-lg-8">
       {# TODO: Replace that with SVG image. #}
```
alastair: no longer need a todo!

PR #16, opened 2015-09-16
Remember current page (timestamp) when switching between pages

gentlecat:
This allows users to provide links to specific pages.


webserver/static/js/listens-viewer.js
```diff
@@ -17,18 +18,42 @@ var ListensViewer = React.createClass({
...
-            }
-        }.bind(this));
+        window.addEventListener('popstate', this.handleLoadData);
+        this.handleLoadData();
     },
     componentDidUpdate: function() {
         jQuery("abbr.timeago").timeago();
     },
+    handleLoadData: function () {
+        if (this.state.ignorePopState) { return; }
```
gentlecat: This part is not working properly. http://chatlogs.metabrainz.org/brainzbot/metabrainz/2015-09-15/?msg=3351147&page=6

mwiencek: You may not need `ignorePopState` at all. If you just want to ignore events that get triggered by `pushState`, but not ones triggered by the user, then https://developer.mozilla.org/en-US/docs/Web/Events/popstate says:

> If the history entry being activated was created by a call to history.pushState() or was affected by a call to history.replaceState(), the popstate event's `state` property contains a copy of the history entry's state object.

So, make sure this function accepts `event` as the first argument, and do `if (event.state) { return; }` instead? (Probably with a comment to explain why it's doing that.)

mwiencek: That might be wrong though, since it could still have the `state` property of the added event, even if it's triggered by the user...

The other alternative I mentioned in IRC is just setting `this.ignorePopState = true/false` instead of using `setState`.

gentlecat: OK, I'll try that today. Thanks a lot for taking a look again!

alastair:
merged as-is for now (it works). We can address outstanding issues in another ticket

PR #20, opened 2015-09-16
Rework the URL structure:

mayhem:
login: /login
profile: /user/rob
import: /user/rob/import
scraper: /user/rob/scraper.js
get listens: /1/user/rob/listens
submit listens: /1/submit-listens

And add a fetch user by auth token method to the db.user object.

### DIFF
--- a/db/user.py
+++ b/db/user.py
@@ -32,6 +32,16 @@ def get_by_mb_id(musicbrainz_id):
     return dict(row) if row else None
 
 
+def get_by_token(token):
+    """Get user from an auth token"""
+    result = db.session.execute("""SELECT id, created, musicbrainz_id
+                                     FROM "user"
+                                    WHERE auth_token = :auth_token""",
+                                {"auth_token": token})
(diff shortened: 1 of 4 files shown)


mayhem:
We just had a length discussion on URL structure in IRC. We agreed to keep the version short.

gentlecat:
It's not about the version that you use in the URL. It's about name of the blueprint that you use in `url_for` calls. `1` is just confusing when I'm looking at `url_for("1.submit_listen", user_id=user_id, _external=True)`.

gentlecat:
And http://chatlogs.metabrainz.org/brainzbot/metabrainz/2015-09-16/?msg=3352140&page=5.

alastair:
This looks good to me. Note thata @gentlecat and I moved import to `/import`, because we can just use the current logged in user to get user details (which makes more sense, because you can only import your data).

alastair:
reminder to me to also deploy api.lb in nginx

PR #22, opened 2015-09-16
Date fix for scraper

mayhem:
Fix the error where an invalid date would throw an error. Now it sets the
timestamp to zero, which is invalid. This is actually a decent solution,
since it allows us to find these and fix them later. Kepstin pointed out
that the data via the API is correct, but via the web pages in incorrect.
This shows the data is still there, but this is caused by a bug in last.fm.

This may fix LB-16. Maybe not.

### DIFF
--- a/webserver/templates/user/scraper.js
+++ b/webserver/templates/user/scraper.js
@@ -53,7 +53,9 @@ var Scrobble = (function() {
 
         this.scrobbledAt = function() {
             var dateContainer = rootScrobbleElement.getElementsByClassName("chartlist-timestamp")[0]
-                var dateElement = dateContainer.getElementsByTagName("span")[0];
+            if (!dateContainer)
+                return 0;
+            var dateElement = dateContainer.getElementsByTagName("span")[0];
             var dateString = dateElement.getAttribute("title");
             //we have to do this because javascript's date parse method doesn't
             //directly accept lastfm's new date format but it does if we add the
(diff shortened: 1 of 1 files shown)


webserver/templates/user/scraper.js
```diff
@@ -53,7 +53,9 @@ var Scrobble = (function() {
 
         this.scrobbledAt = function() {
             var dateContainer = rootScrobbleElement.getElementsByClassName("chartlist-timestamp")[0]
-                var dateElement = dateContainer.getElementsByTagName("span")[0];
+            if (!dateContainer)
+                return 0;
```
gentlecat: Small nitpick: it's probably better to keep this on one line or add brackets around `return`.

alastair:
From https://github.com/metabrainz/listenbrainz-server/blob/master/listenstore/listenstore/listenstore.py#L17, we have a minimum submit time. I don't think this will affect importing though - it's just used as a default when no minimum time is given

alastair:
I've tested this. The scraper continues to run without errors, but the data is not inserted into cassandra. Is this acceptable for now, or do we want to try and insert it with timestamp 0?

mayhem:
Acceptable for now.

alastair:
Fixed - it was a red herring - if you don't give the listenstore a minimum timestamp it'll default from birth of audioscrobbler - the items with time=0 are still there

PR #24, opened 2015-09-16
Use `current_user` in import_data view

gentlecat:


webserver/templates/navbar.html
```diff
@@ -27,11 +27,7 @@
       {# Not showing user stuff on error pages. If attempt to load user info fails there, page will not render. #}
       {% if not error %}
         <ul class="nav navbar-nav navbar-right">
-          {% if not current_user or current_user.is_anonymous() %}
           <li class="import-link"><a href="{{ url_for('index.import_data') }}">Import!</a></li>
-          {% else %}
```
alastair: What does this do? Always send you to import, which redirects if you're logged in?

webserver/views/user.py
```diff
@@ -29,15 +29,18 @@ def profile(user_id):
     return render_template("user/profile.html", user=_get_user(user_id))
 
 
-@user_bp.route("/<user_id>/import")
-def import_data(user_id):
-    user = _get_user(user_id)
+@user_bp.route("/import")
```
alastair: ahh, here's the missing part. I understand now

webserver/templates/navbar.html
gentlecat: Yes, it redirects you to login page if you are not logged in.

PR #25, opened 2015-09-16
Wait until we've sent data to LB before reporting it as a submitted page

alastair:
Fixes LB-17


gentlecat:
I think that would make it easier to understand. :+1:

mayhem:
Without having seen it run, I like it. :+1:  (the bracketed line always broke the cleanliness of the dialog)

alastair:
fixed, with an additional off-by-one error

PR #27, opened 2015-09-17
Encode query before sending it to cassandra

alastair:
There seems to be a problem sending unicode queries to cassandra -
at this time there is an open pull request which offers a fix -
https://github.com/datastax/python-driver/pull/344/files
This change uses the same patch but moves the encoding into our
code.

Fixes LB-33

### DIFF
--- a/listenstore/listenstore/listenstore.py
+++ b/listenstore/listenstore/listenstore.py
@@ -6,6 +6,7 @@ import calendar
 import json
 import time
 import uuid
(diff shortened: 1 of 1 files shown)


mayhem:
Nice research to find this. Looks safe even after they fix the driver, no? :+1:

alastair:
yeah, their if statement won't trigger so it should continue to run fine

PR #29, opened 2015-09-18
Added checking to make sure that a submitted timestamp is an int.

mayhem:
As it says on the tin. Also included a check to make sure that the timestamp is actually present.

### DIFF
--- a/webserver/views/api.py
+++ b/webserver/views/api.py
@@ -205,6 +205,13 @@ def get_messybrainz_data(listen):
 
 def _validate_listen(listen):
     """Make sure that required keys are present, filled out and not too large."""
+
+    if not 'listened_at' in listen:
+        _log_raise_400("JSON document must contain the key listened_at at the top level.", listen)
+        
+    if type(listen['listened_at']) != int:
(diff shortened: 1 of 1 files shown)


webserver/views/api.py
```diff
@@ -205,6 +205,13 @@ def get_messybrainz_data(listen):
 
 def _validate_listen(listen):
     """Make sure that required keys are present, filled out and not too large."""
+
+    if not 'listened_at' in listen:
+        _log_raise_400("JSON document must contain the key listened_at at the top level.", listen)
+        
+    if type(listen['listened_at']) != int:
```
alastair: do we want this beahaviour? an alternative would be 
```
try:
    int(listen['listened_at'])
except ValueError:
    raise here
```
which would also let us accept strings from less-robust clients.

gentlecat: I agree. You can just convert it to integer, if that's what we use.

alastair:
awesome, go for it

gentlecat:
:ship: :shipit: :ship:

PR #30, opened 2015-09-18
use ujson for parsing/generating json

mayhem:
as it says on the tin.


alastair:
If you've tested that `loads` and `dumps` are drop-in replacement for `json`, this looks fine. go for it

mayhem:
Yep, tested a submit and saw it appear on the listens list.

PR #33, opened 2015-09-22
Fixes lb-46

mayhem:
Don't leak filehandles.


gentlecat:
[`finally`](https://docs.python.org/2/tutorial/errors.html#defining-clean-up-actions)?

mayhem:
Does finally get called even when an exception is raised?

gentlecat:
> A finally clause is always executed before leaving the try statement, whether an exception has occurred or not.

mayhem:
\ø/

alastair:
:car:

PR #34, opened 2015-09-22
Add notes about the import process.

mayhem:
Addresses LB-37 and LB-51.

### DIFF
--- a/webserver/templates/user/import.html
+++ b/webserver/templates/user/import.html
@@ -37,6 +37,22 @@
         </div>
      {% else %}
      <p>
+        <strong>Notes about the Last.fm import process:</strong>
+     </p>
+     <p>
+        The Last.fm importer manually steps through your listen history and imports the listens
+        one page at a time. Should it fail for whatever reason, it is safe to restart the import
+        process. Running the import process multiple times <strong>does not</strong> create duplicates in your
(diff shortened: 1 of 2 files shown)


webserver/templates/user/import.html
```diff
@@ -37,6 +37,22 @@ <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
         </div>
      {% else %}
      <p>
+        <strong>Notes about the Last.fm import process:</strong>
+     </p>
+     <p>
+        The Last.fm importer manually steps through your listen history and imports the listens
```
alastair: Here would be a place to additionally add 

 - Don't go away from this tab when it's running
 - It's better to not be listening/scrobbling at the same time

if wanted

alastair:
:+1:

webserver/templates/user/import.html
mayhem: Wanted, but this section is getting big. How about I add them to the second step?

alastair: agreed, it makes more sense there

gentlecat:
:ship:

webserver/templates/user/import.html
mayhem: Added LB-50 for this purpose.

PR #37, opened 2015-09-22
Return cors headers on error pages, submit queue

alastair:
The scraper xhr request to listenbrainz fails on http 4xx and 5xx error pages because the response doesn't include any cors headers.
We should set these headers on the request and check the status code in the javascript on the "success" callback.
We should not retry if the error is 4xx (bad data, bad auth), but retry if it's 5xx (possibly a transient server error)

Still to be done:

 - [x] implement 401 handler
 - [x] retry on the right error codes in the scraper 
 - [x] refactor error handlers to use a helper function

This is http://tickets.musicbrainz.org/browse/LB-21

### DIFF
--- a/webserver/templates/user/scraper.js
+++ b/webserver/templates/user/scraper.js
@@ -154,8 +154,8 @@ function reportScrobbles(struct) {
     xhr.setRequestHeader("Content-Type", "application/json;charset=UTF-8");
     xhr.timeout = 10 * 1000; // 10 seconds
     xhr.onload = function(content) {
-        numCompleted++;
         if (this.status >= 200 && this.status < 300) {
+            numCompleted++;
             console.log("successfully reported page");
         } else if (this.status >= 400 && this.status < 500) {
             console.log("4xx error, skipping");
@@ -164,7 +164,6 @@ function reportScrobbles(struct) {
             enqueueReport(struct);
         }
         getNextPageIfSlots();
-        console.log("successfully reported page");
     };
     xhr.ontimeout = function(context) {
         console.log("timeout, req'ing");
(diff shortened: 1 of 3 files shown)


alastair:
also added submit queue: http://tickets.musicbrainz.org/browse/LB-40

webserver/templates/user/scraper.js
```diff
@@ -121,17 +121,19 @@ function getLastFMPage(page, callback) {
...
 var numCompleted = 0;
+var activeSubmissions = 0;
 
 function dispatch() {
-    for (var i = 0; i < toReport.length; ++i) {
+    for (var i = 0; i < toReport.length; i++) {
         reportScrobbles(toReport[i]);
     }
-    toReport = [];
+    toReport = []
```
gentlecat: Don't need a semicolon there?

alastair: that's what I get for rewriting a block and not checking diff

webserver/templates/user/scraper.js
```diff
@@ -121,17 +121,19 @@ function getLastFMPage(page, callback) {
...
+var version = "1.2";
 var page = 1;
 var numberOfPages = parseInt(document.getElementsByClassName("pages")[0].innerHTML.trim().split(" ")[3]);
 
 var toReport = [];
 var numCompleted = 0;
+var activeSubmissions = 0;
 
 function dispatch() {
-    for (var i = 0; i < toReport.length; ++i) {
```
mayhem: What do you have against pre-decrement?? Hater!

mayhem:
Seems ok, with my limited JS knowledge.

alastair:
I tested this by making the server return 200, 400, 401, 500 and checking the logs to see that it did the right thing

PR #41, opened 2015-09-24
Move Scrobble methods to prototype

mwiencek:
(Untested)

The way it's currently written creates 7 new function objects for every single `new Scrobble`, which is wasteful of time and space. Putting methods on the prototype allows all `Scrobble` instances to share the same function objects, which are only created once.

### DIFF
--- a/webserver/templates/user/scraper.js
+++ b/webserver/templates/user/scraper.js
@@ -32,70 +32,70 @@ function isSpotifyURI(uri) {
     return !!(/open.spotify/.exec(uri));
 }
 
-var Scrobble = (function() {
-    return function(rootScrobbleElement) {
-        this.lastfmID = function() {
-            var loveButtonForm = rootScrobbleElement.getElementsByTagName("form")[0];
-            var loveButtonURL = loveButtonForm.getAttribute("action");
-            return extractlastFMIDFromLoveButtonURL(loveButtonURL);
-        }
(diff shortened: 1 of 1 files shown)


gentlecat:
:+1:

mayhem:
I wonder if this improves our memory leak situation -- even if it makes it much less that would be a great improvement over the current situation. Did you run the code Gentlecat?

alastair:
I'll do a test on safari next time I load my server up.

alastair:
This works, but doesn't affect the memory usage or safari's crashiness

mwiencek:
Yeah, not too surprising I guess, since the limiting factor is gonna be holding all the pages in memory (once for the request, and then again after it's parsed into the DOM, via `innerHTML`).

If all you were doing was creating these objects, you'd see an improvement, but compared to requesting/parsing the pages, this won't register too much.

I still think it's an improvement though. :D

alastair:
Do you have any recommendations on how to delete the data after we've used
it? Does JavaScript not GC when the variable goes out of scope?

mwiencek:
It should definitely GC variables that go out of scope. But I'm not sure how aggressive it is, and there might be variables being kept in-scope somehow. I don't see anything obviously wrong at a glance. If I can get (messy|listen)brainz-server running I'll look again, since I'm curious now...

PR #42, opened 2015-09-25
current-status page (LB-48)

mayhem:
This shows basic kafka stats and load average.


webserver/views/index.py
```diff
@@ -41,3 +45,33 @@ def faq():
 @index_bp.route("/roadmap")
 def roadmap():
     return render_template("index/roadmap.html")
+
+
+# listen-group    listens                        0   282202115       282202115       0               none
```
gentlecat: What's that?

alastair: the output of the java process

webserver/views/index.py
```diff
@@ -41,3 +45,33 @@ def faq():
...
+@index_bp.route("/current-status")
+def current_status():
+
+    load = "%.2f %.2f %.2f" % os.getloadavg()
+
+    process = subprocess.Popen([config.KAFKA_RUN_CLASS_BINARY, "kafka.tools.ConsumerOffsetChecker",
```
alastair: This is a good start, but I'm wary about spinning up all of java each time someone loads this page. Have you looked if there is a way of getting the data from python?

mayhem: I spent 2 hours looking for a way and then you beat me to it by finding the command line program. At this point I feel that:

1. this isn't going to be a popular page, so it isn't going to be loaded often.
2. the VM will already be resident in ram and we have plenty of ram still.
3. If we find a way to do this via the API, then I'm happy to move to it.

In the meantime, I think we should deploy this to get insights into our server going.

alastair: OK, fine by me for now then. If you want I'll merge it to my dev system and
test it.

alastair: do we want to rel=nofollow/robots.txt it so that search engines don't hit it?

mayhem: I'll make a task for adding a proper robots.txt, since we need it anyway.

http://tickets.musicbrainz.org/browse/LB-70

webserver/templates/index/current-status.html
```diff
@@ -0,0 +1,60 @@
...
+         {{ load }}
+      </div>
+
+    </div>
+
+    <div class="col-md-5 col-lg-4">
+
+      <p style="text-align: center">
+         <img style="border-radius: 15px;" 
+              src="https://upload.wikimedia.org/wikipedia/commons/4/4e/Macaca_nigra_self-portrait_large.jpg"
```
alastair: cool photo. Can we have one that's less than 1.7mb?

webserver/templates/index/current-status.html
```diff
@@ -0,0 +1,60 @@
...
+    <div class="col-md-7 col-lg-8">
+      <h3>Incoming listens</h3>
+
+      <p>
+        Stats for our kafka listens stream: 
+      </p>
+
+      <table class="table table-bordered table-striped">
+          <tr>
+              <td>offset</td>
```
alastair: are we happy just putting these names in, or should we describe them?

alastair:
Looks fine. Remember to add the line to the config file when releasing

webserver/templates/index/current-status.html
mayhem: wait, what? You're not going to tell me to take it out??

webserver/templates/index/current-status.html
mayhem: I tried to describe them last night, but wasn't able to find good ways of doing that. Partially because I don't know enough about kafka to do a good job. Good enough for now, I say.

webserver/templates/index/current-status.html
mayhem: Then we have to host the image ourselves. :( WP has no smaller images that I can see.

PR #43, opened 2015-09-27
Fix crashing in Safari (LB-23)

mwiencek:
The importer crashed every single time when I tested it in Safari 8.08, so I was able to reproduce this really easily.

With the changes here, it didn't crash at all after 8 or 9 trials.

This is clearly a bug in Safari, since there's nothing invalid about the code in question. Luckily, the code doesn't actually do anything, so we can just remove it.

Why the code does nothing:

`enqueueReport` pushes a single item onto `toReport`, then calls `dispatch()`, which passes that single item to `reportScrobbles` and immediately clears `toReport`.

So, there is never more than one item in `toReport`, and the code is equivalent to just calling `reportScrobbles` on the item directly.

If there used to be a purpose to that and you'd rather fix it than remove it, let me know and I'll try to fix it instead. :)

### DIFF
--- a/webserver/templates/user/scraper.js
+++ b/webserver/templates/user/scraper.js
@@ -112,11 +112,36 @@ function encodeScrobbles(root) {
     return structure;
 }
 
-function getLastFMPage(page, callback) {
+function getLastFMPage(page) {
+    function retry(reason) {
+        console.warn(reason + ' fetching last.fm page=' + page + ', retrying in 3s');
+        setTimeout(function () {
+            getLastFMPage(page);
+        }, 3000);
+    }
+
     var xhr = new XMLHttpRequest();
+    xhr.timeout = 10 * 1000; // 10 seconds
     xhr.open("GET", encodeURI("http://www.last.fm/user/{{ lastfm_username }}/library?page=" + page + "&_pjax=%23content"));
-    xhr.onload = function(content) {
-        callback(xhr.response);
+    xhr.onload = function () {
+        if (/^2/.test(this.status)) {
+            reportPageAndGetNext(this.response);
+        } else if (/^5/.test(this.status)) {
+            retry('got ' + this.status);
+        } else {
+            // ignore 40x
+            pageDone();
+        }
+    };
+    xhr.ontimeout = function () {
+        retry('timeout');
+    };
+    xhr.onabort = function () {
+        // this should never happen
+        pageDone();
+    };
+    xhr.onerror = function () {
+        retry('error');
     };
     xhr.send();
 }
@@ -136,14 +161,13 @@ var times5Error = 0;
 
 function reportScrobbles(struct) {
     if (!struct.payload.length) {
+        pageDone();
         return;
     }
 
     timesReportScrobbles++;
     //must have a trailing slash
     var reportingURL = "{{ base_url }}";
-    activeSubmissions++;
-
     var xhr = new XMLHttpRequest();
     xhr.open("POST", reportingURL);
     xhr.setRequestHeader("Authorization", "Token {{ user_token }}");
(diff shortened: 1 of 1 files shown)


mwiencek:
This doesn't affect memory usage from what I can see, I just had to fix this first to even look at that in Safari. After actually checking, Firefox is the only browser that uses an insane amount of memory for me (and doesn't release it).

alastair:
this is interesting. It was originally like you had it: https://github.com/metabrainz/listenbrainz-server/commit/e53aa5ca97c215cbaf815b6970fb2eeb8da247ac
I believe it was changed to the enqueue format in an attempt to get a processing queue, except since we use async xhr, the `dispatch` method always runs really quickly. 

I lowered the number of current queries by using `activeSubmissions` instead, but I still like the idea of having a list of perhaps 10 xmlhttpreq objects, which process, finished, get popped off, and then replaced with a new object with new payload. Does this make sense, or does the current system work well enough?

mwiencek:
Hm, well, JavaScript's event loop already acts as a queue, so I'm not sure I understand what the processing queue would do on top of that. `activeSubmissions` should do what you're saying, by limiting the number of concurrent submissions to 10.

I noticed there are some bugs in the way it handles `activeSubmissions`, though, since it regularly get incremented beyond 10. Because it gets incremented asynchronously, inside the `setTimeout`, more than 10 submissions have a chance to make it past the condition. I'll push a commit to fix that too.

webserver/templates/user/scraper.js
```diff
@@ -165,77 +175,76 @@ function reportScrobbles(struct) {
...
 }
 
-function getNextPageIfSlots() {
+function getNextPagesIfSlots() {
     // Get a new lastfm page and queue it only if there are more pages to download and we have
     // less than 10 pages waiting to submit
-    if (page <= numberOfPages && activeSubmissions < 10) {
+    while (page <= numberOfPages && activeSubmissions < 10) {
         page += 1;
-        setTimeout(function() { getLastFMPage(page, reportPageAndGetNext) }, 0 + Math.random()*100);
```
mwiencek: The `setTimeout` isn't needed because the XHR request is already asynchronous.

webserver/templates/user/scraper.js
```diff
@@ -165,77 +175,76 @@ function reportScrobbles(struct) {
...
-        enqueueReport(struct);
+        reportScrobbles(struct);
     };
-    xhr.onloadend = function(context) {
-        activeSubmissions--;
-    }
     xhr.send(JSON.stringify(struct));
 }
 
-function reportPage(response) {
```
mwiencek: This should never be called outside of `reportPageAndGetNext`, so I inlined it there.

webserver/templates/user/scraper.js
```diff
@@ -165,77 +175,76 @@ function reportScrobbles(struct) {
...
-        enqueueReport(struct);
+        reportScrobbles(struct);
     };
     xhr.onerror = function(context) {
         console.log("error, req'ing");
-        enqueueReport(struct);
+        reportScrobbles(struct);
     };
-    xhr.onloadend = function(context) {
-        activeSubmissions--;
```
mwiencek: This was problematic, because `loadend` doesn't fire until after the other handlers (so `activeSubmissions` would get checked again before it got decremented here). Anyway, `pageDone` is now the only place where this gets decremented.

webserver/templates/user/scraper.js
mwiencek: It's also what was causing more than 10 `activeSubmissions` at a time, since the code was incrementing `activeSubmissions` inside the `setTimeout`.

alastair: This was here because originally we didn't have a limit on the number of active submission. This was a random pause between paging between last.fm pages.   I wasn't really worried about having 11 or 12 activeSubmissions going, just as long as it was about 10 instead of about 100

alastair:
This looks good, merged. thanks!

PR #44, opened 2015-09-27
Fix the disappearance of `data` in Listen.from_json

mineo:
Listen.__init__ now correctly sets data to an empty dictionary if no
data was passed in and uses the passed in data if it's not None.


mayhem:
Thank you! I was far too exhausted to see such a simple thing at the time. :(

PR #45, opened 2015-09-27
Output the msid as part of a listen

mayhem:
Fixes an issue mentioned twice this morning in #metabrainz.

### DIFF
--- a/webserver/views/api.py
+++ b/webserver/views/api.py
@@ -85,7 +85,11 @@ def get_listens(user_id):
     )
     listen_data = []
     for listen in listens:
-        listen_data.append({ "track_metadata" : listen.data, "listened_at" : listen.timestamp })
+        listen_data.append({ 
+                             "track_metadata" : listen.data, 
+                             "listened_at" : listen.timestamp,
+                             "recording_msid" : listen.recording_msid,
+                           })
 
     return jsonify({'payload': {
         'user_id': user_id,


alastair:
```
2:38 PM <Freso> ruaok: Is there a reason for the recording_msid to be at a different level than the other _msid's?
2:38 PM <ruaok> it is the primary key
2:38 PM <ruaok> not additional metadata
2:38 PM <ruaok> it does feel odd, no?
2:40 PM <Freso> It looks odd anyway. :) I would either move the other MSIDs up or move that one down.
```

alastair:
I'm happy with having the recording_msid in a different place, especially since we treat it more like a primary key than metadata. This value is guaranteed to exist. Others (especially release msid) won't always be around.

mayhem:
That was the intent with which it was created and I don't really care to change that. So, :sheep:, :ship: or :shipit: or :pineapple:?

alastair:
I wanted to check 2 things: where we put this value in the listen object in the documentation, and if it worked on my VM. I'll approve it for :pineapple:ing once done

mayhem:
I placed it where the Google doc had it. And WTF is :pineapple: for?

alastair:
I think this is fine. I have no idea what :pineapple:s are for. But you can :ship: it if you want

mayhem:
:pineapple::pineapple::pineapple:!!!

PR #47, opened 2015-09-30
Ignore most of initialization when building docs on rtfd.org

gentlecat:


webserver/views/api.py
```diff
@@ -9,7 +10,10 @@
 from webserver.decorators import crossdomain
 import webserver
 import db.user
-import messybrainz
+
+on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
+if not on_rtd:
+    import messybrainz
```
gentlecat: This is kind of ugly, but I'm not sure if there's a better way to handle it.

alastair: I don't like it at all.

How about
```
try:
    import messybrainz
except ImportError:
    if not on_rtd:
        raise
```

gentlecat:
Builds should work now: https://readthedocs.org/projects/lb-test/builds/3364518/, https://lb-test.readthedocs.org/.

webserver/views/api.py
gentlecat: Yeah, I like this better. What about checking in `create_app`?

webserver/__init__.py
```diff
@@ -1,10 +1,6 @@
 from flask import Flask, current_app
 import sys
 import os
-import messybrainz
-import messybrainz.db
-
-_kafka = None
```
gentlecat: I removed this one because it's unused.

webserver/__init__.py
```diff
@@ -1,10 +1,6 @@
 from flask import Flask, current_app
 import sys
 import os
-import messybrainz
```
gentlecat: This import should be unnecessary.

gentlecat:
I didn't test if it works correctly in "normal" mode.

webserver/external/messybrainz.py
```diff
@@ -0,0 +1,20 @@
...
+    on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
+    if not on_rtd:
+        raise
+
+
+def init_db_connection(uri):
+    messybrainz.db.init_db_engine(uri)
+
+
+def submit_listens(listens):
```
alastair: :(

webserver/views/api.py
```diff
@@ -188,7 +189,7 @@ def _messybrainz_lookup(listens):
         msb_listens.append(messy_dict)
 
     try:
-        msb_responses = messybrainz.submit_listens_and_sing_me_a_sweet_song(msb_listens)
+        msb_responses = messybrainz.submit_listens(msb_listens)
```
alastair: this is one of the reasons that I like absolute package names - if here it said `webserver.external.messybrainz` it would be a lot more explicit about what we were calling. As it is, someone who glances at this code could think that this method is in the `messybrainz` package.

alastair:
:+1: Looks good

PR #48, opened 2015-10-03
Import ServiceUnavailable exception

alastair:
Fixes error reported in sentry


gentlecat:
:ship:  :bomb::bug::bomb:

PR #49, opened 2015-10-04
Validate submitted MBIDs

clinty:
Check to see if release_mbid, recording_mbid, or artist_mbids contain
any invalid UUIDs.

This should fix LB-74.

### DIFF
--- a/webserver/views/api.py
+++ b/webserver/views/api.py
@@ -2,6 +2,7 @@ import sys
 import urllib2
 import ujson
 import socket
+import uuid
 from flask import Blueprint, request, current_app, jsonify
 from werkzeug.exceptions import BadRequest, InternalServerError, Unauthorized, ServiceUnavailable
 from kafka import SimpleProducer
@@ -242,8 +243,8 @@ def _validate_listen(listen):
         _log_raise_400("JSON document does not contain a valid metadata.track_name "
                        "and/or track_metadata.artist_name.", listen)
 
-    # Tags
     if 'additional_info' in listen['track_metadata']:
+        # Tags
         if 'tags' in listen['track_metadata']['additional_info']:
             tags = listen['track_metadata']['additional_info']['tags']
             if len(tags) > MAX_TAGS_PER_LISTEN:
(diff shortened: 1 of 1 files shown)


webserver/views/api.py
```diff
@@ -253,7 +253,17 @@ def _validate_listen(listen):
                 if len(tag) > MAX_TAG_SIZE:
                     _log_raise_400("JSON document may not contain track_metadata.additional_info.tags "
                                    "longer than %d characters." % MAX_TAG_SIZE, listen)
-
+        # MBIDs
+        lmbid = listen['track_metadata']['additional_info'].get('release_mbid')
+        if not is_valid_uuid(lmbid):
+                _log_raise_400("Release MBID format invalid.")
```
alastair: Watch indent here

webserver/views/api.py
```diff
@@ -253,7 +253,17 @@ def _validate_listen(listen):
...
+        cmbid = listen['track_metadata']['additional_info'].get('recording_mbid')
+        if not is_valid_uuid(cmbid):
+                _log_raise_400("Recording MBID format invalid.")
+        ambids = listen['track_metadata']['additional_info'].get('artist_mbids')
+        for ambid in ambids:
```
alastair: If `artist_mbids` isn't in additional_info then .get will return None, use `.get('artist_mbids', [])` here

clinty:
better?

mineo:
This would fix [LB-74](http://tickets.musicbrainz.org/browse/LB-74).

AFAICT this would make the recording and release mbids mandatory, but not the artist ones.

webserver/views/api.py
```diff
@@ -253,7 +254,17 @@ def _validate_listen(listen):
                 if len(tag) > MAX_TAG_SIZE:
                     _log_raise_400("JSON document may not contain track_metadata.additional_info.tags "
                                    "longer than %d characters." % MAX_TAG_SIZE, listen)
-
+        # MBIDs
+        lmbid = listen['track_metadata']['additional_info'].get('release_mbid')
+        if not is_valid_uuid(lmbid):
```
gentlecat: It's probably better to check if this value is actually specified first (same with other MBID checks).

clinty:
like that?

gentlecat:
Looks good.

alastair:
Thanks. I made a small change to the dict getting (so that we only have the full key once), but otherwise it's fine.

19379:
this does not work as expected. when supplying an invalid MBID, it throws a 500 internal server error...

`_log_raise_400() takes exactly 2 arguments (1 given)`

alastair:
Yes, we noticed this (in our error logs). It will be fixed in the next
release

PR #53, opened 2015-10-11
Fix LB-76

mayhem:
Ensure that only one max_ts or min_ts can be used, not both.
Make both exclusive.

### DIFF
--- a/listenstore/listenstore/listenstore.py
+++ b/listenstore/listenstore/listenstore.py
@@ -233,7 +233,7 @@ class ListenStore(object):
 
             params = {'uid': uid,
                       'from_id': from_id, 'to_id': to_id,
-                      'limit': this_limit }
+                      'limit': this_limit}
 
             params.update(range_params(date_range))
 
--- a/listenstore/listenstore/listenstore.py
+++ b/listenstore/listenstore/listenstore.py
@@ -171,7 +171,7 @@ class ListenStore(object):
         """
 
         if from_id and to_id:
-            raise ValueError("You cannot specifcy from_id and to_id at the same time.")
+            raise ValueError("You cannot specify from_id and to_id at the same time.")
 
         if from_id:
             order = ORDER_ASC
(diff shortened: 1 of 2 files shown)


gentlecat:
Looks good to me. I guess docs and pagination will need to be updated. :sheep: :honeybee::honeybee::honeybee:

listenstore/listenstore/listenstore.py
```diff
@@ -207,7 +207,7 @@ def fetch_listens_for_range(self, uid, date_range, from_id, to_id, limit=None, o
         """
         query = """SELECT * FROM listens WHERE uid = %(uid)s AND """ + \
                 range_keys(len(date_range)) + \
-                """ AND id >= %(from_id)s AND id <= %(to_id)s
+                """ AND id > %(from_id)s AND id < %(to_id)s
```
alastair: This doesn't validate that min_ts/max_ts are exclusive. Does this matter, with the checks in the view?

webserver/views/api.py
```diff
@@ -75,12 +75,19 @@ def submit_listen():
...
+        _log_and_raise_400("You may only specify max_ts or min_ts, not both.")
+
     cassandra = webserver.create_cassandra()
     listens = cassandra.fetch_listens(
         user_id,
         limit=min(_parse_int_arg("count", DEFAULT_ITEMS_PER_GET), MAX_ITEMS_PER_GET),
-        from_id=_parse_int_arg("max_ts"),
-        to_id=_parse_int_arg("min_ts"),
+        from_id=min_ts,
+        to_id=max_ts,
```
alastair: does listenstore accept None as an argument? what's it's behaviour?

listenstore/listenstore/listenstore.py
alastair: in our user profile view, we call `listenstore.fetch_listens`, so we should have the "one of min or max" check there too

alastair: we can choose asc/desc as the order based on if we get passed a max ts or min ts

webserver/views/api.py
alastair: after looking at the code for `fetch_listens`, None will be converted into a upper or lower bound (beginning of last.fm, or 1 hour in the future from now). The count is used to quit

alastair:
:rocket:

PR #54, opened 2015-10-11
Always return listens in descending order.

mayhem:
:cactus: :monkey:


webserver/views/api.py
```diff
@@ -97,6 +97,9 @@ def get_listens(user_id):
                              "recording_msid" : listen.recording_msid,
                            })
 
+    if min_ts:
+        listen_data = listen_data[::-1]
```
alastair: cute

PR #56, opened 2015-10-13
Create connections to external services on each flask request

alastair:
Instead of having a single connection open for the life of the
server, make a new connection to kafka and listenstore on each
request (this was already the case for listenstore). Store
the connections in flask.g, as that's what it's designed for

A kafka threaded example does basically the same thing: https://github.com/mumrah/kafka-python/blob/master/example.py, creating a connection and producer in each thread. As far as I'm aware, this is about the same kind of thing we get with multiple requests in flask.

Renamed cassandra to listenstore, as it better reflects what we are connecting to. I removed the `from listenstore.listenstore import ...`, so we will then have to actually install listenstore into the virtualenv each time we update it. Either with `setup.py develop`, or we add an install command into `fab deploy`

I've still not tested it, therefore marking as WIP.

I'm unsure if any connections have to be closed after connections are shut down. There is a `producer.stop` method in kafka, so perhaps we should be calling this.

### DIFF
--- a/webserver/__init__.py
+++ b/webserver/__init__.py
@@ -1,10 +1,8 @@
-from flask import Flask, current_app, g
+from flask import Flask, current_app
 import sys
 import os
 import messybrainz
 import messybrainz.db
-import kafka_connection
-import listenstore_connection
 
 
 def create_app():
--- a/webserver/__init__.py
+++ b/webserver/__init__.py
@@ -1,20 +1,13 @@
-from flask import Flask, current_app
+from flask import Flask, current_app, g
 import sys
 import os
 import messybrainz
 import messybrainz.db
-
-_kafka = None
-
-
-def create_cassandra():
-    from cassandra_connection import init_cassandra_connection
-    return init_cassandra_connection(current_app.config['CASSANDRA_SERVER'], current_app.config['CASSANDRA_KEYSPACE'])
+import kafka_connection
+import listenstore_connection
 
 
 def create_app():
-    global _kafka
-
     app = Flask(__name__)
 
     # Configuration
@@ -27,14 +20,17 @@ def create_app():
     from webserver.loggers import init_loggers
     init_loggers(app)
 
-    # Kafka connection
-    from kafka_connection import init_kafka_connection
-    init_kafka_connection(app.config['KAFKA_CONNECT'])
-
     # Database connection
     from db import init_db_connection
     init_db_connection(app)
     messybrainz.db.init_db_engine(app.config['MESSYBRAINZ_SQLALCHEMY_DATABASE_URI'])
+
+    # Connections to external servers
+    @app.before_request
+    def before_reqeust():
+        g.kafka = kafka_connection.init_kafka_connection(app.config['KAFKA_CONNECT'])
+        g.listenstore = listenstore_connection.init_listenstore(current_app.config['CASSANDRA_SERVER'], current_app.config['CASSANDRA_KEYSPACE'])
+
 
     # OAuth
     from webserver.login import login_manager, provider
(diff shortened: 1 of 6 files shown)


webserver/__init__.py
```diff
@@ -27,15 +20,18 @@ def create_app():
...
     from db import init_db_connection
     init_db_connection(app)
     messybrainz.db.init_db_engine(app.config['MESSYBRAINZ_SQLALCHEMY_DATABASE_URI'])
 
+    # Connections to external servers
+    @app.before_request
+    def before_reqeust():
+        g.kafka = kafka_connection.init_kafka_connection(app.config['KAFKA_CONNECT'])
+        g.listenstore = listenstore_connection.init_listenstore(current_app.config['CASSANDRA_SERVER'], current_app.config['CASSANDRA_KEYSPACE'])
+
```
mayhem: Does this really make sense? So far only one of our endpoints talks to kafka, but this will require every connection to have a kafka handle. That's kinda kafkaesque, no? 

Maybe we should make this a decorator that we apply to kafka using endpoints.

alastair: Good point. The pattern I took this from was a database app which obviously needs the database on all pages.  I don't think we need to go as far as a decorator, just a method that can be called without arguments that gives us a kafka or listenstore handle.

alastair:
Now we have a method to get a connection handle from any method. It's only used in submission, api get, and profile get. I did a test import / run cassandra writer / read api / read profile page and all pages work.

mayhem:
This improved granularity is much better. :shell:

webserver/__init__.py
```diff
@@ -1,20 +1,13 @@
...
 import messybrainz.db
-
-_kafka = None
-
-
-def create_cassandra():
-    from cassandra_connection import init_cassandra_connection
-    return init_cassandra_connection(current_app.config['CASSANDRA_SERVER'], current_app.config['CASSANDRA_KEYSPACE'])
+import kafka_connection
+import listenstore_connection
```
gentlecat: Where are these two used?

webserver/__init__.py
```diff
@@ -1,20 +1,13 @@
-from flask import Flask, current_app
+from flask import Flask, current_app, g
```
gentlecat: `g` looks unused as well.

webserver/views/api.py
```diff
@@ -77,14 +78,15 @@ def submit_listen():
 @api_bp.route("/1/user/<user_id>/listens")
 def get_listens(user_id):
 
+    listenstore = listenstore_connection.get_listenstore()
```
gentlecat: Maybe it's better to move that closer to the place where it's used (line 89)? Or just get rid of variable completely?

webserver/listenstore_connection.py
```diff
@@ -0,0 +1,10 @@
+from listenstore import listenstore
+from flask import current_app
+
+def get_listenstore():
+    server = current_app.config['CASSANDRA_SERVER']
+    keyspace = current_app.config['CASSANDRA_KEYSPACE']
```
gentlecat: You can move these assignments directly to dictionary.

    {
        'cassandra_server': current_app.config['CASSANDRA_SERVER'],
        'cassandra_keyspace': current_app.config['CASSANDRA_KEYSPACE'],
    }

webserver/views/api.py
alastair: In the user profile method, the listenstore is created at the top, because it's used many times. I put this one here for symmetry, though perhaps you're right.

gentlecat: Oh, yeah. That makes sense too.

alastair:
It'd be good if someone could try and see if we need to call `producer.stop` as well. It looks like it might happen on GC (https://github.com/mumrah/kafka-python/blob/master/kafka/producer/base.py#L430-L432)
And also if there is a cassandra close method which we should call

gentlecat:
Is this still relevant?

alastair:
nope!

PR #60, opened 2015-10-13
LB-55, LB-64: Show options on completion of import

alastair:
http://tickets.musicbrainz.org/browse/LB-55
http://tickets.musicbrainz.org/browse/LB-64

Add some notes to the import page to remind users to not navigate from
the page, and to not be scrobbling while they import

After the importer finishes, show a link to close the dialog, and
another to go to the profile page on listenbrainz

lb-finished

### DIFF
--- a/webserver/templates/user/scraper.js
+++ b/webserver/templates/user/scraper.js
@@ -146,7 +146,7 @@ function getLastFMPage(page) {
     xhr.send();
 }
 
-var version = "1.3";
+var version = "1.4";
 var page = 1;
 var numberOfPages = parseInt(document.getElementsByClassName("pages")[0].innerHTML.trim().split(" ")[3]);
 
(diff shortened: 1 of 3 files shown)


mayhem:
:shaved_ice: 

I'm curious -- did the order in which we scrape the user profile change? It seems to go from oldest to newest, which isn't so good for doing an incremental import. You have to wait for the whole import to finish to get the latest listen.

alastair:
No? I don't think this is the case at all. Where do you see this happening?

webserver/templates/user/scraper.js
```diff
@@ -192,7 +192,13 @@ function reportScrobbles(struct) {
             pageDone();
         }
         if (numCompleted >= numberOfPages) {
-            updateMessage("<i class='fa fa-check'></i> Import finished<br><span style='font-size:8pt'>Thank you for using ListenBrainz</span>");
+            updateMessage("<i class='fa fa-check'></i> Import finished<br><span><a href='https://listenbrainz.org/user/{{user_id}}'>Go to your ListenBrainz profile</a> | <a href='' id='close-progress-container'>Close</a></span><br><span style='font-size:8pt'>Thank you for using ListenBrainz</span>");
```
gentlecat: How about an exclamation mark at the end of `Thank you for using ListenBrainz`?

alastair: no

gentlecat: :sob:

PR #62, opened 2015-10-17
User token reset

alastair:
http://tickets.musicbrainz.org/browse/LB-71

Only added simple reset. No history, no multiple tokens (not needed for now).

screen shot 2015-10-17 at 12 45 49 pm

---

screen shot 2015-10-17 at 12 45 56 pm

---

screen shot 2015-10-17 at 12 46 09 pm

### DIFF
--- a/db/user.py
+++ b/db/user.py
@@ -11,12 +11,17 @@ def create(musicbrainz_id):
     db.session.commit()
     return result.fetchone()["id"]
 
-def update_token(user_id):
-
-    result = db.session.execute("""INSERT INTO "user" (musicbrainz_id, auth_token)
-                                        VALUES (:mb_id, :token)
-                                     RETURNING id""",
-                                {"mb_id": musicbrainz_id, "token": str(uuid.uuid4())})
+def update_token(id):
+    """Update a user's token to a new UUID
+       Arguments: id - the row id of the user to update
+    """
+    query = """UPDATE "user"
+                  SET auth_token = :token
+                WHERE id = :id
+            """
+    result = db.session.execute(query,
+            {"token": str(uuid.uuid4()),
+             "id": id})
     db.session.commit()
 
 
--- a/db/user.py
+++ b/db/user.py
@@ -11,6 +11,14 @@ def create(musicbrainz_id):
     db.session.commit()
     return result.fetchone()["id"]
 
+def update_token(user_id):
+
+    result = db.session.execute("""INSERT INTO "user" (musicbrainz_id, auth_token)
+                                        VALUES (:mb_id, :token)
+                                     RETURNING id""",
+                                {"mb_id": musicbrainz_id, "token": str(uuid.uuid4())})
+    db.session.commit()
+
 
 def get(id):
     """Get user with a specified ID (integer)."""
--- a/webserver/templates/user/import.html
+++ b/webserver/templates/user/import.html
@@ -80,7 +80,7 @@
 
      <p>If you want to reset your token, click below</p>
       <p>
-      <span class="btn btn-warning btn-lg" style="width: 400px"><a href="{{ url_for("user.reset_token", user_id=user.musicbrainz_id) }}"
+      <span class="btn btn-info btn-lg" style="width: 200px"><a href="{{ url_for("user.reset_token", user_id=user.musicbrainz_id) }}"
          style="color: white;">Reset token</a></span>
       <p>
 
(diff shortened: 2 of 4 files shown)


db/user.py
```diff
@@ -11,6 +11,19 @@ def create(musicbrainz_id):
     db.session.commit()
     return result.fetchone()["id"]
 
+def update_token(id):
+    """Update a user's token to a new UUID
+       Arguments: id - the row id of the user to update
```
gentlecat: I'm usually writing list of arguments like this:
```
Args:
    id: The row of the user to update.
```

webserver/templates/user/import.html
```diff
@@ -78,6 +78,12 @@ <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
         {{ user.auth_token }}
      </div>
 
+     <p>If you want to reset your token, click below</p>
+      <p>
```
mayhem: Not sure if we care, but the indent here is off.

webserver/templates/user/resettoken.html
```diff
@@ -0,0 +1,16 @@
+{% extends 'base.html' %}
+
+{% block title %}Reset token - ListenBrainz{% endblock %}
+
+{% block content %}
+  <h3 class="page-title">Reset token</h3>
+  <p>Are you sure you want to reset your token? If you do, you will need
+  to change it any apps that you have added it to.</p>
+  <form method="post">
```
mayhem: I think this verbiage is a bit scant. How about?

Are you sure you want to reset your token?

If you do, you will need to replace your old token with the newly generated token in any of your applications that use your token. The old token will stop working immediately. Be sure to update the token immediately in your applications to ensure that you have no disruptions in recording your listens.

webserver/views/user.py
```diff
@@ -26,6 +27,17 @@ def lastfmscraper(user_id):
...
     return Response(scraper, content_type="text/javascript")
 
+@user_bp.route("/<user_id>/resettoken", methods=["GET", "POST"])
+@login_required
+def reset_token(user_id):
+    if request.method == "POST":
+        reset = request.form.get("reset")
+        if reset == "yes":
+            db.user.update_token(current_user.id)
+            flash.info("Access token reset")
```
mayhem: What happens if a user submits an invalid token? From the looks of it, it would quietly do nothing. Is that intended due to security?

alastair: What do you mean submit? During this reset process the user gets a new random token. If they continue submitting to the api with the old one they will get 403

webserver/templates/user/import.html
alastair: thank you for caring! 


(i'll fix it)

gentlecat: Not sure?! :angry:

mayhem: Put it this way: I feel like a dick for pointing this out.

webserver/views/user.py
mayhem: What happens when I make a POST call to /1/<user does not exist>/resettoken ?

alastair: It'll ask you to log in. And then it might reset your token. Note that the user_id isn't actually used, it uses `current_user`, which is based on who you are logged in as.  Perhaps we could remove `<user_id>`

gentlecat: I think `user_id` in the URL is kind of useless there. So it will just reset your token.

mayhem: If I call  /1/[user does not exist]/resettoken while I am logged in, it will reset *my* token. I agree the user_id should be removed.

webserver/templates/user/import.html
gentlecat: I just wish we had something like [gofmt](https://golang.org/cmd/gofmt/).

gentlecat:
Why did you decide to have a separate page for resetting instead of just doing this on the import page?

mayhem:
Also your PEP-190 compliance is all over the map. :(

alastair:
I did it on a new page because the import page already has submit logic for step 1/2 and I didn't want to complicate it, and I didn't want to pull in a js dependency to do it inline.

Ideally this should move to a settings page later, but we don't have one of them yet

gentlecat:
Yeah, I think it would make sense to have a separate (more general) page for that since tokens are also used in the API.

gentlecat:
Any updates on this? I can try and finish cleaning this up.

PR #65, opened 2015-10-21
[WIP] [KTHA] rate limiting amidointhisrite?

mayhem:
First cut of the rate limiting stuff seems to be working. The general idea:

1. Rate limiting is done on a per IP or per token basis. If the caller provides the Authorization header with their token, the user gets a better rate limit. If not, the user is rate limited based on IP.
2. Rate limiting is done on a windowing basis, were X requests are allowed in a given window of time. This information is returned in the X-RateLimit-* headers.
2.b. There is no 2.b
3. Rate limits are dynamic and stored in Redis. They can be set with the set_rate_limits.py program. 
4. If you want to try this PR, install redis and python redis by hand for now.
5. I've added a dummy endpoint for easy testing. Use with:

  curl -D - --header "Authorization: token [your token here]" http://10.1.2.3:8080/1/rate-limit-test


webserver/__init__.py
```diff
@@ -31,6 +33,9 @@ def create_app():
     from kafka_connection import init_kafka_connection
     init_kafka_connection(app.config['KAFKA_CONNECT'])
 
+    # Redis connection
+    app.redis = Redis()
```
gentlecat: Are you sure it's supposed to be assigned to the application like that?

webserver/views/api.py
```diff
@@ -22,6 +23,11 @@
 DEFAULT_ITEMS_PER_GET = 25
 MAX_ITEMS_PER_MESSYBRAINZ_LOOKUP = 10
 
+@api_bp.route("/1/rate-limit-test")
```
gentlecat: Why do we need this endpoint?

mayhem: For ease of testing. Once this is done, I'll remove it.

webserver/__init__.py
alastair: yes, I already asked before, and it's fine. this connection is threadsafe

gentlecat: I mean is it actually a good practice to assign things directly to applications?

mayhem: Wasn't there a problem in london related to this? I forget. 

However, as far as I understand you can assign things to the app until you handle the first request -- then you're not allowed anymore.

gentlecat: That seems a bit fragile.

alastair: the specific problem was with postgres connections being shared between threads, which wasn't good. We can use a `Redis` in the whole app with no problem. We might want to use a pattern like @gentlecat suggests, like https://github.com/metabrainz/acousticbrainz-server/blob/6d0099184f00843b3f9c13d94e27803e4d5a28dc/db/__init__.py#L16-L22

gentlecat: If it can be done in the way other connections are created, that seems like a better way. There will be a module that we can extend, and it will be consistent with other parts.

mayhem: OK, can we please focus on the other, more important, parts of this PR? I'll fix this, now move on!

webserver/__init__.py
```diff
@@ -46,6 +51,17 @@ def create_app():
     from webserver.errors import init_error_handlers
     init_error_handlers(app)
 
+    # Add rate limit headers to responses
+    @app.after_request
+    def inject_x_rate_headers(response):
```
gentlecat: I'd make this function more generic and move it into `rate_limiter` module. This way `create_app` will be somewhat cleaner and we can add new after-request tasks easily.

Something like this:
```python
from webserver import rate_limiter

@app.after_request
def after_request_callbacks(response):
    return rate_limiter.inject_headers(response)
```

gentlecat:
Don't forget to add Redis setup into initialization scripts in */admin*.

set_rate_limits.py
```diff
@@ -0,0 +1,51 @@
...
+from webserver.rate_limiter import RATELIMIT_PER_TOKEN_KEY 
+from webserver.rate_limiter import RATELIMIT_PER_IP_KEY
+from webserver.rate_limiter import RATELIMIT_WINDOW_KEY
+
+# Yes, I could use getoptgetargparsewtfbbw, but then I would spend 20 mimnutes re-learning the stupid syntax. 
+# Or, I could just do it myself in the space of seconds.
+# Also, I tried to integrate this script with manage.py, but then I ended up wasting an hour trying
+# to figure out how to do this. So, we have this script. If you want to see it part of manage.py, you'll
+# have to do it.
+
```
mayhem: Even Gentlecat agreed with me.

gentlecat: Yeah, if it's not an easy thing to do, let's just skip it.

gentlecat:
Seems to be working apart from some off-by-one error that we noticed.

gentlecat:
:ship:?

mayhem:
Not ready yet -- I need to get the rate limiter supported in the scraper and I have to somehow get redis installed on the VM.

gentlecat:
Don't forget to remove endpoint for testing. Documentation probably should be updated as well. 🌯

mayhem:
Already deleted, but not pushed yet.

webserver/templates/user/scraper.js
```diff
@@ -174,28 +174,47 @@ function reportScrobbles(struct) {
...
+        else
+            rate_limit_delay = Math.max(0, (reset - current) * 1000);
+
+        console.log("remain: " + remain);
+        console.log("t left: " + (reset - current));
+        console.log("delay " + rate_limit_delay);
+
+        rate_limit_delay = 20000;
+
+        setTimeout(function() {
```
mayhem: @gentlecat: Any idea why the setTimeout() function doesn't actually delay at all?

mayhem: > rate_limit_delay = 20000;

This is there to very clearly observe a 20s timeout, which doesn't happen. :(

webserver/templates/user/scraper.js
```diff
@@ -174,28 +174,47 @@ function reportScrobbles(struct) {
...
+        if (remain)
+            rate_limit_delay = Math.max(0, Math.floor((reset - current) * 1000 / remain));
+        else
+            rate_limit_delay = Math.max(0, (reset - current) * 1000);
+
+        console.log("remain: " + remain);
+        console.log("t left: " + (reset - current));
+        console.log("delay " + rate_limit_delay);
+
+        rate_limit_delay = 20000;
```
gentlecat: `var rate_limit_delay = 20000;`

webserver/templates/user/scraper.js
```diff
@@ -174,28 +174,47 @@ function reportScrobbles(struct) {
...
-            updateMessage("<i class='fa fa-cog fa-spin'></i> Sending page " + numCompleted + " of " + numberOfPages + " to ListenBrainz<br><span style='font-size:8pt'>Please don't navigate while this is running</span>");
-        }
+        remain = parseInt(xhr.getResponseHeader("X-RateLimit-Remaining"));
+        reset = parseInt(xhr.getResponseHeader("X-RateLimit-Reset"));
+        current = new Date().getTime() / 1000;
+
+        if (remain)
+            rate_limit_delay = Math.max(0, Math.floor((reset - current) * 1000 / remain));
+        else
+            rate_limit_delay = Math.max(0, (reset - current) * 1000);
```
gentlecat:
```javascript
if (remain) {
    rate_limit_delay = Math.max(0, Math.floor((reset - current) * 1000 / remain));
} else {
    rate_limit_delay = Math.max(0, (reset - current) * 1000);
}
```

webserver/templates/user/scraper.js
gentlecat: I don't understand why you overwrite this value after all that checking above?

webserver/templates/user/scraper.js
```diff
@@ -174,28 +174,47 @@ function reportScrobbles(struct) {
...
-            pageDone();
-        }
-        if (numCompleted >= numberOfPages) {
-            updateMessage("<i class='fa fa-check'></i> Import finished<br><span style='font-size:8pt'>Thank you for using ListenBrainz</span>");
-        } else {
-            updateMessage("<i class='fa fa-cog fa-spin'></i> Sending page " + numCompleted + " of " + numberOfPages + " to ListenBrainz<br><span style='font-size:8pt'>Please don't navigate while this is running</span>");
-        }
+        remain = parseInt(xhr.getResponseHeader("X-RateLimit-Remaining"));
+        reset = parseInt(xhr.getResponseHeader("X-RateLimit-Reset"));
+        current = new Date().getTime() / 1000;
```
gentlecat: `var`

webserver/templates/user/scraper.js
```diff
@@ -174,28 +174,47 @@ function reportScrobbles(struct) {
...
+            rate_limit_delay = Math.max(0, (reset - current) * 1000);
+
+        console.log("remain: " + remain);
+        console.log("t left: " + (reset - current));
+        console.log("delay " + rate_limit_delay);
+
+        rate_limit_delay = 20000;
+
+        setTimeout(function() {
+            if (this.status >= 200 && this.status < 300) {
```
mwiencek: You have to change `this` to `xhr`, because `setTimeout` callbacks run in the window context,. Right now it's checking `window.status`.

webserver/templates/user/scraper.js
mayhem: > This is there to very clearly observe a 20s timeout, which doesn't happen. :(

mwiencek: Note that https://github.com/metabrainz/listenbrainz-server/pull/65/files#diff-f4bef214d1312fd528a78a3d7726f25fR248 will still perform 10 requests all at once, if it can, regardless of the rate limit.

mayhem:
@mwiencek You were totally right on your comment about it still making requests too fast. Javascript flow control and this scraper's flow control still baffle me. Can you please take a look at this and see if you can make it work as intended? 

Please note that the last commit incorporates a suggestion from Sam, but that code doesn't even seem to compile. All I get is a cryptic error message from some cryptic JS universe. 

HALP?

mwiencek:
Ok, I finally got my LB instance running again, I'll have a look.

hughrawlinson:
Ok, these two commits have added a new object to handle submission to Listenbrainz which manages the rate-limit, and nicely generalises the code for use in other service connectors. I would quite like to separate this out into it's own file so that we don't have to copy it to each of the service connectors (as well as split up the rest of the code a bit), but I'm not totally familiar with best practices for integrating JS build systems into Python projects, so I'll look into that later on at some point.

btw: Hi everyone! :)

webserver/__init__.py
```diff
@@ -31,6 +28,10 @@ def create_app():
     from kafka_connection import init_kafka_connection
     init_kafka_connection(app.config['KAFKA_CONNECT'])
 
+    # Redis connection
+    from redis_connection import init_redis_connection, _redis
```
gentlecat: Why is `_redis` imported there?

webserver/rate_limiter.py
```diff
@@ -0,0 +1,150 @@
...
+# http://flask.pocoo.org/snippets/70/
+#
+import time
+from functools import update_wrapper
+from flask import request, g
+from webserver.redis_connection import _redis
+import db.user
+
+#TODO
+# Add rate limit support to scraper
```
gentlecat: Is that done?

gentlecat:
Python looks fine :snake:

mayhem:
I hope the pyhon looks fine, since that hasn't change since the last commit. It is all about JS in this case. :)

webserver/rate_limiter.py
mayhem: Yeah, that comment is spurious now. I'll remove it when it is time to merge.

webserver/templates/user/scraper.js
```diff
@@ -1,256 +1,296 @@
...
+        xhr.setRequestHeader("Content-Type", "application/json;charset=UTF-8");
+        xhr.timeout = 10 * 1000; // 10 seconds
+        xhr.onload = function(content) {
+            if (xhr.status >= 200 && xhr.status < 300) {
+                console.log("successfully reported page");
+                completed(parseInt(xhr.getResponseHeader("X-RateLimit-Remaining")),parseInt(xhr.getResponseHeader("X-RateLimit-Reset")));
+            } else if (xhr.status == 429) {
+                console.log("received rate limit http error " + xhr.status + " req'ing");
+            } else if (xhr.status >= 400 && xhr.status < 500) {
+                times4Error++;
```
mwiencek: This variable is no longer defined, so this throws an error.

webserver/templates/user/scraper.js
```diff
@@ -1,256 +1,296 @@
...
+                // We mark 4xx errors as completed because we don't
+                // retry them
+                console.log("4xx error, skipping");
+            } else if (xhr.status >= 500) {
+                console.log("received http error " + xhr.status + " req'ing");
+                times5Error++;
```
mwiencek: Same, no longer defined.

webserver/templates/user/scraper.js
```diff
@@ -1,256 +1,296 @@
...
+            }
+        };
+
+        self.done = function(){
+            self.doneSubmissions = true;
+        };
+
+        self.processReq = function(){
+            self.processingReq = true;
+            struct = self.queue.pop();
```
mwiencek: Should be `var struct =`. Right now it's setting a global, so `setTimeout` calls that haven't run yet are reading a different value (possibly `undefined` if there are no items left).

mwiencek: Probably the reason I'm getting `TypeError: Cannot read property 'payload' of undefined at reportScrobbles` while testing. :)

mwiencek:
Worked fine importing a 1000 page library after making the changes I mentioned. Slower than it used to be, which means the rate limiting must be working. :)

mwiencek:
If I set the rate limit to something slow like `./set_rate_limits.py 1 1 5` I get a lot of 429 errors in the scraper. So either the server is returning a bad `reset` or the scraper is calculating things wrong, I guess? But I don't have time to debug this right now.

mayhem:
That's fine -- I can jump in an take a look at this particular issue. Can you please push the changes you've made to this branch?

once that is pushed then this 429 error is the only outstanding thing you see that needs fixing?

mwiencek:
Done. (Forgot I could push stuff here. :)) The 429 errors are the only other issue I saw while testing.

webserver/__init__.py
gentlecat: Still unused.

webserver/rate_limiter.py
```diff
@@ -0,0 +1,150 @@
...
+def get_rate_limit_data(request):
+    '''Fetch key for the given request. If an Authorization header is provided, 
+       the caller will get a better and personalized rate limit. If no header is provided,
+       the caller will be rate limited by IP, which gets an overall lower rate limit.
+       This should encourage callers to always provide the Authorization token
+    '''
+
+    auth_token = request.headers.get('Authorization')
+    if auth_token:
+        auth_token = auth_token[6:]
```
gentlecat: Why only part of it is used here?

alastair: Because the header is "Authorization: Token <thetoken>"
Perhaps it'd be better to split on space, or ensure that the text `Token ` exists

mayhem: ha, this code has been there since day one. works for now.

mayhem:
I pushed a few clean-ups. What else?

webserver/rate_limiter.py
gentlecat: I'd say `auth_header` would be a less confusing variable name to use above. And then `auth_token` on this line and further when it's the actual token inside.

PR #66, opened 2015-10-26
Fix query parameter name in api.py documentation

LordSputnik:
The limiting parameter seems to be "count", rather than "limit".


mayhem:
Thanks!

PR #67, opened 2015-12-07
Update README.md

warpr:
This fixes the link to the API documentation.  The previous URL at rtfd.org was giving me an SSL error.


gentlecat:
LGTM

PR #68, opened 2016-03-04
Fix 'document.getElementByClassName(...) is null'

pinkeshbadjatiya:
When the no of listens in library are less than maximum in a page, then the library is not divided into pages. Hence, class `pages` is not there.
Thus, <code>TypeError: document.getElementsByClassName(...)[0] is undefined on line 151 </code>

### DIFF
--- a/webserver/templates/user/scraper.js
+++ b/webserver/templates/user/scraper.js
@@ -149,7 +149,7 @@ function getLastFMPage(page) {
 var version = "1.4";
 var page = 1;
 var numberOfPages = 1;
-if (document.querySelector('.pages') != null) {
+if (document.getElementsByClassName("pages").length > 0) {
     numberOfPages = parseInt(document.getElementsByClassName("pages")[0].innerHTML.trim().split(" ")[3]);
 }
 
--- a/webserver/templates/user/scraper.js
+++ b/webserver/templates/user/scraper.js
@@ -149,8 +149,9 @@ function getLastFMPage(page) {
 var version = "1.4";
 var page = 1;
 var numberOfPages = 1;
-if (document.getElementsByClassName("pages").length > 0) {
-    numberOfPages = parseInt(document.getElementsByClassName("pages")[0].innerHTML.trim().split(" ")[3]);
+var pages = document.getElementsByClassName("pages");
+if (pages.length > 0) {
+    numberOfPages = parseInt(pages[0].innerHTML.trim().split(" ")[3]);
 }
 
 var toReport = [];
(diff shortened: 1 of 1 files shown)


gentlecat:
LGTM

webserver/templates/user/scraper.js
```diff
@@ -148,7 +148,10 @@ function getLastFMPage(page) {
 
 var version = "1.4";
 var page = 1;
-var numberOfPages = parseInt(document.getElementsByClassName("pages")[0].innerHTML.trim().split(" ")[3]);
+var numberOfPages = 1;
+if (document.querySelector('.pages') != null) {
```
alastair: I'm not sure about the difference between these two usages - what does `document.getElementsByClassName` and `document.querySelector` do differently? My preference would be to use the same one in both places. e.g. you could do 
```javascript
var pages = document.getElementsByClassName("pages");
if (pages !== null) {
  numberPages = pages.something.something;
}
```

alastair: do you need to use `===`/`!==` here when comparing to null? I'm not familiar with the difference

gentlecat: https://dorey.github.io/JavaScript-Equality-Table/

gentlecat: `querySelector` returns `null` if no elements are found. And returns only the first one.

alastair: @pinkeshbadjatiya Please change this to use either `querySelector` or `getElementsByClassName`, not both, and to use `===` equality. Otherwise it's a good patch and we can merge it.

pinkeshbadjatiya:
Made the required changes.

webserver/templates/user/scraper.js
```diff
@@ -148,7 +148,10 @@ function getLastFMPage(page) {
 
 var version = "1.4";
 var page = 1;
-var numberOfPages = parseInt(document.getElementsByClassName("pages")[0].innerHTML.trim().split(" ")[3]);
+var numberOfPages = 1;
+if (document.getElementsByClassName("pages").length > 0) {
+    numberOfPages = parseInt(document.getElementsByClassName("pages")[0].innerHTML.trim().split(" ")[3]);
```
gentlecat: You are doing the sam thing twice: `document.getElementsByClassName("pages")`. Maybe it's worth assigning it to a variable?

pinkeshbadjatiya: It is the same reason why i used <code> queryselector </code>, as it just returned after fetching the first element in the container.

alastair:
Looks good, thanks. I guess this shows that none of us have new last.fm accounts!

PR #69, opened 2016-03-06
Fix [LB-52] [LB-66] [LB-68] - API implemented

pinkeshbadjatiya:
The scrapper is completely built using the audioscrobbler api.

BUG, Line-236: the scrapper was skipping the 1st page because of variable increment <code>page += 1 </code>before function call.

[LB-68] - importer uses lots of ram
GC takes its own time to clean up the xhr variables in closure, which is slower than the amount at which requests were happening. By manually deleting the reference, ram usage is reduced. (Not completely sure, but any comments appreciated)

[LB-66] - import occasionally failing to parse page
Since the scrapper is now completely based on audioscrobbler api, so users does not even need to open the lastfm page. :)

[LB-52] -  last.fm API has better timestamps that last.fm web page
The timestamp issue is covered as the data fetch is done using API, giving accurate timestamps

----
- The<b> LASTFM_API_KEY</b> needs to be changed appropriately


alastair:
This is a great patch, thanks for submitting it. The memory leak especially is a good find.

There are some changes that I think you need to make before we merge this:

 * It is common to use a separate git commit for each independent change that you work on. For example, your change to the page number increment should be separate to the `delete` fix to free up memory.
* You committed a lastfm api key. You should not put private data like this in a git repository. Instead, it could be a value in `config.py`, and you can use template substitution to insert it into the js file.
* For now I think it's a good idea if we keep both the website scraper, and the new api scraper. Can you put this in a different file and create a new website flow to get to it? E.g. on the import page people could choose if they want to use the API or the website scrape.
* In some cases you commented out code, e.g. `//      //Currently does not support spotify`. When you remove things like this you can just delete them completely. We have the history of these files in git, so we do not need to keep the code commented out.

I recommend you create two new pull requests. One to fix some issues in the existing web scraper, and one more to add the new API scraper. The changes look good though, so we should be able to get them merged soon.

gentlecat:
I'm not sure if we should inject stuff into JS using template system. This will be a mess.

pinkeshbadjatiya:
Done as suggested. PR can be closed now.

alastair:
This has been split into different PRs, can be closed

PR #71, opened 2016-03-07
Fix page counter

pinkeshbadjatiya:
The page count was increasing before the call to <code>getLastFMPage()</code>, skipping the first page.


pinkeshbadjatiya:
Ready for merge. :)

alastair:
good catch! thanks

PR #73, opened 2016-03-08
Fix [LB-52] [LB-66] API Scraper

pinkeshbadjatiya:
Screenshot

<b>[LB-66]</b> - import occasionally failing to parse page
Since the scrapper is now completely based on audioscrobbler api, so users does not even need to open the lastfm page. :)

<b>[LB-52]</b> - last.fm API has better timestamps that last.fm web page
The timestamp issue is covered as the data fetch is done using API, giving accurate timestamps


pinkeshbadjatiya:
Made the required changes, as suggested.

alastair:
Why did you make this as a new pull request? If you make a new commit on the branch for your existing pull request then the changes will be listed on that pull request. We can use it to track the changes that you make to the patch and we can keep all of the discussion in one place. Please try and move these changes to the other pull request.

pinkeshbadjatiya:
You asked me in https://github.com/metabrainz/listenbrainz-server/pull/69 to create a new pull request for each change i do, so i thought a clean start of each topic would be a good idea. Should i created a new one for "API Scraper". Should i close this one?

pinkeshbadjatiya:
Merged the changes in #69

pinkeshbadjatiya:
Merged the changes in #69

pinkeshbadjatiya:
Errors related to lastfm api handled.
Screenshot

alastair:
Oh, I'm sorry - I thought that you had already made another pull request with the API scraper, and then you also opened this new one. Looking again at this history I can see that you didn't. That was my mistake for not checking carefully. I will close #69 and we can work on this PR instead.

alastair:
I see you only have 1 commit on this branch, with an unusual commit message. How did you make the commit? You should not be amending commits or performing force pushes when you add work to pull requests.
Please also read this blog post containing suggestions about how to write git log messages: http://chris.beams.io/posts/git-commit/

pinkeshbadjatiya:
Ready for merge. :)

webserver/templates/user/scraper_api.js
```diff
@@ -0,0 +1,309 @@
...
+        pageDone();
+    };
+    xhr.onerror = function () {
+        retry('error');
+    };
+    xhr.send();
+    xhr = null;
+}
+
+function reportScrobbles(struct) {
```
alastair: Is this the same in the api file and the page scraping one? We should abstract it out in to a utilities file so that we can use it in both without code reuse.
@gentlecat has some js compilation tools in acousticbrainz. If possible we could include these tools here too.

webserver/templates/user/scraper_api.js
```diff
@@ -0,0 +1,309 @@
...
+    xhr.send();
+    xhr = null;
+}
+
+var version = "1.4";
+var page = 1;
+var numberOfPages = 1;
+if (document.querySelector('.pages') != null) {
+   numberOfPages = parseInt(document.getElementsByClassName("pages")[0].innerHTML.trim().split(" ")[3]);
+}
```
alastair: This is from the web scraping file. Is it needed here?

webserver/templates/user/scraper_api.js
```diff
@@ -0,0 +1,309 @@
...
+        pageDone();
+    };
+    xhr.onerror = function () {
+        retry('error');
+    };
+    xhr.send();
+    xhr = null;
+}
+
+var version = "1.4";
```
alastair: We should probably start with version 1.0 again!

webserver/templates/user/scraper_api.js
```diff
@@ -0,0 +1,309 @@
...
+};
+
+function extractlastFMIDFromURL(loveButtonURL) {
+    var parts = loveButtonURL.split("/");
+    return parts.slice(0, parts.length-2).join("/");
+}
+
+function encodeScrobbles(jsonstr) {
+    var scrobbles = JSON.parse(jsonstr);
+    scrobbles = scrobbles['recenttracks']['track'];
```
alastair: the API includes much richer data - in particular, it will include a recording MBID if the user's client submitted it. We should include that here

webserver/templates/user/scraper_api.js
```diff
@@ -0,0 +1,309 @@
...
+var times5Error = 0;
+
+function getPageCount() {
+    function getCountAgain(reason) {
+        console.warn(reason + ' fetching last.fm pagecount, retrying in 3s');
+        setTimeout(function () {
+            getPageCount();
+        }, 3000);
+    }
+    var url = "http://ws.audioscrobbler.com/2.0/?method=user.getrecenttracks&user={{ lastfm_username }}&api_key={{ lastfm_api_key }}&format=json&page=" + page;
```
alastair: This function does almost exactly the same as `getLastFMPage`, but it's just used to get the number of pages in the API. Instead, we could combine it into the other query. We can load the first page, and then when it has loaded, read the scrobbles from this page and also the number of pages available.

pinkeshbadjatiya: yeah. I thought of the same. But using a separate function for separate tasks would be a good idea. And extracting the page numbers from the first call would require use of some flag variables which i tried to avoid.

webserver/templates/user/scraper_api.js
pinkeshbadjatiya: I missed it. will do it asap.

webserver/templates/user/scraper_api.js
pinkeshbadjatiya: i forgot to remove these lines.

webserver/templates/user/scraper_api.js
pinkeshbadjatiya: But would it not be a good idea to keep the two separate so that we may continue work on them separately?

webserver/templates/user/scraper_api.js
pinkeshbadjatiya: added in new commit.

pinkeshbadjatiya:
Updated a lot of things. Review needed.

alastair:
We think this is ready. I'll look at merging it. We will replace the html scrape version with this one, though we should keep that js file around in case at some later point we want to go back to it.
Two things to clean up:
 * Factor out duplicated code between html and api tool (interface, submission, Scrobble model)
 * Remove duplicated code or clean up api calls in getNumPages

mayhem:
+1 to what Alastair said. 

Also, given that the scraping method produces worse quality data, I would like to see the workflow improved so that we do NOT give users the option to scrape. Instead, everyone should always use the API import method.

webserver/templates/user/import.html
```diff
@@ -7,22 +7,34 @@ <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
...
+					<a href="{{loader_parse}}" style="color: white;">
+						<p>
+							<span class="btn btn-warning btn-lg" style="width: 400px"> Bookmarklet </span>
+						</p>
+					</a>
         <p>
           <span style="color:red">IMPORTANT</span>: If you've saved this bookmarklet before 29 Sept, please delete the old
           bookmarklet and save it again, since we've moved to https.
         </p>
+					<p style="font-weight:bold;"> or </p>
```
mayhem: As i mentioned in the main comment thread -- we should remove the "scraping" option from the HTML pages -- it will only confuse people and it produces bad data.

webserver/templates/user/scraper_api.js
```diff
@@ -0,0 +1,297 @@
...
+    xhr = null;
+}
+
+function reportScrobbles(struct) {
+    if (!struct.payload.length) {
+        pageDone();
+        return;
+    }
+
+    timesReportScrobbles++;
```
mayhem: One more thing that needs doing here. You need to add rate-limiting to the JS scraper. The API now emits rate limiting headers that tell the client how many requests they can make in a given time window. The scraper needs to pause for the right amount of time in order to not overwhelm the server. See how this is done in the old scraper:

https://github.com/metabrainz/listenbrainz-server/blob/master/webserver/templates/user/scraper.js#L224

You should be able to copy the relevant bits from that JS code.

webserver/templates/user/import.html
pinkeshbadjatiya: done!

webserver/templates/user/scraper_api.js
pinkeshbadjatiya: done!

PR #75, opened 2016-03-23
Fix [LB-84] | Listenstore

tallicamike:
Add validation for count (to be strictly positive) in Listenstore
when fetching from Cassandra.

### DIFF
--- a/listenstore/listenstore/listenstore.py
+++ b/listenstore/listenstore/listenstore.py
@@ -218,6 +218,8 @@ class ListenStore(object):
             This method will limit the amount of rows it fetches with one query,
             issuing multiple queries if the number of rows exceeds self.MAX_FETCH.
         """
+        if  limit is not None and limit <= 0:
+            return
         query = """SELECT * FROM listens WHERE uid = %(uid)s AND """ + \
                 range_keys(len(date_range)) + \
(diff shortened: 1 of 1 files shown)


mayhem:
Hmm. This very trivial PR is not very pythonic. Can you please make it more pythonic?

alastair:
Hi!
Thanks for submitting this PR. In the future when making PRs please also link to the ticket on jira in your original description.

The ticket didn't describe the problem clearly, but I think we should perform the validation differently. The _API_ in the ticket title refers to the web api exposed by the listenbrainz server, not the listenstore.
Here are some suggestions to changes that you could make:
* We already have some verification of the `count` parameter in the `webserver/views/api:get_listens` method, where we check if it's a number, and choose a default value if it isn't set. At this point we should check that it is > 0 and if not, return an HTTP 400 message (`BadRequest` exception) with an error message
* We can also add an extra check to listenstore, but this should be in the main `fetch_listens` method rather than the helper method `fetch_listens_for_range` as you added it. If you see at the top of `fetch_listens`, we already have some checks of `from_id` and `to_id`, which raises `ValueError` if there is a problem. We should check the `limit` parameter here and raise a similar `ValueError` if it's not a good value.

mayhem:
No follow up. Seems like a GsoC trying to get friendly with us PR. Closing.

PR #76, opened 2016-03-25
Fix [LB-41] - Upload lastfm backup

pinkeshbadjatiya:
Fix  	<a href="http://tickets.musicbrainz.org/browse/LB-41">#LB-41</a>
- Upload lastfm backup `json` file
- Supports uploading **.ZIP** files
- ZIP has to be in the same format as downloaded from lastfm.
- Sample file for testing - 
[21192540-QD3jC21IfeW40cZFN7AJ5AY7upcq49FR.zip](https://github.com/metabrainz/listenbrainz-server/files/294122/21192540-QD3jC21IfeW40cZFN7AJ5AY7upcq49FR.zip)

- Screenshots
Screenshot
<hr/>
image
<hr/>
Screenshot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ config.py
 *.log
 .cache
 env
-lastfm-backup-upload
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ config.py
 *.log
 .cache
 env
+lastfm-backup-upload
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 *.pyc
 config.py
 *.log
+.cache
+env
--- a/config.py.sample
+++ b/config.py.sample
@@ -56,4 +56,8 @@ COMPILE_LESS = True
 
 # MAX file size to be allowed for the lastfm-backup import, default is infinite
 # Size is in bytes
-MAX_CONTENT_LENGTH = 2 * 1024
+MAX_CONTENT_LENGTH = 16 * 1024 * 1024  # 16MB
+
+# Specify the upload folder where all the lastfm-backup will be stored
+# The path must be absolute path
+UPLOAD_FOLDER = "METABRAINZ/listenbrainz-server/lastfm-backup-upload"
--- a/config.py.sample
+++ b/config.py.sample
@@ -53,3 +53,7 @@ LASTFM_API_KEY = "USE_LASTFM_API_KEY"
 
 # Set to True if Less should be compiled in browser. Set to False if styling is pre-compiled.
 COMPILE_LESS = True
+
+# MAX file size to be allowed for the lastfm-backup import, default is infinite
+# Size is in bytes
+MAX_CONTENT_LENGTH = 2 * 1024
--- a/config.py.sample
+++ b/config.py.sample
@@ -60,4 +60,4 @@ MAX_CONTENT_LENGTH = 16 * 1024 * 1024  # 16MB
 
 # Specify the upload folder where all the lastfm-backup will be stored
 # The path must be absolute path
-UPLOAD_FOLDER = "METABRAINZ/listenbrainz-server/lastfm-backup-upload"
+UPLOAD_FOLDER = "/tmp/lastfm-backup-upload"
--- a/webserver/templates/user/import.html
+++ b/webserver/templates/user/import.html
@@ -57,7 +57,7 @@
 
      <p><b> or </b></p>
      <p>Import your listens from the lastfm backup file.</p>
-     <div>
+     <div class="well">
          <form action="upload" method="POST" enctype=multipart/form-data>
              <input type=file name=file>
              <input type=submit value=Upload>
--- a/webserver/templates/user/import.html
+++ b/webserver/templates/user/import.html
@@ -55,6 +55,13 @@
         {{ user.auth_token }}
      </div>
 
+     <p><b> or </b></p>
+     <p>Import your listens from the lastfm backup file.</p>
+     <div>
+         <form action="upload" method="POST" enctype=multipart/form-data>
+             <input type=file name=file>
+             <input type=submit value=Upload>
+         </form>
+     </div>
   {% endif %}
-
 {% endblock %}
--- a/webserver/templates/user/import.html
+++ b/webserver/templates/user/import.html
@@ -56,7 +56,7 @@
      </div>
 
      <p><b> or </b></p>
-     <p>Import your listens from the lastfm backup file.</p>
+     <p>Import your listens from a Last.fm backup file (.zip)</p>
      <div class="well">
          <form action="upload" method="POST" enctype=multipart/form-data>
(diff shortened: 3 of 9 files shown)


pinkeshbadjatiya:
Ready for review :)

pinkeshbadjatiya:
Ready for review :)

webserver/views/user.py
```diff
@@ -98,6 +99,63 @@ def import_data():
...
 
+@user_bp.route("/upload", methods=['POST'])
+@login_required
+def upload():
+    if request.method == 'POST':
+        f = request.files['file']
+        if f:
+            augmented_listens = []
+            msb_listens = []
+            try:
```
alastair: The data file that I sent you isn't actually the whole archive file. It is in fact a zip, containing many csvs and many json files. I've attached an example.
[21192540-QD3jC21IfeW40cZFN7AJ5AY7upcq49FR.zip](https://github.com/metabrainz/listenbrainz-server/files/204213/21192540-QD3jC21IfeW40cZFN7AJ5AY7upcq49FR.zip)
We should import the json data.

webserver/views/user.py
```diff
@@ -98,6 +99,63 @@ def import_data():
...
+                    msb_listens = []
+
+            if msb_listens:
+                augmented_listens.extend(_messybrainz_lookup(msb_listens))
+
+            _send_listens_to_kafka("import", augmented_listens)
+    return render_template('user/import.html', user=current_user)
+
+
+def _convert_to_native_format(data):
```
alastair: This is a great example of a function which should be tested!

webserver/views/user.py
```diff
@@ -98,6 +99,63 @@ def import_data():
...
+                msb_listens.append(dic)
+                if len(msb_listens) >= MAX_ITEMS_PER_MESSYBRAINZ_LOOKUP:
+                    augmented_listens.extend(_messybrainz_lookup(msb_listens))
+                    msb_listens = []
+
+            if msb_listens:
+                augmented_listens.extend(_messybrainz_lookup(msb_listens))
+
+            _send_listens_to_kafka("import", augmented_listens)
+    return render_template('user/import.html', user=current_user)
```
alastair: When this page is rendered on a `POST` we should include a message telling the user that their data was imported.

webserver/views/user.py
```diff
@@ -98,6 +99,63 @@ def import_data():
...
+            try:
+                jsonlist = json.load(f)
+                if not isinstance(jsonlist, list):
+                    raise ValueError
+            except ValueError:
+                raise BadRequest("Not a valid lastfmbackup file.")
+
+            for listen in jsonlist:
+                dic = _convert_to_native_format(listen)
+                _validate_listen(dic)
```
alastair: This code is copied from the `submit` method in the API view. We shouldn't be duplicating this code. Instead, it should be put into a utilities module, which upload and submit can both call. In addition, making it a utility method makes it much easier to test.

webserver/views/user.py
pinkeshbadjatiya: Added in commit  2834c7a

webserver/views/user.py
pinkeshbadjatiya: I tested with all the json files and i works fine. :)

webserver/views/user.py
pinkeshbadjatiya: Updated in commit  	ca8d00b

webserver/views/user.py
```diff
@@ -127,7 +127,8 @@ def upload():
                 augmented_listens.extend(_messybrainz_lookup(msb_listens))
 
             _send_listens_to_kafka("import", augmented_listens)
-    return render_template('user/import.html', user=current_user)
+            flash('Congratulations !! Your listens have been uploaded successfully.')
```
gentlecat: `Congratulations! Your listens have been uploaded successfully.`

webserver/templates/user/import.html
```diff
@@ -78,6 +78,13 @@ <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
         {{ user.auth_token }}
      </div>
 
+     <p><b> or </b></p>
+     <p>Import your listens from the lastfm backup file.</p>
+     <div>
+         <form action="upload" method="POST" enctype=multipart/form-data>
+             <input type=file name=file>
+             <input type=submit value=Upload>
+         </form>
```
gentlecat: It's usually a good idea to use Flask-WTF for creating forms.

alastair:
In principle this seems good. I'll review the code again and merge if it's good.
We should put a flag in front of the option so that we can disable it in a hurry if we find that we can't process large files which people send us.

webserver/templates/user/import.html
```diff
@@ -55,6 +55,13 @@ <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
         {{ user.auth_token }}
      </div>
 
+     <p><b> or </b></p>
+     <p>Import your listens from the lastfm backup file.</p>
```
mayhem: Please change the text to:

"Import your listens from a Last.fm backup file (.zip) or from a ListenBrainz export file (.json):"

And make it visually look more like the two sections above.

webserver/views/user.py
```diff
@@ -138,6 +142,57 @@ def export_data():
...
+        # Check upload folder
+        if not 'UPLOAD_FOLDER' in current_app.config:
+            raise InternalServerError("Could not upload the file. Upload folder not specified")
+        upload_path = path.join(path.abspath(current_app.config['UPLOAD_FOLDER']), current_user.musicbrainz_id)
+        if not path.isdir(upload_path):
+            makedirs(upload_path)
+
+        # Write to a file
+        filename = path.join(upload_path, secure_filename(f.filename))
+        f.save(filename)
```
mayhem: This file is never removed when you're done with it.

config.py.sample
```diff
@@ -53,3 +53,11 @@ LASTFM_API_KEY = "USE_LASTFM_API_KEY"
...
 # Set to True if Less should be compiled in browser. Set to False if styling is pre-compiled.
 COMPILE_LESS = True
+
+# MAX file size to be allowed for the lastfm-backup import, default is infinite
+# Size is in bytes
+MAX_CONTENT_LENGTH = 16 * 1024 * 1024  # 16MB
+
+# Specify the upload folder where all the lastfm-backup will be stored
+# The path must be absolute path
+UPLOAD_FOLDER = "METABRAINZ/listenbrainz-server/lastfm-backup-upload"
```
mayhem: Lets use "/tmp/lastfm-backup-upload" as the default here.

.gitignore
```diff
@@ -4,3 +4,6 @@
 *.pyc
 config.py
 *.log
+.cache
+env
+lastfm-backup-upload
```
mayhem: If we use the /tmp/lastfm-backup-upload default, then this doesn't need to be here.

PR #79, opened 2016-05-11
LB-90: link to MB profile from LB

reosarevok:
I don't have a running LB server, but this seems like it should work.

### DIFF
--- a/webserver/templates/user/profile.html
+++ b/webserver/templates/user/profile.html
@@ -6,6 +6,11 @@
   <div id="user-profile">
     <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
 
+    <div>
+      <img src="{{ url_for('static', filename='img/musicbrainz-16.svg') }}" />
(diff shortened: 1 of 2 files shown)


webserver/templates/user/profile.html
```diff
@@ -6,6 +6,11 @@
   <div id="user-profile">
     <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
 
+    <div>
+      <img src="{{ url_for('static', filename='img/musicbrainz-16.svg') }}" />
+      <b><a href="https://musicbrainz.org/user/{{ user.musicbrainz_id }}">See profile on MusicBrainz</a></b>
```
mineo: Does this work correctly for usernames with slashes in them?

reosarevok: Why do we allow those... I wouldn't know, though, to be honest.

mayhem: Yes, those are allowed. Care to update this PR?

PR #80, opened 2016-05-13
Storing Listens in Postgres DB

pinkeshbadjatiya:
Migrate the code to submit listens to Postgres Database.
Some major changes made in the commits:
- Use `postgresql` and `cassandra` both for storing listens. (Just need to uncomment a line to switch b/w the two)
- Use `sqlalchemy` as DB Driver for postgresql commits.
- Single config file, `config.py` for both ListenBrainz & ListenStore. (Was mentioned as TODO inside code.)
- Time batch inserts in postgresql, and log those who exceed the threshold
   -  `PG_QUERY_TIMEOUT` denotes the threshold, in miliseconds, for listen commits
   -  Default threshold: **10000ms**
- Remove listens from Postgres which are older than a particular time.
   -  Uses `apscheduler`
   -  Runs jobs every `24 hours`.
   -  `MAX_POSTGRES_LISTEN_HISTORY` denotes the time-range, in days, for listens that must be kept. Rest of the listens will be deleted. Default value is **90 days**
- `PG_ASYNC_LISTEN_COMMIT` denotes whether the listens must be commited asynchronously. Default value is **False**
- Add `listenstore-kafka-to-postgres-writer.py` to save listens to postgres. Older script for Cassandra also works fine, provided schema for cassandra tables is appropriate.
- Support showing listens from both `Cassandra` and `Postgres`. (Just need to swap a line.)
- Fix [Issue #210 on sentry](https://sentry.metabrainz.org/metabrainz/lb-production/issues/210/)
   - **Issue:** Python does not support direct string() to int() conversion. [Read More](https://github.com/metabrainz/listenbrainz-server/pull/80/files#r64879400)


admin/sql/create_tables.sql
```diff
@@ -8,4 +8,18 @@ CREATE TABLE "user" (
...
+  uid             VARCHAR NOT NULL,
+  year            INT,
+  month           INT,
+  day             INT,
+  id              INT NOT NULL,
+  artist_msid     UUID,
+  album_msid      UUID,
+  recording_msid  UUID,
+  json            VARCHAR,
+  PRIMARY KEY(uid, id)
```
alastair: this should be in the `create_primary_keys.sql` file

listenstore/listenstore/kafkaconsumer.py
```diff
@@ -37,6 +37,7 @@ def start(self, group_name, topic_name):
                 t0 = time()
 
             messages = self.consumer.get_messages(count=CASSANDRA_BATCH_SIZE, block=True, timeout=KAFKA_READ_TIMEOUT)
+            #print 'New batch'
```
alastair: whoops. try not to add print messages, especially if they're commented out

listenstore/listenstore/listenstore.py
```diff
@@ -207,8 +166,9 @@ def fetch_listens(self, uid, from_id=None, to_id=None, limit=None):
                     return
 
     def convert_row(self, row):
-        return Listen(data=ujson.loads(row.json), uid=row.uid, timestamp=row.id, album_msid=row.album_msid,
-                      artist_msid=row.artist_msid, recording_msid=row.recording_msid)
+        print(row)
```
alastair: print!

listenstore/listenstore/listenstore.py
```diff
@@ -221,7 +181,7 @@ def fetch_listens_for_range(self, uid, date_range, from_id, to_id, order, limit=
         query = """SELECT * FROM listens WHERE uid = %(uid)s AND """ + \
                 range_keys(len(date_range)) + \
                 """ AND id > %(from_id)s AND id < %(to_id)s
-                   ORDER BY id """ + ORDER_TEXT[order] + """ LIMIT %(limit)s"""
+                   ORDER BY id """ + ORDER_TEXT[order] + """ LIMIT %(limit)s;"""
```
alastair: I'm pretty sure you don't need a `;` here

listenstore/listenstore/listenstore.py
```diff
@@ -258,6 +218,98 @@ def fetch_listens_for_range(self, uid, date_range, from_id, to_id, order, limit=
             else:
                 return
 
+class PostgresListenStore(ListenStore):
+    def __init__(self, conf):
+        ListenStore.__init__(self, conf)
+        self.postgre_connection = None
+        self.postgre_cursor = None
+        
+        self.postgres_conn = psycopg2.connect("dbname='' user='listenbrainz'")
```
alastair: as I said in irc, make sure that you switch to sqlalchemy. Configuration should be in `config.py`, not in the file

listenstore/listenstore/kafkaconsumer.py
pinkeshbadjatiya: was testing a lot of stuff. Will remove it surely. :)

admin/sql/create_tables.sql
mayhem: Also, it is customary to have the id column for joining purposes:

    id          SERIAL,

Please add that at the top of the table definition.

Next, we need to add more indexes. We'll need an index on uid and possibly on all of the _msid columns. Alastair?

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: Older configurations were hardcoded in listenstore which was not good. It was mentioned as TODO in code. :)
Added support for separate configuration, `listenstore.conf`, for ListenStore.
https://github.com/pinkeshbadjatiya/listenbrainz-server/blob/b1c922ab1c6ba7417160bec7411012df9769948c/listenstore/listenstore.conf.sample

admin/sql/create_tables.sql
pinkeshbadjatiya: - merged `day`, `month` and `year` field into `date` field
- added `id` field

admin/sql/create_tables.sql
```diff
@@ -8,4 +8,15 @@ CREATE TABLE "user" (
 );
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id);
 
+CREATE TABLE listens (
+  id              SERIAL,
+  uid             VARCHAR NOT NULL,
+  date            TIMESTAMP,
```
mayhem: TIMESTAMP WITH TIME ZONE, please.

admin/sql/create_tables.sql
```diff
@@ -8,4 +8,15 @@ CREATE TABLE "user" (
...
+  timestamp       INT NOT NULL,
```
mayhem: Why have two? you don't need a date field if you have a timestamp -- the timestamp includes the date. Please remove the date field and change TIMESTAMP to TIMESTAMP WITH TIME ZONE.

admin/sql/create_tables.sql
```diff
@@ -8,4 +8,15 @@ CREATE TABLE "user" (
...
+  artist_msid     UUID,
+  album_msid      UUID,
+  recording_msid  UUID,
+  json            VARCHAR
```
mayhem: I think all the UUIDs should be NOT NULL. Please confirm @alastair.

admin/sql/create_tables.sql
pinkeshbadjatiya: My idea was,
`date` - The time when the listen was submitted
`timestamp` - Timestamp of the listen.

mayhem: What is the use case for this? When will we ever need to know when it was submitted?

pinkeshbadjatiya: I tried to keep maximal information about the users activity. Did not think much.
I'll remove it. :)

admin/sql/create_tables.sql
pinkeshbadjatiya: I was not sure about the messybrainz DB and what would happen in case listen not indexed.

alastair: our minimum required text data is artist + track name, right? so perhaps those 2 could be not null. album id could be null

mayhem: Ah yes, correct. @pinkeshbadjatiya: Please make the artist and recording_msid NOT NULL.

If it isn't indexed, it will still create an artist msid and rcording_msid, so those should be required.

listenstore/listenstore/listenstore.py
```diff
@@ -104,21 +110,126 @@ class ListenStore(object):
...
+        return self.fetch_listens_from_database(uid, from_id, to_id, limit, order, precision)
+
+
+class PostgresListenStore(ListenStore):
+    def __init__(self, conf):
+        ListenStore.__init__(self, conf)
+        self.log.info('Connecting to postgresql: %s', conf['sqlalchemy_database_uri'])
+        self.engine = create_engine(conf['sqlalchemy_database_uri'], poolclass=NullPool)
+        self.connection = self.engine.connect()
+        self.connection.execute("SET synchronous_commit TO off;")
```
mayhem: What does this do?

listenstore/listenstore/listenstore.py
```diff
@@ -104,21 +110,126 @@ class ListenStore(object):
...
+                VALUES ( %(uid)s, to_timestamp(%(id)s), %(artist_msid)s, %(album_msid)s,
+                %(recording_msid)s, %(json)s) ON CONFLICT DO NOTHING """, values)
+        except sqlalchemy.exc.DataError, e:     # Database error
+            if not self.connection.closed:
+                self.connection.close()
+            print(e)
+
+    @timeout(5)
+    def insert_batch(self, listens):
+        """ Insert a batch of listens, using asynchronous queries.
```
mayhem: In this case we should have only one function, which is the insert_batch function, which takes an array of listens. If you want to insert only one listen, send an array with only one listen.
Next, you need to add transaction support to the insert function. See https://gist.github.com/alastair/1bbdf4284d0de05de6cb5f5afdeda2f1 for how to do this.

If you do not use a transaction, then it will implicitly use a transaction for you, which is really slow. Opening a transaction at the beginning, insert row(s) and then commit the transaction is the way to go here.

admin/sql/create_tables.sql
pinkeshbadjatiya: done.

admin/sql/create_tables.sql
pinkeshbadjatiya: done.

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: done.

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: Asynchronous commit is an option that allows transactions to complete(virtually complete) more quickly. Since the no of commits are quite large in each transaction, and the length of each transaction is about 300 listens, so making the commit process async may boost up the actual speed at which the lastfm receives the listens.

I was not sure how to enable this(other than manually changing the postgres config file), but my guess was to switch it on before every transaction.

I really think this might help get a speed up. More details.
https://www.postgresql.org/docs/9.4/static/wal-async-commit.html

admin/sql/create_tables.sql
```diff
@@ -8,4 +8,14 @@ CREATE TABLE "user" (
 );
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id);
 
+CREATE TABLE listens (
+  id              SERIAL,
+  uid             VARCHAR NOT NULL,
```
gentlecat: What's `uid`?

admin/sql/create_tables.sql
```diff
@@ -8,4 +8,14 @@ CREATE TABLE "user" (
...
+  ts              TIMESTAMP WITH TIME ZONE NOT NULL,
+  artist_msid     UUID NOT NULL,
+  album_msid      UUID,
+  recording_msid  UUID NOT NULL,
+  json            VARCHAR
```
gentlecat: There must be a better name for this column, no?

listenstore/bin/listenstore-kafka-to-cassandra-writer.py
```diff
@@ -13,6 +15,12 @@ def __init__(self):
     def run(self):
         self.kafkaConsumer.start_listens(self.listenStore)
 
+    @property
+    def listenStore(self):
```
gentlecat: `listen_store` http://pep8.org/#method-names-and-instance-variables

listenstore/bin/listenstore-kafka-to-postgres-writer.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def __init__(self):
+        super(KafkaToPostgres, self).__init__()
+        self.log = logging.getLogger(__name__)
+
+    def run(self):
+        self.kafkaConsumer.start_listens(self.listenStore)
+
+    @property
+    def listenStore(self):
```
gentlecat: [Same](https://github.com/metabrainz/listenbrainz-server/pull/80/files#r64876205) here.

listenstore/listenstore/listenstore.py
```diff
@@ -104,21 +110,117 @@ class ListenStore(object):
...
+            order = ORDER_ASC
+        else:
+            order = ORDER_DESC
+
+        precision = 'month'
+        if from_id is None:
+            from_id = MIN_ID
+        if to_id is None:
+            to_id = self.max_id()
+        return self.fetch_listens_from_database(uid, from_id, to_id, limit, order, precision)
```
gentlecat: I don't see `fetch_listens_from_database` method in this class. It might be better to define it here and raise `NotImplementedError`.

gentlecat: Also `fetch_listens_from_storage` or something like that would be a more appropriate name, since you don't know where the actual implementation will store listens. It might be in memory or a file instead of database.

listenstore/listenstore/listenstore.py
```diff
@@ -104,21 +110,117 @@ class ListenStore(object):
...
         self.log = logging.getLogger(__name__)
-        self.replication_factor = conf.get("replication_factor", 1)
-        self.keyspace = conf.get("cassandra_keyspace", "listenbrainz")
-        host = conf.get("cassandra_server", "localhost:9092")
+
+    def max_id(self):
+        return int(self.MAX_FUTURE_SECONDS + calendar.timegm(time.gmtime()))
+
+    def fetch_listens(self, uid, from_id=None, to_id=None, limit=None):
+        """ Setup from_id, to_id, and limit for fetching listens
```
gentlecat: Does this method save `from_id`, `to_id`, and `limit` parameters for fetching listens in the future with some other method?

admin/sql/create_tables.sql
pinkeshbadjatiya: username. (the one we use to login)

listenstore/listenstore/listenstore.py
```diff
@@ -104,21 +110,117 @@ class ListenStore(object):
...
+            to_id = self.max_id()
+        return self.fetch_listens_from_database(uid, from_id, to_id, limit, order, precision)
+
+
+class PostgresListenStore(ListenStore):
+    def __init__(self, conf):
+        ListenStore.__init__(self, conf)
+        self.log.info('Connecting to postgresql: %s', conf['sqlalchemy_database_uri'])
+        self.engine = create_engine(conf['sqlalchemy_database_uri'], poolclass=NullPool)
+        #self.connection.execute("SET synchronous_commit TO off;")
```
gentlecat: You shouldn't keep commented-out code. Just remove it, or at least add a `TODO` comment explaining why it's here.

listenstore/listenstore/listenstore.py
```diff
@@ -104,21 +110,117 @@ class ListenStore(object):
...
+class CassandraListenStore(ListenStore):
+    def __init__(self, conf):
+        ListenStore.__init__(self, conf)
+        self.replication_factor = conf['cassandra_replication_factor']
+        self.keyspace = conf["cassandra_keyspace"]
+        host = conf["cassandra_server"]
         self.log.info('Connecting to cassandra: %s', host)
         self.cluster = Cluster([host])
-
+
```
gentlecat: 💣

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: https://github.com/metabrainz/listenbrainz-server/pull/80#discussion_r64810438

listenstore/listenstore/utils.py
```diff
@@ -3,23 +3,32 @@
...
-    return {
-        "kafka_server": "localhost:9092",
-        "cassandra_server": "localhost",
-        "cassandra_keyspace": "listenbrainz"
-    }
+import ConfigParser
+import os
+
+def config(opt_vars):
+    """ Parse config file else return default values """
```
gentlecat: In other projects we just tend to use Python files for storing configuration. Do we actually need this? @alastair?

pinkeshbadjatiya: We were using builtin flask import feature for importing config files. Since listenstore is not integrated with flask, we can't use that feature.

webserver/__init__.py
```diff
@@ -19,6 +25,9 @@ def create_app():
     import config
     app.config.from_object(config)
 
+    # Run the scheduled jobs
+    scheduledJobs = ScheduledJobs(app.config)
```
gentlecat: I'm not sure if it's a good idea to do this when creating an app.

webserver/views/user.py
```diff
@@ -29,18 +29,20 @@ def lastfmscraper(user_id):
 
 @user_bp.route("/<user_id>")
 def profile(user_id):
-    cassandra = webserver.create_cassandra()
+    # Which database to use to showing user listens.
+    #db_conn = webserver.create_cassandra()
+    db_conn = webserver.create_postgres()
```
gentlecat: Again, commented out code. You can make this an option in config file.

listenstore/listenstore/listenstore.py
mayhem: OK, interesting. Let's keep it as is for now, but make it a configurable option that uses normal commits by default. Then we can play with this once we see the load rising on the server.

webserver/views/user.py
```diff
@@ -29,18 +29,20 @@ def lastfmscraper(user_id):
...
 
     # Getting data for current page
     max_ts = request.args.get("max_ts")
     if max_ts is not None:
         try:
-            max_ts = int(max_ts)
+            max_ts = int(float(max_ts))
```
gentlecat: Why is that needed?

admin/sql/create_tables.sql
mayhem: Maybe data? raw_data? Also, the column type should be jsonb, not VARCHAR.

webserver/__init__.py
pinkeshbadjatiya: do you mean "We can start scheduledJobs it a bit later" or "Their must be a better way to do it" ?

listenstore/listenstore/listenstore.py
gentlecat: @mayhem suggested to make in configurable. I agree with that.

admin/sql/create_tables.sql
gentlecat: What do we store there?

listenstore/listenstore/utils.py
mayhem: In that case, the configuration should be passed in from the main application and the main application can store these bits in python files.

webserver/__init__.py
mayhem: Yeah, scheduled jobs should just be started by the cron daemon.

admin/sql/create_tables.sql
gentlecat: Why not `id`? `id` is a PK in `user` table.

gentlecat: Also, `uid` can be confusing. Better make it `username` or `user_id`.

admin/sql/create_tables.sql
mayhem: The raw data submitted by the user is stored there -- the user is allowed to put other pieces of data into this space that we can't know to stick into the DB. This is all good.

webserver/views/user.py
pinkeshbadjatiya: yeah. that is a good bug.
Python does not provide direct `string` to `int` conversion. It was failing when opening https://listenbrainz.org/user/armalcolite?max_ts=1459858239.0 i.e. when the timestamps are ending with decimals. (It throws on existing LB website as well)

On upgrading to postgres, the timestamp conversion ends with a decimal precision, and thus it was a problem. We could either reduce the precision in postgres, or do this to encorporate this change.

webserver/__init__.py
gentlecat: I'm saying when I run `create_app` function, I don't expect it to do anything else apart from things directly related to creating an instance of Flask app.

listenstore/listenstore/utils.py
pinkeshbadjatiya: That is again a problem.
We access Listenstore from 2 places, once while storing DB which is from the LB app, hence we can take into account what you said, but when accessing Listenstore for showing listens, we are not accessing it from the flask app. 

This was the problem i faced, and the workaround i found was to keep a copy of configuration in `listenstore.conf` and `config.py` (of the main LB app)

listenstore/listenstore/listenstore.py
```diff
@@ -104,21 +110,117 @@ class ListenStore(object):
...
+    def format_dict(self, listen):
+        return { 'uid': listen.uid,
+                 'id': listen.timestamp,
+                 'artist_msid': uuid.UUID(listen.artist_msid),
+                 'album_msid': uuid.UUID(listen.album_msid) if listen.album_msid is not None else None,
+                 'recording_msid': uuid.UUID(listen.recording_msid),
+                 'json': ujson.dumps(listen.data)}
+
+    @timeout(5)
+    def insert_postgresql(self, listens):
```
mayhem: I'm really glad you thought about database timeouts here! However, this mechanism isn't the proper one in this context, since the underlying database supports them. Also, deciding what to do with timeouts is hard to do now. In this case, I would suggest the following:

1. Remove this timeout feature.
2. Add a function to keep track how long postgres spends on the query.
3. Add a configuration option that allows users to specify a threshold at which point slow queries get logged. (e.g. if 3000 milliseconds are specified, then print a log statement for any query that takes longer than threshold (3000ms) to execute. We can then keep our eyes on the logs and see when things start getting into trouble.

listenstore/listenstore/listenstore.py
```diff
@@ -104,21 +110,117 @@ class ListenStore(object):
...
+            for listen in listens:
+                if not listen.validate():
+                    raise ValueError("Invalid listen: %s" % listen)
+                try:
+                    res = connection.execute(
+                    """INSERT INTO listens(uid, ts, artist_msid, album_msid, recording_msid, json)
+                        VALUES ( %(uid)s, to_timestamp(%(id)s), %(artist_msid)s, %(album_msid)s,
+                        %(recording_msid)s, %(json)s) ON CONFLICT DO NOTHING """, self.format_dict(listen))
+                except sqlalchemy.exc.DataError, e:     # Database error
+                    print(e)
```
mayhem: You should setup proper logging for this application and not use print statements. This applies to all places in the app where you are inclined to print debug/informational stuff.

listenstore/listenstore/listenstore.py
mayhem: Whitespace nazi!

admin/sql/create_tables.sql
pinkeshbadjatiya: done!

admin/sql/create_tables.sql
mayhem: +1 to user_id

listenstore/bin/listenstore-kafka-to-cassandra-writer.py
pinkeshbadjatiya: done!

listenstore/bin/listenstore-kafka-to-postgres-writer.py
pinkeshbadjatiya: done!

webserver/views/user.py
pinkeshbadjatiya: I am still not sure if i must remove the cassandra support completely. In that i may need to do some code refactoring, initially i tried to make it compatible with both. I think its better to just remove Cassandra code.

@mayhem fine?

admin/sql/create_tables.sql
pinkeshbadjatiya: done!

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: no. these are passed on every request.

gentlecat: Then this docstring needs an update.

pinkeshbadjatiya: Setup in the sense, if the user has given valid values or not. Also, if none is given, then use default values and pass it to the next method,  `fetch_listens_from_database()`

listenstore/bin/listenstore-kafka-to-cassandra-writer.py
pinkeshbadjatiya: There's a catch :'(
All the methods which are marked as `@property` follow CamelCase notation in the existing LB structure. (ex. listenStore, kafkaConsumer). This might violate PEP8 standards, but i believe its good to keep it the way it is. Sometimes even i find it hard to track variables and methods, considering so diverse structure of ListenBrainz.

Though, changing it will require considerable change in a lot of method names.

gentlecat: Well, you can put that in a separate commit if you don't want to mix things together. But that's the convention we use in all of our projects.

gentlecat:
It would also be nice to see some tests for things that you are adding.

listenstore/listenstore/utils.py
gentlecat: You can load configuration values from config file written in regular Python using this https://github.com/pallets/flask/blob/master/flask/config.py#L111-L137. Can just reuse that class since Flask is one of current requirements.

listenstore/bin/listenstore-kafka-to-cassandra-writer.py
pinkeshbadjatiya: done!

listenstore/listenstore/listenstore.py

listenstore/listenstore/listenstore.py

listenstore/listenstore/utils.py
pinkeshbadjatiya: **Done!**
(This is the minimal what i could create, [https://github.com/pinkeshbadjatiya/listenbrainz-server/blob/040dd4a20fa69e96b29affa15777451d7caffe85/listenstore/listenstore/loadconfig.py](https://github.com/pinkeshbadjatiya/listenbrainz-server/blob/040dd4a20fa69e96b29affa15777451d7caffe85/listenstore/listenstore/loadconfig.py) )

webserver/__init__.py
pinkeshbadjatiya: done!

pinkeshbadjatiya:
Implementation is not yet finalized for every piece of code. It would be difficult to write tests in such a scenario where things keep changing too frequently. 

Also, a lot of stuff is going on in this PR. Its better to keep tests separate. The previous tests also need to be updated. I would be writing tests for a lot of my previous PR's as well once things get shape. :)

listenstore/listenstore/listenstore.py

webserver/views/user.py
mayhem: I'm fine removing it. The old code is in git, so we can always dig it back out.

listenstore/listenstore/listenstore.py

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: Done!

gentlecat:
@brainzbot, add to whitelist.

PR #82, opened 2016-06-20
[WIP] Compatible API for lastfm

pinkeshbadjatiya:
## Compatible API for ListenBrainz (api_compat)
A submission API compatible with Last.fm scrobblers, [more details](https://community.metabrainz.org/t/listenbrainz-a-submission-api-compatible-with-last-fm-scrobblers/2708)

### Features
- Create new schema: `token` & `session` & `listen_json`.
- Add class for `User`, `Token` and `Session`
- Add complete support for the following methods:
  - **auth.getToken**
  - **auth.getSession**
  - **track.scrobble**
  - **track.updateNowPlaying**
  - **user.getinfo**
  - **auth.getsessioninfo**
  - **Approve Token**
- Throw errors with appropriate error codes mentioned [here](http://www.last.fm/api/errorcodes).
- Response rendering can be done in `xml` or `json`
- Generic **xml** -> **json** conversion. Multiple formats can be supported by writing the specific xml parser for that particular format.
- **TOKEN_EXPIRATION_TIME** specifies the max time in minutes before a token expires. Initial value is `60 minutes`
- Separate common bits from the native_API and the compat_API.
- Fixed a few PEP8 and pylint indenting issues.

### More details about the API
- The API key would be the one given on this [page](https://listenbrainz.org/user/import) (Spelled as `User token` on the page)
- Routes for submission: 
  - auth.getToken: `http://listenbrainz.org/2.0/method=auth.getToken&...`
  - auth.getSession: `http://listenbrainz.org/2.0/method=auth.getSession&...`
  - track.scrobble: `http://listenbrainz.org/2.0/method=track.scrobble&..`
  - track.updateNowPlaying: `http://listenbrainz.org/2.0/method=track.updateNowPlaying&...`
  - user.getinfo: `http://listenbrainz.org/2.0/method=user.getinfo&...`
  - auth.getsessioninfo: `http://listenbrainz.org/2.0/method=auth.getsessioninfo&...`
  - Approve Token: `http://listenbrainz.org/api/auth/?api_key=d24..08D&token=a0032a`
- There is ambiguity regarding the method types for some of the methods mentioned above amongst the clients, so, all the methods support both `GET` and `POST` format, except for `Approve Token`.
- API_key is application specific. There is no way to verify if the api_key is correct or not. Last.fm contains that mapping between the api_key and the Application Name.
- Allows single token per user.
- Allows multiple sessions per user.

### Instructions
- The instruction for using the API in `audacious` are mentioned [here](https://github.com/pinkeshbadjatiya/listenbrainz-server/blob/api_compat/API_compatible_README.md).

### Known issues/Open questions for discussion:
- Duplicate token generation might disturb functionality.
- Session limit per API key to avoid overuse of the service? any suggestion?


admin/sql/create_primary_keys.sql
```diff
@@ -2,5 +2,6 @@ BEGIN;
 
 ALTER TABLE "user" ADD CONSTRAINT user_pkey PRIMARY KEY (id);
 ALTER TABLE listen ADD CONSTRAINT listen_pkey PRIMARY KEY (user_id, ts);
+ALTER TABLE sessions ADD CONSTRAINT sessions_pkey PRIMARY KEY (sid);
```
mayhem: You have no indexes on the session table. You need to add appropriate indexes for the types of queries you're going to make.

admin/sql/create_tables.sql
```diff
@@ -18,4 +18,22 @@ CREATE TABLE listen (
...
 );
 
+CREATE TABLE tokens (
+     id               SERIAL,
+     user_id          VARCHAR,
+     token            TEXT NOT NULL,
+     api_key          VARCHAR NOT NULL,
+     ts               TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+ALTER TABLE tokens ADD CONSTRAINT tokens_api_key_uniq UNIQUE (api_key);
```
mayhem: Same here, the tokens table needs indexes.

admin/sql/create_tables.sql
```diff
@@ -18,4 +18,22 @@ CREATE TABLE listen (
...
+
+CREATE TABLE sessions (
+    id        SERIAL,
+    user_id   VARCHAR NOT NULL,
```
mayhem: should this be referring to the user table as a foreign key?

db/mockdata.py
```diff
@@ -0,0 +1,159 @@
+"""
```
mayhem: Why are is this file called mockdata?

admin/sql/create_tables.sql
pinkeshbadjatiya: **Thanks for pointing out! I almost forgot to revert this change.**
Earlier i was trying to keep this column just for storing extra info but afterwards i changed a lot of schema to incorporate api-key checks at every stage which required a lot of table joins to access the **"user"** table.

I'll have to reduce the number of queries for a single batch, to improve performance. Converting columns to foreign keys is part of this task.

db/mockdata.py
pinkeshbadjatiya: I could not find a place to keep these classes so i created a temporary file.
These classes are needed only by the **api_compat** so merging them into existing files is also not a good idea. I can move them to **api_compat.py** or **api_tools.py** to make sure that the api_compat is modular and its bits are not spread across the whole app.

I would love to here any suggestions from you.

admin/sql/create_tables.sql
pinkeshbadjatiya: done!

admin/sql/create_tables.sql
pinkeshbadjatiya: done!

admin/sql/create_primary_keys.sql
pinkeshbadjatiya: done!

db/mockdata.py
mayhem: Lets go with api-tools.py.

If you need to use a non-whitespace separator, please use - and not _. It makes it easier to read overall.

db/mockdata.py
```diff
@@ -0,0 +1,183 @@
...
+This module contains classes required by the compact API.
```
mayhem: s/compact/compat/

db/mockdata.py
```diff
@@ -0,0 +1,183 @@
...
+TOKEN_EXPIRATION_TIME = 60
+
+
+class User(object):
+    def __init__(self, row):
+        serial, created, name, auth_token = row
+        self.id = serial
+        self.name = name
+        self.created = created
+        self.api_key = auth_token
```
mayhem: These 5 lines could be compacted into one line by using the construct on line 18.

db/mockdata.py
```diff
@@ -0,0 +1,183 @@
...
+        self.sid = sid
+        self.api_key = api_key
+        self.timestamp = timestamp
+
+    @staticmethod
+    def load(session, api_key=None):
+        """ Load the session details from the database.
+            If the api_key is also supplied then verify it as well.
+        """
+        dic = {'sid': session}
```
mayhem: dic is a bit of a generic name. Please use something a little more specific for the task.

db/mockdata.py
mayhem: Also, there are multiple classes in one python module. It would be nice to separate those out into different files. Maybe start naming the whole set of files lastfm_session.py, lastfm_token.py ... ?

db/test/test_mockdata.py
```diff
@@ -0,0 +1,185 @@
+# -*- coding: utf-8 -*-
```
mayhem: here as well. please rename the modules and break into smaller files.

listenstore/listenstore/listenstore.py
```diff
@@ -1,42 +1,36 @@
 # coding=utf-8
-from __future__ import division, absolute_import
+from __future__ import absolute_import
 from __future__ import print_function, unicode_literals
 import logging
 import calendar
-import ujson
```
mayhem: Why are you getting rid of ujson? ujson is considerably faster than the json module. We should not use the json module unless we have to.

pinkeshbadjatiya: Earlier alastairp mentioned that they were having issues with `ujson` in AB. So i thought it was better to remove it altogether. Also, the json's what we are dealing with, are not quite big, so the performance gain would not be that significant.

webserver/views/api.py
```diff
@@ -1,34 +1,17 @@
 import sys
-import ujson
```
mayhem: always use ujson

webserver/views/api_compat.py
```diff
@@ -0,0 +1,358 @@
...
+        if 'trackNumber' in data or 'duration' in data or 'albumArtist' in data or \
+                'choosenByUser' in data or 'streamId' in data or 'context' in data:
+            # Album artist, Song Duration, trackNumber on the album, etc
+            # are not supported by the native ListenBrainz API
+            pass
+        listens.append(listen)
+    return listen_type, listens
+
+
+def scrobble_listens(request, data):
```
mayhem: We should not use the word scrobble in any of our code. May use "record_listens"?

webserver/views/api_compat.py
```diff
@@ -0,0 +1,358 @@
...
+        lookup[number][key] = value
+
+    # Convert to native payload then submit 'em.
+    listen_type, native_payload = _to_native_api(lookup, data['method'])
+    insert_payload(native_payload, session.user.name, listen_type=listen_type)
+
+    # With corrections than the original submitted listen.
+    doc, tag, text = Doc().tagtext()
+    with tag('lfm', status="ok"):
+        with tag("nowplaying" if listen_type == "playing_now" else "scrobbles"):
```
mayhem: except here as a string constant, the word scrobbles is allowed. :)

webserver/views/api_compat.py
```diff
@@ -0,0 +1,358 @@
...
+                    corr['artist'] = '1'
+
+                ts = augL['listened_at']
+
+                albumArtist = artist
+                if origL.get('albumArtist', origL['artist']) != artist:
+                    corr['albumArtist'] = '1'
+
+                # TODO: Add the album part
+                album = ""
```
mayhem: Finish this. :)

db/mockdata.py
alastair: you can't import a python module with a - in the name

mayhem: Curses, foiled again. OK, keep the _ in the module names. :)

pinkeshbadjatiya: ok.

db/mockdata.py
pinkeshbadjatiya: done!

db/mockdata.py
pinkeshbadjatiya: done!

db/mockdata.py
pinkeshbadjatiya: done!

db/test/test_mockdata.py
pinkeshbadjatiya: done!

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: Replaced every instance with `ujson`.

webserver/views/api_compat.py
pinkeshbadjatiya: Changed it to `record_listens()`.

mayhem:
The import feature still fails with a 500 error on the latest commit: BadRequest: JSON document must contain the key listened_at at the top level. :(

mayhem:
Please add some documentation on the steps required to configure a last.fm enabled client to scrobble to LB. I would suggest adding a new file in the root of the source tree for this.

API_compatible_README.md
```diff
@@ -0,0 +1,24 @@
...
+1. Install dependencies from **SOURCE_1** then clone the repo and install audacious.
+2. Before installing audacious-plugins, edit the file `audacious-plugins/src/scrobbler2/scrobbler.h` to update the following settings on lines L26-L28.  
+  - `SCROBBLER_API_KEY` to the key given on "listenbrainz.org/user/import" page.
+  - `SCROBBLER_URL` to "http://last.fm/2.0/".  
+  Now compile and install the plugins from the instructions given in **SOURCE_1**.
+3. Register an application on musicbrainz with the following callback URL "http://www.last.fm/login/musicbrainz/post"
+4. Update the received client-id and client-secret in config.py of listenbrainz.
+5. Edit the `/etc/hosts` file and add the following entries.
+
+  > 127.0.0.1 http://www.last.fm
```
mayhem: URLs do no belong in /etc/hosts files, at least not un UNIX systems.

db/lastfm_session.py
```diff
@@ -0,0 +1,44 @@
...
+            If the api_key is also supplied then verify it as well.
+        """
+        params = {'sid': session}
+        query = "SELECT * FROM session WHERE sid=:sid"
+        if api_key:
+            params['api_key'] = api_key
+            query = "SELECT * FROM session WHERE sid=:sid AND api_key=:api_key"
+
+        with db.engine.connect() as connection:
+            result = connection.execute(text(query), params)
```
mayhem: This comment applies to the whole codebase -- please review the entire codebase for instances where this comment might apply.

Anywhere you make calls to the database you need to catch DB exceptions and handle them appropriately.  I would suggest that in the API handlers that you catch the exceptions there and then raise the appropriate HTTP exception. This way the user gets more informative information than just a 500 error that may not make sense to them.

mayhem:
This is coming along nicely. I still haven't resolved what the problem on import is -- we'll likely need to work on that together to see what the problem is.

You'll need to work on adding catching the DB exceptions and returning appropriate HTTP error codes. It might also be nice to add Sentry support for logging exceptions in a nice way. Sentry is a took for logging and showing exceptions that a web site might raise -- if you get all these things that I commented on done, then please touch base with Gentlecat and has him for how to add sentry support.

I will be minimally around this week -- I'm traveling again and am unsure how much net connectivity I'll have. If you still run out of things to do, you can always work to add more tests. :)

API_compatible_README.md
```diff
@@ -0,0 +1,24 @@
+# API Compatible
+
+*Instructions are for setting up API for Audacious client on Ubuntu. The steps can be modified for other clients as well.*
+
+**SOURCE_1**: http://redmine.audacious-media-player.org/boards/1/topics/788
+
+1. Install dependencies from **SOURCE_1** then clone the repo and install audacious.
+2. Before installing audacious-plugins, edit the file `audacious-plugins/src/scrobbler2/scrobbler.h` to update the following settings on lines L26-L28.  
+  - `SCROBBLER_API_KEY` to the key given on "listenbrainz.org/user/import" page.
+  - `SCROBBLER_URL` to "http://last.fm/2.0/".
```
alastair: One of the reasons that we want to have a compatibility API is so that users can submit listens to ListenBrainz without having to change any code or recompile their software (consider users of windows software who can only download a binary, or even linux users who install from their package manager)
In fact, if they are changing code, then there would be nothing stopping us from changing the scrobbler url to listenbrainz.org, for example.

Why do you require that the API key here is set to the user's import page? From my understanding of lastfm authentication, this key is specific to the application, not to each individual user.

API_compatible_README.md
```diff
@@ -0,0 +1,24 @@
...
+*Instructions are for setting up API for Audacious client on Ubuntu. The steps can be modified for other clients as well.*
+
+**SOURCE_1**: http://redmine.audacious-media-player.org/boards/1/topics/788
+
+1. Install dependencies from **SOURCE_1** then clone the repo and install audacious.
+2. Before installing audacious-plugins, edit the file `audacious-plugins/src/scrobbler2/scrobbler.h` to update the following settings on lines L26-L28.  
+  - `SCROBBLER_API_KEY` to the key given on "listenbrainz.org/user/import" page.
+  - `SCROBBLER_URL` to "http://last.fm/2.0/".  
+  Now compile and install the plugins from the instructions given in **SOURCE_1**.
+3. Register an application on musicbrainz with the following callback URL "http://www.last.fm/login/musicbrainz/post"
```
alastair: Why is this step required? Is it because the URL that audacious generates starts as `http://last.fm/api/auth...`? I don't know if it's a good idea for us to respond to requests like this with the listenbrainz site. Perhaps instead our instructions could say something like "When you get a URL from your application which starts `http://last.fm/api/auth`, replace it with `http://listenbrainz.org/.../../...`"

admin/sql/create_tables.sql
```diff
@@ -18,4 +18,22 @@ CREATE TABLE listen (
   raw_data        JSONB
 );
 
+CREATE TABLE token (
+     id               SERIAL,
```
alastair: If these tables are only used for our compatibility API, I'd like to separate them. Either give them a name prefix (`compat`?) or put them in a separate schema.

admin/sql/create_tables.sql
```diff
@@ -18,4 +18,22 @@ CREATE TABLE listen (
...
+     user_id          INTEGER, -- FK to "user".id
+     token            TEXT NOT NULL,
+     api_key          VARCHAR NOT NULL,
+     ts               TIMESTAMP DEFAULT CURRENT_TIMESTAMP
+);
+ALTER TABLE token ADD CONSTRAINT token_api_key_uniq UNIQUE (api_key);
```
alastair: Our policy is to not mix create table and alter statements in this file. If you are doing lookups on this value, perhaps you should have a `UNIQUE INDEX` instead, in which case these statements should be in `create_indexes.sql`

db/lastfm_session.py
alastair: just as an FYI, we've been skipping these checks in AcousticBrainz because of the amount of boilerplate code it added every time we made a query. I agree that it's a good idea to try and catch these errors, perhaps we can take a look at this again and see if we can find a nice way of doing it

db/lastfm_session.py
```diff
@@ -0,0 +1,44 @@
...
+            row = result.fetchone()
+            if row:
+                return Session(row)
```
alastair: There's quite a bit of "magic" in this line. In the select query you get `*` (all columns), and in the constructor for `Session`, it expects that there are 5 values (`self.id, userid, self.sid, self.api_key, self.timestamp = row`). This means that if we ever add/remove a column to this table, and don't update this or have tests, we're going to have a problem.

In AcousticBrainz we treat results of `connection.execute` as a dictionary. In fact, you can do this:
```python
res = connection.execute(text(query), params)
return dict(res)
```
and you'll get a dictionary back.

Our preference for SQL statements is to explicitly list all columns that we want, for example see here: https://github.com/metabrainz/acousticbrainz-server/blob/master/db/dataset.py#L120

My preference would be for these queries to explicitly select the columns that we need, and then explicitly pass 5 parameters into the `Session` object by name.

db/test/test_lastfm_token.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def setUp(self):
+        super(TestAPICompatTokenClass, self).setUp()
+        self.log = logging.getLogger(__name__)
+
+        # Create a user
+        uid = db.user.create("test")
+        self.assertIsNotNone(db.user.get(uid))
+        with db.engine.connect() as connection:
+            result = connection.execute(text('SELECT * FROM "user" WHERE id = :id'),
```
alastair: You're mixing quote styles here. If you need quotes in a string (as in our `"user"` case), I'd prefer we use triple-quotes for strings.
I prefer to write queries in multiple parts for readability, e.g.
```python
query = text("""
    SELECT a
         , b
         , c 
      FROM "user" 
     WHERE id = :id""")
result = connection.execute(query, {"id": uid})
```

Note the alignment of the SQL queries - it's very specific. keywords are right-aligned, column names and expressions are left-aligned, and commas go on a new line (this is so that if you add another WHERE clause or column to select, you only add 1 new line, instead of changing another to add a comma at the end of the line)

db/user.py
```diff
@@ -44,6 +44,8 @@ def get_by_token(token):
         return dict(row) if row else None
 
 
+
```
alastair: Be careful about adding things that are not needed. One way to check this is to use a linter, to check the layout of your code, though we don't enforce this at the moment. I often use `git add -p` to verify the diffs which I add, or at the very least always use `git diff` before `git add`, to see the changes that I'll make.

listenstore/listenstore/listenstore.py
```diff
@@ -89,23 +87,28 @@ def insert_postgresql(self, listens):
...
             res = connection.execute(query, params)
             return res.fetchall()
 
     def fetch_listens_from_storage(self, user_id, from_id, to_id, limit, order, precision):
-        query = """ SELECT id, user_id, extract(epoch from ts), artist_msid, album_msid, recording_msid, raw_data """ + \
-                """ FROM listen WHERE user_id = %(user_id)s """ + \
-                """ AND extract(epoch from ts) > %(from_id)s AND extract(epoch from ts) < %(to_id)s  """ + \
-                """ ORDER BY extract(epoch from ts) """ + ORDER_TEXT[order] + """ LIMIT %(limit)s"""
+        query = """
+                SELECT id, user_id, extract(epoch from ts), artist_msid, album_msid, recording_msid, raw_data
```
alastair: See the other comment I made about layout of queries. I think this was originally copied from the cassandra code, right? In that case it's OK that we've not formatted it yet, but I'd like to see all SQL statements uniformly formatted

webserver/errors.py
```diff
@@ -1,6 +1,41 @@
 from flask import render_template, make_response
+from yattag import Doc
+import yattag
+import ujson
 
 
+# List of errors compatible with LastFM messages for API_compat.
+ERROR_MESSAGES = {
+    1: "This error does not exist",
```
alastair: This is a good way of storing the error messages, but I'm a little worried about using the numbers in the rest of the views. This is what's called a "magic number" (https://en.wikipedia.org/wiki/Magic_number_(programming)), and is generally considered a bad idea, because it makes the context in which the number is being used unclear.
Unfortunately, python 2.7 doesn't have an enumerator type, which makes this kind of pattern really annoying to do. 
This solution is a pretty reasonable idea: http://stackoverflow.com/a/36937
We could have a
```python
class CompatError:
    DOES_NOT_EXIST = 1
    INVALID_SERVICE = 2
   ... etc

ERROR_MESSAGES = {DOES_NOT_EXIST: "This error does not exist", ...}
```
You can make up names for each of the variables, just keep them short but understandable and unique among all errors.

webserver/templates/user/auth.html
```diff
@@ -0,0 +1,32 @@
...
+
+{% block content %}
+  <div id="user-profile">
+    <h2 class="page-title">{{ user_id }}</h2>
+
+    {% if msg %}
+    <p> {{ msg }} </p>
+
+    {% else %}
+    <h3>Approve this client for account <b>{{ user_id }}</b>?</h3>
```
alastair: We can add some more text here. How about:

A legacy last.fm application wishes to have access to your account. If you approve it, this application will be able to submit listens on your behalf.
Please consider asking the developers of your application to add support to submit directly to ListenBrainz.

Allow access?   [Yes please]

webserver/templates/user/auth.html
```diff
@@ -0,0 +1,32 @@
...
+        <input type='hidden' name='token' value='{{ token }}'/>
+        <input type='submit' value='Yes please'/>
+    </form>
+    {% endif %}
+
+  </div>
+{% endblock %}
+
+{% block scripts %}
+  {{ super() }}
```
alastair: What's this? It looks like it's from the user profile page. Make sure that when you copy files for a new task that you remove all unneeded code.

webserver/views/api_compat.py
```diff
@@ -0,0 +1,352 @@
...
+
+
+@api_bp.route('/api/auth/', methods=['POST'])
+@ratelimit()
+@login_required
+def api_auth_approve():
+    """ Authenticate the user token provided.
+    """
+    user = User.load_by_name(current_user.musicbrainz_id)
+    token = Token.load(request.form['token'])
```
alastair: what if this page is submitted with no `token` value in the form?

webserver/views/api_compat.py
```diff
@@ -0,0 +1,352 @@
...
+    method = request.form['method'].lower()
+    def invalid_method_error(*args):
+        raise InvalidAPIUsage(3, output_format=request.args.get('format', "xml"))   # Invalid Method
+
+    return {
+        'track.updatenowplaying': record_listens,
+        'track.scrobble': record_listens,
+        'auth.getsession': get_session,
+        'auth.gettoken': get_token,
+        'user.getinfo': user_info
```
alastair: This is a novel way of choosing which helper function to call based on the `method` parameter, but it took me a few seconds to work out what the code did. There is a lot happening here (return statement, dict get, default value, function execution).

There's a famous programming quote which is worth keeping in mind here:

> “Debugging is twice as hard as writing the code in the first place. Therefore, if you write the code as cleverly as possible, you are, by definition, not smart enough to debug it.”
> https://blogs.msdn.microsoft.com/alfredth/2009/01/08/are-you-smart-enough-to-debug-your-own-code/

I would prefer this to be a series of `if` statements, to make it easier to quickly understand how it works.

webserver/views/api_compat.py
```diff
@@ -0,0 +1,352 @@
...
+    }.get(method, invalid_method_error)(request, request.form)
+
+
+
+def get_token(request, data):
+    """ Issue a token to user after verying his API_KEY
+    """
+    output_format = data.get("format", "xml")
+    api_key = data.get('api_key', None)
```
alastair: `Dict.get()` returns None by default if the key doesn't exist, you dont' need to add it explicitly.

Be careful about using different quotes in the same file. Pick one and stick with it (we tend to use double-quotes)

webserver/views/api_compat.py
```diff
@@ -0,0 +1,352 @@
...
+
+    if not api_key:
+        raise InvalidAPIUsage(6, output_format=output_format)   # Missing required params
+    if not Token.is_valid_api_key(api_key):
+        raise InvalidAPIUsage(10, output_format=output_format)   # Invalid API_KEY
+
+    token = Token.generate(api_key)
+
+    doc, tag, text = Doc().tagtext()
+    with tag('lfm', status="ok"):
```
alastair: two different types of quotes on the same line!

webserver/views/api_compat.py
```diff
@@ -0,0 +1,352 @@
...
+                    with tag('timestamp'):
+                        text(ts)
+                    with tag('ignoredMessage', code="0"):
+                        text('')
+
+    return format_response('<?xml version="1.0" encoding="utf-8"?>\n' + yattag.indent(doc.getvalue()),
+                           output_format)
+
+
+def format_response(data, format="xml"):
```
alastair: This seems like a roundabout way of producing the correct output format. From what I understand, you generate an XML string in each view, and then in this method if the format is json you re-parse the xml and convert it to a dictionary.
I wonder if it would be more understandable to create a python dict in each view instead, and then in this `format_response` method you either convert this dict to XML, or serialise it directly to JSON.

webserver/views/api_compat.py
```diff
@@ -0,0 +1,352 @@
...
+        if 'track' in data:
+            listen['track_metadata']['track_name'] = data['track']
+        if 'artist' in data:
+            listen['track_metadata']['artist_name'] = data['artist']
+        if 'mbid' in data:
+            listen['track_metadata']['release_mbid'] = data['mbid']
+        if 'trackNumber' in data or 'duration' in data or 'albumArtist' in data or \
+                'choosenByUser' in data or 'streamId' in data or 'context' in data:
+            # Album artist, Song Duration, trackNumber on the album, etc
+            # are not supported by the native ListenBrainz API
```
alastair: I don't think this is right - from the documentation at http://listenbrainz.readthedocs.io/en/latest/dev/json.html#payload-json-details:
> At this point, we are not scrubbing any superflous elements that may be submitted via the additional_info element.
Therefore, if you need to add fields that last.fm supports, but are not listed in the documentation, you should propose a key to add to the `additional_info` field, and also add them to our documentation.

admin/sql/create_tables.sql
pinkeshbadjatiya: Creating schema seems to be a good choice.

db/user.py
pinkeshbadjatiya: I do use `-p` after you suggested this a few weeks earlier. I also review twice before the commits. This code change was pushed 15 days ago and i did not touch the line recently. Its also that sometimes such things just slip out of sight. I'll try not to push them again. :)

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: The initial logic was cassandra's but i changed it quite a bit for postgres. I have started refactoring code quite a lot recently. I'll devote more time to it. (Usually i re-factor code as i encounter them, so gradually it will reach the required state. :) )

webserver/views/api_compat.py
pinkeshbadjatiya: I really spent quite sometime to figure out a better way to write this redirection scheme. I really wanted NOT to use those if statements. I think this way its pretty straight forward and future edits do not require much change.

I would really want NOT to use those if statements. They consume too much lines and pretty much do nothing.

webserver/views/api_compat.py
pinkeshbadjatiya: These fields are optional and clients can submit them. I tested this API with audacious on ubuntu and the client sent other parameters like `duration`. Those are just not compulsory. I wanted not to miss this on the code base on order to keep track of things. This is surely a additional feature that might be required while developing user stats and if required we can add them.

webserver/views/api_compat.py
pinkeshbadjatiya: For someone new looking at the code, who does not know about `.get()`, he'll have to spawn a shell to check for the return value in case of key not present.  Specifying it explicitly just makes it easier to read and the number of assumptions regarding the values made are minimal.

API_compatible_README.md
pinkeshbadjatiya: My instructions were for testing the new API for the local server (for the developers). I still haven't thought much from the user's perspective. Your idea seems good, we can go with this.
Also, once the user has changed the hosts file, then, `last.fm` points to `listenbrainz.org` and so the authentication URL can have anything, listenbrainz.org or last.fm

db/lastfm_session.py
pinkeshbadjatiya: sure! i'll wait for a while then. I would love to know if you come across a better way of doing this.

API_compatible_README.md
pinkeshbadjatiya: - Audacious has network issues while connecting to last.fm so i had to change the URL. (Was not sure if its a bug). And regarding the scrobbling URL, i realised this when testing it on a windows client. This problem can be solved by adding 2 entries in **/etc/hosts**:  
`127.0.0.1 ws.audioscrobbler`  
`127.0.0.1 last.fm`   
I'll surely add more instructions on how to setup this week.
- I was quite confused with API key. I had web authentication in mind, which requires the use of API key allocated to the user. But user should never use API_compat for scrobbling using the web-auth. LB has its own native API for that. Would be updating the same in the next commits.

API_compatible_README.md
pinkeshbadjatiya: Updated the instructions.

API_compatible_README.md
pinkeshbadjatiya: - Updated the instructions to prevent changing source code.
- Made changes to the api_key logic in 0fa32e0112b67982dc1fb41e6238a4ddd6721802

API_compatible_README.md
pinkeshbadjatiya: Updated.

API_compatible_README.md
```diff
@@ -0,0 +1,26 @@
+# API Compatible
+
+*Instructions are for setting up API for Audacious client on Ubuntu. The steps can be modified for other clients as well.*
+
+**SOURCE_1**: http://redmine.audacious-media-player.org/boards/1/topics/788
```
gentlecat: An actual link in the text would be easier to access.

pinkeshbadjatiya: done!

db/lastfm_session.py
```diff
@@ -38,7 +39,14 @@ def create(token):
         """
         session = binascii.b2a_hex(os.urandom(20))
         with db.engine.connect() as connection:
-            connection.execute(text("INSERT INTO session (user_id, sid, api_key) VALUES (:user_id, :sid, :api_key)"),
-                               {'user_id': token.user.id, 'sid': session, 'api_key': token.api_key})
-        token.consume()
-        return Session.load(session)
+            result = connection.execute(text("""
+                INSERT INTO session (user_id, sid, api_key)
```
alastair: good formatting :+1:

webserver/views/api_compat.py
alastair: When I see an if statement I immediately know what it means. When I saw this statement it took me about 15 seconds to understand what you were doing and why it works.
We read code much more than we write it, therefore we should help our future selves to understand our code when we write it. 

Your solution also requires hacks with varargs for things like your `invalid_method_error` function (https://github.com/metabrainz/listenbrainz-server/pull/82#discussion-diff-71166506R84), which would not be needed with normal control flow.

Please change to a series of if statements

webserver/views/api_compat.py
alastair: `.get` is already "advanced" usage (compared to using `[]`). If you're already using get, the definition of the function is known, and written in the documentation (https://docs.python.org/2/library/stdtypes.html#dict.get). 

Please remove instances of `None` as a second argument to `.get`.

API_compatible_README.md
```diff
@@ -4,21 +4,23 @@
...
 1. Install dependencies from **SOURCE_1** then clone the repo and install audacious.
-2. Before installing audacious-plugins, edit the file `audacious-plugins/src/scrobbler2/scrobbler.h` to update the following settings on lines L26-L28.  
-  - `SCROBBLER_API_KEY` to the key given on "listenbrainz.org/user/import" page.
-  - `SCROBBLER_URL` to "http://last.fm/2.0/".  
-  Now compile and install the plugins from the instructions given in **SOURCE_1**.
-3. Register an application on musicbrainz with the following callback URL "http://www.last.fm/login/musicbrainz/post"
-4. Update the received client-id and client-secret in config.py of listenbrainz.
-5. Edit the `/etc/hosts` file and add the following entries.
+2. Before installing audacious-plugins, edit the file `audacious-plugins/src/scrobbler2/scrobbler.h` to update the following setting on line L28. This is required only because the local server does not have https support.  
+  - `SCROBBLER_URL` to "http://ws.audioscrobbler.com/2.0/".
```
alastair: Hmm. This is a more serious problem than just not having https support on local development.
If the client checks SSL certificates then we will never be able to intercept requests even if the user changes their hosts file. Can you do a survey of a handful of last.fm clients (we've been recommending a number of them in IRC) and see if you can work out which ones use http and which use https. You can make this list on the jira ticket related to the compatibility API. We will then need to check which https clients have certificate checking.
It may be that many of these are open source, and therefore our options are greater (add a native LB submitter, add an option to change host between last.fm/librefm/LB, etc), but we won't know until we look.

API_compatible_README.md
```diff
@@ -4,21 +4,23 @@
...
-  Flush dns and restart network manager using:  
-  `sudo /etc/init.d/dns-clean start`  
-  `sudo /etc/init.d/networking restart`  
+ Flush dns and restart network manager using:  
+ `sudo /etc/init.d/dns-clean start`  
+ `sudo /etc/init.d/networking restart`  
+5. Register an application on musicbrainz with the following callback URL "http://\<hostname\>:\<port\>/login/musicbrainz/post" and update the received client-id and client-secret in config.py of listenbrainz. hostname and port number are as per the server.  
 6. In audacious, goto, File > Settings > Plugins > Scrobbler2.0 and enable it. Now Open its settings and then authenticate.  
-7. Once done, Enjoy!
+7. When you get a URL from your application which starts http://last.fm/api/auth, replace it with http://\<hostname\>:\<port\>/.../../...". Hostname is listenbrainz.org if you are not running a local server.
```
alastair: be explicit about your instructions here. I think you copied what I said in my previous comment. This is unclear that if the URL says "http://last.fm/api/auth" that you should delete "api/auth". Put yourself in the position of someone who isn't a programmer and wants to use this compatibility API. The instructions should be clear and explicit about exactly what to change.

API_compatible_README.md
alastair: Here's a small list of libre.fm compatibile players: https://git.gnu.io/foocorp/librefm/wikis/Set-Up-Your-Music-Player

webserver/views/api_compat.py
alastair: Are these optional parameters documented on the last.fm website, or from your experience do different programs all send different data?
For this feature to be complete, I believe we need to accept all data sent from other clients, so perhaps we could put all items which we don't explicitly select above this line directly into `additional_info`. If last.fm explicitly documents a whitelist of accepted additional data, we should add all of this data if it is present.

admin/sql/create_tables.sql
pinkeshbadjatiya: done!

webserver/templates/user/auth.html
pinkeshbadjatiya: done!

webserver/views/api_compat.py
pinkeshbadjatiya: fixed it.

webserver/views/api_compat.py
pinkeshbadjatiya: done!

webserver/views/api_compat.py
pinkeshbadjatiya: done!

webserver/views/api_compat.py
pinkeshbadjatiya: done!

API_compatible_README.md
pinkeshbadjatiya: updated the instructions.

admin/sql/create_tables.sql
pinkeshbadjatiya: I already have a index on `(token, api_key)` which would be used for lookups. But each token has to be UNIQUE, that is why i added a UNIQUE constraint on this column. Probably this is a better way to do this,
```
CREATE TABLE token (                                                                                                                                                    
     ... 
     token            TEXT NOT NULL UNIQUE,
     ...
);
```

webserver/views/api_compat.py
pinkeshbadjatiya: A lot a params are mentioned on the website (http://www.last.fm/api/show/track.scrobble).
I have added support for all of them in LB, except for album_artist as we do not accept album artist name in the data.

webserver/views/api_compat.py
pinkeshbadjatiya: I started with exactly the same idea, but then i had to switch to the current one because of the following reason.
- The generic `XML` to `JSON` converting function is using the standard conversion format and it uses generalized rules which are used for any other XML to JSON conversion. The XML object is quite simple to create rather than its corresponding JSON object, using `yattag`. Creating the JSON object for the same was quite messy. So, i preferred with the current approach.

PR #83, opened 2016-06-21
[WIP] Docker support

mayhem:
This WIP PR implements a good chunk of Docker support for ListenBrainz. It uses standard images for Postgres, Redis, and Kafa (through kafka will go away soon) and one customer docker file.

The only awkward thing about this setup is how MessyBrainz is handled. The MSB python module needs to be installed and the database needs to be created, but it lives in a different repo. Not sure what is best on how to solve this, but once the containers are started, you'll need to create the DBs using:

  docker-compose run web python manage.py init_db --skip-create
  docker-compose run web python manage.py init_msb_db

Other than the MSB issues, this is pretty straightforward stuff.


Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN pip install -r requirements.txt
+RUN python setup.py install
+
+# ListenBrainz
+RUN mkdir /code/listenbrainz
+WORKDIR /code/listenbrainz
+ADD requirements.txt /code/listenbrainz/
+RUN pip install -r requirements.txt
+ADD . /code/listenbrainz/
+ADD config.py /code/listenbrainz
```
gentlecat: Some of these `ADD`s seem redundant.

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    image: postgres:9.5.3
+    volumes:
+       - ./pgdata:/pgdata
+    environment:
+       PGDATA: /pgdata
+       POSTGRES_USER: listenbrainz
+       POSTGRES_PASSWD: listenbrainz
+       POSTGRES_DB: listenbrainz
+  redis:
+    image: redis
```
gentlecat: You should reference specific versions of images.

Dockerfile
mayhem: I think I can get rid of one.

mayhem:
Real PR submitted.

PR #85, opened 2016-06-25
Create wrapper function for api_tools + Re-factor

pinkeshbadjatiya:
- Re-factor code and create a wrapper function to access private api_calls.
- Fix wrong function name


mayhem:
Sadly, this didn't turn out as I had hoped, but I did a poor job of describing what I wanted. Let's chat on IRC and give this another shot. Sorry!

PR #88, opened 2016-07-13
Split listen table: listen -> listen + listen_json

pinkeshbadjatiya:


listenstore/listenstore/listenstore.py
```diff
@@ -31,12 +31,10 @@ def __init__(self, conf):
     def max_id(self):
         return int(self.MAX_FUTURE_SECONDS + calendar.timegm(time.gmtime()))
 
-
-    def fetch_listens_from_storage():
+    def fetch_listens_from_storage(*args):
```
gentlecat: Can you explain this?

listenstore/listenstore/listenstore.py
```diff
@@ -89,23 +87,31 @@ def insert_postgresql(self, listens):
...
-        query = """ SELECT id, user_id, extract(epoch from ts), artist_msid, album_msid, recording_msid, raw_data """ + \
-                """ FROM listen WHERE user_id = %(user_id)s """ + \
-                """ AND extract(epoch from ts) > %(from_id)s AND extract(epoch from ts) < %(to_id)s  """ + \
-                """ ORDER BY extract(epoch from ts) """ + ORDER_TEXT[order] + """ LIMIT %(limit)s"""
+        query = """ SELECT listen.id, user_id, extract(epoch from ts), artist_msid, album_msid, recording_msid, data
+                    FROM listen, listen_json
+                    WHERE listen.id = listen_json.id AND user_id = %(user_id)s AND extract(epoch from ts) > %(from_id)s
+                        AND extract(epoch from ts) < %(to_id)s
+                    ORDER BY extract(epoch from ts) """ + ORDER_TEXT[order] + """ LIMIT %(limit)s
+                """
```
gentlecat: This query needs to be formatted a bit better. Split long lines, align statements. See https://github.com/metabrainz/acousticbrainz-server/blob/master/db/dataset_eval.py#L68-L73 for example. @alastair might have more suggestions.

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: The overridden method had arguments but this base method did not. I added `*args` to denote that this method accepts multiple arguments.

gentlecat: Is that actually required?

pinkeshbadjatiya: not really.

listenstore/listenstore/listenstore.py

listenstore/listenstore/listenstore.py
```diff
@@ -106,19 +110,18 @@ def execute(self, query, params={}):
...
-                    WHERE listen.id = listen_json.id AND user_id = %(user_id)s AND extract(epoch from ts) > %(from_id)s
-                        AND extract(epoch from ts) < %(to_id)s
-                    ORDER BY extract(epoch from ts) """ + ORDER_TEXT[order] + """ LIMIT %(limit)s
-                """
-        params = {
+        results = self.execute("""
+            SELECT listen.id, user_id, extract(epoch from ts), artist_msid, album_msid, recording_msid, data
+              FROM listen, listen_json
+             WHERE listen.id = listen_json.id AND user_id = %(user_id)s AND extract(epoch from ts) > %(from_id)s
+               AND extract(epoch from ts) < %(to_id)s
```
gentlecat: If you have to put one of the statements on a new line, make every statement in `WHERE` be on a new line too.

listenstore/listenstore/listenstore.py
```diff
@@ -89,30 +87,48 @@ def insert_postgresql(self, listens):
...
+                     , extract(epoch from ts)
+                     , artist_msid
+                     , album_msid
+                     , recording_msid
+                     , data
+                  FROM listen
+                     , listen_json
+                 WHERE listen.id = listen_json.id
+                   AND user_id = :user_id
+                   AND extract(epoch from ts) > :from_id
```
alastair: A few things that I noticed here:
When you use a function on a where statement, postgres has to compute the value of the function for each row. This means that even if we have an index on `ts`, this query doesn't use it. See a query explication here: https://explain.depesz.com/s/fQTk
The part which checks the ts is listed as a Filter, whereas the part which checks the username is an Index Scan.
There are two ways to fix this:
* Create an index which includes the function name (this is possible in postgres) - `create index x on listen(extract(epoch from ts))`
* [My preference]: Use a comparison against a `datetime` object here too, which postgres will interpret as a timestamp. This is like the opposite of the change we made where we select the timestamp as a datetime object, keeping its semantic meaning

alastair: You've called the query parameters here `from_id` and `to_id`. I suspect this is something that was copied from cassandra where the timestamp actually was the id field of our data. When you copy code like this, please also think about the meaning that you want to convey with variable names. This is no longer an ID. The parameter would make much more sense if you named it `from_ts` or `from_time`

listenstore/listenstore/listenstore.py
```diff
@@ -89,30 +87,48 @@ def insert_postgresql(self, listens):
...
+                     , user_id
+                     , extract(epoch from ts)
+                     , artist_msid
+                     , album_msid
+                     , recording_msid
+                     , data
+                  FROM listen
+                     , listen_json
+                 WHERE listen.id = listen_json.id
+                   AND user_id = :user_id
```
alastair: I just noticed this, which we didn't pick up in the first review of the database code - It's common for fields which end in `_id` to be foreign keys to a separate table. In this case, I would expect `user_id` to be an integer field which references the `user` table instead of a string like it is now.
Normally in databases we don't duplicate data. This is called database normalization. The idea is that by referencing the user_id column to just a number which is the id for the user table, we could easily update this field by changing just the name in the user table (although this example is a bit forced because we don't anticipate changing usernames, the general concept remains the same)

listenstore/listenstore/listenstore.py
```diff
@@ -89,30 +87,48 @@ def insert_postgresql(self, listens):
...
+                 LIMIT :limit
+            """), {
+                'user_id': user_id,
+                'from_id': from_id,
+                'to_id': to_id,
+                'limit': limit
+            })
+
+            for row in results.fetchall():
+                yield self.convert_row(row)
```
alastair: This yield is due to the way that the cassandra library worked. We don't need it here, you can just return a list

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: - I was not aware that you can pass the datetime objects the the query string. I thought only strings are allowed. Committed the changes. (I have used datetime objects on comparison but still i am using `extract(epoch from ts)` while returning the timestamps. The rest of the changes that make listenstore return datetime objects instead of epoch's is already written in [this](https://github.com/metabrainz/listenbrainz-server/pull/89) PR.)
- This PR was meant to be small and simle splitting of the listen table so that i can re-use the code in the api_compat branch. Making such changes was just not the purpose of this PR as i have already made of lot of changes in the api_compat branch. This also has some problems.   
 
  For examle, Consider the above change. I have made changes in the same part of code in api_compat branch, now i also submitted a PR with other changes in show-local-time-to-user branch and finally this PR which also does substantial changes in the same part of code. I now have 3 branches that are open and their changes are dependent on each other, making things difficult for me to manage. So i always aim at making one branch for formatting (for now it is api_compat) while the other branches that make the ground for the updates which i then use in the api_compat. I rebase the branch with those changes as they merge and continue working.
  
Please lets just stick to the main purpose of this PR for now. I'll appreciate your feedback on formatting on the api_compat branch.

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: Even i was confused at first as everywhere in the code, `user_id` denotes the username of the user not a foreign key. I'll try to check if all of these can be changes without much effort else i'll postpone them for now and wait for other works to be finished and then come back later.

listenstore/listenstore/listenstore.py

listenstore/listenstore/listenstore.py
```diff
@@ -119,16 +119,18 @@ def fetch_listens_from_storage(self, user_id, from_id, to_id, limit, order, prec
...
+                   AND ts > :from_ts
+                   AND ts < :to_ts
+              ORDER BY ts """ + ORDER_TEXT[order] + """
                  LIMIT :limit
             """), {
                 'user_id': user_id,
-                'from_id': from_id,
-                'to_id': to_id,
+                'from_ts': datetime.utcfromtimestamp(from_ts),
+                'to_ts': datetime.utcfromtimestamp(to_ts),
```
gentlecat: What types are `from_ts` and `to_ts`? This should be documented.

pinkeshbadjatiya: `float`. Added in the code.

listenstore/listenstore/listenstore.py
alastair: You're right - I've looked at some of the LB code and we call it `user_id` in the API especially. This again is probably a result of the original cassandra code.
However, more important here is the fact that you have a field called `user_id` in the _database_. This has to be fixed before we release the postgres version of listenbrainz as a beta. If you don't want to complicate this PR, you can open a ticket to keep track of the task and do it once we've merged this one.

listenstore/listenstore/listenstore.py
alastair: Right - I'm giving you quite a bit of info here, and I'm doing it in a slightly unrelated PR, however I think that putting it in a PR right now is better because it gives you more context about what I'm saying (I also noticed it when I was reviewing this PR).
It would be fine for you to respond to this by making a ticket in Jira, copying this information, and saying that you'll put it on your list next for when this PR is merged. These things have to be fixed before we release postgres into beta, and so I want to get the info to you as quickly as I can.

alastair: A clarification on this comment, for your information:

> I was not aware that you can pass the datetime objects to the query string. I thought only strings are allowed. 

This is a feature that psycopg2 provides, postgresql doesn't understand python objects. The database driver converts this object to a format that the database engine understands. In this case, a string.

listenstore/listenstore/listenstore.py
alastair: Does it make sense for this value to be a float? What does a timestamp of `1469203685.5` mean?

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: I used the `user_id` field in the table to make things consistent. I have created a ticket to follow the progress.
Ticket: http://tickets.musicbrainz.org/browse/LB-97

listenstore/listenstore/listenstore.py
```diff
@@ -106,6 +106,13 @@ def insert_postgresql(self, listens):
                     self.log.error(e)
 
     def fetch_listens_from_storage(self, user_id, from_ts, to_ts, limit, order, precision):
+        """ The timestamps are stored as UTC in the postgres datebase while on retrieving
+            the value they are converted to the local server's timezone. So to compare
+            datetime object we need to create a object in the same timezone as the server.
```
alastair: Timezones are hard!
Just to let you know, I have a gut feeling that this is wrong. I'm going to look into it and get back to you.

Unix timestamps by definition are in UTC, so interpreting it as a local time is the wrong thing to do. Perhaps what we need to do is make sure the datetime object has an explicit timezone. I'll get back to you on this.

pinkeshbadjatiya: I tested it on my PC and on checking the output of the poastgres query from python i saw it had a `tzinfo` field (it was not None). And then loading using `utcfromtimestamp()` You can see the bug that the previous button on the user profile page will not vanish. The exact problem was, i was comparing the the created UTC datetime object against the postgres object. But postgres outputs the datetime in the local server's timezone (from the postgres documentation: https://ptpb.pw/bHsZ). There was again 2 ways that i could fix this,

**Solution 1:**


**Solution 2:**

listenstore/listenstore/listenstore.py
pinkeshbadjatiya: It is the postgres precision which does this. More specifically, earlier i was using `extract(epoch from ts)` which gave float precision, so typecasting it into int was leading to problems so i used float everywhere. 

But now, since we are switching to using datetime objects into the `listenstore`, then we would have to extract the `epoch` manually everywhere, then we can safely typecast it to int.

This problem can be solved easily once the whole listenstore migration is complete but since the listenstore is not yet complete, hence i can't finish this in this PR. btw, i created a ticket for the same. http://tickets.musicbrainz.org/browse/LB-98

listenstore/listenstore/listenstore.py
```diff
@@ -89,30 +88,63 @@ def insert_postgresql(self, listens):
                 if not listen.validate():
                     raise ValueError("Invalid listen: %s" % listen)
                 try:
-                    res = connection.execute(
-                    """ INSERT INTO listen(user_id, ts, artist_msid, album_msid, recording_msid, raw_data)
-                        VALUES ( %(user_id)s, to_timestamp(%(ts)s), %(artist_msid)s, %(album_msid)s,
-                        %(recording_msid)s, %(raw_data)s) ON CONFLICT DO NOTHING """, self.format_dict(listen))
+                    params = self.format_dict(listen)
```
gentlecat: I don't like how this dictionary with parameters is created in a function. This seems a bit pointless and makes it harder to follow the query process.

listenstore/listenstore/listenstore.py
```diff
@@ -72,12 +71,12 @@ def convert_row(self, row):
...
-                 'recording_msid': uuid.UUID(listen.recording_msid),
-                 'raw_data': ujson.dumps(listen.data)}
+        return {'user_id': listen.user_id,
+                'ts': listen.timestamp,
+                'artist_msid': uuid.UUID(listen.artist_msid),
+                'album_msid': uuid.UUID(listen.album_msid) if listen.album_msid is not None else None,
+                'recording_msid': uuid.UUID(listen.recording_msid),
+                'data': ujson.dumps(listen.data)}
 
     def insert_postgresql(self, listens):
```
gentlecat: `insert_postgresql` can be just called `insert` or `add`. The class is `PostgresListenStore` so everything inside uses PostgreSQL by default, there's no need to specify it here.

gentlecat: I know this is not directly related to this change. I must have missed it in previous code review.

listenstore/tests/util.py
```diff
@@ -3,7 +3,7 @@
 from datetime import datetime, timedelta
 from listenstore.listen import Listen
 import uuid
-
```
gentlecat: ✂️

listenstore/listenstore/listenstore.py

listenstore/listenstore/listenstore.py

pinkeshbadjatiya:
move to next line?

PR #89, opened 2016-07-19
[Fix: LB-96] Show listen time as per the user's timezone

pinkeshbadjatiya:
Fixes [LB-96](http://tickets.musicbrainz.org/browse/LB-96)  
Fix 2 issues:
- The listens should be displayed as per the user's timezone settings of the browser.
- Listenstore must return datetime objects not timestamps.

### DIFF
--- a/listenstore/listenstore/listen.py
+++ b/listenstore/listenstore/listen.py
@@ -23,7 +23,7 @@ class Listen(object):
     def from_json(cls, j):
         """Factory to make Listen() objects from a dict"""
         return cls(  user_id=j['user_id']
-                  , timestamp=datetime.utcfromtimestamp(j['listened_at'])
+                  , timestamp=datetime.utcfromtimestamp(float(j['listened_at']))
                   , artist_msid=j['track_metadata']['additional_info'].get('artist_msid')
                   , album_msid=j['track_metadata']['additional_info'].get('album_msid')
                   , recording_msid=j.get('recording_msid')
--- a/listenstore/listenstore/listen.py
+++ b/listenstore/listenstore/listen.py
@@ -1,7 +1,8 @@
 # coding=utf-8
 from __future__ import division, absolute_import, print_function, unicode_literals
-import datetime
 import ujson
+from datetime import datetime
+import calendar
 
 class Listen(object):
     """ Represents a listen object """
@@ -9,6 +10,7 @@ class Listen(object):
                  recording_msid=None, data=None):
         self.user_id = user_id
         self.timestamp = timestamp
+        self.ts_since_epoch = calendar.timegm(self.timestamp.utctimetuple())
         self.artist_msid = artist_msid
         self.album_msid = album_msid
         self.recording_msid = recording_msid
@@ -21,7 +23,7 @@ class Listen(object):
     def from_json(cls, j):
         """Factory to make Listen() objects from a dict"""
         return cls(  user_id=j['user_id']
-                  , timestamp=j['listened_at']
+                  , timestamp=datetime.utcfromtimestamp(j['listened_at'])
                   , artist_msid=j['track_metadata']['additional_info'].get('artist_msid')
                   , album_msid=j['track_metadata']['additional_info'].get('album_msid')
                   , recording_msid=j.get('recording_msid')
@@ -42,11 +44,11 @@ class Listen(object):
 
     @property
     def date(self):
-        return datetime.datetime.fromtimestamp(self.timestamp)
+        return self.timestamp
 
     def __repr__(self):
         return unicode(self).encode("utf-8")
 
     def __unicode__(self):
         return u"<Listen: user_id: %s, time: %s, artist_msid: %s, album_msid: %s, recording_msid: %s>" % \
-               (self.user_id, self.timestamp, self.artist_msid, self.album_msid, self.recording_msid)
+               (self.user_id, self.ts_since_epoch, self.artist_msid, self.album_msid, self.recording_msid)
--- a/listenstore/listenstore/listen.py
+++ b/listenstore/listenstore/listen.py
@@ -10,7 +10,7 @@ class Listen(object):
                  recording_msid=None, data=None):
         self.user_id = user_id
         self.timestamp = timestamp
-        self.ts_since_epoch = calendar.timegm(self.timestamp.utctimetuple())
+        self.ts_since_epoch = calendar.timegm(self.timestamp.utctimetuple()) if self.timestamp else None
         self.artist_msid = artist_msid
         self.album_msid = album_msid
         self.recording_msid = recording_msid
--- a/listenstore/listenstore/listenstore.py
+++ b/listenstore/listenstore/listenstore.py
@@ -83,7 +83,7 @@ class PostgresListenStore(ListenStore):
                 try:
                     params = {
                         'user_id': listen.user_id,
-                        'ts': listen.timestamp,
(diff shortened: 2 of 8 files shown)


gentlecat:
I don't understand how you determine what time zone the user is in. This should probably be done in the browser.

pinkeshbadjatiya:
Updated using javascript.

listenstore/listenstore/listenstore.py
```diff
@@ -31,7 +31,6 @@ def __init__(self, conf):
     def max_id(self):
         return int(self.MAX_FUTURE_SECONDS + calendar.timegm(time.gmtime()))
 
-
```
alastair: !
Regardless of if this change was needed or not, it's also a good idea to consider the context of each commit you are making. If you make a change to an html template to change timezone layout, an unrelated change to listenstore doesn't make a lot of sense.
If you were specifically making a change to listenstore to help support the different timezones on the template then this would have been an OK change to make.

pinkeshbadjatiya: Sorry!
I always try to clean the code as much as possible. I just can't let these redundant lines pass into the code base. I always try to clean the code gradually, so with every PR i fix a lot of pylint issues and try to improve the logic.
If this needs to be removed then i'll remove it.

webserver/templates/user/profile.html
```diff
@@ -67,8 +67,18 @@ <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
   {{ super() }}
   <script src="{{ url_for('static', filename='js/lib/jquery.timeago.js') }}"></script>
   <script type="text/javascript">
+    var offset = parseInt(new Date().getTimezoneOffset())*-1;
```
alastair: This sounds like something that might be useful in more places than just the user profile for now. How about we move it to a javascript file as a separate function?
We don't have any custom javascript files in ListenBrainz yet (the import/scraper script is an exception because it requires the template engine).
You could copy the setup that we have in AcousticBrainz:
We use some javascript tools to compile javascript and less (css):
* The dependencies are at https://github.com/metabrainz/acousticbrainz-server/blob/master/package.json (@gentlecat or @mwiencek can tell you which dependencies we need for this, I think some of them may not be needed) 
* https://github.com/metabrainz/acousticbrainz-server/blob/master/gulpfile.js and https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/static/gulpfile.js are the build files. They should be able to be copied directly
* https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/static_manager.py is a helper which lets us import a file by its source name, and translate it to its "compiled" name
* https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/__init__.py#L61 is how we make this function available in templates
* https://github.com/metabrainz/acousticbrainz-server/blob/master/webserver/templates/base.html#L59 is an example of using it in templates

alastair:
Before merging this, I'd like to make sure that we fix the way that timestamps are retrieved from postgres. We currently have an issue where timestamps are extracted in UTC, converted to the local time, then interpreted as if they were in UTC again. @pinkeshbadjatiya is working on a fix to this now.

alastair:
much better. check the failing test and then this is OK to merge

pinkeshbadjatiya:
Fixed the tests!

listenstore/listenstore/listen.py
```diff
@@ -9,7 +9,7 @@ def __init__(self, user_id=None, timestamp=None, artist_msid=None, album_msid=No
                  recording_msid=None, data=None):
         self.user_id = user_id
         self.timestamp = timestamp
-        self.ts_since_epoch = calendar.timegm(self.timestamp.utctimetuple())
+        self.ts_since_epoch = calendar.timegm(self.timestamp.utctimetuple()) if self.timestamp else None
```
alastair: What does this mean? Why can a timestamp be `None`? What does it mean to have no timestamp value in the database?

pinkeshbadjatiya: The redis-consumer calls `Listen().from_json()` which needs to initialize the class before making the call to method, as its a `classmethod`. So i had 2 options, either to make it a staticmethod or do the above change.
Also when dealing with `playing_now` listen, they cannot have a timestamp. Currently they have current timestamp (i do not use this anywhere). But ideally their timestamps should be None. And they have an additional `playing_now` attribute to be true.

alastair: There are 2 things here, let's address them one at a time: 

the `Listen` object has always had `from_json` as a static method (e.g., [1 year ago it was there](https://github.com/metabrainz/listenbrainz-server/blob/930b50f51c0623b030434485f675a58ed0d6e096/listenstore/listenstore/listen.py#L20)), so let's use that to fix this problem. 
Putting this change here doesn't sound like the easiest solution to me. This goes back to the question of talking about meaning. What does it mean to have a Listen with no timestamp? with no other data? It sounds like an invalid  state. we have a way to create an object from a json block without running into this bug, so let's do that.

Regarding playing now Listens, how about we have a separate object which doesn't require a timestamp? We can have a Listen and PlayingNow. Perhaps they can share a common base class for their shared fields, but then Listen has this extra field for the timestamp, and PlayingNow has (...? I don't know. Perhaps a date which it expires at?)

pinkeshbadjatiya: - There can be no way a listen having no timestamp. Also the same with raw_data as it contains the json data. (I guess the reason why that condition was there was to make sure it always stays as dict after the datebase query)
- Regarding `playing_now`, the only difference between the normal listen and playing_now is the presence of `listened_at` attribute. As there must be some way to differentiate between the two on the display page so i created a `playing_now:bool`. Creating separate inherited object for playing_now seems to be a nice idea. The only reason i treated playing_now the same as the traditional listen was i wanted to use the existing functions, and not treat it differently as there is no difference actually.

listenstore/listenstore/listen.py
```diff
@@ -21,7 +22,7 @@ def __init__(self, user_id=None, timestamp=None, artist_msid=None, album_msid=No
     def from_json(cls, j):
         """Factory to make Listen() objects from a dict"""
         return cls(  user_id=j['user_id']
-                  , timestamp=j['listened_at']
+                  , timestamp=datetime.utcfromtimestamp(j['listened_at'])
```
gentlecat: Commas should be after arguments. This is not the same as SQL, Python supports trailing commas.

mayhem:
Rebase this if needed and do any fix ups, please.

pinkeshbadjatiya:
Rebased!
Ready for merge.

mayhem:
Same problem as before the latest fix:

consumer_1 | Traceback (most recent call last):
consumer_1 |   File "redis-consumer/redis-consumer.py", line 103, in <module>
consumer_1 |     rc.start(config.SQLALCHEMY_DATABASE_URI, config.REDIS_HOST)
consumer_1 |   File "redis-consumer/redis-consumer.py", line 62, in start
consumer_1 |     l = Listen().from_json(data)
consumer_1 |   File "/code/redis-consumer/../listenstore/listenstore/listen.py", line 26, in from_json
consumer_1 |     , timestamp=datetime.utcfromtimestamp(j['listened_at'])
consumer_1 | TypeError: a float is required
docker_consumer_1 exited with code 1

pinkeshbadjatiya:
Fixed it.

mayhem:
OK, thanks very much. Go update the blog post! :)

PR #91, opened 2016-07-20
[Fix: LB-89] BadRequest error when importing while now_playing

pinkeshbadjatiya:
Fixes [LB-89](http://tickets.musicbrainz.org/browse/LB-89)
The audioscrobbler API's output differs when requesting listens using getrecenttracks method in case when the user is playing song. In case of `now_playing` to be true, the API outputs 1st listen as the now_playing.  
When listening: `{"@attr": {"now_playing":"true"}}`
When not listening: `{"date": {"utc":"12345756", "#text":"21 Jul 2016, 10:22"}}`

### DIFF
--- a/webserver/templates/user/scraper.js
+++ b/webserver/templates/user/scraper.js
@@ -15,7 +15,10 @@ function select(selector, collection) {
 function map(applicable, collection) {
     var newCollection = [];
     for (var i = 0; i < collection.length; i++) {
-        newCollection.push(applicable(collection[i]));
+      var result = applicable(collection[i]);
+      if ('listened_at' in result) {
+        newCollection.push(result);
+      };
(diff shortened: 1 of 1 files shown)


alastair:
This looks good. Can you add a comment which explains why we do this (in case we come back to the code in future weeks or months and forget what it does)

pinkeshbadjatiya:
Updated the description for the same.

alastair:
I meant a comment in the code. Please make sure you use similar words to what are in the code. For example, your code says:

    if ('listened_at' in result) {

and nowhere in the description you added you use the word `listened_at`, only `now_playing`

pinkeshbadjatiya:
Updated the description in code.

PR #92, opened 2016-07-21
[Fix: LB-80] Support playing_now submissions.

pinkeshbadjatiya:
Fix [LB-80](http://tickets.musicbrainz.org/browse/LB-80): **Implement now playing stream**
This PR adds support for the much needed `playing_now` feature for listens.

- Made some changes in the validation for listen payload. The **playing_now** submission need not have `listened_at` attribute.  
- Store playing_now listen in **redis** with key as `playing_now:<user_id>`. The max expiration time is specified by **PLAYING_NOW_MAX_DURATION**. If the user submits the duration of the song then the expiration time is set appropriately otherwise the max time specified.
(The whole discussion can be followed [here](http://tickets.musicbrainz.org/browse/LB-80) ).
- Use the defined constants for checking the listen_type instead of strings.
- Display currently playing song on the user profile page.
_r-t

## Testing
- Since there is no easy way to submit `playing_now` listen to LB in the current setup (Other than manually doing the POST call), so i have created a small patch to make things easier. It submits the playing_now listen which is playing on last.fm from the scraper itself. So, it'll submit all the existing listens along with the now_playing one.
- [Link to Patch](https://ptpb.pw/0Pcv.patch)


webserver/views/api.py
```diff
@@ -29,6 +29,8 @@
 
 MAX_ITEMS_PER_MESSYBRAINZ_LOOKUP = 10
 
+# Max time in seconds after which the playing_now stream will expire.
+PLAYING_NOW_MAX_DURATION = 5 * 60
```
alastair: this should be in the config file

webserver/__init__.py
```diff
@@ -10,6 +10,10 @@ def create_postgres():
     from postgres_connection import init_postgres_connection
     return init_postgres_connection(current_app.config['SQLALCHEMY_DATABASE_URI'])
 
+def create_redis():
```
gentlecat: Where is this used?

webserver/templates/user/profile.html
```diff
@@ -31,11 +31,21 @@ <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
             </thead>
             <tbody>
             {% for listen in listens %}
+
+              {% if listen.playing_now %}
+              <tr id="playing_now">
+                <td>{{ listen.track_metadata.artist_name }}</td>
+                <td>{{ listen.track_metadata.track_name }}</td>
+                <td><span class="glyphicon glyphicon-play" aria-hidden="true"></span> Playing now</td>
+              </tr>
```
gentlecat: Indentation.

webserver/__init__.py
pinkeshbadjatiya: https://github.com/metabrainz/listenbrainz-server/pull/92/files#diff-4f01280a2f2f266a23f31d0a1f922424R45

listenstore/listenstore/listenstore.py
```diff
@@ -116,3 +117,19 @@ def fetch_listens_from_storage(self, user_id, from_id, to_id, limit, order, prec
...
+        self.redis = Redis(conf['REDIS_HOST'])
+
+    def get_playing_now(self, user_id):
+        """ Return the current playing song of the user """
+        data = self.redis.get('playing_now' + ':' + user_id)
+        if not data:
+            return None
+        data = ujson.loads(data)
+        data.update({'listened_at': datetime.utcnow()})
+        return Listen().from_json(data)
```

webserver/__init__.py
gentlecat: So you are doing that on every request to that endpoint? Why not do it once during app initialization process?

gentlecat: In fact, it's already being done: https://github.com/metabrainz/listenbrainz-server/blob/master/webserver/__init__.py#L32. Is there some kind of issue with that?

pinkeshbadjatiya: > So you are doing that on every request to that endpoint? 
Yes.

> Why not do it once during app initialization process?
The same thing happens with the postgres connection as well. So to make things consistent i sticked with the same solution. Also what happens if we create a connection everytime?

> In fact, it's already being done: https://github.com/metabrainz/listenbrainz-server/blob/master/webserver/__init__.py#L32. Is there some kind of issue with that?
Nope. Actually we do have a persistent redis connnection as well with global name as `_redis`, its just the code is getting confusing. On some other branch i made those changes, so i did not want to create duplicates.

pinkeshbadjatiya: I'll update it here as well.

pinkeshbadjatiya: Before i update it, i just wanted to clarify one thing.
The code creates 2 connection as well for the postgres database.
1. To display the listens, which is create at every request. (https://github.com/metabrainz/listenbrainz-server/blob/master/webserver/views/user.py#L42)
2. A persistent connection from `redis_consumer` that inserts the listens.(https://github.com/metabrainz/listenbrainz-server/blob/master/redis-consumer/redis-consumer.py#L31)

So i followed the similar logic,
1. A temporary redis connection to fetch `playing_now`
2. A persistent connection from `redis_consumer` that pushes listens from the redis queue to postgres.

Was this done intentionally?

gentlecat: SQLAlchemy engine should be created once during app creation. Each set of requests is then supposed to create a connection from it.

webserver/templates/user/profile.html
pinkeshbadjatiya: fixed!

gentlecat: Doesn't look like you did. :) Code inside `if`s, `for`s, and other blocks should be indented.

pinkeshbadjatiya: oh, i got it wrong. You mean like this?
```
{% for listen in listens %}
    {% if listen.playing_now %}
        <tr id="playing_now">
        ....
```

webserver/__init__.py
pinkeshbadjatiya: @alastair @gentlecat i did some changes to create single connections for `redis` (RedisListenStore) and `postgres` (PostgresListenStore). It might need some more changes.
Changes: https://github.com/metabrainz/listenbrainz-server/pull/92/commits/b6a17c890d58cb9d6254b36277252d75f8ed1f1f

webserver/views/api.py
pinkeshbadjatiya: done!

config.py.sample
```diff
@@ -37,6 +37,9 @@ MUSICBRAINZ_CLIENT_SECRET = "CLIENT_SECRET"
 LASTFM_API_KEY = "USE_LASTFM_API_KEY"
 
 
+# Max time in seconds after which the playing_now stream will expire.
+PLAYING_NOW_MAX_DURATION = 5 * 60
```
mayhem: 5 minutes is not enough:

  https://musicbrainz.org/release/3591d22e-fdc5-309a-91d3-7bbafe086d6c/disc/1#7aefdb30-ad00-3cde-9072-4e43f4253637

A standard CD is 80 minutes. Let's set this to 80 minutes, which out to be sufficient for now.

webserver/views/api.py
```diff
@@ -248,20 +250,30 @@ def _messybrainz_lookup(listens):
     return augmented_listens
 
 
-def _validate_listen(listen):
+def _validate_listen(listen, listen_type):
```
mayhem: I would prefer to have a numerical constant, rather than a string here. It is much faster to use an int that a string for this. Define and use the LISTEN_TYPE_SINGLE, LISTEN_TYPE_IMPORT and LISTEN_TYPE_NOW_PLAYING.

mayhem:
Overall, this looks good. I would like to see some tests added that test the new code paths.

config.py.sample
alastair: We discussed this in IRC, and a little in the ticket: http://tickets.musicbrainz.org/browse/LB-80
Our other options for the length are
* Use a track length if it's submitted with the listen
* get the length from MB if we have a mbid
* otherwise use some fallback

We should remove the existing now playing if we add another one (since it's a single redis key, it should just get replaced), but we should also remove it if we get a listen submitted with a time within 80 minutes.
Honestly, I think 80 minutes is not a good default. What about the average of all our track lengths in musicbrainz + 1 standard deviation? (although, I have a hunch that that'll be around 5 minutes anyway). We're going to be "more correct" more of the time with a < 10 minute default than with 80.

webserver/views/api.py
alastair: Good point about the constants, I didn't pick that up. Though, I'm suspicious that checking for string membership in a set would be (much) slower than ints.

config.py.sample
mayhem: Ah, I see your point. 

Yes, the removing the last now_playing happens when the next one comes in, the key is overwritten. But you make a good point about what happens when a user stops listening. 80 mins is a bit far out. I would say we should keep it simple for now. I'd say go with 10 minutes, rather than 5. Do I hear a 7.5? :)

alastair: I'm happy with 10
@pinkeshbadjatiya Make sure that you clear playing now if a new listen comes in before the playing now key expires (I don't think we have this)

Let's add a new ticket (or leave LB-80 open with a note) for the more advanced aspects (getting song length from listen metadata or musicbrainz)

pinkeshbadjatiya: the playing_now song is stored with userid as the key, so it gets overwritten when the user listens to a new song with the new expiration as set by us. This is already implemented and works fine.
I'll change the value to 10 mins.

webserver/views/api.py
pinkeshbadjatiya: - The set size is just 3 so it would hardly make any difference.
- Updated the code to use the 3 defined variables.

gentlecat:
@brainzbot, build this please.

gentlecat:
@brainzbot, test this please.

PR #94, opened 2016-07-23
Docker cleanup

gentlecat:


.dockerignore
```diff
@@ -9,26 +9,17 @@ env
...
+/data
 
 # Test results
 htmlcov
 .coverage
 
 # Vagrant
 .vagrant
-
-.transifexrc
```
gentlecat: Hello CB. ✋

README.md
```diff
@@ -54,8 +54,8 @@ This should start the containers and allow you to setup the database.
 
 To initialize the database (create user, tables, etc.) run these commands:
 
-    $ docker-compose run web python manage.py init_db --skip-create
-    $ docker-compose run web python manage.py init_msb_db 
+    $ docker-compose run web python manage.py init_db 
+    $ docker-compose run web python manage.py init_msb_db --create-db
```
gentlecat: I think MSB should be a completely separate Docker image.

alastair: yeah, ruaok and I discussed this a little bit. We should be able to have a docker image for messybrainz which sets up the database, and then (somehow) install the messybrainz database code into the LB env so that it can connect to that database directly

gentlecat: Why not have a separate database for MSB?

mayhem: I think we should work to close some of the open PRs are get armacolite's branch merged before we open the MSB can of worms. But yes, the current state of things is less than ideal.

docker-compose.prod.yml
```diff
@@ -0,0 +1,54 @@
...
+      - db
+
+  consumer:
+    build: 
+      dockerfile: redis-consumer/Dockerfile
+      context: .
+    command: python redis-consumer/redis-consumer.py
+    volumes:
+      - .:/code/listenbrainz-server
+    links:
```
mayhem: Why do you have a links section in the prod version, but not in the normal version?

docker-compose.prod.yml
```diff
@@ -0,0 +1,54 @@
...
+      PGDATA: /var/lib/postgresql/data/pgdata
+
+  redis:
+    image: redis:3.2.1
+    volumes:
+      - ./data/redis:/data
+
+  web:
+    build: .
+    command: uwsgi /code/listenbrainz-server/docker/uwsgi.ini
```
mayhem: I like how the prod version uses uwsgi, but the default is to use server.py. Nice!

Dockerfile
```diff
@@ -1,15 +1,11 @@
 FROM python:2.7.11
 
-MAINTAINER Robert Kaye <rob@metabrainz.org>
```
gentlecat: Didn't actually remove it. It's just the file got moved and diffs here are displayed differently. https://github.com/metabrainz/listenbrainz-server/pull/94/files#diff-0e9722a692af2b244674b56146a1ecd3R3

I don't think it's important in our case. Everyone is working on these files anyway.

README.md
gentlecat: 👌

docker-compose.yml
```diff
@@ -1,51 +1,37 @@
...
     image: postgres:9.5.3
     volumes:
-       - ./pgdata:/pgdata
+      - ./data/pgdata:/var/lib/postgresql/data/pgdata
     environment:
-       PGDATA: /pgdata
-       POSTGRES_USER: listenbrainz
-       POSTGRES_PASSWD: listenbrainz
-       POSTGRES_DB: listenbrainz
+      PGDATA: /var/lib/postgresql/data/pgdata
```
mayhem: Why do we not need these env vars in the -prod version?

docker-compose.prod.yml
gentlecat: That's from original file which I didn't change much. I suspect that `links` might be redundant when there's `depends_on` section with the same set of items. https://docs.docker.com/compose/compose-file/#links

redis-consumer/Dockerfile
```diff
@@ -3,18 +3,10 @@ FROM python:2.7.11
 MAINTAINER Robert Kaye <rob@metabrainz.org>
 
 # General setup
-RUN apt-get update && apt-get install -y --no-install-recommends python-redis redis-tools vim
```
mayhem: Why do not need the PG client here? Is this for the command line tools only and sqlalchemy doesn't need this?

docker-compose.prod.yml
mayhem: Then lets remove the redundant links section.

docker-compose.yml
gentlecat: They should be there. I added a reference to `common.yml`, but forgot to define them.

redis-consumer/Dockerfile
gentlecat: Correct. Since `redis-consumer` doesn't use psql, we don't need to install client here. It can still access the database.

mayhem: 👍

docker-compose.prod.yml
```diff
@@ -0,0 +1,49 @@
+version: "2"
+services:
+
+  db:
+    image: postgres:9.5.3
```
mayhem: the "restart: always" is missing from all the definitions in this compose file.

mayhem:
The restart bits and moving the docker and docker compose files to the docker dir is all that is needed to merge this now.

gentlecat:
OK, that is done.

mayhem:
File layout looks good now, but I can't get the containers to come up:

https://gist.github.com/mayhem/1d68872b942394701db480d35ac8aba4

The message about redis does not appear when I use the non-prod docker compose file.

PR #100, opened 2016-07-29
[WIP] Add basic support for a vagrant VM

mayhem:
First cut at supporting vagrant for listenbrainz. A simple "vagrant up" should do the trick.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,18 @@ ListenBrainz home page.
 
 ### Virtual machine
 
-There will be an easy to use Vagrant setup once the docker branches are merged.
+To use a vagrant setup of the listenbrainz server, install the lastest VirtualBox and the latest Vagrant:
+
+* https://www.virtualbox.org/wiki/Downloads
+* https://www.vagrantup.com/downloads.html
+
+Then build the VM using:
+
+    $ vagrant up
+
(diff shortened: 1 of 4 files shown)


gentlecat:
Not sure I understand, but why do we need a Vagrant VM when there's Docker?

mayhem:
I think there are a few reasons to have a VM: I didn't have much luck with docker on mac and as far as I know there is no docker for windows. Also, installing docker and running docker images is a shitty end user experience -- VB has a bit nicer end user experience than docker does now.

gentlecat:
There are both Mac and Windows versions: https://docs.docker.com/engine/installation/#/on-osx-and-windows.

I'm just trying to understand who's going to use this.

mayhem:
People who don't want to touch docker. Not everyone is geeky like us. Also, I personally despise doing dev work on my mac -- there are always weird problems. A clean VM that I can blow away whenever I want, is a very good working model for me.

I guess I am your target audience. :)

gentlecat:
Alright, sounds reasonable. What's left to be done there?

mayhem:
The postgres permissions thing -- let me see if I can sort that today and then yank this out of WIP.

PR #101, opened 2016-07-29
Add pytest.ini file

pinkeshbadjatiya:
Running py.test outside docker fails with the following error.
elua


mayhem:
The tests should be run inside docker, but if this helps....

gentlecat:
This should be useful with Docker too.

PR #102, opened 2016-08-10
Change listens table merged

mayhem:
This is the change-listens-table branch with the user_id -> int change.


gentlecat:
:shipit:

PR #103, opened 2016-10-02
fix PG_version for docker

samarth:
On doing the docker-compose build , I saw 

E: Version '9.5.4-1.pgdg80+1' for 'postgresql-client-9.5' was not found

which ultimately failed to build the Service "web" . 

I found that the[ postgres version on docker-library ](https://github.com/docker-library/postgres/blob/master/9.5/Dockerfile)was updated to the one I've listed on the PR. 

This fixed the issue .

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update \
 # PostgreSQL client
 RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8
 ENV PG_MAJOR 9.5
-ENV PG_VERSION 9.5.4-1.pgdg80+1
+ENV PG_VERSION 9.5.4-1.pgdg80+2
 RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


gentlecat:
@brainzbot, ok to test.

alastair:
Is it recommended in Docker to pin exact version numbers, or should we just install whatever 9.5 version is available?

samarth:
@alastair I couldn't find if its a recommended practice or not . IMHO , we're overkilling the installation of the postgres client by specifying a minor version . I think we should be cool without pinning the exact number.

mayhem:
@alastair Yes, you want to nail down all version numbers that you can nail down. This gives the best results for complete isolation and repeatability of containers.

samarth:
@mayhem  I agree but when the author first pinned the minor version for post gres , the docker-compose build broke because the version was scrapped off from the postgres ppa . 

This is fixed for now but sometime later we would be facing the same problem again . Can't we just use a major version in this case since it is only the postgres-client ?

mayhem:
In theory that should be ok, but I personally prefer manually fixing/testing when the change happens.

Ideally, the upstream should never destroy the existing containers, since there are people likely to depend on it.

samarth:
Yes , that is an unlikely scenario which happened . 

Fine , I am going to keep it like this then .

ferbncode:
the [postgres version](https://github.com/docker-library/postgres/blob/master/9.5/Dockerfile) seems to be updated again.

PR #104, opened 2016-11-02
Minor fix in documentation: Change token to Token.

paramsingh:
The authorization header in the API should be of the format
"Token <user token>" instead of "token <user token>"

### DIFF
--- a/webserver/views/api.py
+++ b/webserver/views/api.py
@@ -23,7 +23,7 @@ def submit_listen():
 
     For complete details on the format of the JSON to be POSTed to this endpoint, see :ref:`json-doc`.
 
-    :reqheader Authorization: token <user token>
+    :reqheader Authorization: Token <user token>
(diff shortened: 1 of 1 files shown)


pinkeshbadjatiya:
Looks good to me as well.
(Though, we never check the 1st argument so it should be fine. https://github.com/paramsingh/listenbrainz-server/blob/0b05981871ef2e2860e57dbd1d4ba265077a71d4/webserver/views/api.py#L131 )

paramsingh:
@pinkeshbadjatiya Yea, I was on IRC and came across it while talking with alastairp. He said it was more of a convention thing and I had free time on my hands, so a small pull request :)

pinkeshbadjatiya:
@paramsingh no problem. Saw the conversation after my comment. :)

gentlecat:
@brainzbot, ok to test.

alastair:
Thanks for the contribution!

PR #106, opened 2016-12-30
LB-113: Update robots.txt for Scrobble API endpoint

dpmittal:
Changed the robots.txt file to block path to new Lastfm API and Auth pages
Issue Link: https://tickets.metabrainz.org/browse/LB-113

### DIFF
--- a/webserver/static/robots.txt
+++ b/webserver/static/robots.txt
@@ -2,3 +2,5 @@ User-agent: *
 Disallow: /current-status
 Disallow: /user/
 Disallow: /1/
+Disallow: /api/
+Disallow: /2.0/


webserver/static/robots.txt
```diff
@@ -2,3 +2,5 @@ User-agent: *
 Disallow: /current-status
 Disallow: /user/
 Disallow: /1/
+Disallow: /api
+Disallow: /*lastfm
```
pinkeshbadjatiya: Why is this line needed?

dpmittal:
@pinkeshbadjatiya Made the changes

webserver/static/robots.txt
```diff
@@ -2,3 +2,5 @@ User-agent: *
 Disallow: /current-status
 Disallow: /user/
 Disallow: /1/
+Disallow: /api
```
Freso: Is there a reason for using `/api` here and not `/api/`? (I'm a bit rusty on my robots.txt syntax.)

dpmittal: There's a file /api.py which needs to be blocked for bots . /api means every file will be blocked with '/api' where '/api/' will block /api/ folder (which is not present in LB) not a file :)

pinkeshbadjatiya:
This looks good to me.

webserver/static/robots.txt
gentlecat: You are blocking URLs, not files or folders. It is better to have a trailing slash everywhere for consistency sake and to avoid blocking matching prefixes like `/apistuff`.

dpmittal: Actually, I didn't know about robots.txt when I googled about it it said /filename /foldername/ so i thought robots.txt worked like that  but now i came to know that it blocks urls... I have changed the file now

dpmittal:
@mayhem @gentlecat I have made the changes :)

PR #107, opened 2017-01-08
[LB-102] Improve Select query and constructor of Session

paramsingh:
Fix for LB-102. Removed some of the "magic" by making the Session constructor take all 5 values explicitly as arguments and the select query now names all the columns it requires.

I think I've done everything talked about [here](https://github.com/metabrainz/listenbrainz-server/pull/82#discussion-diff-71160085R31).

### DIFF
--- a/db/lastfm_session.py
+++ b/db/lastfm_session.py
@@ -9,8 +9,11 @@ from lastfm_user import User
 class Session(object):
     """ Session class required by the api-compat """
 
-    def __init__(self, row):
-        self.id, userid, self.sid, self.api_key, self.timestamp = row
+    def __init__(self, id, userid, sid, api_key, timestamp):
+        self.id = id
+        self.sid = sid
+        self.api_key = api_key
+        self.timestamp = timestamp
         self.user = User.load_by_id(userid)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


db/lastfm_session.py
```diff
@@ -29,7 +36,8 @@ def load(session_key, api_key):
             })
             row = result.fetchone()
             if row:
-                return Session(row)
+                id, userid, sid, api_key, timestamp = row
```
alastair: Using sqlalchemy, `row` should also be accessible as a dictionary. It would make this line more explicit. What about `return Session(r["id"], r["userid"], ...)`?

paramsingh:
@alastair Done. 

Also, small question: what's the convention on number of columns in one line? Are we using PEP-8 80 lines or are more columns allowed?

alastair:
Cool, thanks!

Lines longer than 80 columns are fine. Just looking at what @gentlecat did for acousticbrainz, it looks like we are using around 130: https://github.com/metabrainz/acousticbrainz-server/blob/master/pylintrc#L31
I normally line wrap when the lines get around 100-120, but won't wrap if the new line only has a few characters.

paramsingh:
> Lines longer than 80 columns are fine. Just looking at what @gentlecat did for acousticbrainz, it looks like we are using around 130: https://github.com/metabrainz/acousticbrainz-server/blob/master/pylintrc#L31
I normally line wrap when the lines get around 100-120, but won't wrap if the new line only has a few characters.

Okay, cool thanks.

Also, are we waiting for someone else to approve the changes before you merge this?

PR #108, opened 2017-01-13
[LB-103] - Remove magic numbers from ERROR_MESSAGES dictionary.

paramsingh:
Fix for [LB-103](https://tickets.metabrainz.org/projects/LB/issues/LB-103).

One question: errors with code 1 and 19 have the same message ("This error does not exist"),  I've given them names DOES_NOT_EXIST and DOESNT_EXIST, but that doesn't look "right" to me. What would be the best course of action here?

### DIFF
--- a/webserver/errors.py
+++ b/webserver/errors.py
@@ -2,39 +2,53 @@ from flask import render_template, make_response
 from yattag import Doc
 import yattag
 import ujson
+import collections
 
+LastFMError = collections.namedtuple('LastFMError', ['code', 'message'])
 
 # List of errors compatible with LastFM messages for API_compat.
-ERROR_MESSAGES = {
-    1: "This error does not exist",
-    2: "Invalid service -This service does not exist",
-    3: "Invalid Method - No method with that name in this package",
-    4: "Invalid Token - Invalid authentication token supplied",
-    5: "Invalid format - This service doesn't exist in that format",
-    6: "Invalid parameters - Your request is missing a required parameter",
-    7: "Invalid resource specified",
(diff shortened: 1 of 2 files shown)


alastair:
I wonder if error 19 used to be an error message, but last.fm got rid of it: http://www.last.fm/api/errorcodes

I would be fine with calling them `DOES_NOT_EXIST` and `DOES_NOT_EXIST_19`

alastair:
These constants are also used when we raise an `InvalidAPIUsage`, you should use the constant here too.
https://github.com/metabrainz/listenbrainz-server/blob/52c4d795823f72d50bb02748a66b2cfc958beb23/webserver/views/api_compat.py#L97

I wonder if there is a cleaner way of creating this mapping, instead of using both the class to define the constants, and the dictionary to map the numbers to the messages. Perhaps we could have a specific class which represents the number, descriptive value, and error message?

alastair:
What about something like this?
```python
import collections
LfmError = collections.namedtuple('LfmError', ['code', 'message'])

class CompatError(object):
    FOO = LfmError(1, 'cannot do stuff')
    BAR = LfmError(2, 'status ok')

In [6]: error.CompatError.FOO
Out[6]: LfmError(code=1, message='cannot do stuff')

In [7]: error.CompatError.FOO.code
Out[7]: 1

In [8]: error.CompatError.FOO.message
Out[8]: 'cannot do stuff'
```

paramsingh:
@alastair How does this look?

I don't think I've missed any usage of these errors this time.

PR #109, opened 2017-01-14
Removing unnecessary AB icons

dpmittal:
There were unnecessary AcousticBrainz icons in the repo which needed to be removed


alastair:
Let's make the listenbrainz icons consistent with the ones in MessyBrainz too (https://github.com/metabrainz/messybrainz-server/pull/16/files)

dpmittal:
@alastair I have made the changes :)

mineo:
`favicon-256.png` looks extremely blurry compared to the previous version.

dpmittal:
@mineo Actually it was resized from the existing 32px file which was blurry so all the files became blurry. Now I have resized the files from [this](https://github.com/metabrainz/metabrainz-logos/blob/master/final-logos/ListenBrainz/VECTOR/SVG/ListenBrainz_logo_icon.svg) so now all the files are not blurry

dpmittal:
An now all the icons are squared which were not earlier :)

PR #112, opened 2017-01-19
[LB-115] Fix bug which didn't allow submission of scrobbles if api_sig was in request

paramsingh:
Small fix for LB-115.

Because the loop didn't skip api_sig, an extra "empty" track gets added to the lookup defaultdict, leading to errors in  messybrainz lookups later.

Fixed this by adding a continue for "api_sig" also. :)

### DIFF
--- a/webserver/views/api_compat.py
+++ b/webserver/views/api_compat.py
@@ -246,7 +246,7 @@ def record_listens(request, data):
 
     lookup = defaultdict(dict)
     for key, value in data.items():
-        if key == "sk" or key == "token" or key == "api_key" or key == "method":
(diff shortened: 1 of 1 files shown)


webserver/views/api_compat.py
```diff
@@ -246,7 +246,7 @@ def record_listens(request, data):
 
     lookup = defaultdict(dict)
     for key, value in data.items():
-        if key == "sk" or key == "token" or key == "api_key" or key == "method":
+        if key == "sk" or key == "token" or key == "api_key" or key == "method" or key == "api_sig":
```
gentlecat: Looks like all of these can be put into a list.

paramsingh: @gentlecat Makes sense. Done!

PR #113, opened 2017-01-27
Write a python script for moving data from LB alpha to beta

paramsingh:
/user/import/alpha pushes a task into a redis queue and
alpha_importer.py keeps track of the queue to find accounts for
which import needs to be done.

### DIFF
--- a/alpha_importer.py
+++ b/alpha_importer.py
@@ -2,9 +2,13 @@ from __future__ import print_function
 from redis import Redis
 import config
 import requests
+from requests.exceptions import HTTPError
 import ujson
 import time
 import sys
+import logging
+from logging.handlers import RotatingFileHandler
+import os
 
 redis_connection = Redis(host = config.REDIS_HOST)
 QUEUE_KEY = 'alphaimporter:queue'
@@ -12,6 +16,18 @@ SET_KEY_PREFIX = 'alphaimporter:set'
 ALPHA_URL = 'https://api.listenbrainz.org/'
 BETA_URL = 'http://0.0.0.0:3031'
 LISTENS_PER_GET = 100
+LOG_FILE = os.path.join(os.getcwd(), 'alpha_importer.log')
+DELAY = 3 # delay if requests don't return http 200
+
+# create a logger to log messages into LOG_FILE
+logger = logging.getLogger('alpha_importer')
+logger.setLevel(logging.DEBUG)
+handler = RotatingFileHandler(LOG_FILE, maxBytes = 512 * 1024, backupCount = 100)
+handler.setLevel(logging.DEBUG)
+formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
+handler.setFormatter(formatter)
+logger.addHandler(handler)
+
 
 def queue_empty():
     return redis_connection.llen(QUEUE_KEY) == 0
@@ -30,11 +46,11 @@ def get_batch(username, max_ts):
     if r.status_code == 200:
         return ujson.loads(r.text)
     else:
-        print("get from alpha http response code: {}".format(r.status_code), file = sys.stderr)
-        return None
+        logger.error("get from alpha http response code: {}".format(r.status_code))
+        raise HTTPError
 
 
-def get_data(batch):
+def extract_data(batch):
     data = {'listen_type':'import', 'payload': []}
     for listen in batch['payload']['listens']:
         data['payload'].append(listen)
@@ -44,30 +60,29 @@ def get_data(batch):
 def send_batch(data, token):
     send_url = '{}/1/submit-listens'.format(BETA_URL)
     r = requests.post(send_url, headers = {'Authorization': 'Token {}'.format(token)}, data = ujson.dumps(data))
-    if r.status_code == 200:
-        return 1
-    else:
-        print("submission to beta returned response code: {}".format(r.status_code), file = sys.stderr)
-        return 0
+    if r.status_code != 200:
+        logger.error("submission to beta returned response code: {}".format(r.status_code))
+        raise HTTPError
 
 
 def import_from_alpha(username, token):
     next_max = int(time.time())
-    cnt = 1
+    page_count = 1
     while True:
-        batch = get_batch(username, next_max)
-        if not batch: # if get_batch doesn't return data, we try again
-            continue
-        if batch['payload']['count'] == 0:
-            print("All pages done")
-            print("total number of pages = {}".format(cnt))
-            return 1
-        data = get_data(batch)
-        sent = send_batch(data, token)
-        if sent:
+        try:
+            batch = get_batch(username, next_max)
+            if batch['payload']['count'] == 0:
+                logger.info('All pages done for user {}'.format(username))
+                logger.info('Total number of pages done: {}'.format(page_count))
+                return True
+            data = extract_data(batch)
+            sent = send_batch(data, token)
             next_max = data['payload'][-1]['listened_at']
-            print("page #{} done".format(cnt))
-            cnt += 1
+            logger.info('Page #{} done.'.format(page_count))
+            page_count += 1
+        except HTTPError:
+            time.sleep(DELAY)
+
 
 def queue_pop():
     redis_connection.lpop(QUEUE_KEY)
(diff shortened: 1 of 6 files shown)


alpha_importer.py
```diff
@@ -0,0 +1,85 @@
...
+import requests
+import ujson
+import time
+import sys
+
+redis_connection = Redis(host = config.REDIS_HOST)
+QUEUE_KEY = 'alphaimporter:queue'
+SET_KEY_PREFIX = 'alphaimporter:set'
+ALPHA_URL = 'https://api.listenbrainz.org/'
+BETA_URL = 'http://0.0.0.0:3031'
```
paramsingh: This will need to be changed later, I guess.

paramsingh:
I haven't cleaned msids from the listen data received from alpha, should we do that as well?

mayhem:
I won't really have a chance to look until monday, sorry!

paramsingh:
@mayhem no prob, I'll see if I can find smaller bugs to fix until then!

alpha_importer.py
mayhem: Let's just pick what we plan to do:

ALPHA_URL: https://alpha.listenbrainz.org
BETA_URL: https://listenbrainz.org

alpha_importer.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def get_batch(username, max_ts):
+    get_url = '{}/1/user/{}/listens'.format(ALPHA_URL, username)
+    payload = {"max_ts": max_ts, "count": LISTENS_PER_GET}
+    r = requests.get(get_url, params = payload)
+    if r.status_code == 200:
+        return ujson.loads(r.text)
+    else:
+        print("get from alpha http response code: {}".format(r.status_code), file = sys.stderr)
```
mayhem: Rather than using a print command here, we should setup a separate logfile using the logging module. The logfile name can be dfined with the other defines near the top of the file.

alpha_importer.py
```diff
@@ -0,0 +1,86 @@
...
+    return data
+
+
+def send_batch(data, token):
+    send_url = '{}/1/submit-listens'.format(BETA_URL)
+    r = requests.post(send_url, headers = {'Authorization': 'Token {}'.format(token)}, data = ujson.dumps(data))
+    if r.status_code == 200:
+        return 1
+    else:
+        print("submission to beta returned response code: {}".format(r.status_code), file = sys.stderr)
```
mayhem: My logging comment clearly applies to all print() statements in this PR. :)

alpha_importer.py
```diff
@@ -0,0 +1,86 @@
...
+    for listen in batch['payload']['listens']:
+        data['payload'].append(listen)
+    return data
+
+
+def send_batch(data, token):
+    send_url = '{}/1/submit-listens'.format(BETA_URL)
+    r = requests.post(send_url, headers = {'Authorization': 'Token {}'.format(token)}, data = ujson.dumps(data))
+    if r.status_code == 200:
+        return 1
```
mayhem: It is more pythonic to return True or False, not 1/0. It is even more pythonic to Raise an exception if something goes wrong. In this case, if a submission fails, the script should wait 3 seconds (maybe configurable) and then try again.

alpha_importer.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def set_remove(username):
+    redis_connection.delete("{} {}".format(SET_KEY_PREFIX, username))
+
+if __name__ == '__main__':
+    while True:
+        if not queue_empty():
+            username, token = queue_front()
+            done = import_from_alpha(username, token)
```
mayhem: It is more pythonic to just check the return value in the If -- there is no need to create a new variable and then just use it once. In this case it would be cleaner to just check the return of the if. However, it would be even better to consider using Exceptions.

alpha_importer.py
```diff
@@ -0,0 +1,86 @@
...
+            if done:
+                queue_pop()
+                set_remove(username)
```
mayhem: Have you considered what happens if the user's data cannot be fully imported? We need to be careful to import all of their data and be 100% certain of that. If we can't be certain we need to inform the user that we failed to import their data.

paramsingh: @mayhem: right now, what the script does is wait for 3 seconds if no reply from either of the servers and resends the requests until all pages are done. So, if we reach the code here, we're sure that everything has been imported. How do I come to know that user's data hasn't been fully imported?

paramsingh:
Logging and try-catch instead of if-else done. Apologies for the less Pythonic code, I think I've fixed all problems mentioned though.

I'll get to work on the email code today, I plan to use python's `email`  and `smtplib` modules for this. I do have a few questions about this though that I hope you'll answer: 

1. We'll have to set up a smtp server on the docker containers for this, right?
2. How do I get the email address of the user? From a cursory look, the User class in `webserver.login` only has `musicbrainz_id` in it.

docker/Dockerfile.prod
```diff
@@ -42,4 +42,4 @@ RUN pip install -r requirements.txt
 # Now install our code, which may change frequently
 COPY . /code/listenbrainz
 
-USER daemon
+# USER daemon
```
paramsingh: I had to comment this out because with the user `daemon`, I wasn't able to create a log file in the container anywhere, it kept throwing `Permission denied` errors. I would love to learn of a better way to fix that.

docker/Dockerfile.prod
```diff
@@ -18,7 +18,7 @@ ENV PG_MAJOR 9.5
 ENV PG_VERSION 9.5.4-1.pgdg80+1
 RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends postgresql-client-$PG_MAJOR=$PG_VERSION \
+    && apt-get install -y --no-install-recommends postgresql-client-$PG_MAJOR \
```
paramsingh: I removed $PG_VERSION because when building using develop.sh, apt-get gave errors saying no such version. I'll fix this by updating with the current PG_VERSION, if you wish, or I'll get this out of the commit.

alpha_importer.py
alastair: In fact, requests already gives us this functionality: http://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status

paramsingh: Cool, I did not know of `raise_for_status`, but right now I don't know if I should use it instead of a simple if because of the logging stuff being done in the `if` block also.

paramsingh:
Updated the verbiage, created a new docker container for alpha_importer to run.

Need some help on the USER daemon stuff, right now I've just commented it out. Because we're using the same image as web, I don't know if I can change stuff for the alpha_importer container while not changing it for web.

alpha_importer.py
```diff
@@ -0,0 +1,107 @@
...
+import sys
+import logging
+from logging.handlers import RotatingFileHandler
+import os
+
+redis_connection = Redis(host = config.REDIS_HOST)
+QUEUE_KEY = 'alphaimporter:queue'
+SET_KEY_PREFIX = 'alphaimporter:set'
+ALPHA_URL = 'https://alpha.listenbrainz.org/'
+BETA_URL = 'https://listenbrainz.org'
```
mayhem: I know I said that we didn't need to have these in the config file, but now that I am testing, it is clear they should be there. Can you please add these to the config.py file so we can change them easily for testing?

paramsingh:
The disappearances were because of bad linking in the template, fixed now. Also moved the variables to config.py.sample.

And as you said, created a /logs directory to store alpha_importer.log and made daemon its owner in the Dockerfile. That should fix the daemon issue I was having.

alpha_importer.py
```diff
@@ -0,0 +1,100 @@
...
+                logger.info('All pages done for user {}'.format(username))
+                logger.info('Total number of pages done: {}'.format(page_count))
+                return True
+            data = extract_data(batch)
+            sent = send_batch(data, token)
+            next_max = data['payload'][-1]['listened_at']
+            logger.info('Page #{} done.'.format(page_count))
+            page_count += 1
+        except HTTPError:
+            time.sleep(DELAY)
```
mayhem: DELAY is not defined here. :(

paramsingh:
Rebased once again from master after the docker-cleanup PR merge.

PR #117, opened 2017-02-07
Fix another small bug where listen_type was not being detected.

paramsingh:
The JSON data contained strings while the variables LISTEN_TYPE_SINGLE
etc are integers, so we need to use _get_listen_type to make sure
that the code works properly. Without this, payloads with listen_type
'single' but more than one listens were being accepted.


mayhem:
Thanks!

PR #118, opened 2017-02-08
Testing in docker

mayhem:
This branch fixes and cleans up the unit test setup. We now have a test.sh script that will start the needed containers and run tests in the most expedient way possible. Also, testing happens in isolated containers that do not interfere with production containers/volumes.


test.sh
```diff
@@ -0,0 +1,128 @@
...
+# See https://github.com/docker/compose/issues/2119.
+COMPOSE_PROJECT_NAME=$(echo $COMPOSE_PROJECT_NAME_ORIGINAL | awk '{print tolower($0)}' | sed 's/[^a-z0-9]*//g')
+TEST_CONTAINER_NAME=listenbrainz
+TEST_CONTAINER_REF="${COMPOSE_PROJECT_NAME}_${TEST_CONTAINER_NAME}_1"
+
+#docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME build
+#    docker ps -a --no-trunc  | grep $COMPOSE_PROJECT_NAME \
+#        | awk '{print $1}' | xargs -r --no-run-if-empty docker stop
+#    docker ps -a --no-trunc  | grep $COMPOSE_PROJECT_NAME \
+#        | awk '{print $1}' | xargs -r --no-run-if-empty docker rm
```
gentlecat: Does this stuff need to be here (commented out)?

test.sh
```diff
@@ -0,0 +1,128 @@
...
+
+
+if [[ ! -d "docker" ]]; then
+    echo "This script must be run from the top level directory of the listenbrainz-server source."
```
gentlecat: Can do something like this to make it easier to use: https://github.com/metabrainz/sir/pull/11#discussion_r98348311 (`cd "$(dirname "${BASH_SOURCE[0]}")/../"`).

test.sh
```diff
@@ -0,0 +1,128 @@
...
+    exit -1
+fi
+
+
+function bring_up_db {
+    docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME up -d db influx redis
```
gentlecat: Why is this separation necessary when the dependencies are defined in the compose file?

test.sh
alastair: no, I don't think so

test.sh
alastair: true, perhaps not needed. I originally had it because I did
```
docker-compose up <databases>
docker-compose run listenbrainz dockerize ... <setup db>
docker-compose up -d
docker-compose logs -f listenbrainz
```

then I realised that we could skip the `up -d` and `logs` and just run py.test straight from `docker-compose run`.
In this case, if we run `docker-compose run listenbrainz dockerize ... <setup db>` then it should load the database containers first, right?

alastair:
Yes, I'd be really happy to remove docker-compose.prod.yml. The idea would be that we have docker-compose.yml for development and .test. for unit tests, right??

Do you think we also need a different Dockerfile and Dockerfile.prod? Looking at the case for LB, I think that they do 99% the same thing?

gentlecat:
You'll probably need a separate Dockerfile if you want to deploy on the new infrastructure.

gentlecat:
As a side note: if you want me to review something next time, I'd appreciate to be given more than half an hour.

PR #119, opened 2017-02-08
Convert tabs to spaces in manage.py

paramsingh:
After the merge of testing-in-docker, manage.py had both tabs and
spaces. Fix this by converting tabs to spaces.


alastair:
I think that was my code :(

paramsingh:
Happens to the best of us :)

PR #120, opened 2017-02-08
Integration tests

mayhem:
Basic framework for integration tests is up and running. Given the current trend I expect the integration-test.sh to be wrapped into develop.sh/test.sh or whatever that morphs into.

For the time being this script is good enough for ilikecomputers to start on writing actual integration tests.

### DIFF
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+COMPOSE_FILE_LOC=docker/docker-compose.integration.yml
+COMPOSE_PROJECT_NAME_ORIGINAL=listenbrainz_int
+
+# Project name is sanitized by Compose, so we need to do the same thing.
+# See https://github.com/docker/compose/issues/2119.
+COMPOSE_PROJECT_NAME=$(echo $COMPOSE_PROJECT_NAME_ORIGINAL | awk '{print tolower($0)}' | sed 's/[^a-z0-9]*//g')
+TEST_CONTAINER_NAME=listenbrainz
+TEST_CONTAINER_REF="${COMPOSE_PROJECT_NAME}_${TEST_CONTAINER_NAME}_1"
+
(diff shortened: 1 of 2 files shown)


integration-test.sh
```diff
@@ -0,0 +1,29 @@
...
+echo "Build current setup"
+docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME build
+
+echo "Running setup"
+docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME run --rm listenbrainz dockerize -wait tcp://db:5432 -timeout 60s \
+                  -wait tcp://influx:8086 -timeout 60s \
+                bash -c "python manage.py init_db --create-db && python manage.py init_msb_db --create-db && python admin/influx/create_db.py"
+
+echo "Bring containers up and run integration tests"
+docker-compose -f docker/docker-compose.integration.yml -p $COMPOSE_PROJECT_NAME up
```
alastair: We need to work out what the termination condition for this will be, because otherwise it will stay running forever until we ^c.
I think the easiest way would be to do what we do in the jenkins script - use `up -d` and then `logs -f listenbrainz`, which will quit when the tests in the `listenbrainz` container finishes.
Otherwise, we `up -d` everything except `listenbrainz`, and use `run listenbrainz py.test [blah]`, then run `down` after that finishes (like I have in test.sh).
Sound good?

mayhem: Yes, that sounds good, but... As I mentioned in IRC, I think it might be good to wrap all three use cases into one script (develop, test, integration-test) since there will be a lot of commonality between all of them. 

This particular script is mainly intended as a stop-gap measure for ilikecomputers to start writing integration tests -- I think it would be acceptable to burden him with ^C at the end of the process until your more refined script is done and dusted.

PR #122, opened 2017-02-10
LB-117: Integration tests for API 

paramsingh:
Moved the tests in `listenbrainz/tests` to `listenbrainz/tests/unit` and created a `listenbrainz/tests/integration` folder which contains tests for the api for now.

I've written the following tests:

* Simple get request to get the last 25 listens
* Unauthorized submission
* Submission of valid listen of type `single`
* invalid submission which has listen_type `single` more thean one listen in payload
* valid submission of listen_type `playing_now`
* invalid submission of listen_type `playing_now` which contains timestamp `listened_at`
* invalid submission of listen_type `playing_now` which has more than one listen in payload
* valid submission of listen_type `import`
* invalid submission in which the overall size of the listens sent is more than 10240 bytes
* invalid submission in which a listen contains more than the allowed number of tags in additional_info.
* invalid submission in which a listen contains a tag of length > 64
* invalid submission in which a listen contains an invalid release_mbid in additional_info
* invalid submission in which a listen contains an invalid artist_mbid in additional_info
* invalid submission in which a listen contains an invalid recording_mbid in additional_info

I've put the json sent in all of these tests into `listenbrainz/testdata` right now, but I'm thinking they
should probably go somewhere else, maybe somewhere in the `tests` folder, I'm not sure where exactly. 
Also, not sure if the changes in the `docker-compose.integration.yml` are the best way to make sure only
integration tests run via `./integration-test.sh`. Would appreciate help on this.


tests/integration/test_api.py
```diff
@@ -0,0 +1,196 @@
...
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", ".."))
+
+from webserver.testing import ServerTestCase
+from flask import url_for
+import db.user
+import time
+import json
+
+# lifted from AcousticBrainz
+def is_valid_uuid(u):
```
paramsingh: This is present in webserver.api_tools but is causing trouble when importing. I've copypastad it right now, but intend to see what's happening when importing from there.

tests/integration/test_api.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def test_get_listens(self):
+        """ Test to make sure that the api sends valid listens on get requests.
+        """
+        with open(os.path.join(os.getcwd(), 'testdata', 'valid_single.json'), 'r') as f:
+            payload = json.load(f)
+        payload['payload'][0]['listened_at'] = int(time.time())
+        response = self.send_data(payload)
+        self.assert200(response)
+        time.sleep(10)
```
paramsingh: When the listens get published to redis, it takes the influx subscriber some time to write them to the db. So we have to wait here otherwise, no data will be returned. Would love suggestions on better ways to get around this problem.

paramsingh:
Note: test_get_listens fails right now, but I've sent PRs fixing those problems.

tests/integration/test_api.py
alastair: As this is an integration test between components I don't think there's much else we can do.

tests/integration/test_api.py
```diff
@@ -0,0 +1,196 @@
...
+        url = url_for('api_v1.get_listens', user_name = self.user['musicbrainz_id'])
+        response = self.client.get(url, query_string = {'count': '1'})
+        self.assert200(response)
+        data = json.loads(response.data)
+        self.assertTrue('payload' in data)
+        data = data['payload']
+        # make sure user id is correct
```
alastair: I like putting a newline between multiple parts of a test

tests/integration/test_api.py
```diff
@@ -0,0 +1,196 @@
...
+        self.assertTrue('user_id' in data)
+        self.assertEquals(data['user_id'], self.user['musicbrainz_id'])
```
alastair: I normally don't have two tests for something like this - if `'user_id'` is missing from the data then the test is going to fail anyway. We may as well make the test a bit smaller and just have the `assertEquals`

tests/integration/test_api.py
```diff
@@ -0,0 +1,196 @@
...
+            url_for('api_v1.submit_listen'),
+            data = json.dumps(payload),
+            headers = {'Authorization': 'Token {}'.format(self.user['auth_token'])},
+            content_type = 'application/json'
+        )
+
+    def test_unauthorized_submission(self):
+        """ Test for checking that unauthorized submissions return 401
+        """
+        with open(os.path.join(os.getcwd(), 'testdata', 'valid_single.json'), 'r') as f:
```
alastair: in AcousticBrainz we have a method to get a filepath for tests: https://github.com/metabrainz/acousticbrainz-server/blob/629431054c6adbcfb80e74336510ef355f014f0d/db/testing.py#L70-L72

See also how the data root is defined in relative terms to the test file, not to the current directory

paramsingh:
Rebased against current master and implemented @alastair's code review suggestions. The file path for the data files currently is `root/testdata` but I can put it inside tests if that would be a better place for them.

alastair:
Looks good!

mayhem:
alastairp says: > yeah, that's reasonable

paramsingh:
@mayhem, sounds good to me. This one has been sitting for a while anyways. I'll send the second PR  tomorrow (holi today). :)

tests/unit/utils.py
```diff
@@ -5,7 +5,7 @@
 import sys
 import os
 
-sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../listenstore"))
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../../listenstore"))
```
alastair: This now causes a failure in lastfm tests:
```
______________________________________________ ERROR collecting db/tests/test_lastfm_user.py ______________________________________________
db/tests/test_lastfm_user.py:5: in <module>
    from tests.utils import generate_data
E   ImportError: No module named utils
```

Just checking... did you run tests as part of this PR?

PR #123, opened 2017-02-18
Import time to fix bug in get_listens.

paramsingh:
When a request was made to get_listens with no timestamp, time() is
used to start from the current time. However it wasn't imported
leading to 500 internal server errors on these requests.


mayhem:
Thanks!

PR #124, opened 2017-02-18
Dockerfile cleanup

mayhem:
This cleans up the unused prod docker compose and Dockerfiles and also removes support for Vagrant.


mayhem:
oops, not what I had intended. :(

PR #125, opened 2017-02-18
Production dockerfile cleanup

mayhem:
Move prod files to non-prod files, since prod is not used in actual production and also nuke Vagrant support.


docker/Dockerfile
```diff
@@ -11,13 +11,14 @@ RUN apt-get update \
...
+RUN pip install uWSGI==2.0.13.1
+
 # PostgreSQL client
 RUN apt-key adv --keyserver ha.pool.sks-keyservers.net --recv-keys B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8
 ENV PG_MAJOR 9.5
-ENV PG_VERSION 9.5.5-1.pgdg80+1
 RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends postgresql-client-$PG_MAJOR=$PG_VERSION \
+    && apt-get install -y postgresql-client-$PG_MAJOR \
```
alastair: :+1:

docker/Dockerfile
```diff
@@ -11,13 +11,14 @@ RUN apt-get update \
                        git \
     && rm -rf /var/lib/apt/lists/*
 
+RUN pip install uWSGI==2.0.13.1
```
alastair: Why is this not in `requirements`?

docker/Dockerfile
```diff
@@ -35,6 +36,12 @@ WORKDIR /code/listenbrainz
...
+# Now install our code, which may change frequently
 COPY . /code/listenbrainz/
 
-CMD python manage.py runserver -h 0.0.0.0 -p 80 -d
+# setup a log dir
+RUN mkdir /logs
+RUN chown -R daemon:daemon /logs
+
+# run as user daemon
+USER daemon
```
alastair: what creates this user?

docker/docker-compose.yml
```diff
@@ -4,47 +4,47 @@ services:
...
   web:
     build:
       context: ..
       dockerfile: ./docker/Dockerfile
+    command: uwsgi /code/listenbrainz/docker/uwsgi.ini
+    image: web
     volumes:
-      - $VAGRANT_DATA_DIR..:/code/listenbrainz
-      - $VAGRANT_DATA_DIR../data/app:/data
+      - ./data/app:/data
```
alastair: does the LB server write anything to `/data`? Is this a copy-over from CB?

docker/docker-compose.yml
```diff
@@ -4,47 +4,47 @@ services:
   db:
     image: postgres:9.5.3
     volumes:
-      - $VAGRANT_DATA_DIR../data/pgdata:/var/lib/postgresql/data/pgdata
-    environment:
-      PGDATA: /var/lib/postgresql/data/pgdata
-      POSTGRES_USER: listenbrainz
-      POSTGRES_PASSWD: listenbrainz
-      POSTGRES_DB: listenbrainz
+      - ./data/pgdata:/var/lib/postgresql
```
alastair: I believe that the postgres image exposes `/var/lib/postgresql/data/` as a volume, so we should use this instead of `/v/l/postgres`.

I don't like this way of persisting data with postgres because the local directory ends up owned by the `postgres` user, which makes it difficult to `rm`.
Instead, I prefer to use a named volume, which is stored in `/var/lib/docker` and can be managed using the `docker volumes` command. See https://github.com/AudioCommons/ac-mediator/blob/master/docker-compose.yml#L2-L3 for an example.

docker/docker-compose.yml
```diff
@@ -4,47 +4,47 @@ services:
...
 
+    ports:
+      - "8086:8086"
+      - "8083:8083"
 
   web:
     build:
       context: ..
       dockerfile: ./docker/Dockerfile
+    command: uwsgi /code/listenbrainz/docker/uwsgi.ini
```
alastair: Does this mean that during development we'll be running with nginx/uwsgi? Do you have a reason why we do this instead of `manage.py runserver`?
I know I normally go on about docker letting us run production setup in devlopment, but I think this is a bit over the top.

If we do decide to go with this, we should configure uwsgi to only use one thread, and to enable the code reloader.

docker/docker-compose.yml
mayhem: Done!

mayhem: Done.

mayhem:
Thanks for the comments, I've cleaned up copy pasta, nuked the nginx/uwsgi and moved the PG volume to a named volume.

mayhem:
Yep, works great, tests run, dev env runs. Clean and reliable -- I too like where this is headed. :)

PR #127, opened 2017-02-21
Show listen count on user page.

paramsingh:
Made modifications to InfluxListenStore to add support for a function
returning total listen count of a user and show this listen count on
the user page.

### DIFF
--- a/listenstore/listenstore/listenstore.py
+++ b/listenstore/listenstore/listenstore.py
@@ -27,6 +27,9 @@ DEFAULT_LISTENS_PER_FETCH = 25
 
 REDIS_USER_TIMESTAMPS = "user.%s.timestamps" # substitute user_name
 USER_CACHE_TIME = 3600 # in seconds. 1 hour
+LISTENCOUNT_CACHE_TIME_LOW = 15 * 60 # in seconds. 15 minutes
+LISTENCOUNT_CACHE_TIME_MEDIUM = 60 * 60 # in seconds. 1 hour
+LISTENCOUNT_CACHE_TIME_HIGH = 24 * 60 * 60 # in seconds. 1 day
 
 # TODO: This needs to be broken into 3 files and moved out of the separate listenstore module,
 #       but I am leaving this for the next PR
@@ -49,6 +52,10 @@ class ListenStore(object):
 
     def get_total_listen_count():
         """ Return the total number of listens stored in the ListenStore """
+        raise NotImplementedError()
+
+    def get_listen_count_for_user(self, user_name, need_exact):
+        """ Override this method in ListenStore implementation class """
         raise NotImplementedError()
 
     def fetch_listens(self, user_name, from_ts=None, to_ts=None, limit=DEFAULT_LISTENS_PER_FETCH):
@@ -204,18 +211,41 @@ class InfluxListenStore(ListenStore):
         self.influx = InfluxDBClient(host=conf['INFLUX_HOST'], port=conf['INFLUX_PORT'], database=conf['INFLUX_DB_NAME'])
 
 
-    def _get_num_listens_for_user(self, user_name):
-        count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_hash(user_name))
-        if count:
-            return int(count)
+    def get_listen_count_for_user(self, user_name, need_exact = False):
+        """ Returns total listen count for user with user_name passed.
 
-        results = self.influx.query("""SELECT count(*)
-                                         FROM listen
-                                        WHERE user_name = '%s'""" % (user_name, from_ts, to_ts, limit))
-        result = results.get_points(measurement='listen')
-#        count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_hash(user_name))
+            Args:
+                user_name: user name of user whose listen count is to be found
+                need_exact: signifies if cached value in redis will suffice or if exact value is needed
+        """
 
-        return 0
+        if not need_exact:
+            # check if the user's listen count is already in redis
+            # if already present return it directly instead of calculating it again
+            count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_name)
+            if count:
+                return int(count)
+
+        try:
+            results = self.influx.query("""SELECT count(*)
+                                             FROM listen
+                                            WHERE user_name = '%s'""" % (user_name))
+        except Exception as e:
+            self.log.error("Cannot query influx: %s" % str(e))
+            raise
+
+        # get the number of listens from the json
+        count = results.raw['series'][0]['values'][0][1]
+
+        # put this value into redis with expiry time based on the number of listens
+        # that the user has
+        if count <= 1000:
+            self.redis.setex(REDIS_INFLUX_USER_LISTEN_COUNT + user_name, count, LISTENCOUNT_CACHE_TIME_LOW)
+        elif count <= 10000:
+            self.redis.setex(REDIS_INFLUX_USER_LISTEN_COUNT + user_name, count, LISTENCOUNT_CACHE_TIME_MEDIUM)
+        else:
+            self.redis.setex(REDIS_INFLUX_USER_LISTEN_COUNT + user_name, count, LISTENCOUNT_CACHE_TIME_HIGH)
+        return int(count)
 
 
     def _select_single_value(self, query):
--- a/listenstore/listenstore/listenstore.py
+++ b/listenstore/listenstore/listenstore.py
@@ -52,7 +52,13 @@ class ListenStore(object):
         raise NotImplementedError()
 
     def get_listen_count_for_user(self, user_name, need_exact):
-        """ Override this method in ListenStore implementation class """
+        """ Override this method in ListenStore implementation class
+
+        Args:
+            user_name: the user to get listens for
+            need_exact: if True, get an exact number of listens directly from the ListenStore
+                        otherwise, can get from a cache also
+        """
         raise NotImplementedError()
 
     def fetch_listens(self, user_name, from_ts=None, to_ts=None, limit=DEFAULT_LISTENS_PER_FETCH):
@@ -214,12 +220,13 @@ class InfluxListenStore(ListenStore):
         self.influx = InfluxDBClient(host=conf['INFLUX_HOST'], port=conf['INFLUX_PORT'], database=conf['INFLUX_DB_NAME'])
 
 
-    def get_listen_count_for_user(self, user_name, need_exact = False):
-        """ Returns total listen count for user with user_name passed.
+    def get_listen_count_for_user(self, user_name, need_exact=False):
+        """Get the total number of listens for a user. The number of listens comes from
+           a redis cache unless an exact number is asked for.
 
-            Args:
-                user_name: user name of user whose listen count is to be found
-                need_exact: signifies if cached value in redis will suffice or if exact value is needed
+        Args:
+            user_name: the user to get listens for
+            need_exact: if True, get an exact number of listens directly from the ListenStore
         """
 
         if not need_exact:
@@ -232,7 +239,7 @@ class InfluxListenStore(ListenStore):
         try:
             results = self.influx.query("""SELECT count(*)
                                              FROM listen
-                                            WHERE user_name = '%s'""" % (user_name))
+                                            WHERE user_name = '%s'""" % (user_name, ))
         except (InfluxDBServerError, InfluxDBClientError) as e:
             self.log.error("Cannot query influx: %s" % str(e))
             raise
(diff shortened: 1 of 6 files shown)


paramsingh:
Screenshots :)

When value is not guaranteed to be exact:
screenshot_2017-02-22_01-46-47

When value is guaranteed to be exact:
screenshot_2017-02-22_01-47-01

listenstore/listenstore/listenstore.py
```diff
@@ -194,18 +201,41 @@ def __init__(self, conf):
...
+            count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_name)
+            if count:
+                return int(count)
 
-        return 0
+        try:
+            results = self.influx.query("""SELECT count(*)
+                                             FROM listen
+                                            WHERE user_name = '%s'""" % (user_name))
+        except Exception as e:
```
mayhem: Ideally we should look up what actual Exception influx might throw, rather than using a generic one here.

listenstore/listenstore/listenstore.py
```diff
@@ -194,18 +201,41 @@ def __init__(self, conf):
...
+            self.log.error("Cannot query influx: %s" % str(e))
+            raise
+
+        # get the number of listens from the json
+        count = results.raw['series'][0]['values'][0][1]
+
+        # put this value into redis with expiry time based on the number of listens
+        # that the user has
+        if count <= 1000:
```
mayhem: These constants should also be defined at the top of the file.

webserver/views/user.py
```diff
@@ -55,6 +55,18 @@ def profile(user_name):
 
     user = _get_user(user_name)
 
+    need_exact = request.args.get("exact")
+    if not need_exact:
+        listen_count = db_conn.get_listen_count_for_user(user_name)
```
mayhem: In the listenstore you catch the exception and then pass it on via Raise. However, here you do not catch the exception. You should catch the exception, but since this is a minor aspect of this page, perhaps show an error message when the count cannot be loaded in time. Then the rest of the page can still be rendered, rather than failing the whole page.

webserver/views/user.py
```diff
@@ -55,6 +55,18 @@ def profile(user_name):
...
+        # round off to nearest 10 as we can't guarantee that the values
+        # are not old, so show approximate values instead
+        if listen_count > 10:
```
mayhem: Again, here define your constants at the top of the file/class.

listenstore/listenstore/listenstore.py
```diff
@@ -25,6 +25,9 @@
 
 REDIS_USER_TIMESTAMPS = "user.%s.timestamps" # substitute user_name
 USER_CACHE_TIME = 3600 # in seconds. 1 hour
+LISTENCOUNT_CACHE_TIME_LOW = 15 * 60 # in seconds. 15 minutes
```
mayhem: We might want to move these defines into the class which uses them. Makes the code a little cleaner/easy to read.

paramsingh:
I have moved the defines into InfluxListenStore, added defines for other constants used and specified which exceptions `query` from influxdbclient can throw. Also, I've done some error handling so that if we aren't able to get listen count, the view won't break but show an error instead.

Also, added a reset_listen_count method that invalidates the data stored in redis and called this from alpha_importer and the last.fm backup import.

Not sure how to get the js last.fm importer to do this also?

paramsingh:
Also, sorry about the whitespace stuff, it was unintentional, I have my text editor set up to automatically trim trailing whitespaces. I will make sure stuff like this doesn't get in the way again.

mayhem:
Ohh, getting the last.fm importer to do this is indeed tricky. What if we add a "close" button that appears on importer when done (I don't think we have that now) and then send the user to the user profile page with an argument like clear-count, which then purges and re-computes the count?

paramsingh:
The js importer already had a close button and a link to the user page! :)

I removed the close button option and modified the link so that it resets the user count.

webserver/views/user.py
```diff
@@ -120,7 +133,10 @@ def profile(user_name):
         listens=listens,
         previous_listen_ts=previous_listen_ts,
         next_listen_ts=next_listen_ts,
-        spotify_uri=_get_spotify_uri_for_listens(listens)
+        spotify_uri=_get_spotify_uri_for_listens(listens),
+        have_listen_count = have_listen_count,
```
alastair: formatting

listenstore/listenstore/listenstore.py
```diff
@@ -188,24 +193,70 @@ def get_playing_now(self, user_id):
...
+    def reset_listen_count(self, user_name):
+        """ Function to reset the listen count of user
+
+            Args:
+                user_name: the musicbrainz id of user whose listen count needs to be reset
+
+            Return value:
+                The exact number of listens of user.
+        """
+        return get_listen_count_for_user(user_name, need_exact = True)
```
alastair: formatting

listenstore/listenstore/listenstore.py
```diff
@@ -188,24 +193,70 @@ def get_playing_now(self, user_id):
...
+        # get the number of listens from the json
+        try:
+            count = results.raw['series'][0]['values'][0][1]
+        except KeyError:
+            count = 0
+
+        # put this value into redis with expiry time based on the number of listens
+        # that the user has
+        if count <= InfluxListenStore.LOW_LISTEN_THRESHOLD:
+            self.redis.setex(REDIS_INFLUX_USER_LISTEN_COUNT + user_name, count, InfluxListenStore.LISTENCOUNT_CACHE_TIME_LOW)
```
alastair: Is this key a string? I'd much prefer that it uses string interpolation instead of concatenation.

This key is used 3 times in 6 lines. You should put it in a variable.

listenstore/listenstore/listenstore.py
```diff
@@ -188,24 +193,70 @@ def get_playing_now(self, user_id):
...
         ListenStore.__init__(self, conf)
         self.redis = Redis(conf['REDIS_HOST'])
         self.influx = InfluxDBClient(host=conf['INFLUX_HOST'], port=conf['INFLUX_PORT'], database=conf['INFLUX_DB_NAME'])
 
 
-    def _get_num_listens_for_user(self, user_name):
-        count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_hash(user_name))
-        if count:
-            return int(count)
+    def get_listen_count_for_user(self, user_name, need_exact = False):
```
alastair: formatting

listenstore/listenstore/listenstore.py
```diff
@@ -45,6 +46,10 @@ def fetch_listens_from_storage():
         """ Override this method in PostgresListenStore class """
         raise NotImplementedError()
 
+    def get_listen_count_for_user(self, user_name, need_exact):
```
alastair: This should have documentation to say what the `need_exact` field does.

Does it make sense to have the function in this class do the redis estimate value and call the more specific implementation only if needed?

listenstore/listenstore/listenstore.py
```diff
@@ -188,24 +193,70 @@ def get_playing_now(self, user_id):
...
+            # check if the user's listen count is already in redis
+            # if already present return it directly instead of calculating it again
+            count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_name)
+            if count:
+                return int(count)
+
+        try:
+            results = self.influx.query("""SELECT count(*)
+                                             FROM listen
+                                            WHERE user_name = '%s'""" % (user_name))
```
alastair: Is this using string formatting? This is almost definitely the wrong thing. Does our influx client not have query parameters?
Small thing - this is not quite correct in the string formatting:
```python

d = "bar"
s = "foo %s" % d # correct for one param, no parentheses
s = "foo %s" % (d, ) # correct, tuple
s = "two %s %s" % (d, d) # correct, multiple values
s = "foo %s" % (d) # incorrect - `(d)` evaluates to `d`, there is no reason for the params unless you use , to turn it into a tuple
```

listenstore/listenstore/listenstore.py
```diff
@@ -188,24 +193,70 @@ def get_playing_now(self, user_id):
...
+                need_exact: signifies if cached value in redis will suffice or if exact value is needed
+        """
 
-        return 0
+        if not need_exact:
+            # check if the user's listen count is already in redis
+            # if already present return it directly instead of calculating it again
+            count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_name)
+            if count:
+                return int(count)
```
alastair: if there is no data for this user, it's going to return `None`. Is that what you want?

listenstore/listenstore/listenstore.py
```diff
@@ -188,24 +193,70 @@ def get_playing_now(self, user_id):
...
+            self.redis.setex(REDIS_INFLUX_USER_LISTEN_COUNT + user_name, count, InfluxListenStore.LISTENCOUNT_CACHE_TIME_LOW)
+        elif count <= InfluxListenStore.MEDIUM_LISTEN_THRESHOLD:
+            self.redis.setex(REDIS_INFLUX_USER_LISTEN_COUNT + user_name, count, InfluxListenStore.LISTENCOUNT_CACHE_TIME_MEDIUM)
+        else:
+            self.redis.setex(REDIS_INFLUX_USER_LISTEN_COUNT + user_name, count, InfluxListenStore.LISTENCOUNT_CACHE_TIME_HIGH)
+        return int(count)
+
+
+    def reset_listen_count(self, user_name):
+        """ Function to reset the listen count of user
```
alastair: Documentation doesn't need to say "function to....". Just say what it does.

This method is named `reset_...`, but returns some data. Choose what it should do and change the name or return value as required

listenstore/listenstore/tests/test_influxlistenstore.py
```diff
@@ -161,3 +161,8 @@ def test_fetch_listens_3(self):
         self.assertEquals(listens[2].ts_since_epoch, 1400000100)
         self.assertEquals(listens[3].ts_since_epoch, 1400000050)
         self.assertEquals(listens[4].ts_since_epoch, 1400000000)
+
+    def test_get_listen_count_for_user(self):
+        count = self._create_test_data()
+        listen_count = self.logstore.get_listen_count_for_user(user_name = self.testuser_name)
```
alastair: formatting

webserver/templates/user/profile.html
```diff
@@ -11,6 +11,13 @@ <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
       <b><a href="https://musicbrainz.org/user/{{ user.musicbrainz_id }}">See profile on MusicBrainz</a></b>
     </div>
 
+    {% if have_listen_count %}
+      <h3> Listen Count: {{ listen_count }}  </h3>
+      <p> Note: This is an approximate value.<a href = "{{ url_for("user.profile", user_name=user.musicbrainz_id, exact = 'y') }}"> Get exact listen count! </a> </p>
```
alastair: It seems strange to have spaces inside the tags. I'd remove them. It's especially odd with the `<a>` link - here the link will extend out into the spaces either side of the phrase, which is not normal.

listenstore/listenstore/listenstore.py
```diff
@@ -188,24 +193,70 @@ def get_playing_now(self, user_id):
...
         self.redis = Redis(conf['REDIS_HOST'])
         self.influx = InfluxDBClient(host=conf['INFLUX_HOST'], port=conf['INFLUX_PORT'], database=conf['INFLUX_DB_NAME'])
 
 
-    def _get_num_listens_for_user(self, user_name):
-        count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_hash(user_name))
-        if count:
-            return int(count)
+    def get_listen_count_for_user(self, user_name, need_exact = False):
+        """ Returns total listen count for user with user_name passed.
```
alastair: Not a great documentation sentence. How about:
```python
"""Get the total number of listens for a user. The number of listens comes from
a redis cache unless an exact number is asked for.
Args:
    user_name: the user to get listens for
    need_exact: if True, get an exact number of listens directly from the ListenStore
"""
```

webserver/views/user.py
```diff
@@ -55,6 +57,17 @@ def profile(user_name):
...
+    need_exact_listen_count = request.args.get("exact") == 'y'
+    try:
+        have_listen_count = True
+        listen_count = db_conn.get_listen_count_for_user(user_name, need_exact = need_exact_listen_count)
+    except (InfluxDBServerError, InfluxDBClientError):
+        have_listen_count = False
+        listen_count = 0
+
+    if have_listen_count and not need_exact_listen_count and listen_count > LISTEN_COUNT_APPROXIMATION:
+        listen_count = (listen_count / LISTEN_COUNT_APPROXIMATION) * LISTEN_COUNT_APPROXIMATION
```
alastair: given the python3 discussion, this won't work, since `/` is now float division.
I just learned this:
```python
In [2]: round(4325, -1)
Out[2]: 4330.0
```
from https://docs.python.org/2/library/functions.html#round: "Values are rounded to the closest multiple of 10 to the power minus ndigits;"
so we could do `int(round(count, -1))` - where `-1` is some variable name

Having said this, I'm not sure why we only show an approximation unless the user requests an exact count. Is there a reason why we don't want to be exact all the time?

listenstore/listenstore/listenstore.py
paramsingh: If there is no data for the user, the code will go ahead and calculate the data. It won't return `None`.

paramsingh:
I think the main reason we're okay with showing approximate values is because the user doesn't really care about the exact value anyways (most of them won't even know about a change of 10 in their listen count) and doing it exactly would make it a teensy bit more complicated to implement. Here's the conversation we had about listen counts before I sent this PR, if you want to take a look: https://chatlogs.metabrainz.org/brainzbot/metabrainz/2017-02-20/?msg=3833504&page=2

I think we could do it so that counts in redis get updated still. Would have to make changes in influx-writer only as far as I can see.

paramsingh:
I've done a rebase with the current master (mainly because the redis port in config.py got added) and fixed the documentation and formatting issues. Sorry for the formatting stuff, I'll see that it doesn't happen again! It's the result of some habits I formed earlier :)

listenstore/listenstore/listenstore.py
paramsingh: >this is not quite correct in the string formatting:

Whoops! Oversight sorry.

> Does our influx client not have query parameters?

The [docs](https://influxdb-python.readthedocs.io/en/latest/api-documentation.html#influxdb.InfluxDBClient.query) are a bit sparse, but I don't think we have query parameters like sqlalchemy does. I was just doing it the way it had already been done in the code elsewhere.

alastair: Sorry, I think this might have been waiting on me to respond to.
You're right, I had a look at the docs and some of the implementation and I really can't find much around query parameters. I'm really worried about us doing this manually instead of getting a tool to do it for us. Any place where users can provide data, we're just asking for trouble if we try and sanitize input before we run a query.
I wonder if since influxql is quite similar to sql, could we use something like http://initd.org/psycopg/docs/cursor.html#cursor.mogrify to escape for us?  It does mean that we need a postgres cursor - perhaps the code is simple enough that we could copy it into a lib and then change all places we use query params in influx to use this wrapper.

mayhem: Yes, query parameter docs are scant and when I asked in their IRC channel about it, I basically got the response that "its not a problem", but that it didn't really sit all that well with me. Still makes me uncomfortable, to be honest.

mayhem: This further goes to another general suitability questions I have around fluxdb. I'll ping ya in IRC.

alastair:
approved! @mayhem and I had a bit of a discussion about integrating a counter system deeper into listenstore so that redis always knows how many items are in the db. We should look into this in the future

PR #128, opened 2017-02-22
LB-71: Token regeneration

paramsingh:
Work continued from #62. 

Did a rebase from current master. I think I have fixed most of the stuff talked about in the comments there.

Also added a test for the update_token function.

### DIFF
--- a/db/user.py
+++ b/db/user.py
@@ -29,14 +29,15 @@ def update_token(id):
     Args:
         id (int) - the row id of the user to update
     """
-    query = """UPDATE "user"
-                  SET auth_token = :token
-                WHERE id = :id
-            """
-    result = db.session.execute(query,
-            {"token": str(uuid.uuid4()),
-             "id": id})
-    db.session.commit()
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            UPDATE "user"
+               SET auth_token = :token
+             WHERE id = :id
+        """), {
+            "token": str(uuid.uuid4()),
+            "id": id
+        })
 
 
 def get(id):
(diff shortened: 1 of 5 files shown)


webserver/templates/user/resettoken.html
```diff
@@ -0,0 +1,18 @@
+{% extends 'base.html' %}
+
+{% block title %}Reset token - ListenBrainz{% endblock %}
+
+{% block content %}
+  <h3 class="page-title">Reset token</h3>
+  <p>Are you sure you want to reset your token? </p>
+
+  <p>If you do, you will need to replace your old token with the newly generated token in any of your applications that use your token. The old token will stop working immediately. Be sure to update the token immediately in your applications to ensure that you have no disruptions in recording your listens.</p>
```
mayhem: Well said!

db/user.py
```diff
@@ -23,6 +23,22 @@ def create(musicbrainz_id):
         })
         return result.fetchone()["id"]
 
+def update_token(id):
+    """Update a user's token to a new UUID
+
+    Args:
+        id (int) - the row id of the user to update
+    """
+    with db.engine.connect() as connection:
```
mayhem: You're not catching any possible DB exceptions here. As per usual, catch exceptions and then show an appropriate error message and log an error as well.

paramsingh:
Added error catching in the db code.

webserver/templates/user/resettoken.html
paramsingh: Haha, I can't take credit for this.

https://github.com/metabrainz/listenbrainz-server/pull/62#r42362760

PR #130, opened 2017-02-23
Fix internal server errors due to None returned from influx.

paramsingh:
Right now, if there is no mbid in a field, influx returns null which
leads to internal server errors on call to fetch_listens because
fetch_listens expects strings everywhere.

### DIFF
--- a/listen.py
+++ b/listen.py
@@ -47,34 +47,34 @@ class Listen(object):
     def from_influx(cls, row):
         """ Factory to make Listen objects from an influx row
         """
-        dt = datetime.strptime(row['time'] , "%Y-%m-%dT%H:%M:%SZ")
+        dt = datetime.strptime(row['time'] , '%Y-%m-%dT%H:%M:%SZ')
         t = int(dt.strftime('%s'))
         mbids = []
         if row.get('artist_mbids'):
-            for mbid in row.get('artist_mbids').split(","):
-                mbids.append(mbid) 
+            for mbid in row.get('artist_mbids').split(','):
+                mbids.append(mbid)
         tags = []
         if row.get('tags'):
-            for tag in row.get('tags').split(","):
+            for tag in row.get('tags').split(','):
                 tags.append(tag)
 
         data = {
-            'artist_mbids' : mbids,
-            'album_msid' : row.get('album_msid'),
-            'album_mbid' : row.get('album_mbid'),
-            'album_name' : row.get('album_name'),
-            'recording_mbid' : row.get('recording_mbid'),
-            'tags' : tags,
+            'artist_mbids': mbids,
+            'album_msid': row.get('album_msid'),
+            'album_mbid': row.get('album_mbid'),
+            'album_name': row.get('album_name'),
+            'recording_mbid': row.get('recording_mbid'),
+            'tags': tags,
         }
         return cls(
-            timestamp = t,
-            user_name = row.get('user_name'),
-            artist_msid = row.get('artist_msid'),
-            recording_msid = row.get('recording_msid'),
-            data = { 
-                'additional_info' : data,
-                'artist_name' : row.get('artist_name'),
-                'track_name' : row.get('track_name'),
+            timestamp=t,
+            user_name=row.get('user_name'),
+            artist_msid=row.get('artist_msid'),
+            recording_msid=row.get('recording_msid'),
+            data={
+                'additional_info': data,
+                'artist_name': row.get('artist_name'),
+                'track_name': row.get('track_name'),
             }
         )
 
--- a/listen.py
+++ b/listen.py
@@ -43,6 +43,41 @@ class Listen(object):
             data=j.get('track_metadata')
         )
 
+    @classmethod
+    def from_influx(cls, row):
+        """ Factory to make Listen objects from an influx row
+        """
+        dt = datetime.strptime(row['time'] , "%Y-%m-%dT%H:%M:%SZ")
+        t = int(dt.strftime('%s'))
+        mbids = []
+        if row.get('artist_mbids'):
+            for mbid in row.get('artist_mbids').split(","):
+                mbids.append(mbid) 
+        tags = []
+        if row.get('tags'):
+            for tag in row.get('tags').split(","):
+                tags.append(tag)
+
+        data = {
+            'artist_mbids' : mbids,
+            'album_msid' : row.get('album_msid'),
+            'album_mbid' : row.get('album_mbid'),
+            'album_name' : row.get('album_name'),
+            'recording_mbid' : row.get('recording_mbid'),
+            'tags' : tags,
+        }
+        return cls(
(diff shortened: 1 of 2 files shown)


alastair:
I think it's better that we continue to use `None` to represent missing data instead of an empty string.  Why not do something like:

```python
ids = result.get('artist_mbids')
if ids:
    for id in result.get('artist_mbids', '').split(' ')
        if id:
            mbids.append(id) 
```
Looking at the rest of the code there might be some other ways of doing this - 
We use `.get(x, '')` everywhere, I'd prefer all these values to be `None`.

We also pass this data into the `Listen` constructor. An alternative would be to have a `Listen.from_influx` classmethod which takes a row from influx and does all of the necessary transformation to make the Listen object.

paramsingh:
Created a from_influx class method and did the work to convert influx row to Listen there. Also, I've changed the code so that instead of `''`, `None` gets returned everywhere. This will cause the api to return null values in fields in the json also.

listen.py
```diff
@@ -43,6 +43,41 @@ def from_json(cls, j):
...
+        data = {
+            'artist_mbids' : mbids,
+            'album_msid' : row.get('album_msid'),
+            'album_mbid' : row.get('album_mbid'),
+            'album_name' : row.get('album_name'),
+            'recording_mbid' : row.get('recording_mbid'),
+            'tags' : tags,
+        }
+        return cls(
+            timestamp = t,
```
alastair: small formatting thing, these are arguments to a function, and so the standard is to have no spaces - 
```python
cls(timestamp=t, x=y, ...)
```

listen.py
```diff
@@ -43,6 +43,41 @@ def from_json(cls, j):
...
+        if row.get('artist_mbids'):
+            for mbid in row.get('artist_mbids').split(","):
+                mbids.append(mbid) 
+        tags = []
+        if row.get('tags'):
+            for tag in row.get('tags').split(","):
+                tags.append(tag)
+
+        data = {
+            'artist_mbids' : mbids,
```
alastair: formatting - no space before a `:`, but space after

listen.py
```diff
@@ -43,6 +43,41 @@ def from_json(cls, j):
...
         )
 
+    @classmethod
+    def from_influx(cls, row):
+        """ Factory to make Listen objects from an influx row
+        """
+        dt = datetime.strptime(row['time'] , "%Y-%m-%dT%H:%M:%SZ")
+        t = int(dt.strftime('%s'))
+        mbids = []
+        if row.get('artist_mbids'):
```
alastair: This looks a little odd to me - If we have two calls to the same function close together my preference is to save it in a variable:
```python
artist_mbids = row.get('artist_mbids')
if artist_mbids:
    for mbid in artist_mbids.split(','):
```
However I think we can take a shortcut here - we can have a default value of `""`, which when split will be an empty array:
```python
for mbid in row.get('artist_mbids', '').split(','):
```
But looking further, if `artist_mbids` isn't in the returned data, this field will be an empty array, so why can't we do this?:
```python
artist_mbids = row.get('artist_mbids', '').split(',')
``` 

and the same with `tags`

listen.py
```diff
@@ -43,6 +43,41 @@ def from_json(cls, j):
...
+            for mbid in row.get('artist_mbids').split(","):
```
alastair: This is something that you copied, but let's really try and be consistent with quotes here. I don't mind which one we use, let's just not have two different ones on the same line! (unless we need to use one type to include the other type)

listen.py
```diff
@@ -43,6 +43,41 @@ def from_json(cls, j):
             data=j.get('track_metadata')
         )
 
+    @classmethod
+    def from_influx(cls, row):
+        """ Factory to make Listen objects from an influx row
+        """
+        dt = datetime.strptime(row['time'] , "%Y-%m-%dT%H:%M:%SZ")
```
alastair: This is the only place that you use dictionary access instead of `.get`. Do we have tests to verify the minimum data which we get from influx?  If not, can you open a ticket to make sure that we write some?

listen.py
paramsingh: @alastair, I wrote it like this to get rid of the internal server errors. What you're suggesting is what was written previously. If some listens do not have artist_mbids, influx will return None in the `artist_mbid` field. Then `row.get('artist_mbids', '')` returns `None` which cannot be split.

alastair: Ah, I didn't realise that the problem was that the data was coming back as `None` from influx. Why is this? Is it because we fill out these fields on insert? or because there is a schema defined which always fills in these fields?
If the fields are optional, we shouldn't add them on insert if they have no data. If you need to check if they're None, use the first suggestion that I gave, where we move the data into a separate variable

listen.py
paramsingh: https://tickets.metabrainz.org/browse/LB-126 :)

paramsingh:
Fixed the formatting problems (haha, this is the third time I'm writing that) and did a rebase from master so that the conflicts are resolved :)

Also, added the ticket https://tickets.metabrainz.org/browse/LB-126 about the tests.

listen.py
paramsingh: @alastair This is how I found the bug, initially I didn't have any listens with artist_mbid in the influx db. Everything worked fine, but as soon as I submitted a listen with artist_mbid, I started getting internal server errors.

It seemed to me that initially when there were no listens with mbids, the row returned by influx did not contain the field even though we always specify `''` (I think this is the equivalent of null in influx) to be put in there if there's no value. But as soon as a single listen with mbids got added, influx started returning `None` for the field for the listens with no mbids.

paramsingh: I'm not really familiar enough with influx to explain what exactly is happening. :/

alastair: Right. That definitely sounds like something that might happen. Let's make sure that we get tests written for this as soon as possible, then. For now how about we change this part to my first suggestion (use `.get()` into a variable and then use the variable)?

listen.py
```diff
@@ -43,6 +43,41 @@ def from_json(cls, j):
...
+    @classmethod
+    def from_influx(cls, row):
+        """ Factory to make Listen objects from an influx row
+        """
+        dt = datetime.strptime(row['time'] , '%Y-%m-%dT%H:%M:%SZ')
+        t = int(dt.strftime('%s'))
+        mbids = []
+        if row.get('artist_mbids'):
+            for mbid in row.get('artist_mbids').split(','):
+                mbids.append(mbid)
```
alastair: We can simplify this a bit more from my original suggestion:
```python
artist_mbids = row.get('artist_mbids')
if artist_mbids:
    mbids = artist_mbids.split(',')
```

paramsingh: Fixed, I've named the variables `artist_mbids` and `influx_tags`. Is that okay, or should I use some other names?

alastair: Looks OK for now.

PR #131, opened 2017-02-27
Bigquery credential cleanup

mayhem:
Do not copy any credentials into the docker containers that we will push to docker-hub.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -15,7 +15,7 @@ pip-log.txt
 pip-delete-this-directory.txt
 
 # Application data
-/data
+/docker/data
 
 # Test results
 htmlcov
@@ -23,3 +23,6 @@ htmlcov
 
 # Vagrant
 .vagrant
+
+# Private credentials
+credentials
--- a/.dockerignore
+++ b/.dockerignore
@@ -15,14 +15,11 @@ pip-log.txt
 pip-delete-this-directory.txt
 
 # Application data
-/docker/data
+docker/data
 
 # Test results
 htmlcov
 .coverage
-
-# Vagrant
-.vagrant
 
 # Private credentials
 credentials
--- a/bigquery-writer/bigquery-writer.py
+++ b/bigquery-writer/bigquery-writer.py
@@ -19,7 +19,7 @@ REPORT_FREQUENCY = 5000
 SUBSCRIBER_NAME = "bq"
 KEYSPACE_NAME_INCOMING = "ilisten"
 KEYSPACE_NAME_UNIQUE = "ulisten"
-APP_CREDENTIALS_FILE = "bigquery-credentials.json"
+APP_CREDENTIALS_FILE = os.environ['GOOGLE_APPLICATION_CREDENTIALS']
 
 # TODO: 
 #   Big query hardcoded data set ids
--- a/bigquery-writer/bigquery-writer.py
+++ b/bigquery-writer/bigquery-writer.py
@@ -19,7 +19,7 @@ REPORT_FREQUENCY = 5000
 SUBSCRIBER_NAME = "bq"
 KEYSPACE_NAME_INCOMING = "ilisten"
 KEYSPACE_NAME_UNIQUE = "ulisten"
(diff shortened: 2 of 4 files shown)


docker/docker-compose.yml
```diff
@@ -1,5 +1,8 @@
 version: "2"
 
+# IMPORTANT NOTE: For the web container we use exposing the code via a volume, not via
```
alastair: Not quite true - [the dockerfile](https://github.com/metabrainz/listenbrainz-server/blob/747118854aa91c877c974ed95b1710ec3c8988e4/docker/Dockerfile#L38) also uses `COPY` to add the code. The use of volumes is for during development - e.g., if you have flask autoreloader configured to restart on file change you need your most up-to-date version of the code always in the container (otherwise you would need to `docker-compose build / down / up` after every file change). I'd say this instead:

```
IMPORTANT NOTE: Volume paths mounted on containers are relative to the directory that this file is in (`docker/`)
and so probably need to start with `../` to refer to a directory in the main code checkout
```

docker/docker-compose.yml
```diff
@@ -62,6 +65,10 @@ services:
     build:
       context: ..
       dockerfile: ./docker/bigquery-writer/Dockerfile
+    volumes:
+      - ../credentials:/code/credentials
```
alastair: We should make a decision here if you want the source code of the bigquery writer mounted into the container as a volume too (for easier code changes). This volume will have to be configured for production (`docker run -v .....`), but if we have a volume `..:/code` then this isn't strictly necessary) - to be continued in our discussion on metabrainz docker style guide

.dockerignore
```diff
@@ -15,11 +15,14 @@ pip-log.txt
 pip-delete-this-directory.txt
 
 # Application data
-/data
+/docker/data
```
alastair: From https://docs.docker.com/engine/reference/builder/#/dockerignore-file:
> For the purposes of matching, the root of the context is considered to be both the working and the root directory. For example, the patterns /foo/bar and foo/bar both exclude a file or directory named bar in the foo subdirectory

This seems correct, but I think that `docker/data` would also be fine, and is consistent with the rest of the file. I don't have a strong preference either way.

.dockerignore
```diff
@@ -15,11 +15,14 @@ pip-log.txt
...
 
 # Test results
 htmlcov
 .coverage
 
 # Vagrant
 .vagrant
```
alastair: we no longer use vagrant

bigquery-writer/bigquery-writer.py
```diff
@@ -19,7 +19,7 @@
 SUBSCRIBER_NAME = "bq"
 KEYSPACE_NAME_INCOMING = "ilisten"
 KEYSPACE_NAME_UNIQUE = "ulisten"
-APP_CREDENTIALS_FILE = "bigquery-credentials.json"
+APP_CREDENTIALS_FILE = os.environ['GOOGLE_APPLICATION_CREDENTIALS']
```
alastair: will this fail if the env variable doesn't exist? Is this the preferred behaviour?

docker/docker-compose.yml
mayhem: I think that makes sense, but for the interest of getting the best test out, I'd prefer not to touch this right this second.

bigquery-writer/bigquery-writer.py
mayhem: It will not fail. The test below catches this and acts appropriately.

alastair: os.environ is a dictionary, if the env variable `GOOGLE_APPLICATION_CREDENTIALS` doesn't exist, this will fail with a `KeyError`. You should use `os.environ.get('GOOGLE_APPLICATION_CREDENTIALS')` and do your error detection by checking if this value is `None` in addition to it being a valid file

docker/docker-compose.yml
alastair: :+1:

bigquery-writer/bigquery-writer.py
mayhem: Oh! If they key doesn't exist. Sorry, misunderstood.

PR #132, opened 2017-03-03
Fix MusicBrainz link to artists on user profile page.

paramsingh:
The link to artists in listens that contained artist mbids was
incorrect previously. I've linked to the first artist in the list
of artist mbids to fix this.

### DIFF
--- a/webserver/templates/user/profile.html
+++ b/webserver/templates/user/profile.html
@@ -41,7 +41,7 @@
                 <tr>
                   <td>
                       {% if listen.track_metadata.additional_info.artist_mbids %}
-                        <a href="http://musicbrainz.org/recording/{{ listen.track_metadata.additional_info.artist_mbids }}">
+                        <a href="http://musicbrainz.org/artist/{{ listen.track_metadata.additional_info.artist_mbids[0] }}">
                       {% endif %}
                       {{ listen.track_metadata.artist_name }}
                       {% if listen.track_metadata.additional_info.artist_mbids %}


paramsingh:
Previously the entire list was being linked to (which is obviously incorrect).

Note: This probably can't be reproduced until #132 is merged, because the code throws internal server errors on loading pages when influx contains some listens with artist_mbids and some listens with no artist_mbids.

alastair:
This looks OK, but the `[0]` worries me. Do we have a failing in our API format that means we can't do proper artist credit phrases?

paramsingh:
@alastair, Seems like something that we should discuss before the beta to make sure the API remains consistent after its release. The `artist_mbids` list in `additional_info` does not discriminate between the main artist and the featuring artists, as far as I can see. 

Maybe breaking the `artist_mbid` field into two parts: one containing the main artist and the other being a list of the featuring artists would be better? I don't know what the names of the fields would be though.

alastair:
The thing about musicbrainz credits is that it's not always a case of "main" and "featuring" artists - https://musicbrainz.org/recording/d386933f-98f6-41fc-bc1c-4c154659e8ff is a good example. Both artists are equal in status.
I think that the ideal case here would be for us to use a recording mbid and get this metadata directly from musicbrainz.

Thinking about the file metadata case, this data is going to come from an audiofile in the way that we currently store it - a string for the artist name, and a comma separated string for the artist ids. There is no metadata scheme that I know that will store a full MB artist credit. Therefore, I think we have a good enough compromise here at the moment.

paramsingh:
That makes sense to me. The `[0]` is a little ugly to me still though. :)

alastair:
Yeah, I don't like the `[0]` either. A more complete solution to this would be to take the recording ID (if it exists) and get the artist credit details directly from the MB database. This is out of the scope of this fix, though, since we don't currently have this connection to MB.
Let's OK it for now. Can you create a ticket on Jira about directly using MusicBrainz to get metadata when we know MBIDs?  - Now that LB is on the main Docker infrastructure this should be possible to do.

PR #134, opened 2017-03-04
LB-122: Fix the issue of album_msid and album_name getting dropped during influx write

paramsingh:
The code was looking for `album_msid` instead of `release_msid`
leading to nothing being stored in the album_msid field in influx.

### DIFF
--- a/listen.py
+++ b/listen.py
@@ -38,7 +38,7 @@ class Listen(object):
             user_name=j.get('user_name', ""),
             timestamp=datetime.utcfromtimestamp(float(j['listened_at'])),
             artist_msid=j['track_metadata']['additional_info'].get('artist_msid'),
-            album_msid=j['track_metadata']['additional_info'].get('album_msid'),
+            album_msid=j['track_metadata']['additional_info'].get('release_msid'),
             recording_msid=j.get('recording_msid'),
             data=j.get('track_metadata')
         )
--- a/listenstore/listenstore/listenstore.py
+++ b/listenstore/listenstore/listenstore.py
@@ -278,7 +278,7 @@ class InfluxListenStore(ListenStore):
                     'artist_name' : listen.data['artist_name'],
                     'artist_msid' : listen.artist_msid,
                     'artist_mbids' : ",".join(listen.data['additional_info'].get('artist_mbids', [])),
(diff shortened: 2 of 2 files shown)


paramsingh:
Note: This partly fixes LB-122 as album names are still not handled. However I think some discussion is required on where release_name belongs exactly, as the changes will have to be made according to that.

listen.py
```diff
@@ -38,7 +38,7 @@ def from_json(cls, j):
             user_name=j.get('user_name', ""),
             timestamp=datetime.utcfromtimestamp(float(j['listened_at'])),
             artist_msid=j['track_metadata']['additional_info'].get('artist_msid'),
-            album_msid=j['track_metadata']['additional_info'].get('album_msid'),
```
alastair: How complex is to to replace `album_*` with `release_*` everywhere in the `Listen` class and places that use it?

paramsingh: From a look at the codebase, seems like `album_*` is used in the listenstores (both postgres and influx) and the bigquery schema has columns named `album_*`. Wouldn't be too complex to change it all to `release_*`, although I was thinking maybe there is some reason we moved from `release_*` to `album_*` related to MB? If there isn't, a refactor wouldn't be much effort I think.

alastair: I suspect it's because listenstore was originally written with `album`. The rest of listenbrainz used `release`, but we've probably been slowly building out features using listenstore as a base.
I'd confirm with @mayhem, but my opinion is that we should use the MusicBrainz standard here, which is release, not album.

paramsingh: Waiting for @mayhem for the final say. I think this refactor should be a different PR anyways.

paramsingh:
I have fixed the dropping of `album_name` also now, after some discussion with @alastair where we both agreed that `release_name` is better placed in the top level and the docs also mention multiple times that `release_name` should be in the top level.

alastair:
Create a ticket to rename album->release in the code, and then I'll approve this 💿

paramsingh:
@alastair, I have created [LB-125](https://tickets.metabrainz.org/browse/LB-125).

The work involved isn't much and should be easy to do. I've started with it and should have the changes done and tested by today hopefully. Mostly will involve changes in the ListenStore and the Listen class.

mayhem:
Bug created, looks fine, merging.

PR #135, opened 2017-03-07
Make redis port a configurable option in config.py

paramsingh:
I've made appropriate changes in influx-writer, bigquery-writer,
the webserver and other places so that now we can run redis from
non-standard ports also.

### DIFF
--- a/alpha_importer.py
+++ b/alpha_importer.py
@@ -10,7 +10,7 @@ import logging
 from logging.handlers import RotatingFileHandler
 import os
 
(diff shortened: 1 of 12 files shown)


paramsingh:
I did a git grep `REDIS_HOST` to make sure that I don't miss anything. I'm pretty sure that I haven't missed anything, it runs well on my system. I've done some testing by hand, submitted a listen, used alpha_importer etc and it all is working for me.

PR #136, opened 2017-03-10
Add alert to inform user that this is just a test version of the beta.

paramsingh:
Made changes to `base.html` so that it is shown on the top of every page in red.


paramsingh:
Screenshots :)
screenshot_2017-03-11_02-16-46
screenshot_2017-03-11_02-16-53

I've tried to keep the verbiage as clear as possible, but improvements are welcome!

PR #137, opened 2017-03-10
Make audioscrobbler link an option in config.py

paramsingh:
Also made it default to https. This should fix the last.fm importer
currently not working in beta.


mayhem:
Looks good!

PR #139, opened 2017-03-11
Add stats about LB to current status page

paramsingh:
Stats include the total number of users registered, the total
number of listens present in the listen store and the number of
people in the queue waiting for an import from alpha_importer.


paramsingh:
Screenshot: 
screenshot_2017-03-11_17-53-36

I've cached everything for 10 minutes. I can change that to something else if it's needed.

PR #141, opened 2017-03-11
Modify last.fm import so that it becomes a one-step process

paramsingh:
Now the user only has to fill her username and click the import
button to start the last.fm import process.


paramsingh:
Moved `loader.js` code to a script in the `import.html` template that gets called on clicking the submit button. The script then gets the last.fm username from the text box and uses it to start with the import. Works well on my end, with only input of the last.fm username and the click of a button needed to start the import.

PR #142, opened 2017-03-12
LB-127: Make position of last.fm importer dialog box fixed

paramsingh:
Also improve the message shown to the user on the start of
the import.


paramsingh:
The stuff done to get the dialog box seems a little hacky to me. Is this enough of a fix to LB-127 or should I do something else, involving a bigger change in the code? This can be a temporary fix until then. Not sure is `position:fixed` is any better than `position:absolute`.

mayhem:
Let's see how this one goes. :)

PR #143, opened 2017-03-12
Modify import template to allow user to re-import from alpha

paramsingh:
Now the user gets the option to import from alpha as many times as she wants.


paramsingh:
Screenshot:
screenshot_2017-03-12_21-29-17

The redis key is updated again when the user presses the button and everything works as expected. Didn't have to change any of the logic to add this functionality. :)

PR #144, opened 2017-03-12
Production docker setup

mayhem:
This PR:

- Compacts 4 services into on docker image.
- Creates a Dockerfile.prod for use in production.
- Add redis port support & persistence
- Big-Query credential support for production
- User & listen count improvements for the status page (needed for debugging)

Sorry for the length, but this wouldn't have made sense to break into smaller PRs -- this one is still reasonably reviewable. Also, there are some references to this branch in order for the code to work (see Dockerfile.prod). I will clean this up when I merge the code.

This PR is live on https://beta.listenbrainz.org


Dockerfile.prod
```diff
@@ -0,0 +1,71 @@
...
+RUN git checkout production-docker-setup
+RUN pip install -r requirements.txt
+
+# Node
+#RUN curl -sL https://deb.nodesource.com/setup_6.x | bash -
+#RUN apt-get install -y nodejs
+#RUN npm install less-plugin-clean-css
+
+# Compile the CSS
+# TODO: Bring this back
```
mayhem: https://tickets.metabrainz.org/browse/LB-128 exists to address this todo.

Dockerfile.prod
```diff
@@ -0,0 +1,71 @@
...
+                       libpq-dev \
+                       libffi-dev \
+                       libssl-dev \
+                       node-less \
+                       node-clean-css \
+                       redis-tools \
+    && rm -rf /var/lib/apt/lists/*
+
+
+# This is needed to circumvent a bug in an ubunutu package
```
alastair: would be nice to say which package

Dockerfile.prod
```diff
@@ -0,0 +1,71 @@
...
+
+# MessyBrainz
+RUN git clone https://github.com/metabrainz/messybrainz-server.git messybrainz
+WORKDIR /code/messybrainz
+RUN pip install -r requirements.txt
+RUN python setup.py install
+
+# ListenBrainz
+WORKDIR /code
+RUN git clone https://github.com/metabrainz/listenbrainz-server.git listenbrainz
```
alastair: Is this the standard MetaBrainz way of building an image? (as opposed to running `docker build` inside a checkout and using `COPY .`)

Dockerfile.prod
```diff
@@ -0,0 +1,71 @@
...
+WORKDIR /code/listenbrainz
+
+# TODO:Remove this for merging!
```
alastair: You'll remember this?? :scream:

influx-writer/requirements.txt
```diff
@@ -1,5 +0,0 @@
-ujson==1.33
```
alastair: Just checking that the main `requirements` file contains all of the items from these deleted files (my guess is yes, since you've got it running)

bigquery-credentials.json.ctmpl
```diff
@@ -0,0 +1,15 @@
+{{- define "KEY" -}}
```
alastair: This is for prod, can it go in `docker/prod`?

consul_config.py.ctmpl
```diff
@@ -0,0 +1,68 @@
...
+{{end}}
+{{end}}
+
+# Not currently used
+MAX_POSTGRES_LISTEN_HISTORY = "-1"
+PG_QUERY_TIMEOUT = "3000"
+PG_ASYNC_LISTEN_COMMIT = False
+
+{{if service "listenbrainz-redis"}}
+{{with index (service "listenbrainz-redis") 0}}
```
alastair: is `service` the name of the container which is running this image? Just to confirm, these config options will only appear in the container for their specific task. Is this OK? Are you OK to have these settings appear in all containers? I have no strong opinions either way, just raising the possibility.

docker/prod/uwsgi/uwsgi.service
```diff
@@ -0,0 +1,23 @@
...
+    cd /code/listenbrainz
+    exec python bigquery-writer/bigquery-writer.py
+fi
+if [ "${CONTAINER_NAME}" = "listenbrainz-alpha-importer" ]
+then
+    cd /code/listenbrainz
+    exec python alpha_importer.py
+fi
+
+echo "init script has no clue which service to start. Set env var CONTAINER_NAME!"
```
alastair: I'm not up to date on my bash - will this happen when the child process finished, or does the `exec` stop the rest of the script from running?
You could have a test for the environment variable at the top of this file instead, and have an explicit `exit 1` so that our init knows that it's failed (instead of exited) - http://stackoverflow.com/a/13864829 ?

alastair: Can you make consul generate this file? Or is there a race condition with the startup of consul and whatever service is running in this container?

Dockerfile.prod
mayhem: Done.

Dockerfile.prod
mayhem: Not really, going back to the traditional way doing things...

bigquery-credentials.json.ctmpl
mayhem: docker hub requires Dockerfiles to be in the top level. Though I am unsure about using automated builds -- for now lets keep this, but if we decided to not do automated builds we can move the Dockerfiles.

consul_config.py.ctmpl
mayhem: No -- service allows you to check if that service is up and running and if so, you can execute the "with" statement to get the IP/port of the service you want to use.

docker/prod/uwsgi/uwsgi.service
mayhem: exec without redirections, replaces this script in memory -- there are no further statements executed in this script after exec.

And this script is the entry point for the container -- this would be a massive chicken-and-the-egg problem that I really don't wish to tackle. Especially with us possibly moving to swarm, all this might change.

influx-writer/requirements.txt
mayhem: Yep, there was loads of duplication.

bigquery-credentials.json.ctmpl
alastair: This is a consul template, not Dockerfile, though!

docker/prod/docker-helpers/install_consul_template.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
```
mwiencek: If you're using the `metabrainz/python` images now, you can remove all of these scripts under `docker/prod/docker-helpers`.

mayhem: Right, yes. Thanks!

PR #146, opened 2017-03-15
LB-133: Make integration tests inherit from DatabaseTestCase

paramsingh:
Running integration tests from test.sh was not working because
database wasn't set up. Inheriting from DatabaseTestCase fixes
this.


paramsingh:
Now the integration tests can be run from `test.sh` although @mayhem did say that we only want to run integration tests from `integration-test.sh`. What would be the best way to make `test.sh` ignore these tests? Changes to `pytest.ini`? Some help would be appreciated here. :)

tests/integration/test_api.py
```diff
@@ -20,9 +21,11 @@ def is_valid_uuid(u):
     except ValueError:
         return False
 
-class APITestCase(ServerTestCase):
+class APITestCase(ServerTestCase, DatabaseTestCase):
```
alastair: Does it make sense to make an IntegrationTestCase which makes this easy as we make more tests?

alastair:
what's in `tests/unit`? I see the pubsub test, which to me sounds more like an integration test (touches an external service). Because of this, my recommendation is that the pytest.ini used for unit tests doesn't include the `tests` folder, and the integration test script only tests this folder.

tests/integration/test_api.py
alastair: I think `tests/integration/__init__.py` would be better

paramsingh: Okay, makes sense.

alastair:
yes to all the above! (note that git only tracks files, not directories. If there is nothing in the directory it will disappear)

paramsingh:
Created an IntegrationTestCase class that inherits from both ServerTestCase and DatabaseTestCase.

Also moved `test_pubsub.py` from tests/unit to tests/integration. Now `unit` won't contain anything because #145 moves utils.py back to `tests`.

mayhem:
When I run test.sh on this branch I get a lot of errors. :(

paramsingh:
> When I run test.sh on this branch I get a lot of errors. :(

Could you please post the errors? I don't get anything on my side. :/

Running `./integration-test.sh` doesn't work in this branch yet, because the docker-compose file does not list the correct dockerfiles. I've fixed that in #148.

mayhem:
>       conn = _connect(dsn, connection_factory=connection_factory, async=async)
E       OperationalError: (psycopg2.OperationalError) FATAL:  role "listenbrainz" does not exist

alastair:
when you run `test.sh` and it sees that the test db container exists, it assumes that it's all set up and doesn't need to do anything. Try and run `./test.sh -d` first. Perhaps we should add a `-f` force option.

mayhem:
@alastair: Yep, that was it. We could also check to see if the the container is running and warn the user if so....

PR #147, opened 2017-03-15
Modify pytest.ini so that test.sh doesn't run integration tests

paramsingh:
`./test.sh` now only runs tests in db, webserver, listenstore and redis-consumer.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,2 @@
 [pytest]
-testpaths = . db webserver listenstore redis-consumer
+testpaths = db webserver listenstore redis-consumer


paramsingh:
I've modified pytest.ini to only run tests from these four folders and not from `listenbrainz-server/tests`. `listenbrainz-server/tests` contains only integration tests right now and `./integration-test.sh` runs these, while `./test.sh` ignores the integration tests.

See comments from @mayhem here: https://chatlogs.metabrainz.org/brainzbot/metabrainz/2017-03-13/?msg=3855330&page=6 and some discussion with @alastair  [here](https://github.com/metabrainz/listenbrainz-server/pull/146#a-286779114).

mayhem:
Thanks!

PR #149, opened 2017-03-17
LB-140: Catch relevant errors during write to influx

paramsingh:
InfluxDB raises InfluxDBClientError or InfluxDBServerError if
something goes kaput during a write. Catch these exceptions so
that scripts like influx-writer don't crash on a failed write.
See LB-140.

### DIFF
--- a/influx-writer/influx-writer.py
+++ b/influx-writer/influx-writer.py
@@ -127,7 +127,7 @@ class InfluxWriterSubscriber(RedisPubSubSubscriber):
                 return
             except WriteFailException as e:
                 self.log.error("InfluxWriterSubscriber failed to write: %s" % str(e))
-                return
+                count = 0
 
             if not count:
                 continue
--- a/influx-writer/influx-writer.py
+++ b/influx-writer/influx-writer.py
@@ -8,6 +8,7 @@ sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", 
 from redis import Redis
 from redis_pubsub import RedisPubSubSubscriber, RedisPubSubPublisher, NoSubscriberNameSetException, WriteFailException, NoSubscribersException
 from influxdb import InfluxDBClient
+from influxdb.exceptions import InfluxDBClientError, InfluxDBServerError
 import ujson
 import logging
 from listen import Listen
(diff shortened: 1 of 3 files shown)


paramsingh:
I don't know exactly why we're catching `ValueError`'s. I didn't remove them but the code [here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenstore/listenstore/listenstore.py#L327-329) should never throw ValueError anyway.

mayhem:
Thanks. This improves the error handling, but it won't actually give us a better clue as to what is going wrong. How can we go one step further with this?

paramsingh:
@mayhem The error is coming from inside influxdb code.From a cursory look [here](https://github.com/influxdata/influxdb-python/blob/master/influxdb/client.py), the client sends a request to the influx http server and throws the error if the response code is not equal to an expected response code. I don't know what more we can do other than looking at the logs to see the exception message and try to get hints from there. I will add more logging so that we know what data is being sent when the error occurs, I guess. Should make it a bit easier to reproduce the problem hopefully.

mayhem:
Ah, got it. OK, in order for us to actually improve things we need to do the following then:

1) When this error happens, log the entire JSON listen data we're trying to push to the DB.
2) Skip and remove the offening listen from the queue. As it stands, the writer dies and the next time it starts up, it dies on exactly the same data, which isn't very useful.

Thanks!

paramsingh:
@mayhem, moved the `WriteFailException` so that it is raised after the cleanup takes place. This should mean that the culprit data gets skipped next time. Also made changes so that influx-writer doesn't die on `WriteFailException`s. Also, logging the entire data that we tried to write now, but we're writing in batches of listens and there's no way to know which listen was the culprit exactly.

mayhem:
> I don't know exactly why we're catching ValueError's. 

When I was coding this, there were a few ValueErrors that were thrown, so I caught them...

mayhem:
Right now this discards the whole block of listens, which is less than ideal if only one of them is problematic. I think we need to keep this in mind for improving once we get closer to the real problem.

paramsingh:
>When I was coding this, there were a few ValueErrors that were thrown, so I caught them...

Okay then, that's why I decided not to remove it. :+1: 

>Right now this discards the whole block of listens, which is less than ideal if only one of them is problematic. I think we need to keep this in mind for improving once we get closer to the real problem.

Noted, I did it this way for now because right now there's no way to know which listen is the problem. We can fix this when we know what the problem is exactly. This is just a stop-gap measure until then. :)

PR #150, opened 2017-03-19
Add logging to stdout to influx-writer

paramsingh:


influx-writer/influx-writer.py
```diff
@@ -117,16 +118,19 @@ def write(self, listen_dicts):
...
         while True:
             try:
                 count = self.subscriber()
             except NoSubscriberNameSetException as e:
                 self.log.error("InfluxWriterSubscriber has no subscriber name set.")
+                print("InfluxWriterSubscriber has no subscriber name set. Exiting.")
                 return
             except WriteFailException as e:
                 self.log.error("InfluxWriterSubscriber failed to write: %s" % str(e))
+                print("InfluxWriterSubscriber failed to write, skipping this batch of data.")
```
zas: You should use the same message imho.

```python
msg = "InfluxWriterSubscriber failed to write: %s" % str(e)
print(msg)
self.log.error(msg)
```

paramsingh: Done! Only errors get logged via the function to stderr. I haven't made a function for the start message because it seemed like overkill to me. Should I make one?

influx-writer/influx-writer.py
```diff
@@ -139,9 +141,15 @@ def start(self):
                 if self.time > 0:
                     self.log.error("Inserted %d rows in %.1fs (%.2f listens/sec). Total %d rows." % \
                         (count, self.time, count / self.time, self.total_inserts))
+                    print("Inserted %d rows in %.1fs (%.2f listens/sec). Total %d rows." % \
```
zas: Use `print_and_log_error()` too here ?

paramsingh: This wasn't an error so I didn't use `print_and_log_error` here, but changed now.

influx-writer/influx-writer.py
zas: That's fine for now. Is autobuild enabled for this project on docker hub ? If yes, merge and test the new container in prod, so we can sort out the issue.

paramsingh: >If yes, merge and test the new container in prod

I don't have write access to the repo, I can't merge.

mayhem:
The auto build on docker hub doesn't work, so I will manually push it. https://www.youtube.com/watch?v=vCadcBR95oU

influx-writer/influx-writer.py
```diff
@@ -117,16 +118,17 @@ def write(self, listen_dicts):
 
     def start(self):
         self.log.info("InfluxWriterSubscriber started")
+        print("InfluxWriterSubscriber started")
```
alastair: We have a logger. Why can't we set up a `StreamHandler` to send the same data to the console too??
https://docs.python.org/2/howto/logging-cookbook.html

PR #151, opened 2017-03-20
LB-140: Upgrade influxdb-python to 4.0.0

paramsingh:
3.0.0 has bugs causing problems in adding listens with track names with `\` in them. 4.0.0 has fixed that. See [this](https://github.com/influxdata/influxdb-python/commit/97952ba1bb0858774a2c10c073db9377adfe57a3) commit.

https://chatlogs.metabrainz.org/brainzbot/metabrainz/2017-03-20/?msg=3861171&page=2


mayhem:
Good work, thanks. Merging and pushing now.

PR #153, opened 2017-03-22
LB-141: Handle bad data in a better way in redis_pubsub

paramsingh:
Previously, if there was a single bad listen, our pubsub would
discard an entire batch of messages. Fix this by making the code
break the batch into half and try to write these halves. This goes
on recursively. See LB-141 for more details.

Also, moved the logger into the RedisPubSubSubscriber class
because previously, error or info messages were just being printed
to stdout.

### DIFF
--- a/redis_pubsub.py
+++ b/redis_pubsub.py
@@ -1,8 +1,10 @@
-from __future__ import print_function
+from __future__ import print_function, division
 import sys
 import ujson
 from redis import Redis
 from time import time, sleep
+import logging
+import json
 
 # For a detailed description of the PubSub Redis mechanism that is being implemented
 # here see this page: https://davidmarquis.wordpress.com/2013/01/03/reliable-delivery-message-queues-with-redis/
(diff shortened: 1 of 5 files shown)


paramsingh:
List of changes:

* modified `subscriber` so that it calls `write_batch` instead of `write` directly
* `write_batch` tries to write the entire batch and if it isn't able to do so, breaks the data into two and tries to write these halves. This is `O(n log n)` in complexity and shouldn't be bad for our use case (mostly batches will have ~1000 messages).
* Moved the logger into `RedisPubSubSubscriber` because all three of our subclasses created a logger of their own and the logging in `RedisPubSubSubscriber` was being done using `print`
* Added stuff to `WriteFailException` so that it stores the number of messages written. We throw `WriteFailException` whenever there is even one bad message in our batch.

redis_pubsub.py
```diff
@@ -1,8 +1,10 @@
-from __future__ import print_function
+from __future__ import print_function, division
```
paramsingh: Yeah, it is used to find the index by which data should be sliced.

PR #154, opened 2017-03-24
LB-142: Handle bad data in a better way in alpha_importer

paramsingh:
Data that is invalid used to make alpha_importer go into an
infinite loop. Change the send_batch function such that it finds
invalid data and skips it.

### DIFF
--- a/alpha_importer.py
+++ b/alpha_importer.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+from __future__ import print_function, division
 from redis import Redis
 import config
 import requests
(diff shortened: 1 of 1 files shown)


alpha_importer.py
```diff
@@ -61,14 +62,53 @@ def extract_data(batch):
...
+        # try to send the bad listen one more time and if it doesn't work
+        # log the error
+        r = requests.post(send_url, headers={'Authorization': 'Token {}'.format(token)}, data = ujson.dumps(data))
+        if r.status_code == 200:
+            return 1
+        else:
+            logger.error("Unable to submit bad listen to beta:")
+            logger.error(json.dumps(data, indent=4))
+            logger.error("Response code from beta: {}".format(r.status_code))
+            logger.error(r.text)
```
mayhem: Great!

PR #155, opened 2017-03-25
LB-143: Make alpha_importer work for users with spaces in their ids

paramsingh:
Change the order of the string pushed into redis and split by first
space only.


alastair:
This sounds like an ideal bug to have a test for to make sure it doesn't happen again! Can you add something that just checks this part?

paramsingh:
@alastair, will do.

PR #158, opened 2017-03-29
[WIP] Trying to fix the tests for jenkins

mayhem:
Still fails with:

py.test: error: argument --cov-report: invalid choice: 'xml:/data/coverage.xml' (choose from 'term', 'term-missing', 'annotate', 'html', 'xml', '')

### DIFF
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -22,7 +22,6 @@ services:
     build:
       context: ..
       dockerfile: Dockerfile
-    command: py.test
     volumes:
       - ..:/code/listenbrainz
     depends_on:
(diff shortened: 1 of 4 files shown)


docker/docker-compose.test.yml
```diff
@@ -22,7 +22,6 @@ services:
     build:
       context: ..
       dockerfile: Dockerfile
-    command: py.test
```
gentlecat: Command should be there if you want to use the same Dockerfile. Otherwise it will run whatever command is used in that Dockerfile, which I think is not `py.test`.

gentlecat: Output (reports) also needs to be specified here. Jenkins doesn't use test.sh script.

PR #159, opened 2017-03-30
LB-139: Make sure user-generated data doesn't get dropped.

paramsingh:
I've done this by flattening the `additional_info` dict in the listen class constructor and then making changes to InfluxListenStore so that it writes all fields present in the `additional_info` of a listen.

### DIFF
--- a/listen.py
+++ b/listen.py
@@ -6,7 +6,7 @@ import calendar
 
 def flatten_dict(d, seperator='', parent_key=''):
     """
-    Flattens a dictionary if it contains dictionaries inside it.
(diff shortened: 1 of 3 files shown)


paramsingh:
Also, I did some refactoring to add functions that convert Listen() objects to proper formats such as the format which is required for writing it into influx and the format that the api returns and then use these functions in the api code and in the listenstore. This seemed like a good idea because listens were being converted to different formats in different places. This task also seemed like a good place  to do the refactor to me, so I didn't make a different PR. However, I can remove that stuff if it should be a different PR or if we don't want to do it that way at all.

PR #162, opened 2017-04-06
[WIP] Test influx with large data

paramsingh:
This branch is not intended to be merged.

It adds a python script that we can use to make sure that influx can handle the data that we expect to be sent to us eventually when we get more popular.

### DIFF
--- a/db/user.py
+++ b/db/user.py
@@ -10,7 +10,7 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 def random_string():
-    l = random.randint(1, 10)
+    l = random.randint(50, 100)
     s = ""
     for _ in xrange(l):
         s += random.choice(string.ascii_lowercase)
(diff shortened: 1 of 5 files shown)


paramsingh:
Added `stress_test_influx.py` which takes as arguments the number of listens to send and the number of users to distribute these listens over. Then, the tester sends these listens with a thread for each user. Had to change the LB code so that it creates a new user for any auth_token that it does not have in the database. 

Also, before running the code, need to go into the `listenbrainz_web` container and use `set_rate_limits.py` to increase the rate limits of the API to very high values.

Usage of `stress_test_influx.py`:

```
usage: stress_test_influx.py [-h] [-l LISTEN_COUNT] [-u USER_COUNT]

optional arguments:
  -h, --help            show this help message and exit
  -l LISTEN_COUNT, --listencount LISTEN_COUNT
                        specify the number of listens to send while testing
  -u USER_COUNT, --usercount USER_COUNT
                        specify the number of users to spread the listens over
```

paramsingh:
My computer was being a little too slow to send 1B listens, I don't think that is a problem with the code. 1B HTTP requests to a local LB instance seemed to be a bit too much to handle for my computer.

paramsingh:
It won't work, if `MAKE_NEW_USERS` is false, because the code creates a new UUID as token for every simulated user.

paramsingh:
The script will get lots of `Unauthorized` responses, that is.

mayhem:
Sounds like I miscommunicated what I wanted earlier -- sorry. The last commit isn't going to be useful at all, can you please roll-back? I'll have to re-think this over the weekend.

paramsingh:
Okay then, removed.

paramsingh:
Maybe this should be closed as all of the work involving this is being done in the `influx-test-scribbles` branch?

PR #163, opened 2017-04-08
LB-125: Refactor code to use release instead of album

paramsingh:
Also, changed the name of the column in the BigQuery schema
and the PostGresListenStore.

### DIFF
--- a/admin/bigquery/listen-schema.json
+++ b/admin/bigquery/listen-schema.json
@@ -26,17 +26,17 @@
     },
     {
         "mode": "NULLABLE",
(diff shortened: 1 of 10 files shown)


mayhem:
Normally schema changing PRs need to be handled carefully and a script that changes existing DBs over is needed. However, after I finish abusing the DB on the beta server I can toss everything and start over, so this is not needed. I'll see about merging this after my abuse is done.

paramsingh:
Yea, I was also thinking of this being in the beta as we start over, not in the test.

PR #164, opened 2017-04-08
LB-147: Merge listenstore into main Python codebase

paramsingh:
Now there's no setup.py etc. files.

Also, removed the appends we were doing to `sys.path` because of listenstore being a different module.


redis-consumer/redis-consumer.py
```diff
@@ -2,7 +2,6 @@
 
 import sys
 import os
-sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../listenstore"))
 sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), ".."))
```
mayhem: While we're moving things around, lets move the redis_pubsub.py from the root into the listenstore as well. Then we can get rid of the second path appending bits.

listenstore/influx_listenstore.py
```diff
@@ -324,12 +135,12 @@ def get_timestamps_for_user(self, user_name):
...
             query = """SELECT first(artist_msid)
                          FROM listen
-                        WHERE user_name = '""" + user_name + "'"
+                        WHERE user_name = '""" + user_name + """'"""
             min_ts = self._select_single_timestamp(query)
 
             query = """SELECT last(artist_msid)
                          FROM listen
-                        WHERE user_name = '""" + user_name + "'"
+                        WHERE user_name = '""" + user_name + """'"""
```
paramsingh: Did this because sublime text's syntax highlighting got confused somehow with the previous version

PR #166, opened 2017-04-10
Do not store all of our data in one measurement

mayhem:
Previous we stored all of our listens in the listens measurement and that does not scale well. Creating a measurement for each user is much better.


paramsingh:
Anything for better performance! LGTM! :shipit:

mayhem:
Merged, but via rebase/merge, so github did't pick up the change.

PR #168, opened 2017-04-19
LB-152: Use new schema when deduplicating listens in influx-writer

paramsingh:
We missed updating the influx-writer code when we changed the influx schema.

### DIFF
--- a/influx-writer/influx-writer.py
+++ b/influx-writer/influx-writer.py
@@ -40,6 +40,11 @@ class InfluxWriterSubscriber(object):
         self.inserts = 0
         self.time = 0
 
+    @staticmethod
+    def escape(value):
+        if not value:
+            return value
+        return "\"{0}\"".format( value.replace("\\", "\\\\").replace( "\"", "\\\"").replace( "\n", "\\n"))
 
     @staticmethod
     def static_callback(ch, method, properties, body, obj):
@@ -68,7 +73,7 @@ class InfluxWriterSubscriber(object):
                 break
             except pika.exceptions.ConnectionClosed:
                 self.connect_to_rabbitmq()
-            
+
         count = len(listens)
         self.redis.decr(INCOMING_QUEUE_SIZE_KEY, count)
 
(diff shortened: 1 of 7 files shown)


paramsingh:
How would I go about writing tests for dedup? There are no values returned anywhere.

influx-writer/influx-writer.py
```diff
@@ -36,6 +36,12 @@ def __init__(self, ls, influx, redis):
         self.inserts = 0
         self.time = 0
 
+    @staticmethod
```
mayhem: We already have this somewhere -- lets not duplicate this and put this into a common module please.

influx-writer/influx-writer.py
```diff
@@ -58,17 +64,13 @@ def write(self, listen_dicts):
...
         # quering for artist name here, since a field must be included in the query.
         query = """SELECT time, artist_name
-                     FROM listen
-                    WHERE user_name = '%s'
-                      AND time >= %d000000000
+                     FROM "\\"%s\\""
+                    WHERE time >= %d000000000
                       AND time <= %d000000000
                 """ % (user_name, min_time, max_time)
+
```
mayhem: Hmm. Does the user name need escaping here?

paramsingh: I did it how it was done here: https://github.com/metabrainz/listenbrainz-server/blob/master/listenstore/influx_listenstore.py#L199

I guess we should escape the username there also, I'll fix and send soon.

influx-writer/influx-writer.py
paramsingh: Would a `utils.py` file in the root folder be acceptable?

mayhem: For now, that is fine. Ideally I would like to get rid of all python modules (not scripts) from root folder... But we can tackle that later.

paramsingh:
Sorry, I haven't been able to get to fixing this, been swamped with lots of stuff, I'll try to do it this week.

paramsingh:
Test added, moved the escape code to `listenstore/utils.py` and  escaped everything in `influx-writer.py` and `InfluxListenStore`. I think everything's fixed now, please take a look again :)

tests/integration/test_influx_writer.py
```diff
@@ -0,0 +1,51 @@
...
+        r = self.send_single_listen(user)
+        self.assert200(r)
+        time.sleep(5)
+        r = self.send_single_listen(user)
+        self.assert200(r)
+        time.sleep(5)
+
+        to_ts = int(time.time())
+        listens = self.ls.fetch_listens(user['musicbrainz_id'], to_ts=to_ts)
+        print(listens)
```
mayhem: i guess this isn't really needed here. I'll take care of removing it.

PR #169, opened 2017-04-24
Use rabbitmq for queueing

mayhem:
This PR swaps out the Redis-pubsub for a proper, less memory hungry, queue. I chose RabbitMQ since we already use that in production. This PR are improves the error handling and resilience of the writer containers and pausing on startup if a service isn't available yet.

New redis keys were needed to keep track of queue stats, so I created a new redis_keys.py module for the purpose of having a central place to track redis keys. 

The per-user listen counts have a bug in them in this PR, but I am already working on an improved PR for counting user and total listens using influxdb's continuous queries that make things like counting listens much easier.


bigquery-writer/bigquery-writer.py
```diff
@@ -1,38 +1,60 @@
...
         self.time = 0
 
-    def write(self, listens):
+    def connect_to_rabbitmq(self):
+        while True:
+            try:
+                self.connection = pika.BlockingConnection(pika.ConnectionParameters(host=config.RABBITMQ_HOST, port=config.RABBITMQ_PORT))
+                break
+            except Exception as e:
+                self.log.error("Cannot connect to rabbitmq: %s, sleeping 2 seconds")
```
paramsingh: Seems like you forgot to add the error message here.

paramsingh:
All tests pass, looks good to me. :+1:

bigquery-writer/bigquery-writer.py
mayhem: Good catch.

bigquery-writer/bigquery-writer.py
```diff
@@ -1,38 +1,60 @@
...
         self.time = 0
 
-    def write(self, listens):
+    def connect_to_rabbitmq(self):
+        while True:
+            try:
+                self.connection = pika.BlockingConnection(pika.ConnectionParameters(host=config.RABBITMQ_HOST, port=config.RABBITMQ_PORT))
+                break
+            except Exception as e:
+                self.log.error("Cannot connect to rabbitmq: %s, sleeping 2 seconds" % str(e))
```
gentlecat: `retrying in 2 seconds`might be a more useful message.

bigquery-writer/bigquery-writer.py
```diff
@@ -1,38 +1,60 @@
...
+                sleep(2)
+
+
+    @staticmethod
+    def static_callback(ch, method, properties, body, obj):
+        return obj.callback(ch, method, properties, body)
+
+
+    def callback(self, ch, method, properties, body):
```
gentlecat: Both `ch` and `properties` arguments are unused.

bigquery-writer/bigquery-writer.py
```diff
@@ -1,38 +1,60 @@
...
+            try:
+                self.connection = pika.BlockingConnection(pika.ConnectionParameters(host=config.RABBITMQ_HOST, port=config.RABBITMQ_PORT))
+                break
+            except Exception as e:
+                self.log.error("Cannot connect to rabbitmq: %s, sleeping 2 seconds" % str(e))
+                sleep(2)
+
+
+    @staticmethod
+    def static_callback(ch, method, properties, body, obj):
```
gentlecat: It might be better to make `obj` a first argument since it's the most important in the context of this method. The rest can be passed as keyword arguments without the need to explicitly specify them. These can be inferred from the `callback` method.

influx-writer/influx-writer.py
```diff
@@ -15,29 +14,78 @@
...
-
-
 DUMP_JSON_WITH_ERRORS = False
+ERROR_RETRY_DELAY = 3 # number of seconds to wait until retrying an operation
+
 
-class InfluxWriterSubscriber(RedisPubSubSubscriber):
-    def __init__(self, ls, influx, redis):
-        RedisPubSubSubscriber.__init__(self, redis, KEYSPACE_NAME_INCOMING, __name__)
+class InfluxWriterSubscriber(object):
```
gentlecat: I see that `InfluxWriterSubscriber` and `BigQueryWriter` have some common methods related to writing. Having an abstract `ListenWriter` class can be useful.

influx-writer/influx-writer.py
```diff
@@ -15,29 +14,78 @@
...
+        return obj.callback(ch, method, properties, body)
+
+
+    def connect_to_rabbitmq(self):
+        while True:
+            try:
+                self.connection = pika.BlockingConnection(pika.ConnectionParameters(host=config.RABBITMQ_HOST, port=config.RABBITMQ_PORT))
+                break
+            except Exception as e:
+                self.log.error("Cannot connect to rabbitmq: %s, sleeping 2 seconds")
```
gentlecat: `%s` is there, but the string is not formatted.

influx-writer/influx-writer.py
gentlecat: ...or not even abstract since `connect_to_rabbitmq` is exactly the same.

influx-writer/influx-writer.py
```diff
@@ -15,29 +14,78 @@
...
+                sleep(2)
+
+
+    def callback(self, ch, method, properties, body):
```
gentlecat: Unused arguments here as well.

influx-writer/influx-writer.py
```diff
@@ -15,29 +14,78 @@
 from time import time, sleep
```
gentlecat: `time` is unused.

bigquery-writer/bigquery-writer.py
```diff
@@ -1,38 +1,60 @@
...
-from redis import Redis
-from listenstore.redis_pubsub import RedisPubSubSubscriber, RedisPubSubPublisher, NoSubscriberNameSetException, WriteFailException
 import ujson
 import logging
-from listen import Listen
+import pika
+from datetime import datetime
 from time import time, sleep
+sys.path.append(os.path.join(os.path.dirname(os.path.realpath(__file__)), ".."))
+from listen import Listen
```
gentlecat: `Listen` and `datetime` are unused.

webserver/rabbitmq_connection.py
```diff
@@ -0,0 +1,23 @@
...
+import pika
+
+_rabbitmq = None
+
+def init_rabbitmq_connection(app):
+    """Create a connection to the RabbitMQ server."""
+    global _rabbitmq
+
+    if not app.config.has_key("RABBITMQ_HOST"):
+        app.logger.error("RabbitMQ host:port not defined. Sleeping 2 seconds, and exiting.")
```
gentlecat: > Sleeping 2 seconds

Where?

bigquery-writer/bigquery-writer.py
gentlecat: In other places too.

bigquery-writer/bigquery-writer.py
```diff
@@ -63,28 +85,69 @@ def write(self, listens):
...
+            try:
+                self.channel.basic_ack(delivery_tag = method.delivery_tag)
+                break
+            except pika.exceptions.ConnectionClosed:
+                self.connect_to_rabbitmq()
+
+        self.redis.decr(UNIQUE_QUEUE_SIZE_KEY, count)
 
         # Clear the start time, since we've cleaned out the batch
         batch_start_time = 0
```
gentlecat: This variable is unused.

bigquery-writer/bigquery-writer.py
```diff
@@ -63,28 +85,69 @@ def write(self, listens):
...
+
+        if not hasattr(config, "RABBITMQ_HOST"):
+            self.log.error("RabbitMQ service not defined. Sleeping 2 seconds and exiting.")
+            sleep(2)
+            return
 
         # if we're not supposed to run, just sleep
         if not config.WRITE_TO_BIGQUERY:
-            sleep(1000)
+            sleep(66666)
```
gentlecat: Why has the duration increased?

bigquery-writer/bigquery-writer.py
```diff
@@ -97,38 +160,37 @@ def start(self):
             sleep(1000)
             return
```
gentlecat: Can all the checks above be consolidated in one function? They all call `sleep` and return.

gentlecat: And why is `sleep` call necessary?

docker/docker-compose.test.yml
```diff
@@ -18,6 +18,13 @@ services:
       INFLUXDB_HTTP_LOG_ENABLED: 'true'
       INFLUXDB_CONTINUOUS_QUERIES_LOG_ENABLED: 'false'
 
+  rabbitmq:
+    image: rabbitmq:3.6.9
```
gentlecat: We use version `3.6.5` in production. https://github.com/metabrainz/docker-server-configs/blob/master/scripts/services.sh#L481

influx-writer/influx-writer.py
```diff
@@ -15,29 +14,78 @@
...
+            if self.time > 0:
+                self.log.info("Inserted %d rows in %.1fs (%.2f listens/sec). Total %d rows." % \
+                    (self.inserts, self.time, self.inserts / self.time, self.total_inserts))
+            self.inserts = 0
+            self.time = 0
+
+        return ret
+
+
+    def write(self, listen_dicts):
```
gentlecat: This method is way too long and complex. I'd at least split it up in multiple methods that are easier to understand and test.

influx-writer/influx-writer.py
```diff
@@ -15,29 +14,78 @@
...
+from redis_keys import INCOMING_QUEUE_SIZE_KEY, UNIQUE_QUEUE_SIZE_KEY
 
 REPORT_FREQUENCY = 5000
-SUBSCRIBER_NAME = "in"
-KEYSPACE_NAME_INCOMING = "ilisten"
-KEYSPACE_NAME_UNIQUE = "ulisten"
-
-
 DUMP_JSON_WITH_ERRORS = False
+ERROR_RETRY_DELAY = 3 # number of seconds to wait until retrying an operation
```
gentlecat: It seems like this value is just used in only one place. Can it be integrated into other parts with retrying within this module?

gentlecat:
*Some of the issues I pointed out can be detected with [Pylint](https://www.pylint.org/).*

bigquery-writer/bigquery-writer.py
mayhem: If there is no sleep, the container will just keep restarting, which just sucks resources.

bigquery-writer/bigquery-writer.py
mayhem: The actual value doesn't matter as long as it is large to prevent the container from restarting too often.

bigquery-writer/bigquery-writer.py
mayhem: This is callback called by pika -- I can't change the signature of the callback.

bigquery-writer/bigquery-writer.py
mayhem: I don't really see the need to consolidate them in one function -- I don't believe in creating new functions just to keep functions below an artificial length. I believe that makes the code harder to read in the end.

influx-writer/influx-writer.py
mayhem: I thought about that,but I didn't really want to create another py module just for one function. I'll do this should I need to add another common method.

influx-writer/influx-writer.py
mayhem: I prefer to keep it as it, in order to ensure readability.

bigquery-writer/bigquery-writer.py
gentlecat: Assuming it's the one in the `start` below, it can be modified like this:
```python
self.channel.basic_consume(
    lambda ch, method, properties, body:
        self.static_callback(self, ch, method, properties, body),
    queue='unique'
)
```

influx-writer/influx-writer.py
gentlecat: That's my point though, it's kind of difficult to read.

Example of one of the changes that can be made:
```python
@staticmethod
def listens_time_range(listens):
    """Calculate the time range that a set of listens covers.
    
    Args:
        listens (list): List of dictionaries with listen data.
    
    Returns:
        Tuple with two values: minimum time and maximum time.
    """
    min_time = 0
    max_time = 0
    for listen in listens:
        t = int(listen['listened_at'])
        if not max_time:
            min_time = max_time = t
            continue
        if t > max_time:
            max_time = t
        if t < min_time:
            min_time = t
    return min_time, max_time

def write(self, listen_dicts):
    submit = []
    unique = []

    min_time, max_time = listens_time_range(listen_dicts)

    ...
```
I don't need to know how `listens_time_range` is implemented to understand how `write` works. `listens_time_range` is also very easy to test now.

influx-writer/influx-writer.py
gentlecat: Fair enough. My point about an abstract writer class still stands. Other methods that can go in it are `callback`, `static_callback`, and `start`. Here's some more information about them and why they can be useful:

* https://www.python.org/dev/peps/pep-3119/
* http://stackoverflow.com/a/3571946/272770
* https://docs.python.org/2/library/abc.html

mayhem: That last comment was great!

(still not enough to sway me, but still. :) )

influx-writer/influx-writer.py
mayhem: time is actually used.

gentlecat: Yep, misread that.

influx-writer/influx-writer.py
mayhem: Done.

influx-writer/influx-writer.py
mayhem: I've spent over an hour trying to remove some args, but it always fails now. I'd rather spent my time doing other things than fixing code that is working.

webserver/rabbitmq_connection.py
mayhem: Added.

influx-writer/influx-writer.py
mayhem: You left out the part where the user_name is determined for this batch of listens, so now the clean single purpose function needs to be have two purposes or the data needs to be processed a second time.

gentlecat: Don't know if all the listens have `user_name` in them. If yes, then it can be retrieved from the first one. If not, get from the first one that's encountered.

PR #170, opened 2017-04-27
Add an option for creating fake users to api

paramsingh:
As we talked about it here: https://chatlogs.metabrainz.org/brainzbot/metabrainz/2017-04-19/?msg=3888764&page=4

### DIFF
--- a/webserver/views/api.py
+++ b/webserver/views/api.py
@@ -10,6 +10,8 @@ from api_tools import insert_payload, log_raise_400, validate_listen, MAX_LISTEN
     DEFAULT_ITEMS_PER_GET, LISTEN_TYPE_SINGLE, LISTEN_TYPE_IMPORT, LISTEN_TYPE_PLAYING_NOW
 import time
 
+CREATE_FAKE_USERS_FOR_STRESS_TESTING = False
+
 api_bp = Blueprint('api_v1', __name__)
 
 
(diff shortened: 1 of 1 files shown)


webserver/views/api.py
```diff
@@ -140,7 +142,10 @@ def _validate_auth_header():
 
     user = db.user.get_by_token(auth_token)
     if user is None:
-        user = db.user.create_user_with_token(auth_token)
+        if CREATE_FAKE_USERS_FOR_STRESS_TESTING:
```
alastair: I know this has been merged, but is there a way that this can be integrated without having a create operation in a verify method?
Is it possible to make the stress-test tool send 2 requests (1 POST to create an account [under a similar guard as this one], and then the POST to save data)?
Is this setting used when `DEBUG` is set to `False`? If no, we could have a guard to make sure that `CREATE_FAKE_USERS_FOR_STRESS_TESTING && !DEBUG` is never true

mayhem: Hmm, I see your point. This is going to be used very rarely, in fact I plan to use it only one more time before our beta. Then this code can go away... But I think adding the guard is a pretty good idea.

alastair: If you pinky swear that we're going to delete this code very soon, I'll let it slide, but a `DEBUG` guard would make me much less anxious

mayhem: https://tickets.metabrainz.org/browse/LB-161 -- I don't do NDAs or pinky swears. :D

PR #171, opened 2017-05-14
LB-159: Change influx-writer so that dedup works correctly for multiple users in same batch

paramsingh:
Previously, if there was more than one user in the same batch of
listens, influx-writer would count listens with the same timestamps
as a duplicate (even though they were from different users). This
change fixes that.

### DIFF
--- a/influx-writer/influx-writer.py
+++ b/influx-writer/influx-writer.py
@@ -89,57 +89,68 @@ class InfluxWriterSubscriber(object):
     def write(self, listen_dicts):
         submit = []
         unique = []
-
-        # Calculate the time range that this set of listens coveres
-        min_time = 0
-        max_time = 0
-        user_name = ""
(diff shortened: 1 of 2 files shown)


influx-writer/influx-writer.py
```diff
@@ -89,57 +89,68 @@ def callback(self, ch, method, properties, body):
...
-        for result in results.get_points(measurement=get_measurement_name(user_name)):
-            dt = datetime.strptime(result['time'] , "%Y-%m-%dT%H:%M:%SZ")
-            timestamps[int(dt.strftime('%s'))] = 1
-
         duplicate_count = 0
         unique_count = 0
+
+        # Partition the listens on the basis of user names
+        # and then store the time range for each user
+        min_times = {}
```
paramsingh: Done!

PR #172, opened 2017-05-14
LB-153: Make api_compat use generic ListenStore objects to get play counts

paramsingh:
Instead of using the PostgresListenStore directly, we now take
a listenstore as argument and get listen count from there.


paramsingh:
Changed the name of the user in the test because the same user was being used in other tests in InfluxListenStore which seemed to be causing tests there to fail. Hopefully that is okay.

PR #174, opened 2017-05-17
LB-62: Begin work on splitting LB into two different flask apps

paramsingh:
I have started with splitting `webserver` into two different apps 
* `webapp`: This contains the `index`, `login` and `user` blueprints.
* `api`: This contains `api` and `api_compat`.

In the dev environment, these are connected by  [DispatcherMiddleware](http://werkzeug.pocoo.org/docs/0.12/middlewares/#werkzeug.wsgi.DispatcherMiddleware) (which is how it works in CB, as far as I have seen).

This just adds initial support for two different flask applications. I haven't gotten into uwsgi configs yet, because that seemed like production stuff to me. I would appreciate some directions on what exactly I need to do there.


paramsingh:
Also, I added another blueprint `api_404` in the webapp, which returns a 404 with relevant message whenever an API request is made to the webapp.

And because `flask_testing.TestCase` requires the presence of a single app, we still use a single app with all the blueprints in the Integration tests. If there is a more preferred way to do testing, I would be happy to change.

webserver/__init__.py
```diff
@@ -92,18 +106,42 @@ def create_app_rtfd():
...
     return app
 
-
-def _register_blueprints(app):
-    from webserver.views.index import index_bp
-    from webserver.views.login import login_bp
-    from webserver.views.api import api_bp
-    from webserver.views.api_compat import api_bp as api_bp_compat
-    from webserver.views.user import user_bp
+def _register_all_blueprints(app):
```
alastair: Is this just the same as 
```
_register_web_blueprints(app)
_register_api_blueprints(app)
```
?

webserver/webapp/views/api_404.py
```diff
@@ -0,0 +1,19 @@
...
+@api_404_bp.route('/1/submit_listens', methods=['GET', 'POST'])
+@api_404_bp.route('/1/user/<user_name>/listens', methods=['GET'])
+@api_404_bp.route('/2.0/', methods=['GET', 'POST'])
+@api_404_bp.route('/api/auth/', methods=['GET', 'POST'])
+def webapp_api_call(user_name=None):
+    """
+    This route is used to raise 404s when api calls are made to the webapp
+    instead of the api.
+    """
+    raise NotFound("Incorrect address used in call to API, please use api.listenbrainz.org instead.")
```
alastair: The body of this message should be JSON, in the same format as other error messages sent by the API.

mayhem:
As agreed, we don't want to do this anymore.

PR #175, opened 2017-05-18
LB-157: Handle bad listens in a smarter way in influx-writer

paramsingh:
Recursively divide the data into two parts until we find the listen
that was failing. This means that only the bad listen gets rejected
and not an entire batch.

### DIFF
--- a/influx-writer/influx-writer.py
+++ b/influx-writer/influx-writer.py
@@ -101,7 +101,6 @@ class InfluxWriterSubscriber(object):
         if not data:
             return 0
 
-        done = False
         failure_count = 0
         while True:
             try:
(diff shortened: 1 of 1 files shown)


paramsingh:
As discussed on IRC, it is hard to find a bad listen that our API accepts but influx rejects, so I haven't added an integration test for this. However this functionality still is important because exceptions have been triggered in the past and can happen in the future also.

influx-writer/influx-writer.py
```diff
@@ -86,6 +86,60 @@ def callback(self, ch, method, properties, body):
...
+
+        if not data:
+            return 0
+
+        done = False
+        failure_count = 0
+        while True:
+            try:
+                self.ls.insert(data)
+                done = True
```
mayhem: You could skip the whole done flag if you just returned len(data) here -- that would also simplify the code below.

paramsingh: Done.

PR #176, opened 2017-05-20
LB-155: Make Redis connection establishment more reliable

paramsingh:
Do a ping and if ConnectionErrors are thrown, sleep for two
seconds and try to establish a new connection.

### DIFF
--- a/alpha_importer.py
+++ b/alpha_importer.py
@@ -10,8 +10,9 @@ import logging
 from logging.handlers import RotatingFileHandler
 import os
 from listenstore import InfluxListenStore
+import redis
 
-redis_connection = Redis(host=config.REDIS_HOST, port=config.REDIS_PORT)
+redis_connection = None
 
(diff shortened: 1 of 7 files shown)


paramsingh:
Doing a ping because `ConnectionError`s get thrown only when the a request is made to the redis server. See here: https://stackoverflow.com/questions/12857604/python-how-to-check-if-redis-server-is-available

Fixed the connection establishment for 

1. `webserver`
2. `alpha_importer`
3. `influx_writer`
4. `bigquery_writer`

The last two had the loops already but didn't ping the server, so only needed to add a ping. :)

alastair:
Does it make sense to quit after some number of connection attempts (say, 5?) instead of continually trying to connect and failing?

mayhem:
The whole point is to keep trying -- containers may come up and down at any point. Waiting nicely and retrying allows the system to self heal, without having to restart containers.

PR #177, opened 2017-05-21
LB-155: Make influx connection establishment more reliable

paramsingh:
If ConnectionErrors are thrown, sleep for 2 seconds and then try again. 

Similar to #176.


paramsingh:
Just did a rebase from master to fix conflicts after merging of #176.

PR #178, opened 2017-05-22
LB-155: Improve Database connection establishment

paramsingh:
Changes to `db.init_db_connection` and `webserver.external.messybrainz.init_db_connection` so that it `OperationalErrors` thrown when db is not up cause the code to wait and try again.

### DIFF
--- a/db/__init__.py
+++ b/db/__init__.py
@@ -1,6 +1,9 @@
+from __future__ import print_function
 import sqlalchemy
 from sqlalchemy import create_engine
 from sqlalchemy.pool import NullPool
+import time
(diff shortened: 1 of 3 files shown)


paramsingh:
Also, removed the connection to PostgresListenStore since that isn't used anymore. Tests don't fail for me, so looks good, but take a look. :)


`create_postgres` creates an object of PostgresListenStore which establishes a different connection to PG in its `__init__` method while `db.init_db_connection` creates a connection which is used by the `db` module to get user data etc. I've removed the first one and not the second one (which is used by the webserver extensively)

PR #179, opened 2017-05-25
LB-138: Create a listenbrainz module so we don't change sys.path anymore

paramsingh:
This PR moves almost all our code into a module called `listenbrainz` so that we don't have to do
all the changes to `sys.path` anymore. Had to change some code in `admin/influx/create_db.py` to do this because that is the only part of the codebase that needs to be somewhere else.


paramsingh:
Note: This PR is mostly inspired from how CritiqueBrainz does things. We have a module called `critiquebrainz` there and everything else is divided into sub-modules. I think this makes for good code organization, but am open to other opinions. 

Also, I know LB-138 was closed but we did have some `sys.path.append`s in the code, so this should fix that. :)

admin/influx/create_db.sql
```diff
@@ -0,0 +1 @@
+CREATE DATABASE listenbrainz;
```
alastair: is it standard in the infux world to name script files `.sql`?

docker/docker-compose.integration.yml
```diff
@@ -22,7 +22,7 @@ services:
     build:
       context: ..
       dockerfile: Dockerfile
-    command: python influx-writer/influx-writer.py
+    command: python -m "listenbrainz.influx-writer.influx-writer"
```
alastair: :+1: this is cool

PR #180, opened 2017-05-27
LB-156: Reconnect to RabbitMQ if connection gets closed 

paramsingh:
Catch `pika.exceptions.ConnectionClosed`, reconnect to RabbitMQ and try to submit again.


paramsingh:
Question: Was there a reason we're doing channel creation and the publish part seperately in two different try blocks? There probably is, BUT if not, I can put both of them inside a single try block, catch `ConnectionClosed` and `NoFreeChannels` seperately. That would make the code a bit easier to understand, I think.

mayhem:
My hope was to have different error messages so that we can tell what exactly happened, rather than having only one. But now that I look at it, the error messages are the same. Can you please make them distinct and then the separate blocks make sense?

paramsingh:
Improved the messages logged.

PR #181, opened 2017-05-28
Add .travis.yml for continuous integration.

paramsingh:
Would just need to enable travis for this repo on https://travis-ci.org/metabrainz after this, and it should run tests on all pull requests then.

I didn't run `test.sh` and `integration-test.sh` directly in travis.yml because they always return 0 in the end.

Example build: https://travis-ci.org/paramsingh/listenbrainz-server/builds/236926159

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,13 @@ install:
 
   # Build the unit test docker-compose project
   - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test build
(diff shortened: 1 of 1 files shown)


paramsingh:
Gonna have to fix this to use `init_influx` now.

paramsingh:
One test seems to be failing, I'm gonna close this for a while and check what the problem is.

paramsingh:
Seemed like the test was failing because `influx-writer` wasn't up before the first test ran. Adding a sleep before running the tests and after bringing up the containers seems to have fixed it.

PR #184, opened 2017-05-30
LB-162: Remove infinite loop when disconnected from RabbitMQ

paramsingh:
Add a test and fix a bug due to which the code always returned 200 even if ServiceUnavailable or InternalServerError was raised and fix a bug in the handling of playing_now that showed its face when I fixed the previous one.

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -67,7 +67,7 @@ def _send_listens_to_queue(listen_type, listens):
                 )
             except Exception, e:
                 current_app.logger.error("Redis rpush playing_now write error: " + str(e))
-                raise InternalServerError("Cannot record playing_now at this time.")
+                raise ServiceUnavailable("Cannot record playing_now at this time.")
         else:
             submit.append(listen)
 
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/views/api_tools.py
```diff
@@ -58,49 +60,46 @@ def _send_listens_to_queue(listen_type, listens):
...
+            channel.exchange_declare(exchange='incoming', type='fanout')
+            channel.queue_declare('incoming', durable=True)
+        except pika.exceptions.ConnectionClosed as e:
+            current_app.logger.error("Connection to rabbitmq closed while creating channel: %s" % str(e))
+            raise ServiceUnavailable("Cannot submit listens to queue, please try again later.")
+        except pika.exceptions.NoFreeChannels as e:
+            if channel:
+                channel.close() # just in case
+            current_app.logger.error("Cannot create a rabbitmq channel: %s" % str(e))
+            raise ServiceUnavailable("Cannot submit listens to queue, please try again later.")
```
alastair: Will this error message be in json format?

paramsingh: Nope, HTML is what is returned as of now because of these error handlers: https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/errors.py#L53

PR #189, opened 2017-06-05
LB-165: Track when user was last logged in  

paramsingh:
This involves a schema change. Not sure if adding `admin/sql/updates/add_last_login.sql` is enough or if I need to do something more.

Added an update_last_login function to listenbrainz.db.user and called it when user gets logged in.

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -4,7 +4,8 @@ CREATE TABLE "user" (
   id             SERIAL,
   created        TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
   musicbrainz_id VARCHAR NOT NULL,
-  auth_token     VARCHAR
+  auth_token     VARCHAR,
+  last_login     TIMESTAMP WITH TIME ZONE
 );
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id);
 
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -4,6 +4,7 @@ import uuid
 import sqlalchemy
 from listenbrainz.db.exceptions import DatabaseException
 import logging
+import time
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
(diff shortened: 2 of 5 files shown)


admin/sql/create_tables.sql
```diff
@@ -4,7 +4,8 @@ CREATE TABLE "user" (
   id             SERIAL,
   created        TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
   musicbrainz_id VARCHAR NOT NULL,
-  auth_token     VARCHAR
+  auth_token     VARCHAR,
+  last_login     TIMESTAMP WITH TIME ZONE
```
alastair: This value can be null. What does a null last login mean? Will you check this correctly every time that you check a user's last login?
Perhaps a better value for this would be `created` for all existing users, then alter the column to make it `NOT NULL` (in the update - perhaps since this has been merged you have another update script to set this)

listenbrainz/db/user.py
```diff
@@ -117,7 +118,7 @@ def get_by_token(token):
     """
     with db.engine.connect() as connection:
         result = connection.execute(sqlalchemy.text("""
-            SELECT id, created, musicbrainz_id
+            SELECT id, created, musicbrainz_id, last_login
```
alastair: I'm just glancing at this duplicated code 3× in the same file and wondering if we could have a `USER_SELECT_FIELDS` variable or something which lists these fields

listenbrainz/db/user.py
```diff
@@ -164,3 +165,25 @@ def get_or_create(musicbrainz_id):
         create(musicbrainz_id)
         user = get_by_mb_id(musicbrainz_id)
     return user
+
+def update_last_login(musicbrainz_id, ts=int(time.time())):
```
alastair: Is there ever a case where we want to set the last login time to a time that isn't now? If not, I would consider removing the argument and making this always set the last_login to the current time. As @mayhem suggested, it would be better to use the database to set this (`SET last_login = now()`)

admin/sql/create_tables.sql
paramsingh: Good point, will do this.

listenbrainz/db/user.py
paramsingh: I guess that makes sense for all the `get_by_*` functions right now. There may be other functions that don't need all the user select fields later.

listenbrainz/db/user.py
paramsingh: >Is there ever a case where we want to set the last login time to a time that isn't now?

Nope, I guess I overthought this a tad. I'll change it. :+1:

PR #190, opened 2017-06-05
Use Flask pika connection pooler

mayhem:
As stated on the tin -- there are still a ton of waring messages that are supposed to be benign. At least the code works every time, regardless of the various messages. That is an improvement. :)

### DIFF
--- a/listenbrainz/webserver/rabbitmq_connection.py
+++ b/listenbrainz/webserver/rabbitmq_connection.py
@@ -1,6 +1,7 @@
 import sys
 from time import sleep
-import pika
+from flask import Flask
+from flask.ext.pika import Pika as FPika
 
 _rabbitmq = None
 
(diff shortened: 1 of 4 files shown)


paramsingh:
Looks good to me. :+1: 

If I had to point out something, adding the version of Flask-Pika to `requirements.txt` would be it.

paramsingh:
:shipit:

listenbrainz/webserver/rabbitmq_connection.py
```diff
@@ -1,25 +1,39 @@
...
     """Create a connection to the RabbitMQ server."""
     global _rabbitmq
 
-    app.logger.info("Call to init_rabbit_mq_connection")
     if "RABBITMQ_HOST" not in app.config:
         app.logger.error("RabbitMQ host:port not defined. Sleeping 2 seconds, and exiting.")
         sleep(2)
         sys.exit(-1)
 
+    FLASK_PIKA_PARAMS = {
```
alastair: should these be here, or in config.py?

listenbrainz/webserver/rabbitmq_connection.py
```diff
@@ -1,25 +1,39 @@
...
+    app.config['FLASK_PIKA_POOL_PARAMS'] = FLASK_PIKA_POOL_PARAMS
+
     while True:
         try:
-            _rabbitmq = pika.BlockingConnection(pika.ConnectionParameters(host=app.config['RABBITMQ_HOST'], port=app.config['RABBITMQ_PORT']))
-            app.logger.info("Connected connect to rabbitmq!")
+            _rabbitmq = FPika(app)
             return
-        except pika.exceptions.ConnectionClosed:
+        except Exception as err:
```
alastair: Why the broadening of the exception? Is there not a generic flask-pika exception that we can catch here?

listenbrainz/webserver/rabbitmq_connection.py
mayhem: Normally it should be in config.py, but that is already getting cluttered. When I work on deployment I'll see about adding production config for the pooling settings.

listenbrainz/webserver/rabbitmq_connection.py
mayhem: Not really. Pika exceptions are generally a mess, which is why I catch broad exceptions for as few lines as possible.

PR #191, opened 2017-06-06
Listen counts

mayhem:
Implement an influxdb suited method for counting total number of listens and at the same time creating data for a timeline of how the count of listens grows over time.

The idea is pretty straightforward:

1) When new listens are entered into the DB, we also write count points with a retention policy of a week.
2) Periodically we tally up those points and then write a total count data point into the "listen_counts" measurement.
3) To figure out the number of listens currently in the DB, we take the last listen_count point and sum up all of the new points added from 1). 

With this setup, figuring out the total number of listens becomes a O(1) operation, as opposed to O(n).


paramsingh:
Tests pass for me locally, but not on travis, I wonder why. :?

alastair:
How can I test this myself? Is the rest of the system stable/bug-free enough that I can run it, or should I wait for you to fix conflicts and rebase/merge master?

mayhem:
superceded by #209

PR #192, opened 2017-06-06
LB-167: Create schema for storing statistics in PostgreSQL

paramsingh:
Create new tables with JSONB columns to store statistics.
One table each for user, artist, release and recordings. All
these tables are inside a new schema `statistics`

Also minor changes in `db/testing.py` to bring the SQL there
out of the code and into the `admin/sql` directory.


paramsingh:
@alastair @mayhem Requesting review :)

One thing that I was thinking of: Is it okay to use msids as primary keys or would using a `SERIAL` like in the `"user"` table be better for performance?

mayhem:
SERIAL is always best for such things.

paramsingh:
Okay, noted. I'll change it to use SERIAL then.

listenbrainz/db/testing.py
```diff
@@ -32,16 +32,10 @@ def init_db(self):
         db.run_sql_script(os.path.join(ADMIN_SQL_DIR, 'create_indexes.sql'))
 
     def drop_tables(self):
-        with db.engine.connect() as connection:
-            connection.execute('DROP TABLE IF EXISTS "user"              CASCADE')
-            connection.execute('DROP TABLE IF EXISTS listen              CASCADE')
-            connection.execute('DROP TABLE IF EXISTS listen_json         CASCADE')
-            connection.execute('DROP TABLE IF EXISTS api_compat.token    CASCADE')
-            connection.execute('DROP TABLE IF EXISTS api_compat.session  CASCADE')
```
alastair: I learned recently that you can use `DROP SCHEMA CASCADE` to drop all tables in a schema. As we have mostly schemas here, perhaps we could just drop the `"user"` table and `statistics` and `api_compat` schemas.
Are we still using the `listen` and `listen_json` tables??

admin/sql/create_tables.sql
```diff
@@ -42,4 +42,45 @@ CREATE TABLE api_compat.session (
     ts        TIMESTAMP DEFAULT CURRENT_TIMESTAMP
 );
 ALTER TABLE api_compat.session ADD CONSTRAINT session_sid_uniq UNIQUE (sid);
+
+CREATE TABLE statistics.user (
+    user_id                 INTEGER NOT NULL, -- PK and FK to "user".id
+    artists                 JSONB,
```
alastair: Can you give an example of what these json fields will look like? How often will you read from them or write to them? Will you want to get a specific value from in the field (and therefore require an index)?

admin/sql/create_tables.sql
```diff
@@ -42,4 +42,45 @@ CREATE TABLE api_compat.session (
...
+    releases                JSONB,
+    recordings              JSONB,
+    last_updated            TIMESTAMP WITH TIME ZONE
+);
+ALTER TABLE statistics.user ADD CONSTRAINT user_stats_user_id_uniq UNIQUE (user_id);
```
alastair: We don't need a `UNIQUE` constraint if we also have a `PRIMARY KEY` (a primary key is basically a unique constraint + index)

admin/sql/create_tables.sql
paramsingh: The plan is to use these fields to store the statistics we calculated from BigQuery. For example, in the artists field of `statistics.user`, we store the top artists listened to by that user over different time intervals. The json would be something like this, I guess: 

```
{
    "all_time": 
    [
        {
            "name": "example artist 1",
            "msid": uuid,
            "listen_count": 87
        },
        {
            "name": "example artist 2",
            "msid": uuid,
            "listen_count": 80
        }
    ],
    "last_month": 
    [
        {
            "name": "example artist 2",
            "msid": uuid,
            "listen_count": 12
        },
        {
            "name": "example artist 1",
            "msid": uuid,
            "listen_count": 3
        }
    ]
}
```

These stats would then be used to draw the graphs on the site. I don't think we will need to query specific values inside them ever, and we'll only write to them during our regular stat calculation timings, I guess.

listenbrainz/db/testing.py
paramsingh: `listen` and `listen_json` aren't being used anymore but the PostgresListenStore tests depend upon them.

I'll update the PR with `DROP SCHEMA CASCADE` for the schemas and drop the remaining tables normally.

admin/sql/create_tables.sql
paramsingh: Ah, sorry about that, will fix.

paramsingh:
@alastair Used `CASCADE` in drop_schema.sql and removed the `DROP TABLE`s, also used SERIAL for primary keys of entity tables, so the unique constraints are needed now and I haven't removed them (except for the user table).

admin/sql/create_tables.sql
alastair: OK, this seems fine. I wasn't sure if it would be a better idea to split this data into separate fields.

It sounds like you're only ever going to write this field in bulk - that is, compute it all at once and then write it to the field, replacing whatever was there. Same for reading - you're always going to want to get all of the data at once in order to display it. If this is the case, then I'm fine with this way of storing the data.

PR #194, opened 2017-06-07
Remove broken stats

mayhem:
I tried to count the actual number of items in rabbitmq (as opposed to a number of batches, which may contain more than 1 listen) but the system is brittle and prone to showing the wrong results. It is unclear that we even need this, so I am removing it.

### DIFF
--- a/listenbrainz/redis_keys.py
+++ b/listenbrainz/redis_keys.py
@@ -1,4 +1,3 @@
 # This module should be used to keep all the of the keys used with redis
 
-INCOMING_QUEUE_SIZE_KEY = "lb.incoming_q_size"
-UNIQUE_QUEUE_SIZE_KEY = "lb.unique_q_size"
+# <crickets> LOL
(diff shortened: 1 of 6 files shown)


listenbrainz/redis_keys.py
```diff
@@ -1,4 +1,3 @@
 # This module should be used to keep all the of the keys used with redis
 
-INCOMING_QUEUE_SIZE_KEY = "lb.incoming_q_size"
-UNIQUE_QUEUE_SIZE_KEY = "lb.unique_q_size"
+# <crickets> LOL
```
paramsingh: I've created an issue for this: https://tickets.metabrainz.org/browse/LB-174

I will do it soon.

listenbrainz/webserver/views/index.py
```diff
@@ -82,8 +83,12 @@ def current_status():
     except (pika.exceptions.ConnectionClosed, AttributeError):
         pass
 
-    incoming_count = _redis.redis.get("lb.incoming_q_size") or 0
-    unique_count = _redis.redis.get("lb.unique_q_size") or 0
+    ls = InfluxListenStore({ 'REDIS_HOST' : config.REDIS_HOST,
```
paramsingh: Wouldn't using `influx_connection._influx` be better here?

paramsingh:
:+1:

PR #195, opened 2017-06-07
LB-173: Fix artist_mbids not being shown correctly on user page

paramsingh:
Also add a test for the from_influx method


listenbrainz/listen.py
```diff
@@ -1,6 +1,5 @@
 # coding=utf-8
 
-import ujson
```
paramsingh: Wasn't being used anywhere

PR #197, opened 2017-06-09
LB-171 / LB-172: Fix bugs in influx-writer

paramsingh:
LB-171: marc2k3's listens weren't being inserted because they contained a list in their additional_info field which led to parse errors (this seems to be a bug in influx because other non-string entities like uuids were inserted correctly). Converting all new fields in additional info to strings explicitly now fixes that. Also added tests for listens like this.

LB-172: While doing this, I found a problem that might be causing [LB-172](https://tickets.metabrainz.org/browse/LB-172). Influx-writer was dying because the debug message tried to serialize an array of Listen objects. However this would only trouble us if `DUMP_JSON_WITH_ERRORS` was `True` in production, and I'm not sure that it was.


mayhem:
DUMP_JSON_WITH_ERRORS is true while we are in beta.

paramsingh:
That was causing LB-172 then for sure.

PR #198, opened 2017-06-10
LB-175: Fix retry when submit request to api times out

paramsingh:
Passing data to submitListens() doesn't do anything as
we pop data from the submitQueue. Eventually, the queue becomes
empty leading to the error in LB-175.


paramsingh:
This is weird, the build for this commit passes on my account. See here: https://travis-ci.org/paramsingh/listenbrainz-server/builds/241565299

Maybe a restart of the build will fix it?

PR #199, opened 2017-06-17
LB-158 / LB-170: Fix listenstore code for users with special characters, add tests and upgrade influx

paramsingh:
So I spent the day looking through the influxdb-python source code. And the reason this was so hard to debug is a combination of random weird behaviour in influx with backslashes and some problems in our code.

Influxdb-python automatically escapes measurement names when we use write_points which can be seen [here](https://github.com/influxdata/influxdb-python/blob/master/influxdb/line_protocol.py#L125), so when we send a measurement with name `"para\m"` to write_points, it shows up in influx as `"para\\m"`. As a result, the line protocol only requires escaping of `\n` characters. On the other hand, the query language requires escaping of everything and since influx had replaced the `\` in the measurement name with `\\`, we need FOUR backslashes in the query to get the data.

So, I have changed the code so that it does this and added a bunch of comments explaining stuff as best I could. I expect more changes will be requested to make it clearer. Also, I have upgraded to influx 1.2.4 and influxdb-python 4.1.0 .

Also, there is some weird behaviour when we backslash quotes in influx. Influx SOMETIMES totally ignores backslashes present in measurement names, if they come before quotes. This behaviour is present in influx itself (not influxdb-python) and I am not sure if this is a bug or if I am misunderstanding something. For example:

Backslash away from quotes:

```
> insert "random\user"",val=1 hello=1
> show measurements
name: measurements
name
----
"random\user""

> select * from "\"random\\user\"\""
name: "random\user""
time                hello val
----                ----- ---
1497719311488498867 1     1
```

Backslash in front of a quote gets ignored

```
> insert "randomuser\"",val=1 hello=1
> show measurements
name: measurements
name
----
"random\user""
"randomuser""

> select * from "\"randomuser\"\""
name: "randomuser""
time                hello val
----                ----- ---
1497719359874380564 1     1
```
Stuff like this is hard to write code for, unless there is something that I am missing completely...


paramsingh:
Clicked `Create Pull Request` before it was ready. :sweat_smile:

listenbrainz/listen.py
```diff
@@ -196,7 +197,7 @@ def to_influx(self, measurement):
         # add the user generated keys present in additional info to fields
         for key, value in self.data['additional_info'].items():
             if key not in Listen.SUPPORTED_KEYS:
-                data['fields'][key] = value
+                data['fields'][key] = escape(value)
```
paramsingh: Did this because `additional_info` variables may have random newlines in them, which would lead to parse errors.

PR #200, opened 2017-06-19
move the redis keys to the correct module, fix a print issue

mayhem:
The set rate limit script was barfing on the prints (I think). While I was at it, I moved the redis keys to the right places.


paramsingh:
Looks good enough to me. +1

paramsingh:
Note: the build failed because of a setuptools issue, it's fixed in the latest version and I've upgraded to that in #199

PR #201, opened 2017-06-19
LB-177: Make influx-writer account for a range of time when looking for duplicates

paramsingh:
LB alpha has data with low precision timestamps, so if a person does both an alpha import and a last.fm import on LB Beta, the data gets duplicated. I've made influx-writer now consider a listen a duplicate if a listen of the same recording exists in a range of that timestamp.

Here's some examples of time differences I found for the same listen in last.fm and LB alpha.

```
listenbrainz	lastfm	        diff
-----------------------------------------
1463598180	1463598185	5
1463597940	1463597956	16
1463597640	1463597671	31
1463597400	1463597440	40
1463597280	1463597323	43
1463597040	1463597061	21
1463596740	1463596794	54
1463596380	1463596385	5
1463596020	1463596024	4
```

The time differences can go as high as 54, as we see. But if we increase the range of time we consider to be duplicates, it can lead to certain use cases not working. For example, if a person listens to a 15-second song consecutively for many times, it will get considered a duplicate here. Not sure if this is something we want to allow?

What is the way forward here? A reasonable compromise by adjusting the value of the time range in which we consider two listens to be duplicates?

### DIFF
--- a/listenbrainz/influx-writer/influx-writer.py
+++ b/listenbrainz/influx-writer/influx-writer.py
@@ -22,7 +22,7 @@ DUMP_JSON_WITH_ERRORS = False
 ERROR_RETRY_DELAY = 3 # number of seconds to wait until retrying an operation
 
 # the difference in timestamps which we consider to be duplicates if same artist msid and recording msid
-TIMESTAMP_DUPLICATE_DIFF = 21
+TIMESTAMP_DUPLICATE_DIFF = 31
 
 
 class InfluxWriterSubscriber(object):
@@ -206,21 +206,16 @@ class InfluxWriterSubscriber(object):
                 # This will check if timestamps in the range (t - TIMESTAMP_DUPLICATE_DIFF, t + TIMESTAMP_DUPLICATE_DIFF)
                 # exist in the database for the user, with the same artist msid and recording msid. If it does, we
                 # consider this listen to be a duplicate and skip it.
-                for delta in range(TIMESTAMP_DUPLICATE_DIFF):
+                deltas = [0] + [sign * val for val in range(1, TIMESTAMP_DUPLICATE_DIFF) for sign in (+1, -1)]
+                for delta in deltas:
                     fuzzed = t + delta
-                    if fuzzed in timestamps:
-                        if artist_msid == timestamps[fuzzed]['artist_msid'] and recording_msid == timestamps[fuzzed]['recording_msid']:
-                            duplicate_count += 1
-                            break
-
-                    fuzzed = t - delta
                     if fuzzed in timestamps:
                         if artist_msid == timestamps[fuzzed]['artist_msid'] and recording_msid == timestamps[fuzzed]['recording_msid']:
                             duplicate_count += 1
                             break
                 else:
                     unique_count += 1
-                    submit.append(Listen().from_json(listen))
+                    submit.append(Listen.from_json(listen))
                     unique.append(listen)
 
         t0 = time()
--- a/listenbrainz/influx-writer/influx-writer.py
+++ b/listenbrainz/influx-writer/influx-writer.py
@@ -21,6 +21,9 @@ REPORT_FREQUENCY = 5000
 DUMP_JSON_WITH_ERRORS = False
 ERROR_RETRY_DELAY = 3 # number of seconds to wait until retrying an operation
 
+# the difference in timestamps which we consider to be duplicates if same artist msid and recording msid
+TIMESTAMP_DUPLICATE_DIFF = 21
+
 
 class InfluxWriterSubscriber(object):
     def __init__(self):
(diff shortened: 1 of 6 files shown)


listenbrainz/influx-writer/influx-writer.py
```diff
@@ -191,18 +196,32 @@ def write(self, listen_dicts):
...
-                # Check to see if the timestamp is already in the DB
                 t = int(listen['listened_at'])
-                if t in timestamps:
-                    duplicate_count += 1
-                    continue
-
-                unique_count += 1
-                submit.append(Listen().from_json(listen))
-                unique.append(listen)
+                artist_msid    = listen['track_metadata']['additional_info']['artist_msid']
```
alastair: are items in `additional_info` optional?

listenbrainz/influx-writer/influx-writer.py
```diff
@@ -191,18 +196,32 @@ def write(self, listen_dicts):
...
+                            break
+
+                    fuzzed = t - delta
+                    if fuzzed in timestamps:
+                        if artist_msid == timestamps[fuzzed]['artist_msid'] and recording_msid == timestamps[fuzzed]['recording_msid']:
+                            duplicate_count += 1
+                            break
+                else:
+                    unique_count += 1
+                    submit.append(Listen().from_json(listen))
```
alastair: if `from_json` is a static method you don't need `()`, just `Listen.from_json`

listenbrainz/influx-writer/influx-writer.py
mayhem: Yes. Only recording_msid is absolutely required.

listenbrainz/influx-writer/influx-writer.py
```diff
@@ -191,18 +196,32 @@ def write(self, listen_dicts):
             timestamps = {}
             for result in results.get_points(measurement=get_measurement_name(user_name)):
                 dt = datetime.strptime(result['time'] , "%Y-%m-%dT%H:%M:%SZ")
```
alastair: do we do stuff with dates in this format from influx often? If so, it'd be nice to have this abstracted away somewhere

listenbrainz/influx-writer/influx-writer.py
```diff
@@ -169,11 +172,13 @@ def write(self, listen_dicts):
...
+            # get the range of time that we need to get from influx for
+            # deduplication of listens
+            min_time = users[user_name]['min_time'] - TIMESTAMP_DUPLICATE_DIFF
+            max_time = users[user_name]['max_time'] + TIMESTAMP_DUPLICATE_DIFF
 
             # quering for artist name here, since a field must be included in the query.
-            query = """SELECT time, artist_name
+            query = """SELECT time, artist_msid, recording_msid
                          FROM %s
                         WHERE time >= %d000000000
```
alastair: can we make this more understandable by having a function like `timestamp_to_influx_crappy_nanoseconds` which we call like `get_escaped_measurement_name`?

listenbrainz/influx-writer/influx-writer.py
```diff
@@ -191,18 +196,32 @@ def write(self, listen_dicts):
...
+                # exist in the database for the user, with the same artist msid and recording msid. If it does, we
+                # consider this listen to be a duplicate and skip it.
+                for delta in range(TIMESTAMP_DUPLICATE_DIFF):
+                    fuzzed = t + delta
+                    if fuzzed in timestamps:
+                        if artist_msid == timestamps[fuzzed]['artist_msid'] and recording_msid == timestamps[fuzzed]['recording_msid']:
+                            duplicate_count += 1
+                            break
+
+                    fuzzed = t - delta
```
alastair: when `delta` is 0, this block is the same as the above block, which is exactly duplicated.
How about creating a list of positive deltas instead:
```
deltas = [0, 1, -1, 2, -2, 3, ....]
for delta in deltas:
   .. code just once, instead of twice
```

creating `deltas` is left as an exercise to the reader

alastair: I do like the positive/negative sweep though, this means that we hit the 0 case first, instead of going from -limit -> +limit

listenbrainz/influx-writer/influx-writer.py
alastair: Ahh, this is an msid, not an mbid. So, it'll be set (because artist names are required)

listenbrainz/influx-writer/influx-writer.py
mayhem: Lets leave this to https://tickets.metabrainz.org/browse/LB-178

listenbrainz/influx-writer/influx-writer.py
mayhem: Correction: artist_msid is also required.

paramsingh: Both of them were required, that's why I chose them. :)

listenbrainz/influx-writer/influx-writer.py
paramsingh: > creating deltas is left as an exercise to the reader

`deltas = [0] + [sign * val for val in range(1, 31) for sign in (+1, -1)]`

No code duplication now :)

>I do like the positive/negative sweep though, this means that we hit the 0 case first, instead of going from -limit -> +limit

Exactly why I did that :D

listenbrainz/influx-writer/influx-writer.py
```diff
@@ -190,19 +195,27 @@ def write(self, listen_dicts):
...
-                unique_count += 1
-                submit.append(Listen().from_json(listen))
-                unique.append(listen)
+                artist_msid    = listen['track_metadata']['additional_info']['artist_msid']
+                recording_msid = listen['recording_msid']
+
+                # This will check if timestamps in the range (t - TIMESTAMP_DUPLICATE_DIFF, t + TIMESTAMP_DUPLICATE_DIFF)
+                # exist in the database for the user, with the same artist msid and recording msid. If it does, we
+                # consider this listen to be a duplicate and skip it.
+                deltas = [0] + [sign * val for val in range(1, TIMESTAMP_DUPLICATE_DIFF) for sign in (+1, -1)]
```
mayhem: https://ih1.redbubble.net/image.11645535.4703/flat,800x800,070,f.jpg

PR #202, opened 2017-06-21
LB-176: Create a stats module and add functions to run queries on Google BigQuery

paramsingh:
This PR adds a stats module that contains `run_query` - a function that takes queries in BigQuery Standard SQL and parameters and then runs them as a synchronous job. It doesn't really depend on #192 but I was waiting on that to get merged before opening this, but now this seemed like a significant amount of work and I thought it best to get some reviews first before continuing further. I've also added a function that uses `run_query` to get the top tracks for a user. Other statistics would be added in a similar way. 

Hopefully, we can ignore the schema changes in this PR for now. Once #192 gets merged, I'll rebase and remove those changes from this PR.

### DIFF
--- a/listenbrainz/stats/user.py
+++ b/listenbrainz/stats/user.py
@@ -30,7 +30,7 @@ def get_top_tracks(musicbrainz_id, time_interval=None):
 
     filter_str = ""
     if time_interval:
-        filter_str = "AND TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL {}".format(time_interval)
+        filter_str = "AND listened_at > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL {})".format(time_interval)
 
     # construct the query string
     query = """SELECT artist_name
--- a/listenbrainz/stats/user.py
+++ b/listenbrainz/stats/user.py
@@ -1,0 +1,65 @@
+"""
+This module contains functions that create jobs and retrieve stats about users
+from Google BigQuery
+"""
+import listenbrainz.config as config
+import json
+from listenbrainz import stats
+
+
+def get_top_tracks(musicbrainz_id, time_interval=None):
+    """ Get top tracks of user with given MusicBrainz ID over a particular time interval
+
+        Args:
+            musicbrainz_id (str): The MusicBrainz ID of the user
+            time_interval  (str): Interval in the BigQuery interval format which can be seen here:
+            https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#timestamp_sub
+
+        Returns:
+            sorted list of the following format
+            [
+                {
+                    "track_name" (str)
+                    "recording_msid" (uuid)
+                    "artist_name" (str)
+                    "artist_msid" (uuid)
+                    "listen_count" (int)
+                }
+            ]
+    """
+
+    filter_str = ""
+    if time_interval:
+        filter_str = "AND TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL {}".format(time_interval)
+
+    # construct the query string
+    query = """SELECT artist_name
+                    , artist_msid
+                    , recording_msid
+                    , track_name
+                    , COUNT(recording_msid) as listen_count
+                 FROM {dataset_id}.{table_id}
+                WHERE user_name = @username
+                {time_filter_clause}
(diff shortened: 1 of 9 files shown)


paramsingh:
Right now, `run_query` doesn't take non-parameterized queries, working on that and thinking about what the ideal way to run the stats calculation at uniform intervals of time is. I think adding jobs to `webserver/scheduler.py` would be okay? I'm thinking of making a file in the stats module that can define a function that makes queries and saves the results and then add it as a job to `webserver/scheduler.py`

paramsingh:
Rebased after merge of #192.

listenbrainz/stats/user.py
```diff
@@ -0,0 +1,176 @@
...
+            time_interval  (str): Interval in the BigQuery interval format
+                                  which can be seen here:
+                                  https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#timestamp_sub
+
+        Returns:
+            sorted list of the following format
+            [
+                {
+                    "track_name" (str)
+                    "recording_msid" (uuid)
```
mayhem: Should we be fetching MBIDs or MSIDs here? Both? Using MBIDs is useful for linking to MB.

listenbrainz/stats/user.py
```diff
@@ -0,0 +1,176 @@
...
+                    "artist_name" (str)
+                    "artist_msid" (uuid)
+                    "listen_count" (int)
+                }
+            ]
+    """
+
+    filter_clause = ""
+    if time_interval:
+        filter_clause = "AND listened_at > TIMESTAMP_SUB(CURRENT_TIMESTAMP(), INTERVAL {})".format(time_interval)
```
mayhem: The interval is not defined anywhere that I can see. Should it be inclusive of the boundaries? (I think so).

listenbrainz/stats/user.py
```diff
@@ -0,0 +1,176 @@
...
+                    , artist_msid
+                    , recording_msid
+                    , track_name
+                    , COUNT(recording_msid) as listen_count
+                 FROM {dataset_id}.{table_id}
+                WHERE user_name = @musicbrainz_id
+                {time_filter_clause}
+             GROUP BY recording_msid, track_name, artist_name, artist_msid
+             ORDER BY listen_count DESC
+            """.format(
```
mayhem: You'll want to LIMIT the data for each and every query in order to keep the amount of traffic low.

listenbrainz/stats/user.py
paramsingh: Both sounds good to me.

listenbrainz/stats/user.py
paramsingh: Interval is a parameter to be passed to the function. But yes, it should be inclusive.

listenbrainz/stats/user.py
paramsingh: What would the ideal value of the limit be? Would top 100 artists / songs be more than enough, I think it would be. Should this value be in the config file?

listenbrainz/stats/user.py
```diff
@@ -34,34 +34,38 @@ def get_top_recordings(musicbrainz_id, time_interval=None):
...
-                    "recording_msid" (uuid)
-                    "artist_name" (str)
-                    "artist_msid" (uuid)
-                    "listen_count" (int)
+                    'track_name' (str)
+                    'recording_msid' (uuid)
+                    'artist_name' (str)
+                    'artist_msid' (uuid)
+                    'artist_mbids' (string of comma-seperated uuids)
+                    'listen_count' (int)
```
mayhem: recording_mbid is missing above, but in the code below.

listenbrainz/stats/user.py
mayhem: 100 seems fine for now -- leave it in the module for now. Easy enough to change once we know what it is we really want.

listenbrainz/stats/user.py
```diff
@@ -136,6 +144,7 @@ def get_top_releases(musicbrainz_id, time_interval=None):
                     {
                         'artist_name' (str),
                         'artist_msid' (uuid),
+                        'artist_mbids' (string of comma seperated uuids),
                         'release_name' (str),
                         'release_msid' (uuid),
```
mayhem: release_mbid is not listed above, but is listed in the code.

PR #205, opened 2017-06-22
Upgrade pika_pool to latest version on github.

paramsingh:
Latest version on PyPI has a bug:
https://github.com/bninja/pika-pool/issues/6

It's been fixed in https://github.com/bninja/pika-pool/pull/11 ,
but no release, so use the version on github instead


alastair:
Do you need `-e`? - https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support

PR #206, opened 2017-06-23
LB-183: Revert to old deduplication in influx-writer

paramsingh:
As we discussed on IRC.


paramsingh:
tbh, that's how I felt after writing the cool dedup code and then having to revert it :persevere:

PR #207, opened 2017-06-24
LB-20 / LB-31 / LB-182: Incremental imports

paramsingh:
* Add a latest_import column to the user table
* Add function in `db` for updating this value
* Add view which can be used to get this value for users and update this value (with authorization the same as API auth)
* Make the alpha importer and last.fm importer get the latest_import value from the server before starting 
* Make the alpha importer and last.fm importer stop at the value they got above
* Make the two importers report the latest import back to the LB server at the end of the import

Had to remove the total number of pages being shown in the last.fm importer because we don't know the number of pages we'll have to get anymore.

### DIFF
--- a/listenbrainz/webserver/templates/user/scraper.js
+++ b/listenbrainz/webserver/templates/user/scraper.js
@@ -147,9 +147,24 @@ function getLastFMPage(page) {
     xhr.onload = function () {
         if (/^2/.test(this.status)) {
             if (numberOfPages <= 1) {
-              numberOfPages = JSON.parse(this.response)['recenttracks']['@attr']['totalPages'];
+                var data = JSON.parse(this.response);
+                numberOfPages = data['recenttracks']['@attr']['totalPages'];
+
+                // initially set the stop page to the total number of pages
+                // we'll later update it if we're able to stop earlier because of previous imports
+                stopPage = numberOfPages;
+
+                if (data['recenttracks']['track'].length > 0) {
+                    // if date is present, use it
+                    if ('date' in data['recenttracks']['track'][0]) {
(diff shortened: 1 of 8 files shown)


listenbrainz/webserver/views/user.py
```diff
@@ -329,3 +331,33 @@ def import_from_alpha():
...
+
+        try:
+            ts = ujson.loads(request.get_data()).get('ts', 0)
+        except ValueError:
+            raise BadRequest('Invalid data sent')
+
+        try:
+            db_user.update_latest_import(user['musicbrainz_id'], int(ts))
+        except DatabaseException as e:
+            raise InternalServerError('Could not update latest_import, try again')
```
mayhem: we should log whatever occurred here.

listenbrainz/webserver/templates/user/scraper.js
```diff
@@ -147,9 +147,24 @@ function getLastFMPage(page) {
...
+                // initially set the stop page to the total number of pages
+                // we'll later update it if we're able to stop earlier because of previous imports
+                stopPage = numberOfPages;
+
+                if (data['recenttracks']['track'].length > 0) {
+                    // if date is present, use it
+                    if ('date' in data['recenttracks']['track'][0]) {
+                        maximumTimestampForImport = data['recenttracks']['track'][0]['date']['uts'];
+                    }
+                    else { // the latest listen is a playing_now, use current time
```
mayhem: I wonder if it is best to ignore playing_now in this context and simply not change the maximumTimestampForImport?

paramsingh: Why? If we've imported the data already, then the variable should be changed, right?

PR #212, opened 2017-06-26
Add artist and recording MSID checks to dedup

paramsingh:
To allow rows with the same timestamp, we have to add `artist_msid` and `recording_msid` to the influx tags too.

### DIFF
--- a/listenbrainz/db/tests/test_lastfm_user.py
+++ b/listenbrainz/db/tests/test_lastfm_user.py
@@ -38,13 +38,6 @@ class TestAPICompatUserClass(DatabaseTestCase):
             row = result.fetchone()
             self.user = User(row['id'], row['created'], row['musicbrainz_id'], row['auth_token'])
 
-        # Insert some listens
-        date = datetime(2015, 9, 3, 0, 0, 0)
-        self.log.info("Inserting test data...")
-        test_data = generate_data(date, 100, self.user.name)
-        self.logstore.insert(test_data)
-        self.log.info("Test data inserted")
-
     def tearDown(self):
         super(TestAPICompatUserClass, self).tearDown()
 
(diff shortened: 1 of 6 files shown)


listenbrainz/db/tests/test_lastfm_user.py
```diff
@@ -63,5 +56,9 @@ def test_user_load_by_id(self):
         self.assertDictEqual(user.__dict__, self.user.__dict__)
 
     def test_user_get_play_count(self):
+        date = datetime(2015, 9, 3, 0, 0, 0)
```
mayhem: why did you move this block?

listenbrainz/listen.py
```diff
@@ -179,6 +179,8 @@ def to_influx(self, measurement):
             'time' : self.ts_since_epoch,
             'tags' : {
                 'user_name' : escape(self.user_name),
+                'artist_msid' : self.artist_msid,
```
mayhem: Why are you using these tags to carry these msids?

paramsingh: influx cannot store two rows with the same tags and time, so if we want tracks with the same timestamp but different metadata to be stored, we have to add `artist_msid` and `recording_msid` to tags as well. Otherwise the old listen will just get overwritten by the new one.

listenbrainz/db/tests/test_lastfm_user.py
paramsingh: The test had started failing for me randomly :(

I had added the reason for moving this block to the commit message https://github.com/metabrainz/listenbrainz-server/pull/212/commits/7056af289d1f21107c1b823d4cec27a17f6acdea

listenbrainz/listen.py
mayhem: Right. Sadly this will bloat our data for little gain, but I can't think of a much better way of doing this.

mayhem: Ok, finally coming back to this. Why do we need both artist_msid and recording_msid? Shouldn't just recording_msid be sufficient?

paramsingh: Hmm, you're right, `recording_msid` should suffice.

paramsingh: Question: do tags not get returned when we do `select count(*) from measurement`? Asking this because I removed them both from fields and added into tags, and the listen count test [here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/listenstore/tests/test_influxlistenstore.py#L197) started to fail because there wasn't a `count_recording_msid` field in the results of the query.

mayhem: I believe that is correct. 

So, given all of the above, all fields should be restored as they were and then only recording_msid added to tags. Then we should be ready to merge this -- this now makes sense to me.

PR #213, opened 2017-06-29
LB-185: Add release name to data submitted by last.fm importer

paramsingh:
Release names are present in the data that last.fm API sends but the importer drops them. They should be submitted to LB as well.

Also, added a license notice :policeman:

### DIFF
--- a/listenbrainz/webserver/templates/user/scraper.js
+++ b/listenbrainz/webserver/templates/user/scraper.js
@@ -1,3 +1,24 @@
+/*
+ * listenbrainz-server - Server for the ListenBrainz project.
+ *
+ * Copyright (C) 2017 MetaBrainz Foundation Inc.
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/templates/user/scraper.js
```diff
@@ -66,6 +87,14 @@ Scrobble.prototype.trackName = function () {
     }
 };
 
+Scrobble.prototype.releaseName = function() {
+    if ('album' in this.rootScrobbleElement && '#text' in this.rootScrobbleElement['album']) {
+        return this.rootScrobbleElement['album']['#text'];
+    } else {
+        return "";
```
alastair: it looks like we have as much of a problem with quote styles in js as we do in python. We should be a bit more careful in the future with consistency.

PR #214, opened 2017-07-03
LB-187: Add NOT NULL constraint to the last_login field

paramsingh:
Fix issues pointed out by @alastair in this review: https://github.com/metabrainz/listenbrainz-server/pull/189#pullrequestreview-45373057


admin/sql/updates/alter_last_login.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
```
alastair: in acousticbrainz we put the date in the filenames of update scripts, to make a it a little easier to know in what order to run things. Should we do this here too?

listenbrainz/db/tests/test_user.py
```diff
@@ -19,10 +19,10 @@ def test_update_token(self):
 
     def test_update_last_login(self):
         user = db_user.get_or_create('testlastloginuser')
-        val = int(time.time())
-        db_user.update_last_login(user['musicbrainz_id'], val)
+        db_user.update_last_login(user['musicbrainz_id'])
         user = db_user.get_by_mb_id(user['musicbrainz_id'])
-        self.assertEqual(val, int(user['last_login'].strftime('%s')))
+        # after update_last_login, the val should be equal to the current time
+        self.assertEqual(int(time.time()), int(user['last_login'].strftime('%s')))
```
alastair: this is dangerous, because 99% of the time it might work, but then suddenly there may be a case where we do the above line just before the time increases by a second.
I would prefer to create a user with a last login time obviously in the past - you could do this by doing an sql query to get `now() - interval 1 day` or choose some time in june 2017 - and then do the update and check that this value is greater than the original value that you set it to

admin/sql/updates/alter_last_login.sql
mayhem: Yes, please.

listenbrainz/db/tests/test_user.py
paramsingh: Hmm, I actually thought that this could be a problem later, should have done this from the beginning, done now!

admin/sql/updates/alter_last_login.sql
paramsingh: Files renamed.

listenbrainz/db/tests/test_user.py
```diff
@@ -18,11 +20,28 @@ def test_update_token(self):
...
+
+        user = db_user.get(user['id'])
+        self.assertEquals(int(user['last_login'].strftime('%s')), 0)
+
+        db_user.update_last_login(user['musicbrainz_id'])
         user = db_user.get_by_mb_id(user['musicbrainz_id'])
-        self.assertEqual(val, int(user['last_login'].strftime('%s')))
+
+        # after update_last_login, the val should be greater than the old value i.e 0
+        self.assertGreater(int(user['last_login'].strftime('%s')), 0)
```
mayhem: This test is a bit iffy to me. I think you should get the current time, save it, then ensure that the DB got updated to that time.

paramsingh: That would require changing the function to take a timestamp parameter and update to that timestamp, which is what alastair wanted changed here: https://github.com/metabrainz/listenbrainz-server/pull/189#discussion_r123195816

Initially I had written the test like you are suggesting [here](https://github.com/paramsingh/listenbrainz-server/blob/da5d86020c0fe8059c0a4a29459a56815b95391f/listenbrainz/db/tests/test_user.py#L20) but we changed it here with alastair's review: https://github.com/metabrainz/listenbrainz-server/pull/214#discussion_r125925710

Should I go back to taking a timestamp in the function and not using `NOW()`?

mayhem: Naw, lets just improve the test once #189 is implemented. But in general giving a timestamp makes more sense, because it allows the caller to determine what to do.

PR #216, opened 2017-07-05
Possible fix for LB-188

mayhem:
Not sure if this is a valid fix for this bug, but it might be. :)


paramsingh:
This solution could cause problems I guess, because if there's a batch containing all invalid listens, it would cause it get stuck in an infinite loop?

Hmm, I had fixed it by doing (was in the process of testing it)

```
if not stop:
    next_max = ts of last listen in list
```

PR #217, opened 2017-07-05
LB-188: Check if we need to stop before getting next timestamp

paramsingh:
Fix for LB-188. We were trying to get the last listen of an empty list at the end of an import. Also, only log the beginning and end of an import now. I'd add a test but alpha_importer is really hard to add tests for.  :slightly_frowning_face:

### DIFF
--- a/listenbrainz/alpha_importer.py
+++ b/listenbrainz/alpha_importer.py
@@ -8,7 +8,6 @@ import time
 import sys
 import logging
 from logging.handlers import RotatingFileHandler
-import os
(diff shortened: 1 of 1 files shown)


listenbrainz/alpha_importer.py
```diff
@@ -8,7 +8,6 @@
 import sys
 import logging
 from logging.handlers import RotatingFileHandler
-import os
```
paramsingh: an unused import

PR #218, opened 2017-07-06
LB-189: Raise 400 Bad Request when artist_name is not a string

paramsingh:
test added too


alastair:
I notice that the documentation at http://listenbrainz.readthedocs.io/en/latest/dev/json.html#payload-json-details doesn't say what the types of the three required fields should be.

PR #221, opened 2017-07-08
Remove retries from insert_to_listenstore in influx-writer

paramsingh:
The retries make influx-writer take too long for a batch of bad listens.


mayhem:
I've debated on this one a lot, but we don't have problems in production and I'm not convinced we need this. Passing for now.

PR #222, opened 2017-07-12
Do not put user name into the tags anymore since tags are optional now.

mayhem:
This should avoid the 100k tags limit for users -- the user name was in the tags only because tags were previously required. I've moved user_name to the fields -- not 100% sure we need it. Thoughts?


paramsingh:
Hmm, well I don't mind having explicit `user_name`s along with the implicit value in the measurement names really. I would be okay with keeping them, provided there are no major reasons to remove them.

mayhem:
That test failure appears to be unrelated to the changes.

PR #223, opened 2017-07-12
Remove recording_msid from tags and use a specific dedup tag 

paramsingh:
Had to make changes to a test in PostgresListenStore because it compared `listen.__dict__` which changed with the introduction of `dedup_tag`.


listenbrainz/listen.py
```diff
@@ -177,8 +180,8 @@ def to_influx(self, measurement):
         data = {
             'measurement' : measurement,
             'time' : self.ts_since_epoch,
-            'tags' : { 
-                'recording_msid' : self.recording_msid
+            'tags' : {
+                'dedup_tag' : self.dedup_tag
```
mayhem: I think it might be nicer to add this key if a dedup value is needed and omit it otherwise. This makes the data a little smaller.

paramsingh: okay

PR #224, opened 2017-07-13
Add message if number of listens is inconsistent in Last.fm

paramsingh:
Also, found a bug that could have caused us to lose a page or two during incremental imports.

### DIFF
--- a/listenbrainz/webserver/templates/user/scraper.js
+++ b/listenbrainz/webserver/templates/user/scraper.js
@@ -433,12 +433,12 @@ function updateLatestImportTimeOnLB() {
         if (this.status == 200) {
             var final_msg = "<i class='fa fa-check'></i> Import finished<br>";
             final_msg += "<span><a href={{ url_for('user.profile', user_name = user_name) }}>Close and go to your ListenBrainz profile</a></span><br>";
-            final_msg += "<span style='font-size:8pt'>Successfully submitted " + countReceived + " listens to ListenBrainz. Please note that some of these listens might be duplicates leading to a lower listen count on LB.";
+            final_msg += "<span style='font-size:8pt'>Successfully submitted " + countReceived + " listens to ListenBrainz."
+                + " Please note that some of these listens might be duplicates leading to a lower listen count on LB.</span></br>";
             if (playCount != -1 && countReceived != playCount) {
-                final_msg += " This number is different from the " + playCount +
-                    " that Last.fm reports due to an inconsistency in their API, sorry!</span><br>";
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/templates/user/scraper.js
```diff
@@ -252,7 +254,8 @@ function reportPageAndGetNext(response, page) {
     }
     var struct = encodeScrobbles(response);
     submitQueue.push(struct);
-    if (struct['payload'][struct['payload'].length - 1]['listened_at'] <= latestImportTime) {
+    countReceived += struct['payload'].length;
+    if (struct['payload'][struct['payload'].length - 1]['listened_at'] < latestImportTime) {
```
paramsingh: From commit message

```
Because we allow more than one listen with the
same timestamp now, we need to stop incremental
imports only if we encounter a listen older than the
latest imported one. Otherwise, some data can be lost,
for example if we have two pages of listens with the
same timestamp, the importer must not stop on the first
page
```

listenbrainz/webserver/templates/user/scraper.js
```diff
@@ -428,7 +431,16 @@ function updateLatestImportTimeOnLB() {
     xhr.timeout = 10 * 1000; // 10 seconds
     xhr.onload = function(content) {
         if (this.status == 200) {
-            updateMessage("<i class='fa fa-check'></i> Import finished<br><span><a href={{ url_for('user.profile', user_name = user_name) }}>Close and go to your ListenBrainz profile</a></span><br><span style='font-size:8pt'>Thank you for using ListenBrainz</span>");
+            var final_msg = "<i class='fa fa-check'></i> Import finished<br>";
+            final_msg += "<span><a href={{ url_for('user.profile', user_name = user_name) }}>Close and go to your ListenBrainz profile</a></span><br>";
+            final_msg += "<span style='font-size:8pt'>Successfully submitted " + countReceived + " listens to ListenBrainz. Please note that some of these listens might be duplicates leading to a lower listen count on LB.";
+            if (playCount != -1 && countReceived != playCount) {
+                final_msg += " This number is different from the " + playCount +
```
mayhem: It would be nice to make this in a bold font in red, so that people have a bit more of a chance to see it.

paramsingh: on it

paramsingh:
This is a screenshot of the message. Note that the numbers match because I removed the `if` condition to get the entire message, the message won't show if the numbers match in reality.

screenshot_2017-07-13_22-42-48

PR #225, opened 2017-07-13
Don't show api inconsistency message on incremental imports

paramsingh:
screenshot_2017-07-14_00-06-57


listenbrainz/webserver/templates/user/scraper.js
```diff
@@ -350,7 +350,7 @@ function submitListens() {
 
                 } else {
                     var msg = "<i class='fa fa-cog fa-spin'></i> Sending page " + numCompleted;
-                    if (showNumberOfPages) {
+                    if (notIncrementalImport) {
```
mayhem: Should notIncrementalImport be declared and initialized somewhere?

PR #226, opened 2017-07-14
Improve the BQ insert id and increase font size on scraper dialog message

mayhem:
Slightly conflated, but tiny overall.

### DIFF
--- a/listenbrainz/bigquery-writer/bigquery-writer.py
+++ b/listenbrainz/bigquery-writer/bigquery-writer.py
@@ -81,7 +81,7 @@ class BigQueryWriter(object):
             }
             bq_data.append({
                 'json': row,
-                'insertId': "%s-%s" % (listen['user_name'], listen['listened_at'])
+                'insertId': "%s-%s-%s" % (listen['user_name'], listen['listened_at'], listen['recording_msid'])
             })
 
         body = { 'rows' : bq_data }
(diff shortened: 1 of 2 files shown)


listenbrainz/bigquery-writer/bigquery-writer.py
```diff
@@ -81,7 +81,7 @@ def callback(self, ch, method, body):
             }
             bq_data.append({
                 'json': row,
-                'insertId': "%s-%s" % (listen['user_name'], listen['listened_at'])
+                'insertId': "%s-%s-%s" % (listen['user_name'], listen['listened_at'], listen['recording_msid'])
```
paramsingh: @alastair, because the `insertId` for two listens with different metadata and same ts was the same, only one of them would get written into BQ. A question is what would be the ideal way to write tests for stuff like this, so that this doesn't break again?

alastair: testing against external services is difficult. We could try and create some sort of 'fake bq' local service, but we would have to replicate behaviour that we don't understand in order to test this, which is almost impossible.

In this case, perhaps we could split the part that generates `bq_data` into a separate method and then have a test for it which checks that we have unique `insertId`s for all items.
However, I'm not sure how useful this check will be - it's not going to help us find a problem in bq until we actually observe the error happening there...

PR #228, opened 2017-07-14
Rename notIncrementalImport to incrementalImport

paramsingh:
There was an incorrect condition in the code and I realized
it was because the name `notIncrementalImport` gets confusing.
So renamed that variable to `incrementalImport` and added a
bunch of comments.

Also, I changed `Please don't navigate` to `Please don't close
this page`.

### DIFF
--- a/listenbrainz/webserver/templates/user/scraper.js
+++ b/listenbrainz/webserver/templates/user/scraper.js
@@ -245,7 +245,7 @@ var latestImportTime = 0;
 // the latest listen found in this import, we'll report back to the server with this
 var maximumTimestampForImport = 0;
 
-var notIncrementalImport = true;
+var incrementalImport = false;
 
 function reportPageAndGetNext(response, page) {
     timesGetPage++;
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/templates/user/scraper.js
```diff
@@ -435,11 +442,16 @@ function updateLatestImportTimeOnLB() {
             final_msg += "<span><a href={{ url_for('user.profile', user_name = user_name) }}>Close and go to your ListenBrainz profile</a></span><br>";
             final_msg += "<span style='font-size:8pt'>Successfully submitted " + countReceived + " listens to ListenBrainz."
                 + " Please note that some of these listens might be duplicates leading to a lower listen count on LB.</span></br>";
-            if (notIncrementalImport == false && playCount != -1 && countReceived != playCount) {
+
+            // if the count received is different from the api count, show a message accordingly
+            // also don't show this message if it's an incremental import, because countReceived
+            // and playCount will be different by definition in incremental imports
+            if (!incrementalImport && playCount != -1 && countReceived != playCount) {
```
mayhem: I think this line might be wrong since the logic was not negated.

paramsingh: The logic was wrong earlier here. We only need to show the inconsistency message if it is not an incremental import. That's why I changed the name of the variable, it confused me initially.

PR #233, opened 2017-07-18
LB-194: Fix error in importer due to spaces in usernames

paramsingh:
Fix for [LB-194](https://tickets.metabrainz.org/projects/LB/issues/LB-194?filter=allopenissues)

The value of scraper_url is already encoded because it is a
result of call to flask's url_for [here](https://github.com/metabrainz/listenbrainz-server/blob/1c31970ba549d7b053430794511ec9567c6c0a81/listenbrainz/webserver/views/user.py#L182). Encoding it again in the scraper causes
an already escaped user_name being passed to the scraper
which led to LB-194.

Also, now after this, we need to escape the user name because `param\singh` shows up as `paramsingh` when passed to the template.

### DIFF
--- a/listenbrainz/webserver/views/test/test_user.py
+++ b/listenbrainz/webserver/views/test/test_user.py
@@ -28,7 +28,7 @@ class UserViewsTestCase(ServerTestCase, DatabaseTestCase):
             }
         )
         self.assert200(response)
-        self.assertIn('var user_name = "{}";'.format(urllib.parse.quote_plus(self.user['musicbrainz_id'])), response.data.decode('utf-8'))
+        self.assertIn('var user_name = "iliekcomputers";', response.data.decode('utf-8'))
 
         response = self.client.get(
             url_for('user.lastfmscraper', user_name=self.weirduser['musicbrainz_id']),
@@ -38,7 +38,7 @@ class UserViewsTestCase(ServerTestCase, DatabaseTestCase):
             }
         )
         self.assert200(response)
-        self.assertIn('var user_name = "{}";'.format(urllib.parse.quote_plus(self.weirduser['musicbrainz_id'])), response.data.decode('utf-8'))
+        self.assertIn('var user_name = "weird%5Cuser%20name";', response.data.decode('utf-8'))
 
     def tearDown(self):
         ServerTestCase.tearDown(self)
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/user.py
```diff
@@ -46,7 +47,7 @@ def lastfmscraper(user_name):
         import_url="{}/1/latest-import".format(config.BETA_URL),
         user_token=user_token,
         lastfm_username=lastfm_username,
-        user_name=user_name,
+        user_name=escape(user_name), # need to escape user_name here because other wise jinja doesn't handle usernames with backslashes correctly
```
alastair: Can we have a few tests with this template to make sure it renders various usernames correctly?

paramsingh:
Added a test and decided to just use `urllib` to encode the username instead of escaping because it's only ever used in urls. Also, found a problem in the profile url generated by url_for because the username gets encoded twice on the call to url_for, so instead I sent a profile_url directly.

listenbrainz/webserver/views/test/test_user.py
```diff
@@ -11,11 +12,34 @@ def setUp(self):
...
+        """ Tests that the username is correctly rendered in the last.fm importer """
+        response = self.client.get(
+            url_for('user.lastfmscraper', user_name=self.user['musicbrainz_id']),
+            query_string={
+                'user_token': self.user['auth_token'],
+                'lastfm_username': 'dummy',
+            }
+        )
+        self.assert200(response)
+        self.assertIn('var user_name = "{}";'.format(urllib.parse.quote_plus(self.user['musicbrainz_id'])), response.data.decode('utf-8'))
```
alastair: It makes sense to use a plain string here as the first argument - `var user_name = "some\\\\name etc";` because we know exactly what it should be after escaping.

As this PR is about fixing users with spaces in their name, it'd be good to have this as a test too

alastair: ah, I just saw that the weird name has slashes and a space in it, never mind

mayhem: @paramsingh THe never mind comment applies to the the second sentence -- the plain string request still stands. Let's do that, then we can merge.

paramsingh: oh, okay. fixing.

PR #234, opened 2017-07-19
LB-191: Report errors on the API in json

paramsingh:
Added return valid json on success with api endpoints `submit-listens` and `latest-import`.

### DIFF
--- a/listenbrainz/webserver/errors.py
+++ b/listenbrainz/webserver/errors.py
@@ -1,4 +1,4 @@
-from flask import render_template, make_response
+from flask import render_template, make_response, jsonify, request
 from yattag import Doc
 import yattag
 import ujson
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/errors.py
```diff
@@ -57,41 +57,61 @@ def error_wrapper(template, error, code):
...
+        """ Returns appropriate error message on HTTP exceptions
+
+            error (werkzeug.exceptions.HTTPException): The exception that needs to be handled
+            code (int): the HTTP error code that should be returned
+
+            Returns:
+                A Response which will be a json error if request was made to the LB api and an html page
+                otherwise
+        """
+        if request.path.startswith('/1/'):
```
alastair: I'm wondering if we could future-proof this or at least get rid of some magic strings.
I'm not sure if there was a reason why our API blueprint is not mounted with a url_prefix: https://github.com/paramsingh/listenbrainz-server/blob/538c0a5ac12dfb72c7a6cfaad6f8cda9b0947962/listenbrainz/webserver/__init__.py#L109
If there was, we could have a const in settings which we can use here.

listenbrainz/webserver/errors.py
```diff
@@ -57,41 +57,61 @@ def error_wrapper(template, error, code):
...
+            return json_error_wrapper(error, code)
+        else:
+            return error_wrapper(f'errors/{code}.html', error, code)
```
alastair: this is going to take a while to get used to...

paramsingh: It looks a bit better `%s` and `.format` to me, and since CB has decided to move to using f-strings, I thought slowly moving to them here would be good.

PR #240, opened 2017-07-31
Fix user export bug caused due to listens with timestamp 0

paramsingh:
listens with timestamp 0 caused `to_ts` to be 0 leading to a bug.

### DIFF
--- a/listenbrainz/testdata/user_export_test.json
+++ b/listenbrainz/testdata/user_export_test.json
@@ -1,0 +1,29 @@
+{
+    "listen_type": "import",
+    "payload": [
+        {
+            "track_metadata": {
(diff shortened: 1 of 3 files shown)


paramsingh:
:india:

listenbrainz/testdata/user_export_test.json
```diff
@@ -0,0 +1,29 @@
...
+            },
+            "listened_at": 1486466992
+        },
+        {
+            "track_metadata": {
+                "track_name": "Sister",
+                "artist_name": "The Black Keys",
+                "release_name": "El Camino"
+            },
+            "listened_at": 0
```
alastair: great that we have a test for this!

PR #241, opened 2017-07-31
LB-197: Allow users to reset latest import timestamp

paramsingh:
Works similar to how users reset tokens.

### DIFF
--- a/listenbrainz/webserver/templates/user/import.html
+++ b/listenbrainz/webserver/templates/user/import.html
@@ -30,6 +30,13 @@
         </form>
       </div>
 
+    <p>If you want to reset your previous import timestamp, click below</p>
+    <p>
+      <span class="btn btn-info btn-lg" style="width: 300px"><a href="{{ url_for("user.reset_latest_import_timestamp") }}" style="color: white;">Reset Import Timestamp</a></span>
+    </p>
+
+
+
     <h3> Import from ListenBrainz Alpha </h3>
       <p>
         <strong>Most users should not use this option! Please read this section carefully to ensure that this is the right option for you!</strong>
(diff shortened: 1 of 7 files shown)


paramsingh:
On import page:

screenshot_2017-07-31_23-37-20

Moves to a confirmation page:

screenshot_2017-07-31_23-37-42

After reset:
screenshot_2017-07-31_23-38-10

listenbrainz/webserver/templates/user/import.html
```diff
@@ -30,6 +30,13 @@ <h2 class="page-title">Import to user {{ user.musicbrainz_id }}</h2>
         </form>
       </div>
 
+    <p>If you want to reset your previous import timestamp, click below</p>
```
mayhem: I think we need more verbiage here. "If you think that a partial import has somehow missed some listens, you may reset your previous import timestamp. This will cause your next import to be a complete import which should add any missing listens while still avoiding adding duplicates to your history."

mayhem: Or something like that.

listenbrainz/webserver/templates/user/resetlatestimportts.html
```diff
@@ -0,0 +1,25 @@
+{% extends 'base.html' %}
+
+{% block title %}Reset import timestamp- ListenBrainz{% endblock %}
+
+{% block content %}
+  <h3 class="page-title">Reset Import Timestamp</h3>
+  <p>Are you sure you want to reset your import timestamp? </p>
+
+  <p>
+     Our last.fm importer keeps track of the listens you've imported and then only imports
```
mayhem: This text is good! I think we need to bring some of this back to the page where we show the button in the first palace.

mayhem:
As agreed in IRC, the verbiage changes will come in the next PR.

PR #242, opened 2017-08-01
LB-114 / LB-135: Add a new user info view

paramsingh:
Move user token, token reset and import timestamp reset
from import page to this view. Mostly inspired from CB.


paramsingh:
screenshot_2017-08-01_21-18-45

PR #247, opened 2017-09-03
LB-201: Make column names in statistics tables singular

paramsingh:


admin/sql/create_tables.sql
```diff
@@ -46,19 +46,19 @@ ALTER TABLE api_compat.session ADD CONSTRAINT session_sid_uniq UNIQUE (sid);
...
 CREATE TABLE statistics.artist (
     id                      SERIAL, -- PK
     msid                    UUID NOT NULL,
     name                    VARCHAR,
-    releases                JSONB,
-    recordings              JSONB,
-    users                   JSONB,
+    release                 JSONB,
+    recording               JSONB,
+    "user"                  JSONB,
```
alastair: this is a bit annoying - I'm in favour of singular names, but user is a reserved keyword in postgres. Is there some other name we could give to this?

paramsingh: What do you suggest? `listener`?

mayhem: listener sounds ok to me.

mayhem:
Merged via command line.

PR #248, opened 2017-09-04
LB-203: Add support for v1.2 of the AudioScrobbler API

paramsingh:
Auth works via LB auth token, we create a new Flask app for
this API because it needs views mounted at /, this new app
is run in a new docker container in development. Also, session
information is stored in the api_compat.session table, so
there are some modifications in the lastfm_session code too.

Also, added tests for the new endpoints added.

### DIFF
--- a/listenbrainz/config.py.sample
+++ b/listenbrainz/config.py.sample
@@ -91,3 +91,6 @@ BETA_URL = 'https://api.listenbrainz.org'
 LISTENS_PER_GET = 100
 IMPORTER_LOG_FILE = '-'
 IMPORTER_DELAY = 3 # delay if requests don't return http 200
+
+
+LASTFM_PROXY_URL = 'http://0.0.0.0:8080/'
--- a/listenbrainz/webserver/templates/api_compat/index.html
+++ b/listenbrainz/webserver/templates/api_compat/index.html
@@ -1,5 +1,0 @@
-<html>
-  <head> <title> LB API Compat Index Page </title> </head>
-
-  <body> Please visit https://listenbrainz.org </body>
-</html>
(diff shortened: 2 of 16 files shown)


paramsingh:
This works with VLC in development for me, VLC has an option for providing API URL, so if we ran the `api_compat_app` on something like `proxy.listenbrainz.org` and asked people to change their URLs there and use the auth tokens as passwords, it would submit data easily. And Spotify would work with `/etc/hosts` modifications, I guess.

listenbrainz/config.py.sample
```diff
@@ -91,3 +91,6 @@ BETA_URL = 'https://api.listenbrainz.org'
 LISTENS_PER_GET = 100
 IMPORTER_LOG_FILE = '-'
 IMPORTER_DELAY = 3 # delay if requests don't return http 200
+
+
+LASTFM_PROXY_URL = 'http://0.0.0.0:8080/'
```
mayhem: What is this used for?

listenbrainz/webserver/templates/api_compat/index.html
```diff
@@ -0,0 +1,5 @@
+<html>
+  <head> <title> LB API Compat Index Page </title> </head>
+
+  <body> Please visit https://listenbrainz.org </body>
```
mayhem: We should have a page with documentation on how to use this (and the other) API. Let's create such a page and then redirect from here to that page.

listenbrainz/config.py.sample
paramsingh: When handshaking, the API returns a URL to be used by clients in the future, so we'd put `proxy.listenbrainz.org` in the config variable here in production and then clients would directly make requests to `proxy.listenbrainz.org`.

See this: https://github.com/metabrainz/listenbrainz-server/pull/248/files#diff-a3c56df132dfb50f5127034a2785c5bfR69

listenbrainz/webserver/templates/api_compat/index.html
paramsingh: Sure, makes sense. I kept this minimal because https://post.audioscrobbler.com/ was pretty small, I'lll change it now.

PR #249, opened 2017-09-09
LB-205: Add new view for now playing notifications

paramsingh:
Also change the response sent to the client when handshaking to
handle VLC problems (hopefully), it needed a newline at the end of the response too, I think.

### DIFF
--- a/listenbrainz/webserver/views/api_compat_deprecated.py
+++ b/listenbrainz/webserver/views/api_compat_deprecated.py
@@ -29,8 +29,9 @@ from hashlib import md5
 from time import time
 
 from flask import Blueprint, request, render_template, redirect
+from werkzeug.exceptions import BadRequest
 from listenbrainz.db.lastfm_session import Session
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/api_compat_deprecated.py
```diff
@@ -67,42 +68,63 @@ def handshake():
         'OK',
         session.sid,
         '{}/np_1.2'.format(config.LASTFM_PROXY_URL),
-        '{}/protocol_1.2'.format(config.LASTFM_PROXY_URL)
+        '{}/protocol_1.2\n'.format(config.LASTFM_PROXY_URL)
```
paramsingh: I read VLC's audioscrobbler plugin source code ([here](https://github.com/videolan/vlc/blob/master/modules/misc/audioscrobbler.c#L626)) and it seems that they require `\n` at the end of each line, I think this might be the reason why vlc wasn't able to send submissions to production even though it worked locally for me. (not sure though)

PR #250, opened 2017-09-12
Make source optional in api_compat_deprecated

paramsingh:
Also, add some logging to help with debugging future problems

### DIFF
--- a/listenbrainz/webserver/views/api_compat_deprecated.py
+++ b/listenbrainz/webserver/views/api_compat_deprecated.py
@@ -22,13 +22,14 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
+import json
 import listenbrainz.config as config
 import listenbrainz.db.user as db_user
 
 from hashlib import md5
 from time import time
 
-from flask import Blueprint, request, render_template, redirect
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/api_compat_deprecated.py
```diff
@@ -143,7 +147,7 @@ def _to_native_api(data, append_key):
                 'track_name': data['t{}'.format(append_key)],
                 'release_name': data['b{}'.format(append_key)],
                 'additional_info': {
-                    'source': data['o{}'.format(append_key)]
+                    'source': data.get('o{}'.format(append_key), '')
```
mayhem: This may add an empty key at times, which I would normally object to, but we don't actually store this data for a now listening request, right?

paramsingh: We don't, but I'll remove it anyways.

listenbrainz/webserver/views/api_compat_deprecated.py
```diff
@@ -134,6 +138,8 @@ def _to_native_api(data, append_key):
                     'track_name': (str)
                 }
             }
+
+        #TODO(param): add tests for this function
```
paramsingh: [LB-208](https://tickets.metabrainz.org/browse/LB-208) - can't right now, have a mid semester to study for, but I'll do this soon.

PR #253, opened 2017-10-01
Style cleanup db

alastair:
This PR cleans up formatting of the listenbrainz.db module according to pep8. I'm submitting each module as a separate PR as I don't want to submit a huge patch that would have to be reverted if we find an issue in the changes.
I used pycharm to rearrange the imports and to find items to change, but skipped a few that would require non-trivial changes.
I have tested unittests and everything still seems to work fine.


alastair:
has an extra commit in it that I didn't want to add; will resubmit

alastair:
back open again; force pushed with proper commits

PR #257, opened 2017-10-02
Improve development environment configuration, support SELinux

jflory7:
This commit makes three changes:

* Supports enforcing SELinux on the host computer by adding the `:z` annotation to the volume mount
* Actually starts the local server on port 8000 on the host system (before, was running on port 80, contradicting the README)
* Create `docker/data` instead of `docker/docker/data`, fixes accidental recursion and plays nicely with the .gitignore

I also added some more paths to ignore in the .gitignore, for Python virtualenvs. Ideally, most people won't have to use them, but I had to when I was playing around to get things working, so I went ahead
and added them.

This should help make it easier for first-time contributors to get started. With the current changes, I'm able to successfully create a development environment on first clone, where I had not been able
to before. Whooo!

**Edit**: I wrote a draft blog post on how to set up a development environment – a temporary preview link is [here](https://blog.justinwflory.com/?p=758&preview=1&_ppp=727602ac6d "How to set up a ListenBrainz development environment")!

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -12,19 +12,21 @@ services:
   db:
     image: postgres:9.5.3
     volumes:
-      - ./dbvolume:/var/lib/postgresql/data/
+      - ./dbvolume:/var/lib/postgresql/data:z
     ports:
       - "15432:5432"
 
   redis:
     image: redis:3.2.1
     volumes:
-      - ./docker/data/redis:/data
+      - ./data/redis:/data:z
+    ports:
+      - "6379:6379"
 
   influx:
     image: influxdb:1.2.4
     volumes:
-      - ./docker/data/influxdb:/var/lib/influxdb
+      - ./data/influxdb:/var/lib/influxdb:z
     environment:
       INFLUXDB_REPORTING_DISABLED: 'true'
       INFLUXDB_META_LOGGING_ENABLED: 'false'
@@ -41,7 +43,7 @@ services:
   rabbitmq:
     image: rabbitmq:3.6.5
     volumes:
-      - ./docker/data/rabbitmq:/var/lib/rabbitmq
+      - ./data/rabbitmq:/var/lib/rabbitmq:z
     ports:
       - "5672:5672"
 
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -52,12 +54,12 @@ services:
...
     volumes:
-      - ..:/code/listenbrainz
-      - ../credentials:/code/credentials
+      - ..:/code/listenbrainz:z
+      - ../credentials:/code/credentials:z
     environment:
       GOOGLE_APPLICATION_CREDENTIALS: '/code/credentials/bigquery-credentials.json'
     ports:
-      - "80:80"
+      - "8000:80"
```
paramsingh: I think we should keep this at 80 and change the docs to 80 from 8000. Other MeB projects use 80 too and it is a good idea to keep as many similarities between projects as possible.

jflory7: @paramsingh What if you're running multiple projects in your environment? For the host machine, I thought port 8000 made more sense since it's an [unprivileged port](https://www.w3.org/Daemon/User/Installation/PrivilegedPorts.html), and you could use a web server (e.g. `nginx`) to do a rewrite rule for port 8000. In this way, port 8000 made more sense for both a development environment and running in production.

If you'd like, I can still change everything to port 80. But from my point of view, port 8000 makes more sense since port 80 is likely to conflict with other services / applications.

paramsingh: You're right, 8000 does make more sense, I guess. I'm just hesitant because of other MeB projects using 80 (and a little, because my setup uses 80 :P). I'd be okay with 80 or 8000, but I guess we should make the port we choose here the standard for other MeB projects too.

paramsingh: @mayhem, any opinions on this? :P

mayhem: I'd prefer consistency with other projects for now -- I agree with the reservations, but until we finally get some "best metabrainz docker practices" defined, lets go with 80.

Regardless of what we pick, people will complain about it. :)

PR #258, opened 2017-10-02
Move private pages for users to /profile

alastair:
We currently have /user/<username> as a namespace for user profiles. We
had a number of subpages of /user which contained tool for users (e.g.
import, reset timestamps, export). This would cause a problem if a user
registered a musicbrainz user with these names and tried to log into
listenbrainz. To fix this, move all user-related pages to /profile.

This PR is based against the `production` branch for quick release. Tests pass and I navigated around the website to make sure that all of the pages work. I was unable to test import from alpha and import last.fm dump (but I'm not sure if this is linked from anywhere)


alastair:
It looks like I didn't get the integration tests quite right, I'll look at them again

PR #262, opened 2017-10-11
LB-217: Handle HTTPErrors in a better way in alpha importer

paramsingh:
We catch most errors but somehow missed this one. Also, some style fixes along the way.

### DIFF
--- a/listenbrainz/alpha_importer.py
+++ b/listenbrainz/alpha_importer.py
@@ -2,7 +2,7 @@
 from redis import Redis
 from listenbrainz import config
 import requests
-from requests.exceptions import HTTPError
+from requests.exceptions import ConnectionError, HTTPError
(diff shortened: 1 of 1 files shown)


listenbrainz/alpha_importer.py
```diff
@@ -153,16 +155,22 @@ def update_latest_import_time(token, ts):
...
 
 def import_from_alpha(username, token):
     """ Function to import listens from alpha for user with given username """
 
-    logger.info('Beginning alpha import for %s' % username)
+    logger.info('Beginning alpha import for %s', username)
 
     # first get the timestamp until which previous imports have already added data
-    latest_import_ts = get_latest_import_time(username)
+    while True:
```
paramsingh: This part is the main change made.

PR #263, opened 2017-10-11
LB-218: Stop alpha import when we get an auth error on beta

paramsingh:
Add an exception, raise it when the error we're talking about happens, catch it and stop the import.

### DIFF
--- a/listenbrainz/alpha_importer.py
+++ b/listenbrainz/alpha_importer.py
@@ -78,6 +78,10 @@ def extract_data(batch):
     return data
 
 
+class BetaAuthorizationException(Exception):
+    pass
+
+
 def send_batch(data, token, retries=5):
     """
     Sends a batch of data. If fails, then breaks the data into two parts and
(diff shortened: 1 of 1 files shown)


listenbrainz/alpha_importer.py
```diff
@@ -202,6 +209,11 @@ def import_from_alpha(username, token):
         except HTTPError as e:
             time.sleep(config.IMPORTER_DELAY)
 
+        except BetaAuthorizationException:
+            logger.error('Invalid authorization token for user %s', username)
+            logger.info('Stopping alpha import for user %s', username)
```
mayhem: Does this effectively remove the user from the alpha queue?

paramsingh: yep, after this it returns back to the main loop. https://github.com/metabrainz/listenbrainz-server/pull/263/files#diff-1adf6645434862a9aa0554be577a691aR215

PR #267, opened 2017-10-12
LB-215: Remove scheduler from webserver and use cron in a different container.

paramsingh:
This means that scheduled jobs should now be independent
of the webserver. Also fixes docs builds for me in case of
no config file.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -53,11 +53,6 @@ services:
       dockerfile: Dockerfile
     command: python3 /code/listenbrainz/manage.py runserver -h 0.0.0.0 -p 80 -d
     image: web
-    volumes:
-      - ..:/code/listenbrainz:z
-      - ../credentials:/code/credentials:z
-    environment:
-      GOOGLE_APPLICATION_CREDENTIALS: '/code/credentials/bigquery-credentials.json'
     ports:
       - "80:80"
     depends_on:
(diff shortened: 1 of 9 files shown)


docker/docker-compose.yml
```diff
@@ -108,3 +103,15 @@ services:
     depends_on:
       - web
       - rabbitmq
+
+  scheduler:
+    image: web
+    command: cron && tail -f /var/log/cron.log
```
mayhem: That's clever! I was wondering how logging was going to work!

listenbrainz/bigquery.py
```diff
@@ -1,10 +1,16 @@
...
 import os
+
 from googleapiclient import discovery
-import googleapiclient
 from oauth2client.client import GoogleCredentials
 
 APP_CREDENTIALS_FILE = os.environ.get('GOOGLE_APPLICATION_CREDENTIALS')
 
+
+logger = logging.getLogger(__name__)
```
mayhem: You added logging to this module, but it is never used. What's your plan here?

listenbrainz/stats/calculate.py
```diff
@@ -24,12 +27,26 @@ def calculate_user_stats():
...
+cli = click.Group()
+
+@cli.command()
+def calculate():
+    """ Command to calculate statistics from Google BigQuery.
+    This can be used from the manage.py file.
+    """
+
+    if not config.WRITE_TO_BIGQUERY:
+        time.sleep(10000)
```
mayhem: This should be in a loop, so it doesn't try to actually use BQ after 10000 seconds....

listenbrainz/bigquery.py
paramsingh: It was used in this file but I had forgotten to create the logger itself :( https://github.com/paramsingh/listenbrainz-server/blob/a1011edd21a5f330bd6617482baeb91664a39257/listenbrainz/bigquery.py#L18

PR #268, opened 2017-10-14
LB-221: Only calculate stats of users who haven't been done recently

paramsingh:
This means that if someone's stats get calculated then the script
needs to start up again, it won't calculate their stats again.

This is part of the better error handling for stats calculation I'm working on.

### DIFF
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -232,16 +232,19 @@ def reset_latest_import(musicbrainz_id):
     update_latest_import(musicbrainz_id, 0)
 
 
-def get_recently_logged_in_users():
-    """Returns a list of users who have logged-in in the
-    last config.STATS_CALCULATION_LOGIN_TIME days
-    """
+def get_users_with_uncalculated_stats():
+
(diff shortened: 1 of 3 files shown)


listenbrainz/db/user.py
```diff
@@ -232,16 +232,19 @@ def reset_latest_import(musicbrainz_id):
...
-                SELECT {columns}
-                  FROM "user"
-                 WHERE last_login >= NOW() - INTERVAL ':x days'
+                SELECT public."user".musicbrainz_id
+                  FROM public."user"
+             LEFT JOIN statistics.user
+                    ON public."user".id = statistics.user.user_id
+                 WHERE public."user".last_login >= NOW() - INTERVAL ':x days'
+                   AND (statistics.user.last_updated IS NULL OR statistics.user.last_updated < NOW() - INTERVAL ':y days')
                 """.format(columns=','.join(USER_GET_COLUMNS))), {
```
mayhem: columns is not longer used, right?

paramsingh:
Addressed the review and fixed the query a bit using shorter table names.

PR #269, opened 2017-10-14
LB-214: Add better exception handling & more logging to the stats calculation

paramsingh:
Also report some info like number of users done and
number of users for whom the process failed etc.

Based on #268 so will need to be merged after that.


mayhem:
I'll review this once #268 is merged.

paramsingh:
Rebased.

PR #272, opened 2017-10-15
LB-224: Add support for dumping PostgreSQL tables

paramsingh:
This commit adds support for dumping our postgres tables.
Dumps are of two types for now:

* private dumps: contains the user table, the api compat sessions
and tokens
* stat dumps: contains the stats tables

TODO: add tests, write code that imports data from dumps, add support for dumping listens from influx

This PR depends on #270, so will need to rebase after that gets merged.

### DIFF
--- a/listenbrainz/db/dump.py
+++ b/listenbrainz/db/dump.py
@@ -1,0 +1,283 @@
+""" This module contains data dump creation and import functions.
+"""
+
+# listenbrainz-server - Server for the ListenBrainz project
+#
+# Copyright (C) 2017 MetaBrainz Foundation Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+
+import listenbrainz.config as config
+import listenbrainz.db as db
+import logging
+import os
+import shutil
+import sqlalchemy
+import subprocess
+import sys
+import tarfile
+import tempfile
+
+from datetime import datetime
+from listenbrainz.utils import create_path
+
+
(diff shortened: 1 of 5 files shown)


paramsingh:
How would we want the notifications if diskspace runs out to work?

mayhem:
I think we need to add Sentry support to LB to alert us of these errors. But, that is beyond the scope of this right now, so I would create a function "notify_sysadmin_monkeys_that_shit_happened" (or something like it) and have it just do more logging for now. In a follow up PR we should add Sentry support.

listenbrainz/db/dump.py
```diff
@@ -0,0 +1,323 @@
...
+        return
+    logger.info('Dump of stats data created at %s!', stats_dump)
+
+
+    logger.info('Creating a new entry in the data_dump table...')
+    try:
+        dump_id = add_dump_entry()
+    except Exception as e:
+        logger.info('Error while adding dump entry: %s', str(e))
+        return
```
paramsingh: Not sure if we should remove the created dump files if adding a new dump entry fails and start again, or if we should loop until the new dump entry gets added. I'm leaning towards looping until new dump entry gets added, because if we error here, most of the heavy work of the dumps has been done and shouldn't be wasted.

paramsingh:
Added [LB-237: Add Sentry support to LB](https://tickets.metabrainz.org/browse/LB-237).

listenbrainz/db/dump.py
```diff
@@ -0,0 +1,323 @@
...
+                        arcname=os.path.join(archive_name, "COPYING"))
+            except IOError as e:
+                logger.error('IOError while adding dump metadata...')
+                raise
+            except Exception as e:
+                logger.error('Exception while adding dump metadata: %s', str(e))
+                raise
+
+
+            archive_tables_dir = os.path.join(temp_dir, 'lb-{}-dump'.format(dump_type), 'lb{}dump'.format(dump_type))
```
mayhem: I think this dir should just always be called lbdump.

paramsingh:
@mayhem do you want to take a look at the last two commits or should I merge and rebase the next PR?

listenbrainz/db/dump.py
```diff
@@ -173,11 +173,12 @@ def dump_postgres_db(location, threads=None):
...
-        dump_id = add_dump_entry()
-    except Exception as e:
-        logger.info('Error while adding dump entry: %s', str(e))
-        return
+    while True:
+        try:
+            dump_id = add_dump_entry()
+            break
+        except Exception as e:
+            logger.error('Error while adding dump entry: %s', str(e))
```
mayhem: There should be a 2 second delay -- we don't want to spin tightly. Add this later, please. I'll merge to keep moving for now.

PR #274, opened 2017-10-17
LB-225: Add support for importing data dumps.

paramsingh:
Based off #272, so will need to be rebased after that.

There are a bunch of things more to do here, for example the stats table has a foreign key to the user table so the stats tables can't be imported alone, it needs a corresponding entry in the user table. Not sure how to solve this, other than sanitizing users and putting it in the stats dump too.

### DIFF
--- a/listenbrainz/db/dump.py
+++ b/listenbrainz/db/dump.py
@@ -30,6 +30,7 @@ import subprocess
 import sys
 import tarfile
 import tempfile
+import time
 
 from datetime import datetime
 from listenbrainz.utils import create_path, log_ioerrors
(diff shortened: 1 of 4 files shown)


paramsingh:
Added tests too.

mayhem:
Rebase, please

paramsingh:
I would love opinions on the problem I mentioned in the OP. The stats dump currently cannot be imported alone, it needs a corresponding private dump.

>the stats table has a foreign key to the user table so the stats tables can't be imported alone, it needs a corresponding entry in the user table. Not sure how to solve this, other than sanitizing users and putting it in the stats dump too.

mayhem:
Oh sorry, didn't see that question:

> Not sure how to solve this, other than sanitizing users and putting it in the stats dump too.

That is exactly how MB does it -- lets do that too.

paramsingh:
[LB-240](https://tickets.metabrainz.org/browse/LB-240) and #293

paramsingh:
Rebased over #293 and added logic to only import user table from stats dump when needed.

PR #277, opened 2017-10-21
LB-227: Dump listens from InfluxListenStore in JSON format

paramsingh:
Create a JSON file for each user and add it to the archive.

### DIFF
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -20,7 +20,7 @@ from listenbrainz.listenstore import ORDER_ASC, ORDER_TEXT, \
     USER_CACHE_TIME, REDIS_USER_TIMESTAMPS
 from listenbrainz.utils import quote, get_escaped_measurement_name, get_measurement_name, get_influx_query_timestamp, \
     convert_influx_nano_to_python_time, convert_python_time_to_nano_int, convert_to_unix_timestamp, \
-    create_path
+    create_path, log_ioerrors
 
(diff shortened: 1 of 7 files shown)


listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -384,20 +386,40 @@ def dump_listens(self, location, dump_time=datetime.today(), threads=None):
...
-                    # loop until we get this user's listens
+                    # Get this user's listens in chunks
                     while True:
-                        try:
-                            results = self.influx.query(query)
+
+                        # loop until we get this chunk of listens
+                        while True:
+                            try:
+                                result = self.influx.query("""
```
mayhem: Given that this dump is about to listens up to, but not beyond the dump timestamp, this query should be limited to all listens up to the timestamp.

listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -384,20 +386,40 @@ def dump_listens(self, location, dump_time=datetime.today(), threads=None):
...
+                        if not rows:
                             break
-                        except Exception as e:
-                            self.log.error('Error while getting listens for user %s', user['musicbrainz_id'])
-                            self.log.error(str(e))
 
-                    listens = []
-                    for row in results.get_points(get_measurement_name(user['musicbrainz_id'])):
-                        listens.append(Listen.from_influx(row).to_api())
+                        for row in rows:
```
mayhem: At this point the rows should be dumped to the file and memory cleared, so the whole listens don't have to be stored in ram.

paramsingh: Hmm, on it.

PR #278, opened 2017-10-22
LB-231: Add code volumes to web and scheduler containers

paramsingh:
This seemed to have gotten removed somehow and it made development
a pain because to get changes into the docker container, you
have to run develop.sh from scratch.


alastair:
Did you know that docker-compose files also have a `volumes_from` command? https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes_from

PR #280, opened 2017-10-24
LB-233: Refactor testcase classes to use single path_to_data_file

paramsingh:
I was working on creating a ListenBrainzTestCase class, but it seemed like too big of a refactor with problems like tests for server would require db setup and stuff. So this PR.

### DIFF
--- a/listenbrainz/db/testing.py
+++ b/listenbrainz/db/testing.py
@@ -6,7 +6,7 @@ from listenbrainz import config
 from listenbrainz import db
 
 ADMIN_SQL_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..','admin', 'sql')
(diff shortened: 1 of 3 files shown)


listenbrainz/db/testing.py
```diff
@@ -6,7 +6,7 @@
 from listenbrainz import db
 
 ADMIN_SQL_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..','admin', 'sql')
-TEST_DATA_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'test_data')
```
alastair: this is `test_data` and all other instances are `testdata` - does this change affect any test files, or are there actually no directories called `test_data` in the repo? (I just checked, it seems like there aren't...)

paramsingh: there were no directories named `test_data`.

alastair: ok, no problem

PR #285, opened 2017-11-11
Use correct port number on final URL

jflory7:
The development environment setup docs still refers to port 8000. This uses the correct default port, 80.

_Edit_: The Travis CI build failed because of `psycopg2`, but this was fixed in #289. Re-running the test will build it fine.


jflory7:
Deprecated by #291.

PR #289, opened 2017-11-11
LB-239: Update psycopg2

paramsingh:
LB's build is broken because of a bug in the version of psycopg2 we use (https://github.com/psycopg/psycopg2/issues/594). Upgrading fixes this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,7 +19,7 @@ pytest==2.9.1
 pytest-cov==2.2.1
 pytz==2016.6.1
 unittest2==1.1.0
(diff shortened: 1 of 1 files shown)


paramsingh:
the build will fail currently because messybrainz also uses old psycopg which causes problems during the build. I've sent a PR upgrading there too: https://github.com/metabrainz/messybrainz-server/pull/23

paramsingh:
Build passing after merge of https://github.com/metabrainz/messybrainz-server/pull/23

PR #292, opened 2017-11-12
LB-238: Use BrainzUtils CustomFlask instead of Flask

paramsingh:
These changes bring a lot of consistency among LB, CB, AB and
MsB. Also, using BrainzUtils means that Flask versions never
conflict each other due to different versions required in BrainzUtils
and our requirements.txt.


listenbrainz/bigquery-writer/bigquery-writer.py
```diff
@@ -16,6 +15,12 @@
 from listenbrainz.bigquery import NoCredentialsVariableException, NoCredentialsFileException
 from oauth2client.client import GoogleCredentials
 
+from listenbrainz import default_config as config
```
paramsingh: Not sure if these imports are worth it or if there is a better way to do this.

CB does it this way but they don't have any config imports outside of Flask apps and AB has a PR https://github.com/metabrainz/acousticbrainz-server/pull/236 replacing `config.py` with `default_config.py` and `custom_config.py`. I would be happy with a finding a consistent way of doing this, so this might be the way.

paramsingh:
This should also make integrating Sentry easier, because after this, we can do exactly the same thing that CB is doing (I assume :stuck_out_tongue:).

PR #293, opened 2017-11-12
LB-240: Dump sanitized version of the user table in the stats dump

paramsingh:
This should make it easier to import the stats dump in absence of a corresponding private dump.

Ideally, this should be reviewed and merged before #274, so that I can rebase and make the changes required to the import process after the extra table is added to the stats dump.

### DIFF
--- a/listenbrainz/db/dump.py
+++ b/listenbrainz/db/dump.py
@@ -46,28 +46,13 @@ formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
 handler.setFormatter(formatter)
 logger.addHandler(handler)
 
-TABLES = {
+# this dict contains the tables dumped in stats dump as keys
+# and a tuple of columns that should be dumped as values
+STATS_TABLES = {
     '"user"': (
         'id',
         'created',
         'musicbrainz_id',
-        'auth_token',
-        'last_login',
-        'latest_import',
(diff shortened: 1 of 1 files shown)


paramsingh:
I've removed the three columns (`auth_token`, `last_login` and `latest_import`) completely from the sanitized dump. Is this the correct way to do this or is there another way?

mayhem:
I think that will still make the FK's fail upon import. We'll need to keep the columns, but put blank and/or dummy values in. For instance, in MB, when we need to dump the passwd field, we dump the data with all the passwords effectively set of "mb".  We should probably do the same here.

listenbrainz/db/dump.py
```diff
@@ -46,28 +46,18 @@
...
-        'id',
-        'user_id',
-        'sid',
-        'api_key',
-        'ts',
+        # the following are dummy values for columns that we do not want to
+        # dump in the stats dump
+        '\'auth_token\'', # auth token
+        'to_timestamp(0)', # last_login
+        'to_timestamp(0)', # latest_import
```
mayhem: I was too tired yesterday when I tried to understand this -- it feels a little odd, but in the general this is a clever way of solving the problem.

paramsingh: thanks!

PR #295, opened 2017-11-15
Put license path in listenbrainz/init for easy future imports

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Put license path in listenbrainz/init for easy future imports

# Problem

Have to import license path in InfluxListenStore for dumping
listens too. Keeping it in db.dump makes it hard to share this variable.

# Solution

Putting the path in `listenbrainz/__init__.py` makes more sense and makes it easier to
import it in the future.

# Action

Just a merge.

### DIFF
--- a/listenbrainz/__init__.py
+++ b/listenbrainz/__init__.py
@@ -1,1 +1,6 @@
+import os
+
 __version__ = '0.1'
+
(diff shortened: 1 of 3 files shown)


paramsingh:
This should be rebased once before merge.

PR #297, opened 2017-11-17
LB-229: Add support for importing listens from data dumps to Influx

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Add a function to `InfluxListenStore` that imports listens from a listen data dump created by LB.

# Problem

* JIRA ticket (_optional_): [LB-229](https://tickets.metabrainz.org/browse/LB-229)

It wasn't possible to actually import back listens into master archive from a data dump.

# Solution

Just wrote some code that opens the tar file, checks the schema version and imports
listens from each user's file into their measurement.

Also added a click command to `manage.py` to import: `python manage.py dump import_listens_dump -l <location>`

And added a test for the import method.

# Action

This is based on #277, so it'll need to be rebased after that gets merged.

### DIFF
--- a/listenbrainz/db/dump_manager.py
+++ b/listenbrainz/db/dump_manager.py
@@ -86,11 +86,13 @@ def import_dump(location):
 
 @cli.command()
 @click.option('--location', '-l')
-def import_listens_dump(location=None):
+@click.option('--threads', '-t', type=int)
+def import_listens_dump(location=None, threads=None):
     """ Import a ListenBrainz listen dump into the Influx database.
 
         Args:
             location (str): path to the listenbrainz listen .tar.xz archive
+            threads (int): the number of threads to use while decompressing, defaults to 1
     """
 
     if not location:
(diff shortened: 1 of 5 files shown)


paramsingh:
I cannot, for the life of me, understand why the test is failing on Travis. It passes locally for me, if someone else can check on their local setup, it would be helpful. In the meanwhile, I'm doing a clean build and trying again.

paramsingh:
Adding a sleep worked. :man_shrugging:

listenbrainz/db/dump_manager.py
```diff
@@ -64,3 +64,21 @@ def create(location, threads):
...
+        return
+
+    ls = init_influx_connection(log,  {
+        'REDIS_HOST': config.REDIS_HOST,
+        'REDIS_PORT': config.REDIS_PORT,
+        'INFLUX_HOST': config.INFLUX_HOST,
+        'INFLUX_PORT': config.INFLUX_PORT,
+        'INFLUX_DB_NAME': config.INFLUX_DB_NAME,
+    })
+    ls.import_listens_dump(location)
```
mayhem: We need more error catching here -- there are a lot of things that could go wrong during the import and we need to handle them here.

PR #298, opened 2017-11-17
LB-85: Make username in the profile URL case insensitive

kartikeyaSh:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Username in the profile URL was case sensitive

* JIRA ticket (_optional_): LB-85

# Solution
After querying the user with given user name from the URL(which is case insensitive), we use the user name of the queried user instead of using the user name from the URL.

### DIFF
--- a/listenbrainz/webserver/views/user.py
+++ b/listenbrainz/webserver/views/user.py
@@ -51,6 +51,8 @@ def profile(user_name):
     playing_now_conn = webserver.redis_connection._redis
 
     user = _get_user(user_name)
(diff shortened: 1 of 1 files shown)


jflory7:
@kartikeyaSh Thanks for making your first contribution to ListenBrainz! 🎉 💯 🎉  Either @paramsingh or @mayhem will take a look here soon for review.

PR #300, opened 2017-11-22
Add config options for RabbitMQ exchanges and queues

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

We would like to use different RabbitMQ pipelines for production and beta. This would need those
options to be configurable in the config file.

# Solution

Just add options to the config files and use them everywhere in the code.

### DIFF
--- a/listenbrainz/custom_config.py.sample
+++ b/listenbrainz/custom_config.py.sample
@@ -37,10 +37,10 @@ RABBITMQ_PASSWORD = "guest"
 RABBITMQ_VHOST = "/"
 
 # RabbitMQ exchanges and queues
-INCOMING_EXCHANGE = "incoming_exchange"
-INCOMING_QUEUE = "incoming_queue"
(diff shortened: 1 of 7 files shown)


listenbrainz/custom_config.py.sample
```diff
@@ -36,6 +36,12 @@ RABBITMQ_USERNAME = "guest"
 RABBITMQ_PASSWORD = "guest"
 RABBITMQ_VHOST = "/"
 
+# RabbitMQ exchanges and queues
+INCOMING_EXCHANGE = "incoming_exchange"
```
mayhem: I think we should ditch the _exchange and _queue bits. I believe they are separate namespaces and therefore a bit redundant. And if we change the names of the queues, I fear that we'll lose some listens when we deploy this.

PR #301, opened 2017-11-22
LB-248: Add on_close_callback to RabbitMQ connections

paramsingh:
Possible fix for LB-248 - BigQuery crashing.

Added on close callbacks to the rabbitmq connections.


paramsingh:
Whoops, this should be based off `production`.

PR #302, opened 2017-11-22
LB-248: Add on_close_callback to RabbitMQ connections

paramsingh:
Possible fix for LB-248 - BigQuery crashing.

Added on close callbacks to the rabbitmq connections.

### DIFF
--- a/listenbrainz/influx-writer/influx-writer.py
+++ b/listenbrainz/influx-writer/influx-writer.py
@@ -41,6 +41,10 @@ class InfluxWriterSubscriber(object):
         self.inserts = 0
         self.time = 0
 
+    @staticmethod
+    def static_on_connection_closed(connection, reply_code, reply_text, obj):
+        obj.on_connection_closed(connection, reply_code, reply_text)
+
 
     def on_connection_closed(self, connection, reply_code, reply_text):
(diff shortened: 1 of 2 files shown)


paramsingh:
Tried to think of ways to add tests somehow for this, but couldn't really come up with anything good.

listenbrainz/influx-writer/influx-writer.py
```diff
@@ -42,6 +42,11 @@ def __init__(self):
         self.time = 0
 
 
+    def on_connection_closed(self, connection, reply_code, reply_text):
```
mayhem: there is no static method as a go between -- how does this work?

paramsingh: this is how it was done in these [usage example docs](https://pika.readthedocs.io/en/0.10.0/examples/asynchronous_consumer_example.html)

but i've changed it to using a static method as done for the other callback.

mayhem: Wow, I had no idea that was possible -- when I tried it, I had no luck. But it seems so:

https://stackoverflow.com/questions/10003833/how-does-python-distinguish-callback-function-which-is-a-member-of-a-class

(I'll remove those static functions next time I'm mucking around there)

PR #304, opened 2017-11-23
LB-247: Paginate listens before sending them to BigQuery

paramsingh:
Keep track of listens and delivery tags in lists and
only ack after the messages have been sent.

# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Paginate listens before sending them to BigQuery. 

# Problem

The bigquery writer is currently reading one "block" of listens from RabbitMQ and then sending it to BigQuery. This is currently 50 listens at a time, so updating a backlog takes a really long time. 

Ideally, the writer would read up to a max number of listens from RabbitMQ, submit a much larger batch, making things much faster. We'll need to take care that we can't possibly lose the listens until they are successfully submitted to BQ. I think this needs careful management of the RabbitMQ consumption acknowledgement.

* JIRA ticket (_optional_): [LB-247](https://tickets.metabrainz.org/browse/LB-247)

# Solution

Increase prefetch count, keep track of listens in a list and only submit when 
the number of listens becomes greater than a certain value. Also, do 
multiple acknowledgements of all delivery tags for a batch.

### DIFF
--- a/listenbrainz/bigquery-writer/bigquery-writer.py
+++ b/listenbrainz/bigquery-writer/bigquery-writer.py
@@ -93,7 +93,7 @@ class BigQueryWriter(ListenWriter):
             except pika.exceptions.ConnectionClosed:
                 self.connect_to_rabbitmq()
 
-            sleep(ERROR_RETRY_DELAY)
+            sleep(self.ERROR_RETRY_DELAY)
 
         # collect and occasionally print some stats
         time_taken = time() - t0
--- a/listenbrainz/bigquery-writer/bigquery-writer.py
+++ b/listenbrainz/bigquery-writer/bigquery-writer.py
@@ -27,7 +27,8 @@ except ImportError:
 
 # the number of listens to send to BQ in one batch
 # NOTE: this MUST be greater than the maximum number of listens sent to us in one
-# RabbitMQ batch
+# RabbitMQ batch, otherwise BigQueryWriter will submit a partial batch and send an ack for
+# the batch.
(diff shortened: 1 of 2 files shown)


paramsingh:
Working on adding a timer that flushes listens after some time.

listenbrainz/bigquery-writer/bigquery-writer.py
```diff
@@ -1,32 +1,38 @@
...
 
-REPORT_FREQUENCY = 5000
 ERROR_RETRY_DELAY = 3 # number of seconds to wait until retrying an operation
 DUMP_JSON_WITH_ERRORS = True
 
-# TODO:
-#   Big query hardcoded data set ids
+# the number of listens to send to BQ in one batch
+# NOTE: this MUST be greater than the maximum number of listens sent to us in one
+# RabbitMQ batch
```
mayhem: what happens if it is not?

listenbrainz/bigquery-writer/bigquery-writer.py
```diff
@@ -120,29 +109,91 @@ def callback(self, ch, method, body):
...
+        # reset back to normal
+        self.bq_data = []
+        self.delivery_tags = []
+
+
+    def callback(self, ch, method, body):
+
+        # if some timeout exists, remove it as we'll add a new one
+        # before this method exits
+        if self.timer_id is not None:
```
mayhem: oh, timer -- good idea!

PR #306, opened 2017-11-24
LB-250: Move test data to json file

kartikeyaSh:
Test data should be present in testdata folder not in python files.
And data preprocessing should be done in util file so that data can
be used for other tests.

# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
We have json data in the python file. but we already have testdata folder for json data.
And it's good to use util.py for generating test data.

# Solution
Create listens.json in testdata folder and put the json data into it.
Add function into util.py for data processing.
Remove json data from python file and use the function created in util.py to get test data.

### DIFF
--- a/listenbrainz/listenstore/tests/util.py
+++ b/listenbrainz/listenstore/tests/util.py
@@ -3,6 +3,11 @@
 from datetime import datetime, timedelta
 from listenbrainz.listen import Listen
 import uuid
+import json
+import os
+
+
+TEST_DATA_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', '..', 'testdata')
 
 
 def generate_data(test_user_id, from_ts, num_records):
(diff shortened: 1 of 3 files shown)


listenbrainz/testdata/listens.json
```diff
@@ -0,0 +1,75 @@
+{
```
paramsingh: this file should be renamed to something more clear like `influx_listenstore_test_listens.json` or something similar.

listenbrainz/listenstore/tests/util.py
```diff
@@ -19,3 +24,16 @@ def generate_data(test_user_id, from_ts, num_records):
 
 def to_epoch(date):
     return int((date - datetime.utcfromtimestamp(0)).total_seconds())
+
+
+def create_test_data_for_influxlistenstore(user_name):
```
paramsingh: A docstring for this function would be great. :)

paramsingh: We try to use something similar to  [Google style docstrings](https://google.github.io/styleguide/pyguide.html?showone=Comments#Comments), [example](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/db/user.py#L93)

mayhem:
Should this PR be against production or master, come to think of it?

kartikeyaSh:
@mayhem for https://tickets.metabrainz.org/projects/LB/issues/LB-85?filter=allopenissues i wrote this test https://github.com/kartikeyaSh/listenbrainz-server/commit/ce97f53ff7a0a546e904da0f23ee9c0049e65c3b . This test needs changes made in above PR. So if you think that LB-85 should be added to production then this PR should be in production branch. otherwise I'll have to either copy this test data to other files or use some other way to create listens for the test.

PR #307, opened 2017-11-28
LB-85:  Make username in the profile URL case insensitive 

kartikeyaSh:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Username in the profile url is case senstive.

* JIRA ticket (_optional_): https://tickets.metabrainz.org/projects/LB/issues/LB-85?filter=allopenissues

# Solution
After querying the user with given user name from the URL(which is case insensitive), we use the user name of the queried user instead of using the user name from the URL.

# Action
1. use the user name of the queried user instead of using the user name from the URL.
2. write an unit test to check case insensitivity.

### DIFF
--- a/listenbrainz/webserver/views/test/test_user.py
+++ b/listenbrainz/webserver/views/test/test_user.py
@@ -1,8 +1,13 @@
-from flask import url_for
+from flask import url_for, current_app
+from influxdb import InfluxDBClient
+
+from listenbrainz.db.testing import DatabaseTestCase
+from listenbrainz.listenstore.tests.util import create_test_data_for_influxlistenstore
+from listenbrainz.webserver.influx_connection import init_influx_connection
+from listenbrainz.webserver.testing import ServerTestCase
 
 import listenbrainz.db.user as db_user
-from listenbrainz.db.testing import DatabaseTestCase
-from listenbrainz.webserver.testing import ServerTestCase
+import logging
 
 
 class UserViewsTestCase(ServerTestCase, DatabaseTestCase):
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/test/test_user.py
```diff
@@ -12,7 +18,22 @@ def setUp(self):
         self.user = db_user.get_or_create('iliekcomputers')
         self.weirduser = db_user.get_or_create('weird\\user name')
 
+        self.log = logging.getLogger(__name__)
+
+        influx = InfluxDBClient(host=config.INFLUX_HOST, port=config.INFLUX_PORT, database=config.INFLUX_DB_NAME)
+        influx.query('''drop database %s''' % config.INFLUX_DB_NAME)
```
paramsingh: I think these lines might not be needed here. They were put in `InfluxListenStore` tests to ensure that the database was empty at the beginning of each test run, but here since there's only one test changing the influx db, it shouldn't be a problem. Also, since we might want to clear the db after the test run, so moving this to `tearDown` might be a good idea.

listenbrainz/webserver/views/test/test_user.py
```diff
@@ -41,3 +62,17 @@ def test_scraper_username(self):
...
         self.assert200(response)
         self.assertIn('var user_name = "weird%5Cuser%20name";', response.data.decode('utf-8'))
+
+    def _create_test_data(self, user_name):
+        test_data = create_test_data_for_influxlistenstore(user_name)
+        self.logstore.insert(test_data)
+
+    def test_username_case(self):
+        """Tests that the username in URL is case insenstive"""
+        self._create_test_data(self.user['musicbrainz_id'])
```
paramsingh: Making the user_name explicit here might be a good idea, considering we use the user name (and a modified version) later.

`self._create_test_data('iliekcomputers')`

listenbrainz/webserver/views/test/test_user.py
```diff
@@ -4,6 +4,12 @@
 from listenbrainz.db.testing import DatabaseTestCase
 from listenbrainz.webserver.testing import ServerTestCase
 
+from listenbrainz.listenstore.tests.util import create_test_data_for_influxlistenstore
```
paramsingh: While you're adding imports, could you please seperate the `import x` and `from x import y` statements and alphabetize them?

listenbrainz/webserver/views/test/test_user.py
```diff
@@ -12,7 +18,23 @@ def setUp(self):
...
+        self.logstore = init_influx_connection(self.log, {
+            'REDIS_HOST': config.REDIS_HOST,
+            'REDIS_PORT': config.REDIS_PORT,
+            'INFLUX_HOST': config.INFLUX_HOST,
+            'INFLUX_PORT': config.INFLUX_PORT,
+            'INFLUX_DB_NAME': config.INFLUX_DB_NAME,
+        })
+
     def tearDown(self):
+        influx = InfluxDBClient(host=config.INFLUX_HOST, port=config.INFLUX_PORT, database=config.INFLUX_DB_NAME)
```
paramsingh: Creating individual clients for both setUp and tearDown would be a bit inefficient, I guess.

PR #308, opened 2017-11-29
LB-213: Add API usage examples to docs

eshansingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
This adds fully executable usage examples in Python 3 for the current API version. Current version includes examples for:

* [x] ~~`submit-listens`~~
* [x] ~~`get-listens`~~
* [x] ~~`set-latest-import`~~
* [x] ~~`get-latest-import`~~

# Problem

* JIRA ticket: https://tickets.metabrainz.org/projects/LB/issues/LB-213

# Solution

Each endpoint has an example file that's executable and demonstrates an example usecase for that particular endpoint.

# Action

Some aspects of how long the examples are, what additional details are needed, what parts are extraneous (for example, is the `verify_args` function useful for examples) etc.

### DIFF
--- a/docs/dev/api-usage.rst
+++ b/docs/dev/api-usage.rst
@@ -1,0 +1,100 @@
+.. role:: json(code)
+   :language: json
+
+.. role:: py(code)
+   :language: python
+
+API Usage Examples
+==================
+
+.. note::
+
+   These examples are written in Python version **3.6.3** and use `requests
+   <http://docs.python-requests.org/en/master/>`_ version **2.18.4**.
+
+Prerequisites
+#############
+
+All the examples assume you have a development version of the ListenBrainz
+server set up on :code:`localhost`. Remember to set :code:`DEBUG` to :py:`True`
+in the config. When in production, you can replace :code:`localhost` with
+:code:`api.listenbrainz.org` to use the real API. In order to use either one,
+you'll need a token. You can find it under :code:`ROOT/profile/` when signed
+in, with :code:`ROOT` being either :code:`localhost` for the dev version or
+:code:`listenbrainz.org` for the real API.
+
+.. caution::
+
+   You should use the token from the API you're using. In production, change the
+   token to one from :code:`listenbrainz.org`.
+
+If you want to run the samples, you'll need to add the imports from this common
+file that are required by each sample:
+
+.. include:: ./api_usage_examples/common.py
+   :code: python
+
+Examples
+########
+
+Submitting Listens
+------------------
+
+See :ref:`json-doc` for details on the format of the Track dictionaries.
+
+If everything goes well, the json response should be :json:`{"status": "ok"}`,
+and you should see a recent listen of "Never Gonna Give You Up" when you visit
+:code:`ROOT/user/{your-user-name}`.
+
+.. include:: ./api_usage_examples/submit_listens.py
+   :code: python
+
+Getting Listen History
+----------------------
+
+See :ref:`json-doc` for details on the format of the Track dictionaries.
+
+If there's nothing in the listen history of your user, you can run
+:code:`submit_listens` before this.
+
+If there is some listen history, you should see a list
+of tracks like this:
+
+.. code::
+
+    Track: Never Gonna Give You Up, listened at 1512040365
+    Track: Never Gonna Give You Up, listened at 1511977429
+    Track: Never Gonna Give You Up, listened at 1511968583
+    Track: Never Gonna Give You Up, listened at 1443521965
+    Track: Never Gonna Give You Up, listened at 42042042
+
+.. include:: ./api_usage_examples/get_listens.py
+   :code: python
+
+Latest Import
+-------------
+
+Set and get the timestamp of the latest import into ListenBrainz.
+
+Setting
+^^^^^^^
+
+.. include:: ./api_usage_examples/set_latest_import.py
+   :code: python
+
+Getting
+^^^^^^^
+
+If your user has never imported before and the latest import has never been
+set by a script, then the server will return :code:`0` by default. Run
+:code:`set_latest_import` before this if you don't want to actually import any
+data.
+
+You should see output like this:
+
+.. code::
+
+   User naiveaiguy last imported on 30 11 2017 at 12:23
+
+.. include:: ./api_usage_examples/get_latest_import.py
+   :code: python
--- a/docs/dev/api_usage_examples/get_latest_import.py
+++ b/docs/dev/api_usage_examples/get_latest_import.py
@@ -1,0 +1,44 @@
+'''
+get_latest_import.py
+'''
+
+import requests
+
+ROOT = '127.0.0.1'
+TOKEN = 'YOUR_TOKEN_HERE'
+USERNAME = 'YOUR_USERNAME_HERE' # Replace with the username from which you got the token above.
+AUTH_HEADER = {
+    "Authorization": "Token {0}".format(TOKEN)
+}
+
+# The AUTH_HEADER token can be any valid token.
+def get_latest_import(username):
+    """Gets the latest import timestamp of a given user.
+
+    Args:
+        username: User to get latest import time of.
+
+    Returns:
+        A Unix timestamp if there's an OK status.
+
+    Raises:
+        An HTTPError if there's a failure.
+        A ValueError if the JSON in the response is invalid.
+        An IndexError if the JSON is not structured as expected.
+    """
+    response = requests.get(
+        url="http://{0}/1/latest-import".format(ROOT),
+        params={
+            "user_name": username
+        },
+        headers=AUTH_HEADER
+    )
+
+    response.raise_for_status()
+
+    return response.json()["latest_import"]
+
+if __name__ == "__main__":
+    timestamp = get_latest_import(USERNAME)
+
+    print("User {0} last imported on {1}".format(USERNAME, timestamp))
--- a/docs/dev/api_usage_examples/get_listens.py
+++ b/docs/dev/api_usage_examples/get_listens.py
@@ -1,0 +1,55 @@
+'''
+get_listens.py
+'''
+
+import requests
+
+ROOT = '127.0.0.1'
+TOKEN = 'YOUR_TOKEN_HERE'
+USERNAME = 'YOUR_USERNAME_HERE' # Replace with the username from which you got the token above.
+AUTH_HEADER = {
+    "Authorization": "Token {0}".format(TOKEN)
+}
+
+# The token in AUTH-HEADER must be valid, but it doesn't have to be the token of the user you're
+# trying to get the listen history of.
+def get_listens(username, min_ts=None, max_ts=None, count=None):
+    """Gets the listen history of a given user.
+
+    Args:
+        username: User to get listen history of.
+        min_ts: History before this timestamp will not be returned.
+                DO NOT USE WITH max_ts.
+        max_ts: History after this timestamp will not be returned.
+                DO NOT USE WITH min_ts.
+        count: How many listens to return. If not specified,
+               uses a default from the server.
+
+    Returns:
+        A list of listen info dictionaries if there's an OK status.
+
+    Raises:
+        An HTTPError if there's a failure.
+        A ValueError if the JSON in the response is invalid.
+        An IndexError if the JSON is not structured as expected.
+    """
+    response = requests.get(
+        url="http://{0}/1/user/{1}/listens".format(ROOT, username),
+        params={
+            "min_ts": min_ts,
+            "max_ts": max_ts,
+            "count": count
+        },
+        headers=AUTH_HEADER
+    )
+
+    response.raise_for_status()
+
+    return response.json()['payload']['listens']
+
+if __name__ == "__main__":
+    listens = get_listens(USERNAME)
+
+    for track in listens:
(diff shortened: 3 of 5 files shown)


eshansingh:
The Travis build is always failing in the same way (`TestInfluxListenStore.test_import_listens` is failing), and apparently, if you restart them they fix themselves.

eshansingh:
All of the example *code* is done, but I haven't added any detail to the documentation regarding those examples. I'll add whatever I feel is necessary right now, but I would love some feedback regarding it. 

Additionally, my `set-latest-import` commit passed Travis on the first try, which is odd.

docs/dev/api-usage.rst
```diff
@@ -0,0 +1,93 @@
...
+Prerequisites
+#############
+
+All the examples assume you have a development version of the ListenBrainz
+server set up on :code:`localhost`. Remember to set :code:`DEBUG` to :py:`True`
+in the config. When in production, you can replace :code:`localhost` with
+:code:`api.listenbrainz.org` to use the real API. In order to use either
+one, you'll need a token. You can find it under :code:`ROOT/profile/`
+when signed in, with :code:`ROOT` being either :code:`localhost` or
+:code:`api.listenbrainz.org`.
```
paramsingh: We don't want to encourage users to use `api.listenbrainz.org` for anything other than sending API requests. We should change this to `listenbrainz.org`.

docs/dev/api-usage.rst
```diff
@@ -0,0 +1,93 @@
...
+
+.. caution::
+
+   Use the token from the version you're testing with. You can't
```
paramsingh: Ideally, we wouldn't want to encourage people to test the API by submitting test data to production. I think we should remove this.

docs/dev/api_usage_examples/common.py
```diff
@@ -0,0 +1,54 @@
...
+
+ROOT = '127.0.0.1'
+API_VERSION = 1
+TOKEN = '93fb3a42-9177-48a0-b07a-e8a7019218f6' # Replace with your dev token.
+USERNAME = 'naiveaiguy' # Replace with the username from which you got the token above.
+AUTH_HEADER = {
+    "Authorization": f"Token {TOKEN}"
+}
+
+class Track(dict):
```
paramsingh: Because these are simple API usage examples, I'd prefer to keep boilerplate code to a minimum. Just using a dictionary instead of creating new classes is something we prefer in LB (and other MeB Python projects), so I'd suggest removing this class and just creating a dict listen like

```
EXAMPLE_PAYLOAD = {
    "listen_type": "single",
    "payload": [
        {
            "listened_at": 1486449409,
            "track_metadata": {
                "artist_name": "Kanye West",
                "track_name": "Fade",
                "release_name": "The Life of Pablo"
            }
        }
    ]
}
```

docs/dev/api_usage_examples/common.py
```diff
@@ -0,0 +1,54 @@
...
+        "release_name": "Whenever you need somebody"
+    })
+
+class ListenType(Enum):
+    Single = "single"
+    Playing = "playing_now"
+    Import = "import"
+
+def construct_url(resource_url):
+    return f"http://{ROOT}/{API_VERSION}/{resource_url}"
```
paramsingh: Although LB runs on Python 3.6, we try to write 3.5 code as readthedocs doesn't support Python 3.6. You could use .format here, although, I would prefer explicit URLs in the `requests.post` and `requests.get` calls as they would make understanding the examples and API endpoints much easier.

docs/dev/api_usage_examples/get_latest_import.py
```diff
@@ -0,0 +1,41 @@
+'''
+get_latest_import.py
+'''
+
+from common import AUTH_HEADER, USERNAME, construct_url
+
+import requests as req
```
paramsingh: I would prefer `import requests`, renaming seems unnecessary.

docs/dev/api_usage_examples/get_latest_import.py
```diff
@@ -0,0 +1,41 @@
+'''
+get_latest_import.py
+'''
+
+from common import AUTH_HEADER, USERNAME, construct_url
```
paramsingh: Since these are examples, I'd prefer to keep these explicit. So that people can just look at the `submit_listen` code and understand it, without having to go into `common.py` to understand what the `AUTH_HEADER` is exactly.

docs/dev/api_usage_examples/get_latest_import.py
```diff
@@ -0,0 +1,41 @@
...
+        A ValueError if the JSON in the response is invalid.
+        An IndexError if the JSON is not structured as expected.
+    """
+    url = construct_url("latest-import")
+
+    params = {
+        "user_name": username
+    }
+
+    response = req.get(url, params=params, headers=AUTH_HEADER)
```
paramsingh: I think most of the code should be in this function.

Something like

```
response = requests.get(
   url,
   params={
       'user_name': username,
   },
   headers={
       'Authorization: 'Token <your token>',
   }
)
```

would be much clearer.

docs/dev/api_usage_examples/get_listens.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # Thanks to requests, if any of these values are None,
+    # they're not going to be added in the first place.
+    params = {
+        "min_ts": min_ts,
+        "max_ts": max_ts,
+        "count": count
+    }
+
+    response = req.get(url, params=params, headers=AUTH_HEADER)
```
paramsingh: Same comment as the one for get_latest_import here.

docs/dev/api_usage_examples/set_latest_import.py
```diff
@@ -0,0 +1,40 @@
...
+        An HTTPError if there's a failure.
+        A ValueError if the JSON response is invalid.
+    """
+    url = construct_url("latest-import")
+
+    json_data = {
+        "ts": timestamp
+    }
+
+    response = req.post(url, json=json_data, headers=AUTH_HEADER)
```
paramsingh: Same here, `json_data` and `AUTH_HEADER` should be much more explicit.

docs/dev/api_usage_examples/submit_listens.py
```diff
@@ -0,0 +1,53 @@
...
+    json_data = {
+        "listen_type": listen_type.value,
+        "payload": payload
+    }
+
+    response = req.post(url, json=json_data, headers=AUTH_HEADER)
+
+    response.raise_for_status() # Will raise in case of a non-OK status
+
+    return response.json()
```
paramsingh: I think an example of the data that is returned here would be good to add too.

docs/dev/api_usage_examples/common.py
```diff
@@ -0,0 +1,54 @@
...
+common.py
+'''
+
+from enum import Enum
+from time import time
+from datetime import datetime
+
+ROOT = '127.0.0.1'
+API_VERSION = 1
+TOKEN = '93fb3a42-9177-48a0-b07a-e8a7019218f6' # Replace with your dev token.
```
paramsingh: Sharing a token is not a good idea, should be replaced with some boilerplate.

docs/dev/api_usage_examples/common.py
```diff
@@ -0,0 +1,54 @@
...
+USERNAME = 'naiveaiguy' # Replace with the username from which you got the token above.
+AUTH_HEADER = {
+    "Authorization": f"Token {TOKEN}"
```
paramsingh: Use `.format` here. Python 3.6 fstrings cause problems when readthedocs builds the documentations.

docs/dev/api_usage_examples/get_listens.py
```diff
@@ -0,0 +1,59 @@
...
+    return response.json()['payload']['listens']
+
+if __name__ == "__main__":
+    listens = get_listens(USERNAME)
+    listens_tracks = [Track(track_dict) for track_dict in listens]
+
+    for track in listens_tracks:
+        print(track.simple_repr())
+
+def verify_args(username, min_ts=None, max_ts=None, count=None):
```
paramsingh: This does not seem to be relevant to the api endpoints much. Maybe it can be removed.

paramsingh:
@jflory7: I think a review from you would be helpful here too.

docs/dev/api-usage.rst
eshansingh: Okay, I'll rewrite this to say "You should use the token from the API you're using. In production, change the token to one from `listenbrainz.org.`" Is that good?

docs/dev/api-usage.rst
eshansingh: I don't quite get what you mean here, apologies. Do you mean for the token? Or in general? Because in production, AFAIK, you have to use `api.listenbrainz.org`, right?

docs/dev/api_usage_examples/common.py
eshansingh: Sure.

docs/dev/api_usage_examples/common.py
eshansingh: My reasoning for this was that it was more semantically understandable in examples. I'm not sure about just using a dictionary.

docs/dev/api_usage_examples/common.py
eshansingh: I figured it would make it easier to update for users, and for us when the API version changes. Though to be fair, if the API version changes, we'll have to change the code anyway, so I'll just remove this function and use explicit URLs.

docs/dev/api_usage_examples/get_latest_import.py
eshansingh: `common.py` is fairly short, and it's given and the top of `api-usage.rst`. Should I just copy the required things that are in `common.py` right into the start of every file? That seems excessive to me, and it would only increase the size of the examples.

docs/dev/api_usage_examples/get_latest_import.py
eshansingh: Sorry, just a habit of mine. I'm a lazy Emacs user 😝

docs/dev/api_usage_examples/get_latest_import.py
eshansingh: Yeah, you're right.

docs/dev/api_usage_examples/get_listens.py
eshansingh: Yeah, even I wasn't sure about these - I'll remove both the `verify_args` functions.

docs/dev/api_usage_examples/submit_listens.py
eshansingh: But this is `submit_listens.py`. If it all goes well, the only thing that's returned is `{"status": "ok"}`, which seems useless to include here.

eshansingh:
Implemented some of the requested changes that I'm sure about. I'm still not sure what to do with the remains of `common.py`. If needed, I'll squash these small commits into a big refactoring commit once this PR is approved.

docs/dev/api_usage_examples/common.py
eshansingh: Now that I think about it, both `Track` and `simple_repr` are only used in one file each. I'll simplify those places.

eshansingh:
I've refactored `common.py` so it only includes constants. I've removed the `Track` class and the `ListenType` enum.

docs/dev/api_usage_examples/common.py
eshansingh: Done.

docs/dev/api-usage.rst
paramsingh: In general, you should get your production auth token from `listenbrainz.org` instead of `api.listenbrainz.org`. `listenbrainz.org` is the user facing site while `api.listenbrainz.org` is the place which handles the API calls.

docs/dev/api_usage_examples/common.py
paramsingh: I know, but still feels weird to share ANY actual token.

docs/dev/api_usage_examples/get_latest_import.py
paramsingh: I think if we create examples that are easy to read individually without having to read the entire document, it would be better. That is why I suggested not using imports because then I have to go up to see what `AUTH_TOKEN` means, which is extra baggage if I'm a random guy who just wants to see how to submit listens and move on.

docs/dev/api_usage_examples/submit_listens.py
paramsingh: That is something that we could mention in `api_usage.rst`.

"If everything worked, you should have received a json response of the form `{"status": "ok"}`"

docs/dev/api_usage_examples/common.py
eshansingh: Changed it to `YOUR_TOKEN_HERE`.

docs/dev/api_usage_examples/submit_listens.py
eshansingh: Ah, I see. Will do.

eshansingh:
Implemented last few requested changes. Again, I'll be happy to `git rebase` the hell out of these commits if they're too many.

eshansingh:
Squashed all of 'em.

PR #310, opened 2017-12-02
LB-196: Remove duplication of RabbitMQ connection code

eshansingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

This change extracts the common RabbitMQ connection code into one function in `utils.py`.

# Problem

* JIRA ticket: https://tickets.metabrainz.org/browse/LB-196

# Solution

`connect_to_rabbitmq` in `utils.py` tries repeatedly until it's able to connect to RabbitMQ and then returns the `Connection` object.

# Action

@paramsingh

### DIFF
--- a/listenbrainz/influx-writer/influx-writer.py
+++ b/listenbrainz/influx-writer/influx-writer.py
@@ -14,6 +14,7 @@ from listenbrainz.listenstore import InfluxListenStore
 from listenbrainz.utils import escape, get_measurement_name, get_escaped_measurement_name, \
                                get_influx_query_timestamp, convert_to_unix_timestamp, \
                                convert_timestamp_to_influx_row_format
+import listenbrainz.utils as utils
 from requests.exceptions import ConnectionError
 from redis import Redis
 from collections import defaultdict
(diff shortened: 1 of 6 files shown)


eshansingh:
I am not sure where this Travis error is coming from. It appeared in #308 as well, seemingly randomly.

eshansingh:
@paramsingh Out of curiosity, I ran `./test.sh` two minutes ago on my local copy of this branch, without any local directory changes, no new commits, nothing. Magically, it worked. So I thought the issue kind of resolved itself here, or at the very least I could make the Travis build pass for now. I made a really minor change to a _docstring_ so I would have something I could run `git commit --amend` on. I force-pushed, and the Travis build fails, once again in the same way.

It seems to be a problem with how the Travis machine is configured or started, or something because I can't seem a pattern here.

listenbrainz/influx-writer/influx-writer.py
```diff
@@ -14,6 +14,8 @@
 from listenbrainz.utils import escape, get_measurement_name, get_escaped_measurement_name, \
                                get_influx_query_timestamp, convert_to_unix_timestamp, \
                                convert_timestamp_to_influx_row_format
+# Doing this so as not to confuse the method names
+from listenbrainz.utils import connect_to_rabbitmq as utils_connect_to_rabbitmq
```
paramsingh: Just do an `import listenbrainz.utils as utils` here too.

paramsingh:
The test failures seem unrelated, I'll take a look when I get the chance.

paramsingh:
I know i said adding to the utils file was what I said in the task but the class seems like a better idea now. 🙄

eshansingh:
@parasingh I have a few questions related to the task. If I create a new class `ListenWriter` and inherit from it in BQ listener and Influx listener, do you want the whole `util.connect_to_rabbitmq` method to go there?

In this case, what should be done for `webserver/rabbitmq_connection.py`, which only has a method called `init_rabbitmq_connection`?

Do you want to move any other methods to this new base class?

eshansingh:
Made a few basic changes - a new `ListenWriter` class, with some functionality I was able to identify as common, is the base class of `BigQueryWriter` and `InfluxWriter` classes. There's some more stuff left, like the initial code in the `start` methods, but I wasn't able to extract any semantic meaning out of those sections myself - I would love some suggestions on it.

listenbrainz/webserver/rabbitmq_connection.py
```diff
@@ -14,26 +15,23 @@ def init_rabbitmq_connection(app):
...
 
-    credentials = pika.PlainCredentials(app.config['RABBITMQ_USERNAME'], app.config['RABBITMQ_PASSWORD'])
-    connection_parameters = pika.ConnectionParameters(
-            host=app.config['RABBITMQ_HOST'],
-            port=app.config['RABBITMQ_PORT'],
-            virtual_host=app.config['RABBITMQ_VHOST'],
-            credentials=credentials
-        )
+    connection_config = {
+        "username": app.config['RABBITMQ_USERNAME'],
```
paramsingh: Consistency here between `"` and `'` would be appreciated.

paramsingh: I prefer `'` :)

eshansingh:
@paramsingh Fixed. Actually I didn't realize `integration_test.sh` existed 😥 But anywho, the failures were caused by the `ListenWriter` not being imported, and me not changing `config` to `self.config` in InfluxWriter. I did it in BQWriter, but not Influx 🤦‍♂️ .

PR #313, opened 2017-12-07
LB-256: Add logging of config values and git commit to LB webserver

paramsingh:
Also, add importing config values from consul_config.py
as done in MeB.org

# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): LB-256

# Solution

Add logging and call `git rev-parse HEAD` to get the git commit on which we're running.

### DIFF
--- a/listenbrainz/utils.py
+++ b/listenbrainz/utils.py
@@ -1,5 +1,6 @@
 import errno
 import os
+import subprocess
 
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


paramsingh:
Created a new PR on `production`.

PR #316, opened 2017-12-16
LB-230: Configure scheduler to do regular data dumps

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-230](https://tickets.metabrainz.org/browse/LB-230)

# Solution

Add an entry to crontab.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,5 +51,3 @@ ADD crontab /etc/cron.d/crontab
 RUN chmod 0644 /etc/cron.d/crontab
 RUN touch /var/log/cron.log
 RUN crontab /etc/cron.d/crontab
-
-CMD service cron stop
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,3 +51,4 @@ ADD crontab /etc/cron.d/crontab
 RUN chmod 0644 /etc/cron.d/crontab
 RUN touch /var/log/cron.log
 RUN crontab /etc/cron.d/crontab
+RUN touch /etc/service/cron/down
--- a/Dockerfile
+++ b/Dockerfile
@@ -51,3 +51,5 @@ ADD crontab /etc/cron.d/crontab
 RUN chmod 0644 /etc/cron.d/crontab
 RUN touch /var/log/cron.log
 RUN crontab /etc/cron.d/crontab
+
+CMD service cron stop
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,6 @@ RUN apt-get update \
                        git \
                        libpq-dev \
                        libffi-dev \
-                       cron \
                        pxz \
     && rm -rf /var/lib/apt/lists/*
 
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -14,7 +14,6 @@ RUN apt-get update \
                        libffi-dev \
                        libssl-dev \
                        redis-tools \
-                       cron \
                        pxz \
     && rm -rf /var/lib/apt/lists/*
 
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -14,6 +14,7 @@ RUN apt-get update \
                        libffi-dev \
                        libssl-dev \
                        redis-tools \
+                       cron \
                        pxz \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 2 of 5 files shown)


paramsingh:
Build should be fixed after https://github.com/metabrainz/listenbrainz-server/pull/304/commits/013d3c64f66cb03be53f51c15e76453045d09cbb gets merged.

mayhem:
@brainzbot, test this please.

paramsingh:
Brainzbot doesn't work right now for LB, let me rebase and travis should retest this.

crontab
```diff
@@ -1,4 +1,5 @@
 PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
 PYTHONPATH=/usr/local/lib/python36.zip:/usr/local/lib/python3.6:/usr/local/lib/python3.6/lib-dynload:/usr/local/lib/python3.6/site-packages:/usr/local/lib/python3.6/site-packages/messybrainz-99.0.0.dev0-py3.6.egg
 GOOGLE_APPLICATION_CREDENTIALS=/code/credentials/bigquery-credentials.json
-* * * * * python /code/listenbrainz/manage.py stats calculate > /var/log/cron.log 2>&1
+* * * * * python /code/listenbrainz/manage.py stats calculate > /var/log/stats.log 2>&1
```
mayhem: Do you really want to calculate stats once a minute?

crontab
```diff
@@ -1,4 +1,5 @@
...
+00 00 * * 0 python /code/listenbrainz/manage.py dump create > /var/log/dump_create.log 2>&1
```
mayhem: And just to be clear, do data dumps once a week?

alastair: I'm not sure if this is a MeB standard, but when I do crontabs I always comment each command saying what it is and when it runs

Dockerfile
```diff
@@ -51,3 +51,5 @@ ADD crontab /etc/cron.d/crontab
 RUN chmod 0644 /etc/cron.d/crontab
 RUN touch /var/log/cron.log
 RUN crontab /etc/cron.d/crontab
+
+CMD service cron stop
```
mayhem: This means that cron is running for no image, right? And it is a bit of a hack, methinks. I suppose we could determine if cron needs starting/stopping in this file : https://github.com/metabrainz/listenbrainz-server/blob/master/docker/beta/uwsgi/uwsgi.service (as well as the prod version of it).

Also, changes to this Dockerfile, if we keep it should be made to Dockerfile.prod

paramsingh: Seemed hacky to me too. Changing the uwsgi.service script then.

>Also, changes to this Dockerfile, if we keep it should be made to Dockerfile.prod

on it.

Dockerfile.prod
```diff
@@ -14,6 +14,7 @@ RUN apt-get update \
                        libffi-dev \
                        libssl-dev \
                        redis-tools \
+                       cron \
```
alastair: it looks like cron is already in the `metabrainz/python:3.6` image, I don't know if you want to add it here again explicitly to make it clear that we use it

docker/beta/uwsgi/uwsgi.service
```diff
@@ -1,25 +1,57 @@
...
+
 if [ "${CONTAINER_NAME}" = "listenbrainz-web-beta" ]
 then
+    bring_cron_down
     exec uwsgi /etc/uwsgi/uwsgi.ini
 fi
 
 if [ "${CONTAINER_NAME}" = "listenbrainz-influx-writer-beta" ]
 then
+    bring_cron_down
```
alastair: I don't fully understand why you bring cron down in all containers. Do you see cron starting automatically in all other containers when you bring them up?
From what I understand cron won't start automatically in all containers when you start from docker, only the specified command will execute. In this case I think it makes sense to run `cron -f` in the scheduler container, but not stop it in the others.

Dockerfile.prod
mayhem: Ok, we now understand that this is not needed.

mayhem:
OK, we're really close! The "down" file works, but simply removing the "down" file isn't enough, we also need to do a:
  sv restart cron
In the scheduler container and cron starts. I've verified this in a production container.

docker/prod/uwsgi/uwsgi.service
```diff
@@ -22,4 +22,11 @@ then
     exec uwsgi /etc/uwsgi/uwsgi-api-compat.ini
 fi
 
+if [ "${CONTAINER_NAME}" = "listenbrainz-scheduler-prod" ]
+then
+    rm /etc/service/cron/down
+    sv restart cron
+    exec cron -f
```
mayhem: This line is not needed. Let's remove it, then we can merge this.

PR #322, opened 2017-12-20
LB-261: Show total listen count on index page ala AB

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-261](https://tickets.metabrainz.org/browse/LB-261)

# Solution

Add a table to which we can pass stats.

This table should also open the doors to showing
other stats such as artist count etc on the home
page.

### DIFF
--- a/listenbrainz/webserver/views/index.py
+++ b/listenbrainz/webserver/views/index.py
@@ -23,7 +23,19 @@ CACHE_TIME = 10 * 60 # time in seconds we cache the stats
 
 @index_bp.route("/")
 def index():
-    return render_template("index/index.html")
+
+    # get total listen count
+    try:
+        listen_count = _influx.get_total_listen_count()
+    except Exception as e:
+        current_app.logger.error('Error while trying to get total listen count: %s', str(e))
+        listen_count = None
+
+
+    return render_template(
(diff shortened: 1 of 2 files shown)


paramsingh:
screenshot_2017-12-21_00-06-21

listenbrainz/webserver/views/index.py
```diff
@@ -23,7 +23,19 @@
 
 @index_bp.route("/")
 def index():
-    return render_template("index/index.html")
+
+    # get total listen count
+    try:
```
mayhem: I think we ought to add some caching to this -- maybe for 2-3 minutes? I'd hate to have to make this (albeit small) query for loading the home page each time.

paramsingh: `get_total_listen_count` already does caching. 

https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/listenstore/influx_listenstore.py#L120

mayhem: Oh right. Grand. :)

PR #326, opened 2017-12-22
Restructure listen dumps

paramsingh:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-268](https://tickets.metabrainz.org/browse/LB-268)

# Solution

We have an index.json file which maps user names to integers and
then listens are stored as listens/(uid%1000)/user-%06d.listens

This was done because having listen files named after users is
just not feasible due to injection problems.

Also, add a test for weird user names.

### DIFF
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -583,8 +583,8 @@ class InfluxListenStore(ListenStore):
                 'size': info['size'],
             })
 
-        for file in file_contents:
-            file_contents[file] = sorted(file_contents[file], key=lambda x: x['offset'])
+        for file_name in file_contents:
+            file_contents[file_name] = sorted(file_contents[file_name], key=lambda x: x['offset'])
 
         pxz = subprocess.Popen(pxz_command, stdout=subprocess.PIPE)
 
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -32,6 +32,8 @@ TIMELINE_COUNT_MEASUREMENT = COUNT_MEASUREMENT_NAME
 
 DUMP_CHUNK_SIZE = 100000
 
+NUMBER_OF_USERS_PER_DIRECTORY = 1000
+
 
 class InfluxListenStore(ListenStore):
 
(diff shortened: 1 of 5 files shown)


paramsingh:
Calling `pxz` twice sounds like a thing that shouldn't be done, but the tarfile is a stream like object so we can't seek back after iterating, and because of this calling it twice seemed like the only way to guarantee that no `.listens` file was skipped. Would be happy to know if I've missed anything and if there is a better way.

listenbrainz/listenstore/tests/test_influxlistenstore.py
```diff
@@ -160,10 +163,160 @@ def test_import_listens(self):
...
+        sleep(1)
+        self.assertTrue(os.path.isfile(dump_location))
+        self.reset_influx_db()
+
+        done = self.logstore.import_listens_dump(dump_location)
+        sleep(1)
+        self.assertEqual(done, len(db_user.get_all_users()))
+
+
+    def create_test_dump(self, archive_name, archive_path, schema_version=None, index=None):
```
mayhem: Why do you have separate code for this here? For more thorough checking?

listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -396,10 +406,15 @@ def dump_listens(self, location, dump_time=datetime.today(), threads=None):
                 # a temporary dir before adding them to the archive
                 for user in users:
                     username = user['musicbrainz_id']
+                    index = user['id']
                     offset = 0
 
-                    user_listens_file = '{username}.listens'.format(username=username)
-                    user_listens_path = os.path.join(listens_path, user_listens_file)
+                    directory = os.path.join(listens_path, '%03d' % (index % NUMBER_OF_USERS_PER_DIRECTORY))
```
mayhem: This will create up to 1,000 top level directories for the first 1,000 users. Also not ideal. I think the path should be: %01d/%02d/%03d , but that might be overkill....

@alastair thoughts?

listenbrainz/listenstore/tests/test_influxlistenstore.py
paramsingh: Yeah, this method is used to create dumps without `index.json` or `SCHEMA_SEQUENCE` to make sure that `SchemaMismatchException` is raised for dumps without these files or with incorrect schema versions.

mayhem:
Besides the error I posted above, I've got the following comments about data dumps: While the dump files seem to have good contents, I'm a little confused when importing data. There are the listens in -listens, which is clear, but then there is the -private data and -stats data. This makes sense our perspective, but from the perspective of a user, where does the user data come from? It seems that there is a data dump file missing, unless that data comes from the -stats file, which is.. misleading. Thoughts?

paramsingh:
You're right, now that I think about it, the name `stats` is a bit misleading if it contains user data (sanitized) as well. A possible solution would be to rename the `stats` dump to `public`?

paramsingh:
`db.engine.raw_connection()` created a connection and a transaction due to which it wasn't possible to change autocommit status (i think this was introduced when we upgraded psycopg2, not sure)

mayhem:
stats -> public makes sense, yes.

PR #331, opened 2017-12-26
LB-259: Begin writing tests for BigQueryWriter

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: refactor BigQueryWriter a bit and write tests

# Problem

* JIRA ticket (_optional_): [LB-259](https://tickets.metabrainz.org/browse/LB-259)

# Solution

I moved the code that converted a listen into BigQuery format into a different function and
added a test for that. Still thinking about the best way to add unit tests for the callback functions.

Also, in order to import BigQueryWriter into the test module, I had to rename the module from
`bigquery-writer` to `bigquery_writer` (did the same for `influx-writer` too).


paramsingh:
depends on #343

PR #333, opened 2018-01-03
LB-198: Use brainzutils for caching instead of directly using redis

kartikeyaSh:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
brainzutils cache has code which can be used for caching instead of using redis directly.

* JIRA ticket (_optional_): [LB-198](https://tickets.metabrainz.org/browse/LB-198)

# Solution
Start using brainzutils instead of redis

### DIFF
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -49,7 +49,7 @@ class InfluxListenStore(ListenStore):
         ListenStore.__init__(self, conf)
         self.influx = InfluxDBClient(host=conf['INFLUX_HOST'], port=conf['INFLUX_PORT'], database=conf['INFLUX_DB_NAME'])
         # Initialize brainzutils cache
-        init_cache()
+        init_cache(host=conf['REDIS_HOST'], port=conf['REDIS_PORT'], namespace=conf['REDIS_NAMESPACE'])
 
     def get_listen_count_for_user(self, user_name, need_exact=False):
         """Get the total number of listens for a user. The number of listens comes from
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -10,11 +10,11 @@ import shutil
 import ujson
 import uuid
 
+from brainzutils import cache
 from collections import defaultdict
 from datetime import datetime
 from influxdb import InfluxDBClient
 from influxdb.exceptions import InfluxDBClientError, InfluxDBServerError
-from redis import Redis
 
 from listenbrainz import DUMP_LICENSE_FILE_PATH
 from listenbrainz.db import DUMP_DEFAULT_THREAD_COUNT
@@ -47,13 +47,11 @@ class InfluxListenStore(ListenStore):
 
     def __init__(self, conf):
         ListenStore.__init__(self, conf)
-        self.redis = Redis(host=conf['REDIS_HOST'], port=conf['REDIS_PORT'], decode_responses=True)
-        self.redis.ping()
         self.influx = InfluxDBClient(host=conf['INFLUX_HOST'], port=conf['INFLUX_PORT'], database=conf['INFLUX_DB_NAME'])
 
     def get_listen_count_for_user(self, user_name, need_exact=False):
         """Get the total number of listens for a user. The number of listens comes from
-           a redis cache unless an exact number is asked for.
+           brainzutils cache unless an exact number is asked for.
 
         Args:
             user_name: the user to get listens for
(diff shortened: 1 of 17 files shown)


kartikeyaSh:
In travis logs line 805 we have: 
"Collecting git+https://github.com/metabrainz/brainzutils-python.git@v1.3.0 (from -r requirements.txt (line 1))" 
in commit "Use version of brainzutils cache that supports increment" I'm trying to use the version of brainzutils with latest commit as I need increment method.
and also I found that in log line 1879  in "listenbrainz/webserver/views/user.py" InfluxDBClientError, InfluxDBServerError has not been imported

paramsingh:
Tests still seem to fail.

kartikeyaSh:
when trying to install brainzutils with latest commit i found that during `pip3 install requirements.txt` we get `Requirement already satisfied (use --upgrade to upgrade): brainzutils==1.1.0` so in actual it does not get installed to the latest commit.
In brainzutils before creating a release the version should have been changed. we can add something like `RUN pip3 install --upgrade --force-reinstall git+https://github.com/metabrainz/brainzutils-python.git@18229350b764201e966fb9d89784d4e2258a95c6` to the Dockerfile to make it work until we get the new version of brainzutils released.

paramsingh:
I guess we'd want to merge this after I release a new version of brainzutils, let me do that and then I'll review this.

paramsingh:
https://github.com/metabrainz/brainzutils-python/releases/tag/v1.4.0

listenbrainz/__init__.py
```diff
@@ -4,3 +4,21 @@
...
+
+from listenbrainz import default_config as config
+try:
+    from listenbrainz import custom_config as config
+except ImportError:
+    pass
+from brainzutils import cache
+import logging
+
+try:
```
paramsingh: doing this here is weird, i think adding a function called `init_cache` in `listenbrainz.utils` and calling it individually in `webserver/__init__.py` (and other places) would be better.

listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -41,23 +41,21 @@ class InfluxListenStore(ListenStore):
...
             user_name: the user to get listens for
             need_exact: if True, get an exact number of listens directly from the ListenStore
         """
 
         if not need_exact:
-            # check if the user's listen count is already in redis
+            # check if the user's listen count is already in cache
             # if already present return it directly instead of calculating it again
-            count = self.redis.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_name)
+            count = cache.get(REDIS_INFLUX_USER_LISTEN_COUNT + user_name, decode=False)
```
paramsingh: I think a comment should be added documenting why `decode=False` is needed here now.

listenbrainz/test_config.py
```diff
@@ -23,6 +23,7 @@
 # Redis
 REDIS_HOST = "redis"
 REDIS_PORT = 6379
+REDIS_NAMESPACE = "listenbrainz"
```
paramsingh: Adding the config changes to the consul config template `consul_config.py.ctmpl` would be helpful too.

listenbrainz/webserver/__init__.py
```diff
@@ -43,8 +43,10 @@ def gen_app(config_path=None, debug=None):
     app = CustomFlask(
         import_name=__name__,
         use_flask_uuid=True,
-        use_debug_toolbar=True,
     )
+    # CustomFlask from brainzutils version 1.4.0 onwards requires calling a seperate
+    # method to initialize debug toolbar
+    app.init_debug_toolbar()
```
paramsingh: this was fixed in #364

listenbrainz/utils.py
```diff
@@ -118,3 +118,25 @@ def connect_to_rabbitmq(username, password,
             error_message = "Cannot connect to RabbitMQ: {error}, retrying in {delay} seconds."
             error_logger(error_message.format(error=str(err), delay=error_retry_delay))
             time.sleep(error_retry_delay)
+
+
+def init_cache():
+    """ Initializes brainzutils cache. """
+    from brainzutils import cache
+    from listenbrainz import default_config as config
```
paramsingh: These imports are a bit ugly (and I plan to remove all such imports soon, using flask app config values), I'd prefer it if init_cache would take args host, port etc. That would also mean we don't need to import logging here.

listenbrainz/webserver/__init__.py
```diff
@@ -43,7 +43,6 @@ def gen_app(config_path=None, debug=None):
     app = CustomFlask(
         import_name=__name__,
         use_flask_uuid=True,
-        use_debug_toolbar=True,
```
paramsingh: This has been fixed in latest master too (changes like these become a bit distracting), please rebase once over master.

PR #334, opened 2018-01-12
Create list of listen submitters and add libra.

kori:
# Summary

* This is a…
    * [] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ x] Other
* **Describe this change in 1-2 sentences**: This adds https://github.com/kori/libra to the list of Listen Submitters.

# Problem

I developed a listen submitter for ListenBrainz and I'd like to add it to the list so people will know it exists.

# Solution

The program was added to the list.

# Action

Please review if the procedure was executed correctly

### DIFF
--- a/listenbrainz/webserver/templates/index/api-docs.html
+++ b/listenbrainz/webserver/templates/index/api-docs.html
@@ -15,7 +15,7 @@
   <h2>Listen Submitters</h2>
     These are programs that submit listens to ListenBrainz.
   <ul>
-    <li><em>mpd</em>: <a href="https://github.com/kori/libra"><code>libra</code></a></li>
+    <li><em><a href="https://www.musicpd.org/">mpd</a></em>: <a href="https://github.com/kori/libra"><code>libra</code></a></li>
   </ul>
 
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/templates/index/api-docs.html
```diff
@@ -12,4 +12,10 @@ <h2>Client Libraries</h2>
     <li><em>Haskell</em>: <a href="http://hackage.haskell.org/package/listenbrainz-client"><code>listenbrainz-client</code></a></li>
   </ul>
 
+  <h2>Listen Submitters</h2>
+    These are programs that submit listens to ListenBrainz.
+  <ul>
+    <li><em>mpd</em>: <a href="https://github.com/kori/libra"><code>libra</code></a></li>
```
mayhem: Might be good to add a link to mpd here so that readers and learn about it.

kori: Alright, I did that in commit 65e1c2a5de99b16b36f8be419ccde07a8c514b08

PR #337, opened 2018-01-20
Change Import dumps interface

paramsingh:
As discussed on IRC, I've changed the manage.py import command to take paths to the archives directly instead of taking a directory as input.

Based on #326, will rebase after that gets merged.

### DIFF
--- a/listenbrainz/db/dump_manager.py
+++ b/listenbrainz/db/dump_manager.py
@@ -69,37 +69,31 @@ def create(location, threads):
 
 
 @cli.command()
-@click.option('--location', '-l', default=os.path.join(os.getcwd(), 'listenbrainz-export'))
+@click.option('--private-archive', '-pr', default=None)
+@click.option('--public-archive', '-pu', default=None)
+@click.option('--listen-archive', '-l', default=None)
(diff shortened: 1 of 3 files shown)


listenbrainz/db/dump_manager.py
```diff
@@ -69,38 +69,32 @@ def create(location, threads):
...
-            threads (int): the number of threads to use while decompressing, defaults to 1
+            private_archive (str): the path to the ListenBrainz private dump to be imported
+            public_archive (str): the path to the ListenBrainz public dump to be imported
+            listen_archive (str): the path to the ListenBrainz listen dump archive to be imported
+            threads (int): the number of threads to use during decompression, defaults to 1
     """
-
-    if not location:
-        print('No location given!')
+    if not private_archive and not public_archive and not listen_archive:
```
mayhem: I think this is too simplistic. It doesn't make sense to import public and private at the same time. Should we allow importing listens without users/stats? I can't think of a compelling reason why not....

paramsingh: Opened [LB-289](https://tickets.metabrainz.org/browse/LB-289) for now.

PR #340, opened 2018-01-28
LB-285: Improve logging in listens dumps for user

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Improved logging so that we know what speed listens are being dumped at.

# Problem

* JIRA ticket (_optional_):  [LB-285](https://tickets.metabrainz.org/browse/LB-285)

### DIFF
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -347,8 +347,6 @@ class InfluxListenStore(ListenStore):
         Returns:
             int: the number of bytes this user's listens take in the dump file
         """
-        self.log.info('Dumping user %s...', username)
-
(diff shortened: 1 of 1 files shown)


listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -348,8 +348,12 @@ def dump_user(self, username, fileobj, dump_time):
             int: the number of bytes this user's listens take in the dump file
         """
         self.log.info('Dumping user %s...', username)
```
mayhem: I think this should go away, its noise in the grand scheme of things.

PR #349, opened 2018-02-04
LB-303: Add instruction to add last.fm api key to custom_config.py to developement environment setup guide

Uditgulati:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [X] Other

There was no instruction given for adding last.fm api to custom_config.py file in the development environment setup guide.

# Problem

* JIRA ticket (_optional_): [LB-303](https://tickets.metabrainz.org/browse/LB-303)

# Solution

Added the instructions to the guide.

### DIFF
--- a/docs/dev/devel-env.rst
+++ b/docs/dev/devel-env.rst
@@ -111,7 +111,7 @@ Update the strings with your client ID and secret. After doing this, your
 ListenBrainz development environment is able to authenticate and log in from
 your MusicBrainz login.
 
-Also, in order for last.fm import to work, you should also update your last.fm api key in this file. Look for the following section in the file.
+Also, in order for Last.FM import to work, you should also update your Last.FM api key in this file. Look for the following section in the file.
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


docs/dev/devel-env.rst
```diff
@@ -111,6 +111,17 @@ Update the strings with your client ID and secret. After doing this, your
...
 
+Also, in order for last.fm import to work, you should also update your last.fm api key in this file. Look for the following section in the file.
+
+.. code-block:: yaml
+
+    # Lastfm API
+    LASTFM_API_URL = "https://ws.audioscrobbler.com/2.0/"
+    LASTFM_API_KEY = "USE_LASTFM_API_KEY"
+
+Update the last.fm api key with your key. After doing this, your
```
paramsingh: I think we should add a explanation on where to get this key from.

docs/dev/devel-env.rst
```diff
@@ -111,6 +111,17 @@ Update the strings with your client ID and secret. After doing this, your
 ListenBrainz development environment is able to authenticate and log in from
 your MusicBrainz login.
 
+Also, in order for last.fm import to work, you should also update your last.fm api key in this file. Look for the following section in the file.
```
paramsingh: We try to write `Last.FM` instead of `last.fm` in documentation.

docs/dev/devel-env.rst
Uditgulati: Just adding link to API form would be enough or should i add something else too?

docs/dev/devel-env.rst
Uditgulati: Ok :+1:

docs/dev/devel-env.rst
paramsingh: You can obtain a Last.FM API key from here: <link>

paramsingh:
@mayhem, you should take a look at the verbiage here once, IMO.

mayhem:
It doesn't flow super well, but really, there is nothing here stopping this from being merged.

PR #352, opened 2018-02-07
LB-305: Improve parameter checking for timestamps

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-305](https://tickets.metabrainz.org/browse/LB-305) 

# Solution

If the timestamp is greater than the time at our server, the listen should be invalid.

Also, added tests.

### DIFF
--- a/listenbrainz/tests/integration/test_api.py
+++ b/listenbrainz/tests/integration/test_api.py
@@ -297,6 +297,7 @@ class APITestCase(IntegrationTestCase):
 
         with open(self.path_to_data_file('timestamp_in_ns.json'), 'r') as f:
             payload = json.load(f)
+        payload['listened_at'] = int(time.time()) * 10**9
         response = self.send_data(payload)
         self.assert400(response)
         self.assertEqual(response.json['code'], 400)
--- a/listenbrainz/tests/integration/test_api.py
+++ b/listenbrainz/tests/integration/test_api.py
@@ -291,3 +291,13 @@ class APITestCase(IntegrationTestCase):
         self.assert400(response)
         self.assertEqual(response.json['code'], 400)
         self.assertEqual('artist_name must be a single string.', response.json['error'])
+
+    def test_too_high_timestamps(self):
+        """ Tests for timestamps greater than current time """
+
+        with open(self.path_to_data_file('timestamp_in_ns.json'), 'r') as f:
+            payload = json.load(f)
+        response = self.send_data(payload)
+        self.assert400(response)
+        self.assertEqual(response.json['code'], 400)
+        self.assertEqual('Value for key listened_at is too high.', response.json['error'])
--- a/listenbrainz/webserver/views/api_compat_deprecated.py
+++ b/listenbrainz/webserver/views/api_compat_deprecated.py
@@ -32,8 +32,7 @@ from time import time
 from flask import current_app, Blueprint, request, render_template, redirect
 from werkzeug.exceptions import BadRequest
 from listenbrainz.db.lastfm_session import Session
-from listenbrainz.webserver import API_LISTENED_AT_ALLOWED_SKEW
-from listenbrainz.webserver.views.api_tools import insert_payload, LISTEN_TYPE_PLAYING_NOW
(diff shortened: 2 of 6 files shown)


paramsingh:
Based this off of production because it is needed there too. We could do a merge into master after this gets merged?

listenbrainz/webserver/views/api_tools.py
```diff
@@ -104,6 +105,9 @@ def validate_listen(listen, listen_type):
             log_raise_400("JSON document may only contain listened_at and "
                            "track_metadata top level keys", listen)
 
+        if listen['listened_at'] > int(time.time()):
```
alastair: perhaps it's a good idea to consider a bit of clock skew. I'd consider allowing anything from 10 minutes to 1h in advance

listenbrainz/tests/integration/test_api.py
```diff
@@ -291,3 +291,14 @@ def test_multiple_artist_names(self):
         self.assert400(response)
         self.assertEqual(response.json['code'], 400)
         self.assertEqual('artist_name must be a single string.', response.json['error'])
+
+    def test_too_high_timestamps(self):
```
mayhem: https://www.youtube.com/watch?v=kcsNbQRU5TI

listenbrainz/webserver/views/api_compat_deprecated.py
```diff
@@ -156,8 +156,14 @@ def _to_native_api(data, append_key):
...
-            listen['listened_at'] = data['i{}'.format(append_key)]
-        except KeyError:
+            listen['listened_at'] = int(data['i{}'.format(append_key)])
+        except (KeyError, ValueError):
+            return None
+
+        # if timestamp is too high, this is an invalid listen
+        # in order to make up for possible clock skew, we allow
+        # timestamps to be one hour ahead of server time
+        if listen['listened_at'] > int(time()) + 86400:
```
mayhem: Now we have two timestamp upper limit checkers, with different values. Why is this one 24 hours? I think we should have one function for this.

listenbrainz/webserver/views/api_tools.py
```diff
@@ -104,6 +105,12 @@ def validate_listen(listen, listen_type):
             log_raise_400("JSON document may only contain listened_at and "
                            "track_metadata top level keys", listen)
 
+        # if timestamp is too high, raise BadRequest
+        # in order to make up for possible clock skew, we allow
+        # timestamps to be one hour ahead of server time
+        if listen['listened_at'] > int(time.time()) + 60 * 60:
```
mayhem: Here is the other instance, at 1 hour.

listenbrainz/webserver/views/api_compat_deprecated.py
paramsingh: whoops.

listenbrainz/webserver/views/api_tools.py
```diff
@@ -104,6 +106,12 @@ def validate_listen(listen, listen_type):
             log_raise_400("JSON document may only contain listened_at and "
                            "track_metadata top level keys", listen)
 
+        # if timestamp is too high, raise BadRequest
+        # in order to make up for possible clock skew, we allow
+        # timestamps to be one hour ahead of server time
+        if listen['listened_at'] > int(time.time()) + API_LISTENED_AT_ALLOWED_SKEW:
```
mayhem: Sorry, I should've seen this earlier, but this allows for only a clock skew for the users clock to be FAST, not SLOW. I think we should test for 30 minutes fast and or 30 minutes slow.

paramsingh: If I do  `current_time - API_LISTENED_AT_ALLOWED_SKEW <= ts <= currentt_time + API_LISTENED_AT_ALLOWED_SKEW`, it wouldn't work for old listens leading to failing imports.

mayhem: Sorry, can you please elaborate? I don't understand what you're saying.

PR #353, opened 2018-02-11
LB-308: API compat fails to record listens

kartikeyaSh:
* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
At present api_compat is broken as it fails to record listens.

* JIRA ticket (_optional_): 

# Solution
In insert_payload the user argument should be of type LB user not a Last.FM user

# Action
Get the user using sessions userid and pass this user into insert_payload.

### DIFF
--- a/listenbrainz/webserver/views/api_compat.py
+++ b/listenbrainz/webserver/views/api_compat.py
@@ -2,6 +2,7 @@
 import time
 import json
 import re
+import listenbrainz.db.user as db_user
 from collections import defaultdict
 from yattag import Doc
 import yattag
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/api_compat.py
```diff
@@ -2,6 +2,7 @@
 import time
 import json
 import re
+import listenbrainz.db.user as db_user
```
paramsingh: In general, if you find unalphabetized imports, you should alphabetize them. Fine for now though.

PR #354, opened 2018-02-11
LB-309: API compat record_listens response does not include accepted attribute

kartikeyaSh:
# Summary
In the XML response of record_listens the scrobbles tag should contain
number of accepted listens and number of ignored listens. As original
Last.FM API response for the same contains these fields.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Last.fm does not accept listens which have bad metadata e.g. artist = "unknown artist" and the method track.scrobble in its response sends the number of accepted listens and number of ignored listens.

<scrobbles accepted="1" ignored="0">

Whereas LB currently accepts all the listens provided to it (if mandatory fields are filled), so such fields are not really important for LB responses but as this API is to work with clients which anticipate that the response will be a Last.FM type of response. And in that case these fields may be used by clients to display number of accepted listens etc.

* JIRA ticket (_optional_): 

# Solution
Add accepted and ignored fields to the response.

### DIFF
--- a/listenbrainz/tests/integration/test_api_compat.py
+++ b/listenbrainz/tests/integration/test_api_compat.py
@@ -124,3 +124,59 @@ class APICompatTestCase(APICompatIntegrationTestCase):
         response = xmltodict.parse(r.data)
         self.assertEqual(response['lfm']['@status'], 'failed')
         self.assertEqual(response['lfm']['error']['@code'], '4')
+
+    def test_record_listen(self):
+        """ Tests if listen is recorded correctly if valid information is provided. """
+
+        token = Token.generate(self.lfm_user.api_key)
+        token.approve(self.lfm_user.name)
+        session = Session.create(token)
+
+        data = {
+            'method': 'track.scrobble',
+            'api_key': self.lfm_user.api_key,
+            'sk': session.sid,
+            'artist[0]': 'Kishore Kumar',
+            'track[0]': 'Saamne Ye Kaun Aya',
+            'album[0]': 'Jawani Diwani',
+            'duration[0]': 300,
+            'timestamp[0]': int(time.time()),
+        }
+
+        r = self.client.post(url_for('api_compat.api_methods'), data=data)
+        self.assert200(r)
+
+        response = xmltodict.parse(r.data)
+        self.assertEqual(response['lfm']['@status'], 'ok')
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/api_compat.py
```diff
@@ -272,47 +272,61 @@ def record_listens(request, data):
...
+            # So ignored will be always '0' which is not the case with Last.FM.
+            accepted = str(len(lookup))
+            with tag('scrobbles', accepted=accepted, ignored='0'):
+                text(create_response_for_recorded_listens(lookup, augmented_listens))
 
     return format_response('<?xml version="1.0" encoding="utf-8"?>\n' + yattag.indent(doc.getvalue()),
                            output_format)
 
 
+def create_response_for_recorded_listens(lookup, augmented_listens):
```
paramsingh: A docstring explaining what this function does would be nice to have. The names `lookup` and `augmented_listens` also don't explain much.

listenbrainz/webserver/views/api_compat.py
```diff
@@ -272,47 +272,61 @@ def record_listens(request, data):
...
+    doc, tag, text = Doc().tagtext()
+
+    for origL, augL in zip(list(lookup.values()), augmented_listens):
```
paramsingh: using `original_listen` and `augmented_listen` here would look better to me.

paramsingh:
I'd like to see unit tests for `create_response_for_recorded_listen` too.

listenbrainz/tests/integration/test_api_compat.py
```diff
@@ -124,3 +124,155 @@ def test_get_session_invalid_token(self):
...
+            'duration[1]': 200,
+            'timestamp[1]': int(time.time())+300,
+        }
+
+        r = self.client.post(url_for('api_compat.api_methods'), data=data)
+        self.assert200(r)
+
+        response = xmltodict.parse(r.data)
+        self.assertEqual(response['lfm']['@status'], 'ok')
+        self.assertEqual(response['lfm']['scrobbles']['@accepted'], '2')
```
paramsingh: We should probably check that the listen has reached the influx listenstore here as well.

PR #356, opened 2018-02-14
Add script to backup ListenBrainz dumps

paramsingh:
Also create a listenbrainz user (similar to MusicBrainz)
which will be needed later as we add ssh keys to rsync
the public dump into the FTP server.

We should now create a volume in prod to copy the dumps into
the host outside the container

### DIFF
--- a/admin/RunExport
+++ b/admin/RunExport
@@ -1,0 +1,54 @@
+#!/bin/bash
+
+# listenbrainz-server - Server for the ListenBrainz project.
+#
+# Copyright (C) 2018 MetaBrainz Foundation Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+echo "This script is being run by the following user: "; whoami
+
+# This is to help with disk space monitoring - run "df" before and after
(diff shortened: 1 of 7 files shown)


admin/RunExport
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
mayhem: The name of the script is my fault (from 17 years ago) -- let's not copy that bad name and go with something more sensical like create-dumps.sh or something...

docker/crontab
```diff
@@ -1,9 +1,9 @@
...
+PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin:/code/listenbrainz/admin
 PYTHONPATH=/usr/local/lib/python36.zip:/usr/local/lib/python3.6:/usr/local/lib/python3.6/lib-dynload:/usr/local/lib/python3.6/site-packages:/usr/local/lib/python3.6/site-packages/messybrainz-99.0.0.dev0-py3.6.egg
 GOOGLE_APPLICATION_CREDENTIALS=/code/listenbrainz/credentials/bigquery-credentials.json
 
 # Statistics calculation via Google BigQuery
 0 0 * * 0 /usr/local/bin/python /code/listenbrainz/manage.py stats calculate >> /var/log/stats.log 2>&1
 
 # Data dumps creation for backup and public use
-15 0 1 * * /usr/local/bin/python /code/listenbrainz/manage.py dump create >> /var/log/dump_create.log 2>&1
+15 0 1 * * RunExport >> /var/log/dump_create.log 2>&1
```
mayhem: For the sake of being explicit, which is a good thing in cron jobs, I think we should use exact paths to the binaries. if that allows us to get rid of the PATH statement, even better.

docker/crontab
```diff
@@ -6,4 +6,4 @@ GOOGLE_APPLICATION_CREDENTIALS=/code/listenbrainz/credentials/bigquery-credentia
 0 0 * * 0 /usr/local/bin/python /code/listenbrainz/manage.py stats calculate >> /var/log/stats.log 2>&1
 
 # Data dumps creation for backup and public use
-15 0 1 * * /usr/local/bin/python /code/listenbrainz/manage.py dump create >> /var/log/dump_create.log 2>&1
+* * * * * /code/listenbrainz/admin/create-dumps.sh >> /var/log/dump_create.log 2>&1
```
mayhem: Whooops, we don't want to dump every minute of every day. :) copypasta!

PR #357, opened 2018-02-14
LB-294: user.py now returns artist_count as None if not calculated

Uditgulati:
# Summary

If a user's artist count hasn't been calculated, we shouldn't show anything. 
Right now, we display artist count = 0 instead.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-294](https://tickets.metabrainz.org/browse/LB-294)

# Solution

Add an if condition to ``user.py`` to return ``None`` if ``artist_count`` isn't calculated.

### DIFF
--- a/listenbrainz/webserver/views/user.py
+++ b/listenbrainz/webserver/views/user.py
@@ -123,7 +123,7 @@ def profile(user_name):
     try:
         artist_count = int(user_stats['artist']['count'])
     except (KeyError, TypeError):
-        artist_count = 0
+        artist_count = None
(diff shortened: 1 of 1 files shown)


paramsingh:
I think adding a test for this might have been a good idea, @uditgulati maybe do that in a new PR?

Uditgulati:
@paramsingh  on it 👍

PR #359, opened 2018-02-17
rsync data dumps to williams

paramsingh:
with some small permission fixes and installing rsync in the container.


paramsingh:
Came across some permission problems after syncing to willliams that I need to fix.

admin/config.sh
```diff
@@ -10,3 +10,9 @@ BACKUP_USER=listenbrainz
 BACKUP_GROUP=listenbrainz
 BACKUP_DIR_MODE=700
 BACKUP_FILE_MODE=600
+
+# rsync to FTP server configuration
+RSYNC_FULLEXPORT_HOST='10.2.2.28'
```
mayhem: Hard coded IP? We can't do that -- we need to get IP addresses from consul, so this file will also need to be generated by consul.

admin/rsync-dump-files.sh
```diff
@@ -0,0 +1,36 @@
...
+LB_SERVER_ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../" && pwd)
+cd "$LB_SERVER_ROOT"
+
+source admin/config.sh
+source admin/functions.sh
+
+retry rsync \
+    --archive \
+    --delete \
+    --exclude='listenbrainz-private-dump*' \
```
mayhem: As mentioned before, there is a small chance of exposing a new private file in the future using this scheme. I think naming files individually is the way to go here.

admin/config.sh
paramsingh: Alright, will fix.

admin/create-dumps.sh
```diff
@@ -47,8 +52,45 @@ chown "$BACKUP_USER:$BACKUP_GROUP" \
...
+# and set appropriate mode for files to be uploaded to
+# the FTP server
+chown "$FTP_USER:$FTP_GROUP" "$DUMP_DIR"/*
+chmod "$FTP_FILE_MODE" "$DUMP_DIR"/*
+retry cp -a "$DUMP_DIR"/* "$FTP_CURRENT_DUMP_DIR"
+
+# create an explicit rsync filter for the new private dump in the
+# ftp folder
+touch "$FTP_CURRENT_DUMP_DIR"/.rsync-filter
+PRIVATE_DUMP_RULE=`printf "exclude listenbrainz-private-dump-%s.tar.xz" $DUMP_TIMESTAMP`
```
mayhem: This is OK, but I would really prefer to see an explicit rule to copy the public files to the public FTP, rather than an explicit rule to exclude the private dump. If something with the exclude private rule goes wrong, private data gets exposed. If something goes with with the include public rule, nothing gets copied. I know which error I prefer to have.

paramsingh: Very reasonable, makes sense. I'll do it.

PR #360, opened 2018-02-20
LB-307: Create a new RabbitMQ connection correctly when acquiring from pool

paramsingh:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-307](https://tickets.metabrainz.org/browse/LB-307) 

Last.FM imports failed with logs showing that rabbitmq connections got closed due to "second
channel.open seen"

# Solution

This caused the bug that led to failed imports, as pika pool
called more than one channel.open for a single channel,
leading to RabbitMQ closing the connection after one
batch of listens were sent. Now the pool should correctly
create a connection when needed.

Also, make the docstring clearer.

### DIFF
--- a/listenbrainz/webserver/rabbitmq_connection.py
+++ b/listenbrainz/webserver/rabbitmq_connection.py
@@ -7,7 +7,11 @@ import listenbrainz.utils as utils
 _rabbitmq = None
 
 def init_rabbitmq_connection(app):
-    """Create a connection to the RabbitMQ server."""
+    """Initialize the webserver rabbitmq connection.
+
+    This initializes _rabbitmq as a connection pool from which new RabbitMQ
+    connections can be acquired.
+    """
     global _rabbitmq
 
     if "RABBITMQ_HOST" not in app.config:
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/rabbitmq_connection.py
```diff
@@ -7,28 +7,27 @@
...
+    connection_parameters = pika.ConnectionParameters(
+        host=app.config['RABBITMQ_HOST'],
+        port=app.config['RABBITMQ_PORT'],
+        virtual_host=app.config['RABBITMQ_VHOST'],
+        credentials=pika.PlainCredentials(app.config['RABBITMQ_USERNAME'], app.config['RABBITMQ_PASSWORD']),
+    )
 
     _rabbitmq = pika_pool.QueuedPool(
-        create=lambda: connection,
+        create=lambda: pika.BlockingConnection(connection_parameters),
```
paramsingh: There is no need (or value) of looping here because connections are acquired here (https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/views/api_tools.py#L72), so the loop is unneccessary.

paramsingh:
There are tests that should have caught this but I'm not sure why they didn't, this one (https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/tests/integration/test_influx_writer.py#L106) in particular.

Anyways, opened [LB-311](https://tickets.metabrainz.org/browse/LB-311) for testing the last.fm import. Maybe selenium or something is needed here, I'm not sure.

PR #362, opened 2018-02-22
LB:316: replace ERROR_RETRY_DELAY with self.ERROR_RETRY_DELAY in influx_writer.py

Uditgulati:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

influx_writer.py checks in a number of places for configuration items, and if they don't exist it tries to sleep and then try again, however it sometime tries to sleep for ERROR_RETRY_DELAY seconds instead of self.ERROR_RETRY_DELAY

* JIRA ticket (_optional_): [LB-316](https://tickets.metabrainz.org/browse/LB-316)

### DIFF
--- a/listenbrainz/influx_writer/influx_writer.py
+++ b/listenbrainz/influx_writer/influx_writer.py
@@ -75,10 +75,10 @@ class InfluxWriterSubscriber(ListenWriter):
                 failure_count += 1
                 if failure_count >= retries:
                     break
-                sleep(ERROR_RETRY_DELAY)
+                sleep(self.ERROR_RETRY_DELAY)
             except ConnectionError as e:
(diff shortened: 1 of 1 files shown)


alastair:
the ticket also has a comment about `DUMP_JSON_WITH_ERRORS` too, will you do that in the same PR or another one?

@paramsingh any ideas on how we could test this? There are enough python tools to hack the time module to make sleeps not happen. You can use mock to make a constructor raise an exception once and succeed the second time.

Uditgulati:
@alastair  didn't see the comment earlier. I'll add it in this PR.

mayhem:
I'm confused by this -- I don't see either where this is defined or imported. What am I missing?

PR #363, opened 2018-02-22
LB-315: import sys in listen_writer.py

Uditgulati:
# Summary

listen_writer.py is missing sys import.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-315](https://tickets.metabrainz.org/browse/LB-315)

### DIFF
--- a/listenbrainz/tests/unit/test_listen_writer.py
+++ b/listenbrainz/tests/unit/test_listen_writer.py
@@ -1,0 +1,34 @@
+import unittest
+from listenbrainz.listen_writer import ListenWriter
+
+class ListenWriterTestCase(unittest.TestCase):
+
+    def setUp(self):
+        self.lwriter = ListenWriter()
+
+
+    def test_verify_hosts_in_config(self):
+        """ Test for the _verify_hosts_in_config method """
+
+        config1 = self.lwriter.config
+        config2 = self.lwriter.config
+
+        if hasattr(config1, "REDIS_HOST"):
+            delattr(config1, "REDIS_HOST")
+
+        self.lwriter.config = config1
+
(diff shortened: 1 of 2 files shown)


mayhem:
What is it you want to test here, @alastair alastairp? That sys.exit works?

alastair:
I want to test that if the config file doesn't contain the redis configuration options that we actually do what the code says - that is, we sleep and then quit. I guess it's our luck that we never hit this line of code - and in fact the behaviour of `sys` being undefined is the same as the intent (the script will raise an exception and quit)

I don't know the best way to test this. It appears that you can catch `SystemExit`, which we could do in a test to verify that we hit this code path when redis conf doesn't exist

alastair:
This is better, but we should mock out `time.sleep` so that the test doesn't take 6 seconds to run
https://stackoverflow.com/questions/22836874/how-to-stub-time-sleep-in-python-unit-testing

listenbrainz/tests/unit/test_listen_writer.py
```diff
@@ -0,0 +1,36 @@
...
+            delattr(config1, "REDIS_HOST")
+
+        self.lwriter.config = config1
+
+        with self.assertRaises(SystemExit) as s:
+            self.lwriter._verify_hosts_in_config()
+
+        self.assertEqual(s.exception.code, -1)
+
+        if hasattr(config2, "REDISMQ_HOST"):
```
paramsingh: REDISMQ?

Uditgulati: A typo :(
I'll fix it. Thanks :+1:

PR #366, opened 2018-02-26
Tune crontab to be easy to grasp for humans

mayhem:
Run stats every monday at midnight and run dumps every 1st and 15th of the month at 4am. I'm picking monday over sunday, since sunday is a high traffic time and monday is low.

The midnight/4am staggering should hopefully prevent both of these tasks from running at the same time.

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

We hadn't yet really thought about when to run the cron jobs. I think this makes sense.

* JIRA ticket (_optional_): 

# Solution

# Action


mayhem:
Easy for humans and hopefully non-conflicting for our servers.

PR #367, opened 2018-02-26
Merge production into master before beta release

paramsingh:


paramsingh:
Yep, nothing other than that.

PR #368, opened 2018-02-27
LB-322: Add import and sign in buttons to error pages

Uditgulati:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-322](https://tickets.metabrainz.org/browse/LB-322)

Sign in and import buttons are not shown on error pages.

# Solution

If it is an error page and user is not signed in, then show import and sign in buttons on that page.


paramsingh:
Please write a test for this too?

Uditgulati:
Currently, drop-down menu isn't shown on error pages if user is signed in. But, we can show it on 40x pages and show only 'sign in' button or nothing on 500 page. Showing nothing on 500 page is a better choice IMHO.

Uditgulati:
@paramsingh i''ll add a commit for tests once the changes are finalized.

paramsingh:
> Currently, drop-down menu isn't shown on error pages if user is signed in. But, we can show it on 40x pages and show only 'sign in' button or nothing on 500 page. Showing nothing on 500 page is a better choice IMHO.

This is what we're doing in AB, should be okay here too.

Uditgulati:
Now, if error code is not 500, buttons are unchanged. If code is 500, then ``sign in`` button is shown if not signed in, else it shows nothing.

listenbrainz/webserver/templates/navbar.html
```diff
@@ -46,6 +50,16 @@
             </li>
           {% endif %}
         </ul>
+      {% else %}
```
mayhem: I'm not really a fan of this block of HTML being duplicated. Can you re-work the IF clause to reduce the duplication?

Uditgulati:
Oh, oops. Actually, it was ``500`` instead of ``404``. In ``500``, the page isn't able to render if drop-down menu is shown. 
While cross-checking the if claus, I replaced ``500`` with ``404`` and forgot to revert it back before pushing. I'll fix it.

alastair:
The core idea here is that we can't do any database queries when showing an error page that may have come about due to an error with the database. We require a db query to get information about the current logged in user and show their name in the top menu if they are signed in. This restriction generally applies to 500 errors. We had a case in AB where we saw this causing issues: https://tickets.metabrainz.org/projects/AB/issues/AB-74

We have a similar PR open on AB (https://github.com/metabrainz/acousticbrainz-server/pull/258), and we changed ideas a bit, arriving to the decision that we could probably do this DB query (and therefore show the user menu) on error pages that aren't 500. I think that this PR is incorrect (just by looking at the code, not running it) - we should show the menu if the error is not *500*, not 404. 
Having said that, we should move logic out of the template. I'd prefer to have a template variable `show_user_menu` or similar, which the error handler for 500 can set to False (perhaps we have to invert this flag so that the error handler sets it to True, and therefore we don't have to set it when showing every other page on the site)

alastair:
> While cross-checking the if claus, I replaced 500 with 404 and forgot to revert it back before pushing. I'll fix it.

This is a great indication that we should have a test for this!

mayhem:
Thanks @alastair and @Uditgulati for the explanations -- now it makes sense. 404 threw me off. :)

Uditgulati:
@alastair  I am getting no clue on how to test this as I can't think of a way to create conditions for 500 page. Any ideas?

paramsingh:
@Uditgulati first thing that comes to my mind is that you could add a view to `self.app` in the test that raises InternalServerError and then call that view?

Uditgulati:
@paramsingh I need to check that import button and drop-down menu aren't shown if user is logged in and there is 500 error. Would I need to login somehow in test or is there a way to mock login?

paramsingh:
There is self.temporary_login

alastair:
hi @Uditgulati!
I wrote some tests to check the user menu in the following cases:
 * user logged out
 * user logged in
 * http400 or http404 (show the menu)
 * http500 (don't show the menu)

I also changed the conditions in which we show the menu a bit, now we always show the 'Import now!' button, and then hide the user menu in a special case (`hide_navbar_user_menu`) so that we don't have to have http status codes in the template. Let me know what you think

alastair:
there's a failing test that I missed, I'll look at it later/tomorrow

Uditgulati:
@alastair  I think maybe in http500, we should only show 'Import!' button coz it could be that before this page, user might be logged in or not, so if he is, showing 'Sign In' button would be a bit weird for UI. Btw, thanks a lot for writing tests :)
If u don't mind, I'd like to fix that failing test :)

alastair:
tests working!

alastair:
@mayhem @paramsingh this is ready for merge

PR #369, opened 2018-02-27
LB-323: Allow users to request statistics calculation

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_):  [LB-323](https://tickets.metabrainz.org/browse/LB-323)

# Solution

Add a stats_calculator script and send users to that queue.

### DIFF
--- a/listenbrainz/stats/stats_calculator.py
+++ b/listenbrainz/stats/stats_calculator.py
@@ -27,6 +27,7 @@ import pika
 import ujson
 
 from listenbrainz import db, stats, utils
+from listenbrainz.stats.utils import construct_stats_queue_key
 from listenbrainz.webserver import create_app
 
 class StatsCalculator:
@@ -61,7 +62,9 @@ class StatsCalculator:
 
         entity_type = data.get('type', '')
         if entity_type == 'user':
-            self.calculate_stats_for_user(data)
+            done = self.calculate_stats_for_user(data)
+            if done:
+                self.redis.set(construct_stats_queue_key(data['musicbrainz_id']), 'done')
         else:
             self.log.info('Cannot recognize the type of entity in queue, ignoring...')
             return
@@ -91,6 +94,7 @@ class StatsCalculator:
         except KeyError:
             self.log.error('Invalid user data sent into queue, ignoring...')
             return False
+
 
         # if this user already has recent stats, ignore
         if db_stats.valid_stats_exist(user['id']):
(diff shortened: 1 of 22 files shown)


paramsingh:
Renamed the scheduler container because it isn't used in production either. Makes sense, I think.

listenbrainz/webserver/views/profile.py
```diff
@@ -202,3 +212,33 @@ def upload():
...
+            'id': current_user.id,
+            'musicbrainz_id': current_user.musicbrainz_id,
+        }
+        publish_data_to_queue(
+            data=data,
+            exchange=current_app.config['STATS_EXCHANGE'],
+            queue=current_app.config['STATS_QUEUE'],
+            error_msg='Could not put user %s into statistics calculation queue, please try again later',
+        )
+        _redis.redis.set(construct_stats_queue_key(current_user.musicbrainz_id), 'queued')
```
mayhem: Now that we merged the brainzutils PR, should this be updated to use brainz util caching instead?

listenbrainz/webserver/views/profile.py
```diff
@@ -202,3 +212,33 @@ def upload():
...
+def request_stats():
+    """ Check if the current user's statistics have been calculated and if not,
+        put them in the stats queue for stats_calculator.
+    """
+    status = _redis.redis.get(construct_stats_queue_key(current_user.musicbrainz_id)) == 'queued'
+    if status == 'queued':
+        flash.info('You have already been added to the stats calculation queue! Please check back later.')
+    elif db_stats.valid_stats_exist(current_user.id):
+        flash.info('Your stats were calculated in the most recent stats calculation interval,'
+            ' please wait until the next interval! We calculate new statistics every Monday at 00:00 UTC.')
```
mayhem: This is good, but out of date the second we change the cronjob. I think we should improve this in the future, but roll with it for now.

listenbrainz/stats/stats_calculator.py
```diff
@@ -0,0 +1,189 @@
...
+    def start(self):
+        """ Starts the stats calculator. This should run perpetually,
+            monitor the stats rabbitmq queue and calculate stats for
+            entities that it receives from the queue.
+        """
+
+        # if no bigquery support, sleep
+        if not self.app.config['WRITE_TO_BIGQUERY']:
+            while True:
+                time.sleep(10000)
```
mayhem: time has not been imported here.

listenbrainz/webserver/views/profile.py
paramsingh: This isn't exactly caching which is why I used our RedisConnection directly similar to how we use it for storing now_playing listens.

PR #371, opened 2018-03-02
LB-302:Add additonal condition for checking LastFM key in config

Uditgulati:
# Summary

No error message is generated if last.fm api key is not added to custom_config.py file

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_):  [LB-302](https://tickets.metabrainz.org/browse/LB-302)

# Solution

# Action

### DIFF
--- a/listenbrainz/webserver/views/profile.py
+++ b/listenbrainz/webserver/views/profile.py
@@ -89,7 +89,9 @@ def import_data():
     """ Displays the import page to user, giving various options """
 
     # Return error if LASTFM_API_KEY is not given in config.py
-    if 'LASTFM_API_KEY' not in current_app.config or current_app.config['LASTFM_API_KEY'] == "":
+    if 'LASTFM_API_KEY' not in current_app.config or \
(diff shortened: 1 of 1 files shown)


Uditgulati:
@paramsingh I replaced the last.FM api key with "USE_LASTFM_API_KEY" and noticed that the scraper isn't going in onerror block and instead the function is getting executed and also, I updated the message with value of playCount at the end of code, it showed -1. So, the scraper isn't stuck anywhere in the code and playCount isn't getting updated.

Can we modify the else at line 508 to instead show an error message?

paramsingh:
@Uditgulati your comment is hard to understand, could you please elaborate?

Uditgulati:
@paramsingh  in other words, I noticed that line 514 isn't getting executed when LastFM API key is invalid.

paramsingh:
@Uditgulati then you should take a look at the response that last.fm sends when it receives a bad api key and see how we can use that response to stop the scraper.

PR #373, opened 2018-03-09
LB-325: Disable import button if no input

Uditgulati:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [LB-325](https://tickets.metabrainz.org/browse/LB-325)

Starting Last.fm import without entering a username breaks the page.

# Solution

Disable the submit button if input is blank.

### DIFF
--- a/listenbrainz/webserver/templates/user/import.html
+++ b/listenbrainz/webserver/templates/user/import.html
@@ -59,6 +59,7 @@
     }
   </script>
   <script type="text/javascript">
+    //disables the import button when no username is entered
     $(document).ready(function() {
       $("#import_button").prop("disabled", true);
       $("#lastfm_input").on("input", function() {
--- a/listenbrainz/webserver/templates/user/import.html
+++ b/listenbrainz/webserver/templates/user/import.html
@@ -24,8 +24,8 @@
       We need to know your Last.fm username:
       <div class="well">
         <form onsubmit="return lastfm_import();">
-          <input id = "lastfm_input" type="text" name="lastfm_username" size="30" placeholder="Last.fm Username" oninput="check()">
-          <input id = "import_button" type="submit" value="Import now!" disabled>
+          <input id="lastfm_input" type="text" name="lastfm_username" size="30" placeholder="Last.fm Username">
+          <input id="import_button" type="submit" value="Import now!">
         </form>
       </div>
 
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/templates/user/import.html
```diff
@@ -58,4 +58,13 @@ <h3>Import a Last.fm backup file</h3>
       return false;
     }
   </script>
+  <script type="text/javascript">
+    function check() {
```
mayhem: check is a very generic name for this. check_import_button() maybe?

listenbrainz/webserver/templates/user/import.html
```diff
@@ -58,4 +58,13 @@ <h3>Import a Last.fm backup file</h3>
...
+      var i = document.getElementById("lastfm_input");
```
mayhem: Our pages have jquery installed on them and this can be done a lot simpler with jquery. Can you please change this code to use jquery?

listenbrainz/webserver/templates/user/import.html
Uditgulati: Seems good. ``check_input()`` or ``check_username()`` maybe?

listenbrainz/webserver/templates/user/import.html
Uditgulati: Ok, I'll modify it :+1:

Uditgulati: I did some research on jquery code for it and found that it is nearly of same length as this one.  Should we keep this or change it then?

alastair: I think that he refers to `$("#lastfm_input")` as being shorter to get the input box. I agree that if we use jquery in other places that we should also use it here. Additionally, in this case I don't think it's necessary to assign it to the variable `i` - the name is not very descriptive and it's not used in multiple places. You could do this entire check inside the if block.

listenbrainz/webserver/templates/user/import.html
```diff
@@ -58,4 +58,13 @@ <h3>Import a Last.fm backup file</h3>
...
+      if(i.value == "")
```
alastair: We don't have any explicit style checks here in the LB codebase, but please be careful about formatting of these if statements. Specifically:
```js
if (something) {
    x;
} else {
    y;
}
```
Always use braces, and be careful about whitespace

listenbrainz/webserver/templates/user/import.html
alastair: similarly for setting the disabled attribute

listenbrainz/webserver/templates/user/import.html
```diff
@@ -58,4 +58,16 @@ <h3>Import a Last.fm backup file</h3>
       return false;
     }
   </script>
+  <script type="text/javascript">
```
paramsingh: Should probably add a one line comment explaining the need for this script.

Uditgulati: :+1:

PR #374, opened 2018-03-12
Remove port forwards for supporting services in the docker-compose file

alastair:
# Summary

* This is a…
    *  Refactoring

Port forwards are not needed for running the services in docker-compose - they only serve to give access to services inside the container from a client on the host.
Services can still be accessed by running the client inside a container with `docker-compose run`

# Action

I'm opening this so that we can have a discussion about the best way to deal with these port forwards.
I specifically ran into a problem because I already have redis running on my host, and so it was unable to bind to the same port number. If we want to delete the port forwards, then we should still have an easy way to run a client and connect to these servers. As an example, you can do this with postgres by running something like

    docker-compose run --rm db psql -h db -U listenbrainz

We could adapt some ideas in [LB-326](https://tickets.metabrainz.org/projects/LB/issues/LB-326) to provide an easy way to run these clients from the develop.sh script.

Alternatively, we can keep the port forwards, but I'd recommend that we prefix port numbers on the host like we had with postgres:

```
    ports:
      - "15432:5432"
```

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -13,15 +13,11 @@ services:
     image: postgres:9.5.3
     volumes:
       - ./dbvolume:/var/lib/postgresql/data:z
-    ports:
-      - "15432:5432"
 
   redis:
     image: redis:3.2.1
     volumes:
       - ./data/redis:/data:z
-    ports:
-      - "6379:6379"
 
   influx:
     image: influxdb:1.2.4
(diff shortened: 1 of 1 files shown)


alastair:
Do we want helper scripts to give client access to the redis/influx/postgres servers, or should we bump this until LB-326? I don't know how often this is needed in day-to-day development.

mayhem:
I think it can wait -- I habitually go into the container to inspect services running therein, so I don't need helper scripts, but I would like to see what ilikecomputers has to say on this.

alastair:
> I go into the containers manually regularly too.

To be fair, it really sounds like we could do with some tools to help you both stop having to always type long and cryptic commands :)

PR #376, opened 2018-03-12
LB-312: Add functionality to delete old dumps

paramsingh:
# Summary

* This is a…

    * Feature addition

# Problem

* JIRA ticket (_optional_):[LB-312](https://tickets.metabrainz.org/browse/LB-312)

# Solution

Add a command to dump_manager and use it. Also, add tests.

### DIFF
--- a/listenbrainz/db/dump_manager.py
+++ b/listenbrainz/db/dump_manager.py
@@ -41,6 +41,9 @@ try:
 except ImportError:
     pass
 
+
+NUMBER_OF_DUMPS_TO_KEEP = 2
+
 log = logging.getLogger(__name__)
 
 cli = click.Group()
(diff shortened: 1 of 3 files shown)


listenbrainz/db/dump_manager.py
```diff
@@ -121,3 +123,41 @@ def import_dump(private_archive, public_archive, listen_archive, threads):
...
+    Returns:
+        (int, int): the number of dumps remaining, the number of dumps deleted
+    """
+    dump_re = re.compile('listenbrainz-dump-[0-9]*-[0-9]*')
+    dumps = [x for x in sorted(os.listdir(location), reverse=True) if dump_re.match(x)]
+    if not dumps:
+        print('No dumps present in specified directory!')
+        return
+
+    keep = dumps[0:2]
```
paramsingh: We're keeping only the latest two dumps right now (like in MB), but I can change this to take an argument, if we want to tinker with how many dumps we want to keep on ftp simultaneously.

mayhem: It would be nice to have it defined as a constant at the top of the file for clarity,but no need to make it a config item.

PR #378, opened 2018-03-13
Use docker volume instead of local dbvolume

LeoVerto:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

docker-compose.yml set up a named volume for the DB but falsely continued to use a local one. Now the named volume is used instead.

This previously caused ./develop.sh to fail for me with a "Permission Denied" error after the DB had been set up ([log](https://gist.github.com/LeoVerto/c5b88ce6965bff63dd467cb98b5f56d8)) but that has also been fixed as a side effect of this PR.

@mayhem [came up with this fix](https://chatlogs.metabrainz.org/brainzbot/metabrainz/2018-03-13/?msg=4138493&page=3).

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -12,7 +12,7 @@ services:
   db:
     image: postgres:9.5.3
     volumes:
-      - ./dbvolume:/var/lib/postgresql/data:z
+      - dbvolume:/var/lib/postgresql/data:z
 
   redis:
     image: redis:3.2.1


alastair:
While we're here do we also want to switch to named volumes for redis, influx, and rabbitmq?

advantages:
  * no more root-owned files in your home directory
  * easily delete volumes with docker volume rm

disadvantages:
  * for docker-mac/windows people, data counts towards the usage of the vm image disk usage

mayhem:
I've opened https://github.com/metabrainz/listenbrainz-server/pull/380 to replace this PR since alastairp brings up a good point.

PR #379, opened 2018-03-14
Fix missing query string format in error message

alastair:
# Summary

* This is a…
    * Bug fix

Some error strings were missing their format parameters

# Solution
Add format parameters to strings and add tests

### DIFF
--- a/listenbrainz/webserver/views/test/test_user.py
+++ b/listenbrainz/webserver/views/test/test_user.py
@@ -1,6 +1,7 @@
 import listenbrainz.db.stats as db_stats
 import listenbrainz.db.user as db_user
 import ujson
+from unittest import mock
 
 from flask import url_for, current_app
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/test/test_user.py
```diff
@@ -134,3 +132,45 @@ def test_username_case(self):
         self.assert200(response1)
         self.assert200(response2)
         self.assertEqual(response1.data.decode('utf-8'), response2.data.decode('utf-8'))
+
+    @mock.patch('listenbrainz.webserver.views.user.time')
+    @mock.patch('listenbrainz.webserver.influx_connection._influx.fetch_listens')
+    def test_ts_filters(self, influx, m_time):
+        """Check that max_ts and min_ts are passed to the influx """
```
alastair: comment isn't finished

PR #381, opened 2018-03-15
LB-329: Add supported keys to Influx rows explicitly

paramsingh:
# Summary

* This is a bug fix

# Problem

* JIRA ticket (_optional_): [LB-329](https://tickets.metabrainz.org/browse/LB-329)

Spotify ID's weren't being returned by the API when listens were requested. This was due to the fact that we didn't add them explicitly into the influx db.

# Solution

Add all supported keys explicitly and add a test for this.

### DIFF
--- a/listenbrainz/testdata/additional_info.json
+++ b/listenbrainz/testdata/additional_info.json
@@ -19,10 +19,10 @@
                         "REMIX"
                     ],
                     "spotify_id": "http://open.spotify.com/track/1rrgWMXGCGHru5bIRxGFV0",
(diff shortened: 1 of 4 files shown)


listenbrainz/testdata/additional_info.json
```diff
@@ -17,7 +17,14 @@
                     "release_type": [
                         "ALBUM",
                         "REMIX"
-                    ]
+                    ],
+                    "spotify_id": "http://open.spotify.com/track/1rrgWMXGCGHru5bIRxGFV0",
+                    "artist_mbids": ["abaa7001-0d80-4e58-be5d-d2d246fd9d87"],
```
mayhem: should we test multiple ids here?

PR #384, opened 2018-03-16
Single config file please

mayhem:
Just here for running tests..


listenbrainz/webserver/__init__.py
```diff
@@ -49,37 +49,23 @@ def gen_app(config_path=None, debug=None):
...
-        print("loading consul config file %s)" % os.path.join( os.path.dirname(os.path.realpath(__file__)), '..', 'custom_config.py'))
-        app.config.from_pyfile(custom_config)
+        print("loading consul config file %s)" % config_file)
+        app.config.from_pyfile(config_file)
 
     else:
-        print("loading custom config %s" % custom_config)
-        app.config.from_pyfile(custom_config, silent=True)
-
-    if config_path:
```
paramsingh: Removing this means that config_path passed to gen_app wouldn't work anymore?

mayhem: That seems right. Should we keep it? I was just wanting to remove all places where we could get this problem...

paramsingh: I've never come across a usecase for passing a config path yet, so I'd be okay with removing the functionality.

mayhem: Its in git -- we can always bring it back if we come across the use case.

PR #391, opened 2018-03-20
Disable fsync for integration tests

paramsingh:
Should make them faster too.

### DIFF
--- a/docker/docker-compose.integration.yml
+++ b/docker/docker-compose.integration.yml
@@ -8,6 +8,7 @@ services:
 
   db:
     image: postgres:9.5.3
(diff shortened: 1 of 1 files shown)


alastair:
Nice. Do you know if influx also has this setting? (Although we have to be careful in some cases - for example an fsync might be needed to be able to write and then read something again...)
https://github.com/influxdata/influxdb/blob/49ed5f3f5ba91b77feb8ea7888c2d75186da09b2/etc/config.sample.toml#L54

mayhem:
Err, sorry! 🍰 !

PR #393, opened 2018-03-27
Improved Spotify Integration

LeoVerto:
# Summary

This PR does a couple things at once:

It provides a temporary fix for [LB-338](https://tickets.metabrainz.org/browse/LB-338).
Unfortunately with the current state of the Spotify play button widget we can only play one song at a time although I'd like to implement a fully-featured player in the future, so the big embedded player space feels kind of wasted right now.

It adds Spotify icons to listens with associated Spotify IDs
screenshot-2018-3-27 user leo verto - listenbrainz

### DIFF
--- a/listenbrainz/webserver/templates/user/profile.html
+++ b/listenbrainz/webserver/templates/user/profile.html
@@ -98,7 +98,7 @@
 
         {% if spotify_uri %}
           <div class="col-md-4 text-right">
-            <iframe id="player" frameborder="0" allowtransparency="true"
+            <iframe id="player" frameborder="0" allowtransparency="true" allow="encrypted-media"
                     src="https://embed.spotify.com/?uri={{ spotify_uri }}&theme=white"></iframe>
           </div>
         {% endif %}
--- a/listenbrainz/webserver/templates/user/profile.html
+++ b/listenbrainz/webserver/templates/user/profile.html
@@ -68,6 +68,9 @@
                       {% endif %}
                   </td>
                   <td>
+                      {% if spotify_uri %}
+                        <span class="fab fa-spotify"></span>
+                      {% endif %}
                       {% if listen.track_metadata.additional_info.recording_mbid %}
                         <a href="http://musicbrainz.org/recording/{{ listen.track_metadata.additional_info.recording_mbid }}">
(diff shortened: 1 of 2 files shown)


LeoVerto:
I was planning to continue working on building a complete player but I suppose I can do that in a future PR.

LeoVerto:
Looks like I need to fix some tests for this PR to go through.

mayhem:
How do we add a test to see if this is working or not?

paramsingh:
we could mock the function that gets listens to send a listen with a spotify id, and then check that the template is called with the correct context.

Something like the following pseudocode:

```
def test_spotify_id_integration_on_user_page(self, mock_get_listens):
    mock_get_listens returns a listen with a spotify id
    response = client.get(userpage)
    assertContext('spotify_uri', the correct uri)
```

LeoVerto:
Wouldn't that test succeed on the current master even though the player is essentially broken there? The iframe has the correct URI but contains a 404 page.

This really isn't an issue on our side and I don't think we should test for it, especially considering we will have to drop the widget sooner or later if we want a player with more features.

LeoVerto:
I just learned that this PR is severly abusing Spotify's [Branding Guidelines](https://beta.developer.spotify.com/branding-guidelines/) so I'll have to come up with a fix for that.

mayhem:
I'm tempted to merge this for now -- to get it out of the way and so we can push it to beta for us to play with. Clearly we need to fix this better before we actually release it.

Thoughts?

chhavijustme:
Hi, if I understand right. The problem is with the spacing and size of the spotify icon according to the guidelines?
I can think of two ways to do it.
**One**
screen shot 2018-05-04 at 8 00 42 pm

**Two**
screen shot 2018-05-04 at 8 00 33 pm
See if any of these options work @LeoVerto! :)

paramsingh:
>I'm tempted to merge this for now -- to get it out of the way and so we can push it to beta for us to play with. Clearly we need to fix this better before we actually release it.

More than 2 months later, I'd tend to agree :P

mayhem:
I like chhavi's suggestions -- lets implement one of those. But for now, merging. :)

PR #400, opened 2018-04-22
Restream data to BigQuery.

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): 

We missed some data and needed schema changes in BQ, so this script restreams data into
Google BigQuery. It then streams each user's listens into the appropriate table.

# Solution

This script creates new tables for each year, putting all data before 2002 (the year that 
last.fm started) into a `before_2002` table.

# Action

We'll probably need to first get bigquery_writer to work with the new schema before running this.
We should discuss this a bit more.

### DIFF
--- a/listenbrainz/__init__.py
+++ b/listenbrainz/__init__.py
@@ -4,3 +4,7 @@ __version__ = '0.1'
 
 DUMP_LICENSE_FILE_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                                       'db', 'licenses', 'COPYING-PublicDomain')
+
+
+LAST_FM_FOUNDING_YEAR = 2002
+INVALID_LISTENS_BIGQUERY_TABLE_NAME = 'before_2002'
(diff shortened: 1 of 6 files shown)


paramsingh:
I didn't bother adding tests because this will only be a one-off script. There are some ideas in this such as creating the BQ schema that I think would have a place in manage.py though.

mayhem:
We'll also need to get the reports to work with the new format before we merge/run this.

paramsingh:
A problem that may lead to duplication:

1. We start the restreamer which streams listens according to batches of time (first all listens before 2002, then all listens before 2003 and so on)
2. The streamer sends some listens and dies for some reason (internal error, bigquery bad, anything)
3. We now have no idea about which listens have been sent and which haven't.
4. If we decide to start from the batch we were doing, we risk duplicates.

Possible way to avoid this: We could try resetting the bq dataset and starting over if the restreamer dies.

paramsingh:
Closing for now.

PR #402, opened 2018-04-24
Make bigquery-writer send listens to yearly tables.

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: We're doing new tables in BQ, so bqwriter needs to stream listens to these new tables.

# Problem

* JIRA ticket (_optional_): 

BigQueryWriter currently streams listens to a single table which is getting out of hand really quickly.

# Solution

Make bigquerywriter stream listens to yearwise tables.


paramsingh:
Moved functionality here to #400, makes more sense and makes it easier to share code. closing this. If #400 gets too big, I'll break it into parts.

PR #403, opened 2018-04-27
Create md5sums of dump files during dump creation

paramsingh:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): 

The dumps created did not have any md5sums created.

# Solution

Create md5s and send them to the FTP server as well.

### DIFF
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -534,6 +534,7 @@ class InfluxListenStore(ListenStore):
 
             pxz.stdin.close()
 
+        pxz.wait()
         self.log.info('ListenBrainz listen dump done!')
         self.log.info('Dump present at %s!', archive_path)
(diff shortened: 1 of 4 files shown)


listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -534,6 +534,7 @@ def dump_listens(self, location, dump_time=datetime.today(), threads=DUMP_DEFAUL
 
             pxz.stdin.close()
 
+        pxz.wait()
```
paramsingh: I think the absence of this might have been the cause of the dump file corruption. We didn't wait for the pxz command to exit, leading to a race condition between the cp and this, leading to corrupted files in some places.

I came across this because the hashes created and printed were different from the hashes of the actual files in dev also.

PR #404, opened 2018-04-27
Create md5sums of dump files during dump creation

paramsingh:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): 

The dumps created did not have any md5sums created.

# Solution

Create md5s and send them to the FTP server as well.

### DIFF
--- a/listenbrainz/db/dump_manager.py
+++ b/listenbrainz/db/dump_manager.py
@@ -26,6 +26,7 @@ import logging
 import os
 import re
 import shutil
+import subprocess
 import sys
 
 from datetime import datetime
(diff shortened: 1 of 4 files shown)


listenbrainz/db/dump_manager.py
```diff
@@ -70,6 +71,7 @@ def create(location, threads):
     create_path(dump_path)
     db_dump.dump_postgres_db(dump_path, time_now, threads)
     ls.dump_listens(dump_path, time_now, threads)
+    write_hashes(dump_path)
```
mayhem: There is no error handling here. If the disk runs out of space during hash creation, the hash creation fails silently.

PR #405, opened 2018-05-05
Keep track of when a listen was inserted into influx.

paramsingh:
# Summary

Keep track of when a listen was inserted in an extra field `inserted_timestamp` in influx

# Problem

It is hard to restream listens from influx into BQ without duplication, without row ids, as the influx db is dynamic and listens keep getting inserted.

# Solution

With this change, we could keep track of when we start a restream and only restream
those listens that were inserted before a particular timestamp, while the rest of the listens
get streamed to BQ by the bigquery-writer.

### DIFF
--- a/listenbrainz/listen.py
+++ b/listenbrainz/listen.py
@@ -1,7 +1,8 @@
 # coding=utf-8
+import calendar
+import time
 
 from datetime import datetime
-import calendar
 from listenbrainz.utils import escape, convert_to_unix_timestamp
 
 def flatten_dict(d, seperator='', parent_key=''):
@@ -206,6 +207,7 @@ class Listen(object):
                 'tracknumber': self.data['additional_info'].get('tracknumber', ''),
                 'isrc': self.data['additional_info'].get('isrc', ''),
                 'spotify_id': self.data['additional_info'].get('spotify_id', ''),
+                'inserted_timestamp': int(time.time()),
             }
         }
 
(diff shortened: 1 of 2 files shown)


paramsingh:
Note that there is a problems with this approach:  Influx doesn't have a query to get only rows where a field is missing (https://stackoverflow.com/questions/30091980/select-from-influxdb-where-value-is-null) and for all rows present in the influx db currently, the new field would be missing, meaning that for the listens currently, we wouldn't be able to do queries liek `select * from measurement where inserted_timestamp is null or inserted_timestamp > x` so I'm not sure if this is the best way to do this.

Decided to do this in a seperate PR to get opinions.

listenbrainz/listen.py
```diff
@@ -206,6 +214,7 @@ def to_influx(self, measurement):
                 'tracknumber': self.data['additional_info'].get('tracknumber', ''),
                 'isrc': self.data['additional_info'].get('isrc', ''),
                 'spotify_id': self.data['additional_info'].get('spotify_id', ''),
+                'inserted_timestamp': int(time.time()),
```
mayhem: Why is there an inserted_timestamp both in the top level and in the data??

mayhem:
I think having somewhat slow restreams and doing additional filtering in python is OK in my book. Hopefully we won't be doing it too often.

paramsingh:
@mayhem, cool. This method should work then.

mayhem:
This PR is ready to merge, but I am not certain that we should merge this into master -- production since we want to get this out with our GDPR/Spotify release?

paramsingh:
See #410.

PR #407, opened 2018-05-14
Spotify listen importer

alastair:
# Summary

* This is a…
    *  Feature addition
* **Describe this change in 1-2 sentences**: Add the ability to connect your spotify account to Listenbrainz and automatically import listens

# Solution

Spotify has an API that can read the 50 most recent songs played by a user. We can do an oauth connection and read the user's recent plays and import them as listens.

The user profile page has a connect to spotify button which does an oauth workflow.
A separate script loops through all users who have performed this login flow and gets listens for each of them, converting them to listens and importing into LB.
Listens are imported starting from the user who last had their listens imported least recently.

# Action

I tried a few new things with code layout, specifically a `domain` package that contains methods relevant to spotify, but disconnected from our database methods. This allows us to test database methods in db directly with the database and not worry about interaction, and methods that "do stuff", while letting us mock database access to run tests faster.
I'm open to feedback on if this is a direction that we want to take with this code, but if it's to confusing we can rearrange it into a format closer to the existing code.

There are some tests missing, especially in the domain package. There are a number of comments in the `spotify_read_listens.py`script especially regarding error catching from Spotify about what we should do in particular failure cases. This isn't implemented yet.

I don't know the preferred way of starting scripts in the LB production environment, especially regarding config file reading, etc. This will need to be updated. 

We have to keep an eye on how long this operation will take to run. At only 50 songs per history for a user, we may find that with too many users that we start running behind. Perhaps we'll have to run multiple scripts. This is a future concern.


alastair:
Last.fm's spotify connect documentation (https://getsatisfaction.com/lastfm/topics/spotify-scrobbling-gvfb1uv1ff1cs) says

> Private sessions are not supported
> We don't currently support Private Sessions so all listens will appear on your Last.fm profile. If you're using Private Sessions to hide your listening history, then please don't enable the new scrobbler yet.

We should confirm if this is a concern that we might have... perhaps spotify returns these listens in their play history but with a flag in the json? If so we should skip it. If they report it with no flag we should ping Hugh

listenbrainz/domain/spotify.py
```diff
@@ -0,0 +1,102 @@
...
+        self.token_expires = token_expires
+        self.refresh_token = refresh_token
+        self.last_updated = last_updated
+        self.active = active
+        self.update_error_message = update_error_message
+
+    def get_spotipy_client(self):
+        return spotipy.Spotify(auth=self.user_token)
+
+    def nice_last_updated(self):
```
mayhem: Is this a function to make it nice and human readable? If so, maybe should also also use an "ago" function to make the timestamp more approachable.

alastair:
@mayhem requested a breakdown of what I think should be done before and after a merege:
before:
 * Config/running of script
 * Handle failing error codes from spotify: https://github.com/metabrainz/listenbrainz-server/blob/fa9e1da5665193cbbb688dde1f383dce5414d3b1/listenbrainz/spotify_updater/spotify_read_listens.py#L61-L70
 * Handle failures from adding listens: https://github.com/metabrainz/listenbrainz-server/blob/fa9e1da5665193cbbb688dde1f383dce5414d3b1/listenbrainz/spotify_updater/spotify_read_listens.py#L87-L88

After:

 * Finalise tests
 * Confirm that we don't record private listens
 * decide if the domain package is nice

listenbrainz/domain/spotify.py
```diff
@@ -0,0 +1,102 @@
...
+        if not self.last_updated:
+            return 'never'
+        return self.last_updated
+
+    @property
+    def token_expired(self):
+        print(self.token_expires)
```
mayhem: cruft?

listenbrainz/domain/spotify.py
```diff
@@ -0,0 +1,102 @@
...
+                       token_expires=row['token_expires'],
+                       refresh_token=row['refresh_token'],
+                       last_updated=row['last_updated'],
+                       active=row['active'],
+                       update_error_message=row['update_error_message'])
+
+    def __str__(self):
+        return "<Spotify(user:%s)>" % self.user_id
+
+# TODO: All documentation for these
```
mayhem: @paramsingh : please add tickets for all the TODOs in this PR.

listenbrainz/domain/spotify.py
```diff
@@ -0,0 +1,102 @@
...
+            'profile.connect_spotify_callback',
+            _external=True)
+    return spotipy.oauth2.SpotifyOAuth(client_id, client_secret, redirect_uri=redirect_url, scope=scope)
+
+
+def get_user(user_id):
+    row = db_spotify.get_user(user_id)
+    if row:
+        return Spotify.from_dbrow(row)
+
```
mayhem: Hmm. Not sure I'm a fan of implicit returns..

alastair:
One final thing - we should decide if we're happy with the data that we add to listenbrainz:
https://github.com/metabrainz/listenbrainz-server/blob/fa9e1da5665193cbbb688dde1f383dce5414d3b1/listenbrainz/spotify_updater/spotify_read_listens.py#L21-L33

listenbrainz/domain/tests/test_spotify.py
```diff
@@ -0,0 +1,7 @@
+from listenbrainz.webserver.testing import ServerTestCase
+
+
+class SpotifyDomainTestCase(ServerTestCase):
+
+    def test(self):
+        pass
```
mayhem: pass what??

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -0,0 +1,111 @@
...
+
+    listenbrainz_user = db_user.get(user.user_id)
+
+    spotipy_client = user.get_spotipy_client()
+
+    recently_played = spotipy_client._get("me/player/recently-played", limit=50)
+    # TODO: if error
+    # If it's 401, try and refresh one more time and get it
+    # If that time fails, report the failure
+    # If refresh fails, report the failure
```
mayhem: Param: Please report failures to sentry.

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -0,0 +1,111 @@
...
+            listens.append(_convert_spotify_play_to_listen(item))
+
+    try:
+        for listen in listens:
+            validate_listen(listen, LISTEN_TYPE_IMPORT)
+    except werkzeug.exceptions.BadRequest:
+        spotify.update_last_updated(user.user_id, False, 'Unable to validate listens')
+        return
+        # TODO: This is our problem, report it to LB developers with a logger (sentry?)
+        # TODO: Skip invalid listens instead of skipping all listens?
```
mayhem: Yes, get as many valid listens in as possible.

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -0,0 +1,111 @@
...
+    for u in users:
+        print("user ")
+        print(u)
+        process_one_user(u)
+
+
+def main():
+    app = listenbrainz.webserver.create_app()
+
+    with app.app_context():
```
mayhem: Should there be some throttling here? While we don't have a lot of users, this tight loop would hammer the spotify API unnecessarily. It would be ideal to find some delay here that we can sleep for and then wake up again to start over. I think this point will need some discussion...

listenbrainz/webserver/views/profile.py
```diff
@@ -202,3 +204,44 @@ def upload():
...
+@login_required
+def connect_spotify_callback():
+    code = request.args.get('code')
+    if not code:
+        raise BadRequest('missing code')
+    sp_oauth = db_spotify._get_spotify_oauth()
+
+    try:
+        token = sp_oauth.get_access_token(code)
+        print("code: {}".format(code))
```
mayhem: also cruft?

listenbrainz/spotify_updater/spotify_read_listens.py
alastair: In addition to developer reports, I had the idea of a field `spotify.update_error` which could be used to report to the user if an error happened while doing an update (e.g. 'we no longer have permission to read your listens, please re-auth with spotify'). I've not used this yet, but my plan with this comment was to report this user to the user (this specific error is that if we try and refresh the user's oauth access key and it fails we should tell this to the user)

listenbrainz/domain/tests/test_spotify.py
alastair: pass the tests, please

listenbrainz/domain/spotify.py
alastair: explicit `else return None`? Fine by me

listenbrainz/domain/spotify.py
alastair: I think this was for a potential status message on the user profile page ("You have connected your Spotify account to ListenBrainz and it was last updated x", 'never' is for the case where a user connects but it hasn't run yet) - you're right that "ago" would be a good update here.

listenbrainz/spotify_updater/spotify_read_listens.py
alastair: Yes good, but we need to keep track of this in the medium term (with a logging system) - we need to know how long it takes for a run to happen and how many users we have - if it takes more than 150 minutes to run (50 songs * 3 minutes) then we could potentially miss listens

mayhem: Agreed -- we need to build a system that detects when we're getting close to falling behind so we can take corrective measures.

mayhem:
I also added https://tickets.metabrainz.org/browse/LB-361

mayhem:
And also: https://tickets.metabrainz.org/browse/LB-362 , https://tickets.metabrainz.org/browse/LB-363 , https://tickets.metabrainz.org/browse/LB-364 .

We have quite a lot of stuff to do here -- I wonder if we should merge this to another branch and not quite commit to merging to master yet.

paramsingh:
> I wonder if we should merge this to another branch and not quite commit to merging to master yet.

+1, there's a lot of stuff to do here.

mayhem:
Well, the branch is already in this repo, so we can close this PR for now and get back to it in a bit.

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -0,0 +1,111 @@
...
+
+    spotipy_client = user.get_spotipy_client()
+
+    recently_played = spotipy_client._get("me/player/recently-played", limit=50)
+    # TODO: if error
+    # If it's 401, try and refresh one more time and get it
+    # If that time fails, report the failure
+    # If refresh fails, report the failure
+    # What other types of error can we make? (Try bad/invalid keys), 404?
+    # If you get status code 429, it means that you have sent too many requests. If this happens, have a look in the Retry-After header
```
paramsingh: Spotipy is nice enough to do this for us, it seems. :)

https://github.com/plamere/spotipy/blob/master/spotipy/client.py#L152

PR #409, opened 2018-05-15
LB-354: Musicbrainz delete endpoint

paramsingh:
This PR adds the MusicBrainz `delete-user` endpoint as specified by https://tickets.metabrainz.org/browse/MBS-9680.

### DIFF
--- a/listenbrainz/webserver/views/index.py
+++ b/listenbrainz/webserver/views/index.py
@@ -1,15 +1,19 @@
 
 #TODO(param): alphabetize these
 from brainzutils import cache
-from flask import Blueprint, render_template, current_app, redirect, url_for, request
+from flask import Blueprint, render_template, current_app, redirect, url_for, request, jsonify
 from flask_login import current_user, login_required
+from werkzeug.exceptions import Unauthorized
+from requests.exceptions import HTTPError
 import os
 import subprocess
+import requests
 import locale
 import listenbrainz.db.user as db_user
 from listenbrainz.db.exceptions import DatabaseException
 from listenbrainz import webserver
 from listenbrainz.webserver.influx_connection import _influx
+from listenbrainz.webserver.views.profile import delete_user
 from influxdb.exceptions import InfluxDBClientError, InfluxDBServerError
 import pika
 import listenbrainz.webserver.rabbitmq_connection as rabbitmq_connection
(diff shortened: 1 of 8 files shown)


paramsingh:
@mwiencek, please take a look at the last commit (https://github.com/metabrainz/listenbrainz-server/pull/409/commits/ba533f8d8af4ef9550c02fb9298186929e9febcb) if you get the time.

mayhem:
Rebase, please.

listenbrainz/webserver/views/index.py
```diff
@@ -120,6 +124,35 @@ def current_status():
...
+
+
+def _authorize_mb_user_deleter(auth_token):
+    query_url = 'https://musicbrainz.org/oauth2/userinfo'
+    headers = {'Authorization': 'Bearer {}'.format(auth_token)}
+    r = requests.get(query_url, headers=headers)
+    try:
+        r.raise_for_status()
+    except HTTPError:
+        raise Unauthorized('Not authorized to use this view')
```
mayhem: I like how you give no useful info the caller. Well done!

paramsingh:
From the ticket, it seemed like the creation of the `UserDeleter` account would be handled in MusicBrainz.

listenbrainz/webserver/views/index.py
LordSputnik: https://httpstatuses.com/401 - `Unauthorised` requires a "WWW-Authenticate" header. Does this get set internally?

paramsingh: Doesn't seem like werkzeug sets it (https://github.com/pallets/werkzeug/blob/master/werkzeug/exceptions.py#L214)

listenbrainz/webserver/views/index.py
```diff
@@ -120,6 +124,35 @@ def current_status():
     )
 
 
+@index_bp.route('/delete-user/<musicbrainz_id>')
+def mb_user_deleter(musicbrainz_id):
```
mwiencek: We decided in last week's meeting that storing the editor IDs in LB et al. is the way to go, so MB will actually send the editor row ID here and not the name. (Not sure if you've already started updating it to expect that, or if `musicbrainz_id` is actually the user name here.)

listenbrainz/webserver/views/index.py
```diff
@@ -120,6 +124,35 @@ def current_status():
...
+    _authorize_mb_user_deleter(request.args.get('auth_token', ''))
+    delete_user(musicbrainz_id)
+    return jsonify({'status': 'ok'}), 200
+
+
+def _authorize_mb_user_deleter(auth_token):
+    query_url = 'https://musicbrainz.org/oauth2/userinfo'
```
mwiencek: (I mentioned this in the BookBrainz PR too, but) it'd be nice if the MusicBrainz server URL were configurable, so we don't have to hand-edit source files when testing this locally (or against test.mb, say). Not just here, but for OAuth logins and all other references too.

mwiencek:
> This looks good to me -- one question -- what setup do we need to create the user that can delete users? Should there be a script for that?

I just created the user we need manually and documented it here: https://github.com/metabrainz/syswiki/blob/master/MusicBrainzUserDeleter.md

mayhem:
OK, I think this is ready to merge.

marc2k3:
Is this working? I deleted my MEB account ages ago yet my LB profile is still present...

https://listenbrainz.org/user/marc2k3

PR #412, opened 2018-06-07
Upgrade brainzutils to fix listen count problem

paramsingh:
The last time we upgraded brainzutils, I forgot to update
`__version__` in `__init__.py` which meant that the new changes
fixing cache expiration didn't get installed in the container.
This new release should fix that.

Also, brainzutils uses a new version of SQLAlchemy, so upgrade
SQLAlchemy too. Testing doesn't show any problems.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ Flask-WTF == 0.12
 Flask-SQLAlchemy == 2.0
 Jinja2 == 2.8
 Werkzeug == 0.10.4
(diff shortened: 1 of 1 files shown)


paramsingh:
This is the commit in brainzutils that fixes the problem: https://github.com/metabrainz/brainzutils-python/commit/56bcef9e4cbb6c8d8ff1cb08ef6a3f58e6678194

PR #413, opened 2018-06-08
Merge production into master.

paramsingh:
This should bring master back in sync with the gdpr changes we released.

Only one new commit to fix some tests that started failing because of gdpr redirects.


listenbrainz/webserver/views/test/test_index.py
```diff
@@ -152,8 +159,6 @@ def view500():
         self.assertNotIn('Your Listens', data)
         self.assertIn('Sign in', data)
         self.assertIn('Import!', data)
-        mock_user_get.assert_not_called()
```
paramsingh: this needs to be removed, because the `before_request` stuff we do now means that user_get gets called no matter what if the user is logged-in.

paramsingh:
:tada: :tada:

PR #415, opened 2018-06-12
(WIP) Import musicbrainz row ids for users in LB

paramsingh:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: Add a script to import musicbrainz row ids to users in LB.

# Problem

* JIRA ticket (_optional_): 

Need to save musicbrainz row ids in LB too.

# Solution

Connect to the musicbrainz db, query it for the user ID and put it in for each user.

### DIFF
--- a/import_musicbrainz_row_ids.py
+++ b/import_musicbrainz_row_ids.py
@@ -64,7 +64,7 @@ def import_musicbrainz_rows(musicbrainz_db_uri, dry_run=True):
     with musicbrainz_db.engine.connect() as mb_connection:
         with db.engine.connect() as connection:
             for user in users:
-                if user['musicbrainz_row_id'] is not None:
+                if user.get('musicbrainz_row_id') is not None:
                     already_imported += 1
                     continue
                 name = user['musicbrainz_id']
@@ -81,11 +81,12 @@ def import_musicbrainz_rows(musicbrainz_db_uri, dry_run=True):
                     import_count += 1
                 else:
                     print('No user with specified username in the MusicBrainz db: %s' % name)
-                    print('Deleting user %s' % name)
-                    try:
-                        delete_user(user)
-                    except NotFound:
-                        print('User %s not found in LB...' % name)
+                    if not dry_run:
+                        print('Deleting user %s' % name)
+                        try:
+                            delete_user(user)
+                        except NotFound:
+                            print('User %s not found in LB...' % name)
                     not_found += 1
                     continue
 
@@ -99,6 +100,7 @@ def import_musicbrainz_rows(musicbrainz_db_uri, dry_run=True):
                             'id': user['id'],
                         })
                     print('Inserted row_id %d for user %s' % (musicbrainz_row_id, name))
+
     print('Total number of ListenBrainz users: %d' % len(users))
     print('Total number of ListenBrainz users with already imported row ids: %d' % already_imported)
     print('Total number of ListenBrainz users whose row ids can be imported: %d' % import_count)
--- a/import_musicbrainz_row_ids.py
+++ b/import_musicbrainz_row_ids.py
@@ -4,14 +4,19 @@ from brainzutils import musicbrainz_db
 import listenbrainz.db.user as db_user
 import sqlalchemy
 
-def import_musicbrainz_rows(musicbrainz_db_uri):
+
+def import_musicbrainz_rows(musicbrainz_db_uri, dry_run=True):
     musicbrainz_db.init_db_engine(musicbrainz_db_uri)
     db.init_db_connection(config.SQLALCHEMY_DATABASE_URI)
     users = db_user.get_all_users()
+    import_count = 0
+    already_imported = 0
+    not_found = 0
     with musicbrainz_db.engine.connect() as mb_connection:
         with db.engine.connect() as connection:
             for user in users:
                 if user['musicbrainz_row_id'] is not None:
+                    already_imported += 1
                     continue
                 name = user['musicbrainz_id']
                 result = mb_connection.execute(sqlalchemy.text("""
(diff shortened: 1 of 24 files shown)


import_musicbrainz_row_ids.py
```diff
@@ -0,0 +1,41 @@
...
+import listenbrainz.db.user as db_user
+import sqlalchemy
+
+def import_musicbrainz_rows(musicbrainz_db_uri):
+    musicbrainz_db.init_db_engine(musicbrainz_db_uri)
+    db.init_db_connection(config.SQLALCHEMY_DATABASE_URI)
+    users = db_user.get_all_users()
+    with musicbrainz_db.engine.connect() as mb_connection:
+        for user in users:
+            name = user['musicbrainz_id']
```
mwiencek: Just a suggestion, you can `continue` if `user['musicbrainz_row_id']` is already set, if you think it's possible the script may have to be run more than once (so it'll be much faster subsequent times).

Though if you make sure new accounts are created with a `musicbrainz_row_id` set, and deploy that patch first, this should only need to be run once.

import_musicbrainz_row_ids.py
```diff
@@ -0,0 +1,41 @@
...
+                    'name': name,
+                })
+            musicbrainz_row_id = None
+            if result.rowcount > 0:
+                musicbrainz_row_id = result.fetchone()['id']
+            else:
+                print('No user with specified username in the MusicBrainz db: %s' % name)
+                continue
+
+            with db.engine.connect() as connection:
```
mwiencek: If LB has a lot of users, it could be beneficial to create this connection outside the loop.

mayhem:
Logging everyone out would only happen for every time we run this script, which would ideally only be once, correct?

paramsingh:
I just pushed a commit that imports musicbrainz_row_id's from login info. If we log out everyone, then it would import row ids for everyone once they log in, with no need for running the script. 

However, this wouldn't work for people who don't log in, meaning that we won't have row ids for them, meaning we wouldn't be able to delete their accounts via the `/delete-user` endpoint.

mwiencek:
Oh, maybe they don't need to be logged out. Is `get_user` called on most pages? If so, it seems like it should automatically update `musicbrainz_row_id` by just them visiting a page. Or if not, that at least seems technically possible.

listenbrainz/webserver/login/provider.py
```diff
@@ -30,8 +30,11 @@ def get_user():
         'redirect_uri': url_for('login.musicbrainz_post', _external=True)
     }, decoder=lambda b: ujson.loads(b.decode("utf-8")))
     data = s.get('oauth2/userinfo').json()
-    user = db_user.get_or_create(data.get('sub'))
+    musicbrainz_id = data.get('sub')
+    user = db_user.get_or_create(musicbrainz_id)
```
mwiencek: This'll need to look up users based on the `metabrainz_user_id ` too (otherwise there's no point in storing it :)). It'll have to fallback to using the `sub` where `musicbrainz_row_id` isn't set, but should still check that it isn't clobbering another user:

```SQL
SELECT *
  FROM "user"
 WHERE musicbrainz_row_id = :metabrainz_user_id
    OR (LOWER(musicbrainz_id) = LOWER(:sub) AND musicbrainz_row_id IS NULL);
```

mwiencek: Not sure what you're supposed to do if there's already a user with that name having a different row ID. Since there's a UNIQUE constraint on the name. You can log the issue somewhere and change the new username to something unique (append an underscore or something). Hopefully it never happens in practice.

paramsingh:
>Oh, maybe they don't need to be logged out. Is get_user called on most pages?

`get_user` is only called once during login.

We could add some stuff to `before_request` [here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/__init__.py#L139) to check if the row id is filled for the user, but that would still require people to come to LB before it gets imported.

import_musicbrainz_row_ids.py
```diff
@@ -0,0 +1,115 @@
...
+        connection.execute(sqlalchemy.text("""
+            UPDATE "user"
+               SET musicbrainz_id = 'Fée Deuspi'
+             WHERE id = 2106
+            """))
+
+        # 243 - ClæpsHydra
+        connection.execute(sqlalchemy.text("""
+            UPDATE "user"
+               SET musicbrainz_id = 'ClæpsHydra'
```
mayhem: LOL

import_musicbrainz_row_ids.py
```diff
@@ -0,0 +1,115 @@
...
+            'musicbrainz_id': user['musicbrainz_id'],
+        },
+        exchange=app.config['BIGQUERY_EXCHANGE'],
+        queue=app.config['BIGQUERY_QUEUE'],
+        error_msg='Could not upt user %s into queue for bq deletion.' % user['musicbrainz_id'],
+    )
+    db_user.delete(user['id'])
+
+
+def import_musicbrainz_rows(musicbrainz_db_uri, dry_run=True):
```
mayhem: Normally I would say that we should be catching exceptions here, but given that this script will only ever be run by you and hopefully only once, I suppose we can skip it.

import_musicbrainz_row_ids.py
paramsingh: just realized, we would need to fix the influx measurement name for these users too. :/

Influx doesn't have an easy way to rename measurements (and there's a lot of listens)

https://listenbrainz.org/user/Cl%EF%BF%BDpsHydra
https://listenbrainz.org/user/F%EF%BF%BDe%20Deuspi

Maybe I should just write the musicbrainz user id's of these two users by hand...
Then, once we allow renaming user accounts, their names would get fixed automatically.

PR #416, opened 2018-06-14
Continue with app creation even if rabbitmq is absent

paramsingh:
Also, check every request if consul has provided host and port values and connect if possible.

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -70,6 +70,16 @@ def _send_listens_to_queue(listen_type, listens):
             submit.append(listen)
 
     if submit:
+        # check if rabbitmq connection exists or not
+        # and if not then try to connect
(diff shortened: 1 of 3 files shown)


paramsingh:
The builds are failing because of Postgres installation issues, I'll check it out.

listenbrainz/webserver/views/api_tools.py
```diff
@@ -70,6 +70,16 @@ def _send_listens_to_queue(listen_type, listens):
             submit.append(listen)
 
     if submit:
+        # check if rabbitmq connection exists or not
+        # and if not then try to connect
+        if rabbitmq_connection._rabbitmq is None:
```
mayhem: Hmm. Using a _function from outside the module bugs me. I think we should add an accessor function here. Either that or always call init_rabbitmq_connection and have it do nothing if the connection is already established.

PR #417, opened 2018-06-19
Get MusicBrainz row ids for exceptions by hand.

paramsingh:
Just realized that we can't willy nilly rename users in postgres, we would need to fix the influx measurement name for these users too. :/

Influx doesn't have an easy way to rename measurements (and there's around 200k listens for one of the users)

https://listenbrainz.org/user/Cl%EF%BF%BDpsHydra
https://listenbrainz.org/user/F%EF%BF%BDe%20Deuspi

I decided to just write the musicbrainz row id's of these two users by hand...
Then, once we allow renaming user accounts, their names would get fixed automatically.

### DIFF
--- a/import_musicbrainz_row_ids.py
+++ b/import_musicbrainz_row_ids.py
@@ -22,20 +22,39 @@ influx = init_influx_connection(logging, {
 
 
 def fix_username_for_exceptions():
-    with db.engine.connect() as connection:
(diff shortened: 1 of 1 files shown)


mayhem:
willy nilly. Just like loosey goosey, they are great technical terms!

mayhem:
How will we ever rename accounts if we can't rename measurements?

PR #422, opened 2018-06-25
Spotify Listen Importer

paramsingh:
# Summary

* This is a **Feature addition**.
* **Describe this change in 1-2 sentences**: Allow users to import listens from Spotify

# Problem

Opening a new PR because I can't reopen #407. We had a bunch of things that needed
to be done before we could deploy the spotify importer.

# Solution

I've fixed most of the stuff

* [x] error handling and logging while getting listens from spotify
* [x] error handling and logging during listen validation and submission to LB
* [x] adding an admin view that allows for activation / deactivation of imports
* [x] add tests and documentation
* [x] fix lots of bugs
* [x] remove code cruft and duplicated code
* [ ] show user feedback on when listens were last imported and if anything broke.

# Action

I would like to get this reviewed and merged, closing existing tickets. I'll work on the showing feedback work in the meanwhile.


alastair:
> *  fix lots of bugs

:joy:

listenbrainz/webserver/templates/user/import.html
```diff
@@ -29,6 +29,15 @@ <h3>Direct import from Last.fm</h3>
         </form>
       </div>
 
+    <h3> Import from Spotify </h3>
+
+    <p>
+      ListenBrainz can import future listens from your Spotify account.
```
alastair: I think this is a bit confusing, I guess you're trying to say that it can't import the past listening history (or did I write this?)
How about:
> ListenBrainz can automatically import songs from Spotify as you listen to them

listenbrainz/domain/spotify.py
```diff
@@ -0,0 +1,150 @@
...
+           active=row['active'],
+           update_error_message=row['update_error'],
+           musicbrainz_id=row['musicbrainz_id'],
+        )
+
+    def __str__(self):
+        return "<Spotify(user:%s): %s>" % (self.user_id, self.musicbrainz_id)
+
+
+def refresh_user_token(spotify_user):
```
mayhem: Should there be any sort of validation on spotify_user?

listenbrainz/domain/spotify.py
```diff
@@ -0,0 +1,150 @@
...
+    Args:
+        user_id (int): the ListenBrainz row ID of the user
+    """
+    row = db_spotify.get_user(user_id)
+    if row:
+        return Spotify.from_dbrow(row)
+    return None
+
+
+def delete_spotify(user_id):
```
mayhem: This and create_spotify are not very descriptive function names.

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -0,0 +1,246 @@
...
+                # Rate Limit Problems -- the client handles these, but it can still give up
+                # after a certain number of retries, so we look at the header and try the
+                # request again, if the error is raised
+                time_to_sleep = e.headers.get('Retry-After', delay)
+                current_app.logger.warn('Encountered a rate limit, sleeping %d seconds and trying again...', time_to_sleep)
+                time.sleep(time_to_sleep)
+                delay += 1
+                if retries == 0:
+                    raise spotify.SpotifyListenBrainzError('Encountered a rate limit.')
+            elif e.http_status in (400, 403, 404) or (e.http_status >= 500 and e.http_status < 600):
```
mayhem: The 400 series basically means "caller's fault" and 500 is "server's fault". Are we sure we want to retry 400 errors?? Retying 403 and 404 doesn't really make sense to me...

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -0,0 +1,246 @@
...
+    listenbrainz_user = db_user.get(user.user_id)
+    try:
+        recently_played = get_user_recently_played(user)
+    except (spotify.SpotifyListenBrainzError, spotify.SpotifyAPIError) as e:
+        raise
+
+    # convert listens to ListenBrainz format and validate them
+    listens = []
+    if 'items' in recently_played:
+        current_app.logger.info('Received %d listens from Spotify for %s', len(recently_played['items']),  str(user))
```
mayhem: This is going to yield large logs and raise zas' ire. Perhaps make this debug?

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -0,0 +1,246 @@
...
+    app = listenbrainz.webserver.create_app()
+    with app.app_context():
+        current_app.logger.info('Spotify Reader started...')
+        while True:
+            t = time.time()
+            success, failure = process_all_spotify_users()
+            if success + failure > 0:
+                current_app.logger.info('All %d users in batch have been processed.', success + failure)
+                current_app.logger.info('Total time taken: %.2f s, average time per user: %.2f s.', time.time() - t, (time.time() - t) / (success + failure))
+            time.sleep(10)
```
mayhem: I wonder if this should be dynamic -- 10 seconds should work fine for now, but over time as more people use our stuff 10 seconds is time wasted we may need to use. 

For instance, if the last update run didn't fetch anything new, we can wait longer. But if we got a pile of updates, we may not want to wait 10 seconds.

paramsingh:
>A user who is not actively listening should be checked less frequently as one who is listening, no?

It is hard to judge which user's are actively listening because spotify always returns 50 plays and we have no idea how many of them are dupes. I guess we could somehow try to find that out.

mayhem:
What if we say that a user who has had a listen in the last hour is considered to be active?

listenbrainz/domain/spotify.py
paramsingh: I don't think it's needed. Bad values would lead to raising `SpotifyException` which would be caught later in the script anyways.

paramsingh:
Nice idea, we could store the timestamp of the last listen for the user and sort via that.

mayhem:
Yes, that. Lets do that.

LeoVerto:
I don't think a simple LIFO queue (if that is what you're suggesting) would work too well for listeners who have "burst" listening habits, i.e. don't listen for a couple days and then listen for a couple hours nonstop.

Maybe categorizing users into different queues by last listen time and then assigning the queues different pick chances would work better to prevent users not being checked in time?

E.g. 50% chance to pick a user with last listen during the last hour, 25% to pick one with last listen during the last day, 15% during last week, 10% further back (this last queue would probably end up being the largest).

listenbrainz/domain/spotify.py
alastair: additionally, this is only called by us, with a value from the database, originally populated by an API call to spotify, so I don't think it's is actually possible to have a bad value here

mayhem:
LeoVerto: The current implementation is a straight FIFO (as opposed to my favorite data structure, a FINO) and I think that is too simplistic. Which is why I am suggesting tracking the last listen timestamp which then lets us tune when to check the user again.

If we assume for a moment that an inactive listener goes on a binge to listen to a shit-town of 60 second tracks. This is quite unlikely now since tracks are still mostly longer than 60 seconds. But, assuming that, we've got 50 minutes to not lose any listens by the user. If we check inactive listeners once every 30 minutes we should be fine. 

I think the key thing is to track the last timestamp -- we can fine tune the algorithm quickly based on feedback as long as we can differentiate active from inactive users.

paramsingh:
https://github.com/metabrainz/listenbrainz-server/commit/edbe6034893c151e98804736325c4e78976b06d3 and https://github.com/metabrainz/listenbrainz-server/commit/12cb19802e3678be74e7363de7271d98baf2240f add the `listened at` column.

LeoVerto:
@mayhem Okay, if we are able to hit each users at least once an hour we should be mostly on the safe side. I had assumed that checking each of our users within that timeframe would run into API limits.

Can we monitor instances of possible recording loss (i.e. lack of overlap between recordings) or would that be outside the scope of this PR? In the end that's the metric we want to optimize, right?

mayhem:
@LeoVerto Outside of the scope of this ballooning PR -- can you please add a ticket and we'll get that taken care of after this is merged?

paramsingh:
[LB-375](https://tickets.metabrainz.org/browse/LB-375)

listenbrainz/spotify_updater/spotify_read_listens.py
paramsingh: [LB-376](https://tickets.metabrainz.org/browse/LB-376)

PR #425, opened 2018-07-05
Use flask app logger and config in scripts

paramsingh:
# Summary

* This is an *improvement*
* **Describe this change in 1-2 sentences**: The first part of a series of PRs to improve logging and error reporting for LB.

# Problem

* JIRA ticket (_optional_): 

Our major scripts (`influx-writer` and `bigquery-writer`) did not report errors to Sentry at all.

# Solution

Instead of creating an independent logger, just run the script in the flask app context so that
errors are reported to sentry automatically.

### DIFF
--- a/listenbrainz/bigquery_writer/bigquery_writer.py
+++ b/listenbrainz/bigquery_writer/bigquery_writer.py
@@ -9,10 +9,9 @@ import sys
 import ujson
 import time
 
-from listenbrainz.utils import safely_import_config
+from listenbrainz.webserver import create_app
 
-safely_import_config()
-
+from flask import current_app
 from googleapiclient import discovery
 from googleapiclient.errors import HttpError
 from listenbrainz.listen_writer import ListenWriter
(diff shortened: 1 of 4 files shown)


listenbrainz/bigquery_writer/bigquery_writer.py
```diff
@@ -69,17 +67,15 @@ def submit_data(self):
...
-                    datasetId=self.config.BIGQUERY_DATASET_ID,
-                    tableId=self.config.BIGQUERY_TABLE_ID,
+                    projectId=current_app.config['BIGQUERY_PROJECT_ID'],
+                    datasetId=current_app.config['BIGQUERY_DATASET_ID'],
+                    tableId=current_app.config['BIGQUERY_TABLE_ID'],
                     body=body).execute(num_retries=5)
                 break
             except HttpError as e:
-                self.log.error("Submit to BigQuery failed: %s. Retrying in 3 seconds." % str(e))
+                current_app.logger.error("Submit to BigQuery failed: %s. Retrying in 3 seconds. Data: %s", str(e), json.dumps(body, indent=3), exc_info=True)
```
mayhem: should we be using ujson here since it was imported?

paramsingh: `json.dumps` has an `indent=3` option which leads to more readable json, that's why i used it.

PR #434, opened 2018-10-24
Make sure that old lists in the database are parsed correctly too

paramsingh:
there are some lists in the database that were converted to  via str(list) so they can't be loaded via json.

Example: "['Blank & Jones']"

However, yaml parses them safely and correctly


mayhem:
Lol, my listens are breaking shit.

PR #435, opened 2018-10-24
Import DatabaseException before using it

paramsingh:
Small bugfix from an error in sentry.


paramsingh:
https://sentry.metabrainz.org/metabrainz/listenbrainz-beta/issues/15714/

PR #439, opened 2018-10-25
Use localhost instead of 0.0.0.0 in spotify callback url

paramsingh:
After the discussion on https://github.com/metabrainz/listenbrainz-server/pull/436#discussion_r228117531

PR #440, opened 2018-10-25
Add documentation on how to setup Spotify Importer

paramsingh:
As discussed in #439

### DIFF
--- a/docs/dev/devel-env.rst
+++ b/docs/dev/devel-env.rst
@@ -142,6 +142,12 @@ section of the file.
     SPOTIFY_CLIENT_ID = ''
     SPOTIFY_CLIENT_SECRET = ''
 
+.. note::
+
+    The hostname on the callback URL must be the same as the host you use to
+    access your development server. If you use something other than ``localhost``, you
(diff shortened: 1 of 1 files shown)


docs/dev/devel-env.rst
```diff
@@ -127,9 +127,25 @@ In case you don't have a Last.FM API key, you can get it from `Last.FM API page`
 
 You also need to update the ``API_URL`` field value to ``http://localhost``.
 
+We also have a Spotify importer script which imports listens from
+Spotify automatically using the Spotify API. In order to run this in your
+local development environment, you'll have to register an application on the
+`Spotify Developer Dashboard`_. Use ``http://localhost/profile/connect-spotify/callback``
+as the callback URL.
```
alastair: Something about how to change the config file to change the callback url if you want/need the hostname to be different?

"The hostname on this callback URL must be the same as the host that you access your development server from. If this is different then you should also change SPOTIFY_CALLBACK_HOST blah blah blah"

paramsingh:
@alastair, done.

PR #445, opened 2018-10-29
Remove pika pool and use own connection pooling

paramsingh:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Pika pool's connection pooling wasn't working and a new connection was being created
for every `submit-listens` request.

# Solution

I spent a bunch of time trying to debug this but didn't really get anywhere.

So, I just wrote a simple one for us.


listenbrainz/webserver/rabbitmq_connection.py
```diff
@@ -26,11 +26,41 @@ def init_rabbitmq_connection(app):
...
 
-    _rabbitmq = pika_pool.QueuedPool(
-        create=lambda: pika.BlockingConnection(connection_parameters),
-        max_size=100,
-        max_overflow=10,
-        timeout=10,
-        recycle=3600,
-        stale=45,
-    )
+    _rabbitmq = RabbitMQConnectionPool(connection_parameters, 10)
```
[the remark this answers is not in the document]
mayhem: No idea, really. Let's make it configurable and start with 20 connections. Also add a sentry log message for when we're trying to get a connection, but none are available, so we can know to tune the setting.

listenbrainz/webserver/rabbitmq_connection.py
```diff
@@ -26,11 +26,44 @@ def init_rabbitmq_connection(app):
...
+            self.log.error('Tried to add a new connection into a full queue...', exc_info=True)
+
+    def get(self):
+        while True:
+            try:
+                connection, channel = self.queue.get_nowait()
+                if connection.is_open:
+                    return connection, channel
+                else:
+                    return self.create()
```
mayhem: Instead of creating a new connection right away if a connection is closed, I think it makes more sense to try and fetch the next possibly open connection in the queue and see if it is open and if so use it. That should reduce the overall number of connections used, no?

listenbrainz/webserver/views/api_tools.py
```diff
@@ -326,15 +325,16 @@ def publish_data_to_queue(data, exchange, queue, error_msg):
...
-        with rabbitmq_connection._rabbitmq.acquire() as cxn:
-            cxn.channel.exchange_declare(exchange=exchange, exchange_type='fanout')
-            cxn.channel.queue_declare(queue, durable=True)
-            cxn.channel.basic_publish(
-                exchange=exchange,
-                routing_key='',
-                body=ujson.dumps(data),
-                properties=pika.BasicProperties(delivery_mode=2, ),
-            )
+        cxn, channel = rabbitmq_connection._rabbitmq.get()
```
mayhem: I gotta say, I like the concept of the with ... as: statement here. It isn't hard to implement -- do you think we should do this?

PR #448, opened 2018-11-01
Retry some times while trying to refresh user tokens

paramsingh:
There was a sentry error for this in the Spotify importer.

### DIFF
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -5,6 +5,8 @@ import spotipy.oauth2
 from listenbrainz.db import spotify as db_spotify
 import datetime
 
+SPOTIFY_API_RETRIES = 5
+
 
 class Spotify:
     def __init__(self, user_id, musicbrainz_id, user_token, token_expires, refresh_token,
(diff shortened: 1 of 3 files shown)


listenbrainz/domain/spotify.py
```diff
@@ -68,7 +68,17 @@ def refresh_user_token(spotify_user):
         user (domain.spotify.Spotify): the same user with updated tokens
     """
     auth = get_spotify_oauth()
-    new_token = auth.refresh_access_token(spotify_user.refresh_token)
+
+    retries = 5
```
mayhem: either make this a config or a constant we can define at the top. Burying this in the code is no good.

listenbrainz/domain/spotify.py
```diff
@@ -68,7 +68,17 @@ def refresh_user_token(spotify_user):
...
+    new_token = None
+    while retries > 0:
+        new_token = auth.refresh_access_token(spotify_user.refresh_token)
+        if new_token is not None:
```
mayhem: Or just `if new_token:`...

PR #452, opened 2018-11-05
LB-393: Upgrade flask to 1.0.2

Ge0rg3:
Moving to new PR due to issues with my commit.


Ge0rg3:
Please note that the reason that requirements.txt looks as if much has been changed is only because I hadn't previously updated it. The only line that has actually been changed there is:
```
git+https://github.com/metabrainz/brainzutils-python.git@v1.5.0
```
(Which was updated to: )
```
git+https://github.com/metabrainz/brainzutils-python.git@v1.7.0
```

PR #453, opened 2018-11-05
LB-393: Upgrade flask to 1.0.2

Ge0rg3:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
Updating Flask to 1.0.4.

# Problem

The previous version of Flask was outdated, so this updates the BrainzUtils package (and therefore Flask). 
* JIRA ticket: [LB-393](https://tickets.metabrainz.org/browse/LB-393)

# Solution

The response in one of the test cases includes debug tools data, including text such as:
```
CPU: 144.13ms (159.74ms)
```
This means that the two responses will never be identical.
By comparing only the first 100 characters, we compare enough to know whether the case sensitivity worked, and we also ignore all of the debugging info.

### DIFF
--- a/listenbrainz/webserver/views/test/test_user.py
+++ b/listenbrainz/webserver/views/test/test_user.py
@@ -8,6 +8,7 @@ from influxdb import InfluxDBClient
 from listenbrainz.db.testing import DatabaseTestCase
 from listenbrainz.listenstore.tests.util import create_test_data_for_influxlistenstore
 from listenbrainz.webserver.influx_connection import init_influx_connection
+from listenbrainz.webserver.login import User
 from listenbrainz.webserver.testing import ServerTestCase
 
 import listenbrainz.db.user as db_user
@@ -18,8 +19,6 @@ class UserViewsTestCase(ServerTestCase, DatabaseTestCase):
     def setUp(self):
         ServerTestCase.setUp(self)
         DatabaseTestCase.setUp(self)
-        self.user = db_user.get_or_create(1, 'iliekcomputers')
-        self.weirduser = db_user.get_or_create(2, 'weird\\user name')
 
         self.log = logging.getLogger(__name__)
         self.influx = InfluxDBClient(
@@ -39,6 +38,14 @@ class UserViewsTestCase(ServerTestCase, DatabaseTestCase):
             'INFLUX_DB_NAME': current_app.config['INFLUX_DB_NAME'],
         })
 
+        db_user.get_or_create(1, 'iliekcomputers')
+        user = db_user.get_by_mb_id('iliekcomputers')
+        self.user = User.from_dbrow(user)
+
+        db_user.get_or_create(2, 'weird\\user name')
+        weirduser = db_user.get_by_mb_id('weird\\user name')
+        self.weirduser = User.from_dbrow(weirduser)
+
     def tearDown(self):
         self.influx.query('''drop database %s''' % current_app.config['INFLUX_DB_NAME'])
         self.logstore = None
@@ -47,25 +54,25 @@ class UserViewsTestCase(ServerTestCase, DatabaseTestCase):
         DatabaseTestCase.tearDown(self)
 
     def test_user_page(self):
-        response = self.client.get(url_for('user.profile', user_name=self.user['musicbrainz_id']))
+        response = self.client.get(url_for('user.profile', user_name=self.user.musicbrainz_id))
         self.assert200(response)
         self.assertContext('section', 'listens')
 
         # check that artist count is not shown if stats haven't been calculated yet
-        response = self.client.get(url_for('user.profile', user_name=self.user['musicbrainz_id']))
+        response = self.client.get(url_for('user.profile', user_name=self.user.musicbrainz_id))
         self.assert200(response)
         self.assertTemplateUsed('user/profile.html')
         self.assertContext('artist_count', None)
 
         # check that artist count is shown if stats have been calculated
         db_stats.insert_user_stats(
-            user_id=self.user['id'],
+            user_id=self.user.id,
             artists={},
             recordings={},
             releases={},
             artist_count=2,
         )
-        response = self.client.get(url_for('user.profile', user_name=self.user['musicbrainz_id']))
+        response = self.client.get(url_for('user.profile', user_name=self.user.musicbrainz_id))
         self.assert200(response)
         self.assertTemplateUsed('user/profile.html')
         self.assertContext('artist_count', '2')
@@ -73,9 +80,9 @@ class UserViewsTestCase(ServerTestCase, DatabaseTestCase):
     def test_scraper_username(self):
         """ Tests that the username is correctly rendered in the last.fm importer """
         response = self.client.get(
-            url_for('user.lastfmscraper', user_name=self.user['musicbrainz_id']),
+            url_for('user.lastfmscraper', user_name=self.user.musicbrainz_id),
             query_string={
-                'user_token': self.user['auth_token'],
+                'user_token': self.user.auth_token,
                 'lastfm_username': 'dummy',
             }
         )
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/test/test_user.py
```diff
@@ -131,7 +131,7 @@ def test_username_case(self):
         response2 = self.client.get(url_for('user.profile', user_name='IlieKcomPUteRs'))
         self.assert200(response1)
         self.assert200(response2)
-        self.assertEqual(response1.data.decode('utf-8'), response2.data.decode('utf-8'))
+        self.assertEqual(response1.data.decode('utf-8')[:100], response2.data.decode('utf-8')[:100])
```
alastair: I wonder if we can be smarter here. 2 things come to mind:
 - We probably shouldn't enable the debug toolbar during testing - we probably want to be closer to "production" here rather than development. I don't know if there's a way to disable DEBUG during tests, or explicitly disable the debug toolbar. I'll look into it
- An alternative to checking the whole response page is to just check the user that we pass to the template. I know that in django you can check the template context in a test: https://docs.djangoproject.com/en/2.1/topics/testing/tools/#django.test.Response.context
Let's see if we can do the same in flask

Ge0rg3: Turning off debugging during tests sounds great, but I wonder why this error didn't occur during the previous version of flask? I'm not sure why it didn't run into this issue previously.

Ge0rg3: Looks like flask also has a `render_template()` function :)

Ge0rg3: It might be worth just doing:
```python
        self.assertTrue("iliekcomputers" in response1.data.decode('utf-8'))
        self.assertTrue("iliekcomputers" in response2.data.decode('utf-8'))
```
For readability & efficiency?
I imagine that searching for that exact string would be far faster than rendering the template to check it.

paramsingh: There is a `self.assertIn` in Flask-Testing

alastair: The django method doesn't test the rendered html - instead it gives you the actual parameter that was passed to the template ('user' in this case). This test is really only that you get the correct username even if the capitalisation is different, so ideally we should only check this user object.

Ge0rg3: Ohh okay that makes more sense 👍

paramsingh: https://github.com/metabrainz/listenbrainz-server/blob/60f66aea236901ae225b7b495d4da0afc9fb9fb7/listenbrainz/webserver/views/test/test_user.py#L58

paramsingh: I think we're talking about `self.assertContext`

alastair: perfect, that's exactly what I mean! I didn't know what it was called in flask. let's replace this assertEqual with an assertContext for the 'user' parameter.
@Ge0rg3 has a good point though - it's interesting to know why this didn't happen when we were using older versions of flask. I know they made some changes to the way that debug mode works some time around flask 0.11 or 0.12, but I'm not sure exactly what these changes were

Ge0rg3: That looks good, but I'm currently getting:
```
AssertionError: Context variable does not exist: user.musicbrainz_id
```
From what I can tell from the [template](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/templates/user/profile.html), "user" is an object that contains the ID that we want.
So it doesn't look like assertContext can do properties. Might be worth [getting the context variable itself](https://github.com/jarus/flask-testing/blob/master/flask_testing/utils.py#L253) and checking the musicbrainz_id from there.

Ge0rg3: Okay--I've got it working as follows:
```python
self.assertEqual(self.get_context_variable('user').musicbrainz_id, "iliekcomputers")
```
Am I okay to commit that, or is there any way I can improve it? :)

alastair: Is there a reason why `self.assertContext('user', self.user)` doesn't work? We should be able to directly check these objects instead of accessing fields within them.

Ge0rg3: From what I can tell from the [assertContext source code](https://github.com/jarus/flask-testing/blob/master/flask_testing/utils.py#L283), it can't read the (sub?)attributes of user, since we need to read the `musicbrainz_id` attribute of `user`. It also uses the get_context_variable function itself.

alastair: I see what you're trying to do. I don't think that we need to specifically compare the `musicbrainz_id` attribute. Instead, we can directly compare the user object, which uses `flask_login`'s UserMixin and its equals method (to compare user id) https://flask-login.readthedocs.io/en/latest/_modules/flask_login/mixins.html#UserMixin

Ge0rg3: (See edited comment above)

alastair: you're right that the definition of "equals" could be different. We are lucky because our User object is a flask_login UserMixin. This allows us to use the definition of `__eq__` which checks `self.id`. If we were not using this, we could have an issue where the two objects that we wanted to test might have been different.

alastair: You should make sure that you add 2 `self.asserContext`s. One for each `self.client.get`. These checks should be directly after the get - because the context changes after each one

Ge0rg3: Ah okay, will do :)

Ge0rg3: Interesting, I'm getting this error when running the tests:
```

    self.assertEqual(self.get_context_variable(name), value)
E   AssertionError: <listenbrainz.webserver.login.User object[14 chars]cfd0> != (1, datetime.datetime(2018, 11, 6, 11, 14[348 chars]e, 1)
```

When at the line:
```
        response1 = self.client.get(url_for('user.profile', user_name='iliekcomputers'))
>       self.assertContext('user', self.user)
```

Ge0rg3: So at that stage, self.user is equal to:
```
(1, datetime.datetime(2018, 11, 6, 11, 31, 17, 726742, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), 'iliekcomputers', '5b879824-9591-48dd-82e6-268400d77633', datetime.datetime(2018, 11, 6, 11, 31, 17, 726742, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), datetime.datetime(1970, 1, 1, 0, 0, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None)), None, 1)
```

Ge0rg3: > we could have an issue where the two objects that we wanted to test might have been different.

I think that this might actually be the case :/ @alastair

Ge0rg3: Thanks for the info. I'll get to work on that 👍

Ge0rg3: @alastair I've implemented these channges, and added them in the new commit.
Just for reference, I used `get_or_create` instead of just `create` so that the user is only created if it doesn't exist (and it saves me doing a database search for it).

listenbrainz/webserver/views/test/test_user.py
```diff
@@ -39,6 +38,14 @@ def setUp(self):
             'INFLUX_DB_NAME': current_app.config['INFLUX_DB_NAME'],
         })
 
+        db_user.get_or_create(1, 'iliekcomputers')
+        user = db_user.get_by_mb_id('iliekcomputers')
```
alastair: We don't need to have this line twice, `get_or_create` will also return the db row, so we can just use `user = get_or_create(blah)`

Ge0rg3: Okay, fixed.

PR #456, opened 2018-11-14
Make sure the webserver stays up even if RabbitMQ is down.

paramsingh:
Trille went down today, which led to the ListenBrainz app not being created and listenbrainz.org erroring out.

These changes should make it so that the webserver still works but the rabbitmq thing is logged to sentry and 503s are returned until RabbitMQ comes back.

### DIFF
--- a/listenbrainz/webserver/rabbitmq_connection.py
+++ b/listenbrainz/webserver/rabbitmq_connection.py
@@ -19,12 +19,14 @@ def init_rabbitmq_connection(app):
     # if RabbitMQ config values are not in the config file, the app should
     # wait for them to come back. Consul will bring the values back into
     # config once the RabbitMQ service comes up.
-    while True:
+    for _ in range(CONNECTION_RETRIES):
         if "RABBITMQ_HOST" not in app.config:
             app.logger.critical("RabbitMQ host:port not defined. Sleeping 2 seconds and trying again...")
             sleep(2)
         else:
             break
+    else:
+        raise ConnectionError("RabbitMQ service is not up!")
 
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/rabbitmq_connection.py
```diff
@@ -1,4 +1,3 @@
-import sys
 import queue
 from time import sleep
 import pika
```
mayhem: pika?

listenbrainz/webserver/rabbitmq_connection.py
```diff
@@ -16,16 +16,11 @@ def init_rabbitmq_connection(app):
...
     # config once the RabbitMQ service comes up.
-    for _ in range(CONNECTION_RETRIES):
-        if "RABBITMQ_HOST" not in app.config:
-            app.logger.critical("RabbitMQ host:port not defined. Sleeping 2 seconds and trying again...")
-            sleep(2)
-        else:
-            break
-    else:
+    if "RABBITMQ_HOST" not in app.config:
+        app.logger.critical("RabbitMQ host:port not defined. Sleeping 2 seconds and trying again...")
```
mayhem: Message isn't quite correct. :-p

listenbrainz/webserver/rabbitmq_connection.py
paramsingh: `pika` is still used, we removed `pika_pool`

mayhem: Oh, right. Thanks.

PR #457, opened 2018-11-17
Allow deletion of users who have activated the spotify import

paramsingh:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: You can't delete your account if spotify import is activated. I've added an `ON DELETE CASCADE` to the fk constraint to fix this.


paramsingh:
https://sentry.metabrainz.org/metabrainz/listenbrainz-beta/issues/17209/

PR #463, opened 2018-11-22
Fix bug in the UUID validation function

paramsingh:
uuid.UUID can raise attribute errors if passed an integer

https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/7140/


alastair:
`uuid.UUID(datetime.datetime.now())` raises `TypeError`. why not test `uuid.UUID(str(u))` instead?

paramsingh:
I really should have thought of that.

_me gets to making new PR_

PR #467, opened 2018-12-11
Email notifications to users on spotify errors

paramsingh:
add a feature emailing users when we stop importing  their listens for some reason or error.


listenbrainz/config.py.sample
```diff
@@ -19,6 +19,7 @@ PG_QUERY_TIMEOUT = "3000"
 # Set to True to enable 'synchronous_commit' for Postgres. Default: False
 PG_ASYNC_LISTEN_COMMIT = False
 
+MUSICBRAINZ_DB_URI = ""
```
mayhem: Does this need to be set in the ctmpl above?

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -14,6 +14,28 @@
...
+    """ Notifies specified user via email about error during Spotify import.
+
+    Args:
+        musicbrainz_row_id (int): the MusicBrainz row ID of the user
+        error (str): a description of the error encountered.
+    """
+    user_email = mb_editor.get_editor_by_id(musicbrainz_row_id)['email']
+    text = ('Hi, we encountered an error while importing your listens from Spotify.'
+            'The error was as follows: "%s" '
+            'Please take a look at your profile page for more information.\n'
```
mayhem: Can we include a link to their profile in the mail?

mayhem: What email address is the email being sent from?

paramsingh: We can set it here: https://github.com/metabrainz/listenbrainz-server/pull/467/files#diff-1e4b2ed41bf1b80350b586a316431d06R141

I have it currently as `listenbrainz.org` so it tries to send the mail from `noreply@listenbrainz.org`, I intended to ask if there's more setup needed for the email.

listenbrainz/config.py.sample
paramsingh: We have it already set in the consul template from when we were importing musicbrainz row IDs.

https://github.com/metabrainz/listenbrainz-server/blob/master/consul_config.py.ctmpl#L24

listenbrainz/spotify_updater/spotify_read_listens.py
mayhem: noreply is great -- consistent with our other services. That's all I wanted to hear. :)

listenbrainz/config.py.sample
mayhem: kewl.

mayhem:
OK, add the link and I'll give it another glance before we merge.

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -14,6 +14,28 @@
...
+
+def notify_error(musicbrainz_row_id, error):
+    """ Notifies specified user via email about error during Spotify import.
+
+    Args:
+        musicbrainz_row_id (int): the MusicBrainz row ID of the user
+        error (str): a description of the error encountered.
+    """
+    user_email = mb_editor.get_editor_by_id(musicbrainz_row_id)['email']
+    text = ('Hi, we encountered an error while importing your listens from Spotify.'
```
alastair: I typically don't use multiline strings like this -- instead I'd use `"""`. However, as an alternative we could use jinja here and have a separate template file for the email. It might be nice to make it more consistent like other templates

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -234,6 +256,7 @@ def process_all_spotify_users():
                 success=False,
                 error_message=str(e),
             )
+            notify_error(u.musicbrainz_row_id, str(e))
```
alastair: To confirm, this means that we'll send an email to a user a maximum of one time? (because we'll also set `success=False`, so we won't process this user again). I want to make sure that we don't send a user an email every 10 minutes.

How will a user "reset" their account to reprocess it? Will we remove the last-updated row? or will something set `success`?

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -252,6 +275,7 @@ def process_all_spotify_users():
 def main():
     app = listenbrainz.webserver.create_app()
     with app.app_context():
+        musicbrainz_db.init_db_engine(current_app.config['MUSICBRAINZ_DB_URI'])
```
alastair: Why is this not done in `create_app()`? (https://github.com/metabrainz/listenbrainz-server/blob/72694d6ef073bfa2a4c0506e2ca8ee39348c5a7a/listenbrainz/webserver/__init__.py#L116)
Why does this script specifically need musicbrainz access? (and not the rest of LB?)

listenbrainz/spotify_updater/spotify_read_listens.py
paramsingh: Yes, if `success == False` then the user isn't processed again, there are tests for this.

Right now, the user would just unlink their account and relink again. There is probably a better flow for this though.

listenbrainz/spotify_updater/spotify_read_listens.py
paramsingh: Changed to use jinja

listenbrainz/spotify_updater/spotify_read_listens.py
paramsingh: Moved to create_app. I had some flawed thinking for keeping it in spotify_read_listens, right now this is the only script that requires access to the MB db to get user emails, so I was trying to keep it here, but we'll have to eventually move it to create_app anyways, so that is the best place.

PR #468, opened 2018-12-12
use --no-cache-dir in pip install during docker build

alastair:
In Docker the pip cache uses space and doesn't make anything faster, so don't use it

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,19 +25,19 @@ RUN apt-get update \
 RUN mkdir /code
 WORKDIR /code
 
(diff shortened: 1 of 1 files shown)


alastair:
You're effectively doubling the size of python packages that you're installing (once in the cache, once in site-packages).
On listenbrainz this looks like it's ~100MB:
```
root@7beef0a4045f:/usr/local/lib/python3.6# du -sh site-packages/
91M    site-packages/
```

mayhem:
I'm copying this to other projects, thanks!

PR #469, opened 2018-12-13
LB-401: Fix logging config format

paramsingh:
The logging config values needed has changed with us starting to use brainzutils loggers. This PR changes the config.py.sample to reflect this change, so that users have the correct format to fill.

### DIFF
--- a/listenbrainz/config.py.sample
+++ b/listenbrainz/config.py.sample
@@ -94,6 +94,7 @@ PLAYING_NOW_MAX_DURATION = 10 * 60
 # LOG_SENTRY = {
 #    'dsn':'',
 #    'environment': 'development',
+#    'level': 'ERROR',
 # }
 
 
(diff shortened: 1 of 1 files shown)


listenbrainz/config.py.sample
```diff
@@ -73,15 +73,29 @@ PLAYING_NOW_MAX_DURATION = 10 * 60
 
 # LOGGING
```
alastair: In my changes to this in AB I added an extra few lines here:

```
# Uncomment any of the following logging stubs if you want to enable logging
# during development. In general you shouldn't need to do this
```

listenbrainz/config.py.sample
```diff
@@ -73,15 +73,29 @@ PLAYING_NOW_MAX_DURATION = 10 * 60
...
+#    ],
+#    'log_email_topic': 'ListenBrainz Error',
+#    'level': 'ERROR',
+# }
+
+# LOG_SENTRY = {
+#    'dsn':'',
+#    'environment': 'development',
+# }
+# SENTRY_DSN_PUBLIC = ''
```
alastair: What is this? is this for sentry reporting from javascript? It doesn't look like it's used anywhere in LB yet. What do we prefer to do? Leave it here for consistency over all projects, or skip it until it's required?

PR #470, opened 2018-12-17
Introduced 'Remember me' functionality

vansika:
* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Some users note that they get logged out of ListenBrainz quite often.

# Solution
Introduced a 'Remember me' checkbox and linked it with 'Sign in with Musicbrainz' button. Created a form to send checkbox data to server. If the checkbox is checked then the user will be logged in for specified duration even if the browser is accidentally closed.
Have cleared session before logging out to clear the 'Remember me' token that hinders signing out.

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -12,6 +12,7 @@ CREATE TABLE "user" (
 );
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id);
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_row_id_key UNIQUE (musicbrainz_row_id);
+ALTER TABLE "user" ADD COLUMN user_login_id VARCHAR;
 
 CREATE TABLE spotify_auth (
   user_id                   INTEGER NOT NULL, -- PK and FK to user.id
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -8,11 +8,11 @@ CREATE TABLE "user" (
   last_login     TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
   latest_import  TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT TIMESTAMP 'epoch',
   gdpr_agreed    TIMESTAMP WITH TIME ZONE,
-  musicbrainz_row_id INTEGER NOT NULL
+  musicbrainz_row_id INTEGER NOT NULL,
+  user_login_id  UUID NOT NULL DEFAULT uuid_generate_v4()
 );
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id);
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_row_id_key UNIQUE (musicbrainz_row_id);
-ALTER TABLE "user" ADD COLUMN user_login_id VARCHAR;
 
 CREATE TABLE spotify_auth (
   user_id                   INTEGER NOT NULL, -- PK and FK to user.id
--- a/admin/sql/updates/2019-01-04-add-user-login-id.sql
+++ b/admin/sql/updates/2019-01-04-add-user-login-id.sql
@@ -1,0 +1,6 @@
+BEGIN;
+
+-- Add column for alternative user id for login purpose to user table
+ALTER TABLE "user" ADD COLUMN user_login_id VARCHAR;
+
+COMMIT;
--- a/admin/sql/updates/2019-01-04-add-user-login-id.sql
+++ b/admin/sql/updates/2019-01-04-add-user-login-id.sql
@@ -1,6 +1,6 @@
 BEGIN;
 
 -- Add column for alternative user id for login purpose to user table
-ALTER TABLE "user" ADD COLUMN user_login_id VARCHAR;
+ALTER TABLE "user" ADD COLUMN user_login_id UUID NOT NULL DEFAULT uuid_generate_v4();
 
 COMMIT;
--- a/listenbrainz/config.py.sample
+++ b/listenbrainz/config.py.sample
@@ -145,3 +145,7 @@ TESTING = True # this flag is used by the mail module to check if mails need to 
 SMTP_SERVER = 'listenbrainz-mail'
 SMTP_PORT = 25
 MAIL_FROM_DOMAIN = 'listenbrainz.org'
+
+# expiration of 'Remember me' cookie
+SESSION_REMEMBER_ME_DURATION = 365
+
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -26,13 +26,14 @@ def create(musicbrainz_row_id, musicbrainz_id):
     """
     with db.engine.connect() as connection:
         result = connection.execute(sqlalchemy.text("""
-            INSERT INTO "user" (musicbrainz_id, musicbrainz_row_id, auth_token)
-                 VALUES (:mb_id, :mb_row_id, :token)
+            INSERT INTO "user" (musicbrainz_id, musicbrainz_row_id, auth_token, user_login_id)
+                 VALUES (:mb_id, :mb_row_id, :token, :login_id)
               RETURNING id
         """), {
             "mb_id": musicbrainz_id,
             "token": str(uuid.uuid4()),
             "mb_row_id": musicbrainz_row_id,
+            "login_id" : str(uuid.uuid4()),
         })
         return result.fetchone()["id"]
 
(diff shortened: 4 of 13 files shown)


vansika:
Hello @alstair

I implemented alternate id, making changes in database and other files. Everything is working fine on localhost, when I change alternative id I get logged out otherwise the session is restored according to remember me cookie. 
The logs seem fine yet there are errors while passing tests. The get_id method of user class is returning alt_id which is a string (have checked that) yet the tests say that it returns an integer.
Any help would be great. 
Thankyou.

admin/sql/create_tables.sql
```diff
@@ -12,6 +12,7 @@ CREATE TABLE "user" (
 );
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id);
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_row_id_key UNIQUE (musicbrainz_row_id);
+ALTER TABLE "user" ADD COLUMN alt_id VARCHAR;
```
paramsingh: @vansika, the bug seems to be here, you're declaring the `alt_id` column as varchar, but then comparing it with integers later.

Also, this column should be inside the `CREATE TABLE` statement instead of being a seperate `ALTER` statement.

listenbrainz/config.py.sample
```diff
@@ -117,3 +117,6 @@ SPOTIFY_CALLBACK_URL = 'http://localhost/profile/connect-spotify/callback'
 # We do not use this feature at all, so it is safe to set to False.
 # See https://stackoverflow.com/a/33790196 for more details.
 SQLALCHEMY_TRACK_MODIFICATIONS = False
+
+# expiration of 'Remember me' cookie
+days = 365
```
paramsingh: This name can be more descriptive. `SESSION_REMEMBER_ME_DURATION` maybe.

listenbrainz/db/user.py
```diff
@@ -84,6 +84,31 @@ def get(id):
...
+        }
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT {columns}
+              FROM "user"
+             WHERE alt_id = :alt_id
+        """.format(columns=','.join(USER_GET_COLUMNS))), {"alt_id": alt_id})
+        row = result.fetchone()
+        print (row)
```
paramsingh: This print can probably be removed.

listenbrainz/webserver/views/login.py
```diff
@@ -29,7 +30,7 @@ def musicbrainz_post():
     if provider.validate_post_login():
         user = provider.get_user()
         db_user.update_last_login(user.musicbrainz_id)
-        login_user(user)
+        login_user(user, remember=True, duration=datetime.timedelta(days=config.days))
```
paramsingh: please use `current_app.config['days']` here instead of importing config explicitly.

listenbrainz/config.py.sample
```diff
@@ -98,7 +98,7 @@ MAX_CONTENT_LENGTH = 16 * 1024 * 1024  # 16MB
 UPLOAD_FOLDER = "/tmp/lastfm-backup-upload"
 
 
-API_URL = 'https://api.listenbrainz.org'
+API_URL = 'http://localhost'
```
paramsingh: this change isn't needed, in my opinion

admin/sql/create_tables.sql
paramsingh: Please note that the automated tests on travis do a fresh build of the code you push, while your local environment may be different because of experimentation etc.

paramsingh:
Also, we try to do descriptive commit messages in comparison to `Update <file>`. See this guide for examples: https://juffalow.com/other/write-good-git-commit-message

admin/sql/create_tables.sql
vansika: @paramsingh, get_id( ) passes alt_id to load_user( ) which is a string, but the redirects in tests are passing user_id to load_user( ), which is the cause of error because the check function (filter_by_alt_id( )) inside load_user( ) can return user by checking against alt_id (string) and not user_id (integer). 
I will try to debug. Thank you

vansika:
@alastair 
I have implemented alternate tokens and the other requirements of the PR except disabling session cookies on API requests. API auth doesn't seem to involve flask-login, how exactly do i disable it.

admin/sql/create_tables.sql
```diff
@@ -12,6 +12,7 @@ CREATE TABLE "user" (
 );
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_id_key UNIQUE (musicbrainz_id);
 ALTER TABLE "user" ADD CONSTRAINT user_musicbrainz_row_id_key UNIQUE (musicbrainz_row_id);
+ALTER TABLE "user" ADD COLUMN user_login_id VARCHAR;
```
paramsingh: this column should be added inside the create table function call instead of creating the table and then altering it.

admin/sql/updates/2019-01-04-add-user-login-id.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+-- Add column for alternative user id for login purpose to user table
+ALTER TABLE "user" ADD COLUMN user_login_id VARCHAR;
```
paramsingh: we will need this script to fill out the column for existing users also, because each user must have a login ID, no matter what.

admin/sql/create_tables.sql
paramsingh: Also, it should be `NOT NULL`, I guess.

listenbrainz/db/user.py
```diff
@@ -85,6 +86,30 @@ def get(id):
         row = result.fetchone()
         return dict(row) if row else None
 
+def get_by_user_login_id(login_id):
+    """Get user with a specified login ID.
+
+    Args:
+        id (int): login ID of a user.
```
paramsingh: this should be updated because login ID of the user isn't an integer.

admin/sql/create_tables.sql
paramsingh: Also, If the only thing we're filling this column with is UUIDs, we should use the postgres uuid data type instead. (https://www.postgresql.org/docs/9.6/datatype-uuid.html)

This would make it more efficient. And you could set a default value for all users as well.

vansika: "uuid-ossp" extension is not being created due to some missing files. I will try to fix it and then would use UUID and postgres functions to generate user_login_id.

paramsingh: I've fixed the problem in #479. Thanks for the bug report! As soon as it gets merged, you can rebase this PR and everything should work fine. In the meanwhile, you can just install the extension in your "listenbrainz" db and continue.

paramsingh: merged now. =)

alastair:
> I have implemented alternate tokens and the other requirements of the PR except disabling session cookies on API requests. API auth doesn't seem to involve flask-login, how exactly do i disable it.

From the documentation, https://flask-login.readthedocs.io/en/latest/#disabling-session-cookie-for-apis, this step only appears to be necessary if we use `header_loader` or `request_loader` for authentication in the API. We don't use these, instead we do authentication manually using [_validate_auth_header](https://github.com/metabrainz/listenbrainz-server/blob/505e1974ce5a209f937fa130409eabb13cf723bc/listenbrainz/webserver/views/api.py#L261), so it looks like this step is unnecessary.

vansika:
@alastair 
Okay, thank you. Is there any thing else I should implement?

listenbrainz/webserver/views/test/test_index.py
```diff
@@ -118,9 +118,13 @@ def test_menu_logged_in(self, mock_user_get):
         self.assertIn('iliekcomputers', data)
         self.assertIn('Import', data)
         # item in user menu
+
         self.assertIn('Your Listens', data)
         mock_user_get.assert_called_with(user['user_login_id'])
 
+        self.assertIn('My Listens', data)
```
mayhem: We've changed "Your Listens" to "My listens", so the "Your Listens" are no longer needed.

vansika: Okay

listenbrainz/webserver/views/login.py
```diff
@@ -41,6 +49,6 @@ def musicbrainz_post():
 @login_bp.route('/logout/')
 @login_required
 def logout():
-    logout_user()
     session.clear()
+    logout_user()
```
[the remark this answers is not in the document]
vansika: Session variables need to be cleared before logging out so that remember me cookie is deleted. Otherwise, user won't be able to log out.

admin/sql/updates/2019-01-04-add-user-login-id.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+-- Add column for alternative user id for login purpose to user table
+ALTER TABLE "user" ADD COLUMN user_login_id UUID NOT NULL DEFAULT uuid_generate_v4();
```
alastair: I just noticed that there's no index on this field, there should be if we use it for lookups

PR #472, opened 2018-12-21
Fix docs build error due to code block formatting

paramsingh:
There was no newline which meant that the the code block took
every line after that as arguments, leading to docs build errors.


alastair:
Looks good, while I was looking at it I took the chance to update some of the text to make it easier to understand

paramsingh:
Thanks for that!

PR #474, opened 2018-12-28
Only create extension uuid-ossp if it doesn't exist

vansika:
* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Error: (psycopg2.ProgrammingError) extension "uuid-ossp" already exists

# Solution
Changing "create extension" to "create extension if not exists" would create extension only if does not exists during the creation of database.


vansika:
@paramsingh 😊

PR #475, opened 2019-01-07
make timestamp optional for playing now listens in RedisListenStore

vansika:
* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

A random value is used as timestamp for 'playing now' listens. 

* JIRA ticket (_optional_): [LB-404](https://tickets.metabrainz.org/browse/LB-404)

# Solution

Timestamp value in Listen object is assigned after checking the 'Listened_at' value which would be None in case of 'playing_now' listens.

### DIFF
--- a/listenbrainz/listen.py
+++ b/listenbrainz/listen.py
@@ -6,7 +6,6 @@ import yaml
 
 from datetime import datetime
 from listenbrainz.utils import escape, convert_to_unix_timestamp
-from flask import current_app
 
 def flatten_dict(d, seperator='', parent_key=''):
     """
@@ -77,8 +76,8 @@ class Listen(object):
                 self.timestamp = timestamp
                 self.ts_since_epoch = calendar.timegm(self.timestamp.utctimetuple())
             else:
-                self.timestamp = 0
-                self.ts_since_epoch = 0
+                self.timestamp = None
+                self.ts_since_epoch = None
 
         self.artist_msid = artist_msid
         self.release_msid = release_msid
--- a/listenbrainz/listen.py
+++ b/listenbrainz/listen.py
@@ -6,6 +6,7 @@ import yaml
 
 from datetime import datetime
 from listenbrainz.utils import escape, convert_to_unix_timestamp
+from flask import current_app
 
 def flatten_dict(d, seperator='', parent_key=''):
     """
@@ -99,7 +100,10 @@ class Listen(object):
     @classmethod
     def from_json(cls, j):
         """Factory to make Listen() objects from a dict"""
-        if j['listened_at']:
+
+        if 'playing_now' in j:
+            j.update({'listened_at': None})
+        else:
             j['listened_at']=datetime.utcfromtimestamp(float(j['listened_at']))
         return cls(
             user_id=j.get('user_id'),
--- a/listenbrainz/listen.py
+++ b/listenbrainz/listen.py
@@ -99,10 +99,12 @@ class Listen(object):
     @classmethod
     def from_json(cls, j):
         """Factory to make Listen() objects from a dict"""
+        if j['listened_at']:
+            j['listened_at']=datetime.utcfromtimestamp(float(j['listened_at']))
         return cls(
             user_id=j.get('user_id'),
             user_name=j.get('user_name', ''),
-            timestamp=datetime.utcfromtimestamp(float(j['listened_at'])),
+            timestamp=j['listened_at'],
             artist_msid=j['track_metadata']['additional_info'].get('artist_msid'),
             release_msid=j['track_metadata']['additional_info'].get('release_msid'),
             recording_msid=j.get('recording_msid'),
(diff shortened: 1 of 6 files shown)


listenbrainz/listenstore/redis_listenstore.py
```diff
@@ -30,7 +30,7 @@ def get_playing_now(self, user_id):
         if not data:
             return None
         data = ujson.loads(data)
-        data.update({'listened_at': MIN_ID+1})
+        data.update({'listened_at': None})
```
paramsingh: Instead of directly putting None here, I would prefer it if we added a `playing_now` key to the data and then set the timestamp attribute in the from_json function there accordingly.

paramsingh: Also, we should look into removing MIN_ID altogether from the codebase. Is it used anywhere else?

listenbrainz/listen.py
```diff
@@ -99,10 +99,12 @@ def __init__(self, user_id=None, user_name=None, timestamp=None, artist_msid=Non
     @classmethod
     def from_json(cls, j):
         """Factory to make Listen() objects from a dict"""
+        if j['listened_at']:
```
paramsingh: I would like a test for `from_json` where we create pass a json file with `playing_now` set to true and check that the timestamp is None and then pass a json file for a normal listen (non `playing_now`) and check that the timestamp is valid in the Listen object. 

You can add this test to `test_listen.py`.

listenbrainz/listenstore/redis_listenstore.py
vansika: MIN_ID has been used as 'listened_at' value for playing now listens. If we put appropriate checks on timestamp value according to listen type , MIN_ID can be removed from the codebase.

listenbrainz/listen.py
```diff
@@ -6,6 +6,7 @@
 
 from datetime import datetime
 from listenbrainz.utils import escape, convert_to_unix_timestamp
+from flask import current_app
```
paramsingh: this seems extraneous.

listenbrainz/listenstore/__init__.py
```diff
@@ -1,6 +1,5 @@
 import logging
 
-MIN_ID = 1033430400     # approx when audioscrobbler was created
```
paramsingh: woohoo! :tada: :tada:

listenbrainz/listenstore/tests/util.py
```diff
@@ -15,12 +15,17 @@ def generate_data(test_user_id, user_name, from_ts, num_records):
     test_data = []
     artist_msid = str(uuid.uuid4())
 
-    for i in range(num_records):
+    if (from_ts == None):  #check for playing now listens
```
paramsingh: We generally don't do parantheses in if conditions in Python. :)

this could be better written as `if from_ts is None`

listenbrainz/listenstore/tests/util.py
```diff
@@ -15,12 +15,17 @@ def generate_data(test_user_id, user_name, from_ts, num_records):
...
+    else:
         from_ts += 1   # Add one second
+        timestamp = datetime.utcfromtimestamp(from_ts)
+
+    for i in range(num_records):
         item = Listen(
             user_name=user_name,
             user_id=test_user_id,
-            timestamp=datetime.utcfromtimestamp(from_ts),
+            timestamp= timestamp,
```
paramsingh: not to nitpick, but there is an extra space here. [PEP-8](https://www.python.org/dev/peps/pep-0008/), the style guide we try to loosely follow, suggests spaces between operands but not between arguments in functions.

so ```timestamp=timestamp``` instead of the current line.

listenbrainz/tests/unit/test_listen.py
```diff
@@ -100,3 +100,37 @@ def test_to_influx(self):
         self.assertEqual(data['fields']['artist_name'], listen.data['artist_name'])
 
         self.assertIn('inserted_timestamp', data['fields'])
+
+
+    def test_from_json(self):
+        json_row = {
```
paramsingh: Could we possibly use `testdata/valid_single.json` here? (https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/testdata/valid_single.json). It would reduce the complexity and size of the test a lot.

listenbrainz/tests/unit/test_listen.py
```diff
@@ -100,3 +100,37 @@ def test_to_influx(self):
...
+        json_row.update({'listened_at': 123456})
+        listen = Listen.from_json(json_row) 
+
+        self.assertEqual(listen.timestamp, json_row['listened_at'])
+
+        del json_row['listened_at']
+        json_row.update({'playing_now': True})
+        listen = Listen.from_json(json_row)
+
+        self.assertEqual(listen.timestamp, 0)
```
paramsingh: Shouldn't this be None and not 0? we have actual listens in the db with timestamp 0.

vansika:
Happy to work☺

PR #478, opened 2019-01-15
Add dump function to dump in spark-friendly format

paramsingh:
As discussed on IRC.

### DIFF
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -20,13 +20,13 @@ from influxdb.exceptions import InfluxDBClientError, InfluxDBServerError
 from listenbrainz import DUMP_LICENSE_FILE_PATH
 from listenbrainz.db import DUMP_DEFAULT_THREAD_COUNT
 from listenbrainz.db.dump import SchemaMismatchException
-from listenbrainz.listen import Listen
+from listenbrainz.listen import Listen, convert_influx_row_to_spark_row
 from listenbrainz.listenstore import ListenStore
 from listenbrainz.listenstore import ORDER_ASC, ORDER_TEXT, \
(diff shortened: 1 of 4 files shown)


listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -348,13 +348,14 @@ def fetch_listens_from_storage(self, user_name, from_ts, to_ts, limit, order):
...
 
-    def dump_user(self, username, fileobj, dump_time):
+    def dump_user(self, username, fileobj, dump_time, spark_format=False):
         """ Dump specified user's listens into specified file object.
 
         Args:
             username (str): the MusicBrainz ID of the user whose listens are to be dumped
             fileobj (file): the file into which listens should be written
             dump_time (datetime): the time at which the specific data dump was initiated
+            spark_format (bool): dump files in Apache Spark friendly format if True, else full dumps
```
alastair: would it be possible to do these dumps without making the dump methods have to understand an implementation detail (spark)? e.g., could you have a method which converts the result of `to_api()` to the spark format, and you call that in the spark dumper?

mayhem: In general I agree with this, but this may end up being premature optimization. Part of it is that we simply do not yet know what what we really need and how we need it, in order to feed Spark. At this point I would be happy to enter a ticket for us to revisit this and proceed for now.

listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -347,6 +347,73 @@ def fetch_listens_from_storage(self, user_name, from_ts, to_ts, limit, order):
 
         return listens
 
+    def get_listens_batch_for_dump(self, username, dump_time, offset):
```
mayhem: I'm not convinced that this is the fastest way to do this -- but I think it is good enough for now. Like my other comment, we don't know what is going to stick just yet.

PR #480, opened 2019-01-19
LB-414: Spotify now playing

paramsingh:
# Summary

* This is a Feature Addtion

* **Describe this change in 1-2 sentences**: Make the spotify importer send now playing listens 

# Problem

* JIRA ticket (_optional_): [LB-414](https://tickets.metabrainz.org/browse/LB-414)

the importer did not notify listenbrainz about the user's currently playing tracks.

# Solution

Make requests to the currently-playing endpoint and submit that data as well.

# Action

This increases the scope of the permissions we need. So people will have to unlink
and link their accounts. I'm not sure if there is a way around this.

### DIFF
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -95,7 +95,7 @@ def get_spotify_oauth():
     """
     client_id = current_app.config['SPOTIFY_CLIENT_ID']
     client_secret = current_app.config['SPOTIFY_CLIENT_SECRET']
-    scope = 'user-read-recently-played'
+    scope = 'user-read-recently-played user-read-currently-playing'
     redirect_url = current_app.config['SPOTIFY_CALLBACK_URL']
     return spotipy.oauth2.SpotifyOAuth(client_id, client_secret, redirect_uri=redirect_url, scope=scope)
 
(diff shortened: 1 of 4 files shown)


listenbrainz/domain/spotify.py
```diff
@@ -95,7 +95,7 @@ def get_spotify_oauth():
     """
     client_id = current_app.config['SPOTIFY_CLIENT_ID']
     client_secret = current_app.config['SPOTIFY_CLIENT_SECRET']
-    scope = 'user-read-recently-played'
+    scope = 'user-read-recently-played user-read-currently-playing'
```
paramsingh: Should this be in a config file?

mayhem: I'm ok with it being in the code. Because you need to change more code if you change these values, no?

listenbrainz/domain/tests/test_spotify.py
```diff
@@ -54,7 +54,7 @@ def test_get_spotify_oauth(self):
         self.assertEqual(func_oauth.client_id, current_app.config['SPOTIFY_CLIENT_ID'])
         self.assertEqual(func_oauth.client_secret, current_app.config['SPOTIFY_CLIENT_SECRET'])
         self.assertEqual(func_oauth.redirect_uri, 'http://localhost/profile/connect-spotify/callback')
-        self.assertEqual(func_oauth.scope, 'user-read-recently-played')
+        self.assertEqual(func_oauth.scope, 'user-read-currently-playing user-read-recently-played')
```
mayhem: If they are used in multiple places a constant near the top might be good.

PR #481, opened 2019-01-19
Get duration from `duration_ms` for playing_now listens

paramsingh:
# Summary

* This is a Feature Addition

* **Describe this change in 1-2 sentences**: The Spotify API sends us the durations in ms, so if we get a listen with the key `duration_ms` we should use it to calculate the expiration time of currently playing tracks.

# Problem

We were ignoring the `duration_ms` key.

# Solution

Use the `duration_ms` key and add tests for expiration.

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -56,12 +56,12 @@ def _send_listens_to_queue(listen_type, listens):
         if listen_type == LISTEN_TYPE_PLAYING_NOW:
             try:
                 if 'duration' in listen['track_metadata']['additional_info']:
-                    expire_time = listen['track_metadata']['additional_info']['duration']
+                    listen_timeout = listen['track_metadata']['additional_info']['duration']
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/api_tools.py
```diff
@@ -55,8 +55,12 @@ def _send_listens_to_queue(listen_type, listens):
     for listen in listens:
         if listen_type == LISTEN_TYPE_PLAYING_NOW:
             try:
-                expire_time = listen["track_metadata"]["additional_info"].get("duration",
-                                    current_app.config['PLAYING_NOW_MAX_DURATION'])
+                if 'duration' in listen['track_metadata']['additional_info']:
+                    expire_time = listen['track_metadata']['additional_info']['duration']
```
mayhem: This whole block bugs me. Expire_time to me suggest that an absolute time of when something happens and duration is an interval of time. While the code looks correct, it feels awkward to read.

PR #485, opened 2019-01-19
LB-406: Stops import if listen count is zero 

vansika:
* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Users who accidentally try to import their listens with zero listen count found that the imports stuck in an infinite loop until the page is refreshed. 

* JIRA ticket (_optional_):  [LB-406](https://tickets.metabrainz.org/browse/LB-406) 

# Solution
Imports with listen count zero do not have 'listened_at" attribute. Imports were halted at a point because of key error. A simple check for length would do.


vansika:
😊

PR #487, opened 2019-01-21
LB-416: Change data type of auth_token from string to UUID

vansika:
* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

Datatype of auth_token is string when it actually stores UUID. 

* JIRA ticket (_optional_): [LB-416](https://tickets.metabrainz.org/browse/LB-416) 

# Solution

Changing datatype from string/varchar to UUID. Since request headers return auth_token as string, it should be explicitely converted to UUID.

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -4,7 +4,7 @@ CREATE TABLE "user" (
   id             SERIAL,
   created        TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
   musicbrainz_id VARCHAR NOT NULL,
-  auth_token     VARCHAR,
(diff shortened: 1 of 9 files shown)


alastair:
Hi!
Thanks for the patch.
I'm not sure we need this change - while it's true that the value of this key is a UUID, this is an implementation detail of how we generate them. There's no slowdown using a string type in the database here rather than a uuid type, and it means that if we ever change this key in the future, we don't need to make another database schema change.
I'm in favour of keeping the code as-is at the moment. I mentioned it to @paramsingh and he agreed with me in keeping it as a string for now.

vansika:
Okay. Sounds good to me:)

PR #490, opened 2019-01-22
Add a RabbitMQ consumer to send data to sockets

paramsingh:
Used a SelectConnection instead of creating two threads, seemed cleaner and safer.


consul_config.py.ctmpl
```diff
@@ -68,6 +68,7 @@ INCOMING_EXCHANGE = '''{{template "KEY" "incoming_exchange"}}'''
 INCOMING_QUEUE = '''{{template "KEY" "incoming_queue"}}'''
 UNIQUE_EXCHANGE = '''{{template "KEY" "unique_exchange"}}'''
 UNIQUE_QUEUE = '''{{template "KEY" "unique_queue"}}'''
+UNIQUE_PLAYER_QUEUE = '''{{template "KEY" "unique_player_queue"}}'''
```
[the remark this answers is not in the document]
mayhem: ahh, that is how that works. got it.

listenbrainz/config.py.sample
```diff
@@ -45,6 +45,7 @@ INCOMING_EXCHANGE = "incoming"
 INCOMING_QUEUE = "incoming"
 UNIQUE_EXCHANGE = "unique"
 UNIQUE_QUEUE = "unique"
+UNIQUE_PLAYER_QUEUE = "unique_player"
```
mayhem: Let call this FOLLOW_LIST_QUEUE

listenbrainz/spotify_player_writer/spotify_player_writer.py
```diff
@@ -0,0 +1,77 @@
...
+import pika
+import time
+
+from flask import current_app
+
+
+class PlayerWriter:
+
+    def __init__(self):
+        self.app = listenbrainz.webserver.create_app()
```
mayhem: This will need changing, since we're building a different app module.

PR #495, opened 2019-01-22
Fix Werkzeug TypeError: '<' not supported between instances of 'str' and 'int

vansika:
* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
TypeError: ('<' not supported between instances of 'str' and 'int ) on every request.

# Solution

Upgrade Werkzeug version


vansika:
Yeah :)

PR #496, opened 2019-01-23
Submit data coming from RMQ to the socketio server

paramsingh:
and then push it to the room it needs to go to.

### DIFF
--- a/listenbrainz/follow_server/dispatcher.py
+++ b/listenbrainz/follow_server/dispatcher.py
@@ -3,10 +3,8 @@ import listenbrainz.webserver
 import pika
 import time
 import threading
-import socketio
 
 from flask import current_app
-from flask_socketio import emit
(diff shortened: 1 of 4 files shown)


listenbrainz/follow_server/dispatcher.py
```diff
@@ -3,24 +3,41 @@
...
+from listenbrainz.webserver.views.api_tools import LISTEN_TYPE_PLAYING_NOW, LISTEN_TYPE_IMPORT
 
 class FollowDispatcher(threading.Thread):
 
     def __init__(self, app):
         threading.Thread.__init__(self)
         self.app = app
 
+    def send_listens(self, listens, listen_type):
+        sio = socketio.Client()
```
paramsingh: Not sure how to best do this without creating a new connection to the server everytime. Do this in another thread?

PR #504, opened 2019-01-30
Npm install

paramsingh:
# Summary

* This is a…
    * ( ) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * (x) Other
* **Describe this change in 1-2 sentences**: Add a package.json and install the less dependencies via that.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,8 +23,8 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Node
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
-RUN apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
+    apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir /code
 WORKDIR /code
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,10 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends postgresql-client-$PG_MAJOR \
     && rm -rf /var/lib/apt/lists/*
 
+# Node
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
+
 RUN mkdir /code
 WORKDIR /code
 
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -44,6 +44,13 @@ COPY ./requirements.txt .
 RUN pip3 install -r requirements.txt
 RUN pip3 install setuptools==36.0.1
 
+# Node
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
+    apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
+COPY ./package.json /code/listenbrainz/
+COPY ./package-lock.json /code/listenbrainz/
+RUN npm install
+
 COPY . /code/listenbrainz
 WORKDIR /code/listenbrainz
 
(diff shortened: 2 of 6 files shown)


Dockerfile.prod
```diff
@@ -65,10 +65,11 @@ RUN touch /var/log/stats.log /var/log/dump_create.log && chown listenbrainz:list
...
 
 # Node
-RUN curl -sL https://deb.nodesource.com/setup_6.x | bash -
-RUN apt-get install -y nodejs
-RUN npm install -g less@2.7.2
-RUN npm install less-plugin-clean-css
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
+COPY ./package.json /code/listenbrainz/
+COPY ./npm-shrinkwrap.json /code/listenbrainz/
```
MonkeyDo: I believe Node 10.x (npm 6.4.1) should generate a `package-lock.json` instead of `npm-shrinkwrap.json`, even though shrinkwrap is taken into account as a backup.
Not sure this is super important. https://docs.npmjs.com/files/package-lock.json

Dockerfile
```diff
@@ -22,6 +22,10 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends postgresql-client-$PG_MAJOR \
     && rm -rf /var/lib/apt/lists/*
 
+# Node
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
```
alastair: This should be in a single RUN line, as the curl|bash will run `apt-get update`, and so the rm has no effect on the previous layer

docs/dev/devel-env.rst
```diff
@@ -186,6 +186,16 @@ Your development environment is now ready. Now, let's actually see ListenBrainz
...
 
 
+Install node dependencies
+-------------------------
+
+You also need to install some JavaScript dependencies.
+
+.. code-block:: bash
+
+    docker-compose -f docker/docker-compose.yml -p listenbrainz run --rm web npm install
```
MonkeyDo: If there are build steps after installation (`postinstall` script, like running webpack and such), we'll need to run this command as the LB user.
I assumer we're not risking anything?
`docker-compose -f docker/docker-compose.yml -p listenbrainz run --rm -u listenbrainz web npm install`

Dockerfile.prod
```diff
@@ -65,10 +65,11 @@ RUN touch /var/log/stats.log /var/log/dump_create.log && chown listenbrainz:list
 RUN touch /etc/service/cron/down
 
 # Node
-RUN curl -sL https://deb.nodesource.com/setup_6.x | bash -
-RUN apt-get install -y nodejs
-RUN npm install -g less@2.7.2
-RUN npm install less-plugin-clean-css
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
+RUN apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
+COPY ./package.json /code/listenbrainz/
```
MonkeyDo: If I change my package.json, is that change not going to be picked up by the `COPY`command at line 47 ?
Does line 47 need to be modified to not copy the JS-related files?
https://github.com/metabrainz/listenbrainz-server/pull/504/files#diff-0e9722a692af2b244674b56146a1ecd3R47

docs/dev/devel-env.rst
MonkeyDo: Out of curiosity, what is the reason for running this command separately and not at some point in the Dockerfile?

Dockerfile.prod
alastair: you're right, normally specific copies should be done earlier in the file (package.json, requirements.txt) and bulk copies (COPY .) later in the file. This way, changes to other files in the repository won't cause all the future commands to be re-run.
Having said this, I guess this is the production dockerfile, so perhaps it's not as much as an issue because this isn't run as often, but I agree that this file should be re-organised a bit.

Dockerfile.prod
```diff
@@ -44,6 +44,13 @@ COPY ./requirements.txt .
 RUN pip3 install -r requirements.txt
 RUN pip3 install setuptools==36.0.1
 
+# Node
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
```
alastair: I would consider moving the apt-get install above the python/pip installation, because as it is now, you're going to re-install node from apt every time you change something in requirements (that is, consider putting things in order from "least likely to change" to "most likely to change")

PR #509, opened 2019-02-05
Add Webpack to compile JS

MonkeyDo:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem
Unable to import JS libraries or modularise files since they would need a module loader or a compiler
Flaky loading of js libraries was causing race-condition errors
No support for older browsers

# Solution
Use Webpack with babel to transpile and bundle the JS code

# Action
For development environment, I am currently running `./develop.sh` in one terminal window, and `npm run build:dev` in another, which is a shortcut for running webpack in dev and watch mode.
Is that the expected workflow? I'll need to write something up in the dev instructions.

For prod environment, the prod dockerfile runs npm install, which in turn will run webpack in prod mode.


MonkeyDo:
I'm working on cleaning up the conflicts. I shit myself a bit on this PR.

MonkeyDo:
That's a clean stopping point for this PR I think, if y'all want to test now.

PR #510, opened 2019-02-05
Follow feature fixes

mayhem:
# Summary

* This is a…
    * (x) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

Add/remove users from rooms correctly. Remove some debug prints. Add a rudimentary test for the follow feature. Fix problem with going to follow page without being logged in.

# Problem

* JIRA ticket (_optional_): 

# Solution

# Action

### DIFF
--- a/listenbrainz/webserver/views/test/test_follow.py
+++ b/listenbrainz/webserver/views/test/test_follow.py
@@ -1,0 +1,45 @@
+import listenbrainz.db.stats as db_stats
+import listenbrainz.db.user as db_user
+import ujson
+from unittest import mock
+
+from flask import url_for, current_app
+from influxdb import InfluxDBClient
+from listenbrainz.db.testing import DatabaseTestCase
+from listenbrainz.listenstore.tests.util import create_test_data_for_influxlistenstore
+from listenbrainz.webserver.influx_connection import init_influx_connection
+from listenbrainz.webserver.login import User
+from listenbrainz.webserver.testing import ServerTestCase
+
+import listenbrainz.db.user as db_user
+import logging
+
+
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/test/test_follow.py
```diff
@@ -0,0 +1,45 @@
...
+
+class FollowViewsTestCase(ServerTestCase, DatabaseTestCase):
+    def setUp(self):
+        ServerTestCase.setUp(self)
+        DatabaseTestCase.setUp(self)
+
+        self.log = logging.getLogger(__name__)
+
+        user = db_user.get_or_create(1, 'iliekcomputers')
+        self.user = User.from_dbrow(user)
```
paramsingh: this isn't needed either

listenbrainz/webserver/views/test/test_follow.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+    def tearDown(self):
+        self.logstore = None
```
paramsingh: doesn't seem needed.

listenbrainz/webserver/views/test/test_follow.py
```diff
@@ -0,0 +1,45 @@
+import listenbrainz.db.stats as db_stats
```
paramsingh: we should probably only use the imports we need.

listenbrainz/webserver/views/test/test_follow.py
mayhem: self.user does actually get used below.

paramsingh: oh ok. right. :+1:

PR #511, opened 2019-02-06
Allow the option to choose which spotify features to opt in to

paramsingh:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

* **Describe this change in 1-2 sentences**: The user now has the option to opt in to the import without the music playing capabilities and vice versa

### DIFF
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -25,7 +25,8 @@ SPOTIFY_LISTEN_PERMISSIONS = (
 
 class Spotify:
     def __init__(self, user_id, musicbrainz_id, musicbrainz_row_id, user_token, token_expires,
-                 refresh_token, last_updated, record_listens, error_message, latest_listened_at):
+                 refresh_token, last_updated, record_listens, error_message, latest_listened_at,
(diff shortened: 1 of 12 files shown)


listenbrainz/domain/spotify.py
```diff
@@ -173,6 +190,42 @@ def update_latest_listened_at(user_id, timestamp):
     db_spotify.update_latest_listened_at(user_id, timestamp)
 
 
+def get_access_token(code):
```
paramsingh: the spotipy lib has a bug which makes it lose the scope returned by the Spotify API. the scope is needed if we want to know whether the user wants their listens imported or not. So I've used the Spotify endpoint directly here. I'll open a ticket in spotipy but the last release was in 2017, so not sure how active it is.

mayhem: Don't forget to make a ticket to remind you to look into this.

paramsingh: [LB-421](https://tickets.metabrainz.org/browse/LB-421)

PR #514, opened 2019-02-08
Consumer for Listenbrainz-server 

vansika:
* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

The following scripts connect to RabbitMQ at LB-playground side, consume the data and push it into the database.


vansika:
Sure :)

listenbrainz/db/stats.py
```diff
@@ -43,16 +43,16 @@ def insert_user_stats(user_id, artists, recordings, releases, artist_count):
 
     artist_stats = {
         'count': artist_count,
-        'all_time': artists,
+        'prev_month': artists,
     }
 
     recording_stats = {
-        'all_time': recordings,
+        'prev_month': recordings,
```
paramsingh: We should prolly just specify the month here in the key, like `2019-01` so that we know exactly which month the data is for. With prev month, we might have data for different months for different users.

listenbrainz/spark/spark_reader.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        data = ujson.loads(body)
+        user_name = next(iter(data[0]))
+        user = db_user.get_by_mb_id(user_name)
+        artists = data[0][user_name]['artists']['artist_stats']
+        recordings = data[0][user_name]['recordings']
+        releases = data[0][user_name]['releases']
+        artist_count = data[0][user_name]['artists']['artist_count']
+        db_stats.insert_user_stats(user['id'], artists, recordings, releases, artist_count)
+        print ("data for {} published".format(user_name))
```
paramsingh: This print seems extraneous, or it should be current_app.logger.info

listenbrainz/spark/spark_reader.py
```diff
@@ -0,0 +1,85 @@
...
+            password=current_app.config['RABBITMQ_PASSWORD'],
+            host=current_app.config['RABBITMQ_HOST'],
+            port=current_app.config['RABBITMQ_PORT'],
+            virtual_host=current_app.config['RABBITMQ_VHOST'],
+            error_logger=current_app.logger.error,
+        )
+
+
+    def callback(self, ch, method, properties, body):
+        """ Handle the data received from the queue and works accordingly.
```
paramsingh: This docstring could be better. Handle the stats received from the queue and insert into the database accordingly.

listenbrainz/spark/spark_reader.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        data = ujson.loads(body)
+        user_name = next(iter(data[0]))
```
paramsingh: What exactly does this mean? I wonder if we can make it a bit clearer?

vansika: This is used to retrieve the first key of a dictionary, which would always be the user name in this case.

listenbrainz/spark/spark_reader.py
```diff
@@ -0,0 +1,85 @@
...
+                self.init_rabbitmq_connection()
+                self.incoming_ch = utils.create_channel_to_consume(
+                    connection=self.connection,
+                    exchange=current_app.config['SPARK_EXCHANGE'],
+                    queue=current_app.config['SPARK_QUEUE'],
+                    callback_function=self.callback,
+                )
+                current_app.logger.info('Stats calculator started!')
+                try:
+                    print("consuming")
```
paramsingh: this print isn't needed as well.

listenbrainz/spark/spark_reader.py
paramsingh: I think this should just be data.keys()[0]?

listenbrainz/spark/spark_reader.py
```diff
@@ -0,0 +1,85 @@
...
+            port=current_app.config['RABBITMQ_PORT'],
+            virtual_host=current_app.config['RABBITMQ_VHOST'],
+            error_logger=current_app.logger.error,
+        )
+
+
+    def callback(self, ch, method, properties, body):
+        """ Handle the data received from the queue and works accordingly.
+        """
+        data = ujson.loads(body)
```
paramsingh: why is data a list? I think we could get away with just a dict for now?

listenbrainz/spark/spark_reader.py
alastair: In anything other than python 3.7, dictionaries are not guaranteed to be in any specific order. Especially when reading from json I would be very careful about this. What does the format of the body look like? Is it always {username: {somedata}, otherusername: {moredata}} ? How many user datas are returned? always 1 or more?
You could use `for username, stats in data.items()`, which would work even if there was only data for 1 username. It would also clean up the accesses later in this method

listenbrainz/spark/spark_reader.py
```diff
@@ -34,17 +35,20 @@ def init_rabbitmq_connection(self):
...
+            insert into the database accordingly.
         """
         data = ujson.loads(body)
-        user_name = next(iter(data[0]))
-        user = db_user.get_by_mb_id(user_name)
-        artists = data[0][user_name]['artists']['artist_stats']
-        recordings = data[0][user_name]['recordings']
-        releases = data[0][user_name]['releases']
-        artist_count = data[0][user_name]['artists']['artist_count']
+        for key, value in data.items():
```
alastair: this whole block, including the insert_user_stats and logger call should all be indented in this for loop. Even though we only expect there to be one item here, it doesn't matter that we do it in all in the loop.
Use more specific names in the forloop, e.g. `username, metadata`
Confirm with @paramsingh if we should do early return if the user doesn't exist, of if we want to always do the basic_ack on the channel below.

listenbrainz/db/stats.py
```diff
@@ -41,18 +42,23 @@ def insert_user_stats(user_id, artists, recordings, releases, artist_count):
              artist_count (int): the total number of artists listened to by the user
     """
 
+    month = datetime.now().month
```
alastair: You have this code in three places, it should be moved into a utility so that it's only called once.
A warning: Dates are hard. really hard. I'd like to see some changes made here to be a little bit more careful about our calculations:
Call `datetime.now()` only once and use the same value to get both the year and the month
`datetime.now()` returns the _local_ time of the server. This could be any value depending on what the timezone of the server is set to. We should  always use UTC here - `datetime.utcnow()`
For a method of getting previous/next month, I'd prefer to use libraries that help manage these kinds of calculations. for example:

    >>> t = datetime.datetime.utcnow().replace(day=1)
    >>> t + dateutil.relativedelta.relativedelta(months=-1)
    datetime.datetime(2019, 1, 1, 9, 21, 27, 478427)
    >>> t + dateutil.relativedelta.relativedelta(months=-2)
   datetime.datetime(2018, 12, 1, 9, 21, 27, 478427)

Note that going back 2 months from feb will take us to 2018 automatically. The `relativedelta` method has some smart functionality, for example going back 1 month from March 31 will go to Feb 28 or 29 correctly, but in order to make the expected behaviour more explicit I use `.replace(day=1)` to set the date of the month to the 1st.

Dates are typically stored with a - as a separator, not a . like in your previous example. You should also 0-pad 1 digit months: 2019-02 (https://en.wikipedia.org/wiki/ISO_8601#Calendar_dates)

I'm not exactly sure why you want this date information, but it would probably be a good idea to pass this information into the `insert_user_stats` method as parameters. This would make it easier to test, and clearer to know where the values were calculated from. Do some form of these dates come from bigquery?

vansika: No, these dates are not coming from bigquery. We are storing date information to know the data/stats belong to which month. We might want to implement a functionality in future that will allow users to retrieve data of any month/week and not just the previous month, in that case these dates would be useful.

listenbrainz/spark/spark_reader.py
```diff
@@ -0,0 +1,100 @@
...
+
+from flask import current_app
+from listenbrainz import utils
+from listenbrainz.db import user as db_user, stats as db_stats
+from listenbrainz.webserver import create_app
+from listenbrainz.db.exceptions import DatabaseException
+from listenbrainz import config
+import sqlalchemy
+
+TEST_DATA_PATH = os.path.join('/code', 'listenbrainz', 'listenbrainz', 'testdata/')
```
alastair: what's `testdata` in this case? This doesn't sound like something that should be in a production script. We shouldn't store runtime data in the source directory. This should be in a separate volume, and the configuration for the path location should be in the config file.

listenbrainz/db/stats.py
```diff
@@ -25,10 +25,9 @@
 import ujson
 
 from listenbrainz import db
-from listenbrainz import config
+from listenbrainz import config 
 
-
-def insert_user_stats(user_id, artists, recordings, releases, artist_count):
+def insert_user_stats(user_id, artists, recordings, releases, artist_count, timestamp):
     """Inserts user stats calculated from Google BigQuery into the database.
```
alastair: is BigQuery true here now? This docstring should probably be updated to say Spark, or to be technology independent.

listenbrainz/db/stats.py
```diff
@@ -25,10 +25,9 @@
 import ujson
 
 from listenbrainz import db
-from listenbrainz import config
+from listenbrainz import config 
 
-
-def insert_user_stats(user_id, artists, recordings, releases, artist_count):
+def insert_user_stats(user_id, artists, recordings, releases, artist_count, timestamp):
```
alastair: `timestamp` isn't a good name here - traditionally I would expect this to refer to a unix timestamp (number of seconds since Jan 1 1970), but it looks like you're using it to mean `yyyy.mm` instead. Maybe call it `yearmonth` instead?

listenbrainz/db/stats.py
alastair: thanks for the clarification, I guess we should be talking about spark and not bigquery here, but I guess the result is the same.
It's unclear to me how this script is going to be run, so maybe I'm missing some understanding on exactly what time-frame this data covers, and what happens when you run the consumer:
 * How often is this script run?
 * At the point that this script is run, what data is in spark? Is it summary data for all users for a given month? Do we know what data has been calculated? For what users, and for what timeframe this data covers? I understand that this is summary data for an entire month, therefore I find it surprising that the month that the data covers isn't part of the information.
 * What happens if we run this script twice in a row? Will this override data for the last month or append the data to the table? Does this mean that it might be possible to have two rows containing summary data for a user for the same month?
 * What happens if we run it at 23:59 on the last day of the month and we continue running over the border of the month? This means that the "last month" for two subsequently inserted rows would be different, even though I suspect that they are actually for the same month.

I think that its probably necessary to have:
 * A sentinel table that keeps track of what month/user combinations have been inserted (this could be part of the `statistics.user` table, either by adding a date field or by creating an index into the date field of the inner json file.
* and/or: report the month that these statistics are for directly from spark so that there is no confusion as to what month they cover. Use this data to insert the stats into the database.

listenbrainz/db/stats.py
vansika: Yes. I think it should be replaced with 'Spark' because that is the technology we are using.

listenbrainz/db/stats.py
vansika: yearmonth sounds good. It basically tells us the month and year in which the stats were calculated.
Maybe YearMonth?

listenbrainz/spark/spark_reader.py
vansika: The YearMonth would be imported from Listenbrainz-playground and the database would look like "2019-01" : {artist_stats} and so on. This value ("2019-01") should be available to test scripts and hence I stored a single JSON file in testdata from which the value can be retrieved. I will make a separate volume though :)

listenbrainz/db/stats.py
vansika: - At present we decided to run the script monthly. Every month, previous month data of users would be exported from lb-server to playground where we will calculate stats of users and push them into rabbitmq. Then at lb-server end data will be consumed and fed into the database. Therefore, the script will run as many times as the playground producer runs which is monthly as of now.
- When this script runs, we have all the listen data of users for the previous month in spark. Though we will have data of months before the previous months (spark will basically store all the listen data) but we will be processing only the data for last month. Yes, the script at spark end tells us what data has been calculated. Currently, it is top 100 of artists, recordings, and releases.  I have modified the spark script, it will contain a key timestamp (which we changed to YearMonth) that will tell us the year and month in which the data was calculated. For eg, 'YearMonth' : '2019-01'. Also, the data in stored in json files like 1.json, 2.json which tells us the month and the main directory would be like 2019 which tells us the year. 
- If the script runs in a row, the data will be overridden. At any point we will only store data of the previous month of a user in the DB as of now. We won't have two rows for the same month. We can talk over if we need to keep record of every month's data in DB. But, I think this would un-necessarily flood the DB. If at any point the user wants to query data of some sepcific period we can queue the request and send it to spark. 
- There won't be two subsequent rows for a user as of now but this point is something which I was thinking about. Presently, if we calculate data at 23:59 and this is March and we keep calculating till the morning of 1st april the YearMonth would be "2019-03", a month before April but actually, it should be february. Presently, We assumed that the calculation would be done by the mid of a month but this is something that should be discussed upon. Thanks for pointing out :)

vansika: The spark script reports the month for which these stats have been calculated. This info will be useful if we decide not to override the data, and store json data in db of months other than the previous month.

listenbrainz/db/stats.py
alastair: in python we never name variables with upper case letters.

listenbrainz/spark/spark_reader.py
alastair: I don't understand what you mean by test scripts. Do you mean unit tests? or do you mean experiments that you want to run on the data? Why are you putting it in a file and not in the database?

listenbrainz/db/stats.py
alastair: > The spark script reports the month for which these stats have been calculated.

I think this is a key point. If you have the data for the month that you can read from spark, then use that! This will solve lots of issues about getting the correct value for the previous month when adding the data to the database.

listenbrainz/db/stats.py
vansika: I thought yearmonth lacks distinction. But yeah sure :) Pythonic way is better.

listenbrainz/spark/spark_reader.py
vansika: I meant unit tests (./test.sh). I should have been clear with the word. 

Since it is for unit test, I did not saved it in the db.

listenbrainz/db/stats.py
vansika: Yes sure :)

listenbrainz/spark/spark_reader.py
alastair: So it sounds like this code shouldn't be here then. If it's just to generate data for using in a test then it shouldn't be run during a production script. Consider making a separate script to just retrieve and download this data if you need.

vansika:
Thanks:)

vansika:
@alastair Thank you.

PR #516, opened 2019-02-13
Change login id column type to string

paramsingh:
Changed the type to string as discussed on IRC with default values as uuids. The migration script puts the default values for existing users as their user IDs which should make sure that they do not get logged out.

Also create index and add a constraint and rename to column to login_id from user_login_id.

### DIFF
--- a/admin/sql/updates/2019-02-13-change-login-id-type.sql
+++ b/admin/sql/updates/2019-02-13-change-login-id-type.sql
@@ -1,0 +1,13 @@
+BEGIN;
+
+ALTER TABLE "user" DROP COLUMN user_login_id;
+ALTER TABLE "user" ADD COLUMN login_id TEXT NOT NULL DEFAULT uuid_generate_v4()::text;
+
+UPDATE "user"
+   SET login_id = id::text;
+
+ALTER TABLE "user" user_login_id_key UNIQUE (login_id);
+CREATE UNIQUE INDEX login_id_ndx_user ON "user" (login_id);
+
+COMMIT;
+
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -85,7 +85,8 @@ def get(id):
         row = result.fetchone()
         return dict(row) if row else None
 
-def get_by_user_login_id(login_id):
+
+def get_by_login_id(login_id):
     """Get user with a specified login ID.
 
     Args:
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -58,7 +58,7 @@ def update_token(id):
             raise
 
 
-USER_GET_COLUMNS = ['id', 'created', 'musicbrainz_id', 'auth_token', 'last_login', 'latest_import', 'gdpr_agreed', 'musicbrainz_row_id', 'user_login_id']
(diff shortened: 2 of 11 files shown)


admin/sql/updates/2019-02-13-change-login-id-type.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+ALTER TABLE "user" DROP COLUMN user_login_id;
+ALTER TABLE "user" ADD COLUMN login_id TEXT NOT NULL DEFAULT uuid_generate_v4()::text;
+
+UPDATE "user"
+   SET login_id = id::text;
```
alastair: nice 🕶

admin/sql/updates/2019-02-13-change-login-id-type.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+ALTER TABLE "user" DROP COLUMN user_login_id;
```
alastair: is the idea that we'll run both updates in a row while releasing in prod?

listenbrainz/db/user.py
```diff
@@ -104,7 +104,7 @@ def get_by_user_login_id(login_id):
         result = connection.execute(sqlalchemy.text("""
             SELECT {columns}
               FROM "user"
-             WHERE user_login_id = :user_login_id
+             WHERE login_id = :user_login_id
```
alastair: with this change do we also want to change the name of this method? (get_by_user_login_id -> get_by_login_id)

admin/sql/updates/2019-02-13-change-login-id-type.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE "user" ADD COLUMN login_id TEXT NOT NULL DEFAULT uuid_generate_v4()::text;
+
+UPDATE "user"
+   SET login_id = id::text;
+
+ALTER TABLE "user" user_login_id_key UNIQUE (login_id);
+CREATE UNIQUE INDEX login_id_ndx_user ON "user" (login_id);
```
alastair: I believe that as an implementation detail, in postgres unique constraints are implemented by creating a unique index. With a constraint _and_ index on this field, does it mean that there are two indexes? Do we mind?

admin/sql/updates/2019-02-13-change-login-id-type.sql
paramsingh: the first one is already run, i'll run this while releasing on beta for testing.

listenbrainz/db/user.py
paramsingh: right, good point.

admin/sql/updates/2019-02-13-change-login-id-type.sql
paramsingh: this is how it is done for other columns, i'd keep it as is and change them all if needed.

listenbrainz/db/user.py
alastair: Perfect

admin/sql/updates/2019-02-13-change-login-id-type.sql
alastair: Fine

PR #518, opened 2019-02-15
Don't create websocket connection for 'recent' page

MonkeyDo:
# Summary

* This is a…
    * (x) Refactoring

# Problem

Useless creation of websocket connection in a mode that does not use it.

### DIFF
--- a/listenbrainz/webserver/static/js/jsx/profile.jsx
+++ b/listenbrainz/webserver/static/js/jsx/profile.jsx
@@ -36,7 +36,9 @@ class RecentListens extends React.Component {
 	}
 
 	componentDidMount(){
-		this.connectWebsockets();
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/static/js/jsx/profile.jsx
```diff
@@ -36,7 +36,9 @@ class RecentListens extends React.Component {
 	}
 
 	componentDidMount(){
-		this.connectWebsockets();
+    if(this.state.mode === "listens" || this.state.mode === "follow"){
```
mayhem: indent problem? tabs vs spaces?

MonkeyDo: Yeah, my text editor keeps taking rash decisions without consulting me.
Fixed with new commit.

PR #519, opened 2019-02-17
remove bigquery job runner files to simplify tests

vansika:
* This is a…
    * ( ) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * (x ) Other

Since Big query is no more used to calculate stats, big query runner files are not required. It would reduce the overhead to make changes in job runner files while making changes in the codebase.


paramsingh:
We can also probably get rid of stuff in the stats directory. Maybe open a ticket for that? I'll merge this for now.

PR #522, opened 2019-02-21
Create separate follow page template

MonkeyDo:
# Summary

* This is a…
    * (x) Refactoring

Sure enough, it came quick. I got annoyed at seeing the user name and stats block on the follow page, so here's a fix.

Considering the similarities between user/profile and index/follow templates, it may be worth refactoring further to have for example the  scripts block in common.
This could turn out to be a hindrance later, so I abstained for now (*insert premature optimization comic strip reference*)

### DIFF
--- a/listenbrainz/webserver/templates/index/follow.html
+++ b/listenbrainz/webserver/templates/index/follow.html
@@ -1,17 +1,8 @@
-{% extends 'base.html' %}
+{%- extends 'base-react.html' -%}
 
 {% block title %}Follow users {{ follow_list|join(', ') }} - ListenBrainz{% endblock %}
 
-{% block content %}
-
-    <div id="react-listens">
-    </div>
-
-{% endblock %}
-
 {% block scripts %}
   {{ super() }}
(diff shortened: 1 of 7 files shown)


listenbrainz/webserver/templates/index/follow.html
```diff
@@ -0,0 +1,17 @@
...
+{% block content %}
+
+    <div id="react-listens">
+    </div>
+
+{% endblock %}
+
+{% block scripts %}
+  {{ super() }}
+  <script defer src="https://use.fontawesome.com/releases/v5.0.8/js/all.js" integrity="sha384-SlE991lGASHoBfWbelyBPLsUlwY1GwNDJo3jSJO04KZ33K2bwfV9YBauFfnzvynJ" crossorigin="anonymous"></script>
```
paramsingh: is this code reused in another template? I'm not sure if it is, but we should try to not duplicate it if it is being duplicated.

MonkeyDo: Yes, here:
https://github.com/metabrainz/listenbrainz-server/blob/5c3d6fdab340180111e8a16ec5d62c5484a0f3d9/listenbrainz/webserver/templates/user/profile.html#L19

What's the best way to refactor?
Create and extend a react template?

MonkeyDo: OK, I refactored with a react base template

PR #527, opened 2019-02-26
Make the follow page require logged in users

paramsingh:
We already didn't have functionality to allow for playing songs for non-logged in users. Should just make the page only available to logged-in users.


mayhem:
What was missing to allow non logged in users to play? A spotify account-link?

paramsingh:
Yes, only users who have authrd with Spotify would be able to play

mayhem:
We hope to change this soon with a more intelligent player, but when we do we can undo this change as well.

PR #528, opened 2019-02-27
Fix playlist mechanism for follow page

MonkeyDo:
# Summary

* This is a…
    * (x) Bug fix
    
# Problem

The playlist functionality was not working on the follow page. In short, the list of listens was never updated properly after initially being set.
So on the follow page, the player only had an empty listens list (whereas on user/recent pages an array of listens is loaded initially)

# Solution

Dummy move, I set the state in the constructor, but we don't need to store listens in the state as we don't modify it. Using the component prop directly means we're back to working order.


MonkeyDo:
> This looks like one of those, "how did this ever work" sort of fixes. :)

It can't ever have worked on the follow page. Same component was working for recent/user pages.
It took to long to realize !

PR #530, opened 2019-02-28
Save follow lists in the database

paramsingh:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Action

I added a table and save the follow lists inside them. Then the follow page by default shows the latest list that was saved by the user. 

* Need to carefully look at the JSX code, there might be bad practices there.
* ~Need to look at how lists are being updated.~
* A general UX review by Mr_Monkey would be nice
* I have a hacky way of putting the list name inside the name input via placeholder, not exactly sure what the best way forward there is.
* Should `/follow/save` be an API endpoint? If so, how exactly would user authentication work?
* Need to get the base url from the server. I saw that this was a problem everywhere and rob was fixing it so i left it for now.
* ~Need to add docstrings.~

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -99,13 +99,14 @@ CREATE TABLE follow_list (
   creator           INTEGER NOT NULL, -- FK to "user".id
   private           BOOLEAN NOT NULL DEFAULT FALSE,
   created           TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
-  last_listened     TIMESTAMP WITH TIME ZONE
+  last_saved        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
 
 CREATE TABLE follow_list_member (
   list_id      INTEGER NOT NULL,
   user_id      INTEGER NOT NULL,
+  priority     INTEGER NOT NULL,
   added        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
 
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -88,7 +88,6 @@ CREATE TABLE statistics.recording (
 );
 ALTER TABLE statistics.recording ADD CONSTRAINT recording_stats_msid_uniq UNIQUE (msid);
 
-
 CREATE TABLE data_dump (
   id          SERIAL,
   created     TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
@@ -101,6 +100,7 @@ CREATE TABLE follow_list (
   private       BOOLEAN NOT NULL DEFAULT FALSE,
   created       TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
+ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_key UNIQUE (name);
 
 CREATE TABLE follow_list_member (
   list_id      INTEGER NOT NULL,
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -94,11 +94,12 @@ CREATE TABLE data_dump (
 );
 
 CREATE TABLE follow_list (
-  id            SERIAL, -- PK
-  name          TEXT NOT NULL,
-  creator       INTEGER NOT NULL, -- FK to "user".id
-  private       BOOLEAN NOT NULL DEFAULT FALSE,
-  created       TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
+  id                SERIAL, -- PK
+  name              TEXT NOT NULL,
+  creator           INTEGER NOT NULL, -- FK to "user".id
+  private           BOOLEAN NOT NULL DEFAULT FALSE,
+  created           TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+  last_listened     TIMESTAMP WITH TIME ZONE
 );
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
 
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -98,7 +98,7 @@ CREATE TABLE follow_list (
   name              TEXT NOT NULL,
   creator           INTEGER NOT NULL, -- FK to "user".id
   private           BOOLEAN NOT NULL DEFAULT FALSE,
-  member            INTEGER ARRAY DEFAULT ARRAY[]::INTEGER[], -- TODO: look into indexes
+  member            INTEGER ARRAY NOT NULL DEFAULT ARRAY[]::INTEGER[],
   created           TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
   last_saved        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -98,7 +98,7 @@ CREATE TABLE follow_list (
   name              TEXT NOT NULL,
   creator           INTEGER NOT NULL, -- FK to "user".id
   private           BOOLEAN NOT NULL DEFAULT FALSE,
-  member            INTEGER ARRAY NOT NULL DEFAULT ARRAY[]::INTEGER[],
+  members           INTEGER ARRAY NOT NULL DEFAULT ARRAY[]::INTEGER[],
   created           TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
   last_saved        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -98,16 +98,10 @@ CREATE TABLE follow_list (
   name              TEXT NOT NULL,
   creator           INTEGER NOT NULL, -- FK to "user".id
   private           BOOLEAN NOT NULL DEFAULT FALSE,
+  member            INTEGER ARRAY DEFAULT ARRAY[]::INTEGER[], -- TODO: look into indexes
   created           TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
   last_saved        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
-
-CREATE TABLE follow_list_member (
-  list_id      INTEGER NOT NULL,
-  user_id      INTEGER NOT NULL,
-  priority     INTEGER NOT NULL,
-  added        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
-);
 
 COMMIT;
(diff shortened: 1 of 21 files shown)


listenbrainz/db/follow_list.py
```diff
@@ -0,0 +1,144 @@
...
+    })
+    return [dict(row) for row in result.fetchall()]
+
+
+def _get_by_creator_and_name(connection, creator, list_name):
+    r = connection.execute(sqlalchemy.text("""
+        SELECT id
+          FROM follow_list
+         WHERE creator = :creator
+           AND LOWER(name) = LOWER(:list_name)
```
mayhem: Is this the way to do this in postgres? Seems to me that that would result in a table scan for each query....

mayhem: I wonder if there is an index trick for this one...

listenbrainz/db/follow_list.py
```diff
@@ -0,0 +1,144 @@
...
+            return None
+        else:
+            row = dict(result.fetchone())
+            return row
+
+
+def update(list_id, name, members):
+    with db.engine.begin() as connection:
+        _update_name(connection, list_id, name)
+        _remove_users(connection, list_id)
```
mayhem: The remove all and re-add all pattern doesn't scale very well -- it is fine for small lists, but once the lists get large you're creating too many operations for a simple "add this user to my list" operation. Comparing the existing and new lists and then creating a DELETE and an INSERT statement ensures that this operation touches the least number of rows necessary.

paramsingh: Using a postgres array now.

listenbrainz/db/follow_list.py
paramsingh: There is an index on (creator, list_name) already. I'll look more into it.

listenbrainz/webserver/static/js/jsx/follow-users.jsx
```diff
@@ -6,7 +6,10 @@ export class FollowUsers extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      users: props.followList || []
+      users: props.followList || [],
+      saveUrl: props.saveUrl || "https://listenbrainz.org/follow/save",
```
MonkeyDo: For the default value, I would suggest getting the current origin, so that there should be nothing to change if you are running locally:

` `${window.location.origin}/follow/save` `

paramsingh: Right, makes sense.

listenbrainz/webserver/static/js/jsx/follow-users.jsx
```diff
@@ -42,6 +46,40 @@ export class FollowUsers extends React.Component {
...
+    })
+    .then(response => response.json())
+    .then(data => {
+      console.debug(data);
+      console.debug("old List ID: " + this.state.listId);
+      this.setState(prevState => {
+        return {listId: data.list_id};
+      });
+      console.debug("new List ID: " + this.state.listId);
+    })
```
MonkeyDo: You also want to `.catch` any error with the promise, and do something with the error. I'd say just write it to console.error for now, I want to come up with a good way to show errors for this component.

listenbrainz/webserver/views/follow.py
```diff
@@ -1,47 +1,100 @@
...
+
+        # do some validation
+        current_list = db_follow_list.get(list_id)
+        if current_list is None:
+            return jsonify({
+                'code': 404,
+                'message': 'List not found',
+            }), 404
+        if current_list['creator'] != current_user.id:
+            raise Unauthorized("Can only edit your own lists")
```
mayhem: "You can only edit your own lists."

listenbrainz/webserver/views/follow.py
```diff
@@ -1,47 +1,100 @@
...
+        # update the old list
+        db_follow_list.update(
+            list_id=list_id,
+            name=list_name,
+            member=[user['id'] for user in users],
+        )
+
+    return jsonify({
+        "code": 200,
+        "message": "it worked!",
```
mayhem: lol

listenbrainz/webserver/static/js/jsx/follow-users.jsx
```diff
@@ -42,6 +46,43 @@ export class FollowUsers extends React.Component {
...
+      }),
+    })
+    .then(response => response.json())
+    .then(data => {
+      console.debug(data);
+      console.debug("old List ID: " + this.state.listId);
+      this.setState(prevState => {
+        return {listId: data.list_id};
+      });
+      console.debug("new List ID: " + this.state.listId);
```
MonkeyDo: This console log won't necessarily work because setState is asynchronous.

You can pass an optional callback to setState as a second argument, and in that callback you can be sure the value has been updated. That's where you'll want that new List ID log.

listenbrainz/webserver/static/js/jsx/follow-users.jsx
```diff
@@ -42,6 +46,43 @@ export class FollowUsers extends React.Component {
...
+  }
+
+  newFollowList(event) {
+    this.setState(prevState => {
+      return {
+        users: [],
+        listId: null,
+     };
+    });
+    this.nameInput.value = "Untitled Follow List";
```
MonkeyDo: Just set the input value to `null` or an empty string to show the placeholder text (which has a sane default in profile.jsx)

listenbrainz/webserver/views/follow.py
```diff
@@ -1,47 +1,100 @@
...
-        Allow an LB user to follow the stream of one or more other LB users.
-    """
-
-    follow_list = []
-    for to_follow in user_list.split(","):
-        to_follow = to_follow.strip()
-        if not to_follow:
-            continue
-        follow_list.append(to_follow)
+    if user_list:
```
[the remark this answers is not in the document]
paramsingh: if a list of users was specified, I assumed that it is a new list without a name, so called it untitled. Let's discuss this on IRC.

listenbrainz/db/follow_list.py
```diff
@@ -0,0 +1,196 @@
...
+
+def update(list_id, name, member):
+    """ Update the name and members of the list with specified ID.
+
+    Args:
+        list_id (int): the row ID of th follow list
+        name (str): the new name of the list
+        member ([int]): ordered list of row IDs of new members of the follow list
+    """
+    with db.engine.begin() as connection:
```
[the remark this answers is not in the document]
paramsingh: there is sanity checking done in the `/follow/save` endpoint.

admin/sql/create_tables.sql
```diff
@@ -88,10 +88,20 @@ CREATE TABLE statistics.recording (
...
   id          SERIAL,
   created     TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
 
+CREATE TABLE follow_list (
+  id                SERIAL, -- PK
+  name              TEXT NOT NULL,
+  creator           INTEGER NOT NULL, -- FK to "user".id
+  private           BOOLEAN NOT NULL DEFAULT FALSE,
+  member            INTEGER ARRAY NOT NULL DEFAULT ARRAY[]::INTEGER[],
```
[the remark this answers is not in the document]
paramsingh: me thinks so too.

listenbrainz/webserver/static/js/jsx/follow-users.jsx
```diff
@@ -51,14 +92,32 @@ export class FollowUsers extends React.Component {
...
+              <button className="btn btn-primary" type="button" onClick={this.newFollowList.bind(this)}>
+                <span className="fa fa-plus" aria-hidden="true"></span> New List
+              </button>
+            </span>
+            <div style={{"width": "1px", "height": "100%"}}/>
+            <span className="input-group-btn btn-primary">
+              <button className="btn btn-primary" type="button" onClick={this.saveFollowList.bind(this)}>
+                <span className="fa fa-save" aria-hidden="true"></span> Save
+              </button>
+            </span>
```
MonkeyDo: UI-wise, I think the group input+New List+Save should appear on the right of what is currently the panel-body.

I would suggest putting two divs in panel-body, with `col-md-9` and `col-md-3` classes respectively, and arrange input+New List+Save in that second column.

Quick sketch:
capture d ecran 2019-03-01 a 12 24 29

listenbrainz/webserver/static/js/jsx/follow-users.jsx
```diff
@@ -42,6 +46,43 @@ export class FollowUsers extends React.Component {
     });
   }
 
+  saveFollowList(event) {
+    var listName = this.state.listName;
+    if (this.nameInput.value != undefined && this.nameInput.value != "") {
```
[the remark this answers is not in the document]
paramsingh: whoops, i've read about this. didn't cross my mind at all, coming from python land.

listenbrainz/webserver/static/js/jsx/follow-users.jsx
paramsingh: the placeholder text is the name of the old list, so if we create a new list, we can't make it show the placeholder text. :sheepish:

listenbrainz/webserver/static/js/jsx/follow-users.jsx
paramsingh: whoops.

listenbrainz/webserver/static/js/jsx/follow-users.jsx
paramsingh: Ok, fixed now.

listenbrainz/webserver/views/follow.py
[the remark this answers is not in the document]
paramsingh: I've made this more clear, the new list doesn't have a name but has members.

listenbrainz/webserver/views/follow.py
```diff
@@ -1,47 +1,100 @@
...
-        "mode"                 : "follow",
-        "follow_list"          : follow_list,
-        "spotify_access_token" : spotify_access_token,
-        "web_sockets_server_url": current_app.config['WEBSOCKETS_SERVER_URL'],
+        "user": user_data,
+        "mode": "follow",
+        "follow_list": follow_list_members,
+        "spotify_access_token": spotify_access_token,
+        "web_sockets_server_url": current_app.config["WEBSOCKETS_SERVER_URL"],
+        "save_url": "http://0.0.0.0/follow/save", #TODO
```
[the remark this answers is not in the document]
paramsingh: done.

paramsingh:
Ok, I've addressed all issues mentioned now. Take a look again!

listenbrainz/webserver/static/js/jsx/follow-users.jsx
paramsingh: We discussed that it'd be better for @MonkeyDo to do this in a new PR, given my great frontend skillz.

listenbrainz/webserver/views/follow_api.py
```diff
@@ -0,0 +1,55 @@
...
+@follow_api_bp.route("/save", methods=["POST", "OPTIONS"])
+def save_list():
+    creator = _validate_auth_header()
+    raw_data = request.get_data()
+    try:
+        data = ujson.loads(raw_data.decode("utf-8"))
+    except ValueError as e:
+        log_raise_400("Cannot parse JSON document: %s" % str(e), raw_data)
+
+    list_name = data['name']
```
mayhem: If name is missing we'll throw a 500. Should be 400. Let's improve the param checking, please.

listenbrainz/webserver/views/follow_api.py
```diff
@@ -0,0 +1,55 @@
...
+    members = db_user.validate_usernames(data['users'])
```
mayhem: Same.

listenbrainz/webserver/views/follow_api.py
```diff
@@ -0,0 +1,55 @@
...
+            raise Unauthorized("You can only edit your own lists.")
+
+        # update the old list
+        db_follow_list.update(
+            list_id=list_id,
+            name=list_name,
+            members=[member['id'] for member in members],
+        )
+
+    return jsonify({
```
mayhem: Maybe user json_error_wrapper here too?

PR #533, opened 2019-03-01
 Push latest_listen_ts in the listens page 

paramsingh:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Minor / simple change (like a typo)
    * ( ) Other

# Problem

we need the timestamp of the latest listen for each user.

# Solution

Send it and add test.

PR #535, opened 2019-03-02
Add latest listen ts to get_listens endpoint

paramsingh:
#533 but for the API endpoint.

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -116,6 +116,14 @@ def get_listens(user_name):
     for listen in listens:
         listen_data.append(listen.to_api())
 
+    latest_listen = db_conn.fetch_listens(
+        user_name,
+        limit=1,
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/api.py
```diff
@@ -116,13 +116,22 @@ def get_listens(user_name):
     for listen in listens:
         listen_data.append(listen.to_api())
 
+    latest_listen = db_conn.fetch_listens(
+        user_name,
+        limit=1,
+        to_ts=max_ts,
+    )
+    latest_listen_ts = latest_listen[0].ts_since_epoch if len(latest_listen) > 0 else 0
```
alastair: This assumes that the user has at least one listen, it should have a guard around it, because this won't always be the case

alastair: my bad, I didn't see if the if-check the first time. Ignore me!

PR #537, opened 2019-03-04
fix LB-426: Fallback to embedded player for non-premium spotify users

MonkeyDo:
# Summary

* This is a…
    * (x) Bug fix

# Problem

I misunderstood the Spotify docs: non-premium users can't use the Web Playback SDK (the new player in the react pages).

# Solution

Thankfully it's easy to check if a user is premium, and we can fall back to the embedded iframe player if it's not the case.
Non-premium users should be able to play the full tracks in the embedded player. Unfortunately, they'll need to press 'play' every new song.

### DIFF
--- a/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
+++ b/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
@@ -20,12 +20,12 @@ function getSpotifyUriFromListen(listen) {
 export class SpotifyPlayer extends React.Component {
 
   _spotifyPlayer;
-  _accessToken;
   _firstRun = true;
 
   constructor(props) {
     super(props);
     this.state = {
+      accessToken: props.spotify_access_token,
       currentSpotifyTrack: null,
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/static/js/jsx/spotify-player.jsx
```diff
@@ -142,6 +150,15 @@ export class SpotifyPlayer extends React.Component {
     this.setState({ errorMessage: error });
   }
 
+  handleAccountError(error) {
+    const errorMessage = 'Failed to validate Spotify premium account';
```
paramsingh: Is this function only called when validating premium accounts? 401s and 403s might be due to different reasons?

MonkeyDo: I modified the error message, removing 'premium'.
403 is for non-premium users, 401 if the spotify token is missing or invalid.
Both should result in the same action and trigger a fallback to embedded iframe

PR #538, opened 2019-03-04
Open links in new tab

MonkeyDo:
# Summary

* This is a…

# Problem
Playback stops when you click links.
Open in new tab instead.

### DIFF
--- a/listenbrainz/webserver/static/js/jsx/profile.jsx
+++ b/listenbrainz/webserver/static/js/jsx/profile.jsx
@@ -282,7 +282,9 @@ class RecentListens extends React.Component {
                                 </abbr>
                               </td>
                             }
-                            {(this.state.mode === "follow" || this.state.mode === "recent") && <td><a href={`/user/${listen.user_name}`}>{listen.user_name}</a></td>}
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/static/js/jsx/profile.jsx
```diff
@@ -282,7 +282,9 @@ class RecentListens extends React.Component {
                                 </abbr>
                               </td>
                             }
-                            {(this.state.mode === "follow" || this.state.mode === "recent") && <td><a href={`/user/${listen.user_name}`}>{listen.user_name}</a></td>}
+                            {(this.state.mode === "follow" || this.state.mode === "recent") &&
```
mayhem: Hmmm. this is starting to get ugly. :(

MonkeyDo: Agreed. I've started thinking about what needs to be refactored, after initial beta testing.
The 'mode' certainly is part of that.

paramsingh:
Merging this, considering we don't have any tests for the frontend. speaking of which, we should have some tests :D

PR #540, opened 2019-03-07
Add a new endpoint to refresh spotify tokens

paramsingh:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem

* JIRA ticket (_optional_): 

Tokens expired, leading to errors in the frontend.

# Solution

I added an endpoint to which you can send GET requests to get a new token if needed.

Also, I improved the error handling to raise API specific errors which return json instead of html.

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -1,6 +1,6 @@
 import ujson
 from flask import Blueprint, request, jsonify, current_app
-from werkzeug.exceptions import BadRequest, InternalServerError, Unauthorized, ServiceUnavailable, NotFound
+from listenbrainz.webserver.errors import APIBadRequest, APIInternalServerError, APIUnauthorized, APINotFound
 from listenbrainz.db.exceptions import DatabaseException
 from listenbrainz.webserver.decorators import crossdomain
 from listenbrainz.webserver.views.follow import parse_user_list
@@ -68,10 +68,10 @@ def submit_listen():
 
     try:
         insert_payload(payload, user, listen_type=_get_listen_type(data['listen_type']))
-    except ServiceUnavailable as e:
+    except APIServiceUnavailable as e:
         raise
     except Exception as e:
-        raise InternalServerError("Something went wrong. Please try again.")
+        raise APIInternalServerError("Something went wrong. Please try again.")
 
     return jsonify({'status': 'ok'})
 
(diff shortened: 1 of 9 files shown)


listenbrainz/webserver/views/api.py
```diff
@@ -290,29 +290,36 @@ def validate_token():
             'message': 'Token valid.'
         })
 
+
+@api_bp.route('/refresh-spotify-token', methods=['POST'])
+@crossdomain(headers='Authorization, Content-Type')
+def refresh_spotify_token():
```
alastair: What's this method used for?

listenbrainz/webserver/views/profile.py
```diff
@@ -258,3 +261,23 @@ def connect_spotify_callback():
         flash.warn('Unable to authenticate with Spotify (error {})'.format(e.args[0]))
 
     return redirect(url_for('profile.connect_spotify'))
+
+
+@profile_bp.route('/refresh-spotify-token', methods=['GET'])
```
alastair: This feels like it should be POST, not GET

listenbrainz/webserver/views/profile.py
```diff
@@ -258,3 +261,23 @@ def connect_spotify_callback():
...
+@crossdomain()
+@api_login_required
+def refresh_spotify_token():
+    spotify_user = spotify.get_user(current_user.id)
+    if not spotify_user:
+        raise APINotFound("User has not authenticated to Spotify")
```
alastair: I don't know if 404 is a good error here. The page still exists, it just can't be run. 400 doesn't seem right either:
> the server cannot or will not process the request due to something that is perceived to be a client error

Maybe 409?

listenbrainz/webserver/views/profile.py
MonkeyDo: It's defined as such for the api:
https://github.com/metabrainz/listenbrainz-server/blob/f61e811ddee8911990dcfdf9893115be0646c816/listenbrainz/webserver/views/api.py#L294

listenbrainz/webserver/views/profile.py
```diff
@@ -258,3 +261,23 @@ def connect_spotify_callback():
...
+    if spotify_user.token_expired:
+        try:
+            spotify_user = spotify.refresh_user_token(spotify_user)
+        except spotify.SpotifyAPIError:
+            raise APIServiceUnvailable("Cannot refresh Spotify token right now")
```
alastair: Unavailable

PR #543, opened 2019-03-08
Remove the sort option from the recent page

paramsingh:
Simple really.


paramsingh:
I only delete already written code.

PR #545, opened 2019-03-11
Refresh the Spotify token for the front-end player

MonkeyDo:
# Summary

* This is a…
    * (x) Feature addition

# Problem

The Spotify player's token expires after an hour. That obviously sucks.

# Solution

After merging PR #540, use the new endpoint to get a refreshed token (or the existing one if still valid) from the server.

### DIFF
--- a/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
+++ b/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
@@ -32,18 +32,19 @@ export class SpotifyPlayer extends React.Component {
       durationMs: 0,
       direction: props.direction || "down"
     };
+    this.connectSpotifyPlayer = this.connectSpotifyPlayer.bind(this);
+    this.disconnectSpotifyPlayer = this.disconnectSpotifyPlayer.bind(this);
+    this.getAlbumArt = this.getAlbumArt.bind(this);
+    this.handleError = this.handleError.bind(this);
+    this.handlePlayerStateChanged = this.handlePlayerStateChanged.bind(this);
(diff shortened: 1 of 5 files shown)


listenbrainz/webserver/views/profile.py
```diff
@@ -263,7 +263,7 @@ def connect_spotify_callback():
     return redirect(url_for('profile.connect_spotify'))
 
 
-@profile_bp.route('/refresh-spotify-token', methods=['GET'])
+@profile_bp.route('/refresh-spotify-token', methods=['POST'])
```
alastair: there are tests that use `client.get` which will fail now

MonkeyDo:
Pretty hard to test, but I believe this is now working.

I had a look at spotify's web player source. It's partially obfuscated, but it seems that they use the token's expiry timestamp, which I currently don't have access to.

If the current mechanism doesn't work well as is, we'll need to return token expiry info from the API and implement a timer on the front-end to get a new token.

listenbrainz/webserver/static/js/jsx/spotify-player.jsx
```diff
@@ -63,7 +64,10 @@ export class SpotifyPlayer extends React.Component {
       },
     })
     .then(response =>{
-      if(response.status === 403 || response.status === 401){
+      if(response.status === 401){
```
mayhem: You were worried about this working well yesterday. I trust you're happy with it now?

MonkeyDo: After more testing and docs reading, yes, I think it should work.

PR #547, opened 2019-03-12
Follow page playlist improvements

MonkeyDo:
# Summary

* This is a…
    * (x) Bug fix

# Problem
- Sorting tracks by time on the follow page should have newest at the bottom (ascending, got it wrong in the commit message)
- Tracks were being skipped when detecting the end of the previous track: we receive multiple events in a short burst at the end of a track

### DIFF
--- a/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
+++ b/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
@@ -26,7 +26,7 @@ export class SpotifyPlayer extends React.Component {
     super(props);
     this.state = {
       accessToken: props.spotify_access_token,
-      currentSpotifyTrack: null,
+      currentSpotifyTrack: {},
       playerPaused: true,
       progressMs: 0,
       durationMs: 0,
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/static/js/jsx/spotify-player.jsx
```diff
@@ -42,6 +42,7 @@ export class SpotifyPlayer extends React.Component {
     this.isCurrentListen = this.isCurrentListen.bind(this);
     this.playListen = this.playListen.bind(this);
     this.playNextTrack = this.playNextTrack.bind(this);
+    this.debouncedPlayNextTrack = _.debounce(this.playNextTrack, 500, {leading:true, trailing:false});
```
mayhem: what is 500 here? half a second?

MonkeyDo:
Not ready to merge just yet

listenbrainz/webserver/static/js/jsx/spotify-player.jsx
MonkeyDo: Correct. This being the only place we use the debounced function, you can still click next like  a madman on the controls

PR #558, opened 2019-03-18
Fix LB-432: Check spotify token scopes and fallback graciously

MonkeyDo:
# Summary

* This is a…
    * (x) Bug fix

# Problem
https://tickets.metabrainz.org/projects/LB/issues/LB-432
With a spotify account linked for only recording listens, the resulting scope error was not recognised as such, and the player entered a loop of reconnection/retry until Spotify complained about too many attempts.

I implemented an initial check of the token scopes to save some other error being thrown (and caught adequately) by the Spotify web player.

While I was at it I improved the component unmounting in case of fallback to embedded player.

### DIFF
--- a/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
+++ b/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
@@ -33,6 +33,7 @@ export class SpotifyPlayer extends React.Component {
       durationMs: 0,
       direction: props.direction || "down"
     };
+
     this.connectSpotifyPlayer = this.connectSpotifyPlayer.bind(this);
     this.disconnectSpotifyPlayer = this.disconnectSpotifyPlayer.bind(this);
     this.getAlbumArt = this.getAlbumArt.bind(this);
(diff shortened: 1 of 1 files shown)


MonkeyDo:
> We store the permissions we've taken from the user in the db. Should we just use that instead? I can pass it in props with the token?

I hadn't though about it. Yes, that's much better. Is there stored somewhere the distinction wether the user is linked with the feature "play listens", or just an array of permissions?

paramsingh:
@MonkeyDo We have a flag for `record_listens` but not for `play listens`. But the permissions for both features don't intersect at all, and i think it would just be a simple check of whether `streaming` is in the list of permissions or not.

MonkeyDo:
Spotify does things a bit weird sometimes. As far as I know, for the front-end player to work properly, I need to check the user has either:
- `https://api.spotify.com/v1/melody/v1/check_scope?scope=web-playback`; note the scope `web-playback`, which is not [one of the existing scopes](https://developer.spotify.com/documentation/general/guides/scopes/) but probably a shortcut for:
- ["streaming", "user-read-birthdate", "user-read-email", "user-read-private"] scopes.


I think the simplest will be to return the array of scopes along with the token on page load, as well as during a refresh-token API call.
I can then compare with the array of scopes mentioned above, saving an HTTP call.

MonkeyDo:
Incorporated into #560

PR #560, opened 2019-03-19
Put spotify permissions in props as well

paramsingh:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem

* JIRA ticket (_optional_): 

We had to make requests to the spotify API to check if streaming worked in #558 .

# Solution

Return the permissions we stored in props and the refresh-token endpoint.


listenbrainz/webserver/views/follow.py
```diff
@@ -39,12 +40,13 @@ def follow(user_list):
...
     }
-    spotify_access_token = db_spotify.get_token_for_user(current_user.id)
+    spotify_data = spotify.get_user_dict(current_user.id)
     props = {
         "user": user_data,
         "mode": "follow",
         "follow_list": follow_list_members,
-        "spotify_access_token": spotify_access_token,
+        "spotify_access_token": spotify_data.get("access_token", ""),
+        "spotify_permission": spotify_data.get("permission", ""),
```
MonkeyDo: At this point, maybe we should consider returning an object for spotify props in all the views?  I.E:
```
"spotify": {
    "access_token": XYZ,
    "permissions":XYZ
}
```

It'll be a bit less polluting and better match the refresh-token response structure.

paramsingh: Makes sense, I didn't want to break existing stuff which is why I didn't change it. I'll do it.

listenbrainz/webserver/static/js/jsx/profile.jsx
```diff
@@ -352,13 +352,13 @@ class RecentListens extends React.Component {
...
           <div className="col-md-4" style={{ position: "-webkit-sticky", position: "sticky", top: 20 }}>
-            {this.props.spotify_access_token && this.state.isSpotifyPremium !== false ?
+            {this.props.spotify.access_token && this.state.isSpotifyPremium !== false ?
               <SpotifyPlayer
                 APIService={this.APIService}
                 ref={this.spotifyPlayer}
                 listens={spotifyListens}
                 direction={this.state.direction}
-                spotify_access_token={this.props.spotify_access_token}
+                spotify={this.props.spotify}
```
MonkeyDo: Maybe rename that prop to `spotify_user` for clarity ?

paramsingh:
I've incorporated the changes in #558 here. Added a new function for permission errors because we don't want a popup on every page for users who don't want to listen to music.

PR #563, opened 2019-03-19
Get latest listen correctly in the API endpoint

paramsingh:
Fixes https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/54067

### DIFF
--- a/listenbrainz/tests/integration/test_api.py
+++ b/listenbrainz/tests/integration/test_api.py
@@ -66,6 +66,11 @@ class APITestCase(IntegrationTestCase):
         # check for latest listen timestamp
         self.assertEqual(data['latest_listen_ts'], ts)
 
+        # request with min_ts should work
+        response = self.client.get(url, query_string = {'min_ts': int(time.time())})
+        self.assert200(response)
+
+
         # checkt that recent listens are fectched correctly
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/api.py
```diff
@@ -119,7 +119,7 @@ def get_listens(user_name):
     latest_listen = db_conn.fetch_listens(
         user_name,
         limit=1,
-        to_ts=max_ts,
+        to_ts=int(time.time()),
```
alastair: I think it'd make sense to only calculate time.time() once (it's also used if max_ts and min_ts aren't set)

listenbrainz/tests/integration/test_api.py
```diff
@@ -66,6 +66,11 @@ def test_get_listens(self):
         # check for latest listen timestamp
         self.assertEqual(data['latest_listen_ts'], ts)
 
+        # request with min_ts should work
+        response = self.client.get(url, query_string = {'min_ts': int(time.time())})
+        self.assert200(response)
+
```
alastair: It'd also be good to have a test where you select a max_ts lower than the most recent listen to ensure that you get both the listens that you query for and the timestamp of the most recent one.

PR #564, opened 2019-03-19
Make spotify reader parse listens in a less restrictive way

paramsingh:
# Summary

* This is a…
    * (x) Bug fix
    * () Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem

* JIRA ticket (_optional_): 

There were errors in JIRA because we expected some keys from spotify that didn't exist. 
# Solution
This PR makes spotify reader more resistant to missing keys.

### DIFF
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -59,26 +59,32 @@ def _convert_spotify_play_to_listen(play, listen_type):
             'listened_at': int(parser.parse(play['played_at']).timestamp()),
         }
 
+    if track is None:
+        return None
 
-    album = track['album']
-    artists = track['artists']
-    album_artists = album['artists']
-    artist_name = ', '.join([a['name'] for a in artists])
-    album_artist_name = ', '.join([a['name'] for a in album_artists])
+    album = track.get('album', {})
+    artists = track.get('artists', [])
+    album_artists = album.get('artists', [])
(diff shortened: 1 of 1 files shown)


listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -59,26 +59,32 @@ def _convert_spotify_play_to_listen(play, listen_type):
...
+        'duration_ms': track.get('duration_ms'),
+        'spotify_album_id': album.get('external_urls', {}).get('spotify'),
         # Named 'release_*' because 'release_name' is an official name in the docs
         'release_artist_name': album_artist_name,
-        'release_artist_names': [a['name'] for a in album_artists],
+        'release_artist_names': [a.get('name') for a in album_artists if a.get('name')],
         # Named 'album_*' because Spotify calls it album and this is spotify-specific
-        'spotify_album_artist_ids': [a['external_urls']['spotify'] for a in album_artists],
+        'spotify_album_artist_ids': [
+            a.get('external_urls', {}).get('spotify') for a in album_artists if a.get('external_urls', {}).get('spotify')
```
mayhem: I generally dislike this form -- it is far too perl like for my taste. And the get is done twice in this case -- I believe it to be efficient, but still done twice and the code is less readable. I would prefer an unrolled loop with the assignment and then an if and then re-use the variable if the if condition is true. 

This is mostly a philosophical question -- this should not prevent you from merging this code.

paramsingh:
>Are there any tests that should be updated? Or added?

I'll have to take a look again.

PR #566, opened 2019-03-22
LB-439: Allow cross-origin requests for playing now and listens api endpoints

bluefantail:
# Summary

* This is a…
    * ( ) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * (x) Other

* **Describe this change in 1-2 sentences**: Enables cross-origin requests for two API endpoints so requests can be made to retrieve basic user activity like recent listen info via the browser / javascript.

# Problem

It's currently not possible for a website to make a request to any of the endpoints that get data about a users current or recent listens, as a cross-origin request error is thrown in the console (even though it's possible in other non-browser based applications).

* JIRA ticket: [LB-439](https://tickets.metabrainz.org/browse/LB-439)

# Solution

As the GET endpoints don't use any authorisation, this change adds `@crossdomain()` (used in a few other places already) to `listens` and `playing-now`, allowing cross-origin requests from all domains.

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -77,6 +77,7 @@ def submit_listen():
 
 
 @api_bp.route("/user/<user_name>/listens")
+@crossdomain()
 @ratelimit()
 def get_listens(user_name):
     """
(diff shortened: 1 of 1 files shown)


bluefantail:
https://tickets.metabrainz.org/browse/LB-439

bluefantail:
> Not sure why they didn't have this before, but it makes sense to me. Thanks for the PR!

No problem!

PR #573, opened 2019-05-17
Fix and improve spotify search functionality

MonkeyDo:
# Summary

* This is a…
    * (x) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

- properly display currently playing track in list
- skip to next track if no track is found
- fix call to wrong function

### DIFF
--- a/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
+++ b/listenbrainz/webserver/static/js/jsx/spotify-player.jsx
@@ -89,11 +89,13 @@ export class SpotifyPlayer extends React.Component {
         </span>, "Found a match");
         return this.play_spotify_uri(track.uri);
       }
-      return this.handleWarning("Could not find track on Spotify");
+      this.handleWarning("Could not find track on Spotify");
+      this.playNextTrack();
     })
     .catch(errorObject => {
       if(errorObject.status === 401){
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/static/js/jsx/spotify-player.jsx
```diff
@@ -89,11 +89,13 @@ export class SpotifyPlayer extends React.Component {
         </span>, "Found a match");
         return this.play_spotify_uri(track.uri);
       }
-      return this.handleWarning("Could not find track on Spotify");
+      this.handleWarning("Could not find track on Spotify");
+      this.playNextTrack();
```
paramsingh: Not returning anything here?

MonkeyDo: No need since it's the last link in our promise chain. We don't have anything to return, and it's not strictly needed to return anything (https://stackoverflow.com/questions/53499297/is-the-return-type-of-then-a-promise-even-with-no-return-statement#comment93867901_53499297)

paramsingh: Alright, makes sense.

PR #575, opened 2019-05-22
Influx bug

mayhem:
Fix the test in param's PR


mayhem:
Merging as per our IRC discussion.

PR #577, opened 2019-05-22
Skip listens which cause client errors in influx

paramsingh:
This is a temporary fix to get the pipeline moving,
while we look into better solutions to the type mismatch problem.


paramsingh:
the function `current_app.logger.error` pushes every error to sentry. It's why we use the flask app to log even in scripts.

PR #578, opened 2019-06-04
table to store recommended playlist of every user

vansika:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

Table *recommendation* consists of columns namely, "user_id", "daily_delight", "last_updated" and "listened_to".

The column 'user_id' is a foreign key to table 'user', which will store the user id of the user. 'daily_delight' will store recommended tracks of a user in JSON format. 'last-updated' as the name suggests will keep track of when the playlist of a user was last updated. Finally, 'listened_to' will store two values at any instant, 'true' or 'false'. True indicates that the user has at least listened to one song from the recommended playlist whereas false indicates that the user has not listened to any of the songs. A new playlist will be generated for users whose 'listened_to' field is true.

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -105,7 +105,7 @@ CREATE TABLE follow_list (
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
 
 CREATE TABLE recommendation (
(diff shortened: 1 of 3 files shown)


vansika:
Since this PR has been opened, the Listenbrainz-labs code base has undergone a lot changes, and hence a new PR should be made.

PR #580, opened 2019-06-14
LB-452: Add script to replay user listens to fix bad data

paramsingh:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem

* JIRA ticket (_optional_): 

Some users have some bad data which is hard to fix via influx itself. This script replay's user listens.

# Solution

Write a class with a filter function that can be used to fix data as needed. 

# Action

We need to consider listens coming in via the API too while writing replaying users. A solution to this could probably just be us stopping influx-writer while replaying listens.

### DIFF
--- a/listenbrainz/listen_replay/replay_user.py
+++ b/listenbrainz/listen_replay/replay_user.py
@@ -1,5 +1,6 @@
 """ This script should be used to replay user listens to fix bad data.
 """
+import abc
 import json
 import uuid
 
@@ -11,15 +12,21 @@ from listenbrainz.webserver.influx_connection import init_influx_connection
 from listenbrainz.listenstore.influx_listenstore import DUMP_CHUNK_SIZE
 
 
-class UserReplayer:
+class UserReplayer(abc.ABC):
     def __init__(self, user_name):
         self.user_name = user_name
         self.max_time = datetime.now()
         self.app = create_app()
 
 
-    def filter_function(self, row):
-        return row
+    @abc.abstractmethod
+    def filter(self, row):
+        """ Modify row as needed by the subclass
+
+        Returns:
+            row (dict): modified row as needed
+                or None if row needs to be removed
+        """
 
 
     def convert_to_influx_insert_format(self, row, measurement):
--- a/listenbrainz/listen_replay/replay_user.py
+++ b/listenbrainz/listen_replay/replay_user.py
@@ -27,6 +27,7 @@ class UserReplayer(abc.ABC):
             row (dict): modified row as needed
                 or None if row needs to be removed
         """
+        pass
 
 
     def convert_to_influx_insert_format(self, row, measurement):
--- a/listenbrainz/replay_user.py
+++ b/listenbrainz/replay_user.py
@@ -51,7 +51,6 @@ class UserReplayer:
                 count += 1
                 if apply_filter:
                     row = self.filter_function(row)
-                current_app.logger.info(json.dumps(row, indent=4))
                 if row:
                     rows.append(self.convert_to_influx_insert_format(row, quote(dest)))
             self.ls.write_points_to_db(rows)
--- a/listenbrainz/replay_user.py
+++ b/listenbrainz/replay_user.py
@@ -1,0 +1,97 @@
+""" This script should be used to replay user listens to fix bad data.
+"""
+import json
+import uuid
+
+from datetime import datetime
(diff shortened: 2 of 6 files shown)


listenbrainz/replay_user.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class UserReplayer:
+    def __init__(self, user_name):
+        self.user_name = user_name
+        self.max_time = datetime.now()
+        self.app = create_app()
+
+
+    def filter_function(self, row):
```
mayhem: I think we should make this an abstract function that must be overriden. Right now this could effectively be the same as a quiet failure...

mayhem: In fact, it might be nice to provide a skeleton example script that overrides the abstract function and serves as a template for replaying other users.

listenbrainz/replay_user.py
```diff
@@ -0,0 +1,93 @@
...
+        return row
+
+
+    def convert_to_influx_insert_format(self, row, measurement):
+        data = {
+            'measurement': measurement,
+            'time': convert_to_unix_timestamp(row['time']),
+        }
+
+        if 'dedup_tag' in row:
```
mayhem: Hmm. This is the innermost loop, meaning that this will be executed gazillions of times and ought to be tight. I wonder if it would be faster to copy all the fields over (using a built-in optimized function and then overwrite the time/dedup tags?

listenbrainz/replay_user.py
```diff
@@ -0,0 +1,93 @@
...
+
+        data['fields'] = {}
+        for field in row:
+            if field != 'time' and field != 'dedup_tag':
+                data['fields'][field] = row[field]
+
+        return data
+
+
+    def copy_measurement(self, src, dest, apply_filter=False):
```
mayhem: Why would we not want to apply a filter?

listenbrainz/replay_user.py
```diff
@@ -0,0 +1,93 @@
...
+        offset = 0
+        while True:
+            result = self.ls.get_listens_batch_for_dump(src, self.max_time, offset)
+            rows = []
+            count = 0
+            for row in result.get_points(get_measurement_name(src)):
+                count += 1
+                if apply_filter:
+                    row = self.filter_function(row)
+                current_app.logger.info(json.dumps(row, indent=4))
```
mayhem: I suppose this should go too.

listenbrainz/replay_user.py
```diff
@@ -0,0 +1,93 @@
...
+                'REDIS_HOST': current_app.config['REDIS_HOST'],
+                'REDIS_PORT': current_app.config['REDIS_PORT'],
+                'REDIS_NAMESPACE': current_app.config['REDIS_NAMESPACE'],
+                'INFLUX_HOST': current_app.config['INFLUX_HOST'],
+                'INFLUX_PORT': current_app.config['INFLUX_PORT'],
+                'INFLUX_DB_NAME': current_app.config['INFLUX_DB_NAME'],
+            })
+            current_app.logger.info("Done!")
+
+            new_measurement_name = str(uuid.uuid4())
```
mayhem: There is no exception handling here at all. Let's put that into the demo main function as I suggested above.

listenbrainz/replay_user.py
paramsingh: We do A -> B with filter, delete A, then do B -> A without filter, and delete B. the B -> A copy would not need a filter.

mayhem: Got it -- thanks!

paramsingh:
I've added a demo class. I also added a manage.py cli to make usage easier.

listenbrainz/listen_replay/replay_user.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @abc.abstractmethod
+    def filter(self, row):
+        """ Modify row as needed by the subclass
+
+        Returns:
+            row (dict): modified row as needed
+                or None if row needs to be removed
+        """
+
```
mayhem: pass ?

paramsingh: I remember reading some PEP8 / flake8 rules about not using passes for empty functions / classes with docstrings but I can't find it now. I'll add the pass.

PR #582, opened 2019-06-14
Upgrade pyyaml and clean-css due to security vulnerability

paramsingh:
As it says on the tin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1725,21 +1725,17 @@
       }
     },
     "clean-css": {
-      "version": "3.4.28",
-      "resolved": "https://registry.npmjs.org/clean-css/-/clean-css-3.4.28.tgz",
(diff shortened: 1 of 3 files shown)


paramsingh:
@MonkeyDo, the clean-css vulnerability came from package-lock.json which wasn't in package.json, i guess it came from `less-plugin-clean-css`, I manually installed it, is that okay? Is there a better way?

MonkeyDo:
> @MonkeyDo, the clean-css vulnerability came from package-lock.json which wasn't in package.json, i guess it came from `less-plugin-clean-css`, I manually installed it, is that okay? Is there a better way?

Usually, you would update the root package that the project depends upon, in this case updating  `less-plugin-clean-css` to a newer version if there is one, but it doesn't seem to be the case here, so I think you did it correctly.

I think `less-plugin-clean-css` and `clean-css` should be in devDependencies though, unless they are used once bundled?

PR #589, opened 2019-06-24
Don't stop getting user listens on 404s

paramsingh:
spotify is being flaky and the api endpoint 404s randomly, we need to retry, but not give up completely if 404s cause problems.


paramsingh:
Small fix and seems urgent, so I'll merge and deploy for now.

PR #591, opened 2019-06-25
Add an API endpoint to get information about data dumps

paramsingh:
This endpoint is gonna be useful for getting information
about dumps when we need to download and import them
into the spark cluster.


listenbrainz/webserver/views/dump_api.py
```diff
@@ -0,0 +1,62 @@
+from flask import Blueprint, request, jsonify
+from listenbrainz.webserver.errors import APIBadRequest, APINotFound
+from listenbrainz.webserver.rate_limiter import ratelimit
+
+import listenbrainz.db.dump as db_dump
+
+dump_api_bp = Blueprint("dump_api_v1", __name__)
```
mayhem: Hmm, I'm not a fan of that name. I think we should think bigger and make a more generic "health-stats" or some such thing that doesn't narrow the scope too much.

listenbrainz/webserver/views/dump_api.py
```diff
@@ -0,0 +1,62 @@
...
+
+    **Example response**:
+
+    .. sourcecode:: json
+        {
+            "id": 1,
+            "timestamp": "20190625-165900"
+        }
+
+    :query id: *Required.* Integer specifying the ID of the dump
```
mayhem: What if I don't have one? If not specified, I think we should return the date and id of the latest dump...

paramsingh: I was returning 400 if there wasn't an ID provided, I've changed it now to return the latest.

listenbrainz/webserver/views/dump_api.py
paramsingh: Done after IRC discussion!

PR #592, opened 2019-06-29
Refactor dockerfiles

paramsingh:
Use docker multi-stage builds to only have the single Dockerfile
for both production and development. Also, some other
improvements.

* Install requirements for docs too for easier building of docs
* Remove unneeded bigquery service from integration tests
* Refactor dockerfiles to copy code as late as possible, to
make use of docker cache well


Dockerfile
```diff
@@ -40,16 +41,23 @@ WORKDIR /code/listenbrainz
...
 COPY requirements.txt /code/listenbrainz/
 RUN pip3 install --no-cache-dir -r requirements.txt
+RUN useradd --create-home --shell /bin/bash listenbrainz
 
-# Now install our code, which may change frequently
-COPY . /code/listenbrainz/
 
-# create a user named listenbrainz for storing dump file backups
-RUN useradd --create-home --shell /bin/bash listenbrainz
+# NOTE: The development image starts here
```
alastair: is it worth adding some whitespace here?

Dockerfile
```diff
@@ -59,3 +67,32 @@ RUN touch /var/log/stats.log /var/log/dump_create.log && chown listenbrainz:list
...
+RUN chown -R listenbrainz:listenbrainz /home/listenbrainz/backup /home/listenbrainz/ftp
+
+# Now install our code, which may change frequently
+COPY . /code/listenbrainz/
+
+RUN mkdir /static
+WORKDIR /static
+
+# Compile static files
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
```
alastair: not a good idea to run apt after copying code. If this is something that only happens in this image then it should be done as the first step.
If you don't want node_modules in production, you could do npm install in the base image, and then in the prod image just copy node_modules from the base image, run npm, and then delete it (I'm not sure with x thousand small files if this will be efficient or not). It'll prevent us from having to download from npm on every deploy, but it'd require adding package.json to the base image.

docker/docker-compose.integration.yml
```diff
@@ -24,35 +24,27 @@ services:
...
 
-  influx_writer:
+  listenbrainz:
     build:
       context: ..
       dockerfile: Dockerfile
-    command: python3 -m "listenbrainz.influx_writer.influx_writer"
+      target: listenbrainz-dev
+    command: py.test listenbrainz/tests/integration
+    image: listenbrainz
```
alastair: this has a build key _and_ an image key? what does this mean?

alastair:
looks good

Dockerfile
paramsingh: This was a mistake, it now copies the static files and installs and then copies code.

docker/docker-compose.integration.yml
paramsingh: This is to specify the name of the service image so it can be reused for the influx-writer service. (https://stackoverflow.com/a/35662191)

alastair: cool, I hadn’t seen this behaviour before.

PR #595, opened 2019-07-07
Fix bug where listens weren't added to spark dump archive

paramsingh:
This was caused by the recent refactor to add files to normal
dumps one by one. To do something similar for spark dumps would
require a relatively big refactor, so we just add all the files
at once for the spark dumps for now.

Added a more comprehensive test for spark dumps to make sure
this doesn't happen again.

### DIFF
--- a/listenbrainz/listenstore/tests/test_influxlistenstore.py
+++ b/listenbrainz/listenstore/tests/test_influxlistenstore.py
@@ -172,7 +172,7 @@ class TestInfluxListenStore(DatabaseTestCase):
         shutil.rmtree(temp_dir)
 
     def test_dump_listens_spark_format(self):
-        self._create_test_data(self.testuser_name)
+        expected_count = self._create_test_data(self.testuser_name)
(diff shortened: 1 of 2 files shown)


listenbrainz/listenstore/tests/test_influxlistenstore.py
```diff
@@ -172,20 +172,32 @@ def test_dump_listens(self):
...
         shutil.rmtree(temp_dir)
 
+    def assert_spark_dump_contains_listens(self, dump, expected_listen_count):
+        pxz_command = ['pxz', '--decompress', '--stdout', dump]
+        pxz = subprocess.Popen(pxz_command, stdout=subprocess.PIPE)
+
+        listen_count = 0
+        with tarfile.open(fileobj=pxz.stdout, mode='r|') as tar:
+            for member in tar:
+                file_name = member.name.split('/')[-1]
```
mayhem: Normally this should use os.path..., but I think in this case we can skip it.

PR #596, opened 2019-07-08
Skip listens with inserted_timestamps not in time range

paramsingh:
Theoretically, without this check we'd get listens that weren't inserted in the time range of the dump, leading to inconsistencies. We also can't do this in queries because influx is weird with fields with null values being queried, it doesn't return those listens with null `inserted_timestamp`. 

So we do the check in python.

### DIFF
--- a/listenbrainz/listenstore/influx_listenstore.py
+++ b/listenbrainz/listenstore/influx_listenstore.py
@@ -475,6 +475,14 @@ class InfluxListenStore(ListenStore):
                 result = self.get_incremental_listens_batch(username, start_time, end_time, offset)
             rows_added = 0
             for row in result.get_points(get_measurement_name(username)):
+
+                # make sure that listen was inserted in current dump's time range
+                # need to do this check in python, because influx doesn't
+                # do "IS NULL" operations and we have null inserted_timestamps from
+                # old data
+                if datetime.utcfromtimestamp(row.get('inserted_timestamp', 0)) > end_time:
+                    continue
+
                 listen = convert_influx_row_to_spark_row(row)
                 timestamp = convert_influx_to_datetime(row['time'])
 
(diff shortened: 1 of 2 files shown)


paramsingh:
I rebased this over #595 so that we could also test spark dumps for the edge cases like inserted_timestamps absence as well.

listenbrainz/listenstore/tests/test_influxlistenstore.py
```diff
@@ -238,6 +241,128 @@ def test_incremental_dump(self):
...
+        """ Incremental dumps should only consider listens that have
+        inserted_timestamps.
+        """
+        t = datetime.now()
+        sleep(1)
+        listens = generate_data(1, self.testuser_name, 1, 5)
+
+        # insert these listens into influx without an insert_timestamp
+        influx_rows = [listen.to_influx(quote(self.testuser_name)) for listen in listens]
+        for row in influx_rows[1:]:
```
mayhem: Why is the timestamp being popped off? To make comparisons in the test easier?

paramsingh: This is to test that the code doesn't bork with listens that do not have an `inserted_timestamp`. initial releases of LB didn't track when listens were inserted into influx, so we need to take care of that as well.

listenbrainz/listenstore/influx_listenstore.py
```diff
@@ -453,6 +453,22 @@ def write_spark_listens_to_disk(self, listens, temp_dir):
                     f.write('\n'.join([ujson.dumps(listen) for listen in listens[year][month]]))
                     f.write('\n')
 
+    def row_inserted_before(self, row, timestamp):
+        """ Check that the influx row passed was inserted to influx before the specified timestamp
```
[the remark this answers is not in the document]
paramsingh: before_or_equal, I'll rename.

PR #597, opened 2019-07-09
Schema to store recordings recommended to users

vansika:
As discussed on IRC.

### DIFF
--- a/admin/sql/create_foreign_keys.sql
+++ b/admin/sql/create_foreign_keys.sql
@@ -30,4 +30,28 @@ ALTER TABLE follow_list
     REFERENCES "user" (id)
     ON DELETE CASCADE;
 
+ALTER TABLE recommendation.cf_recording
+    ADD CONSTRAINT cf_recording_user_id_foreign_key
+    FOREIGN KEY (user_id)
+    REFERENCES "user" (id)
+    ON DELETE CASCADE;
+
+ALTER TABLE recommendation.cf_recording
+    ADD CONSTRAINT cf_recording_recommender_id_foreign_key
+    FOREIGN KEY (recommender_id)
+    REFERENCES recommendation.recommender (id)
+    ON DELETE CASCADE;
+
+ALTER TABLE recommendation.cf_recording_recommender_join
+    ADD CONSTRAINT cf_recording_recommender_join_recommender_id_foreign_key
+    FOREIGN KEY (recommender_id)
+    REFERENCES recommendation.recommender (id)
+    ON DELETE CASCADE;
+
+ALTER TABLE recommendation.cf_recording_recommender_join
+    ADD CONSTRAINT cf_recording_recommender_join_cf_recording_id_foreign_key
+    FOREIGN KEY (cf_recording_id)
+    REFERENCES recommendation.cf_recording (id)
+    ON DELETE CASCADE;
+
 COMMIT;
--- a/admin/sql/create_foreign_keys.sql
+++ b/admin/sql/create_foreign_keys.sql
@@ -36,12 +36,6 @@ ALTER TABLE recommendation.cf_recording
     REFERENCES "user" (id)
     ON DELETE CASCADE;
 
-ALTER TABLE recommendation.cf_recording
-    ADD CONSTRAINT cf_recording_recommender_id_foreign_key
-    FOREIGN KEY (recommender_id)
-    REFERENCES recommendation.recommender (id)
-    ON DELETE CASCADE;
-
 ALTER TABLE recommendation.cf_recording_recommender_join
     ADD CONSTRAINT cf_recording_recommender_join_recommender_id_foreign_key
     FOREIGN KEY (recommender_id)
--- a/admin/sql/create_foreign_keys.sql
+++ b/admin/sql/create_foreign_keys.sql
@@ -36,16 +36,22 @@ ALTER TABLE recommendation.cf_recording
     REFERENCES "user" (id)
     ON DELETE CASCADE;
 
-ALTER TABLE recommendation.cf_recording_recommender_join
-    ADD CONSTRAINT cf_recording_recommender_join_recommender_id_foreign_key
-    FOREIGN KEY (recommender_id)
-    REFERENCES recommendation.recommender (id)
+ALTER TABLE recommendation.recommender
+    ADD CONSTRAINT recommender_user_id_foreign_key
+    FOREIGN KEY (user_id)
+    REFERENCES "user" (id)
     ON DELETE CASCADE;
 
-ALTER TABLE recommendation.cf_recording_recommender_join
-    ADD CONSTRAINT cf_recording_recommender_join_cf_recording_id_foreign_key
-    FOREIGN KEY (cf_recording_id)
-    REFERENCES recommendation.cf_recording (id)
+ALTER TABLE recommendation.recording_session
+    ADD CONSTRAINT recording_session_recommender_session_foreign_key
+    FOREIGN KEY (session_id)
+    REFERENCES recommendation.recommender_session (id)
+    ON DELETE CASCADE;
+
+ALTER TABLE recommendation.recommender_session
+    ADD CONSTRAINT recommender_session_recommender_foreign_key
+    FOREIGN KEY (recommender_id)
+    REFERENCES recommendation.recommender (id)
     ON DELETE CASCADE;
 
 COMMIT;
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -54,8 +54,8 @@ CREATE TABLE recommendation.cf_recording (
   id                  SERIAL, -- PK
   user_id             INTEGER NOT NULL, --FK to "user".id
   recording_msid      UUID NOT NULL,
-  type                cf_recording_type,
-  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+  type                cf_recording_type NOT NULL,
+  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
 );
 
 CREATE TABLE recommendation.recommender (
@@ -63,14 +63,14 @@ CREATE TABLE recommendation.recommender (
   user_id             INTEGER NOT NULL, --FK to "user".id, denotes user who wrote software for this recommender.
   repository          TEXT NOT NULL,
   name                TEXT NOT NULL,
-  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
 );
 
 CREATE TABLE recommendation.recommender_session (
   id                  SERIAL, --PK
   recommender_id      INTEGER NOT NULL, --FK to recommendation.recommender.id
   user_id             INTEGER NOT NULL, --FK to "user".id, user for whom the recommendations are generated.
-  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
   data                JSONB
 );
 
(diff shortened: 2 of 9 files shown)


admin/sql/create_tables.sql
```diff
@@ -104,4 +104,11 @@ CREATE TABLE follow_list (
 );
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
 
+CREATE TABLE recommendation (
+    row_id                  SERIAL, --PK
+    user_id                 INTEGER NOT NULL, -- FK to "user".id
+    recording_msid          UUID NOT NULL,
+	last_used		        TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
```
paramsingh: we should add a `created` timestamp column too. it'll help later with dumps.

admin/sql/create_tables.sql
```diff
@@ -104,4 +104,11 @@ CREATE TABLE follow_list (
 );
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
 
+CREATE TABLE recommendation (
+    row_id                  SERIAL, --PK
```
paramsingh: this should probably just be `id`

admin/sql/create_tables.sql
```diff
@@ -104,4 +104,11 @@ CREATE TABLE follow_list (
 );
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
 
+CREATE TABLE recommendation (
```
paramsingh: I thought we were gonna create a seperate [schema](https://www.postgresql.org/docs/9.5/ddl-schemas.html) for this.

paramsingh:
The tables are named top_artist and similar_artist but they contain recording_msids, this is confusing, I'm not sure if I'm missing something here?

vansika:
> The tables are named top_artist and similar_artist but they contain recording_msids, this is confusing, I'm not sure if I'm missing something here?

@paramsingh  Top artist table will contain recordings that belong to top artists listened to by the user, similar is the case with similar artists. recording_msid can be used to recognize the recordings. Do you want to suggest some good names if it is confusing? :)

admin/sql/create_tables.sql
```diff
@@ -104,4 +104,27 @@ CREATE TABLE follow_list (
 );
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
 
+CREATE TABLE recommendation.cf_recording (
+  id                  SERIAL, -- PK
+  user_id             INTEGER NOT NULL, --FK to "user".id
+  msid                UUID NOT NULL,
+  recommender_id      INTEGER, --FK to recommendation.recommender.id
```
mayhem: should this also be NOT NULL?

admin/sql/create_tables.sql
```diff
@@ -104,4 +104,27 @@ CREATE TABLE follow_list (
...
+CREATE TABLE recommendation.recommender (
+  id                  SERIAL, --PK
+  repository          TEXT NOT NULL,
+  author_email        TEXT NOT NULL,
+  name                TEXT NOT NULL,
+  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+);
+
+CREATE TABLE recommendation.cf_recording_recommender_join(
+  last_used           TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
```
mayhem: also NOT NULL

admin/sql/create_tables.sql
vansika: Not sure. Whilst pushing recordings in the table, we won't have a recommender script, so I guess it will be left blank till we have a recommender script. What do you think?

admin/sql/create_tables.sql
vansika: Not null with Default now()?

mayhem: Yes, please.

admin/sql/create_tables.sql
mayhem: Ohh! I think we just found a bug! recommender_id should not be in this table -- it should only be in the join table, where it already is.

admin/sql/create_tables.sql
mayhem: yes.

admin/sql/create_tables.sql
```diff
@@ -104,4 +104,26 @@ CREATE TABLE follow_list (
...
+CREATE TABLE recommendation.cf_recording (
+  id                  SERIAL, -- PK
+  user_id             INTEGER NOT NULL, --FK to "user".id
+  msid                UUID NOT NULL,
+  type                recording_type,
+  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+);
+
+CREATE TABLE recommendation.recommender (
+  id                  SERIAL, --PK
```
alastair: recommenders don't have to be made by existing MB user? Does it make sense for us to link them here so that we can group better (e.g. list all of someone's recommenders)

mayhem: Are you suggesting that we link to MB accounts, rather than having random emails, etc? (just trying to understand your comment). Sounds good though.

alastair:
How do batches of recommendations work here? I understand a 'recommender' to be a tool written by a person that can give a bunch of recommendations to a user. How often are these tools run?
Should we have a session/run identifier to say that this block of recommendations comes from running the process once?
Do we expect recommenders to work on only one type (the `recording_type` type)? If so, perhaps this could be on the recommender table instead? If recommenders can do many types of recommendations, perhaps it's better on session instead?
It'd be nice to see some documentation of what the values in the `recording_type` type mean. Looking at their names, it's not immediately clear to me what these represent. Perhaps just a comment in the create_types.sql file is enough for now?
Is there a record of the input to the tool that performed this recommendation? That is, some way of saying "we took these recordings and mappings as input and then created x recommendations". I feel this is somewhat important because if we run the same tool more than once and get a different result, what changed to cause the difference in recommendations?

admin/sql/create_tables.sql
vansika: Recommenders of existing MB users sounds like a good idea.

vansika:
> Right now we have no idea on how often recommenders would run. We are trying to call people and build recommenders to be able to filter the one that generates best recommendations. ( We are thinking to invite people and organise a challenge after  summers for the same.

> cf_recording is one such table (using CF as an algorithm) that the recommenders can use to generate recommendations. recording_type are kind of candidate sets for recommenders to generate recommendations from. In future, we may have other tables containing candidate sets based on some other algo.

vansika:
There should be a comment for recording_type, true.

vansika:
If you mean keeping track of input recordings given to recommender scripts then there is a way out. In the join table, we can get cf_recording id corresponding to recommender id, cf_recording id can then be used to get the input track info from cf_recording table. ( I am not sure if I understood it right)

admin/sql/create_foreign_keys.sql
```diff
@@ -30,4 +30,22 @@ ALTER TABLE follow_list
...
+    REFERENCES "user" (id)
+    ON DELETE CASCADE;
+
+ALTER TABLE recommendation.cf_recording_recommender_join
+    ADD CONSTRAINT cf_recording_recommender_join_recommender_id_foreign_key
+    FOREIGN KEY (recommender_id)
+    REFERENCES recommendation.recommender (id)
+    ON DELETE CASCADE;
+
+ALTER TABLE recommendation.cf_recording_recommender_join
```
mayhem: This FK is not needed.

admin/sql/updates/2019-07-09-add-recommendation-table.sql
```diff
@@ -0,0 +1,48 @@
...
+    FOREIGN KEY (user_id)
+    REFERENCES "user" (id)
+    ON DELETE CASCADE;
+
+ALTER TABLE recommendation.cf_recording_recommender_join
+    ADD CONSTRAINT cf_recording_recommender_join_recommender_id_foreign_key
+    FOREIGN KEY (recommender_id)
+    REFERENCES recommendation.recommender (id);
+
+ALTER TABLE recommendation.cf_recording_recommender_join
```
mayhem: Nuke here too.

admin/sql/updates/2019-07-09-add-recommendation-table.sql
```diff
@@ -0,0 +1,48 @@
...
+  id                  SERIAL, --PK
+  repository          TEXT NOT NULL,
+  author_email        TEXT NOT NULL,
+  name                TEXT NOT NULL,
+  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW()
+);
+
+CREATE TABLE recommendation.cf_recording_recommender_join(
+  last_used           TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+  cf_recording_id     INTEGER, --FK to recommendation.cf_recording.id
```
mayhem: Since we're going to change the tracks periodically, we could not delete them from cf_recording if we had this FK. We should change this to: "recording_msid UUID NOT NULL"

admin/sql/create_tables.sql
```diff
@@ -104,4 +104,26 @@ CREATE TABLE follow_list (
 );
 ALTER TABLE follow_list ADD CONSTRAINT follow_list_name_creator_key UNIQUE (name, creator);
 
+CREATE TABLE recommendation.cf_recording (
```
mayhem: This file should be sorted on table name -- this wasn't quite done when you started working with it, but would you mind putting them in order?

admin/sql/create_tables.sql
mayhem: OK, then this should be a FK to the "user" table, which makes a lot more sense. Also we can remove the author_email then.

admin/sql/updates/2019-07-09-add-recommendation-table.sql
```diff
@@ -0,0 +1,65 @@
+BEGIN;
+
+CREATE SCHEMA recommendation;
+
+CREATE TABLE recommendation.cf_recording (
+  id                  SERIAL, -- PK
+  user_id             INTEGER NOT NULL, --FK to "user".id
+  recording_msid      UUID NOT NULL,
+  type                cf_recording_type,
+  created             TIMESTAMP WITH TIME ZONE DEFAULT NOW()
```
alastair: perhaps not null for this and the recommender table? What about `type` too?

mayhem: I wondered about this too, but I couldn't remember best practices. I'd like them and certainly on type. Good catch.

admin/sql/updates/2019-07-09-add-recommendation-table.sql
```diff
@@ -0,0 +1,65 @@
+BEGIN;
+
+CREATE SCHEMA recommendation;
+
+CREATE TABLE recommendation.cf_recording (
+  id                  SERIAL, -- PK
+  user_id             INTEGER NOT NULL, --FK to "user".id
+  recording_msid      UUID NOT NULL,
+  type                cf_recording_type,
```
paramsingh: cf_recording_type isn't created in this script.

vansika: <3

mayhem:
The failing build is a recurring transient failure, I think it is ok to merge, but we should look at the failure later today.

admin/sql/updates/2019-07-09-add-recommendation-table.sql
```diff
@@ -2,20 +2,22 @@ BEGIN;
 
 CREATE SCHEMA recommendation;
 
+CREATE TYPE cf_recording_type AS ENUM('top_artist', 'similar_artist');
```
mayhem: Oh whoops. Good you caught that.

PR #598, opened 2019-07-11
Make sure both timestamps in dumps are in UTC before comparing

paramsingh:
The database contains timestamps which are timezone aware and comparing them to timezone naive timestamps caused a bug (https://sentry.metabrainz.org/share/issue/d14b420991c045a48bfe1617e7e16c72/), make them both timestamp aware and then compare.

Missed this because there were no tests and it only happened when we got the end timestamp from the database. I've added a bunch of tests for the dump_manage.py script now.


paramsingh:
definitely get motivated to write more tests when you see a script error out at 4 o'clock in the morning :P

PR #606, opened 2019-08-21
Set up spark_reader config and uwsgi stuff

paramsingh:
This is to start a consumer that processes statistics.

## Open Questions

* Do we just want an independent container for stats? Or maybe we want to merge it with a generic `listenbrainz-jobs` container? For example: [LB-465](https://tickets.metabrainz.org/projects/LB/issues/LB-465?filter=allopenissues) shows that user export dump creation should also probably be done as a rabbitmq job instead of being done inside the request right now. Do we want to start a new container for that?

* Do we want a container running cron jobs to push statistics into the queue on the cluster now? Because that is something we can do just now to get stats back.

### DIFF
--- a/docker/beta/uwsgi/uwsgi.service
+++ b/docker/beta/uwsgi/uwsgi.service
@@ -43,5 +43,10 @@ then
     exec python3 manage.py run_follow_server -h 0.0.0.0 -p 3031
 fi
 
+if [ "${CONTAINER_NAME}" = "listenbrainz-stats-consumer-beta" ]
+then
+    cd /code/listenbrainz
+    exec python3 -m listenbrainz.spark.spark_reader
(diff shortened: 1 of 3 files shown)


docker/beta/uwsgi/uwsgi.service
```diff
@@ -43,5 +43,10 @@ then
     exec python3 manage.py run_follow_server -h 0.0.0.0 -p 3031
 fi
 
+if [ "${CONTAINER_NAME}" = "listenbrainz-stats-consumer-beta" ]
```
mayhem: Should we add this to the prod section as well, so we don't forget it?

paramsingh:
@mayhem @vansika ready for review.

PR #635, opened 2019-09-10
Revert "Update the musicbrainz ID of users if it has changed"

paramsingh:
So I realized from an email thread that user renames are not as easy as this. Every user has a measurement in influx which is their username basically, so if we change the name here, their listens won't show up.

I'd like to revert this while we think of better solutions.

Influx doesn't readily rename measurements. One solution that comes to mind is adding a new column with influx measurement name to the user table and using it everywhere to get listens.

### DIFF
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -246,24 +246,6 @@ def update_latest_import(musicbrainz_id, ts):
             raise DatabaseException
 
 
-def update_musicbrainz_id(id, musicbrainz_id):
-    """ Update the username (musicbrainz_id) of the user with specified row ID
-    """
-    with db.engine.connect() as connection:
-        try:
-            connection.execute(sqlalchemy.text("""
-                UPDATE "user"
-                   SET musicbrainz_id = :musicbrainz_id
-                 WHERE id = :id
-                """), {
-                'id': id,
-                'musicbrainz_id': musicbrainz_id,
-            })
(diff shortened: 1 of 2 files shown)


paramsingh:
Where is this docker hub integration coming from? Can't seem to remove it for the life of me.

alastair:
> Influx doesn't readily rename measurements. One solution that comes to mind is adding a new column with influx measurement name to the user table and using it everywhere to get listens.

I guess this was part of the reason that we moved to using user ids in the oauth login part. What if instead we stored measurements based on userid instead of username? We'd have to migrate all existing measurements, but once done it should be easy to change names.
A middle-ground could be to do exactly what you suggest here - set the measurement name to the username for all existing users, but when a new user signs up, set it to the user id.

mayhem:
> Where is this docker hub integration coming from? Can't seem to remove it for the life of me.

That was in the metabrainzdocker account -- I've disabled automated builds. I think we can delete that project since we're not using it.

paramsingh:
> A middle-ground could be to do exactly what you suggest here - set the measurement name to the username for all existing users, but when a new user signs up, set it to the user id.

Yep, agreed. This is basically what I was suggesting. I'm not sure how much work it is yet, but I'm gonna look at it. In the meanwhile, I think we should merge this so that no users lose their listens after renaming their MB accounts.

paramsingh:
@mayhem @vansika please take a look?

PR #647, opened 2019-09-24
Add script to request user stats from the spark cluster

paramsingh:
# Summary

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

* **Describe this change in 1-2 sentences**: Script that should be run in cron to request the cluster to do a job.

I've refactored the populate.py script that we used to make bigquery requests to do this.

# Future work

* Add this to cron
* Clean up the consumer script on labs side and open PR.
* Seperate the cron stuff from the web container.


mayhem:
Is this a WIP PR? The use of the BigQuery exchange is odd -- still needs to be cleaned up?

paramsingh:
This is not WIP, the bigquery exchange is used for bigquery-writer which runs in prod right now.

listenbrainz/spark/request_manage.py
```diff
@@ -0,0 +1,92 @@
...
+
+    possible_queries = _get_possible_queries()
+    if query not in possible_queries:
+        raise InvalidSparkRequestError(query)
+
+    message = {'query': possible_queries[query]['name']}
+
+    required_params = set(possible_queries[query]['params'])
+    given_params = set(params.keys())
+    if required_params != given_params:
```
mayhem: slick! :)

listenbrainz/spark/request_manage.py
```diff
@@ -0,0 +1,92 @@
...
+        raise InvalidSparkRequestError
+
+    if params:
+        message['params'] = {}
+    for key, value in params.items():
```
mayhem: I'd indent this block as well, less to do, but we'll never notice a difference in real life...

PR #657, opened 2019-10-19
Remove volume created with tests

vansika:
Every iteration of tests should build everything from scratch and destroy everything before the tests terminates. Tests should not save anything.


vansika:
If the volume is removed, then the tests should be build at every run so that we have recent changes in the codebase. I will open a ticket for that.

PR #667, opened 2019-11-12
Improve the messages of the spotify recorder

mayhem:
Improve the status messages of the reader to print how many listens were imported, but only if some listens were imported. This should improve the ability to check the reader's health.

### DIFF
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -271,6 +271,8 @@ def process_one_user(user):
     spotify.update_latest_listened_at(user.user_id, latest_listened_at)
     spotify.update_last_updated(user.user_id)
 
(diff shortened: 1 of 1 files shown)


listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -271,6 +271,8 @@ def process_one_user(user):
     spotify.update_latest_listened_at(user.user_id, latest_listened_at)
     spotify.update_last_updated(user.user_id)
 
+    current_app.logger.info('imported %d listens for %s' % (len(listens), str(user)))
```
alastair: fyi, you can do this by doing `.info("string %s %s", formatparam, formatparam2)` instead of doing a string format with `"str" % (params)`
the idea is that it'll only do the string interpolation if logging is enabled for this level, which theoretically is an optimisation, but in this case probably isn't important

PR #682, opened 2020-01-04
Fixed npm install in develop.sh

sjainit:
# Summary

* This is a…
    * ( ) Bug fix

# Problem

While setting up the environment for LB-server, the `./develop.sh npm` script didn't work for installing the node dependencies.

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -24,7 +24,7 @@ function open_psql_shell {
 }
 
 function npm_install {
-	invoke_docker_compose run --rm --user `id -u`:`id -g` -e \
+	invoke_docker_compose run --rm -e \
 				HOME=/tmp static_builder npm install
 }
 # Arguments following "manage" are as it is passed to function "invoke_manage" and executed.


develop.sh
```diff
@@ -24,7 +24,7 @@ function open_psql_shell {
 }
 
 function npm_install {
-	invoke_docker_compose run --rm --user `id -u`:`id -g` -e \
```
alastair: This is required for installing npm packages in Linux. Can you explain what error you got and why this change is needed?

sjainit: Hi @alastair. I was running it on Linux only. But I got the following error on running the `./develop.sh npm` script earlier:
[](https://gist.github.com/sjainit/f10642821ec102dd41e9a888bd51b384).

paramsingh: alastairp: it seemed to me that we were passing the incorrect user and group ids, because the bash script gets ids from the host and the command runs in the container. removing them made it work.

alastair: that's right, if you run a command in the container as root, the bind mount will cause newly created files to be owned by root, which makes it difficult to clean up your home directory. We added the --user flag to run the command in the container as the host user so that the owner/group is correct.
The only way I can see this error occurring is if you incorrectly ran npm install first without using the --user flag or develop.sh

PR #686, opened 2020-01-08
LB-179: Show error message if Last.fm import fails

ishaanshah:
# Summary

* This is a…
    * (X ) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem
Last.fm importer doesn't indicate if it receives a error from Last.fm.

* JIRA ticket: [LB-179](https://tickets.metabrainz.org/browse/LB-179)

# Solution
If a non OK response code is received from the AJAX request in the function getTotalNumberOfScrobbles(), then an error message is shown.

### DIFF
--- a/listenbrainz/webserver/templates/user/scraper.js
+++ b/listenbrainz/webserver/templates/user/scraper.js
@@ -500,12 +500,17 @@ function getTotalNumberOfScrobbles() {
     xhr.timeout = 10 * 1000; // 10 seconds
     xhr.open('GET', encodeURI(url));
     xhr.onload = function () {
-        var data = JSON.parse(this.response)['user'];
-        if ('playcount' in data) {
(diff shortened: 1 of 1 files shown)


paramsingh:
Hi @ishaanshah, thanks for helping us with this task. The changes look very sane to me. However, since this part of the codebase isn't covered by automated tests, can you list the manual tests you performed?

ishaanshah:
- Tried a valid username, in which songs were successfully imported. 
- Tried a username which doesn't exist, which returned a 404 error.
- Put an invalid Last.fm API key, which returned a 403 error.

In both of the above cases the error message was displayed.

However if the any network error isn't handled because AJAX requests are sent every 3 seconds in such a case.

PR #691, opened 2020-01-19
Merge listenbrainz-labs into listenbrainz-server

paramsingh:
# Summary

* This is a major refactor merging the listenbrainz-labs codebase (https://github.com/metabrainz/listenbrainz-labs) into the main codebase.

# Problem

It is hard to develop actual major features that need data from spark-land and need to show that data in the site proper. This is because for almost everything, there is a need to open two different pull requests, manage two different repositories. There's also a lot of code that needs to be copy pasted between the two repositories. There's also infrastructure that should be shared in the development environment such as the rabbitmq docker container that runs on the local machine.

# Solution

Merge the master branch of labs into listenbrainz-server. I've seperated most of the stuff right now. There is a listenbrainz package which contains the server code and a listenbrainz_spark package which contains the spark code for now. However, as more code is written, we'll f`ind ways to share code between the two modules which sounds pretty nice. 

The spark tests run independently from the server tests. The spark tests run in Jenkins while the server tests run on travis for now (another divergence that we should fix). For now, this is okay, but in the future we should look into moving everything to jenkins.

There are different `manage.py` and `requirements` files for both environments for now, but that should be fixed soon too.

I've tested this myself by running a local instance of listenbrainz, running the tests locally and on CI.
Once this is safely merged, I'll look into deploying the spark code to the cluster from here, and once that's done, we should safely archive the labs repository (and move all open pull requests / in development branches to here).

# Action

I think we should merge this and deploy to beta. Then look into moving open pull requests from labs to here.

*NOTE*: Ideally I should be the one to merge this, however if someone else wants to, please make sure that you do not squash or rebase the commits. Future action items involve moving existing branches and pull requests from labs to here and a rebase or squash will make things more difficult.


vansika:
Ideally there should be a develop.sh for building and running spark containers and one for building and running LB containers since the Dockerfiles and docker-compose files are different. When I open spark-develop.sh I see build instructions for LB containers. I think this should be develop.sh and we should have a separate file to build and run spark containers called spark develop.sh. 

or we can write a single develop.sh for both spark and LB and include commands to shoot correct Dockerfiles and docker-compose files. 

Thanks for reporting @MonkeyDo :)

paramsingh:
>Apparently develop.sh was renamed to spark_develop.sh

Indeed that was unintentional. I made develop.sh and spark_develop.sh both, but I must have forgotten to commit it.

PR #703, opened 2020-01-25
update develop.sh to wrap LB and spark build commands in one file.

vansika:
# Description
* This is a…
    * (x) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem

develop.sh for spark missing. 
develop.sh for LB named to spark_develop.sh

# Solution

update spark_develop.sh to build and run containers for LB and spark and rename the file to develop.sh


develop.sh
```diff
@@ -5,6 +5,12 @@ if [[ ! -d "docker" ]]; then
     exit -1
 fi
 
+function invoke_docker_compose_spark {
+	docker-compose -f docker/docker-compose.dev.yml \
```
paramsingh: Can we rename this docker-compose file to docker-compose.spark.yml

vansika: Yes, we can.

develop.sh
```diff
@@ -46,6 +52,10 @@ elif [ "$1" == "npm" ]; then
 	npm_install
 	exit
 
+elif [ "$1" == "spark" ]; then shift
+	invoke_docker_compose_spark "$@"
```
alastair: I wonder if we could use `exec` in this script and every place that we call docker-compose: https://askubuntu.com/a/819960. We should be able to add it in the `invoke_` commands
It would prevent us from having to use exit in each block, and puts control back in the hands of docker-compose

PR #704, opened 2020-01-25
Make stats rabbitmq messages smaller 

paramsingh:
# Description

* This is an improvement.

Make the rabbitmq messages being transmitted around for stats smaller, so that the pipeline doesn't error so much.

# Problem

We were sending one message as result per query to the spark cluster. This meant that we sent all stats of all users in a single json document. This document was generally gigabytes in size. This isn't something rabbitmq handles well and json parsing is slow anyways.
 

# Solution

This pull request changes the code so that we send one user's statistics in one json document.
I've tested this on the cluster.

### DIFF
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -13,10 +13,3 @@ def handle_user_artist(data):
     artist_count = data['artist_count']
     db_stats.insert_user_stats(user['id'], artists, {}, {}, artist_count)
     current_app.logger.info('Processed artist data for %s', musicbrainz_id)
-
-
-def handle_user_release(data):
-    pass #TODO: add releases
-
-def handle_user_track(data):
-    pass #TODO: add track
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -1,0 +1,22 @@
+import listenbrainz.db.user as db_user
+import listenbrainz.db.stats as db_stats
+
+from flask import current_app
+
+
+def handle_user_artist(data):
+    musicbrainz_id = data['musicbrainz_id']
+    user = db_user.get_by_mb_id(musicbrainz_id)
+    if not user:
+        return
+    artists = data['artist_stats']
+    artist_count = data['artist_count']
+    db_stats.insert_user_stats(user['id'], artists, {}, {}, artist_count)
+    current_app.logger.info('Processed artist data for %s', musicbrainz_id)
+
+
+def handle_user_release(data):
+    pass #TODO: add releases
+
+def handle_user_track(data):
+    pass #TODO: add track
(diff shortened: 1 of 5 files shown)


listenbrainz/spark/handlers.py
```diff
@@ -0,0 +1,15 @@
+import listenbrainz.db.user as db_user
```
mayhem: This file feels rather bare without a single comment in it -- maybe add a comment near the top describing the point of these handlers?

listenbrainz/spark/spark_reader.py
```diff
@@ -10,12 +10,20 @@
...
 from listenbrainz.db.exceptions import DatabaseException
+from listenbrainz.spark.handlers import handle_user_artist
 import sqlalchemy
 
 class SparkReader:
     def __init__(self):
         self.app = create_app() # creating a flask app for config values and logging to Sentry
 
+    def get_response_handler(self, response_type):
+        response_handler_map = {
```
mayhem: If wonder if we should declare the handler map at the top of the file for greater visibility?

listenbrainz_spark/request_consumer/request_consumer.py
```diff
@@ -58,34 +58,35 @@ def get_result(self, request):
...
-                self.result_channel = self.rabbitmq.channel()
-                self.result_channel.exchange_declare(exchange=current_app.config['SPARK_RESULT_EXCHANGE'], exchange_type='fanout')
+    def push_to_result_queue(self, messages):
+        for message in messages:
+            while True:
+                try:
+                    self.result_channel.basic_publish(
+                        exchange=current_app.config['SPARK_RESULT_EXCHANGE'],
+                        routing_key='',
+                        body=json.dumps(message),
```
mayhem: Hmm. now that we merged in the -labs repo we're using json and ujson -- we should make an effort to replace all json with ujson...

listenbrainz_spark/request_consumer/request_consumer.py
```diff
@@ -58,34 +58,35 @@ def get_result(self, request):
...
+                except pika.exceptions.ConnectionClosed:
+                    self.connect_to_rabbitmq()
+                    time.sleep(1)
+                except pika.exceptions.ChannelClosed:
+                    self.result_channel = self.rabbitmq.channel()
+                    self.result_channel.exchange_declare(exchange=current_app.config['SPARK_RESULT_EXCHANGE'], exchange_type='fanout')
 
 
 
     def callback(self, channel, method, properties, body):
```
mayhem: The messages in this function seem a bit verbose to me. These are pretty good for debugging, but for production it yields very chatty logs. Overall for production, I wonder if we should aim to have the logs be silent, or near silent, except for when errors happen?

listenbrainz_spark/request_consumer/request_consumer.py
paramsingh: This is a bit more complicated and we should take a look at dependencies a bit more holistically. I'm not gonna do it here, but open a ticket.

listenbrainz/spark/handlers.py
paramsingh: Added a docstring for the file and for the handler function

PR #707, opened 2020-01-26
Fix build and run spark containers

shivam-kapila:
#11 Description
Fix Spark containers not being build and run in develop.sh (In continuation to PR #703 ).

* This is a…
    * (x) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem
Even after combining the code to create Spark containers in PR #703 the function `invoke_docker_compose_spark` was not being called. Hence the Spark containers were not run using develop.sh.

# Solution

* Running both `docker-compose.yml` and `docker-compose.spark.yml` using a single `docker-compose` command
* Remove the `network` configured in `docker-compose.spark.yml` as now `listenbrainz-default` is in the same network.

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -13,6 +13,7 @@ function invoke_docker_compose_spark {
 
 function invoke_docker_compose {
 	docker-compose -f docker/docker-compose.yml \
+	 			-f docker/docker-compose.spark.yml \
 				-p listenbrainz \
 				"$@"
 }
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,6 @@ ENV/
 
 config.py
 docker/tunnel/keys
+
+# temporary files
+tmp/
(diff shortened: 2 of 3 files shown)


paramsingh:
@brainzbot add to whitelist

develop.sh
```diff
@@ -13,6 +13,7 @@ function invoke_docker_compose_spark {
 
 function invoke_docker_compose {
 	docker-compose -f docker/docker-compose.yml \
+	 			-f docker/docker-compose.spark.yml \
```
paramsingh: I'm not sure what specifying two `-f` params does exactly. What happens if we run `./develop.sh up`? Does it bring up both the spark containers and the webserver containers?

shivam-kapila: Mentioning multiple `-f` combines the `.yml` files in the order they are mentioned and then brings them up. Yes `./develop.sh up` brings both spark and webserver containers up

PR #708, opened 2020-01-26
LB-484: Stream response from /profile/export endpoint

ruuda:
# Description

Do not materialize all listens in memory when exporting them, stream them instead.

* This is a…
    * (x) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * ( ) Other

* **Describe this change in 1-2 sentences**: Previously, the export endpoint would fetch all listens from the listenstore into a list in memory, then build a second list in memory, then serialize that as json, and only then the response could be served. This pull request replaces the lists with generators, so serving the response and fetching from the database can be interleaved, for faster time-to-first-byte and lower memory consumption.

# Problem

* JIRA ticket: [LB-484](https://tickets.metabrainz.org/browse/LB-484)

When I try to export the listens for my profile, I consistently get a gateway timeout after 60 seconds. I suspect that this is because fetching my almost 150k listens into a list in memory takes more than 60 seconds, and the response cannot be served until all listens have been fetched.

Building that list also consumes memory proportional to the amount of listens, but I don't know if that has been a problem for you.

# Solution

Rather than fetching all listens into memory first, we can stream them. If we move the loops into a function and replace `append` with `yield`, then we get a generator that produces the listens on demand. Instead of building a second list of result dicts, we can map a function over the listens to produce the dicts. And instead of building a list and serializing the entire thing in one go, add a generator that streams a json array. Flask supports generators that produce the response body. This way, the response can be served immediately after fetching the first batch of listens.

I hope that starting the response earlier will avoid the gateway timeout, but I have not been able to verify this.

# Action

I have not been able to set up a local development environment for this yet, but Travis does run the test suite, and there was already a test covering export, which still passes. I also added an extra test in `test_profile.py` that tests the structure of the response and calls to `fetch_listens`.

I tried to keep the style consistent with the code that was already there, but please let me know if anything can be improved.

### DIFF
--- a/listenbrainz/webserver/views/profile.py
+++ b/listenbrainz/webserver/views/profile.py
@@ -9,7 +9,8 @@ import zipfile
 
 
 from datetime import datetime
-from flask import Blueprint, render_template, request, url_for, redirect, current_app, make_response, jsonify
+from flask import Blueprint, Response, render_template, request, url_for, \
+    redirect, current_app, make_response, jsonify, stream_with_context
 from flask_login import current_user, login_required
 import spotipy.oauth2
 from werkzeug.exceptions import NotFound, BadRequest, RequestEntityTooLarge, InternalServerError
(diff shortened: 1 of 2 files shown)


paramsingh:
@brainzbot add to whitelist

listenbrainz/webserver/views/profile.py
```diff
@@ -123,6 +124,39 @@ def import_data():
...
+    db_conn = webserver.create_influx(current_app)
+    while True:
+        batch = db_conn.fetch_listens(current_user.musicbrainz_id, to_ts=to_ts, limit=EXPORT_FETCH_COUNT)
+        if not batch:
+            break
+        yield from batch
+        to_ts = batch[-1].ts_since_epoch  # new to_ts will be the the timestamp of the last listen fetched
+
+
+def map_listen_for_export(obj):
```
paramsingh: we have a function for this (https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/listen.py#L186) that might be helpful.

ruuda: Thanks for taking a look! `map_listen_for_export` here is what used to be [this loop body](https://github.com/metabrainz/listenbrainz-server/pull/708/files#diff-5c1b6caf808da8edd5873fe0fc2e5d22L147-L152). It looks like `to_api` returns the data in a slightly different format. If we use it here, that would change the schema of `/profile/export`. Is that desired?

paramsingh: I think so, yes! We're hoping to consolidate the data format. The old version seems to be from some old code that never got changed.

ruuda: Okay, I changed it to use `to_api` and I updated the test to match the new format.

paramsingh:
I've deployed this to https://beta.listenbrainz.org if you want to check that the gateway timeout is fixed now?

ruuda:
It works, I was able to export my listens from beta.listenbrainz.org. Thanks a lot for picking this up so quickly!

paramsingh:
Awesome, thanks for contributing! If you have any more feedback or feature requests, feel free to hop on over to our irc channel on freenode: #metabrainz :)

PR #709, opened 2020-01-26
fix sentry config for spark

paramsingh:
# Description

* This is a bugfix

* **Describe this change in 1-2 sentences**: The config file didn't have a variable for the public DSN.

# Problem

The Sentry config in lsitenbrainz_spark didn't have a variable for the public DSN.

# Solution

I've added a seperate DSN variable as is required by brainzutils. Ideally it should be a part of the dict, but that would require changing the brainzutils library which I don't think is worth the effort right now.


paramsingh:
I've deployed the request consumer on leader with the ListenBrainz beta sentry config, so we should start seeing errors come in from the spark side too now.

PR #712, opened 2020-01-28
format namenode and datanode together

vansika:
# Description
* This is a…
    * ( ) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * (x ) Other

# Problem
datnode was not starting up when I tried running the containers after formatting namenode .

# Solution
if namenode is formatted without formatting datadnode, the datanode keeps referencing with the previous namenode cluster id. 
Formatting namenode and datanode is a good way out.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -44,7 +44,7 @@ To run the integration tests:
 - ##### Format namenode
     Before formatting namenode, datanode must be formatted. To format namenode and datanode:
 
-        docker-compose -f docker/docker-compose.spark.yml -p listenbrainzspark run --rm hadoop-master hdfs namenode -format -clusterId <datanode cluster ID>
+        ./spark_develop.sh format <datanode cluster ID>
 
     To open datanode command prompt:
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -39,3 +39,17 @@ To run the unit tests:
 To run the integration tests:
 
     ./integration-test.sh
+
+### Spark
+- ##### Format namenode
+    Before formatting namenode, datanode must be formatted. To format namenode and datanode:
+
+        docker-compose -f docker/docker-compose.spark.yml -p listenbrainzspark run --rm hadoop-master hdfs namenode -format -clusterId <datanode cluster ID>
+
+    To open datanode command prompt:
+
+        docker exec -it listenbrainzspark_datanode_1 bash
+
+    To get the datanode cluster id:
+
+        cat /home/hadoop/hdfs/datanode/current/VERSION | grep "clusterID"
--- a/HACKING.md
+++ b/HACKING.md
@@ -41,6 +41,12 @@ To run the integration tests:
     ./integration-test.sh
 
 ### Spark
+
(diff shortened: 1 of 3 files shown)


HACKING.md
```diff
@@ -39,3 +39,17 @@ To run the unit tests:
 To run the integration tests:
 
     ./integration-test.sh
+
+### Spark
+- ##### Format namenode
+    Before formatting namenode, datanode must be formatted. To format namenode and datanode:
+
+        docker-compose -f docker/docker-compose.spark.yml -p listenbrainzspark run --rm hadoop-master hdfs namenode -format -clusterId <datanode cluster ID>
```
paramsingh: Can we add thiis command to `spark_develop.sh`? Like we have different commands for develop.sh.

HACKING.md
```diff
@@ -39,3 +39,17 @@ To run the unit tests:
 To run the integration tests:
 
     ./integration-test.sh
+
+### Spark
```
mayhem: This comment doesn't really pertain to your work on this PR, but reading the file over, I feel that there is some general introductory stuff missing. We talk about the various bits of technology we use in this file, but never why or how they fit together. Which up until now was fine, but now that we have spark, a new developer might feel that they need to setup spark in order to setup LB, which isn't the case. Adding more verbiage describing what these bits are used for and why most people won't need to setup spark will make this document much much better.

vansika: I was assuming that Readme will contain all the information about LB and spark, about how they are related, and how they are not and yes, why it is not necessary to setup both. HACKING.md is more like a file for quick reference to know the correct commands. I am not sure how good of a plan it is. Otherwise HACKING.md can be descriptive. It is just that Readme is first or may be most obvious place to look for detailed setup info. What do you think?

mayhem: Yes, I see that, but now that we've merged the spark portions into this repo, I think we ought to at least put a note into HACKING that lets people know that not everyone needs to setup spark.

And if we want to do a bit of a tech overview in README, that would be fine too -- but right now neither of them really explain how spark fits into LB and that will generate confusion.

vansika: Sounds good. We can refactor both the files later to make them fit together well. Thanks :)

develop.sh
```diff
@@ -5,63 +5,53 @@ if [[ ! -d "docker" ]]; then
...
-	docker-compose -f docker/docker-compose.spark.yml \
-				-p listenbrainzspark \
-				"$@"
-}
-
 function invoke_docker_compose {
-	docker-compose -f docker/docker-compose.yml \
-				-p listenbrainz \
-				"$@"
+    docker-compose -f docker/docker-compose.yml \
```
paramsingh: It'd be great if we could remove the whitespace changes in this file.

HACKING.md
vansika: I missed it before but for labs we need to setup LB once. 
https://github.com/metabrainz/listenbrainz-server/blob/master/docker/docker-compose.spark.yml#L47

develop.sh
vansika: I think indentation looks better now. Not sure if there is a standard we follow.

PR #720, opened 2020-02-09
Add clarification about Last.fm privacy settings to import page.

agates:
# Description

Adds a clarification to the import page notifying the user that import will not work without disabling the "Hide recent listening information" privacy setting.

* This is a…
    * ( ) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem

Last.fm import doesn't work when the user has "Hide recent listening information" enabled in Last.fm.  The import page doesn't specify this.

# Solution

Specify to disable "Hide recent listening information" in Last.fm settings.

### DIFF
--- a/listenbrainz/webserver/templates/user/import.html
+++ b/listenbrainz/webserver/templates/user/import.html
@@ -17,10 +17,13 @@
         process. Running the import process multiple times <strong>does not</strong> create duplicates in your
         ListenBrainz listen history.
       </p>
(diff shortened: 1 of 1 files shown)


paramsingh:
@brainzbot add to whitelist

mayhem:
The failed test is unrelated to this, so I'm happy to merge.

PR #726, opened 2020-02-10
Add "delete listens" button

shivam-kapila:
# Description
Add "Delete your listens" button to delete listens history for a user 

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem
There didn't exist any button to access the `delete` function present in _influx_listenstore.py_.

# Solution

* Added a button to redirect to `/delete-listens` route.
* If GET request it renders a page asking user for confirmation. If POST request it deletes the listens history for the logged-in user. 
(Similar to delete user feature.)


shivam-kapila:
@sjainit I have added the front-end part. Kindly add the `delete_listens_history` function in _user.py_ to call the delete function existing in _influx_listenstore.py_.

sjainit:
Hey @shivam-kapila. The build isn't passing. Can you please check?

listenbrainz/webserver/templates/profile/deletelistens.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'base.html' %}
```
paramsingh: I'd prefer to name this `delete_listens.html` for consistency.

listenbrainz/webserver/views/user.py
```diff
@@ -231,7 +231,7 @@ def get_track_id_from_listen(listen):
 
 def delete_user(musicbrainz_id):
     """ Delete a user from ListenBrainz completely.
-    First, drops the user's influx measurement and then deletes her from the
+    First, drops the user's influx measurement and then deletes him/her from the
```
paramsingh: given that you're fixing this anyways, we've decided to avoid gendered pronouns. Let's fix this to be `them` or `the user`.

listenbrainz/webserver/views/user.py
```diff
@@ -253,3 +253,18 @@ def delete_user(musicbrainz_id):
         error_msg='Could not put user %s into queue for deletion, please try again later' % musicbrainz_id,
     )
     db_user.delete(user.id)
+
+def delete_listens_history(musicbrainz_id):
+    """ Delete a user's listens from ListenBrainz completely.
+    This, drops the user's influx measurement and resets his/her listen count.
```
paramsingh: same for gendered pronoun here.

listenbrainz/webserver/views/test/test_profile.py
```diff
@@ -94,6 +94,15 @@ def test_delete(self, mock_publish_data_to_queue):
...
         self.assertIsNone(user)
 
+    
+    def test_delete_listens(self):
+        self.temporary_login(self.user['login_id'])
+        r = self.client.get(url_for('profile.delete_listens'))
+        self.assert200(r)
+
+        r = self.client.post(url_for('profile.delete_listens'), data={'token': self.user['auth_token']})
+        self.assertRedirects(r, url_for('user.profile', user_name=self.user['musicbrainz_id']))
```
paramsingh: So we should also test that the listens have actually gotten deleted. 

Let's add an integration test which does the following:
1. submits a bunch of listens
2. checks that the listens have been written
2. makes a post request to this endpoint
3. checks that the listens have been deleted.

listenbrainz/webserver/views/test/test_profile.py
```diff
@@ -94,6 +94,15 @@ def test_delete(self, mock_publish_data_to_queue):
         user = db_user.get(self.user['id'])
         self.assertIsNone(user)
 
+    
+    def test_delete_listens(self):
+        self.temporary_login(self.user['login_id'])
```
paramsingh: let's test that the page redirects to login if the user isn't logged in also.

listenbrainz/webserver/views/test/test_profile.py
```diff
@@ -94,6 +94,15 @@ def test_delete(self, mock_publish_data_to_queue):
...
+        r = self.client.get(url_for('profile.delete_listens'))
+        self.assert200(r)
+
+        r = self.client.post(url_for('profile.delete_listens'), data={'token': self.user['auth_token']})
```
paramsingh: also, let's add a test that makes sure that the endpoint returns a 4xx if the auth token isn't provided or if the auth token is incorrect.

listenbrainz/webserver/views/profile.py
```diff
@@ -241,6 +241,36 @@ def delete():
...
+            try:
+                delete_listens_history(current_user.musicbrainz_id)
+            except Exception as e:
+                current_app.logger.error('Error while deleting listens for %s: %s', current_user.musicbrainz_id, str(e))
+                flash.error('Error while deleting listens for %s, please try again later.' % current_user.musicbrainz_id)
+                return redirect(url_for('profile.info'))
+            flash.info('Successfully deleted listens for %s.' % current_user.musicbrainz_id)
+            return redirect(url_for('user.profile', user_name=current_user.musicbrainz_id))
+        else:
+            flash.error('Cannot delete the user\'s listens due to error during authentication, please try again later.')
```
paramsingh: This message is shown to the user. Let's change it to something more user friendly.

"We couldn't delete your listens due to an error. Please try again later."

listenbrainz/webserver/views/test/test_profile.py
shivam-kapila: Is it ok if I test against 302 because in case auth token is missing or invalid the user is redirected to profile page. If I explicitly specify 4xx then the page is not automatically redirected.

paramsingh: The behaviour seems incorrect then. The code should return we should `raise Unauthorized` if the token is not valid.

Let's fix the actual code so that the test which checks for a 401 Unauthorized status passes

shivam-kapila: Did exactly the same.

shivam-kapila: I have added the unauthorized check. The test for invalid token passed but if I don't pass any token or pass None to it then it gets status 200 in tests but in actual if I delete token it raises unauthorized. Can you please look at th [test code](https://gist.github.com/shivam-kapila/d14dbde481c0208bd95bd87c61eabe7e).

The modified actual code [here](https://gist.github.com/shivam-kapila/246b8b2ac3722a04ee7ea45a8308242a#file-delete_listens-modified-code-L14)

paramsingh: In the test, the request response is stored in `r` and the assertion is on `response`

listenbrainz/webserver/views/test/test_profile.py
shivam-kapila: I have added the test.

listenbrainz/webserver/views/test/test_profile.py
shivam-kapila: Thanks a lot. I have pushed all the tests.

listenbrainz/webserver/views/profile.py
shivam-kapila: I have updated it with ### Auth token invalid or missing.

listenbrainz/webserver/views/test/test_profile.py
shivam-kapila: The integration tests have been added.

shivam-kapila:
@paramsingh This is the gif for how the functionality works. 
delete-listens

listenbrainz/webserver/templates/profile/info.html
```diff
@@ -12,6 +12,15 @@ <h3> Delete your account </h3>
       <a href="{{ url_for("profile.delete") }}" style="color: white;"><span class="btn btn-info btn-lg" style="width: 300px">Delete your account</span></a>
     </p>
 
+    <h3> Delete your listens </h3>
+    <p>
+      This will delete all your listens imported into Listenbrainz.
```
mayhem: I think this text makes it sound like if you press the button your listens are gone, when in fact you'll be taken to another page. I personally don't think we need the extra page just yet -- all the other actions can be taken directly from the profile page. I think for now it would be better to add the delete listens function to the profile page.

listenbrainz/webserver/views/user.py
```diff
@@ -253,3 +253,18 @@ def delete_user(musicbrainz_id):
...
+    This, drops the user's influx measurement and resets their listen count. 
+
+    Args:
+        musicbrainz_id (str): the MusicBrainz ID of the user
+
+    Raises:
+        NotFound if user isn't present in the database
+    """
+
+    user = _get_user(musicbrainz_id)
```
mayhem: You need to catch and handle exceptions from here.

shivam-kapila: @mayhem Thanks for the review. Most of the exceptions are handled in the `delete` function in _listenstore.py_ [here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/listenstore/influx_listenstore.py#L916). Can you give me an idea which exceptions can I handle in the function I created.

listenbrainz/webserver/templates/profile/info.html
shivam-kapila: Okay.

listenbrainz/webserver/views/user.py
mayhem: Sorry, never mind this point.

listenbrainz/webserver/templates/profile/info.html
shivam-kapila: @mayhem Done :)

shivam-kapila:
@mayhem Thanks for the review. I have added the confirmation page back. Kindly have a look when free :).

This is the demo of the latest changes:

delete-listens_new_with_confirmation

PR #728, opened 2020-02-13
Move set_rate_limits.py

mayhem:
# Description
set rate limiits is in the wrong dir for it to run...

* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

# Problem

* JIRA ticket (_optional_): 

# Solution

# Action

### DIFF
--- a/set_rate_limits.py
+++ b/set_rate_limits.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+
+import sys
+from redis import Redis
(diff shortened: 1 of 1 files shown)


mayhem:
Note -- this just moves a file -- it changes nothing else.

paramsingh:
Where was this file earlier?

mayhem:
in listenbrainz-server/listenbrainz , so it could not import python modules.

paramsingh:
Should we remove the old file?

paramsingh:
Although, technically it should work from the dir it was in anyways. Not sure why it doesn't.

mayhem:
Something about this PR turned out wrong. I'm going to toss it and just make the change against master.

PR #729, opened 2020-02-13
create dumps for candidate recordings 

vansika:
# Description
* This is a…
    * ( ) Bug fix
    * (x) Feature addition
    * ( ) Refactoring
    * ( ) Other

This PR aims at creating candidate recording dumps i.e. dumps of recommendations.

# Problem
It is a draft PR since the script is failing for large number of users. It did not come to me that candidate_sets.py would fail since spark works well in iterative jobs. The script is failing because of GC overhead limit exceeded. I tried to generate candidate sets for around 1900 users (on a months data) . In my opinion, it is failing because top artists and similar artists are being generated for every user individually. 
For a small number of users (~500)  it works well though. 

Also, the union function may be one of the reasons for the errors since with each union operation a new dataframe is created, old data is copied to it and fresh data is appended. 

# Solution
I think it will be good to make a datafame of the top artist and similar artists for all the users in one go, just like how we do in stats.


listenbrainz_spark/recommendations/recommend.py
```diff
@@ -33,6 +41,149 @@ def load_model(path):
...
+def dump_candidate_recordings(recommendations, temp_file, playlist_type):
+    """ Write candidate recordings to a file based on type of playlist.
+
+        Args:
+            recommendations (dict): Recommended recordings for users.
+            temp_file (str) : File path to dump recordings.
+            playlist_type (str): Type of playlist.
+    """
+
+    with open(temp_file, 'w') as f:
```
mayhem: One thing that I really like doing when generating dumps is to have python compress the data as it is being output. It makes the process a bit slower, but overall it saves time from having to compress the output in a second step. See https://docs.python.org/3/library/lzma.html

mayhem: Though my point may be moot, since you do use the very intelligent multithreaded pxz compression below. Perhaps this comment really isn't actionable. 😇

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -267,5 +421,8 @@ def main():
     # persisted data must be cleared from memory after usage to avoid OOM
     recordings_df.unpersist()
 
+    if create_dump:
+        create_candidate_recordings_dump(recommendations, threads)
```
mayhem: This lacks top-level exception handling

paramsingh:
Closing for inactivity, feel free to reopen when needed.

PR #732, opened 2020-02-19
Upgrade influxdb

mayhem:
# Description

We need to upgrade to the latest influxdb before we start performance tuning on our currently lagging setup.

* This is a…
    * ( ) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * (x) Other

# Problem

Our DB is old.

# Solution

Upgrade it!

# Action


mayhem:
Against master? FFS!

PR #733, opened 2020-02-19
Bump version numbers

mayhem:
# Description

We need to upgrade to the latest influxdb before we start performance tuning on our currently lagging setup.

* This is a…
    * ( ) Bug fix
    * ( ) Feature addition
    * ( ) Refactoring
    * (x) Other

# Problem

Our install is outdated and needs upgrading


mayhem:
As per our conversation online, the spark related builds failing is expected.

mayhem:
I've added some finer grained timing debug info to the writer -- its kinda ugly and should get reverted once we do the next release, but for now this should do.

PR #734, opened 2020-02-19
Add cron job requesting stats from the spark cluster

paramsingh:
# Description

Add a cronjob asking for stats every saturday, I'd like to keep it every saturday for now and monitor. When it's stable we can change the time as we want.

# Problem

User stats do not get automatically updated.

# Solution

A cron job 

# Action

Merge and deploy.


paramsingh:
I agree. I'll change it to every day for now.

PR #736, opened 2020-02-19
Send a notification email when new data dumps are created

paramsingh:
# Description

This should help us monitor that data dumps get created correctly.

# Problem

Data dumps were not being created for a while and it took us much longer than expected to actually discover the regression.

# Solution

This pull request adds code that sends an email when the dump has been created with a link to ftp. 

# Action

I think we might want to create a `listenbrainz-observability@metabrainz.org` or similar mailer that we can individually subscribe to if interested. Right now, I've made the code send an email to me, but the mailing list seems the way forward.

### DIFF
--- a/listenbrainz/db/dump_manager.py
+++ b/listenbrainz/db/dump_manager.py
@@ -29,6 +29,7 @@ import shutil
 import subprocess
 import sys
 
+from brainzutils.mail import send_mail
 from datetime import datetime
 from flask import current_app
 from influxdb.exceptions import InfluxDBClientError, InfluxDBServerError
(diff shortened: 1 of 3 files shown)


listenbrainz/db/dump_manager.py
```diff
@@ -44,6 +45,17 @@
 
 cli = click.Group()
 
+def send_dump_creation_notification(dump_name, dump_type):
+    if not current_app.config['TESTING']:
+        dump_link = 'http://ftp.musicbrainz.org/pub/musicbrainz/listenbrainz/{}/{}'.format(dump_type, dump_name)
+        send_mail(
+            subject="ListenBrainz dump created - {}".format(dump_name),
+            text=render_template('emails/data_dump_created_notification.txt', dump_name=dump_name, dump_link=dump_link),
+            recipients=['param@metabrainz.org'],
```
paramsingh: I think this recipient should be a mailing list. `listenbrainz-observability@metabrainz.org`?

paramsingh:
@brainzbot retest this please

paramsingh:
@brainzbot retest this please

PR #738, opened 2020-02-20
LB-494: Make Last.fm importer modal responsive

shivam-kapila:
# Description

Make Last.fm importer modal responsive and add redirect link to user profile in it

# Problem

The last.fm importer modal is not currently responsive after it was ported to react. Also, the link that redirects the user to his/her profile after importing the listens from Last.fm is missing.

# Solution

- Make the  Last.fm importer modal responsive
- Add redirect to user profile in the modal


shivam-kapila:
Sure

shivam-kapila:
These are the screenshots in desktop and mobile view.
Modal_Desktop_View
Modal_Mobile_View

PR #739, opened 2020-02-20
Write tests for the LastFM Importer

ishaanshah:
# Description
Write tests for LastFM Importer using Jest and Enzyme.

# Problem
Currently the JavaScript part of the code base isn't covered in the test suite, which means one has to manually test the code if any changes are made to the code base.

# Solution
Setting up a test suite for React and JavaScript can help in writing code faster and with more confidence

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,13 +30,14 @@ install:
 
 script:
   # First run unit tests and bring containers down then
-  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test up -d db influx redis
+  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test up -d db influx redis static_tester
   - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm listenbrainz
                 dockerize
                 -wait tcp://db:5432 -timeout 60s
                 -wait tcp://influx:8086 -timeout 60s
                 -wait tcp://redis:6379 -timeout 60s
-                py.test
+                "py.test"
+  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm static_tester npm test
   - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test down
 
   # Bring up integration test containers, run tests and bring them down
@@ -50,16 +51,3 @@ script:
                                      -wait tcp://rabbitmq:5672 -timeout 60s
                                      bash -c "py.test listenbrainz/tests/integration"
   - docker-compose -f docker/docker-compose.integration.yml -p listenbrainz_int down
-
-language: node_js
-node_js:
-- lts/*
-cache:
-  directories:
-    - node_modules
-before_install:
-- npm update
-install:
-- npm run build:dev
-script:
-- npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,5 +51,5 @@ script:
                                      bash -c "py.test listenbrainz/tests/integration"
   - docker-compose -f docker/docker-compose.integration.yml -p listenbrainz_int down
 
-  - echo "JavaScripts tests start here"
+  - echo "JavaScript tests start here"
   - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm static_tester npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,14 +30,13 @@ install:
 
 script:
   # First run unit tests and bring containers down then
-  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test up -d db influx redis static_tester
+  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test up -d db influx redis
   - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm listenbrainz
                 dockerize
                 -wait tcp://db:5432 -timeout 60s
                 -wait tcp://influx:8086 -timeout 60s
                 -wait tcp://redis:6379 -timeout 60s
                 "py.test"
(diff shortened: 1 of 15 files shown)


.travis.yml
```diff
@@ -30,13 +30,14 @@ install:
 
 script:
   # First run unit tests and bring containers down then
-  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test up -d db influx redis
+  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test up -d db influx redis static_tester
```
paramsingh: do we need to bring the container up here? Seems like the run command down below should just work.

listenbrainz/webserver/static/js/__tests__/lastFmImporter.test.js
```diff
@@ -0,0 +1,15 @@
...
+import ReactDOM from 'react-dom'
+
+import LastFmImporter from '../jsx/lastFmImporter'
+
+let props;
+
+describe('LastFmImporter is working correctly', () => {
+  it('renders without crashing', () => {
+    const div = document.createElement('div');
+    ReactDOM.render(<LastFmImporter {...props} />, div);
```
paramsingh: why are we using ReactDom here? 

Ideally I'd like to use enzyme to test the components, something liek mounting the component and checking that it is truthy. https://airbnb.io/enzyme/docs/api/mount.html

```
it ('renders', () => {
    const wrapper = mount(<LastFmImporter {...props />);
    expect(wrapper).toBeTruthy();
})
```

Something around these lines.

test.sh
```diff
@@ -77,7 +77,13 @@ function dcdown {
     # Shutting down all containers associated with this project
     docker-compose -f $COMPOSE_FILE_LOC \
                    -p $COMPOSE_PROJECT_NAME \
-               down
+                    down 
+}
+
+function static_tester_up {
```
paramsingh: while I'm happy that the same travis build runs the frontend tests, I'd prefer to have a different `frontend-test.sh` to test the frontend stuff. (or a way to specify that we only want to run the frontend tests in `test.sh`, maybe something like ./`test.sh --js-only`). I'll let you choose whatever seems more convenient to you.

This gives us the benefit of not having to set up the database and other random services, just to run the js tests which will never touch those services.

docker/Dockerfile.jest
```diff
@@ -0,0 +1,7 @@
+FROM node:10.15-alpine
```
paramsingh: Do we need this dockerfile? What's the difference between this and [Dockerfile.webpack](https://github.com/metabrainz/listenbrainz-server/blob/52705348/docker/Dockerfile.webpack#L1) and could we just use the same dockerfile for both running tests and compiling the jsx?

docker/docker-compose.test.yml
```diff
@@ -43,3 +43,10 @@ services:
       - db
       - influx
       - rabbitmq
+ 
+  static_tester:
```
paramsingh: Would prefer it if this was named `frontend_tester`.

.travis.yml
```diff
@@ -30,13 +30,14 @@ install:
...
   - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm listenbrainz
                 dockerize
                 -wait tcp://db:5432 -timeout 60s
                 -wait tcp://influx:8086 -timeout 60s
                 -wait tcp://redis:6379 -timeout 60s
-                py.test
+                "py.test"
+  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm static_tester npm test
```
paramsingh: I would prefer it if the frontend tests were run either after all the python tests were done and containers brought down OR before any containers for the python tests are started. Right now it sits between the python unit tests and integration tests which is confusing when reading the logs.

jest.config.js
```diff
@@ -0,0 +1,37 @@
...
+    moduleFileExtensions: ['js', 'json', 'jsx'],
+  
+    // The paths to modules that run some code to configure or set up the testing environment before each test
+    setupFiles: ['<rootDir>/enzyme.config.js'],
+  
+    // The test environment that will be used for testing
+    testEnvironment: 'jsdom',
+  
+    // The glob patterns Jest uses to detect test files
+    testMatch: ['<rootDir>/static/**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+(spec|test).js?(x)'],
```
paramsingh: I think it might be better to not have the `__tests__` directory here. Instead let's just keep the `.test.js` files next to the normal `js` files.

listenbrainz/webserver/static/js/__tests__/lastFmImporter.test.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
+import LastFmImporter from '../jsx/lastFmImporter'
+
+let props;
```
paramsingh: If we're not passing any props, let's just remove this variable entirely.

test.sh
```diff
@@ -126,6 +126,7 @@ if [ $DB_EXISTS -eq 1 -a $DB_RUNNING -eq 1 ]; then
     dcdown
 else
     # Else, we have containers, just run tests
+    static_tester_up
```
paramsingh: did you miss removeing this?

frontend-test.sh
```diff
@@ -0,0 +1,12 @@
+COMPOSE_FILE_LOC=docker/docker-compose.test.yml
```
paramsingh:
```suggestion
#!/bin/bash

COMPOSE_FILE_LOC=docker/docker-compose.test.yml
```

having a shebang for scripts is generally a good idea: https://en.wikipedia.org/wiki/Shebang_(Unix)

PR #740, opened 2020-02-20
Send a mail when new user stats start getting written into the db from spark

paramsingh:
# Description

This PR sends an email when new user statistics start coming in. If the user stats table was last updated 12 hours ago, we send a mail.

# Problem

Observability on listenbrainz stats generation is non-existent.

# Solution

Start sending emails so that we know that stuff is up and working.

### DIFF
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -10,7 +10,13 @@ from datetime import datetime, timezone, timedelta
 
 TIME_TO_CONSIDER_STATS_AS_OLD = 12 # hours
 
-def new_user_stats():
+def is_new_user_stats_batch():
+    """ Returns True if this batch of user stats is new, False otherwise
+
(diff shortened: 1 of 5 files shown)


listenbrainz/spark/handlers.py
```diff
@@ -4,7 +4,25 @@
...
 import listenbrainz.db.stats as db_stats
 
-from flask import current_app
+from flask import current_app, render_template
+from brainzutils.mail import send_mail
+from datetime import datetime, timezone, timedelta
+
+TIME_TO_CONSIDER_STATS_AS_OLD = 12 # hours
+
+def new_user_stats():
```
mayhem: this function name doesn't exactly make it clear what is going on. Stats for a new user? Or new stats about users?

paramsingh:
@brainzbot retest this please

PR #742, opened 2020-02-21
Fix all.py path bug

paramsingh:
# Description

Fix the regression in stats calculation.

# Problem

There was a bug in our user stats calculation. Essentially we were passing the HDFS_CLUSTER_URI to the path twice.

# Solution

Remove the extra path param added and add a test.

# Action

Merge and deploy.

### DIFF
--- a/listenbrainz_spark/stats/user/tests/test_all.py
+++ b/listenbrainz_spark/stats/user/tests/test_all.py
@@ -1,0 +1,55 @@
+from pyspark.sql import Row
+
+from listenbrainz_spark.tests import SparkTestCase
+from listenbrainz_spark.stats.user.all import calculate
+from listenbrainz_spark import utils
+
+class UserStatsAllTestCase(SparkTestCase):
+
+    def save_dataframe(self):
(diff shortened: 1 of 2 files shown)


listenbrainz_spark/stats/user/tests/test_all.py
```diff
@@ -0,0 +1,55 @@
...
+
+class UserStatsAllTestCase(SparkTestCase):
+
+    def save_dataframe(self):
+        df = utils.create_dataframe(Row(user_name='user2', artist_name='artist1', artist_msid='1',artist_mbids='1',
+            track_name='test', recording_msid='1', recording_mbid='1', release_name='test',release_msid='1',
+            release_mbid='1'), schema=None)
+        df1 = utils.create_dataframe(Row(user_name='user1',artist_name='artist1', artist_msid='1',artist_mbids='1',
+            track_name='test', recording_msid='1', recording_mbid='1', release_name='test',release_msid='1',
+             release_mbid='1'), schema=None)
```
mayhem: indent? 😆

PR #743, opened 2020-02-22
Download the latest dump from FTP instead of using config.

paramsingh:
# Description

For some reason, the code for our FTP stuff used a config variable to decide which dump to download. This PR removes that and downloads the latest dump instead.

# Problem

Using config variables to decide which dump to download makes it hard to automate this flow.

# Solution

Remove the config variable, sort the output of ftp ls by dump ID and then download the dump with the greatest ID. 

Also, this pull request skips a flaky test that has been a pain point for the last week. I've added a TODO to fix it, but I don't have context on what it's doing actually so it's probably gonna be easier if @vansika looks at it.


vansika:
config variables to decide which dump to download are meant for users who need a small dump ( Incremental till we have some test dump). Users can update the var with the dump path, not having to download big dumps worth GBs.

paramsingh:
>config variables to decide which dump to download are meant for users who need a small dump ( Incremental till we have some test dump). Users can update the var with the dump path, not having to download big dumps worth GBs.

That should be done with a different command that downloads and imports the latest incremental dump instead of a config variable. The variable was confusing because I'm not sure whether it is supposed to be an absolute path to a file or something else altogether.

vansika:
Right! It was just FYI. I did not automate the listens downloading process because when I last checked FTP, there was a folder for latest dump with no tarfile. It was a temporary sol. Thanks for the PR.

vansika:
Having said that, it will great if you could add a command/function for incremental dump with lil comments/docs indicating it for users who want small dumps or dumps preferable for local environment

paramsingh:
@vansika ptal

paramsingh:
I'm merging this in interest of moving forward :ship:. Feel free to review more as needed, I'll make the  changes in future PRs.

PR #744, opened 2020-02-23
Automate data dump imports into the spark cluster

paramsingh:
# Description

Create a spark job that downloads and imports a full dump from FTP. Mostly based off of @vansika's code. 

I also added a manage command on the server side which we can use to make a request for new dump import via the same infra we use to request stats.

# Problem

 

There's no good way to update the data in the spark cluster. Right now we just do it willy nilly on our own, running commands by hand. This should be automated. Also, there's no real point of updating stats automatically if the data doesn't change.

# Solution

* Add a command that would run from lemmy to request that the cluster updates it's data. 
* Add a consumer in spark-land which takes the aforementioned request, imports the data and sends an acknowledgement
* Add a consumer in webserver-land which takes the acknowledgement and sends `listenbrainz-observability@metabrainz.org` an email.
* Add a bunch of tests.
* Add a cron job.

The schedule of the cron job is staggered 7 days behind the date of actual generation. Full dump generation is really slow, it literally takes days, so for now 7 seems safe to me. The actual way forward here would be to fix incremental dumps and then only import incremental dumps, which take minutes to generate.

# Action

I've deployed this (without the cron job) to `spark-reader-beta` on lemmy and to `request-consumer` on leader and have triggered an import.


paramsingh:
WhatsApp Image 2020-02-23 at 23 56 35

:tada: :tada: :tada:

docker/stats-crontab
```diff
@@ -1,3 +1,7 @@
 PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin
 
+# Request user stats every day at 12:00
 00 12 * * * /usr/local/bin/python /code/listenbrainz/manage.py spark request_all_user_stats
+
+# Request full data dump import into spark cluster on 8th and 22nd every month
+00 00 8,22 * * /usr/local/bin/python /code/listenbrainz/manage.py spark request_import_full
```
mayhem: Remind me what are the other dates that we do stuff on?

PR #747, opened 2020-03-01
Write tests for LastFM Importer

ishaanshah:
# Description
Write tests for the LastFM Importer written in ReactJS

This is a continuation of #739


listenbrainz/webserver/static/js/jsx/importer.test.js
```diff
@@ -0,0 +1,310 @@
...
+    });
+
+    importer.submitPage();
+    jest.runAllTimers();
+
+    // Flush all promises
+    // https://stackoverflow.com/questions/51126786/jest-fake-timers-with-promises
+    await new Promise(resolve => setImmediate(resolve));
+    
+    expect(importer.APIService.submitListens).toHaveBeenCalledWith("foobar", "import", undefined);
```
MonkeyDo: Here you check if it's been called, but you could also check that it's been called only once, as your test description suggests.

listenbrainz/webserver/static/js/jsx/importer.js
```diff
@@ -43,7 +43,7 @@ export default class Importer {
     this.canClose = false; // Disable the close button
     this.updateMessage("Your import from Last.fm is starting!");
     this.playCount = await this.getTotalNumberOfScrobbles();
-    this.latestImportTime = await this.APIService.getLatestImport(this.userName); // TODO: Error handling, test usernames having special characters
```
MonkeyDo: I see no test currently for `getLatestImport`.
I think it would be good to add a test as describe in this comment before removing it.

listenbrainz/webserver/static/js/jsx/lastFmImporterModal.test.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import LastFmImporterModal from './lastFmImporterModal'
+
+let wrapper = null;
+describe('LastFmImporterModal is working correctly', () => {
```
MonkeyDo: There's not much else to test on the modal, but you can check if close the button is disabled or enabled when expected when changing the props.

ishaanshah:
Thanks for the review @MonkeyDo. @paramsingh asked me to keep the test files and source files in the same directory, so as of now I haven't changed the location of the test files. I have added tests for the functions in APIService which were used by the importer. Please have a look.

paramsingh:
Can you rebase so this can be merged?

PR #754, opened 2020-03-13
create multiple sessions for spark applications

vansika:
Ever since request consumer is being used in production, we are not able to run other applications on leader because they get stuck. I tried to run scripts a few weeks ago and one thing that bothered me was that there was no mention of executors in logs which means executors were not getting created. I checked configs which were as follows:
- MAX_CORES = 21
- EXECUTOR_CORES = 5

I then checked the configs used by request consumer, which were identical. So here is the catch, since request consumer is alive all the time, 5 cores of each node are never free, therefore, on every node 3 cores are free to be used by other applications.

So I changed the configs accordingly. This time the executors showed up but another error followed. 
https://gist.github.com/vansika/310b3f20056133060090ecb2c98f7e96

Here is a proposed solution but I am still not sure: 
https://stackoverflow.com/questions/35485662/local-class-incompatible-exception-when-running-spark-standalone-from-ide

Another thing which we know is that there can be one spark context on one JVM, though we can change the configs to allow multiple spark contexts but this practice is highly discouraged. Hence creating multiple spark sessions is good idea to run applications parallelly.

https://www.waitingforcode.com/apache-spark-sql/multiple-sparksession-one-sparkcontext/read

https://stackoverflow.com/questions/43516948/how-do-i-run-multiple-spark-applications-in-parallel-in-standalone-master

The above two links give a good insight into this. 

There can be two options I can think as of now, first is to keep the request consumer alive and the script that initiates request consumer's spark session gives new sessions to new applications as and when required. But I doubt if it can happen because imo it is only possible if all the applications start 
at the same time and acquire their spark sessions.

The second way as @paramsingh suggested is to create separate instance of Spark and use the available resources. 

This is just a draft PR to explain my thoughts and the possibilities.

### DIFF
--- a/listenbrainz_spark/__init__.py
+++ b/listenbrainz_spark/__init__.py
@@ -5,17 +5,12 @@ from listenbrainz_spark.exceptions import SparkSessionNotInitializedException
 from pyspark import SparkContext
 from pyspark.sql import SparkSession, SQLContext
 
-session = None
-context = None
-sql_context = None
-
 def init_spark_session(app_name):
     """ Initializes a Spark Session with the given application name.
 
         Args:
             app_name (str): Name of the Spark application. This will also occur in the Spark UI.
     """
-    global session, context, sql_context
     try:
         session = SparkSession \
                 .builder \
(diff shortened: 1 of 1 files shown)


vansika:
if spark context is shared automatically most of the time then there may be a possibility that request consumer was not deployed correctly leading to the version incompatibility error.

paramsingh:
We're gonna fix this by investing into the developer environment, it should be possible to do development on local environments, deploy to production and then check what happens. 

https://tickets.metabrainz.org/browse/LB-537

Running in-development non code-reviewed scripts on production is risky and should be highly discouraged anyways.

I don't think we want to solve the problem where multiple spark contexts can't exist on the same JVM right now. For now, let's make everything go through request consumer and once we start hitting bottlenecks there, see what we can do.

paramsingh:
I'm gonna close this because it has been lingering a long time, but feel free to discuss more about this on IRC or reopen.

PR #767, opened 2020-03-27
LB-503: Add command to connect to Timescale DB via develop.sh

shivam-kapila:
# Description

Just like `./develop.sh psql` we can have a command to connect to Timmescale DB directly.

# Problem

There didn't exist any command to directly enter into PSQL shell and connect to timescale DB.

# Solution

- Add  `timescale` option to `./develop.sh` to enter into PSQL shell and connect to timescale DB automatically.
- Use URLs to connect to the PSQL dbs.

# Action

- Enter the `./develop.sh timescale` command in the 
Done. You can now query the Timescale Db directly.

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -23,6 +23,12 @@ function open_psql_shell {
             -h db listenbrainz
 }
 
+function open_timescale_shell {
+    invoke_docker_compose run --rm web psql \
+            -U timescale_lb  \
+            -h timescale timescale_lb
+}
+
 function npm_install {
     invoke_docker_compose run --rm -e \
                 HOME=/tmp static_builder npm install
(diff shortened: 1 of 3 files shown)


alastair:
Because this is only for development, and the local user password is known and always the same, we can specify the password in the run command so that you don't have to enter it all the time. It looks like there are a few different ways of doing this here: https://stackoverflow.com/questions/6405127/how-do-i-specify-a-password-to-psql-non-interactively/6405162
Specifying a connect string like we already have in the config file is probably a good solution (SQLALCHEMY_TIMESCALE_URI), it keeps us consistent in different parts of the project

shivam-kapila:
@alastair As you mentioned I added URL to connect to PSQL db keeping it consistent with _config.py_. I tried to use variables from _config.py_ but its not possible due to `import os` and spaces in variable declaration.

Using these URLs the user doesn't need to enter the passwords now. Please have a look.

PR #778, opened 2020-04-05
Add ESLint and Prettier to ListenBrainz

ishaanshah:
# Description

Add `ESLint` and `Prettier` to ListenBrainz codebase

# Problem

Currently there is no coding style/standard enforced on the frontend codebase. This leads to inconsistent code across different files. 

# Solution

Adding `ESLint` and `Prettier` helps enforce standard coding style across the codebase.


.travis.yml
```diff
@@ -54,3 +54,4 @@ script:
   - echo "JavaScript tests start here"
   - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm frontend_tester npm test
   - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm frontend_tester npm run type-check
+  - docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test run --rm frontend_tester npm run format
```
paramsingh, line 57: Does this actually format the file?

The behaviour in travis I'd expect is for this return a non-zero exit code if there's any file which is not linted correctly and fail the build.

docs/dev/devel-env.rst
```diff
@@ -288,8 +288,22 @@ once so that subsequent running of the tests is faster:
 If you made any changes to the frontend, you can run the tests for frontend using
 
 .. code-block:: bash
+
     ./frontend-test.sh
 
+ListenBrainz uses ESLint to lint the frontend codebase, you can run ESLint using,
```
paramsingh, line 294:
```suggestion
ListenBrainz uses ESLint to lint the frontend codebase, please make sure you lint all new frontend code using,
```

docs/dev/devel-env.rst
```diff
@@ -288,8 +288,22 @@ once so that subsequent running of the tests is faster:
...
+
+.. code-block:: bash
+
+    ./lint.sh
+
+You can also setup a pre-commit hook to do the above process before each commit.
```
paramsingh, line 300:
```suggestion
You can also setup a pre-commit hook to do the above process automatically before each commit.
```

.eslintrc.js
```diff
@@ -0,0 +1,49 @@
...
+    'react',
+    '@typescript-eslint',
+    'prettier',
+  ],
+  rules: {
+    'prettier/prettier': 'error',
+  },
+  overrides: [
+    {
+      files: '**/*.test.js',
```
paramsingh, line 35: If our test files will be `test.ts` we should add that here too, maybe

.travis.yml
ishaanshah: It doesn't actually format the files, a non-zero exit code along with the errors made will be returned.

lint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+docker-compose -f docker/docker-compose.yml \
```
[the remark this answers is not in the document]
ishaanshah: This script is also called in the pre-commit hook, which means the image will be built each time a commit is made. This will slow down committing and linting significantly.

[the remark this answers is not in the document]
paramsingh: Yeah, that does make sense. Let's add a comment in the documentation that the docker images needs to be built before lint.sh works correctly

.travis.yml
paramsingh: Sgtm, thanks for clarifying!

PR #788, opened 2020-04-11
Fix LastFM importer icons

ishaanshah:
# Description

# Problem

# Solution

# Action


listenbrainz/webserver/static/js/src/LastFMImporterModal.tsx
```diff
@@ -1,6 +1,7 @@
+/* eslint-disable no-unused-vars */
```
paramsingh, line 1: Do we have a typescript equivalent of this rule? I'd just prefer to disable it everywhere

PR #789, opened 2020-04-11
Convert playback-controls to typescript.

paramsingh:
# Description

Converted the playback controls component to use typescript.

# Problem

Didn't use typescript, hard to understand.

# Solution

Uses types and is much easier to understand now.

Also added a snapshot test.


listenbrainz/webserver/static/js/src/PlaybackControls.tsx
```diff
@@ -0,0 +1,154 @@
...
+  faFastBackward,
+  faFastForward,
+  faPauseCircle,
+  faPlayCircle,
+  faSortAmountDown,
+  faSortAmountUp,
+} from "@fortawesome/free-solid-svg-icons";
+
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { IconDefinition } from "@fortawesome/fontawesome-common-types"; // eslint-disable-line import/no-unresolved
```
paramsingh, line 14: I'm not sure why this was getting triggered, typescript had no problems with the import and we have it installed in package.json

listenbrainz/webserver/static/js/src/PlaybackControls.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { IconProp } from "@fortawesome/fontawesome-svg-core";
+
+export type SpotifyPlayDirection = "up" | "down" | "hidden";
```
ishaanshah, line 17: I think we should keep all the types in the types.d.ts file, that ensures re usability

paramsingh: Moved the non prop and non state types to types.d.ts. The propstype and statetype are gonna stay here, as agreed on irc.

PR #791, opened 2020-04-11
Port RecentListens to Typescript

ishaanshah:
# Description

We're moving over to Typescript for our frontend code. This pull request migrates the profile.jsx file to typescript.

We also do some linting fixes here.

# Problem
The LB frontend code is very hard to read or modify because it's complex and javascript provides no guarantee of what is available where etc.

# Solution
Use typescript instead, which helps us have types on react components and functions and catch errors at compile time.

The linter has also caught a few actual bugs too.


listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -31,6 +33,47 @@ declare interface ModalProps {
...
+  save_url?: string;
+  spotify: {
+    access_token?: string;
+    permission?: string;
+  };
+  user: User;
+  web_sockets_server_url: string;
+}
+
+declare interface RecentListensState {
```
paramsingh, line 61: I'd like to keep component specific props and states in the component file itself.

* We shouldn't need to import these types much anyways other than tests. 
* It is gonna be easier to read the props etc if they're in the file vs having to open a file which will potentially get huge if we put ALL types in it

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -90,23 +90,7 @@ document.addEventListener("DOMContentLoaded", () => {
...
-    <LastFmImporter
-      user={user}
-      profileUrl={profile_url}
-      apiUrl={api_url}
-      lastfmApiKey={lastfm_api_key}
-      lastfmApiUrl={lastfm_api_url}
-    />,
-    domContainer
-  );
+  ReactDOM.render(<LastFmImporter {...reactProps} />, domContainer);
```
paramsingh, line 95: Let's keep the props camelCase here.

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -35,77 +45,57 @@ class RecentListens extends React.Component {
...
     }
-  }
+  };
 
-  connectWebsockets() {
-    this._socket = io.connect(this.props.web_sockets_server_url);
-    this._socket.on("connect", () => {
-      console.debug("Connected to websocket!");
-      switch (this.state.mode) {
+  connectWebsockets = () => {
```
paramsingh, line 68: Let's add return types to these functions as well.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -31,6 +35,46 @@ declare interface ModalProps {
   onClose(event: React.MouseEvent<HTMLButtonElement>): void;
 }
 
+declare interface RecentListensProps {
+  api_url: string;
```
paramsingh, line 39: Let's keep the props camelCase.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -7,12 +11,16 @@ declare type SpotifyPlayDirection = "up" | "down" | "hidden";
...
 
+declare type AlertType = "danger" | "warning" | "success";
+
+declare type User = {
+  id?: string;
+  name: string;
+  auth_token?: string;
+};
+
 declare interface ImporterProps {
```
ishaanshah, line 22: Has been changed in #793

listenbrainz/webserver/static/js/src/SpotifyPlayer.tsx
```diff
@@ -141,7 +141,6 @@ export default class SpotifyPlayer extends React.Component<
         }
         this.handleWarning("Could not find track on Spotify");
         this.playNextTrack();
-        return;
```
ishaanshah, line 144: Extra return statements, removed by the linter. I checked if player was working as expected and the alert bug didn't appear. This should be safe to remove.

PR #792, opened 2020-04-11
Convert SpotifyPlayer to typescript

paramsingh:
This PR finishes off #789 and converts the SpotifyPlayer to Typescript.

While testing I found out that the SpotifyPlayer doesn't work in Chrome right now, even on production. Need to look into that. :D

I've tested it out on firefox and it works though. The way forward here to me seems to be to merge this and then look into why the player doesn't work on Chrome.

https://tickets.metabrainz.org/browse/LB-521


ishaanshah:
I couldn't test the player as I don't have a premium account. The player however is showing a lot of popups. This is not happening on production.
Peek 2020-04-12 12-14

paramsingh:
@ishaanshah, can you test again? I've pushed a fix.

PR #796, opened 2020-04-12
Fix the url for next and previous page in the recent listens page.

paramsingh:
# Description

This PR fixes the bug on beta where the next page and previous page button doesn't work.

# Problem

We accidentally changed the parameter we send to the backend. 

# Solution

Fix the param. I have a test for this but that'll only pass when FollowUsers is ported.

Also fixed some types to match the data being sent from the backend.

# Action

Merge and deploy asap.


paramsingh:
This is relatively urgent, the next button on beta is broken right now, so I'm merging this. Feel free to leave comments and I'll fix them in subsequent PRs.

PR #797, opened 2020-04-12
Add a snapshot test for the recent listens page

paramsingh:
Add a snapshot test with real data from my profile.

The test doesn't pass right now, because jest doesn't handle the `@ts-ignore` before the follow-users import well. 

We can merge this test once follow users is ported over.


listenbrainz/webserver/static/js/src/__mocks__/recentListensProfilePageProps.json
```diff
@@ -0,0 +1,1074 @@
+{
+    "user": {
+        "id": 1,
+        "name": "iliekcomputers"
+    },
+    "listens": [
```
paramsingh, line 6: We can use this data to make a pretty comprehensive listen type.

PR #807, opened 2020-04-13
Fix RecentListens snapshot test

ishaanshah:
# Description
Fixes the issue with failing snapshot test for RecentListens

# Problem
The timestamp on RecentListens is calculated relative to the current time. This cause the snapshot tests to fail as the timestamp changes.

# Solution
Mocking the `timago.ago` function to return a constant value make the snapshot test independent of 
current time


ishaanshah:
> Why do you want to change it to shallow rendering here?

Mount renders child components too. The Spotify Player was causing some warnings to pop up. As we are testing Spotify Player in it's own snapshot test and we want the tests to be independent, I think we should use `shallow` rendering.

PR #808, opened 2020-04-15
Add definition to remaining types

ishaanshah:
# Description
Adds definition to remaining types.

# Problem
Some types were defined as `any`, which doesn't help in development or to catch errors.

# Solution
Adding correct definitions to these types fixes the problem

### DIFF
--- a/frontend-test.sh
+++ b/frontend-test.sh
@@ -2,6 +2,7 @@
 
 # ./frontend-test.sh                    Run tests
 # ./frontend-test.sh -u                 Run tests, update snapshots
+# ./frontend-test.sh -t                 Run type-checker
 # ./frontend-test.sh -b                 Build containers
 
 if [ "$1" == "-u" ]; then
@@ -13,6 +14,13 @@ fi
 
 if [ "$1" == "-b" ]; then
     docker-compose -f docker/docker-compose.test.yml -p listenbrainz_test build
+    exit 0
+fi
+
+if [ "$1" == "-t" ]; then
+    docker-compose -f docker/docker-compose.test.yml \
+                   -p listenbrainz_test \
+                   run --rm frontend_tester npm run type-check
     exit 0
 fi
 
--- a/listenbrainz/webserver/static/js/src/APIService.test.ts
+++ b/listenbrainz/webserver/static/js/src/APIService.test.ts
@@ -16,7 +16,15 @@ describe("submitListens", () => {
   });
 
   it("calls fetch with correct parameters", async () => {
-    await apiService.submitListens("foobar", "import", "foobar");
+    await apiService.submitListens("foobar", "import", [
+      {
+        listened_at: 1000,
+        track_metadata: {
+          artist_name: "foobar",
+          track_name: "bazfoo",
+        },
+      },
+    ]);
     expect(window.fetch).toHaveBeenCalledWith("foobar/1/submit-listens", {
       method: "POST",
       headers: {
@@ -25,7 +33,15 @@ describe("submitListens", () => {
       },
       body: JSON.stringify({
         listen_type: "import",
-        payload: "foobar",
+        payload: [
+          {
+            listened_at: 1000,
+            track_metadata: {
+              artist_name: "foobar",
+              track_name: "bazfoo",
+            },
+          },
+        ],
       }),
     });
   });
@@ -44,7 +60,15 @@ describe("submitListens", () => {
         });
       });
 
-    await apiService.submitListens("foobar", "import", "foobar");
+    await apiService.submitListens("foobar", "import", [
+      {
+        listened_at: 1000,
+        track_metadata: {
+          artist_name: "foobar",
+          track_name: "bazfoo",
+        },
+      },
+    ]);
     expect(setTimeout).toHaveBeenCalledTimes(1);
   });
 
@@ -57,7 +81,15 @@ describe("submitListens", () => {
       });
     });
 
-    await apiService.submitListens("foobar", "import", "foobar");
+    await apiService.submitListens("foobar", "import", [
+      {
+        listened_at: 1000,
+        track_metadata: {
+          artist_name: "foobar",
+          track_name: "bazfoo",
+        },
+      },
+    ]);
     expect(setTimeout).toHaveBeenCalledTimes(1);
   });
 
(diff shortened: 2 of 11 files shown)


listenbrainz/webserver/static/js/src/APIService.test.ts
```diff
@@ -70,29 +102,90 @@ describe("submitListens", () => {
...
+        },
+      ])
     ).resolves.toEqual({
       ok: true,
       status: 200,
     });
   });
 
   it("calls itself recursively if size of payload exceeds MAX_LISTEN_SIZE", async () => {
-    // Change MAX_LISTEN_SIZE to 0
```
ishaanshah, line 87: `MAX_LISTEN_SIZE` can't be 0 as it would lead to infinite recursion.

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -206,6 +212,7 @@ export default class RecentListens extends React.Component<
         return {
           playingNowByUser: {
             ...prevState.playingNowByUser,
+            // @ts-ignore
```
paramsingh, line 215: what was the error here?

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -410,7 +417,7 @@ export default class RecentListens extends React.Component<
                       .map((listen) => {
                         return (
                           <tr
-                            key={listen}
+                            key={listen.listened_at}
```
paramsingh, line 420: let's use `listened_at-recording_msid-username` here. `listened_at` can still be duplicated, for example in follow page.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -2,8 +2,43 @@ declare module "spotify-web-playback-sdk";
 declare module "react-bs-notifier";
 declare module "time-ago";
 
-// TODO: make this type specific
-declare type Listen = any;
+// TODO: Remove "| null" when backend stops sending fields with null
+declare type Listen = {
+  listened_at: number;
+  listened_at_iso?: string | null;
```
paramsingh, line 8: Is there a way to get rid of these `| null`s, they seem non-ideal.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -2,8 +2,43 @@ declare module "spotify-web-playback-sdk";
...
+      release_msid?: string | null;
+      spotify_album_artist_ids?: Array<string> | null;
+      spotify_album_id?: string | null;
+      spotify_artist_ids?: Array<string> | null;
+      spotify_id?: string | null;
+      tags?: Array<string> | null;
+      track_mbid?: string | null;
+      tracknumber?: number | null;
+      work_mbids?: Array<string> | null;
+      [key: string]: any;
```
paramsingh, line 37: if you make additional info a seperate interface, you might not need this line.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -13,38 +48,75 @@ declare type ListenBrainzUser = {
...
   access_token: string;
   permission: string;
 };
 
-declare type SpotifyPermission = any; // TODO
-declare type SpotifyTrack = any; // TODO
-declare type SpotifyArtist = any; // TODO
-declare type SpotifyDeviceID = any; // TODO
-declare interface SpotifyImage {
+declare type SpotifyPermission = string;
```
paramsingh, line 62: we only have a few specific permissions for spotify. 

Let's make the type reflect that. (https://github.com/metabrainz/listenbrainz-server/blob/5b1a66ea/listenbrainz/domain/spotify.py#L14)

listenbrainz/webserver/static/js/src/types.d.ts
ishaanshah: We can turn off `strictNullChecks` in typescript which will treat `null` and `undefined` as the same everywhere. But then we will also lose other errors such as **Possibly null/undefined**.

listenbrainz/webserver/static/js/src/RecentListens.tsx
ishaanshah: Fixed it!

listenbrainz/webserver/static/js/src/types.d.ts
ishaanshah: I did not understand how making `additional_info` into a separate interface will help solve the issue. Can you please explain how it can be done.

listenbrainz/webserver/static/js/src/types.d.ts
paramsingh: That does not sound like a good idea. Let's keep the nulls for now.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -13,38 +49,81 @@ declare type ListenBrainzUser = {
...
+  name: string;
+};
+
+declare type SpotifyTrack = {
+  album: {
+    uri: string;
+    name: string;
+    images: Array<SpotifyImage>;
+  };
+  artists: Array<SpotifyTrack>;
```
paramsingh, line 87:
```suggestion
  artists: Array<SpotifyArtist>;
```

?

listenbrainz/webserver/static/js/src/SpotifyPlayer.tsx
```diff
@@ -82,6 +82,7 @@ export default class SpotifyPlayer extends React.Component<
     this.state = {
       accessToken: props.spotifyUser.access_token,
       permission: props.spotifyUser.permission,
+      // @ts-ignore
```
paramsingh, line 85: why is this needed?

listenbrainz/webserver/static/js/src/Scrobble.ts
```diff
@@ -57,7 +57,7 @@ export default class Scrobble {
     return "";
   }
 
-  scrobbledAt(): string {
+  scrobbledAt(): number {
     /* Returns scrobbledAt if present, else returns an empty string */
```
paramsingh, line 61: let's update the comment here too, we return a -1 now.

frontend-test.sh
```diff
@@ -19,3 +19,7 @@ fi
 docker-compose -f docker/docker-compose.test.yml \
                -p listenbrainz_test \
                run --rm frontend_tester npm test
+
+docker-compose -f docker/docker-compose.test.yml \
```
paramsingh, line 23: why did you add this back? We removed it in #790. If we want to keep it, I think frontend.sh should have options that allow for running just the tests or just the type checker.

ishaanshah: I added it so that, type checking can be done on all files once before commit rather than opening each file and checking for errors. I have added an option to run type checker only, using `./frontend-test.sh -t`.

paramsingh:
For the future, can you make changes after a review in a different commit without rebasing / amending the previous commit? It makes things easier to review.

PR #810, opened 2020-04-19
Add tests for RecentListens

ishaanshah:
# Description

Adds tests for the RecentListens.tsx file

Note: This is a continuation of #797


paramsingh:
@ishaanshah i've fixed the problems you mentioned in the two commits I added, take a look. the test still fails because the `expect(handler).toHaveBeenCalled` is incorrect. The handler won't be called until an actual connection happens which we do not want in tests.

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -1,10 +1,17 @@
+/* eslint-disable dot-notation */
```
paramsingh, line 1: Can we do this on the specific lines where it causes problems, doing eslint-disables on entire files is something i'd like to avoid.

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -43,7 +50,353 @@ const props = {
...
+      track_name: "Viva La Vida",
+    },
+    listened_at: 1586580524,
+    listened_at_iso: "2020-04-10T10:12:04Z",
+  };
+
+  it("crops the listens array if length is more than or equal to 100", () => {
+    const wrapper = shallow<RecentListens>(
+      <RecentListens
+        {...(JSON.parse(JSON.stringify(tooManyListens)) as RecentListensProps)}
```
paramsingh, line 247: do we need the json.parse(json.dumps(object)) here?

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -43,7 +50,353 @@ const props = {
...
+    const instance = wrapper.instance();
+
+    wrapper.setState({ mode: "follow" });
+    instance.receiveNewListen(JSON.stringify(mockListen));
+
+    expect(wrapper.state()["listens"].length).toBeLessThanOrEqual(100);
+
+    wrapper.setState({
+      mode: "listens",
+      listens: JSON.parse(JSON.stringify(tooManyListens["listens"])),
```
paramsingh, line 259: do we need the json.parse(json.dumps(object)) here?

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -43,7 +50,353 @@ const props = {
...
+    });
+    instance.receiveNewListen(JSON.stringify(mockListen));
+
+    expect(wrapper.state()["listens"].length).toBeLessThanOrEqual(100);
+  });
+
+  it('inserts the recieved listen for "follow"', () => {
```
paramsingh, line 266:
```suggestion
  it('inserts the received listen for "follow"', () => {
```

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -43,7 +50,353 @@ const props = {
...
+    wrapper.setState({ mode: "follow" });
+    let result: Array<Listen> = recentListenPropOneListen["listens"]
+    result.push(mockListen);
+    instance.receiveNewListen(JSON.stringify(mockListen));
+
+    expect(wrapper.state('listens').length).toEqual(result.length)
+    expect(wrapper.state("listens")).toEqual(result);
+  });
+
+  it("inserts the recieved listen for other modes", () => {
```
paramsingh, line 282:
```suggestion
  it("inserts the received listen for other modes", () => {
```

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
ishaanshah: Yes, we do, otherwise  `tooManyListens` gets passed by reference. Which means `result` and and the `state` will always be the same. `JSON.parse(HSON.stringify(object))` is a fast way to deep copy so that this doesn't occur.

paramsingh: Cool, could you add a comment explaining that? So that the next person who sees the code understands it.

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -43,7 +50,493 @@ const props = {
...
+
+describe("handleFollowUserListChange", () => {
+  it("sets the state correctly", () => {
+    const wrapper = shallow<RecentListens>(<RecentListens {...props} />);
+    const instance = wrapper.instance();
+
+    instance.handleFollowUserListChange(["foo", "bar"], true);
+
+    // eslint-disable-next-line dot-notation
+    expect(wrapper.state()["followList"]).toEqual(["foo", "bar"]);
```
paramsingh, line 142:
```suggestion
    expect(wrapper.state("followList")).toEqual(["foo", "bar"]);
```

You could do this and get rid of the eslint disable

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -43,7 +50,493 @@ const props = {
...
+      mode: "listens",
+      listens: JSON.parse(
+        // eslint-disable-next-line dot-notation
+        JSON.stringify(recentListensPropsTooManyListens["listens"])
+      ),
+    });
+    instance.receiveNewListen(JSON.stringify(mockListen));
+
+    // eslint-disable-next-line dot-notation
+    expect(wrapper.state()["listens"].length).toBeLessThanOrEqual(100);
```
paramsingh, line 287: Same here, don't need the eslint-disable

PR #825, opened 2020-05-01
Spark development environment and documentation improvements

paramsingh:
# Problem

It is not very easy to develop for spark locally. The documentation is a bit too verbose,
and there is no documentation of request consumer.

# Solution

* Split the spark environment setup instructions into a different page.
* Add a page explaining how request consumer works and how to start it locally.
* Change the order of pages on the docs index page to focus on API docs first and then ListenBrainz development docs.
* Move client libraries list from html to readthedocs

# Action

cc @ishaanshah @ferbncode, i'd be interested in feedback for this PR from the POV of newcomers to the spark environment.


docs/dev/devel-env.rst
```diff
@@ -321,112 +321,4 @@ all errors.
...
-data.
-
-When the tests complete, you will see if your changes are valid or not. These tests
-are a helpful way to validate new changes without a lot of work.
-
-Refer the `FAQs`_ to resolve the common errors that may arise when setting up 
-the development environment.
-
-.. _FAQs: https://github.com/metabrainz/listenbrainz-server/blob/master/docs/dev/faqs.rst
+        ./develop build static_builder
```
shivam-kapila, line 324: It should be `./develop.sh build static_builder`

docs/dev/spark-architecture.rst
```diff
@@ -0,0 +1,57 @@
...
+
+.. code-block:: bash
+
+    ./develop.sh run web python -m listenbrainz.spark.spark_reader
+
+Now, you have everything needed to work with Spark. You can trigger a request like this
+
+.. code-block:: bash
+
+    ./develop.sh run web python manage.py spark request_all_user_stats
```
shivam-kapila, line 57: We can simplify it to
`./develop.sh manage spark request_all_user_stats`.

docs/dev/devel-env.rst
shivam-kapila, line 324: We should add a Link to FAQs from here too as we do in spark-devel-env.rst

docs/dev/spark-architecture.rst
paramsingh: I'm not using that because I have a feeling that it will change. `develop.sh` is mostly supposed to be a wrapper around docker-compose but I feel it's gotten a bit too complicated right now.

shivam-kapila:
@paramsingh Can you fix the **Here is some example code to explain the mentioned way of parsing the listens dump.** too.
I had removed the include code block by mistake and replace it with an anchor link.
Please refer the line just above `Incremental dumps` [here](https://github.com/metabrainz/listenbrainz-server/blob/master/docs/dev/listenbrainz-dumps.rst#incremental-dumps)

docs/dev/spark-architecture.rst
alastair: as a clarification here, this isn't strictly true any more, we're open to adding subcommands to develop.sh to do more complex things, as a way of centralising all commands.
I see this as a perfect use of `develop.sh manage` (in fact, it's better than your command in this case because it also has the --rm flag)

shivam-kapila: To clarify we don't need to add a new clause to `./develop.sh` for the command I listed. It works now also because we have imported request_manage.py  in manage.py [here](https://github.com/metabrainz/listenbrainz-server/blob/master/manage.py#L191).

paramsingh: oh ok, fixed that then.

docs/dev/spark-devel-env.rst
```diff
@@ -0,0 +1,110 @@
...
+Import data into the spark environment
+--------------------------------------
+
+We provide small data dumps that are helpful for working with real ListenBrainz data.
+Download and import a data dump into your spark environment using the following 
+commands.
+
+.. code-block:: bash
+
+    ./spark_develop.sh run playground python spark_manage.py upload_listens -i
```
shivam-kapila, line 68: It should be `./spark_develop.sh run request_consumer python spark_manage.py upload_listens -i`

paramsingh: Nice catch! Thanks!

shivam-kapila:
[This](https://github.com/metabrainz/listenbrainz-server#development-environment) section needs some changes. The production version in readthedocs is the latest one. Maybe we should remove v:latest from readthedocs

PR #826, opened 2020-05-02
LB-539: request consumer rabbitmq improvements

paramsingh:
# Problem

The RabbitMQ connection seems to close after handling one request on production.

LB-539

# Solution

* Increase the heartbeat, so that the connection doesn't die because of our long running tasks.
* Add more logs 
* Close the connection and make a new one everytime there's a `ChannelClosed` or `ConnectionClosed` error.


paramsingh:
Need to configure the line length to be a bit higher. maybe 130/140?

PR #828, opened 2020-05-02
Fix last_update_ts for user stats

shivam-kapila:
# Problem

The spark reader raised an error when the user stats are calculated for the first time as `last_update_ts` is None and hence can't be used in datetime operations.

# Solution

Using a default min value for datetime (in UTC timezone) for last_update_ts if its None fixes the issue.

### DIFF
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -17,7 +17,11 @@ def is_new_user_stats_batch():
     So, we check the database and see if the difference between the last time stats were updated
     and right now is greater than 12 hours.
     """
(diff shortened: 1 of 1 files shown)


listenbrainz/spark/handlers.py
```diff
@@ -17,7 +17,11 @@ def is_new_user_stats_batch():
     So, we check the database and see if the difference between the last time stats were updated
     and right now is greater than 12 hours.
     """
-    return datetime.now(timezone.utc) - db_stats.get_timestamp_for_last_user_stats_update() > timedelta(hours=TIME_TO_CONSIDER_STATS_AS_OLD)
+    last_update_ts = db_stats.get_timestamp_for_last_user_stats_update()
+    if(last_update_ts is None):
```
paramsingh, line 21: brackets around if conditions isn't really pythonic.

shivam-kapila: My bad

PR #829, opened 2020-05-03
Fix Spark test container build

shivam-kapila:
# Problem

The spark tests are not running on the local machine due to error in the docker-compose file.

# Solution

- Fix the dockerfile for the test container
- Also, clean up the `develop.sh` and `spark_develop.sh files`.

### DIFF
--- a/spark_develop.sh
+++ b/spark_develop.sh
@@ -17,7 +17,7 @@ function format_namenode_and_datanode {
 }
 
 if [ "$1" == "format" ]; then shift
-	echo "Formatting NameNode..."
+    echo "Formatting NameNode..."
     format_namenode_and_datanode "$@"
     exit
 else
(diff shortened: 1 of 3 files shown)


spark_develop.sh
```diff
@@ -12,12 +12,12 @@ function invoke_docker_compose_spark {
...
 function format_namenode_and_datanode {
-    docker-compose -f docker/docker-compose.spark.yml \
-        -p listenbrainzspark run --rm hadoop-master hdfs namenode \
-        -format -nonInteractive -force
+    invoke_docker_compose_spark run --rm hadoop-master \
+        hdfs namenode -format -nonInteractive -force
 }
 
 if [ "$1" == "format" ]; then shift
+	echo "Formatting NameNode..."
```
paramsingh, line 20: Indent problem here, but looks good to me otherwise.

shivam-kapila: I don't know why it was not showing up in VS Code. But I force indented it.

PR #830, opened 2020-05-04
Add option to download a specific dump

ishaanshah:
Adds option to download a specific dump using `spark_manage`. Fixes some bugs and adds tests.
# Problem
Currently the `spark_manage` script allows to download the latest dump only. However this is inconvenient if a developer wants to use a older dump.

# Solution
The `download_listens` function has an option to download a dump by index. Exposing this option in the `spark_manage` script fixes the issue.

### DIFF
--- a/spark_manage.py
+++ b/spark_manage.py
@@ -90,7 +90,8 @@ def upload_mapping(force):
 @cli.command(name='upload_listens')
 @click.option('--incremental', '-i', is_flag=True, default=False, help="Use a smaller dump (more for testing purposes)")
 @click.option("--force", "-f", is_flag=True, help="Deletes existing listens.")
-def upload_listens(force, incremental):
+@click.option("--idx", default=None, type=int, help="Get a specific dump based on index")
+def upload_listens(force, incremental, idx):
     """ Invoke script to upload listens to HDFS.
     """
     from listenbrainz_spark.ftp.download import ListenbrainzDataDownloader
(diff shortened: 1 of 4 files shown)


spark_manage.py
```diff
@@ -90,15 +90,16 @@ def upload_mapping(force):
...
+def upload_listens(force, incremental, idx):
     """ Invoke script to upload listens to HDFS.
     """
     from listenbrainz_spark.ftp.download import ListenbrainzDataDownloader
     from listenbrainz_spark.hdfs.upload import ListenbrainzDataUploader
     with app.app_context():
         downloader_obj = ListenbrainzDataDownloader()
         dump_type = 'incremental' if incremental else 'full'
-        src, _ = downloader_obj.download_listens(directory=path.FTP_FILES_PATH, dump_type=dump_type)
+        src, _ = downloader_obj.download_listens(directory=path.FTP_FILES_PATH, listens_dump_id=idx, dump_type=dump_type)
```
paramsingh, line 102: Can we add a test that checks that download_listens is called with listens_dump_id when passed?

spark_manage.py
```diff
@@ -90,15 +90,16 @@ def upload_mapping(force):
 @cli.command(name='upload_listens')
 @click.option('--incremental', '-i', is_flag=True, default=False, help="Use a smaller dump (more for testing purposes)")
 @click.option("--force", "-f", is_flag=True, help="Deletes existing listens.")
-def upload_listens(force, incremental):
+@click.option("--idx", default=None, type=int, help="Get a specific dump based on index")
```
paramsingh, line 93: Let's make this `--id` for consistency, we call them dump IDs, not indexes.

PR #831, opened 2020-05-04
LB-511: Raise 400 if the payload doesn't contain any listens

paramsingh:
# Problem

LB-511

We returned a inconsistent "success" string if the payload to `submit-listens` contained no listens.

# Solution

Raise a 400 with an appropriate message and improve documentation.

### DIFF
--- a/listenbrainz/tests/integration/test_api.py
+++ b/listenbrainz/tests/integration/test_api.py
@@ -110,6 +110,18 @@ class APITestCase(IntegrationTestCase):
             content_type = 'application/json'
         )
 
+    def test_zero_listens_payload(self):
+        """ Test that API returns 400 for payloads with no listens
+        """
+
+        for listen_type in ('single', 'playing_now', 'import'):
+            payload = {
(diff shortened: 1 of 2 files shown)


alastair:
documentation at https://listenbrainz.readthedocs.io/en/production/dev/json/#submission-json should be updated

paramsingh:
@alastair I added a line to the documentation of the `submit-listens` endpoint.

alastair:
It'd be nice to update the json docs to say that a `playing_now` submission must contain exactly 1 submission. Currently it's still unclear as to if it can contain 0.

Zastai:
@alastair You point to the production docs for a change that I don't think is live yet.
https://listenbrainz.readthedocs.io/en/latest/dev/api/ does show the updated doc already.

Zastai:
Ah never mind, I see what you mean. The JSON docs for the 'import' type don't explicitly say there needs to be at least one. And the others say 'only' which is more ambiguous than 'exactly'.
I'll draft a PR for that part of the docs.

PR #833, opened 2020-05-04
Make payload content requirements more explicit

Zastai:
Based on comment on the PR for LB-511.

For single/playing-now, use "exactly one" instead of "only one".
For import, explicitly state "at least one", and reword the item describing the document size limit.

This also makes line spacing and indentation consistent across the three parts.


alastair:
this looks great, thanks

PR #834, opened 2020-05-05
LB-542: Add API endpoint for Artist Statistics

ishaanshah:
Adds API endpoint to get statistics for an user
# Problem
There is no way to programmatically get statistics for an user. 

# Solution
Adding an API endpoint `/1/stats/<user-name>/stats/artist` fixes this issue


listenbrainz/webserver/views/api.py
```diff
@@ -177,6 +177,36 @@ def get_playing_now(user_name):
     })
 
 
+@api_bp.route("user/<user_name>/stats/artist")
```
paramsingh, line 180: Can we move this endpoint to a different file and a different blueprint? We'll probably have lots of these endpoints soon.

docs/dev/json.rst
```diff
@@ -175,3 +175,62 @@ submitted via the ``additional_info`` element. We're open to see how people
...
+       "last_updated": "Sun, 03 May 2020 08:26:01 GMT",
+       "user_id": "John Doe"
+     }
+   }
+
+.. note::
+   `artist_mbids` and `artist_msid` are optional fields and may not be present in all the responses   
+
+.. warning::
+   The statistics API endpoint is still in beta and the `JSON` format may change in the future.
```
paramsingh, line 236: I'm happy to keep the beta tag, but let's try not to change the the json format, otherwise clients have no real incentive to actually use the endpoint.

Let's remove the format change warning

docs/dev/json.rst
```diff
@@ -175,3 +175,62 @@ submitted via the ``additional_info`` element. We're open to see how people
...
+                  "artist_mbids": [cc197bad-dc9c-440d-a5b5-d52ba2e14234],
+                  "artist_msid": "6599e41e-390c-4855-a2ac-68ee798538b4",
+                  "artist_name": "Coldplay",
+                  "listen_count": 321
+               },
+            ]
+         },
+         "count": 3
+       },
+       "last_updated": "Sun, 03 May 2020 08:26:01 GMT",
```
paramsingh, line 227: Could this last_uodated be a Unix timestamp, that would be much easier to parse?

docs/dev/json.rst
```diff
@@ -175,3 +175,62 @@ submitted via the ``additional_info`` element. We're open to see how people
 will make use of these unspecified fields and may decide to formally specify or
 scrub elements in the future.
 
+Artist Statistics JSON
+----------------------
+The JSON documents returned from our API look like the following::
+
+   {
+     "payload": {
+        "user_id": "-- the MusicBrainz ID of the user --",
```
paramsingh, line 184: Is this the string musicbrainz id or the database int id? If it's the string, can we call it musicbrainz_id

docs/dev/json.rst
```diff
@@ -175,3 +175,62 @@ submitted via the ``additional_info`` element. We're open to see how people
...
+        "artist": {
+           "all_time": {
```
paramsingh, line 186: Can we document that we only return all time stats for now but the artist field here could eventually have more time intervals?

listenbrainz/webserver/views/api.py
```diff
@@ -177,6 +177,36 @@ def get_playing_now(user_name):
...
+@crossdomain()
+@ratelimit()
+def get_artist(user_name):
```
paramsingh, line 183: We could add get params to specify the number of artists to return

listenbrainz/webserver/views/api.py
```diff
@@ -177,6 +177,36 @@ def get_playing_now(user_name):
...
+    :resheader Content-Type: *application/json*
+    """
+
+    user = db_user.get_by_mb_id(user_name)
+    if user is None:
+        raise APINotFound("Cannot find user: %s" % user_name)
+
+    stats = db_stats.get_user_artists(user['id'])
+    if stats is None:
+        return '', 204
```
paramsingh, line 201: We should always return valid json. Maybe return the same format but with an empty artist list?

listenbrainz/webserver/views/api.py
```diff
@@ -247,9 +277,9 @@ def latest_import():
         if user is None:
             raise APINotFound("Cannot find user: {user_name}".format(user_name=user_name))
         return jsonify({
-                'musicbrainz_id': user['musicbrainz_id'],
-                'latest_import': 0 if not user['latest_import'] else int(user['latest_import'].strftime('%s'))
-            })
+            'musicbrainz_id': user['musicbrainz_id'],
```
paramsingh, line 280: Try to avoid unrelated pep-8 changes. It makes things harder to review.

docs/dev/json.rst
```diff
@@ -175,3 +175,62 @@ submitted via the ``additional_info`` element. We're open to see how people
...
+              "artists": [
+                 "-- Artist stats here --"
```
paramsingh, line 188: I feel an example of what the artist stats look would be nice here.

docs/dev/json.rst
```diff
@@ -175,3 +175,62 @@ submitted via the ``additional_info`` element. We're open to see how people
 will make use of these unspecified fields and may decide to formally specify or
 scrub elements in the future.
 
+Artist Statistics JSON
```
paramsingh, line 178: I'm not sure it makes sense to put this in the json file which is pretty big already.  Do you think we could do a different file or put it in the endpoint documentation itself.

docs/dev/json.rst
ishaanshah: It is username, that is string Musicbrainz ID of the user

listenbrainz/webserver/views/api.py
ishaanshah: Sorry for that, I configured my editor to format on saves. It won't happen in the future.

listenbrainz/webserver/views/api.py
ishaanshah: A 204 response does not include a message body and is terminated after first empty line after response headers.
https://restfulapi.net/http-status-204-no-content/

docs/dev/json.rst
ishaanshah: The example of how the whole response will be has been given at the end. Will that suffice?

listenbrainz/webserver/views/stats_api.py
```diff
@@ -0,0 +1,57 @@
...
+    :statuscode 400: Bad request, `count` should be of type integer
+    :statuscode 404: User not found
+    :resheader Content-Type: *application/json*
+    """
+
+    count = request.args.get('count', default=-1)
+    try:
+        count = int(count)
+    except ValueError:
+        raise APIBadRequest("Bad request, 'count' should be an integer")
```
paramsingh, line 34: we should check that count is non-negative too, should raise a 400 if someone sends us -1

listenbrainz/webserver/views/stats_api.py
```diff
@@ -0,0 +1,57 @@
...
+@crossdomain()
+@ratelimit()
+def get_artist(user_name):
+    """
+    Get top artists for user ``user_name``. The format for the JSON returned is defined in our :ref:`json-doc`.
+
+    **Note**: This endpoint is currently in beta
+
+    :param count: Optional, number of artists to return.
+    :statuscode 200: Successfull query
```
paramsingh, line 23:
```suggestion
    :statuscode 200: Successful query, you have data!
```

listenbrainz/tests/integration/test_stats_api.py
```diff
@@ -0,0 +1,75 @@
+import json
+
+from flask import url_for
+
+import listenbrainz.db.stats as db_stats
+import listenbrainz.db.user as db_user
+from listenbrainz.tests.integration import IntegrationTestCase
+
+
+class APITestCase(IntegrationTestCase):
```
paramsingh, line 10: let's call this class StatsAPITestCase

listenbrainz/webserver/views/stats_api.py
```diff
@@ -0,0 +1,57 @@
...
+    if stats is None:
+        return '', 204
+
+    artist_list = stats['artist']['all_time']['artists']
+    if count >= 0:
+        artist_list = stats['artist']['all_time']['artists'][:count]
+
+    return jsonify({'payload': {
+        'user_id': user_name,
+        'artist': {
```
paramsingh, line 50: I don't think we need this `artist` dict here in the response. We can just keep the structure payload->all_time->artists

docs/dev/api.rst
```diff
@@ -61,6 +61,14 @@ Core API Endpoints
   :statuscode 400: invalid JSON sent, see error message for details.
   :statuscode 401: invalid authorization. See error message for details.
 
+Statistics API Endpoints
+^^^^^^^^^^^^^^^^^^^^
+
```
paramsingh, line 66: Let's add a small paragraph here about what stats are, how often they're updated and how they are in beta right now.

listenbrainz/tests/integration/test_stats_api.py
```diff
@@ -24,14 +24,43 @@ def test_artist(self):
         self.assert200(response)
         data = json.loads(response.data)['payload']
         sent_count = payload['count']
-        recieved_count = data['all_time']['count']
+        recieved_count = data['count']
```
Zastai, line 27: It's only a variable name in code, so not super important, but it's "receive", not "recieve"

listenbrainz/webserver/views/stats_api.py
```diff
@@ -17,55 +17,59 @@ def get_artist(user_name):
...
         }
 
     .. note::
-       - ``artist_mbids`` and ``artist_msid`` are optional fields and may not be present in all the responses
-       - As of now we are only calculating ``all_time`` statistics for artist.
-         However, we plan to add other time intervals in the future.
+        - This endpoint is currently in beta
+        - ``artist_mbids`` and ``artist_msid`` are optional fields and may not be present in all the responses
+        - As of now we are only calculating ``all_time`` statistics for artist.
+          However, we plan to add other time intervals in the future.
```
Zastai, line 56: If different ranges are going to be supported, would it not make sense not to use named ones in the response, but instead use something like `timespan: { from: <unix_ts>, to: <unix_ts> }`?
The current always-all-time-stats could set those either to the timestamps of the user's oldest and newest listen (if readily available), or even just the unix epoch as `from` and the current timestamp as `to`. That way, the payload is stable and usable, regardless of what kinds of ranges get supported later.

Zastai: If the user's oldest and newest listen timestamps are known, it would especially useful to have them, because you could show "Top <count> Artists for <user> (over <textual description of timespan>)", e.g "Top 10 Artists for Zastai (over 10 years and 5 months)".

"All time" is not very useful, because it could cover anything from "5 minutes" (if the user only has one submitted listen) to "18 years" (if the user has been using last.fm since it started).

ishaanshah: We currently calculate user statistics in a batch every 15 days so answering queries for arbitrary time periods is not possible. Can you open a ticket for the functionality so that we can look into it in the future. Thanks.

PR #838, opened 2020-05-07
LB-557, LB-570: Port User Artist graph to ReactJS

ishaanshah:
Ports the User Artist graph to ReactJS

# Problem
The current User Artist graph is written in vanilla JavaScript using `d3.js`. However, this makes the code hard to understand and maintain, especially because all other code has been written in ReactJS.

# Solution
Porting the User Artist Graph will improve the readability and re-usability of the code drastically. The PR makes use of `nivo` a charting library based on `d3.js` and `ReactJS`. `nivo` makes it easy to create interactive graphs while having control over how the graphs look.


ishaanshah:
2020-05-08-212721_1364x637_scrot
2020-05-08-212842_596x636_scrot
@paramsingh @MonkeyDo Please have look at the attached screenshot. I would love to have your feedback on it.

shivam-kapila:
I like wrapping long strings thing :)

listenbrainz/webserver/__init__.py
```diff
@@ -149,7 +149,7 @@ def create_app(config_path=None, debug=None):
     app = gen_app(config_path=config_path, debug=debug)
 
     # Static files
-    import listenbrainz.webserver.static_manager
+    import listenbrainz.webserver.static_manager as static_manager
```
paramsingh, line 152: why was this change needed?

listenbrainz/webserver/static/js/src/APIService.ts
```diff
@@ -83,14 +192,14 @@ export default class APIService {
...
 
-  /*
-    Send a POST request to the ListenBrainz server to submit a listen
-  */
   submitListens = async (
     userToken: string,
     listenType: ListenType,
     payload: Array<Listen>
   ): Promise<Response> => {
+    /*
```
paramsingh, line 200: I actually moved them above the function definition consciously. It makes sense to me there because we'd want to keep comments about the actual implementation inside the function.

listenbrainz/webserver/static/js/src/UserArtist.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+export type UserArtistProps = {
+  user: ListenBrainzUser;
+  apiUrl: string;
+};
+
+export type UserArtistState = {
+  data: Array<{
+    id: string;
+    Listens: number;
```
paramsingh, line 43:
```suggestion
    count: number;
```
makes more sense here

listenbrainz/webserver/static/js/src/UserArtist.tsx
```diff
@@ -0,0 +1,130 @@
...
+    this.APIService = new APIService(
+      props.apiUrl || `${window.location.origin}/1`
+    ); // Used to access LB API
+
+    this.state = {
+      data: [] as UserArtistState["data"],
+    };
+  }
+
+  processData = (data: any): UserArtistState["data"] => {
```
paramsingh, line 67: Let's define `UserArtistState["data"]` as an individual type instead of doing this. It's hard to understand.

listenbrainz/webserver/static/js/src/APIService.ts
ishaanshah: Sorry for that, fixed it.

listenbrainz/webserver/__init__.py
ishaanshah: There seems to be some bug with `static_manager`,  especially the `read_manifest` function. I thought the above change would fix the error. The change  worked once but did not work the next time. I committed the change by mistake. I will look into the bug and make appropriate changes, till then I am rolling back to the previous version without `as static_manger`.

ishaanshah:
> I like wrapping long strings thing :)

Even though wrapping strings would work on desktop, on mobile devices the string will wrap around twice or thrice because of the less width available which I think is not optimal. Also, the full artist name is shown when mouse is hovered upon the bar graph, so the information is not completely lost. @paramsingh @MonkeyDo any thoughts on this?

ishaanshah:
> I'm having some issue running locally, but from what I see of the screenshots it's looking pretty good. Similar to what was there previously but improved :)

@MonkeyDo Yes there is some issue with static file manager code, but I haven't gotten around to fix that yet. Can you try doing `./develop.sh` one or two more times, it generally works for me. I suspect that the manifest file is read before being updated hence the file is not found.

listenbrainz/webserver/static/js/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    if (hasError) {
+      document.title = "Something went wrong - Listenbrainz";
+      return (
+        <div>
+          <h2 className="page-title">{error.name}</h2>
+          <p>{error.message}</p>
+          <p>
+            <a href={`//${window.location.href.split("/")[2]}`}>
+              Back to home page
```
paramsingh, line 27: Could we possibly just do a reload instead? This error boundary is for frontend errors only, right?

listenbrainz/webserver/static/js/src/UserArtists.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  render() {
+    const { data } = this.state;
+    const prevPage = this.currPage - 1;
+    const nextPage = this.currPage + 1;
+
+    return (
+      <div>
+        <div className="row">
+          <div className="col-md-8" style={{ height: "75em" }}>
```
MonkeyDo, line 112: Here and for the pager below, why not use the full available width?
It looks a bit odd on a wide screen at the moment.

listenbrainz/webserver/static/js/src/UserArtists.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <Bar data={data} maxValue={this.maxListens} />
+          </div>
+        </div>
+        <div className="row">
+          <div className="col-md-8">
+            <ul className="pager">
+              <li className={`previous ${!(prevPage > 0) ? "hidden" : ""}`}>
```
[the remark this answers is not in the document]
MonkeyDo: Bootstrap's docs also suggest:
> You can optionally swap out active or disabled anchors for `<span>`, or omit the anchor in the case of the prev/next arrows, to remove click functionality and prevent keyboard focus while retaining intended styles.

listenbrainz/webserver/static/js/src/UserArtists.tsx
```diff
@@ -0,0 +1,153 @@
...
+                <a href={`${this.baseUrl}?page=${prevPage}`}>&larr; Previous</a>
+              </li>
+              <li
+                className={`next ${
+                  !(nextPage <= this.totalPages) ? "hidden" : ""
+                }`}
+              >
+                <a href={`${this.baseUrl}?page=${nextPage}`}>Next &rarr;</a>
```
MonkeyDo, line 127: Have you thought about how you could load the next/previous page without reloading the page?
I think if you extract some of the mechanisms in `componentDidMount` to be reusable (specifically the data loading), you could simply call it again with a page offset and that would recreate the graph (considering `this.state.data` would change).

That might also reduce the number of calls to `this.APIService.getUserStats` (like those to calls on startup) as you can store `maxListens` on the state once in `componentDidMount`

listenbrainz/webserver/static/js/src/Bar.tsx
```diff
@@ -0,0 +1,109 @@
...
+    });
+  };
+
+  render() {
+    const { data, maxValue } = this.props;
+    const { marginLeft } = this.state;
+
+    const leftAlignedTick = (tick: any) => {
+      let { value } = tick;
+      if (value.length > marginLeft / 10) {
```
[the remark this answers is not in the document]
ishaanshah: According to `nivo`'s documentation, `tick` object will always have the `value` property. I had made a syntax error before, which was causing the issue. However, that will also be caught in the unit tests, so I think we should be fine without the check.

listenbrainz/webserver/static/js/src/UserArtists.tsx
ishaanshah: This really boosted the page speed, thanks for the tip :rocket:

listenbrainz/webserver/static/js/src/UserArtists.tsx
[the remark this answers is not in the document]
ishaanshah: Should be fixed with [LB-579](https://tickets.metabrainz.org/browse/LB-579)

listenbrainz/webserver/static/js/src/Bar.tsx
```diff
@@ -0,0 +1,115 @@
...
+    this.setState({
+      marginLeft: window.innerWidth / 5,
+    });
+  };
+
+  render() {
+    const { data, maxValue } = this.props;
+    const { marginLeft } = this.state;
+
+    const leftAlignedTick = (tick: any) => {
```
paramsingh, line 40: Can we fix this `any`?

listenbrainz/webserver/static/js/src/Bar.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from "react";
+import { ResponsiveBar, LabelFormatter } from "@nivo/bar";
+
+export type BarProps = {
+  data: Array<object>;
```
paramsingh, line 5: Is there a way to make this more specific than `Array<object>`?

listenbrainz/webserver/static/js/src/Bar.tsx
```diff
@@ -0,0 +1,115 @@
...
+      );
+    };
+
+    const labelFormatter = (((label: string) => {
+      return (
+        <tspan x={5} textAnchor="start">
+          {label}
+        </tspan>
+      );
+    }) as unknown) as LabelFormatter;
```
paramsingh, line 61: unknown?

listenbrainz/webserver/static/js/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react";
+
+export default class ErrorBoundary extends React.Component<any, any> {
```
paramsingh, line 3: can we remove the any's here?

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    // We don't need to call componentDidMount during "mount" because we are
+    // passing the data manually, so mock the implementation once.
+    jest
+      .spyOn(UserArtists.prototype, "componentDidMount")
+      .mockImplementationOnce((): any => {});
+
+    const wrapper = mount<UserArtists>(<UserArtists {...props} />);
+    const instance = wrapper.instance();
+
+    // eslint-disable-next-line dot-notation
```
paramsingh, line 27: I don't like so many `eslint-disable-next-line`s, can we just make these things public?

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const wrapper = shallow<UserArtists>(<UserArtists {...props} />);
+    const instance = wrapper.instance();
+
+    delete window.location;
+    window.location = {
+      href: "https://foobar/user/bazfoo/artists?page=3",
+    } as any;
+    instance.componentDidMount();
+
+    // eslint-disable-next-line dot-notation
```
paramsingh, line 47: same  with the `eslint-disable-next-line` here

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    expect(instance["currPage"]).toBe(3);
+  });
+
+  it("sets maxListens and totalPages correctly", async () => {
+    const wrapper = shallow<UserArtists>(<UserArtists {...props} />);
+    const instance = wrapper.instance();
+
+    // eslint-disable-next-line dot-notation
```
paramsingh, line 55: same  with the `eslint-disable-next-line` here

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const spy = jest.spyOn(instance["APIService"], "getUserStats");
+    spy.mockImplementation(() => {
+      return Promise.resolve(userArtistsResponse as any);
+    });
+
+    await instance.componentDidMount();
+    expect(spy).toHaveBeenCalledWith("dummyUser", undefined, undefined, 1);
+    // eslint-disable-next-line dot-notation
```
paramsingh, line 63: same  with the `eslint-disable-next-line` here

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    expect(instance["maxListens"]).toBe(385);
+    // eslint-disable-next-line dot-notation
```
paramsingh, line 65: same  with the `eslint-disable-next-line` here

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    expect(instance["totalPages"]).toBe(7);
+  });
+
+  it("calls handlePageChange", async () => {
+    const wrapper = shallow<UserArtists>(<UserArtists {...props} />);
+    const instance = wrapper.instance();
+
+    instance.handlePageChange = jest.fn();
+    // eslint-disable-next-line dot-notation
```
paramsingh, line 74: same  with the `eslint-disable-next-line` here

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    );
+  });
+});
+
+describe("handlePageChange", () => {
+  it("sets state correctly and updates currPage", async () => {
+    const wrapper = shallow<UserArtists>(<UserArtists {...props} />);
+    const instance = wrapper.instance();
+
+    // eslint-disable-next-line dot-notation
```
paramsingh, line 121: same  with the `eslint-disable-next-line` here

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+      return Promise.resolve(userArtistsResponse);
+    });
+    instance.processData = jest.fn().mockImplementationOnce(() => {
+      return userArtistsProcessDataOutput;
+    });
+    await instance.handlePageChange(1);
+
+    expect(instance.processData).toHaveBeenCalledWith(userArtistsResponse, 0);
+    expect(wrapper.state("data")).toEqual(userArtistsProcessDataOutput);
+    // eslint-disable-next-line dot-notation
```
paramsingh, line 133: same  with the `eslint-disable-next-line` here

listenbrainz/webserver/static/js/src/UserArtists.tsx
```diff
@@ -0,0 +1,192 @@
...
+  };
+
+  handleError = (error: Error): void => {
+    // Error Boundaries don't catch errors in async code.
+    // Throwing an error in setState fixes this.
+    // This is a hacky solution but should be fixed with upcoming concurrent mode in React.
+    const { user } = this.props;
+
+    if (
+      error.name === "SyntaxError" &&
```
paramsingh, line 120: could we check the http code instead somehow, this seems risky and something that could break easily?

listenbrainz/webserver/static/js/src/UserArtists.tsx
```diff
@@ -0,0 +1,192 @@
...
+        <div className="row">
+          <div className="col-md-12" style={{ height: "75em" }}>
+            <Bar data={data} maxValue={this.maxListens} />
+          </div>
+        </div>
+        <div className="row">
+          <div className="col-md-12">
+            <ul className="pager">
+              <li className={`previous ${!(prevPage > 0) ? "hidden" : ""}`}>
+                {/* eslint-disable-next-line */}
```
paramsingh, line 153: why do we need this eslint-disable?

listenbrainz/webserver/static/js/src/ErrorBoundary.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const { hasError, error } = this.state;
+
+    if (hasError) {
+      document.title = "Something went wrong - ListenBrainz";
+      return (
+        <div>
+          <h2 className="page-title">{error.name}</h2>
+          <p>{error.message}</p>
+          <p>
+            <a href={window.location.origin + window.location.pathname}>
```
paramsingh, line 26: Maybe we should make this a button and call `window.location.reload()` here instead? Would be easier to test.

listenbrainz/webserver/static/js/src/UserArtists.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    expect(spy).toHaveBeenCalledWith("dummyUser", undefined, 0, 25);
+  });
+});
+
+describe("processData", () => {
+  it("processes data correctly", () => {
+    const wrapper = shallow<UserArtists>(<UserArtists {...props} />);
+    const instance = wrapper.instance();
+
+    expect(instance.processData(userArtistsResponse as any, 0)).toEqual(
```
paramsingh, line 110: as any?

listenbrainz/webserver/static/js/src/UserArtists.tsx
```diff
@@ -0,0 +1,192 @@
...
+                <a onClick={() => this.handlePageChange(prevPage)}>
+                  &larr; Previous
+                </a>
+              </li>
+              <li
+                className={`next ${
+                  !(nextPage <= this.totalPages) ? "hidden" : ""
+                }`}
+              >
+                {/* eslint-disable-next-line */}
```
paramsingh, line 163: why do we have this eslint disable?

listenbrainz/webserver/static/js/src/ErrorBoundary.tsx
paramsingh, line 3: Could we also add tests for this component?

listenbrainz/webserver/static/js/src/Bar.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from "react";
```
paramsingh, line 1: let's also move this and `UserArtist` and related test files to a subdir, maybe `graphs` or `stats`?

listenbrainz/webserver/static/js/src/UserArtists.tsx
```diff
@@ -114,6 +98,42 @@ export default class UserArtists extends React.Component<
...
+          `Statistics for user: ${user.name} have not been calculated yet. Please try again later.`
+        );
+      });
+    }
+
+    this.setState(() => {
+      throw error;
+    });
+  };
+
```
MonkeyDo, line 136: Not sure what's happening here. Are you setting the state to an error? how does that work? Is it something to do with the ErrorBoundary component?

ishaanshah: `Error Boundary` doesn't catch errors in `async` functions. Throwing an error in `setState` fixes this issue. More information about his can be found [here](https://medium.com/trabe/catching-asynchronous-errors-in-react-using-error-boundaries-5e8a5fd7b971)

listenbrainz/webserver/static/js/src/Bar.tsx
ishaanshah: `nivo` itself doesn't provide type definition for the `tick` variable, I will add a definition for it using `console.log` statement although it might change in the future.

listenbrainz/webserver/static/js/src/Bar.tsx
ishaanshah: The definition of `LabelFormatter` expects a function which returns a `string` or a `number`, even though `JSX` elements work. I will file an issue in `nivo` regarding this but until then we will have to keep it as it is.

listenbrainz/webserver/static/js/src/UserArtists.tsx
ishaanshah: Ideally we should not be using anchor tags with `onClick` handler, but changing it to buttons leads to lot of styling issues. It should be fixed with [LB-579](https://tickets.metabrainz.org/browse/LB-579).

listenbrainz/webserver/static/js/src/UserArtists.tsx
ishaanshah: Same reason as above.

PR #839, opened 2020-05-08
Don’t show user token on profile page by default

vansika:
# Problem
This PR addresses [LB-492](https://tickets.metabrainz.org/browse/LB-492).  As of now, user token is displayed on user profile page by default which makes the auth token vulnerable. 

# Solution

Give the option to copy the user token without showing it. Give a separate option to hide and show the user token.

[Screenshot](https://drive.google.com/open?id=1B-Z1X_UX7X4xEiZ16dvOZEsB0RkRTynC)


listenbrainz/webserver/templates/profile/info.html
```diff
@@ -35,13 +35,21 @@ <h3> Reset Last.FM Import timestamp </h3>
...
     <p>
-      If you would like to use an external program to submit data to ListenBrainz you will need the following
-      user token:
+      If you would like to use an external program to submit data to ListenBrainz you will need the user token:
     </p>
 
     <div class="form-inline form-group">
-      <input class="form-control" id="auth-token" type="text" size="36" value="{{ user.auth_token }}" readonly>
+      <p><b>Copy User Token</b>
+      <input class="form-control" id="auth-token" type="hidden" size="36" value="{{ user.auth_token }}" readonly>
```
shivam-kapila, line 43:
```suggestion
      <input class="form-control" id="auth-token" type="password" size="36" value="{{ user.auth_token }}" readonly>
```

listenbrainz/webserver/templates/profile/info.html
```diff
@@ -35,13 +35,21 @@ <h3> Reset Last.FM Import timestamp </h3>
...
       <button class="btn btn-info btn-sm glyphicon glyphicon-copy" id="copy-token" title="Copy token to clipboard"></button>
+      </p>
+    </div>
+
+    <div class="divcontainer">
+      <p><b>User Token</b>
+        <span class="btn btn-info glyphicon glyphicon-dropdown glyphicon-chevron-down" data-toggle="collapse" data-target="#collapsecontent" style="width: 50px"></span>
+        <div class="collapse" id="collapsecontent">{{ user.auth_token }}</div>
+      </p>
```
shivam-kapila, line 52: Instead of this, we can simply add an "eye" icon which just toggles the type of `#copy-token` field from `password` to `text`, to reveal the key. I like [this](https://codepen.io/prasanthmj/pen/KqqBVo) one.

vansika: Ideally, we'd want to have copy button to copy token ( without the need to copy-paste manually) and an option to show/hide. I am not sure if the link serves that purpose?
A token is not a password though, showing it in that sense can be a bit off imo.

shivam-kapila: The copy button is to stay there. I was saying that having a copy button even though nothing is visible there is bad UX.

What I am saying is that we can have 2 buttons
1) Show/Hide
2) Copy

The copy button should copy the user token in both the cases when its plain string or password like string. 

The user can toggle the show/hide button to effectively show or hide the user token.

The user should know "Ah there is a string that I am trying to copy".  Currently the copy button simply copies and its confusing which string its talking about when there is none adjacent string whether text/password like.

But this is just an opinion. Lets wait for @paramsingh to review this.

vansika: Gotcha. Sounds good to me.

paramsingh: I think what @shivam-kapila says makes sense. the UX in the screenshot is a bit non-obvious.

vansika:
Looks like [this](https://drive.google.com/open?id=1zyS8HP-fJHif6VNbeLB4pSY64nX1fFtA) now. I can open a ticket to move this to react.

listenbrainz/webserver/templates/profile/info.html
```diff
@@ -35,14 +35,14 @@ <h3> Reset Last.FM Import timestamp </h3>
...
     </p>
 
-    <div class="form-inline form-group">
-      <input class="form-control" id="auth-token" type="text" size="36" value="{{ user.auth_token }}" readonly>
-      <button class="btn btn-info btn-sm glyphicon glyphicon-copy" id="copy-token" title="Copy token to clipboard"></button>
-    </div>
+    <p><form class="form-inline">
+      <input type="password" class="form-control" id="auth-token" style="width: 400px; height:30px;" value="{{ user.auth_token }}" readonly>
+      <button type="button" class="btn btn-info glyphicon glyphicon-eye-open" id="show-hide-token" style="width: 50px; height:30px;" title="Show user token"></button>
+      <button type="button" class="btn btn-info btn-sm glyphicon glyphicon-copy" id="copy-token" style="width: 50px; height:30px;" title="Copy token to clipboard" ></button>
```
shivam-kapila, line 44:
```suggestion
      <button type="button" class="btn btn-info glyphicon glyphicon-copy" id="copy-token" style="width: 50px; height:30px;" title="Copy token to clipboard" ></button>
```

shivam-kapila: `btn-sm` is not required as we are using inline styling

listenbrainz/webserver/static/js/info.js
```diff
@@ -1,4 +1,29 @@
-document.getElementById("copy-token").addEventListener("click", function() {
-    document.getElementById("auth-token").select();
+document.getElementById("show-hide-token").addEventListener("click", function() {
+    var token = document.getElementById("auth-token");
+    var btn = document.getElementById("show-hide-token");
+
+    if (token.type === 'password'){
```
shivam-kapila, line 5:
```suggestion
    if (token.type === 'password') {
```

listenbrainz/webserver/static/js/info.js
```diff
@@ -1,4 +1,29 @@
...
+        token.type = 'text';
+        btn.className = 'btn btn-info glyphicon glyphicon-eye-close';
+        btn.title = "Hide user token";
+    } else{
```
shivam-kapila, line 9:
```suggestion
    } else {
```

listenbrainz/webserver/static/js/info.js
```diff
@@ -1,4 +1,29 @@
...
+});
+
+function copy(token) {
+    token.select();
     document.execCommand("copy");
+}
+
+document.getElementById("copy-token").addEventListener("click", function() {
+    var token = document.getElementById("auth-token");
+    if (token.type === 'password'){
```
shivam-kapila, line 23:
```suggestion
    if (token.type === 'password') {
```

shivam-kapila:
Overall it's good now. Just a few changes to keep the code stylish :)

paramsingh:
@brainzbot retest this please

PR #841, opened 2020-05-09
Merge develop.sh and spark_develop.sh 

vansika:
In this PR, develop.sh and spark_develop.sh are combined to keep one file for development tasks.


vansika:
We want to automate "namenode format" which should be invoked only once during setup. I am not able to think of any other way except this PR. 

But I think there is a glitch here. If the user wants to issue build command after the setup then it can be fatal since the link between namenode and datanode will break and data will be lost completely. 

One way is `./develop.sh spark format` for which we should have a separate function in develop.sh.

I would love to know if there is any other way which can ensure that the format is issued only once during setup.

develop.sh
```diff
@@ -47,6 +53,15 @@ elif [ "$1" == "npm" ]; then
     npm_install
     exit
 
+elif [ "$1" == "spark" ]; then shift
+    if [ "$1" == "build" ]; then
```
alastair, line 57: if you have no containers, then I'm not sure if this is going to run as expected. You're going to do `docker-compose run`, and docker-compose will see that the container doesn't exist, and build it. then it'll run it. then it'll execute your command a few lines down, and run `docker-compose build` again, building the same containers that you just built.
Additionally, this will format the namenode _every time_ you do a build. Is that something that you want? It might make sense to put this in a separate `init` method that you can call just once, or when you need to wipe your setup and start again

shivam-kapila:
Formatting the namenode is a dangerous thing. We only need to format the namenode

shivam-kapila:
A separate `./develop.sh spark format` can be used as we only need to format namenode once in the beginning. Formatting after that without a cluster ID will format the namenode and arbitrarily assign a new ID thus breaking the link between datanode and namenode. Its not possible to relink them unless we manually edit in the config xml files to update the Cluster ID or wipe out the associated directories in HDFS.

vansika:
I have removed spark commands from HACKING.md since most of them are covered in docs. Regarding Format with cluster ID, I want to make look into it more.

develop.sh
vansika: separate init method which will be called just one in shell script? I am not sure if I follow this :(

develop.sh
```diff
@@ -47,6 +58,15 @@ elif [ "$1" == "npm" ]; then
     npm_install
     exit
 
+elif [ "$1" == "spark" ]; then shift
+    if [ "$1" == 'format' ]; then
```
shivam-kapila, line 62:
```suggestion
    if [ "$2" == 'format' ]; then
```

vansika: `shift` will move the pointer.

paramsingh: Not shifting might be a better idea, it keeps the code easier to understand.

vansika: Keeping uniformity in mind, I think we should retain `shift`  imo.

paramsingh: Uniformity with what?

vansika: With other functions in the script.

paramsingh: The other functions only shift when they need to pass the rest of the args into a different command. I don't have stong opinions on this, happy to keep the shift here, but if I were writing this, I would do the following.

```
if "$1" == spark:
    if "$2" == format:
        format()
    else:
        shift 
        pass the remaining args to docker-compose
````

That way we only change the list when we actually need to change it.

develop.sh
paramsingh: I don't think this is still an issue, there's a different spark format command for now, we'll look into automating that in the future.

PR #842, opened 2020-05-10
LB-495 - Document JSON response format for "/1/validate-token"

ishaanshah:
Adds documentation for the JSON response format for `/1/validate-token` route.

# Problem
The documentation does not provide the format of the JSON response received on querying the above endpoint.

# Solution
Adding the response format solves the problem

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -275,7 +275,7 @@ def validate_token():
     In order to query this endpoint, send a GET request with the token to check
     as the `token` argument (example: /validate-token?token=token-to-check)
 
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/views/api.py
```diff
@@ -275,7 +275,24 @@ def validate_token():
     In order to query this endpoint, send a GET request with the token to check
     as the `token` argument (example: /validate-token?token=token-to-check)
 
-    A JSON response will be returned, with one of two codes.
+    A JSON response will the following format be returned,
```
paramsingh, line 278:
```suggestion
    A JSON response, with the following format, will be returned,
```

PR #844, opened 2020-05-13
LB-574: Avoid huge results for statistics endpoint

ishaanshah:
# Problem
There is no upper limit on number of artists returned when no count is provide. This leads to user having to download JSON payload of bigger size than needed and is a waste of server's resources.

# Solution
Adding a max limit of 100 artists per request and a default of 25, if count parameter is not provided, solves the issue.


listenbrainz/webserver/views/stats_api.py
```diff
@@ -86,7 +92,8 @@ def get_artist(user_name):
         return '', 204
 
     if count is None:
-        count = stats['artist']['count']
+        count = DEFAULT_ITEMS_PER_GET
```
paramsingh, line 95: instead of doing this check here, we could just do `count = _get_non_negative_param('count', default=DEFAULT_ITEMS_PER_GET) above on line 84, where we first retrieve count,

PR #847, opened 2020-05-13
BrainzPlayer page

MonkeyDo:
This POST endpoint is meant to be used with the recommendation engine playground at first.
You can POST it a form with `{listens: "[stringifiedArrayOfListens]"}` and get a playable playlist page as response.

To call it, you'll need an html page that does the POST call for you. Here's an example gist (Adjust the port on line 38 to your configuration):
https://gist.github.com/MonkeyDo/6d2f46a15731a36027fd9e099177dcc9

We require a list of listens-formatted items but there is currently no check of the structure of items passed, which will result in a broken page with no indications (try sending `{listens: "[1,2,3]"}` for example)


shivam-kapila:
@brainzbot restest this please

paramsingh:
Thanks!

PR #849, opened 2020-05-13
Improve spotify track search

MonkeyDo:
# Problem

Fix issue where a track with the same name as album might not appear first in results. Spotify is likely to return the most popular song of the album first instead.

# Solution

Specify `track:` in front of the trackname

I also remove the quotes around the trackname that have no effect I could measure, while possibly preventing to find the song with a slightly different name (I'm thinking "feat.. XYZ", "- Radio Edit" etc.)


MonkeyDo:
https://tickets.metabrainz.org/browse/LB-583

paramsingh:
Thanks!

PR #851, opened 2020-05-14
update model iteration value

vansika:
# Problem
Higher iteration value leads to spark errors

# Solution
reduce the number of iterations while training the data to get the model


paramsingh:
Why not just push to the branch directly?

PR #853, opened 2020-05-15
Add artist/artist_credit relations

mayhem:
This is the first step in adding support for artist-artist and artist_credit-artist_credit relations support for our recommendation efforts.

This PR aims to get the tables generated in a new schema inside a (replicated) MB database. It also has support for making file dumps that can be used as part of the spark CF tools. The get_relations.py is mostly there as an example and a base for a very rudimentary test.

This PR has nothing to do with deploying this in any way -- that will be left for a future PR.

### DIFF
--- a/relations/tests/test_artist_relations.py
+++ b/relations/tests/test_artist_relations.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python3
+
+import psycopg2
+
+from get_relations import get_artist_credit_similarities, get_artist_similarities
+import config
+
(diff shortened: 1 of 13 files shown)


relations/tests/test_artist_relations.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
```
paramsingh, line 1: This test requires a musicbrainz-docker setup, so I've removed it from the travis setup for now.

PR #856, opened 2020-05-16
LB-575: Add support for more time ranges for "Top Artists" (Spark and DB Insert)

ishaanshah:
We currently only show "All Time" statistics for "Top Artists". We should support more time ranges like "Last Week", "Last Month" and "Last Year". This PR adds support for calculation and storage of these stats. Further PRs to update API and front end will be made.


listenbrainz_spark/stats/user/utils.py
```diff
@@ -3,49 +3,10 @@
...
-        artists[row.user_name].append({
-            'artist_name': row.artist_name,
-            'artist_msid': row.artist_msid,
-            'artist_mbids': row.artist_mbids,
-            'listen_count': row.cnt,
-        })
-    print("Query to calculate artist stats processed in %.2f s" % (time.time() - t0))
-    return artists
+def filter_listens(df, from_date, to_date):
+    result = df.filter(df.listened_at.between(from_date, to_date))
```
vansika, line 8: <3

listenbrainz_spark/stats/user/artist.py
```diff
@@ -0,0 +1,100 @@
...
+    listens_df = get_listens(from_date, to_date, LISTENBRAINZ_DATA_DIRECTORY)
+    filtered_df = filter_listens(listens_df, from_date, to_date)
+    filtered_df.createOrReplaceTempView('user_artist_last_week')
+
+    return get_artists('user_artist_last_week')
+
+
+def get_artists_last_month():
+    from_date = adjust_months(datetime.now(), 1)
+    to_date = datetime.now()
```
paramsingh, line 73: instead of using datetime.now here and in the `from_date`, it would be better if we used the timestamp of the latest listen we have on the spark cluster, that way we still calculate something even if the dump import pipeline breaks for some reason. The way it works right now, we will delete old calculated stats for the user and start showing them nothing instead.

listenbrainz/db/stats.py
```diff
@@ -52,11 +53,10 @@ def insert_user_stats(user_id, artists, recordings, releases, artist_count):
...
     """
 
+    artist_range = artists['range']
+    del artists['range']
     artist_stats = {
-        'count': artist_count,
-        'all_time': {
-            'artists': artists,
-        }
+        artist_range: artists
```
paramsingh, line 59: this json should also contain the timestamp for which the stats are. We should be able to easily find out and show that the stats are for the month of July or for the week of May 18.

listenbrainz/db/stats.py
```diff
@@ -52,11 +53,10 @@ def insert_user_stats(user_id, artists, recordings, releases, artist_count):
              artist_count (int): the total number of artists listened to by the user
     """
 
+    artist_range = artists['range']
+    del artists['range']
     artist_stats = {
-        'count': artist_count,
```
paramsingh, line 56: this count is used in the user page, if we're removing this, we should also fix that page. I'm ok with keeping it too.

listenbrainz_spark/stats/user/all.py
```diff
@@ -1,31 +1,58 @@
...
-    table_name = 'stats_user_all'
-    listens_df.createOrReplaceTempView(table_name)
 
+def calculate():
     # calculate and put artist stats into the result
     current_app.logger.debug("Running query...")
-    artist_data = get_artists(table_name)
     messages = []
+
+    current_app.logger.debug("Calculating artist_all_time...")
```
paramsingh, line 12: I think it would be better if we had different commands for the new stats you're adding. Right now, if this breaks, it'll break the all time artist calculations too, which seems risky. It's definitely better to keep them seperate.

listenbrainz/db/stats.py
```diff
@@ -52,11 +53,10 @@ def insert_user_stats(user_id, artists, recordings, releases, artist_count):
              artist_count (int): the total number of artists listened to by the user
     """
 
+    artist_range = artists['range']
+    del artists['range']
```
paramsingh, line 57: modifying the dict in place can lead to problems.

```
# suppose we have a variable called artist_stats
db.insert_user_stats(artist_stats)

# get range for some other use
range = artists['range']
```
This will error out, but it's not obvious to the user of the insert function.

It is a good idea to keep functions pure as much as possible, that is take input -> give output, without modifying the input. We should probably make a copy here.

listenbrainz/spark/handlers.py
```diff
@@ -48,9 +49,12 @@ def handle_user_artist(data):
     if is_new_user_stats_batch():
         notify_user_stats_update()
     current_app.logger.debug("inserting stats for user %s", musicbrainz_id)
-    artists = data['artist_stats']
-    artist_count = data['artist_count']
-    db_stats.insert_user_stats(user['id'], artists, {}, {}, artist_count)
+
+    # delete extra data
+    del data['musicbrainz_id']
```
paramsingh, line 54: again, let's not modify data passed into the function, instead copy the data as you need.

listenbrainz/spark/request_manage.py
```diff
@@ -91,6 +92,35 @@ def request_all_user_stats():
     """
     send_request_to_spark_cluster(_prepare_query_message('stats.user.all'))
 
+
+@cli.command(name="request_last_week_user_artists")
+def request_all_user_stats():
```
paramsingh, line 97: all these functions have the same name, we should change that.

listenbrainz/spark/request_queries.json
```diff
@@ -1,17 +1,37 @@
...
+    "name": "stats.user.artist.last_year",
+    "description": "Artist statistics for all users in the last year",
+    "params": []
+  },
+  "stats.user.artist.all_time": {
+    "name": "stats.user.artist.all_time",
+    "description": "All time artist statistics for all users",
+    "params": []
+  },
+  "stats.user.for_user": {
```
paramsingh, line 27: let's remove this command for now, we're not gonna build it anytime soon, imo.

listenbrainz/spark/request_manage.py
paramsingh: Maybe we should make this just one command

```
python manage.py spark request_user_artists --month
python manage.py spark request_user_artists --year
```
and so on.

listenbrainz/spark/request_queries.json
```diff
@@ -1,17 +1,37 @@
...
-        "description": "Import a new full dump into the spark cluster",
-        "params": []
-    }
+  "stats.user.all": {
+    "name": "stats.user.all",
+    "description": "All user statistics that ListenBrainz calculates for all ListenBrainz users",
+    "params": []
+  },
+  "stats.user.artist.last_week": {
+    "name": "stats.user.artist.last_week",
```
paramsingh, line 8: the `last_week` name could be misleading here, we're actually calculating the stats for the most recent week that we have in the spark cluster. I'd just rename this `week` everywhere and similar for other time frames.

listenbrainz_spark/stats/user/utils.py
```diff
@@ -1,51 +1,31 @@
...
 
-    Args:
-        table: name of the temporary table.
+def get_latest_listen_ts():
+    now = datetime.now()
+    while True:
+        try:
+            df = get_listens(now, now, LISTENBRAINZ_DATA_DIRECTORY)
+            break
+        except:
```
paramsingh, line 17: this `except` catches ALL exceptions, that's bad, because it might catch exceptions that you don't want it to catch, I'd only catch the exception that you care about here.

listenbrainz_spark/stats/user/utils.py
```diff
@@ -1,51 +1,31 @@
...
-                 , artist_name
-                 , artist_msid
-                 , artist_mbids
-                 , count(artist_name) as cnt
-              FROM %s
-          GROUP BY user_name, artist_name, artist_msid, artist_mbids
-          ORDER BY cnt DESC
-        """ % (table))
+    df.createOrReplaceTempView('latest_listen_ts')
+    result = run_query("SELECT MAX(listened_at) as max_date FROM latest_listen_ts")
```
paramsingh, line 21: maybe call it `max_timestamp` or `max_listened_at` instead, clearer and more accurate.

listenbrainz_spark/stats/user/utils.py
```diff
@@ -1,51 +1,31 @@
...
 
-def get_artists(table):
-    """
-    Get artist information (artist_name, artist_msid etc) for every user
-    ordered by listen count (number of times a user has listened to tracks
-    which belong to a particular artist).
 
-    Args:
-        table: name of the temporary table.
+def get_latest_listen_ts():
```
paramsingh, line 11: We should add a docstring to this function specifying what it returns. See this for details: https://github.com/metabrainz/guidelines/blob/master/Python.md#docstrings

listenbrainz_spark/stats/user/utils.py
```diff
@@ -1,51 +1,31 @@
...
-            SELECT user_name
-                 , artist_name
-                 , artist_msid
-                 , artist_mbids
-                 , count(artist_name) as cnt
-              FROM %s
-          GROUP BY user_name, artist_name, artist_msid, artist_mbids
-          ORDER BY cnt DESC
-        """ % (table))
+    df.createOrReplaceTempView('latest_listen_ts')
```
paramsingh, line 20: Should we also delete this view when we're done with it?

listenbrainz_spark/stats/user/artist.py
```diff
@@ -0,0 +1,146 @@
...
+    return artists
+
+
+def get_artists_last_week():
+    current_app.logger.debug("Calculating artist_last_week...")
+
+    date = get_latest_listen_ts()
+
+    # Get date for Monday before "date"
+    to_date = adjust_days(date, date.weekday())
```
paramsingh, line 67: the adjust_days name here is very confusing, can you open a ticket to rename it

paramsingh, line 67: We should move this logic out to a seperate function, that would make it easier to test.

listenbrainz_spark/stats/user/artist.py
```diff
@@ -0,0 +1,146 @@
...
+    from_date = adjust_days(to_date, 7)
+
+    listens_df = get_listens(from_date, to_date, LISTENBRAINZ_DATA_DIRECTORY)
+    filtered_df = filter_listens(listens_df, from_date, to_date)
+    filtered_df.createOrReplaceTempView('user_artists_last_week')
+
+    artist_data = get_artists('user_artists_last_week')
+    messages = create_messages(artist_data, 'user_artists', 'last_week', from_date.timestamp(), to_date.timestamp())
```
paramsingh, line 75: can we use named parameters here, it would make the code much much easier to read.

listenbrainz_spark/stats/user/artist.py
```diff
@@ -0,0 +1,146 @@
...
+
+    artist_data = get_artists('user_artists_all_time')
+    messages = create_messages(artist_data, 'user_artists', 'all_time', from_date.timestamp(), to_date.timestamp())
+
+    current_app.logger.debug("Done!")
+
+    return messages
+
+
+def create_messages(artist_data, _type, _range, from_ts, to_ts):
```
paramsingh, line 133: Let's add a docstring here.

listenbrainz/db/stats.py
```diff
@@ -52,11 +53,10 @@ def insert_user_stats(user_id, artists, recordings, releases, artist_count):
              artist_count (int): the total number of artists listened to by the user
```
paramsingh, line 53: if we're removing the artist_count param, we should fix the docstring as well.

listenbrainz/db/stats.py
```diff
@@ -65,29 +65,27 @@ def insert_user_stats(user_id, artists, recordings, releases, artist_count):
...
     }
 
-
     with db.engine.connect() as connection:
         connection.execute(sqlalchemy.text("""
             INSERT INTO statistics.user (user_id, artist, recording, release)
                  VALUES (:user_id, :artists, :recordings, :releases)
             ON CONFLICT (user_id)
-          DO UPDATE SET artist = :artists,
+          DO UPDATE SET artist = statistics.user.artist || :artists,
```
paramsingh, line 79: nice use of the concatenate operator, can we add tests to make sure that this works as we expect?

listenbrainz_spark/stats/user/artist.py
ishaanshah: Should be fixed with ticket [LB-594](https://tickets.metabrainz.org/browse/LB-594)

listenbrainz_spark/stats/user/utils.py
```diff
@@ -1,51 +1,48 @@
...
+
+
+def get_latest_listen_ts():
+    """ Get the timestamp of the latest timestamp present in spark cluster """
+    now = datetime.now()
+    while True:
+        try:
+            df = get_listens(now, now, LISTENBRAINZ_DATA_DIRECTORY)
+            break
+        except SystemExit:
```
paramsingh, line 18: we should make `get_listens` raise something else, `SystemExit` here makes no sense.

paramsingh:
Chatted on IRC. Moving forward, we should

* add unit tests with mocks for `get_user_artists_last_week` etc, but add a ticket for integration tests.
* Fix name of `_type` and `_range`.

listenbrainz_spark/stats/user/utils.py
ishaanshah: We do not need to do that, because the table gets replaced whenever we call `createOrReplaceTempView` again.

listenbrainz/spark/request_manage.py
```diff
@@ -85,12 +86,29 @@ def send_request_to_spark_cluster(message):
...
-@cli.command(name="request_all_user_stats")
-def request_all_user_stats():
+@cli.command(name="request_user_stats")
+@click.option("--week", is_flag=True, help="Request weekly statistics")
+@click.option("--month", is_flag=True, help="Request monthly statistics")
+@click.option("--week", is_flag=True, help="Request yearly statistics")
+def request_user_stats(week, month, year):
     """ Send a user stats request to the spark cluster
     """
+    if (week):
```
paramsingh, line 96: can we remove these brackets?

ishaanshah: Sorry JS habits :man_facepalming:

listenbrainz/spark/test_request_manage.py
```diff
@@ -27,49 +27,56 @@
 from click.testing import CliRunner
 from listenbrainz.spark import request_manage
 
+
```
ishaanshah, line 30: Sorry for the extra PEP8 changes, this file had too many and bot kept complaining

PR #858, opened 2020-05-16
update recommendation schema to use recording mbid

vansika:


vansika:
@mayhem  though lgtm but do we need any further changes in the schema?

admin/sql/updates/2019-07-09-add-recommendation-table.sql
```diff
@@ -7,7 +7,7 @@ CREATE TYPE cf_recording_type AS ENUM('top_artist', 'similar_artist');
 CREATE TABLE recommendation.cf_recording (
   id                  SERIAL, -- PK
   user_id             INTEGER NOT NULL, --FK to "user".id
-  recording_msid      UUID NOT NULL,
+  recording_mbid      UUID NOT NULL,
```
paramsingh, line 10: This would need a new schema change script which we can apply on production.

vansika: Gosh, I missed it. Thanks :)

vansika:
@mayhem  I think it will be better to have the type of recording_mbid col as JSONB wherein we can insert a list of recording mbids recommended to a user rather than inserting a separate entry for each recommended track in the db.

admin/sql/updates/2019-07-09-add-recommendation-table.sql
paramsingh, line 10: has this schema change already been run in production? Either way, we shouldn't change this here, in case we want to apply all schema changes in the order of their creation.

PR #859, opened 2020-05-16
Generate recommendations to send over lemmy

vansika:
As decided on IRC, recommendation HTML is of little use when the recommendations are shipped to lemmy so the file has been deleted. In this PR recmmend.py has been updated with changes in create_dataframes and candidate_Sets.py


listenbrainz_spark/recommendations/recommend.py
```diff
@@ -32,203 +25,135 @@ def load_model(path):
...
-                        ('xxx', 'xxx', 'xxx', 'xxx', 'xxx', 'xxx', 'xxx')
-                    ]
-                    'similar_artists_recordings' : [
-                        ('xxx', 'xxx', 'xxx', 'xxx', 'xxx', 'xxx', 'xxx'),
-                        ...
-                        ('xxx', 'xxx', 'xxx', 'xxx', 'xxx', 'xxx', 'xxx')
-                        ]
-                }
+            user_recommendations_top_artist: list of recommended recordings of top artist.
+            user_recommendations_top_artist: list of recommended recordings of similar artist.
```
shivam-kapila, line 74:
```suggestion
            user_recommendations_similar_artist: list of recommended recordings of similar artist.
```

vansika: Oops. Thanks <3

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -32,203 +25,135 @@ def load_model(path):
...
+        user_recommendations_top_artist, user_recommendations_similar_artist = get_recommendations_for_user(
+                model, user_id,
+                user_name, recordings_df,
+                top_artists_candidate_set,
+                similar_artists_candidate_set
+        )
+
+        messages.append({
+            'musicbrainz_id': user_name,
+            'type': 'recommendations',
```
paramsingh, line 137: this name can probably be more specific.

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -32,203 +25,135 @@ def load_model(path):
...
+    # active users in the last week/month.
+    # users for whom recommendations will be generated.
+    users_df = top_artists_candidate_set.select('user_id', 'user_name').distinct()
+
+    for row in users_df.collect():
+        user_name = row.user_name
+        user_id = row.user_id
+
+        user_recommendations_top_artist, user_recommendations_similar_artist = get_recommendations_for_user(
+                model, user_id,
```
paramsingh, line 129: indentation seems off here?

listenbrainz_spark/recommendations/recommend.py
vansika: Want to suggest something? I could not think of anything else.

vansika: Want to suggest something? I could not think of anything else.

paramsingh: `cf_recording_recommendations` at least, i assume eventually we'll have other recommendations coming in from the cluster too.

vansika: Thanks.

PR #861, opened 2020-05-17
LB-591: Upgrade postgresql-client to 12 from 9.5

shivam-kapila:
# Problem

After the lastest upgrade to postgresql 12, the postgresql-client was left for upgrade.

# Solution

- Update the _Dockerfile_ and _Dockerfile.test_ files
- Update _develop.sh_ to use URL to log into psql.

# Action

The steps I followed are
1. Copy configs
2. Update dockerfile
3. `./develop.sh build`
4. Remove old image 
5. `./develop.sh psql`

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+POSTGRES_ADMIN_LB_URI="postgresql://postgres:postgres@db/listenbrainz"
+
 if [[ ! -d "docker" ]]; then
     echo "This script must be run from the top level directory of the listenbrainz-server source."
     exit -1
@@ -19,8 +21,7 @@ function invoke_manage {
 
 function open_psql_shell {
     invoke_docker_compose run --rm web psql \
-            -U listenbrainz  \
-            -h db listenbrainz
+        $POSTGRES_ADMIN_LB_URI
 }
 
 function npm_install {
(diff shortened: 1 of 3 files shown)


develop.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+POSTGRES_ADMIN_LB_URI="postgresql://postgres:postgres@db/listenbrainz"
```
paramsingh, line 3: By default, we'd want to log in as the `listenbrainz` user not as `postgres`

develop.sh
```diff
@@ -19,8 +21,7 @@ function invoke_manage {
 
 function open_psql_shell {
     invoke_docker_compose run --rm web psql \
-            -U listenbrainz  \
-            -h db listenbrainz
+        $POSTGRES_ADMIN_LB_URI
```
paramsingh, line 24: can we just do `psql -U listenbrainz -h db` here instead of duplicating the variable from config.py? Is there a way to not make this change at all?

shivam-kapila: If we keep the old setup it will prompt us to add password everytime

paramsingh: Actually, there is `-p`, we could set the PGPASSWORD env variable instead...

develop.sh
shivam-kapila: Oh sorry

develop.sh
paramsingh, line 24: Let's change the variable name here too

shivam-kapila: Oh sorry :/

PR #866, opened 2020-05-19
Add python module to interact with lovehate table

shivam-kapila:
This PR aims to add a python module to perform `insert`, `delete`, `update` and `fetch` queries on the feedback schema.


listenbrainz/db/lovehate.py
```diff
@@ -0,0 +1,138 @@
+import sqlalchemy
+
+from listenbrainz import db
+
+
+def insert_lovehate_record(user_id, recording_msid, score):
```
mayhem, line 6: I suppose at some point we should also change this to insert_feedback_record. And if the module is called feedback, then this could just be called insert(). Pretty simple...

listenbrainz/db/lovehate.py
```diff
@@ -0,0 +1,138 @@
...
+            user_id (int): the row id of the user
+            recording_msid (string): the MessyBrainz ID of the recording
+            score (int): The score associated with the recording (+1/-1 for love/hate respectively)
+    """
+
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            INSERT INTO lovehate (user_id, recording_msid, score)
+                 VALUES (:user_id, :recording_msid, :score)
+            ON CONFLICT (user_id, recording_msid)
```
mayhem, line 20: Yay for the on conflict clause!

listenbrainz/db/lovehate.py
```diff
@@ -0,0 +1,138 @@
...
+             WHERE user_id = :user_id
+               AND recording_msid = :recording_msid
+            """), {
+                'user_id': user_id,
+                'recording_msid': recording_msid,
+            }
+        )
+
+
+def get_user_loved_or_hated_recordings(user_id, score):
```
mayhem, line 50: get_feedback() ?

mayhem: This function and the next feel weird to me -- not sure we ever want to fetch just one. Buy ok, it exists... But the names are confusing -- you have to read the docs before you get what they mean. 

@paramsingh what do you think?

listenbrainz/db/lovehate.py
shivam-kapila: Makes sense. I m already on it

listenbrainz/db/lovehate.py
shivam-kapila: I tried to resolve the names to better names. Perhaps these will be good now

shivam-kapila:
Tests are failing because it's waiting for #868 to merge.

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+
+from listenbrainz import db
+
+
+def insert(user_id, recording_msid, score):
+    """ Inserts a feedback record for a user's loved/hated recording into the database.
+        If the record is already present for the user, the score is updated to the new
+        value passed.
+        Args:
+            user_id (int): the row id of the user
```
alastair, line 11: let's start using python type hints in new code, and slowly start migrating other code as we touch it.
In this case, this is as simple as putting these types in the function definition: `def insert(user_id: int, recording_msid: str, score: int):` and removing it from the docstring. Nothing else needed for now

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+            recording_msid (string): the MessyBrainz ID of the recording
+            score (int): The score associated with the recording (+1/-1 for love/hate respectively)
```
alastair, line 13: where will validation on this value happen? do other numbers have meaning? -10? 0? 20? Do we want to limit the ranges here at all?

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            INSERT INTO recording_feedback (user_id, recording_msid, score)
+                 VALUES (:user_id, :recording_msid, :score)
+            ON CONFLICT (user_id, recording_msid)
```
alastair, line 20: Did we want to consider if a user's feedback for a song changes over time? Maybe they used to like it, but no longer do. Or they like it in the mornings, but not in the evenings, or they only want it when they're in a particular mood.
These are all possible future uses of this data, and I understand if we don't want to support it right now, but perhaps it could be interesting to store multiple feedbacks rather than overwriting them?

listenbrainz/db/feedback.py
shivam-kapila: I was thinking to validate the values at the API endpoint function to prevent unnecessary db calls if the value is not in [-1,01].

listenbrainz/db/feedback.py
shivam-kapila: Sure

listenbrainz/db/feedback.py
shivam-kapila: I am not quite sure about allowing multiple feedbacks. I mean if we want to know how the feedbacks for a recording drift over time, we may like to allow multiple records. 
I would like to know what @mayhem and @paramsingh have to say about this?

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+    """ Format the given entires into a dict.
+        Args:
+            user_id (int): the row id of the user
+            recording_msid (string): the MessyBrainz ID of the recording
+            score (int): The score associated with the recording
+        Returns:
+            A dict formed from the args.
+    """
+    record = {
+        'user_id': user_id,
```
shivam-kapila, line 148: I am somewhat inclined to return the `username` rather than the `id`. @mayhem Please let me know your opinion.

alastair: It depends on what we want to use this data for. If it's to give to an external endpoint for a user to use, it should definitely be a musicbrainz login name. If we need to use this information for internal stuff, it might be better to be an id, so it definitely depends on the final requirement.

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+             WHERE user_id = :user_id
+               AND score = :score
+          ORDER BY created DESC
+            """), {
+                'user_id': user_id,
+                'score': score
+            }
+        )
+        while True:
+            row = result.fetchone()
```
alastair, line 100: Take a look at how we return this kind of data in other places in LB:
https://github.com/metabrainz/listenbrainz-server/blob/b73de5f5/listenbrainz/db/spotify.py#L187
https://github.com/metabrainz/listenbrainz-server/blob/b73de5f5/listenbrainz/db/user.py#L412

2 things:
 * we use `fetchall` when we know that there are many rows to be returned
 * we just use `dict(row)` to turn the row into a dictionary instead of using an external formatting function. Note that an item from a db cursor is both a list _and_ a dictionary. If needed you can use `row[0]` but also `row['user_id']` if needed

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT user_id, recording_msid, score
+              FROM recording_feedback
+          ORDER BY created DESC
+            """), {
+                'user_id': user_id
+            }
+        )
+        while True:
```
paramsingh, line 67: this `while true` is a bit scary to me. I know the loop will eventually break, but I'd still prefer if we just used a for loop through all the records, instead of a `while True` and a `break`

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+            row = result.fetchone()
+            if not row:
+                break
+
+            records.append(_format_recording_feedback_row(row[0], row[1], row[2]))
```
paramsingh, line 72:
```suggestion
            records.append(_format_recording_feedback_row(row['user_id'], row['recording_msid'], row['score']))
```

Something around these lines will probably work too, and is easier to understand. Also, maybe it's a good idea to use named parameters for `format_recording_feedback_row`

listenbrainz/db/feedback.py
paramsingh: ++

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+        result = connection.execute(sqlalchemy.text("""
+            SELECT user_id, recording_msid, score
+              FROM recording_feedback
+             WHERE recording_msid = :recording_msid
+          ORDER BY created DESC
+            """), {
+                'recording_msid': recording_msid
+            }
+        )
+        while True:
```
paramsingh, line 128: same here, let's do it how it's done in other places.

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+            row = result.fetchone()
+            if not row:
+                break
+
+            records.append(_format_recording_feedback_row(row[0], row[1], row[2]))
```
paramsingh, line 133: instead of the integer indices, let's use strings

listenbrainz/db/tests/test_feedback.py
```diff
@@ -0,0 +1,116 @@
...
+    def setUp(self):
+        DatabaseTestCase.setUp(self)
+        self.user = db_user.get_or_create(1, 'recording_feedback_user')
+
+    def insert_test_data(self, user_id, neg_score=False):
+        """ Insert test data into the database """
+
+        if neg_score:
+            # Insert the two records with negative score
+            with open(self.path_to_data_file('recording_feedback_negative_score.json')) as f:
```
paramsingh, line 21: I don't think we need a seperate json file for these tests, I'd prefer if you just kept the dict in code, it's easier to read that way.

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,152 @@
...
+
+    return records
+
+
+def get_feedback_single_type(user_id: int, score: int):
```
paramsingh, line 77:
```suggestion
def get_feedback_with_score(user_id: int, score: int):
```

listenbrainz/db/feedback.py
paramsingh: why not both?

listenbrainz/db/feedback.py
paramsingh: not exactly ideal, but something along these lines would be nice. we have `get_user_by_mbid` in `db/user.py`.

listenbrainz/db/feedback.py
paramsingh: We could store multiple rows and just show the latest one on the listen page. I think storing as much of this data as possible might be a good idea?

listenbrainz/db/feedback.py
shivam-kapila: I don't think we should return the user_id in the case of API endpoints. I guess the class approach can get it solved with the class approach. We can have to methods
1. to_api()
2. to_internal_methods() ("yet to think its name")

paramsingh: right, but this is the database function, just return both here and take the one you want for the API.

shivam-kapila: @paramsingh DB doesn't store user_name though. Makes sense to return only user_id and change it in feedback_api class, when needed?

paramsingh: if you're gonna need the username later, joining in this query and returning here might be a better idea than having to run a different query later. I don't know how this will be used exactly, so I can't say.

listenbrainz/db/feedback.py
mayhem: As per IRC discussion, I think this data is going to be sparse and we shouldn't bother collecting it now. If we need it later, then we can always turn it on. Let's keep the on conflict clause.

listenbrainz/db/lovehate.py
shivam-kapila: Added better names

listenbrainz/db/feedback.py
shivam-kapila: Added a validator check in the class.

listenbrainz/db/feedback.py
shivam-kapila: Standardized the returned object.

listenbrainz/db/feedback.py
shivam-kapila: Updated

listenbrainz/db/feedback.py
shivam-kapila: Removed

listenbrainz/db/feedback.py
shivam-kapila: Added named parameters.

listenbrainz/db/feedback.py
shivam-kapila: Done

listenbrainz/db/feedback.py
shivam-kapila: Done

listenbrainz/db/tests/test_feedback.py
shivam-kapila: Removed.

listenbrainz/db/feedback.py
shivam-kapila: Updated.

listenbrainz/feedback.py
```diff
@@ -0,0 +1,24 @@
...
+    def check_score_is_valid(cls, scr):
+        if scr not in [-1, 0, 1]:
+            raise ValueError('Score can have a value of 1, 0 or -1.')
+        return scr
+
+    @validator('recording_msid')
+    def check_recording_msid_is_valid_uuid(cls, rec_msid):
+        try:
+            rec_msid = uuid.UUID(rec_msid)
+            return rec_msid
```
alastair, line 22: just a thing to be aware of here, this will be a UUID object, not a string. In most cases this probably won't be a problem - postgres will happily accept it as a value, but it does mean that the type definition is incorrect. Also, you might find that some places don't like it.
Maybe better to use `str(rec_msid)` here

shivam-kapila: Makes sense

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,115 @@
...
+        result = connection.execute(sqlalchemy.text("""
+            SELECT user_id, recording_msid, score
+              FROM recording_feedback
+          ORDER BY created DESC
+            """), {
+                'user_id': user_id
+            }
+        )
+        return [Feedback(user_id=row["user_id"], recording_msid=str(row["recording_msid"]), score=row["score"])
+                for row in result.fetchall() if row['user_id'] is not None]
```
alastair, line 66: the definition of the user_id field in the database schema is `NOT NULL`, so this will never be None. Therefore I don't think the `if` is needed.

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,115 @@
...
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT user_id, recording_msid, score
+              FROM recording_feedback
+          ORDER BY created DESC
+            """), {
+                'user_id': user_id
+            }
+        )
+        return [Feedback(user_id=row["user_id"], recording_msid=str(row["recording_msid"]), score=row["score"])
```
alastair, line 65: If pydantic works well for us, we need to make a decision on how to construct these objects. This works fine, but we could also do `[Feedback(**dict(row)) for row in result.fetchall()]`

mayhem:
> I find PRs like these hard to review because it isn't clear how the functions being added here will be used. Maybe next time, it would be better to divide work based on API endpoints.

To be fair, I was the one who suggested this workflow.

listenbrainz/db/feedback.py
shivam-kapila: Yep

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,115 @@
...
+                for row in result.fetchall() if row['user_id'] is not None]
+
+
+def get_feedback_by_user__id_and_score(user_id: int, score: int):
```
alastair, line 69: too many __ here in `user__id`

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,115 @@
...
+    """ Get a list of recording feedbacks of particular type (loved or hated) given by the user.
+        If score is +1 then get loved recordings, if -1 then get hated recordings.
+        Args:
+            user_id: the row ID of the user
+        Returns:
+            A list of user's feedback
```
alastair, line 75: it'd be nice for this to say that it returns a `Feedback` object.

listenbrainz/db/feedback.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+
+    records = []
```
alastair, line 78: this isn't needed here and in a few other methods

listenbrainz/feedback.py
```diff
@@ -0,0 +1,24 @@
+import uuid
+
+from typing import List
+from pydantic import BaseModel, ValidationError, validator
+
+
+class Feedback(BaseModel):
```
alastair, line 7: should be some basic documentation here explaining what a feedback is.
@paramsingh what do you think about location of this? We could either keep it in the module in db associated with reading/writing them, or make a separate `models` package. Thoughts?

listenbrainz/db/feedback.py
shivam-kapila: Updated

listenbrainz/db/feedback.py
shivam-kapila: Oops. Removed

listenbrainz/feedback.py
shivam-kapila: Documentation added.

listenbrainz/db/feedback.py
shivam-kapila: Updated

listenbrainz/db/feedback.py
shivam-kapila: Removed

listenbrainz/feedback.py
paramsingh: hmm, i'm thinking of how we'll import it and I think for now it makes sense to keep in the db module.

PR #867, opened 2020-05-19
Request consumer: cf_recording recommendations from spark cluster

vansika:
This PR aims at requesting recommendations from spark cluster and inserting them in the db

### DIFF
--- a/admin/sql/updates/2020-05-20-change-recommendation-cf-recording-col-type.sql
+++ b/admin/sql/updates/2020-05-20-change-recommendation-cf-recording-col-type.sql
@@ -1,0 +1,8 @@
+BEGIN;
+
+ALTER TABLE recommendation.cf_recording DROP recording_mbid;
+ALTER TABLE recommendation.cf_recording ADD COLUMN recording_mbid JSONB NOT NULL;
+
+ALTER TABLE recommendation.cf_recording ADD CONSTRAINT user_id_unique UNIQUE (user_id);
+
+COMMIT;
--- a/listenbrainz/db/recommendations_cf_recording.py
+++ b/listenbrainz/db/recommendations_cf_recording.py
@@ -1,0 +1,57 @@
+"""This module contains functions to insert and retrieve recommendations
+   generated from Apache Spark into the database.
+"""
+
+# listenbrainz-server - Server for the ListenBrainz project.
+#
+# Copyright (C) 2020 MetaBrainz Foundation Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+
+import sqlalchemy
+
+from listenbrainz import db
+from flask import current_app
+
+def get_timestamp_for_last_recording_recommended():
+    """ Get the time when recommendation_cf_recording table was last updated
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT MAX(created) as created_ts
+              FROM recommendation.cf_recording
+            """
+        ))
+        row = result.fetchone()
+        return row['created_ts'] if row else None
+
+def insert_user_recommendation(user_id, recording_mbid, artist_type=None):
+    """ Insert recommended recording for a user in the db.
+    """
+    if artist_type is None:
+        current_app.logger.critical('artist type for recommendation not specified')
+        return
+
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            INSERT INTO recommendation.cf_recording (user_id, recording_mbid, type)
+                 VALUES (:user_id, :recording_mbid, :artist_type)
+        """), {
+                'user_id': user_id,
+                'recording_mbid': recording_mbid,
+                'artist_type': artist_type,
+        }
+    )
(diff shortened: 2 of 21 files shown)


listenbrainz/db/recommendations_cf_recording.py
```diff
@@ -0,0 +1,57 @@
...
+def insert_user_recommendation(user_id, recording_mbid, artist_type=None):
+    """ Insert recommended recording for a user in the db.
+    """
+    if artist_type is None:
+        current_app.logger.critical('artist type for recommendation not specified')
+        return
+
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            INSERT INTO recommendation.cf_recording (user_id, recording_mbid, type)
```
vansika, line 50: @mayhem I am not really sure how to update the table with every batch of recommendations.

vansika: with JSONB all the recommendation will live in one row which can be updated easily by referring the user_id. In the present scenario, we have a multiple rows for one user id, so either we have to delete the whole data before inserting next batch (which is not promising) or keep past recommendations and just insert the new.

mayhem: Oh! I think I finally understand why you wanted this change. Sorry for that -- yes, putting all the recording_mbis into a JSON is a much better approach, since it is one DB operation, you're right. Let's do that.

vansika: thanks :)

listenbrainz/spark/request_manage.py
```diff
@@ -91,8 +90,37 @@ def request_all_user_stats():
...
+@cli.command(name="request_dataframes")
+def request_dataframes():
+    """ Send the cluster a request to create dataframes.
+    """
+    send_request_to_spark_cluster(_prepare_query_message('cf_recording.recommendations.create_dataframes'))
+
+
+@cli.command(name='request_model')
+def request_model():
+    """ Send the cluster request to train the data.
```
shivam-kapila, line 110: "train the data" sounds somewhat wrong. IMO "train the model" might be better

listenbrainz/spark/request_manage.py
```diff
@@ -91,8 +90,37 @@ def request_all_user_stats():
...
+    """
+    send_request_to_spark_cluster(_prepare_query_message('cf_recording.recommendations.train_model'))
+
+
+@cli.command(name='request_candidate_sets')
+def request_candidate_sets():
+    """ Send the cluster the request to generate candidate sets.
```
shivam-kapila, line 117:
```suggestion
    """ Send the cluster a request to generate candidate sets.
```

listenbrainz/spark/request_manage.py
```diff
@@ -91,8 +90,37 @@ def request_all_user_stats():
...
+    """
+    send_request_to_spark_cluster(_prepare_query_message('cf_recording.recommendations.candidate_sets'))
+
+
+@cli.command(name='request_recommendations')
+def request_recommendations():
+    """ Send the cluster the request to generate recommendations.
```
shivam-kapila, line 124:
```suggestion
    """ Send the cluster a request to generate recommendations.
```

listenbrainz/spark/request_manage.py
```diff
@@ -91,8 +90,37 @@ def request_all_user_stats():
...
+
 @cli.command(name="request_import_full")
 def request_import_new_full_dump():
     """ Send the cluster a request to import a new full data dump
     """
     send_request_to_spark_cluster(_prepare_query_message('import.dump.full'))
+
+
+@cli.command(name="request_dataframes")
+def request_dataframes():
```
paramsingh, line 102: we should have tests for each of these commands in `test_request_manage.py` ideally.

listenbrainz/spark/handlers.py
```diff
@@ -71,3 +97,83 @@ def handle_dump_imported(data):
         from_name='ListenBrainz',
         from_addr='noreply@'+current_app.config['MAIL_FROM_DOMAIN'],
     )
+
+
+def handle_dataframes(data):
```
paramsingh, line 102: we should have tests for this handler.

listenbrainz/spark/handlers.py
```diff
@@ -24,6 +26,31 @@ def is_new_user_stats_batch():
     return datetime.now(timezone.utc) - last_update_ts > timedelta(hours=TIME_TO_CONSIDER_STATS_AS_OLD)
 
 
+def is_new_cf_recording_recommendation_batch():
```
paramsingh, line 29: can we add a test for this?

listenbrainz/spark/handlers.py
```diff
@@ -3,12 +3,14 @@
 """
 import listenbrainz.db.user as db_user
 import listenbrainz.db.stats as db_stats
+import listenbrainz.db.recommendations_cf_recording as db_rec
```
paramsingh, line 6: let's do the full name here instead of db_rec.

listenbrainz/db/recommendations_cf_recording.py
```diff
@@ -0,0 +1,64 @@
...
+        result = connection.execute(sqlalchemy.text("""
+            SELECT MAX(created) as created_ts
+              FROM recommendation.cf_recording
+            """)
+        )
+        row = result.fetchone()
+        return row['created_ts'] if row else None
+
+
+def insert_user_recommendation(user_id, top_artist_recording, similar_artist_recording):
```
paramsingh, line 44: Can we add tests for this function, and document what the top_artist_recording and similar_artist_recording fields look like?

listenbrainz/db/recommendations_cf_recording.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+import ujson
+import sqlalchemy
+
+from listenbrainz import db
+from flask import current_app
+
+
+def get_timestamp_for_last_recording_recommended():
```
paramsingh, line 31: Would like tests for this one too.

vansika:
I was waiting for a heads up to move on to tests. Thanks.

admin/sql/updates/2020-05-20-change-recommendation-cf-recording-col-type.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
```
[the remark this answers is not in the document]
vansika: will keep in mind. thanks :)

paramsingh:
For next time, I would have preferred to merge this when we're ready to deploy.

PR #869, opened 2020-05-20
Add Feedback API endpoint

shivam-kapila:
This PR aims to

- Add APIs to submit and retrieve the feedbacks.
- Add the required API tools.
- Add reference JSON for feedback APIs.
- Add tests for the APIs

### DIFF
--- a/docs/dev/feedback-json.rst
+++ b/docs/dev/feedback-json.rst
@@ -6,8 +6,8 @@ Feedback JSON Documentation
 Submission JSON
 ---------------
 
-To submit a feedback via our API (see: :doc:`api`), ``POST`` a JSON document to
-the ``submit-feedback`` endpoint.
+To submit recording feedback via our API (see: :doc:`api`), ``POST`` a JSON document to
+the ``recording-feedback`` endpoint.
 
 A sample feedback may look like::
 
@@ -34,16 +34,24 @@ Score can have one of these three values:
 Fetching feedback JSON
 ----------------------
 
-The JSON documents returned from our API for user feedback look like the following::
+The JSON documents returned from our API for recording feedback look like the following::
 
-[
-  {
-  "user_id": "-- the MusicBrainz ID of the user --",
-  "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
-  "score": 1
-  },
-  "-- more feedback data here ---"
-]
+{
+   count: 1,
+   feedback: [
+      {
+         "user_id": "-- the MusicBrainz ID of the user --",
+         "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
+         "score": 1
+      },
+      "-- more feedback data here ---"
+   ],
+   offset: 0,
+   total_count: 1
+}
 
+The number of feedback in the document are returned by the top-level ``count`` element. The total number of
+feedback for the user/recording are returned by the top-level ``total_count``. ``offset`` speicifies the
+number of feedback to skip from the beginning, for pagination.  The other element is the ``feedback`` element.
 This is a list which contains the feedback JSON elements having a ``user_id`` the MusicBrainz ID of the user,
 a ``recording_msid`` and a ``score`` key.
--- a/docs/dev/feedback-json.rst
+++ b/docs/dev/feedback-json.rst
@@ -3,17 +3,11 @@
 Feedback JSON Documentation
 ===========================
 
-.. note:: Do not submit copyrighted information in these fields!
-
 Submission JSON
 ---------------
 
 To submit a feedback via our API (see: :doc:`api`), ``POST`` a JSON document to
 the ``submit-feedback`` endpoint.
-
-   - The complete JSON document may not exceed :data:`~webserver.views.api.MAX_FEEDBACK_SIZE` 
-     bytes in size
-
 
 A sample feedback may look like::
 
@@ -42,35 +36,14 @@ Fetching feedback JSON
 
 The JSON documents returned from our API for user feedback look like the following::
 
-    {
-      "user_id": "-- the MusicBrainz ID of the user --",
-      "feedback": [
-        {
-        "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
-        "score": 1
-        },
-        "-- more feedback data here ---"
-      ]
-    }
+[
+  {
+  "user_id": "-- the MusicBrainz ID of the user --",
+  "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
+  "score": 1
+  },
+  "-- more feedback data here ---"
+]
 
-The ``user_id`` element contains the MusicBrainz ID of the user whose feedback are
-being returned. The other element is the ``feedback`` element. This is a list which contains
-the feedback JSON elements having a ``recording_msid`` and a ``score`` key.
-
-The JSON documents returned from our API for recording feedback look like the following::
-
-    {
-      "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
-      "feedback": [
-        {
-        "user_id": "-- the MusicBrainz ID of the user --",
-        "score": 1
-        },
-        "-- more feedback data here ---"
-      ]
-    }
-
-The ``recording_msid`` element contains the MessyBrainz ID of the recording 
-whose feedback are being returned. The other element is the ``feedback`` element. 
-This is a list which contains the feedback JSON elements having a ``user_id`` 
-and a ``score`` key.
+This is a list which contains the feedback JSON elements having a ``user_id`` the MusicBrainz ID of the user,
+a ``recording_msid`` and a ``score`` key.
--- a/docs/dev/feedback-json.rst
+++ b/docs/dev/feedback-json.rst
@@ -9,26 +9,17 @@ Submission JSON
 ---------------
 
 To submit a feedback via our API (see: :doc:`api`), ``POST`` a JSON document to
-the ``submit-feedback`` endpoint. Submit JSON document of the following type:
-
-- ``feedback``: Submit feedback for a recording
-
-   - ``payload`` should contain information about *exactly one* recording feedback
+the ``submit-feedback`` endpoint.
 
    - The complete JSON document may not exceed :data:`~webserver.views.api.MAX_FEEDBACK_SIZE` 
      bytes in size
 
 
-The ``submission_type`` element defines different types of submissions. The element
-is placed at the top-most level of the JSON document. The only other required
-element is the ``payload`` element. This provides an array of feedbacks – the
-payload must contain only one feedback
-
-A sample feedback payload may look like::
+A sample feedback may look like::
 
     {
         "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
-        "score": +1
+        "score": 1
     }
 
 Score can have one of these three values:
@@ -39,69 +30,47 @@ Score can have one of these three values:
 
 - ``-1``: Mark the track as hated
 
-   - Indicates user has marked the recordings as hated
+   - Indicates user has marked the recording as hated
 
 - ``0``: Remove the feedback from the track
 
    - Indicates user wants to remove the feedback (loved or hated) from the recording
 
-A complete submit listen JSON document may look like::
-
-    {
-      "submission_type": "feedback",
-      "payload": [
-        {
-        "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
-        "score": +1
-        }
-      ]
-    }
-
 
 Fetching feedback JSON
 ----------------------
 
-The JSON documents returned from our API for feedbacks given by a user look
-like the following::
+The JSON documents returned from our API for user feedback look like the following::
 
     {
-      "payload": {
-        "count": 25,
-        "user_id": "-- the MusicBrainz ID of the user --",
-        "feedbacks": [
-          {
-          "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
-          "score": +1
-          },
-          "-- more feedback data here ---"
-        ]
-      }
+      "user_id": "-- the MusicBrainz ID of the user --",
+      "feedback": [
+        {
+        "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
+        "score": 1
+        },
+        "-- more feedback data here ---"
+      ]
     }
 
-The number of feedbacks in the document are returned by the top-level ``count``
-element. The ``user_id`` element contains the MusicBrainz ID of the user whose feedbacks are
-being returned. The other element is the ``feedbacks`` element. This is a list which contains
+The ``user_id`` element contains the MusicBrainz ID of the user whose feedback are
+being returned. The other element is the ``feedback`` element. This is a list which contains
 the feedback JSON elements having a ``recording_msid`` and a ``score`` key.
 
-The JSON documents returned from our API for feedbacks given for a recording look
-like the following::
+The JSON documents returned from our API for recording feedback look like the following::
 
     {
-      "payload": {
-        "count": 25,
-        "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
-        "feedbacks": [
-          {
-          "user_id": "-- the MusicBrainz ID of the user --",
-          "score": +1
-          },
-          "-- more feedback data here ---"
-        ]
-      }
+      "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
+      "feedback": [
+        {
+        "user_id": "-- the MusicBrainz ID of the user --",
+        "score": 1
+        },
+        "-- more feedback data here ---"
+      ]
     }
 
-The number of feedbacks in the document are returned by the top-level ``count``
-element. The ``recording_msid`` element contains the MeesyBrainz ID of the recording 
-whose feedbacks are being returned. The other element is the ``feedback`` element. 
+The ``recording_msid`` element contains the MessyBrainz ID of the recording 
+whose feedback are being returned. The other element is the ``feedback`` element. 
 This is a list which contains the feedback JSON elements having a ``user_id`` 
 and a ``score`` key.
--- a/docs/dev/feedback-json.rst
+++ b/docs/dev/feedback-json.rst
@@ -50,8 +50,8 @@ The JSON documents returned from our API for recording feedback look like the fo
    total_count: 1
 }
 
-The number of feedback in the document are returned by the top-level ``count`` element. The total number of
-feedback for the user/recording are returned by the top-level ``total_count``. ``offset`` speicifies the
+The number of feedback items in the document are returned by the top-level ``count`` element. The total number of
+feedback items for the user/recording are returned by the top-level ``total_count``. ``offset`` specifies the
 number of feedback to skip from the beginning, for pagination.  The other element is the ``feedback`` element.
 This is a list which contains the feedback JSON elements having a ``user_id`` the MusicBrainz ID of the user,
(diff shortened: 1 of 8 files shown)


docs/dev/feedback-json.rst
```diff
@@ -0,0 +1,76 @@
...
+
+.. note:: Do not submit copyrighted information in these fields!
+
+Submission JSON
+---------------
+
+To submit a feedback via our API (see: :doc:`api`), ``POST`` a JSON document to
+the ``submit-feedback`` endpoint.
+
+   - The complete JSON document may not exceed :data:`~webserver.views.api.MAX_FEEDBACK_SIZE`
```
mayhem, line 14: This is also not needed since a feedback object is nearly fixed in size.

docs/dev/feedback-json.rst
```diff
@@ -0,0 +1,76 @@
...
+
+
+Fetching feedback JSON
+----------------------
+
+The JSON documents returned from our API for user feedback look like the following::
+
+    {
+      "user_id": "-- the MusicBrainz ID of the user --",
+      "feedback": [
```
mayhem, line 47: There is not need for this extra layer. Just have the recording_msid and score on the same level as user_id

docs/dev/feedback-json.rst
```diff
@@ -0,0 +1,76 @@
...
+
+The ``user_id`` element contains the MusicBrainz ID of the user whose feedback are
+being returned. The other element is the ``feedback`` element. This is a list which contains
+the feedback JSON elements having a ``recording_msid`` and a ``score`` key.
+
+The JSON documents returned from our API for recording feedback look like the following::
+
+    {
+      "recording_msid": "d23f4719-9212-49f0-ad08-ddbfbfc50d6f",
+      "feedback": [
```
mayhem, line 64: obvs make this one match the one above.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,74 @@
...
+                                           APIServiceUnavailable,
+                                           APIUnauthorized)
+from listenbrainz.webserver.rate_limiter import ratelimit
+from listenbrainz.webserver.views.api import _validate_auth_header
+from listenbrainz.webserver.views.api_tools import log_raise_400, DEFAULT_ITEMS_PER_GET, MAX_ITEMS_PER_GET
+from listenbrainz.feedback import Feedback
+from pydantic import ValidationError
+
+#: Maximum overall feedback size in bytes, to prevent egregious spamming.
+MAX_FEEDBACK_SIZE = 10240
```
mayhem, line 19: if we even keep this, we should drastically reduce this to something near the actual size.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,74 @@
...
+@feedback_api_bp.route("submit-feedback", methods=["POST", "OPTIONS"])
+@crossdomain(headers="Authorization, Content-Type")
+@ratelimit()
+def submit_feedback():
+    """
+    Submit recording feedback (love/hate) to the server. A user token (found on  https://listenbrainz.org/profile/ )
+    must be provided in the Authorization header! Each request should contain only one feedback in the payload.
+
+    Feedback should be submitted for tracks when the user has listened to them and they exist in their listen history.
+    If the user hasn't listened to the track or hasn't submitted it to ListenBrainz, it doesn't fully count as a feedback
```
mayhem, line 32: I think we can nuke this part.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,74 @@
...
+
+    :reqheader Authorization: Token <user token>
+    :statuscode 200: feedback accepted.
+    :statuscode 400: invalid JSON sent, see error message for details.
+    :statuscode 401: invalid authorization. See error message for details.
+    :resheader Content-Type: *application/json*
+    """
+    user = _validate_auth_header()
+
+    raw_data = request.get_data()
```
mayhem, line 45: The JSON being posted to the endpoint is simple enough that I think we should use request.json and leave the parsing of json to flask. I think you can drop all the code from here until line 55 and use the built in json support.

docs/dev/feedback-json.rst
```diff
@@ -0,0 +1,76 @@
+.. _feedback-json-doc:
+
+Feedback JSON Documentation
+===========================
+
+.. note:: Do not submit copyrighted information in these fields!
```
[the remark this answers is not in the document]
shivam-kapila: Nuked

docs/dev/feedback-json.rst
shivam-kapila: Removed

docs/dev/feedback-json.rst
shivam-kapila: Simplified

docs/dev/feedback-json.rst
shivam-kapila: Done

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Removed

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Removed

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Removed

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+
+    try:
+        if feedback.score == 0:
+            db_feedback.delete(feedback)
+        else:
+            db_feedback.insert(feedback)
+    except APIServiceUnavailable as e:
+        raise
+    except Exception as e:
+        raise APIInternalServerError("Something went wrong. Please try again.")
```
mayhem, line 63: you should log the error internally (self.log.error or so), and include the complete error message you can get from e.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+    if user is None:
+        raise APINotFound("Cannot find user: %s" % user_name)
+
+    if score is None:
+        feedback = db_feedback.get_feedback_by_user_id(user["id"])
+    else:
+        if score in [-1, 1]:
+            feedback = db_feedback.get_feedback_by_user_id_and_score(user["id"], score)
+        else:
+            log_raise_400("Score can have a value of 1 or -1.", request.args)
```
mayhem, line 95: The request.args seems like a leftover, here. If you intend to print out the value that was sent, you should improve the format string. I'd say take it out.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+    :statuscode 200: Yay, you have data!
+    :resheader Content-Type: *application/json*
+    """
+
+    if not is_valid_uuid(recording_msid):
+        log_raise_400("%s MSID format invalid." % recording_msid)
+
+    feedback = db_feedback.get_feedback_by_recording_msid(recording_msid)
+
+    for i, fb in enumerate(feedback):
```
mayhem, line 121: This is a bad DB access pattern. If a recording has 1000 rows of feedback by 1000 different users, you'll execute 1,000 queries. You should make a list of users and then fetch them all using an SELECT ... FROM table WHERE user_name in ([list]).

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+    except ValidationError as e:
+        log_raise_400("Invalid JSON document submitted: %s" % str(e).replace("\n ", ":").replace("\n", " "),
+                      data)  # str.replace() to tidy up the error message
+
+    try:
+        if feedback.score == 0:
+            db_feedback.delete(feedback)
+        else:
+            db_feedback.insert(feedback)
+    except APIServiceUnavailable as e:
```
paramsingh, line 60: why are we catching an API error from the database?

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+
+    if 'recording_msid' in data and 'score' in data and len(data) > 2:
+        log_raise_400("JSON document may only contain recording_msid and "
+                      "score top level keys", data)
+
+    try:
+        feedback = Feedback(user_id=user["id"], recording_msid=data["recording_msid"], score=data["score"])
+    except ValidationError as e:
+        log_raise_400("Invalid JSON document submitted: %s" % str(e).replace("\n ", ":").replace("\n", " "),
+                      data)  # str.replace() to tidy up the error message
```
paramsingh, line 53: this seems random, do we really need to do this? If yes, we should have a more detailed comment saying what this does.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    if not is_valid_uuid(recording_msid):
+        log_raise_400("%s MSID format invalid." % recording_msid)
+
+    feedback = db_feedback.get_feedback_by_recording_msid(recording_msid)
+
+    for i, fb in enumerate(feedback):
+        fb.user_id = db_user.get(fb.user_id)["musicbrainz_id"]
+        feedback[i] = fb.__dict__
```
paramsingh, line 123: this does not seem the right way to do what you're trying to do. I think you want the dict() method on pydantic models here?

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+    except Exception as e:
+        raise APIInternalServerError("Something went wrong. Please try again.")
+
+    return jsonify({'status': 'ok'})
+
+
+@feedback_api_bp.route("/user/<user_name>/get-feedback", methods=["GET"])
+@crossdomain()
+@ratelimit()
+def get_feedback_for_user(user_name):
```
paramsingh, line 71: Should this endpoint have a limit and offset for pagination?

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+        fb.user_id = user_name
+        feedback[i] = fb.__dict__
+
+    return jsonify(feedback)
+
+
+@feedback_api_bp.route("/recording/<recording_msid>/get-feedback", methods=["GET"])
+@crossdomain()
+@ratelimit()
+def get_feedback_for_recording(recording_msid):
```
paramsingh, line 107: Should this endpoint have a limit and offset for pagination?

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+from listenbrainz.webserver.rate_limiter import ratelimit
+from listenbrainz.webserver.views.api import _validate_auth_header, _parse_int_arg
+from listenbrainz.webserver.views.api_tools import log_raise_400, is_valid_uuid
+from listenbrainz.feedback import Feedback
+from pydantic import ValidationError
+
+feedback_api_bp = Blueprint('feedback_api_v1', __name__)
+
+
+@feedback_api_bp.route("submit-feedback", methods=["POST", "OPTIONS"])
```
paramsingh, line 21: `/1/feedback/submit-feedback`? I'm wondering if this should be recording feedback? Feedback could refer to many things.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,125 @@
...
+        feedback = db_feedback.get_feedback_by_user_id(user["id"])
+    else:
+        if score in [-1, 1]:
+            feedback = db_feedback.get_feedback_by_user_id_and_score(user["id"], score)
+        else:
+            log_raise_400("Score can have a value of 1 or -1.", request.args)
+
+    for i, fb in enumerate(feedback):
+        fb.user_id = user_name
+        feedback[i] = fb.__dict__
```
paramsingh, line 99:
```suggestion
        feedback[i] = fb.dict()
```

docs/dev/feedback-json.rst
```diff
@@ -0,0 +1,49 @@
+.. _feedback-json-doc:
+
+Feedback JSON Documentation
+===========================
+
+Submission JSON
+---------------
+
+To submit a feedback via our API (see: :doc:`api`), ``POST`` a JSON document to
```
paramsingh, line 9:
```suggestion
To submit feedback via our API (see: :doc:`api`), ``POST`` a JSON document to
```

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Added

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: The `request.args` is passed so that it can be logged as an error internally. See [here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/views/api_tools.py#L278)

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Shouldn't be here

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Oops. Improved.

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Improved the comment.

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Updated.

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Added

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Updated to /1/feedback/recording-feedback

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Updated. Thanks

docs/dev/feedback-json.rst
shivam-kapila: Oops. Updated.

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Added.

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Oops. Updated. Thanks.

docs/dev/feedback-json.rst
```diff
@@ -0,0 +1,57 @@
...
+         "score": 1
+      },
+      "-- more feedback data here ---"
+   ],
+   offset: 0,
+   total_count: 1
+}
+
+The number of feedback in the document are returned by the top-level ``count`` element. The total number of
+feedback for the user/recording are returned by the top-level ``total_count``. ``offset`` speicifies the
```
mayhem, line 54: typo "speicifies"

mayhem: Better wording for stupid english: "The total number of feedback _items_ for the"

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,167 @@
...
+        log_raise_400("Invalid JSON document submitted: %s" % str(e).replace("\n ", ":").replace("\n", " "),
+                      data)
+    try:
+        if feedback.score == 0:
+            db_feedback.delete(feedback)
+        else:
+            db_feedback.insert(feedback)
+    except Exception as e:
+        raise APIInternalServerError("Something went wrong. Please try again.")
+        current_app.logger.error("Error while inserting recording feedback: {}".format(e))
```
mayhem, line 63: Better, but the raise will never return, so you need to swap this line with the raise.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,167 @@
...
+def get_feedback_for_user(user_name):
+    """
+    Get feedback given by user ``user_name``. The format for the JSON returned is defined in our :ref:`feedback-json-doc`.
+
+    If the optional argument ``score`` is not given, this endpoint will return all the feedback submitted by the user.
+    Otherwise filters the feedback to be returned by score.
+
+    :param score: Optional, If 1 then returns the loved recordings, if -1 returns hated recordings.
+    :type score: ``int``
+    :param count: Optional, number of feedback to return, Default: :data:`~webserver.views.api.DEFAULT_ITEMS_PER_GET`
```
mayhem, line 80: feedback items

listenbrainz/webserver/views/feedback_api.py
mayhem: Ahhh, right!

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,167 @@
...
+
+    total_count = len(feedback)
+    feedback_list = feedback[offset:(offset+count)]
+
+    user_id_list = [fb.user_id for fb in feedback_list]
+    user_name_list = db_user.get_users_in_order(user_id_list)
+
+    for i, fb in enumerate(feedback_list):
+        fb.user_id = user_name_list[i]["musicbrainz_id"]
+        feedback_list[i] = fb.dict()
```
alastair, line 160: When getting data from the database, it's often a good idea to let the database do as much work as possible. Partially so that you don't make a mistake when joining data together, but also because the database is much more efficient at collecting this data and returning if you request it all in one go.

In this case I see a few things that can be improved:
 * You perform the offset / count in python
 * you perform 2 queries - one for the feedback and one for the users
 * you manually link feedback and users together

I would make the following changes:
* In the database query, use OFFSET ... LIMIT to do the paging. You can pass these parameters into the `get_feedback_by_recording_msid` method.
* Directly join the user table against the feedback table (`select ... blah, u.username from recording_feedback join "user" u on u.id = recording_feedback.user_id`), so that you get the feedback and the username in a single query. I think it makes sense to also include username in the `Feedback` model here. You could include both the user id and the username, making the id required (as it's needed for the database row). In some cases the username could be optional (e.g. when you're making an object to pass it to the insert method).

shivam-kapila: Done in #883

docs/dev/feedback-json.rst
shivam-kapila: Done

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Done

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Done

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -0,0 +1,180 @@
...
+        log_raise_400("%s MSID format invalid." % recording_msid)
+
+    score = _parse_int_arg('score')
+
+    offset = _get_non_negative_param('offset', default=0)
+    count = _get_non_negative_param('count', default=DEFAULT_ITEMS_PER_GET)
+
+    count = min(count, MAX_ITEMS_PER_GET)
+
+    args = {}
```
paramsingh, line 157: This seems hard to understand to me. Could we keep the args as variables and pass them explicitly into the function? That would be much more readable.

paramsingh: Same for the other endpoint

shivam-kapila: Fixed

PR #870, opened 2020-05-20
Add youtube player

MonkeyDo:
# Problem

Music player currently limited to Spotify.

# Solution

This PR adds a mechanism to have multiple music sources and try to play track on each in order if a track isn't found.
The first new datasource is Youtube. Search results are a bit so-so, but it does find more results and increases the playability of playlists

Ignore the name of the branch, that's just me being a dummy.

# Action

Still need to:
- work on tests
- reimplement Spotify embedded iframe (in case LB user isn't connected to/has no Spotify account). This has limited interest, considering it's 30second snippets, so maybe they should be lower in order of datasources.

Future improvements:
- mechanism for user to define preference/order of data sources
- implement more data sources (break the silos !)
- feedback mechanism to confirm/infirm correct track (to send to MsB?)


MonkeyDo:
Currently tests are failing because of a typescript definition issue for react-youtube. There's [an open PR](https://github.com/tjallingt/react-youtube/pull/241) for it but not yet merged/published.

Should I try circumventing that? Maybe make a local copy of the type file ?

Is there a point in reimplementing the Spotify embedded iframe? I feel like it wasn't a great experience, considering we couldn't control it with the player controls (i.e it doesn't play the next song automatically) and it was limited to 30sec previews when not connected to an account.
The youtube source seems to be working fairly well (for users who don't have a Spotify account).

In any case, testing would be appreciated !

paramsingh:
About the typescript error, yeah, I think we should duplicate the types file and add a ticket to track moving back once the fix is released.

paramsingh:
About the spotify iframe, I'd say let's not reimplement and see if users come to us with issues. If no one complains, no one was using it.

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -0,0 +1,403 @@
...
+  handleError: (error: string | Error, title?: string) => void;
+  handleWarning: (message: string | JSX.Element, title?: string) => void;
+  handleSuccess: (message: string | JSX.Element, title?: string) => void;
+  onInvalidateDataSource: (
+    dataSource?: DataSourceTypes,
+    message?: string | JSX.Element
+  ) => void;
+};
+
+type BrainzPlayerProps = {
```
paramsingh, line 35: I wonder if we should consider adding to the props the sources we want brainzplayer to use, in priority order. Maybe not right now though, once we find a usecase.

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -0,0 +1,403 @@
...
+
+export default class BrainzPlayer extends React.Component<
+  BrainzPlayerProps,
+  BrainzPlayerState
+> {
+  spotifyPlayer?: React.RefObject<SpotifyPlayer>;
+  youtubePlayer?: React.RefObject<YoutubePlayer>;
+  dataSources: Array<React.RefObject<DataSourceTypes>> = [];
+
+  firstRun: boolean = true;
```
paramsingh, line 68: A comment here describing why this is needed would be very helpful. Otherwise, we forget. :D

listenbrainz/webserver/static/js/src/BrainzPlayer.test.tsx
```diff
@@ -0,0 +1,271 @@
...
+  newAlert: (
+    alertType: AlertType,
+    title: string,
+    message: string | JSX.Element
+  ) => {},
+  apiService: new APIService("base-uri"),
+};
+
+// Give yourself a two minute break and go listen to this gem
+// https://musicbrainz.org/recording/7fcaf5b3-e682-4ce6-be61-d3bce775a43f
```
paramsingh, line 25: :tada:

listenbrainz/webserver/static/js/src/BrainzPlayer.test.tsx
```diff
@@ -0,0 +1,271 @@
...
+    expect(instance.dataSources).toHaveLength(1);
+    expect(instance.dataSources[0].current).toBeInstanceOf(YoutubePlayer);
+  });
+
+  it("creates a Spotify datasource when passed a spotify user with right permissions", () => {
+    const mockProps = {
+      ...props,
+      spotifyUser: {
+        access_token: "haveyouseenthefnords",
+        permission: "streaming user-read-email user-read-private" as SpotifyPermission,
```
paramsingh, line 56: the SpotifyPermission type only contains individual permissions like `streaming` and `user-read-email`. I think we should change the code so that `spotifyUser.permission` is  a list of `SpotifyPermission`.

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -0,0 +1,403 @@
...
+            onTrackEnd={this.playNextTrack}
+            onTrackNotFound={this.failedToFindTrack}
+            handleError={this.handleError}
+            handleWarning={this.handleWarning}
+            handleSuccess={this.handleSuccess}
+          />
+        </PlaybackControls>
+      </div>
+
+      // // Fallback embedded player
```
paramsingh, line 389: I'm ok with removing this for now.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
```
paramsingh, line 1: Should we just disable this everywhere? I'd like to keep the convention of using camelcase everywhere, but there are some places where it's just not possible (stuff coming from python etc), so maybe we should just get rid of this.

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
MonkeyDo: That's definitely on my mind.
The way I see it, at some point we'll need an overhaul of what is currently the connect-spotify page to allow for other music services we want to connect to.

Along with that, we'll want somewhere in the user profile a list of the services the user is connected to and a way to arrange them by order of preference.
Of course after that the BrainzPlayer will need to be modified to take that order into account.

listenbrainz/webserver/static/js/src/types.d.ts
MonkeyDo: I think it's fine to keep it for the jsx pages. I don't mind changing that file-wide disable into specific blocks disables for the items we know for sure like spotify API and python types

listenbrainz/webserver/static/js/src/BrainzPlayer.test.tsx
MonkeyDo: Yeah, I think that would be a good idea.
Maybe in another PR?
I can do the front-end if you want to do the back-end?

MonkeyDo:
@paramsingh and @mayhem , IMO this is ready to make its way onto test or beta.

listenbrainz/webserver/static/js/src/types.d.ts
paramsingh: Sounds good to me!

paramsingh: Sounds good to me!

listenbrainz/webserver/static/js/src/BrainzPlayer.test.tsx
paramsingh: Sounds great to me! Could you please open a ticket for this, so we don't forget?

PR #871, opened 2020-05-20
Add new cron jobs for weekly, monthly and yearly artist stats

paramsingh:
Also, fix up the observability email thingy, and add the
type of stat to the email.

The cron job schedule isn't that important right now, we're just trying to make sure this stuff is stable, we can fix the timelines later.


paramsingh:
I'm gonna merge this when the build passes to deploy, the cron job schedule is something we're using to test the stability of the pipeline for now, so it isn't very important. If anyone has any suggestions, please feel free to add them and I'll action them.

PR #874, opened 2020-05-21
LB-540: Use a pydantic model for artist stats 

paramsingh:
# Problem

LB-540

The schema of the artist stats is complex and not explicitly documented, making any code changes hard.

# Solution

Add a pydantic model for the stats, meaning that it gets documented.

This is a proposal of how we could do this. I'm still on the fence about it because it makes sharing code between artist, recording and release stats harder. Maybe an idea for this particular case is to just do a jsonschema validation before putting the data into the db. That would document the structure of the data in the column, which is what I really want to do here.

# Action

NA


listenbrainz/db/stats.py
```diff
@@ -120,62 +116,19 @@ def get_user_stats(user_id, columns):
...
-            }
-
-            In general, the `artist` dict will contain all artist related stats
-            calculated for the user in listenbrainz.stats.user, keyed by stat name.
+            UserArtistStatRow: a row of statistics.user but only with artist json, user id and last_updated
     """
-    return get_user_stats(user_id, 'artist')
-
-
-def get_all_user_stats(user_id):
```
paramsingh, line 153: Removed this function because it was dead code not used anywhere, we can bring it back if needed later.

listenbrainz/db/stats.py
```diff
@@ -40,24 +42,18 @@ def get_timestamp_for_last_user_stats_update():
         return row['last_update_ts'] if row else None
 
 
-def insert_user_stats(user_id, artists, recordings, releases):
+def insert_user_stats(user_id: int, artists: UserArtistStatJson, recordings: dict, releases: dict):
```
paramsingh, line 45: We should consider refactoring this into three different functions, this does not seem ideal to me.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -151,3 +147,16 @@ def _is_valid_range(stats_range):
         result (bool): True if given range is valid
     """
     return stats_range in StatisticsRange.__members__
+
+
+def _get_stats_for_range(data: UserArtistStatRow, stats_range: StatisticsRange) -> UserArtistStatForRange:
+    if stats_range == 'all_time':
```
paramsingh, line 153: @alastair any suggestions here? this function seems non ideal, it was just `data['artist'][stats_range]` before this, also, i'm wary of doing this for release and recording as well...

ishaanshah: I may be wrong, but can't we convert `data` to a dictionary and revert back to the original code.
Something like, `return dict(data)['artist'][stats_range]`

ishaanshah:
Overall, this looks very good. Merging #873 with this will be difficult, so I will revert back all files in `listenbrainz` directory and open a subsequent PR for writing the remaining stats into the DB.

paramsingh:
@alastairp, could you take a high-level look at this, very interested in feedback

listenbrainz/db/stats.py
alastair: at the moment different types of stats are in different columns though - any thoughts on this? Would you have a single column with a dict of artist/recording/release counts, or separate tables, or a type column? Do you expect to calculate all types of stats for a user at once, or a single type of stats for all users at once?

listenbrainz/db/stats.py
```diff
@@ -120,62 +116,19 @@ def get_user_stats(user_id, columns):
         return dict(row) if row else None
 
 
-def get_user_artists(user_id):
+def get_user_artists(user_id: int) -> Optional[UserArtistStatRow]:
     """Get top artists for user with given ID.
 
         Args:
             user_id (int): the row ID of the user in the DB
```
alastair, line 123: no longer need types here if we have them in the method signature

listenbrainz/db/stats.py
```diff
@@ -120,62 +116,19 @@ def get_user_stats(user_id, columns):
...
-            {
-                'all_time': all time artist listen counts for the user
-                            calculated by listenbrainz.stats.user.get_top_artists
-                'count': the total number of artists this user has listened to
-                                calculated by listenbrainz.stats.user.get_artist_count
-            }
-
-            In general, the `artist` dict will contain all artist related stats
-            calculated for the user in listenbrainz.stats.user, keyed by stat name.
+            UserArtistStatRow: a row of statistics.user but only with artist json, user id and last_updated
```
alastair, line 126: ... or None if there are no statistics for this user id

listenbrainz/db/stats.py
```diff
@@ -120,62 +116,19 @@ def get_user_stats(user_id, columns):
...
-
-            Similarly, the recording and release dicts contain all recording and release
-            related stats calculated for the user keyed by stat name.
-    """
-
-    return get_user_stats(user_id, 'artist, recording, release')
+    data = _get_user_stats(user_id, 'artist')
+    if not data:
+        return None
+    return UserArtistStatRow(**data)
```
alastair, line 131: normally I would expect to see this in the method that gets the row from the database. I assume that it's not there because one of these rows includes other types of stats too, and we only want the artist stat from this method?

listenbrainz/db/model/stats/user/artist.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class UserArtistStatJson(pydantic.BaseModel):
+    week: Optional[UserArtistStatForRange]
+    year: Optional[UserArtistStatForRange]
+    month: Optional[UserArtistStatForRange]
+    all_time: Optional[UserArtistStatForRange]
+
+
+class UserArtistStatRow(pydantic.BaseModel):
```
alastair, line 36: terminology: maybe remove `Row`? We know it's a single thing, so maybe it doesn't need it.

listenbrainz/db/model/stats/user/artist.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class UserArtistRecord(pydantic.BaseModel):
+    artist_msid: Optional[str]
+    artist_mbids: List[str] = []
+    listen_count: int
+    artist_name: str
+
+
+class UserArtistStatForRange(pydantic.BaseModel):
```
alastair, line 22: terminology: `UserArtistStatRange`. the `for` makes it sound more like a verb (method) rather than a noun (model)

listenbrainz/db/model/stats/user/artist.py
```diff
@@ -0,0 +1,39 @@
...
+    to_ts: float
+    from_ts: float
+    count: int
+    artists: List[UserArtistRecord]
+
+
+class UserArtistStatJson(pydantic.BaseModel):
```
alastair, line 29: terminology: I'm not sure about Json - does this imply too much about the underlying structure of the data? I was going to suggest `UserArtistStatResponse` if this was only used to serialise the data when returning it in the API, but I see that it's used in a few other places in the code too, so maybe that term isn't very good.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -96,29 +91,30 @@ def get_artist(user_name):
     if user is None:
         raise APINotFound("Cannot find user: %s" % user_name)
 
-    stats = db_stats.get_user_artists(user['id'])
-    if stats is None:
+    all_stats: UserArtistStatRow = db_stats.get_user_artists(user['id'])
```
alastair, line 94: is this type needed here if we know the return type of the `get_user_artists` method?

listenbrainz/webserver/views/stats_api.py
alastair: in this case you're passing a `str` into this method instead of a `StatisticsRange`. What if we did validation in the view to ensure that the parameter was a valid value for the enum? Then we could maybe do something like

`getattr(data.artist, stats_range.value)`

because all of the validation is already done. You could also do `getattr(data.artist, stats_range.value, None)` and raise the ApiBadRequest if the value is None

listenbrainz/webserver/views/stats_api.py
```diff
@@ -151,3 +147,16 @@ def _is_valid_range(stats_range):
         result (bool): True if given range is valid
     """
     return stats_range in StatisticsRange.__members__
+
+
+def _get_stats_for_range(data: UserArtistStatRow, stats_range: StatisticsRange) -> UserArtistStatForRange:
```
alastair, line 152: docs/tests?

listenbrainz/db/stats.py
paramsingh: A single type of stat for all users, this has been refactored by @ishaanshah  already.

listenbrainz/webserver/views/stats_api.py
paramsingh: Used getattr in some places, in some places I thought it was better to just be more explicit and duplicate code.

listenbrainz/spark/test_handlers.py
```diff
@@ -30,7 +34,7 @@ def test_handle_user_artist(self, mock_send_mail, mock_new_user_stats, mock_get_
             current_app.config['TESTING'] = False  # set testing to false to check the notifications
             handle_user_artist(data)
 
-        mock_db_insert.assert_called_with(1, {'artists': data['artists']}, {}, {})
+        mock_db_insert.assert_called_once()
```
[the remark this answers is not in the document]
paramsingh: Done.

listenbrainz/db/model/stats/user/artist.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class StatisticsRange(Enum):
+    week = 'week'
+    month = 'month'
+    year = 'year'
+    all_time = 'all_time'
+
+
+class UserArtistRecord(pydantic.BaseModel):
```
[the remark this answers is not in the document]
paramsingh: True, Done!

listenbrainz/db/model/stats/user/artist.py
```diff
@@ -0,0 +1,39 @@
+import pydantic
```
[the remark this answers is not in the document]
paramsingh: Moved to `listenbrainz.db.model`.

listenbrainz/db/model/stats/user/artist.py
```diff
@@ -0,0 +1,39 @@
...
+
+class UserArtistRecord(pydantic.BaseModel):
+    artist_msid: Optional[str]
+    artist_mbids: List[str] = []
+    listen_count: int
+    artist_name: str
+
+
+class UserArtistStatForRange(pydantic.BaseModel):
+    to_ts: float
```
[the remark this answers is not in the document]
paramsingh: I think this was a mistake, moved to int.

listenbrainz/db/stats.py
paramsingh: Done!

listenbrainz/db/stats.py
paramsingh: Yep, that's it.

listenbrainz/db/model/stats/user/artist.py
paramsingh: I think Json is the only correct thing that fits here.

listenbrainz/webserver/views/stats_api.py
paramsingh: Added docs to a similar function after rebase, there are pretty extensive tests testing this codepath already.

PR #875, opened 2020-05-24
send train model window with spark request  to generate dataframes

vansika:
# Problem
As of now, the script that creates dataframe uses config value to fetch listens. If someone wants to change the window (number of days on which data is to be trained), it is a bit inflexible since the config value should be changed on the spark cluster  

# Solution
Send the window with the request from lemmy.


vansika:
@mayhem I think it is ready to be merged

listenbrainz/spark/request_queries.json
```diff
@@ -1,57 +1,52 @@
 {
```
mayhem: Why are there fewer handlers now? I'm generally confused by this commit....

vansika: Right. Have just resolved the merge conflict in light with recent changes to master.

mayhem: ok, @paramsingh  does this look ok?

PR #879, opened 2020-05-28
LB-602: Add "Top Release" (API + Frontend)

ishaanshah:
This PR aims to -
- Rename `Artists` page to `History` page
- Add `Top Release` API endpoint
- Add `Top Release` graph on the `History` page

[UI Demo](https://drive.google.com/file/d/1KO5Kcv3SSYsClpB3OVm5y3VPqc83_vbZ/view)


listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
```diff
@@ -0,0 +1,422 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
paramsingh, line 1: Can we fix this? https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md

ishaanshah: Bootstrap doesn't support `button` tag in `nav-pills` which leads to formatting issues. Also the drop-down CSS is written with `<a>` in  mind, so changing it to `<button>` is not possible. The issue is not that serious and can be fixed with the upcoming redesign, so I would say we can leave it disabled for now.

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
```diff
@@ -0,0 +1,452 @@
...
+      entity,
+      entityCount,
+      currPage,
+      maxListens,
+      totalPages,
+      startDate,
+    } = this.state;
+    const prevPage = currPage - 1;
+    const nextPage = currPage + 1;
+    const month = [
```
MonkeyDo, line 273: This works. I can suggest a more state-of-the-art solution with the internationalization API that could help in the future if we implement other languages:

`startDate.toLocaleString('default', { month: 'long' });` will give you the month name in the user's default language.
`startDate.toLocaleString('en-us', { month: 'long' });` will do the same but in english (what you want here since the rest of the page is in english).


You can also get the year in the same string if you want: `date.toLocaleString('en-us', { month:'long', year: 'numeric' });` will output `May 2020`

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
```diff
@@ -0,0 +1,452 @@
...
+            className="col-md-12"
+            style={{ height: `${(75 / this.ROWS_PER_PAGE) * data.length}em` }}
+          >
+            <Bar data={data} maxValue={maxListens} />
+          </div>
+        </div>
+        <div className="row">
+          <div className="col-xs-12">
+            <ul className="pager">
+              <li className={`previous ${!(prevPage > 0) ? "hidden" : ""}`}>
```
MonkeyDo, line 408: I usually prefer buttons to be visible but disabled, but it's personal preference.
Also, it would mean changing it in other pages with pagination, so maybe for another day.

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
ishaanshah: Didn't know about this, much more elegant than the previous solution :)

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
ishaanshah: Should be fixed with [LB-608](https://tickets.metabrainz.org/browse/LB-608)

ishaanshah:
Thanks for the feedback @MonkeyDo, I have incorporated the changes that you asked.

[Updated UI](https://drive.google.com/open?id=1bcYToFqaW_t08oRYbZahVgkw8BPPyjN-)

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
```diff
@@ -0,0 +1,449 @@
...
+  range: UserEntityAPIRange;
+  entity: Entity;
+  currPage: number;
+  entityCount: number;
+  totalPages: number;
+  maxListens: number;
+  startDate: Date;
+};
+
+export default class UserEntity extends React.Component<
```
paramsingh, line 30: Should we call this `UserEntityGraph` or similar? `UserEntity` is a bit unclear to me.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -84,42 +86,141 @@ def get_artist(user_name):
...
+    """
+
+    stats_range = request.args.get('range', default='all_time')
+    if not _is_valid_range(stats_range):
+        raise APIBadRequest("Invalid range: {}".format(stats_range))
+
+    offset = _get_non_negative_param('offset', default=0)
+    count = _get_non_negative_param('count', default=DEFAULT_ITEMS_PER_GET)
+
+    plural_entity = entity + 's'
```
paramsingh, line 202: This is not ideal, but I guess we'll fix this with the pydantic models

listenbrainz/webserver/views/stats_api.py
```diff
@@ -84,42 +86,141 @@ def get_artist(user_name):
...
-    artist_list = stats['artist'][stats_range]['artists'][offset:count]
+    entity_list = stats[entity][stats_range][plural_entity][offset:count]
 
-    return jsonify({'payload': {
+    return {'payload': {
         "user_id": user_name,
-        "artists": artist_list,
-        "count": len(artist_list),
-        "total_artist_count": total_artist_count,
+        plural_entity: entity_list,
```
paramsingh, line 215: Instead of doing this, could we handle this more explicitly in the actual API view, like return  entity_list as a list and then put that it json inside the view. It'll duplicate a little code but it'll make the json more explicit, and we can change the json of each view individually if we need it.

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
```diff
@@ -0,0 +1,449 @@
...
+  };
+
+  processData = (
+    data: UserArtistsResponse | UserReleasesResponse,
+    page: number,
+    entity?: Entity
+  ): UserEntityData => {
+    if (!entity) {
+      // eslint-disable-next-line no-param-reassign
+      ({ entity } = this.state);
```
paramsingh, line 228: what does this do?

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
```diff
@@ -0,0 +1,449 @@
...
+  startDate: Date;
+};
+
+export default class UserEntity extends React.Component<
+  UserEntityProps,
+  UserEntityState
+> {
+  APIService: APIService;
+
+  ROWS_PER_PAGE = 25; // Number of atists to be shown on each page
```
paramsingh, line 36:
```suggestion
  ROWS_PER_PAGE = 25; // Number of rows to be shown on each page
```

listenbrainz/webserver/errors.py
```diff
@@ -55,46 +61,46 @@ def __init__(self, message, payload=None):
...
-                                                                "for radio")
-    NO_PEAK_RADIO            = LastFMError(code = 24, message = "No Peak Radio - This user is not allowed to listen to " \
-                                                                "radio during peak usage")
-    RADIO_NOT_FOUND          = LastFMError(code = 25, message = "Radio Not Found - Radio station not found")
-    API_KEY_SUSPENDED        = LastFMError(code = 26, message = "API Key Suspended - This application is not allowed to make "
-                                                                "requests to the web services")
-    DEPRECATED               = LastFMError(code = 27, message = "Deprecated - This type of request is no longer supported")
-    RATE_LIMIT_EXCEEDED      = LastFMError(code = 29, message = "Rate Limit Exceded - Your IP has made too many requests in " \
-                                                                "exceeding our API guidelines")
+    DOES_NOT_EXIST = LastFMError(code=1, message="This error does not exist")
```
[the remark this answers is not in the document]
paramsingh: Should be easy to do if you start using `git add -p`

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
paramsingh: Wait, I guess I understand what it does, do we need the parameter from the function at all? Why not just always use the state?

MonkeyDo: I wonder if this call will be an issue: https://github.com/metabrainz/listenbrainz-server/pull/879/files#diff-1812b9221be0ce4e18b25519f801eabcR190
You'd probably have to set the newEntity before calling processData in a callback.

listenbrainz/webserver/static/js/src/stats/__snapshots__/UserEntity.test.tsx.snap
```diff
@@ -15,105 +15,139 @@ exports[`User Artists Page renders 1`] = `
...
+          >
+            <button
+              className="dropdown-togle btn-transparent"
+              data-toggle="dropdown"
+              style={
+                Object {
+                  "fontWeight": "bold",
+                  "textTransform": "capitalize",
+                }
+              }
```
MonkeyDo, line 33: Maybe it would be worth to put this in one of the less stylesheets and use a class

listenbrainz/webserver/errors.py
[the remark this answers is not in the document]
ishaanshah: Sorry for that, I have reverted back the file to remove the extra changes. I will keep the `-p` flag in mind next time

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
ishaanshah: Yes, we need the `newEntity` to be set before calling `processData` in `setState` as MonkeyDo said.

listenbrainz/webserver/static/js/src/stats/__snapshots__/UserEntity.test.tsx.snap
ishaanshah: Done

listenbrainz/webserver/static/js/src/stats/UserEntity.tsx
ishaanshah: Changed to `UserHistory`

PR #881, opened 2020-05-29
Send params with spark request 

vansika:
This PR is a follow up of [#875](https://github.com/metabrainz/listenbrainz-server/pull/875). It is better to send params with spark request then having to change them on spark cluster.


vansika:
Yes, of course. I plan to open a different PR on that.

PR #882, opened 2020-05-29
BrainzPlayer: Automatically detect listen datasource and use youtubeID if available

MonkeyDo:
If origin_url is set on the listen additional info, use Youtube as datasource with the id extracted from the URL.

Also added tests for BrainzPlayer to ensure it detects the datasource, and on each datasource to make sure it can extract the id properly.

When we implement other datasources with recognizable attributes and track IDs, we'll be able to implement the same mechanism.

### DIFF
--- a/listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
+++ b/listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
@@ -1,5 +1,9 @@
 import * as React from "react";
-import { isEqual as _isEqual, get as _get } from "lodash";
+import {
+  isEqual as _isEqual,
+  isString as _isString,
+  get as _get,
+} from "lodash";
 import * as _ from "lodash";
 import PlaybackControls from "./PlaybackControls";
 import APIService from "./APIService";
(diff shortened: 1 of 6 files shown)


listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -186,12 +190,46 @@ export default class BrainzPlayer extends React.Component<
...
+    let selectedDatasourceIndex = datasourceIndex;
     const { onCurrentListenChange } = this.props;
     onCurrentListenChange(listen);
     if (this.firstRun) {
       this.firstRun = false;
     }
-    this.setState({ currentDataSourceIndex: datasourceIndex }, () => {
+
+    /** If available, retreive the service the listen was listened with */
+    const listeningFrom = _get(
```
paramsingh, line 201: Could we put this in a new function? `sortSourcesByListenData` maybe? I think the `playListen` function is getting a bit too big.

MonkeyDo: Addressed in PR #904

PR #883, opened 2020-05-29
Improve feedback module

shivam-kapila:
# Problem

The module did not serve the purpose properly for the APIs.

- We had to make separate access for getting `user_name`
- We were fetching all the records and then applying offset and count at the API level.
- There was a lot of repetitive code.

# Solution

This PR aims to address these issues by:

- Using SQL JOINs to fetch `username` along with feedback records.
- Apply LIMIT and OFFSET in SQL queries.
- Remove repetitive code.

### DIFF
--- a/listenbrainz/db/feedback.py
+++ b/listenbrainz/db/feedback.py
@@ -8,6 +8,7 @@ def insert(feedback: Feedback):
     """ Inserts a feedback record for a user's loved/hated recording into the database.
         If the record is already present for the user, the score is updated to the new
         value passed.
+
         Args:
             feedback: An object of class Feedback
     """
@@ -28,7 +29,8 @@ def insert(feedback: Feedback):
 
 
 def delete(feedback: Feedback):
-    """ Deletes the feedback record for a given recording for the user from the database.
+    """ Deletes the feedback record for a given recording for the user from the database
+
         Args:
             feedback: An object of class Feedback
     """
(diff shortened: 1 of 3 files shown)


listenbrainz/db/feedback.py
```diff
@@ -45,66 +47,111 @@ def delete(feedback: Feedback):
         )
 
 
-def get_feedback_by_user_id(user_id: int):
-    """ Get a list of recording feedbacks given by the user.
+def get_feedback_for_user(user_id: int, limit: int, offset: int, score: int = None):
+    """ Get a list of recording feedbacks given by the user
```
alastair, line 51: You should add something here to explicitly say that they're returned in newest-to-oldest order.

listenbrainz/db/feedback.py
```diff
@@ -45,66 +47,111 @@ def delete(feedback: Feedback):
...
-        )
+        result = connection.execute(sqlalchemy.text(query), args)
         return [Feedback(**dict(row)) for row in result.fetchall()]
 
 
-def get_feedback_by_user_id_and_score(user_id: int, score: int):
-    """ Get a list of recording feedbacks of particular type (loved or hated) given by the user.
-        If score is +1 then get loved recordings, if -1 then get hated recordings.
+def get_feedback_count_for_user(user_id: int):
+    """ Get total number recording feedbacks given by the user
```
alastair, line 84: spelling: total number of recording feedbacks

listenbrainz/db/feedback.py
```diff
@@ -45,66 +47,111 @@ def delete(feedback: Feedback):
...
+
+
+def get_feedback_count_for_recording(recording_msid: str):
+    """ Get total number recording feedbacks for a given recording
+
+        Args:
+            recording_msid: the MessyBrainz ID of the recording
+
+        Returns:
+            The total number recording feedbacks given by the user
```
alastair, line 145: this is wrong, should be recordings?

listenbrainz/db/feedback.py
shivam-kapila: Added.

listenbrainz/db/feedback.py
shivam-kapila: Oops

listenbrainz/db/feedback.py
shivam-kapila: Yeah. Sorry. Corrected.

PR #897, opened 2020-06-01
LB-603: Fix spark crash on calculation of all_time release stats

ishaanshah:
# Problem
The spark cluster crashes when calculating all_time release stats because of memory issues.

# Solution
The `collect` command collects all the data and dumps it into the driver, hence running out of memory. We should use `toLocalIterator` to iterate over the instead of collecting it at once.

# Action
We should test if the updated code works correctly without crashing on production. The flag `--all-time-release` can be used to trigger all_time release stats calculation only.

### DIFF
--- a/listenbrainz/webserver/views/stats_api.py
+++ b/listenbrainz/webserver/views/stats_api.py
@@ -237,7 +237,7 @@ def _process_entity(user_name, stats, stats_range, offset, count, entity):
     count = min(count, MAX_ITEMS_PER_GET)
     try:
         total_entity_count = stats[entity][stats_range]['count']
-    except KeyError:
+    except (TypeError, KeyError):
         raise APINoContent('')
 
     count = count + offset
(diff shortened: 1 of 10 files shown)


listenbrainz/webserver/views/stats_api.py
```diff
@@ -237,7 +237,7 @@ def _process_entity(user_name, stats, stats_range, offset, count, entity):
     count = min(count, MAX_ITEMS_PER_GET)
     try:
         total_entity_count = stats[entity][stats_range]['count']
-    except KeyError:
+    except (TypeError, KeyError):
```
ishaanshah, line 240: I will add a test for this in a subsequent PR

paramsingh:
Running this in prod.

paramsingh:
This is going reasonably well, as far as I can see. There are around 4.8k messages in the rabbitmq queue right now, so the processing worked, i'm gonna wait until it reports completion.

listenbrainz/webserver/views/stats_api.py
ishaanshah: Added in #898

PR #898, opened 2020-06-02
LB-609: Fix TypeError if stats are not calculated

ishaanshah:
# Problem
`APIError` doesn't sound the `Acces-Control-Origin` header, which leads to `CORS` error on frontend.

# Solution
Sending the above mention header should fix the problem.

# Action
We will have to test this after it is deployed to production as the `CORS` issue doesn't come up in local dev environment.
https://listenbrainz.org/user/puneruns/history

### DIFF
--- a/listenbrainz/webserver/errors.py
+++ b/listenbrainz/webserver/errors.py
@@ -4,6 +4,8 @@ import yattag
 import ujson
 import collections
 
+from listenbrainz.webserver.decorators import crossdomain
+
 LastFMError = collections.namedtuple('LastFMError', ['code', 'message'])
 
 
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/errors.py
```diff
@@ -166,7 +167,9 @@ def service_unavailable(error):
 
     @app.errorhandler(APIError)
     def api_error(error):
-        return jsonify(error.to_dict()), error.status_code
+        resp = make_response(jsonify(error.to_dict()))
+        resp.headers['Access-Control-Allow-Origin'] = '*'
```
paramsingh, line 171: the cross-domain decorator should already handle this right? https://github.com/metabrainz/listenbrainz-server/blob/2d47f7b4/listenbrainz/webserver/decorators.py#L7

Any way that we can go through that code path here also?

That way we'll have a single implementation of these headers, leading to less chances of security issues. It also gives us the allowed headers header too.

ishaanshah: Done

PR #900, opened 2020-06-02
LB-619: Improve page navigation on History page

ishaanshah:
# Problem
- If an user directly goes to a page `/user/u/history?page=4&range=all_time&entity=artist`, they see the page flash with the contents of page 1, before being replaced with the contents of page 4. Ideally the page should just show the data of page 4 immediately.
- If an user goes to a page, and then click the next button, the URL updates, and the data updates, but if they click the browser's back button, the URL changes but the data on the page doesn't change.

Jira Ticket - [LB-619](https://tickets.metabrainz.org/browse/LB-619)
# Solution
Using the `popstate` event to restore the data when back or forward button is pressed solves the issue.

### DIFF
--- a/docs/dev/devel-env.rst
+++ b/docs/dev/devel-env.rst
@@ -276,6 +276,13 @@ Also, run the **integration tests** for ListenBrainz.
 When the tests complete, you will see if your changes are valid or not. These tests
 are a helpful way to validate new changes without a lot of work.
 
+.. note::
+
+    If you are working on frontend for statistics and don't wan't to go through the
+    process of generating the statistics locally, you can change the `API_URL` in config
+    to `api.listenbrainz.org`_ so that the statistics are fetched from the production server.
+
+.. _api.listenbrainz.org: https://api.listenbrainz.org
 
 Lint your code
 --------------
--- a/docs/dev/devel-env.rst
+++ b/docs/dev/devel-env.rst
@@ -276,14 +276,6 @@ Also, run the **integration tests** for ListenBrainz.
 When the tests complete, you will see if your changes are valid or not. These tests
 are a helpful way to validate new changes without a lot of work.
 
-.. note::
-
-    If you are working on frontend for statistics and don't wan't to go through the
-    process of generating the statistics locally, you can change the `API_URL` in config
-    to `api.listenbrainz.org`_ so that the statistics are fetched from the production server.
-
-.. _api.listenbrainz.org: https://api.listenbrainz.org
-
 Lint your code
 --------------
 
--- a/docs/dev/devel-env.rst
+++ b/docs/dev/devel-env.rst
@@ -276,6 +276,7 @@ Also, run the **integration tests** for ListenBrainz.
 When the tests complete, you will see if your changes are valid or not. These tests
 are a helpful way to validate new changes without a lot of work.
 
+
 Lint your code
 --------------
 
--- a/listenbrainz/webserver/static/js/src/Loader.tsx
+++ b/listenbrainz/webserver/static/js/src/Loader.tsx
@@ -1,0 +1,25 @@
+/* eslint-disable react/prefer-stateless-function */
+import * as React from "react";
+import { faSpinner } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { IconProp } from "@fortawesome/fontawesome-svg-core";
+
+type withLoaderProps = {
+  isLoading: boolean;
+};
+
+const withLoader = <P extends object>(Component: React.ComponentType<P>) => {
+  return class WithLoader extends React.Component<P & withLoaderProps> {
+    render() {
+      const { isLoading, ...props } = this.props;
+      // TODO: Replace with Bootstrap 4 spinner when we upgrade
+      return isLoading ? (
+        <FontAwesomeIcon icon={faSpinner as IconProp} size="4x" spin />
+      ) : (
+        <Component {...(props as P)} />
+      );
+    }
+  };
+};
+
+export default withLoader;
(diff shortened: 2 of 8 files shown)


listenbrainz/webserver/static/js/src/stats/UserHistory.tsx
```diff
@@ -303,24 +315,24 @@ export default class UserHistory extends React.Component<
...
                 >
-                  {entity}s
+                  {`${entity}s`}
                   <span className="caret" />
                 </button>
                 <ul className="dropdown-menu" role="menu">
                   <li>
                     <a
-                      href="#"
+                      href=""
```
ishaanshah, line 324: I know this is not a good idea but bootstrap supports only `anchor` tags in dropdown-menu and pager, a solution for this can be using `react-bootstrap`, which provides us with the components needed in a `React` friendly way.

MonkeyDo: Does this work without an `href` attribute at all?
The [HTML5 specs](https://www.w3.org/TR/2016/REC-html51-20161101/textlevel-semantics.html#the-a-element) say it should be valid:
>If the a element has no href attribute, then the element represents a placeholder for where a link might otherwise have been placed, if it had been relevant, consisting of just the element’s contents.

ishaanshah: It does work but the cursor switches to an `I-Beam` and `ESlint` throws more errors.

ishaanshah:
> Well done on getting the navigation to work.
> The component seems a bit cleaner too.
> 
> I can't test this locally (no stats), but I'm happy to do a round of testing on beta.

You can change the API URL in your config file to api.listenbrainz.org so that it would work without stats being present.

paramsingh:
>You can change the API URL in your config file to api.listenbrainz.org so that it would work without stats being present.

We should document this somewhere.

listenbrainz/webserver/static/js/src/stats/UserHistory.tsx
```diff
@@ -256,27 +210,85 @@ export default class UserHistory extends React.Component<
...
+          totalPages,
+          maxListens,
+          entityCount,
+        });
+      }
+      const data = await this.getData(page, range, entity);
+      this.setState({
+        data: this.processData(data, page, entity),
+        currPage: page,
+      });
```
MonkeyDo, line 237: I think the weird transition jump that @paramsingh mentioned (https://github.com/metabrainz/listenbrainz-server/pull/900#pullrequestreview-425086819) might be caused by how setState is being called in succession here, considering setState is asynchronous.

I would suggest doing a single call, maybe something like this (pseudocode warning):
```
let initData = {};
if (range !== currRange || entity !== currEntity) {
  initData = await this.getInitData(range, entity);
}
const data = await this.getData(page, range, entity);
this.setState({
  data: this.processData(data, page, entity),
  currPage: page,
  ...initData
});
```

Let me know if that works and fixes the issue.

ishaanshah: Yep, that fixed it, thanks for the tip :smiley:

listenbrainz/webserver/static/js/src/Loader.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const withLoader = <P extends object>(Component: React.ComponentType<P>) => {
+  return class WithLoader extends React.Component<P & withLoaderProps> {
+    render() {
+      const { isLoading, ...props } = this.props;
+      // TODO: Replace with Bootstrap 4 spinner when we upgrade
+      return isLoading ? (
+        <FontAwesomeIcon icon={faSpinner as IconProp} size="4x" spin />
+      ) : (
+        <Component {...(props as P)} />
```
MonkeyDo, line 19: I would do this using `{props.children}` here instead of `<Component … />`.
Then you'll be able to simply invoke your component with:
```
<Loader isLoading={loading}>
    <Bar data={data} maxValue={maxListens} />
</Loader>
```
And your Loader component can be a regular React.Component class like the others instead of a factory, as an added bonus.

Here's a good breakdown of props.children in React: https://codeburst.io/a-quick-intro-to-reacts-props-children-cb3d2fce4891

ishaanshah: Yes, this is simpler, should have thought about it before :D

ishaanshah:
Demo of the loader
https://drive.google.com/file/d/1OEoEta8uNwRYrrHnUPC1anKqHp9lR-y-/view?usp=sharing

shivam-kapila:
@ishaanshah Can you please enable link sharing access for the loader demo.

docs/dev/devel-env.rst
```diff
@@ -276,6 +276,13 @@ Also, run the **integration tests** for ListenBrainz.
 When the tests complete, you will see if your changes are valid or not. These tests
 are a helpful way to validate new changes without a lot of work.
 
+.. note::
```
paramsingh, line 279: Putting this after tests doesn't make sense. Can we just put  it in the update config.py section?

docs/dev/devel-env.rst
```diff
@@ -276,6 +276,13 @@ Also, run the **integration tests** for ListenBrainz.
...
+
+    If you are working on frontend for statistics and don't wan't to go through the
```
paramsingh, line 281: Want* typo

listenbrainz/webserver/static/js/src/Loader.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { faSpinner } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { IconProp } from "@fortawesome/fontawesome-svg-core";
+
+type LoaderProps = {
+  isLoading: boolean;
+};
+
+const Loader: React.SFC<LoaderProps> = (props) => {
+  // eslint-disable-next-line react/prop-types
```
paramsingh, line 11: Do we need this rule? Can we disable it entirely?

ishaanshah: Actually we already have disabled it entirely, https://github.com/metabrainz/listenbrainz-server/blob/master/.eslintrc.js#L27. However it does not seem to work in this case. Some bug in `ESLint` I guess.

PR #901, opened 2020-06-04
LB-617: Fix empty fields in release stats

ishaanshah:
# Problem
The release page shows blank entries for listens not having release name present. We should filter out listens not having the release page to solve the issue.

Example-
https://listenbrainz.org/user/marc2k3/history?page=3&range=month&entity=release
https://api.listenbrainz.org/1/stats/user/marc2k3/releases?offset=50&range=month&count=25

# Solution
Adding a `WHERE` clause for listens not having release name in it fixes the issue


ishaanshah:
> Can we add a test to make sure this doesn't regress?

I have modified the test-data to reflect this

PR #902, opened 2020-06-05
update mapping downloader with change in FTP mapping dir format 

vansika:
As discussed on IRC.

### DIFF
--- a/listenbrainz_spark/config.py.sample
+++ b/listenbrainz_spark/config.py.sample
@@ -42,7 +42,4 @@ FTP_MSID_MBID_DIR = '/pub/musicbrainz/listenbrainz/labs/mappings/msid-mbid-mappi
 FTP_ARTIST_RELATION_DIR = '/pub/musicbrainz/listenbrainz/labs/artist-credit-artist-credit-relations/'
 FTP_LISTENS_DIR = '/pub/musicbrainz/listenbrainz/'
 
-# There are three types of mapping, namely 'msid-mbid-mapping', 'msid-mbid-mapping-with-text' and 'msid-mbid-mapping-with-matchable'
-# The type can be changed to download desired mapping
-# Note: Right now, we only have schema to support 'msid-mbid-mapping-with-matchable'
-MAPPING_TYPE = 'msid-mbid-mapping-with-matchable'
+MAPPING_NAME_PREFIX = 'msid-mbid-mapping-with-matchable'
--- a/listenbrainz_spark/config.py.sample
+++ b/listenbrainz_spark/config.py.sample
@@ -38,6 +38,11 @@ SENTRY_DSN_PUBLIC = ''
 MODEL_ID_PREFIX = 'listenbrainz-recommendation-model'
 
 FTP_SERVER_URI = 'ftp.eu.metabrainz.org'
-FTP_MSID_MBID_DIR = '/pub/musicbrainz/listenbrainz/labs/mappings/msid-mbid-mapping-with-matchable/'
+FTP_MSID_MBID_DIR = '/pub/musicbrainz/listenbrainz/labs/mappings/msid-mbid-mapping/'
 FTP_ARTIST_RELATION_DIR = '/pub/musicbrainz/listenbrainz/labs/artist-credit-artist-credit-relations/'
 FTP_LISTENS_DIR = '/pub/musicbrainz/listenbrainz/'
+
+# There are three types of mapping, namely 'msid-mbid-mapping', 'msid-mbid-mapping-with-text' and 'msid-mbid-mapping-with-matchable'
+# The type can be changed to download desired mapping
+# Note: Right now, we only have schema to support 'msid-mbid-mapping-with-matchable'
+MAPPING_TYPE = 'msid-mbid-mapping-with-matchable'
--- a/listenbrainz_spark/ftp/download.py
+++ b/listenbrainz_spark/ftp/download.py
@@ -1,13 +1,13 @@
+import re
 from time import time
 
 from listenbrainz_spark.ftp import ListenBrainzFTPDownloader
-from listenbrainz_spark.exceptions import DumpNotFoundException
+from listenbrainz_spark.exceptions import DumpNotFoundException, MissingMappingTypeException
 
 from flask import current_app
 
 # mbid_msid_mapping_with_matchable is used.
 # refer to: http://ftp.musicbrainz.org/pub/musicbrainz/listenbrainz/labs/mappings/
-MAPPING_DUMP_ID_POS = 5
 ARTIST_RELATION_DUMP_ID_POS = 5
 
 FULL = 'full'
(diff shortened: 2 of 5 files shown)


vansika:
I think we don't need this [downloader](https://github.com/metabrainz/listenbrainz-server/blob/master/spark_manage.py#L80) func anymore because ideally request to download and upload mapping should be sent from lemmy in light of the workflow we have recently adopted. I want to open a separate PR to move all such functions to [request_manage.py](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/spark/request_manage.py)

vansika:
Have the field names/col names/attributes in updated mapping-matchable changed?
I have only tested mapping downloader and not mapping uploader in this PR since my local machine does not allow uploading ~10GB of data into HDFS. 
 
if the attributes have not changed, I am pretty much sure the uploader script will work but we can test it on the spark cluster since we don't have any critical data in the cluster related to recommendations.

listenbrainz_spark/ftp/download.py
```diff
@@ -94,21 +94,68 @@ def download_spark_dump_and_get_path(self, directory, dump_id, ftp_dump_dir, dum
...
+                mapping: list of mapping dump names of given type in the current working directory.
+        """
+        if mapping_type is None:
+            err_msg = 'Please provide a valid mapping type from the following:' \
+                      '\nmsid-mbid-mapping\nmsid-mbid-mapping-with-text\nmsid-mbid-mapping-matchable\n'
+            raise MissingMappingTypeException(err_msg)
+
+        mapping = list()
+        for mapping_name in dump:
+            mapping_pattern = '{}-\\d+-\\d+(.tar.bz2)$'.format(mapping_type)
```
vansika, line 114: @alastair I really think that the pattern can be improved? Can you have a look?

mayhem:
The columns have not changed -- I did fix the problem with that matchable field that wasn't matchable, but all else remains the same.

listenbrainz_spark/ftp/download.py
```diff
@@ -94,21 +94,68 @@ def download_spark_dump_and_get_path(self, directory, dump_id, ftp_dump_dir, dum
...
+        """ Get latest mapping name.
+
+            Args:
+                mapping: list of mapping dump names.
+
+            Returns:
+                latest_mapping (str): latest mapping dump name.
+        """
+        # sort the mappings on timestamp
+        sorted_mapping = sorted(mapping, key=lambda x: int(x.split('-')[-2] + x.split('-')[-1].split('.')[0]))
```
mayhem, line 135: This is a bit confusing -- you have a good regexp up above -- why not use that to parse the filename? Also, you should sort on the date string, not on the timestamp. The timestamp is only there to provide disambiguation in case we ever produce more than 1 dump in any given day.

listenbrainz_spark/config.py.sample
```diff
@@ -38,6 +38,11 @@ SENTRY_DSN_PUBLIC = ''
...
 FTP_SERVER_URI = 'ftp.eu.metabrainz.org'
-FTP_MSID_MBID_DIR = '/pub/musicbrainz/listenbrainz/labs/mappings/msid-mbid-mapping-with-matchable/'
+FTP_MSID_MBID_DIR = '/pub/musicbrainz/listenbrainz/labs/mappings/msid-mbid-mapping/'
 FTP_ARTIST_RELATION_DIR = '/pub/musicbrainz/listenbrainz/labs/artist-credit-artist-credit-relations/'
 FTP_LISTENS_DIR = '/pub/musicbrainz/listenbrainz/'
+
+# There are three types of mapping, namely 'msid-mbid-mapping', 'msid-mbid-mapping-with-text' and 'msid-mbid-mapping-with-matchable'
+# The type can be changed to download desired mapping
+# Note: Right now, we only have schema to support 'msid-mbid-mapping-with-matchable'
+MAPPING_TYPE = 'msid-mbid-mapping-with-matchable'
```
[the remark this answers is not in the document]
vansika: I am taking out this config option for now.

listenbrainz_spark/ftp/download.py
vansika: ummmm, so if we produce more than one dump in any given day, we would want to download the latest mapping of the two (of greater datetime)?

mayhem: Yes

vansika: So we should sort on date and timestamp both?

mayhem: Yes

vansika: Yeah. Was a bit confused by the original comment. Thanks :))

yvanzo:
Not sure if it applies here, but maybe the directory structure should implement [OTHER-17](https://tickets.metabrainz.org/browse/OTHER-17) as in <http://ftp.musicbrainz.org/pub/musicbrainz/data/json-dumps/>?

listenbrainz_spark/ftp/download.py
```diff
@@ -94,21 +94,69 @@ def download_spark_dump_and_get_path(self, directory, dump_id, ftp_dump_dir, dum
...
+                latest_mapping (str): latest mapping dump name.
+        """
+        # sort the mappings on timestamp
+        def callback(mapping_name):
+            res = re.findall("\\d+", mapping_name)
+            _date = res[0]
+            _time = res[1]
+            return int(_date + _time)
+
+        sorted_mapping = sorted(mapping, key=callback)
```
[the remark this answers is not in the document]
vansika: umm... if there are no mappings available, an error will be raised by self.get_mapping_dump_name and this func would never be called. But can make the change. Looks good :)

listenbrainz_spark/ftp/download.py
```diff
@@ -94,21 +94,69 @@ def download_spark_dump_and_get_path(self, directory, dump_id, ftp_dump_dir, dum
         current_app.logger.info('Done. Total time: {:.2f} sec'.format(time() - t0))
         return dest_path
 
-    def download_msid_mbid_mapping(self, directory, mapping_dump_id=None):
-        """ Download msid_mbid_mapping to dir passed as an argument.
+    def get_mapping_dump_name(self, dump, mapping_name_prefix):
```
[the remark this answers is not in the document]
vansika: cool :)

vansika:
`download_spark_dump_and_get_path` was earlier used for mapping as well as artist relations. Since we have changed the dir format of mappings, the function is not of much use. I have removed it and updated `download_artist_relation` accordingly. The PR now looks good to me.

mayhem:
@yvanzo : It does, but not directly to this project, since this project is not generating this data. Next time I work with the project that does generate this data, I'll all the LATEST file.

listenbrainz_spark/ftp/download.py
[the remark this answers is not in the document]
mayhem: ah, ok.

vansika:
@paramsingh ready for merge.

PR #905, opened 2020-06-06
Soundcloud player

MonkeyDo:
This PR adds a new datasource for SoundCloud listens.

As this is a [js-controlled Iframe widget ](https://developers.soundcloud.com/docs/api/html5-widget) situation, rather than using SoundCloud's [HTTP API](https://developers.soundcloud.com/docs/api/guide), we don't currently have search enabled, so it will only play listens where `track_metadata.additional_info.origin_url` matches a soudcloud URL.

We can think about implementing  the API approach, but that would require implementing an [authorization workflow](https://developers.soundcloud.com/docs/api/guide#authentication) on the server similar to what we do for Spotify tokens. I don't this is necessary, considering SoundCloud doesn't have a regular catalog.

I still need to write tests.


MonkeyDo:
Tests currently missing, will get on it this coming week.

paramsingh:
Thanks @MonkeyDo! Could you assign me to this when you want me to review it?

PR #907, opened 2020-06-08
update candidate sets path to avoid race condition

vansika:
# Problem
Spark creates _temporary files in the parent directory of parquet files while writing to storage and later deletes them. The temporary files created for top_artist and similar_artist [candidate sets](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz_spark/recommendations/candidate_sets.py#L214) are saved inside `/recommendation/candidate_sets`.

While the temporary files are still being accessed for similar_artist parquet, spark deletes them considering them to be temp files for top_artists which leads to error.

# Solution
make a different directory for each parquet. Now the temp files will be stored in
- `/recommendation/candidate_sets/top_artist` for top artist parquet
- `/recommendation/candidate_sets/similar_artist` for similar artist parquet.


vansika:
@paramsingh ready for merge.

PR #912, opened 2020-06-11
Add delete a listen feature

shivam-kapila:
This PR aims to add a feature to delete listen for particular `recording_msid` for the given user.

## Note

This PR is dependent on Timescale and can't be directly merged to master as Influx has some issues in deleting listens.

### DIFF
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -671,6 +671,30 @@ class TimescaleListenStore(ListenStore):
             self.log.error("Cannot delete listens for user: %s" % str(e))
             raise
 
+    def delete_listen(self, listened_at: int, user_name: str, recording_msid: str):
+        """ Delete a particular listen for user with specified MusicBrainz ID.
+        Args:
+            listened_at: The timestamp of the listen
+            user_name: the username of the user
+            recording_msid: the MessyBrainz ID of the recording
+        Raises: TimescaleListenStoreException if unable to delete the listen
+        """
+
+        args = {'listened_at': listened_at, 'user_name': user_name, 'recording_msid': recording_msid}
+        query = """DELETE FROM listen
+                    WHERE listened_at = :listened_at
+                      AND user_name = :user_name
+                      AND data -> 'track_metadata' -> 'additional_info' ->> 'recording_msid' = :recording_msid """
+
+        try:
+            with timescale.engine.connect() as connection:
+                connection.execute(sqlalchemy.text(query), args)
+
+            user_key = "{}{}".format(self.ns + REDIS_TIMESCALE_USER_LISTEN_COUNT, user_name)
+            cache.delete(user_key)
+        except psycopg2.OperationalError as e:
+            self.log.error("Cannot delete listen for user: %s" % str(e))
+            raise TimescaleListenStoreException
 
 class TimescaleListenStoreException(Exception):
     pass
--- a/listenbrainz/webserver/views/user.py
+++ b/listenbrainz/webserver/views/user.py
@@ -4,6 +4,7 @@ import urllib
 import ujson
 import psycopg2
 import datetime
+import time
 
 from flask import Blueprint, render_template, request, url_for, Response, redirect, flash, current_app, jsonify
 from flask_login import current_user, login_required
@@ -15,10 +16,11 @@ from listenbrainz.webserver.decorators import crossdomain
 from listenbrainz.webserver.login import User
 from listenbrainz.webserver.redis_connection import _redis
 from listenbrainz.webserver.timescale_connection import _ts
-from listenbrainz.webserver.views.api_tools import publish_data_to_queue
-import time
+from listenbrainz.webserver.views.api_tools import publish_data_to_queue, log_raise_400, is_valid_uuid
 from datetime import datetime
-from werkzeug.exceptions import NotFound, BadRequest, RequestEntityTooLarge, InternalServerError
+from werkzeug.exceptions import NotFound, BadRequest, RequestEntityTooLarge, ServiceUnavailable, Unauthorized, InternalServerError
+from listenbrainz.webserver.views.stats_api import _get_non_negative_param
+from listenbrainz.listenstore.timescale_listenstore import TimescaleListenStoreException
 from pydantic import ValidationError
 
 LISTENS_PER_PAGE = 25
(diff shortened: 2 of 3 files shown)


listenbrainz/webserver/views/user.py
```diff
@@ -180,6 +181,31 @@ def artists(user_name):
...
+def delete_listen():
+    """ Delete a particular listen from the currently logged-in user's listen history.
+
+    This checks for the correct authorization token and deletes the listen.
+    """
+    if request.form.get('token') and (request.form.get('token') == current_user.auth_token):
+        try:
+            recording_msid = request.form.get('recording_msid')
+            if recording_msid is None:
+                log_raise_400("%s Recording MSID missing.")
```
mayhem, line 195: You're missing the `% recording_msid` here.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -695,6 +695,36 @@ def delete(self, musicbrainz_id):
...
+        try:
+            with timescale.engine.connect() as connection:
+                connection.execute(sqlalchemy.text(query), args)
+
+            listen_count = cache.get(self.ns + TimescaleListenStore.REDIS_TIMESCALE_TOTAL_LISTEN_COUNT, decode=False)
+            if listen_count:
+                cache.set(
+                    self.ns + TimescaleListenStore.REDIS_TIMESCALE_TOTAL_LISTEN_COUNT,
+                    int(listen_count) - 1,
+                    TimescaleListenStore.TOTAL_LISTEN_COUNT_CACHE_TIME,
```
mayhem, line 722: Perhaps this not ideal, perhaps it doesn't matter. But if the listen count just near expiry and we delete a listen and set the decremented listen count for the same duration again the date may be stale for longer. How about just deleting the key from redis and forcing a fetch next time?

listenbrainz/webserver/views/user.py
```diff
@@ -180,6 +181,31 @@ def artists(user_name):
...
+
+            if not is_valid_uuid(recording_msid):
+                log_raise_400("%s MSID format invalid." % recording_msid)
+
+            delete_listen(user_name=current_user.musicbrainz_id, recording_msid=recording_msid)
+        except Exception as e:
```
mayhem, line 201: I think we can do this better -- the exception being caught is very general and we're sending the details of DB exception to the user, which isn't very nice. We should just tell them sorry, we could not delete this please try again later. I personally think you should catch the DB exception in the timescale module and raise something more generic that you can catch here. Make sure to log the error though.
If we can't delete listen, that is our fault so it should be a 500 error, not a 400. (remember 4xx is their fault, 5xx is our fault)

listenbrainz/webserver/views/user.py
```diff
@@ -252,3 +278,18 @@ def delete_listens_history(musicbrainz_id):
     _ts.reset_listen_count(user.musicbrainz_id)
     db_user.reset_latest_import(user.musicbrainz_id)
     db_stats.delete_user_stats(user.id)
+
+
+def delete_listen(musicbrainz_id: str, recording_msid: str):
```
mayhem, line 283: You have two delete_listen functions here. And this one does very little, do we really need it?

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -695,6 +695,36 @@ def delete(self, musicbrainz_id):
...
+            musicbrainz_id: the MusicBrainz ID of the user
+            recording_msid: the MessyBrainz ID of the recording
+
+        Raises: Exception if unable to delete the listen
+        """
+
+        args = {'user_name': musicbrainz_id, 'recording_msid': recording_msid}
+        query = """DELETE FROM listen
+                    WHERE user_name = :user_name
+                      AND recording_msid = :recording_msid"""
```
paramsingh, line 711: this would delete ALL listens with a particular recording msid, is that what we want to do? If so, we should fix the name of the function to make sure that it's clear that it will probably delete more than one listen.

listenbrainz/webserver/views/user.py
shivam-kapila: Actually the %s is redundant. I have removed it.

listenbrainz/listenstore/timescale_listenstore.py
shivam-kapila: From your suggestion, I think we can just delete the key and also decrement the count at frontend on success. If the user reloads the page then the new listen_count will be updated in the cache also. Sounds good to you?

listenbrainz/webserver/views/user.py
shivam-kapila: I raised `TimescaleListenStoreException` and when its caught then return InternalServerError. If any other exception then raise BadRequest. Is it fine now?

listenbrainz/webserver/views/user.py
shivam-kapila: Actually seems unnecessary. So nuked it

listenbrainz/listenstore/timescale_listenstore.py
shivam-kapila: I changed it to delete the listen based on (listened_at, track_name, user_name) since the `listen` table in timescale is now based on track_name rather than recording_feedback as a key. Hopefully, your issue is addressed now.

paramsingh: Why not use recording msid instead of track name?

listenbrainz/listenstore/timescale_listenstore.py
mayhem: OK, that works for me.

listenbrainz/listenstore/timescale_listenstore.py
mayhem: +1

shivam-kapila: Reverted

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -695,6 +695,32 @@ def delete(self, musicbrainz_id):
...
+            recording_msid: the MessyBrainz ID of the recording
+
+        Raises: TimescaleListenStoreException if unable to delete the listen
+        """
+
+        args = {'listened_at': listened_at, 'user_name': user_name, 'recording_msid': recording_msid}
+        query = """DELETE FROM listen
+                    WHERE listened_at = :listened_at
+                      AND user_name = :user_name
+                      AND data -> 'track_metadata' -> 'additional_info' ->> 'recording_msid' = :recording_msid """
```
paramsingh, line 713: Can we add an index for this? Otherwise it'll probably be very slow

paramsingh: If possible, otherwise I'm happy to merge this.

shivam-kapila: `CREATE INDEX recording_msid_ndx_listen ON listen (data -> 'track_metadata' -> 'additional_info' ->> 'recording_msid');`
Does this solve the issue

mayhem: As per our IRC discussion indexes are not needed.

shivam-kapila: Upon some discussion and confusion on IRC it was noticed that index isn't much of use here as we already have an index on `user_name, listened_at`. So we think its not required as of now.

listenbrainz/webserver/views/user.py
shivam-kapila: 500 for general exceptions
503 for all Timescale exceptions

mayhem: yep

mayhem:
We should fix the merge conflict and the merge this code now that we can.

PR #914, opened 2020-06-13
send mapping and artist relation import request to spark cluster

vansika:
Rn, there is no way to send a request to import mapping and artist relation into the cluster. This PR resolves the issue.


vansika:
again I have not tested this fully on my local machine. Can be tested on the cluster.

listenbrainz/spark/handlers.py
```diff
@@ -203,3 +203,38 @@ def handle_recommendations(data):
                                                                similar_artist_recording_mbids)
 
     current_app.logger.debug("recommendation for {} inserted".format(musicbrainz_id))
+
+
+def handle_mapping(data):
```
mayhem, line 208: this name does not reflect what the function does

listenbrainz/spark/handlers.py
```diff
@@ -203,3 +203,38 @@ def handle_recommendations(data):
...
+    send_mail(
+        subject='MSID MBID mapping has been imported into the Spark cluster',
+        text=render_template('emails/mapping_import_notification.txt', mapping_name=mapping_name, time=import_completion_time),
+        recipients=['listenbrainz-observability@metabrainz.org'],
+        from_name='ListenBrainz',
+        from_addr='noreply@'+current_app.config['MAIL_FROM_DOMAIN'],
+    )
+
+
+def handle_artist_relation(data):
```
mayhem, line 225: same here

listenbrainz/spark/handlers.py
vansika: handle_mapping_imported?

mayhem: Your doc string clearly names the key function of this method. Why is that not part of the name?

vansika: `notify_mapping_import` ? 

Otherwise you tell me one :p

mayhem: Yep, that works!

vansika:
@paramsingh ready for merge

PR #917, opened 2020-06-15
LB-621: Treat empty string as NULL for optional fields

ishaanshah:
# Problem
Some clients send empty strings for fields not present in song metadata while some send NULL. This creates duplicates in the statistics for the user.

# Solution
We should treat empty strings as NULL so as to fix this issue.


listenbrainz_spark/stats/user/tests/test_release.py
```diff
@@ -51,10 +51,10 @@ def test_get_releases(self):
             if entry['release_name'] != '':
                 expected[entry['user_name']].append({
                     'release_name': entry['release_name'],
-                    'release_msid': entry['release_msid'],
-                    'release_mbid': entry['release_mbid'],
+                    'release_msid': entry['release_msid'] or None,
```
paramsingh, line 54: Nitty but entry.get for the same thing

alastair: I don't think this is the same thing right? Because this will fail with a KeyError if the key doesn't exist, and will return None if `entry['release_msid']` is falsy (empty list, empty dict, empty string, 0), whereas `get` will return None if the key doesn't exist.

PR #918, opened 2020-06-15
Create dumps for recording_feedback

shivam-kapila:
This PR adds the `recording_feedback` table to the list of PUBLIC_TABLES so that it can be dumped into the LB public dumps

### DIFF
--- a/listenbrainz/db/dump.py
+++ b/listenbrainz/db/dump.py
@@ -285,7 +285,7 @@ def create_private_dump(location, dump_time, threads=DUMP_DEFAULT_THREAD_COUNT):
 
 def create_public_dump(location, dump_time, threads=DUMP_DEFAULT_THREAD_COUNT):
     """ Create postgres database dump for statistics and user info in db.
-        This includes a sanitized dump of the "user"' and recording_feedback table and dumps of all tables
+        This includes a sanitized dump of the "user" table and dumps of all tables
         in the statistics schema:
             statistics.user
             statistics.artist
--- a/listenbrainz/db/dump.py
+++ b/listenbrainz/db/dump.py
@@ -90,6 +90,13 @@ PUBLIC_TABLES = {
         'listen_count',
         'last_updated',
     ),
+    'recording_feedback': (
+        'id',
+        'user_id',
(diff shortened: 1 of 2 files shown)


shivam-kapila:
I am not sure about the tests that can be possible. I don't think we have any table based tests for dumps.

listenbrainz/db/dump.py
```diff
@@ -278,7 +285,7 @@ def create_private_dump(location, dump_time, threads=DUMP_DEFAULT_THREAD_COUNT):
 
 def create_public_dump(location, dump_time, threads=DUMP_DEFAULT_THREAD_COUNT):
     """ Create postgres database dump for statistics and user info in db.
-        This includes a sanitized dump of the "user"' table and dumps of all tables
+        This includes a sanitized dump of the "user"' and recording_feedback table and dumps of all tables
```
mayhem, line 288: The recording_feedback table is included in "dumps of all tables". This addition makes the comment read poorly.

shivam-kapila: Not sure I get you correctly. It reads _dumps of all tables in statistics schema_ or I misunderstood this?

paramsingh:
>I don't think we have any table based tests for dumps.

You could create a dump and import it back. We do something similar for listens in influx. Take a look at the influx listen store tests

listenbrainz/db/dump.py
mayhem: Do you really need to add "and recording_feedback" here?

shivam-kapila: I don't feel strongly about it. I added it to specify that we also dump this table in the public dump. I can remove that if it makes the comment awkward.

mayhem: Let's do that. No need to highlight this table in particular.

shivam-kapila: Okay

shivam-kapila: Okay

shivam-kapila:
@paramsingh I have added a test for recording_feedback

PR #922, opened 2020-06-18
use parameter object to handle function arguments

vansika:
# Problem
rn, too many args are being passed to functions in recommend.py ([func1](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz_spark/recommendations/recommend.py#L59), [func2](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz_spark/recommendations/recommend.py#L106), etc.)

# Solution
Create a class to handle function params.


vansika:
too many arguments (5-6) are being passed to functions in recommend.py. I have created a class to create a parameter object and use it wherever necessary, now we only have to pass the parameter object and not too many arguments.

PR #927, opened 2020-06-23
Add API endpoint for fetching recommendations

vansika:
As discussed on IRC.

### DIFF
--- a/listenbrainz/webserver/views/recommendations_cf_recording_api.py
+++ b/listenbrainz/webserver/views/recommendations_cf_recording_api.py
@@ -1,0 +1,135 @@
+import listenbrainz.db.user as db_user
+import listenbrainz.db.recommendations_cf_recording as db_recommendations_cf_recording
+
+from listenbrainz.webserver.errors import APIBadRequest, APINotFound
+
+from flask import Blueprint, jsonify, request
+from listenbrainz.webserver.decorators import crossdomain
+from listenbrainz.webserver.rate_limiter import ratelimit
+
+recommendations_cf_recording_api_bp = Blueprint('recommendations_cf_recording_v1', __name__)
+
+
+@recommendations_cf_recording_api_bp.route("/user/<user_name>/artist")
+@crossdomain()
+@ratelimit()
+def get_recommendations(user_name):
+    """ Get recommendations for user ``user_name``.
+
+        A sample response from the endpoint may look like::
+
+        {
+          "payload": {
+
+            "last_updated": "Sat, 06 Jun 2020 08:18:50 GMT",
+
+            "<artist_type>": {
+
+              "recording_mbids": [
+                "526bd613-fddd-4bd6-9137-ab709ac74cab",
+                "a6081bc1-2a76-4984-b21f-38bc3dcca3a5",
+                "a6ad0205-6e96-416d-a4e8-edd1773dac09",
+                "d8783d03-8a3b-4269-8261-00709d2cfee8"
+              ]
+            },
+
+            "user_name": "unclejohn69"
+          }
+        }
+
+        <artist_type>: 'top' or 'similar'
+
+        :statuscode 200: Successful query, you have data!
+        :statuscode 400: Bad request, check ``response['error']`` for more details
+        :statuscode 404: User not found or recommendations not generated
+    """
+    user = db_user.get_by_mb_id(user_name)
+    if user is None:
+        raise APINotFound("Cannot find user: {}".format(user_name))
+
+    artist_type = request.args.get('type')
+    if not _is_valid_artist(artist_type):
+        raise APIBadRequest("Invalid artist type: {}".format(artist_type))
+
+    recommendations = db_recommendations_cf_recording.get_user_recommendation(user['id'])
+
+    if recommendations is None:
+        err_msg = 'Oops! Looks like {} doesn\'t have recent listening history.' \
+                  ' Recommendations not generated :('.format(user_name)
+        raise APINotFound(err_msg)
+
+    if artist_type == 'top':
+        try:
+            payload = _get_top_artist_recommendations(recommendations, user_name)
+        except TopArtistRecommendationsNotFound:
+            err_msg = 'Oops! Looks like there was some Spark issue while generating top artist' \
+                      ' recommendations for {}. We are working on it. Please check back later!!'.format(user_name)
+            raise APINotFound(err_msg, payload={'last_updated': recommendations['created']})
+
+    if artist_type == 'similar':
+        try:
+            payload = _get_similar_artist_recommendations(recommendations, user_name)
+        except SimilarArtistRecommendationsNotFound:
+            err_msg = 'Oops! Looks like there was some Spark issue while generating similar artist' \
+                      ' recommendations for {}. We are working on it. Please check back later!!'.format(user_name)
+            raise APINotFound(err_msg, payload={'last_updated': recommendations['created']})
+
+    return jsonify(payload)
+
+
+def _get_top_artist_recommendations(recommendations, user_name):
+    """ Get top artist recommended recording mbids.
+
+        if recommendations not found raises TopArtistRecommendationsNotFound exception.
+    """
+    top_artist_recommendations = recommendations['recording_mbid']['top_artist']
+
+    if len(top_artist_recommendations) == 0:
+        raise TopArtistRecommendationsNotFound()
+
+    payload = {'payload': {
+        'top-artist': {
+            'recording_mbids': top_artist_recommendations
+        },
+        'user_name': user_name,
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+from listenbrainz.webserver.rate_limiter import ratelimit
+
+recommendations_cf_recording_api_bp = Blueprint('recommendations_cf_recording_v1', __name__)
+
+
+@recommendations_cf_recording_api_bp.route("/user/<user_name>/artist")
+@crossdomain()
+@ratelimit()
+def get_recommendations(user_name):
+    """ Get recommendations for user ``user_name``.
```
paramsingh, line 17: We should add a note/warning here that this endpoint is experimental and probably will change in the future.

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+
+        A sample response from the endpoint may look like::
+
+        {
+          "payload": {
+
+            "last_updated": "Sat, 06 Jun 2020 08:18:50 GMT",
```
paramsingh, line 24: Would prefer if we were consistent with other endpoints and kept this a unix timestamp.

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+    recommendations = db_recommendations_cf_recording.get_user_recommendation(user['id'])
+
+    if recommendations is None:
+        err_msg = 'Oops! Looks like {} doesn\'t have recent listening history.' \
+                  ' Recommendations not generated :('.format(user_name)
+        raise APINotFound(err_msg)
+
+    if artist_type == 'top':
+        try:
+            payload = _get_top_artist_recommendations(recommendations, user_name)
```
paramsingh, line 63: it doesn't make much sense that a `_get_recommendations` function takes recommendations, this function should either be renamed to something like `format_recommendations` or the database call should be done inside this function.

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+        :statuscode 200: Successful query, you have data!
+        :statuscode 400: Bad request, check ``response['error']`` for more details
+        :statuscode 404: User not found or recommendations not generated
+    """
+    user = db_user.get_by_mb_id(user_name)
+    if user is None:
+        raise APINotFound("Cannot find user: {}".format(user_name))
+
+    artist_type = request.args.get('type')
+    if not _is_valid_artist(artist_type):
```
paramsingh, line 51: the name `_is_valid_artist` is misleading, we're not checking if it's a valid artist, we're checking that it's a valid artist type

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+        raise APIBadRequest("Invalid artist type: {}".format(artist_type))
+
+    recommendations = db_recommendations_cf_recording.get_user_recommendation(user['id'])
+
+    if recommendations is None:
+        err_msg = 'Oops! Looks like {} doesn\'t have recent listening history.' \
```
paramsingh, line 57: The audience of the API endpoints are developers, I think we should keep these messages short and simple. `No recommendations due to absence of recent listening history for user "username"`

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+                  ' Recommendations not generated :('.format(user_name)
+        raise APINotFound(err_msg)
+
+    if artist_type == 'top':
+        try:
+            payload = _get_top_artist_recommendations(recommendations, user_name)
+        except TopArtistRecommendationsNotFound:
+            err_msg = 'Oops! Looks like there was some Spark issue while generating top artist' \
```
paramsingh, line 65: Same here, no need to mention spark here at all. `"No recommendations for user {}, please try again later.`

We also use an `APINoContent` instead of `NotFound` in the stats endpoints, it's probably a good idea to keep consistency here.

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+                      ' recommendations for {}. We are working on it. Please check back later!!'.format(user_name)
+            raise APINotFound(err_msg, payload={'last_updated': recommendations['created']})
+
+    if artist_type == 'similar':
+        try:
+            payload = _get_similar_artist_recommendations(recommendations, user_name)
+        except SimilarArtistRecommendationsNotFound:
+            err_msg = 'Oops! Looks like there was some Spark issue while generating similar artist' \
```
paramsingh, line 73: Let's improve this message too.

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+        'last_updated': recommendations['created'],
+    }}
+
+    return payload
+
+
+def _is_valid_artist(artist_type):
+    """ Check if artist type is valid.
+    """
+    if artist_type not in ['top', 'similar']:
```
paramsingh, line 125: instead of keeping a list here, you should create an enum and return `artist_type in RecommendationArtistType`

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+
+def _get_top_artist_recommendations(recommendations, user_name):
+    """ Get top artist recommended recording mbids.
+
+        if recommendations not found raises TopArtistRecommendationsNotFound exception.
+    """
+    top_artist_recommendations = recommendations['recording_mbid']['top_artist']
+
+    if len(top_artist_recommendations) == 0:
+        raise TopArtistRecommendationsNotFound()
```
paramsingh, line 88: these braces are not needed here. also, maybe just raise the `APIException` here instead of doing two layers of exceptions.

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+
+def _get_similar_artist_recommendations(recommendations, user_name):
+    """ Get similar artist recommended recording mbids.
+
+        if recommendations not found raises SimilarArtistRecommendationsNotFound exception.
+    """
+    similar_artist_recommendations = recommendations['recording_mbid']['similar_artist']
+
+    if len(similar_artist_recommendations) == 0:
+        raise SimilarArtistRecommendationsNotFound()
```
paramsingh, line 109: parentheses not needed, also could just directly raise the exception here.

listenbrainz/webserver/views/recommendations_cf_recording_api.py
paramsingh: We mention nowhere that this endpoint needs a `type` parameter that takes values `top` or `similar`. We should do that somewhere, see https://github.com/metabrainz/listenbrainz-server/blob/32461e3f/listenbrainz/webserver/views/stats_api.py#L90 for an example

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+
+        if recommendations not found raises TopArtistRecommendationsNotFound exception.
+    """
+    top_artist_recommendations = recommendations['recording_mbid']['top_artist']
+
+    if len(top_artist_recommendations) == 0:
+        raise TopArtistRecommendationsNotFound()
+
+    payload = {'payload': {
+        'top-artist': {
```
paramsingh, line 91: Generally, we avoid dashes (`-`) in JSON fields because it makes the object a bit harder to consume in Javascript and similar languages. Could we use underscore here instead?

listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -0,0 +1,135 @@
...
+                "a6ad0205-6e96-416d-a4e8-edd1773dac09",
+                "d8783d03-8a3b-4269-8261-00709d2cfee8"
+              ]
+            },
+
+            "user_name": "unclejohn69"
+          }
+        }
+
+        <artist_type>: 'top' or 'similar'
```
paramsingh, line 40: the docs say that artist_type is `top` or `similar` but the actual code returns `top-artists` and `similar-artists`, we should be consistent here.

PR #931, opened 2020-06-25
Improve query for all time "Listening Activity" statsistics

ishaanshah:
# Problem
The current query for calculation "Listening Activity" is not efficient and does not work for calculating all time statistics.

# Solution
A better way to calculate all time statistics is to calculate the data year wise and merge it in a later stage.

### DIFF
--- a/listenbrainz_spark/stats/user/listening_activity.py
+++ b/listenbrainz_spark/stats/user/listening_activity.py
@@ -7,6 +7,7 @@ from pydantic import ValidationError
 import listenbrainz_spark
 from data.model.user_listening_activity import UserListeningActivityStatMessage
 from listenbrainz_spark.constants import LAST_FM_FOUNDING_YEAR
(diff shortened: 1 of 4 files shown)


listenbrainz_spark/stats/user/listening_activity.py
```diff
@@ -156,16 +157,32 @@ def get_listening_activity_all_time() -> Iterator[UserListeningActivityStatMessa
...
+            _get_listens(year_start, year_end)
+        except HDFSException:
+            # Skip if no listens present in df
+            continue
+        year_df = run_query("""
+                    SELECT user_name,
+                           count(user_name) as listen_count
+                      FROM listens
+                  GROUP BY user_name
+                """.format(year=year))
```
paramsingh, line 174: This format isn't needed.

ishaanshah: Removed

PR #934, opened 2020-06-26
Request and generate recommendations for list of users 

vansika:
As discussed on IRC in light of [LB-648](https://tickets.metabrainz.org/browse/LB-648)

### DIFF
--- a/listenbrainz/spark/request_manage.py
+++ b/listenbrainz/spark/request_manage.py
@@ -165,12 +165,15 @@ def request_candidate_sets(days, top, similar):
 @cli.command(name='request_recommendations')
 @click.option("--top", type=int, default=200, help="Generate given number of top artist recommendations")
 @click.option("--similar", type=int, default=200, help="Generate given number of similar artist recommendations")
-def request_recommendations(top, similar):
+@click.option("--mb_id", callback=parse_list, default=[], multiple=True, help="Generate recommendations for given users" \
+              " Generate recommendation for all users by default.")
+def request_recommendations(top, similar, mb_id):
     """ Send the cluster a request to generate recommendations.
     """
     params = {
         'recommendation_top_artist_limit': top,
         'recommendation_similar_artist_limit': similar,
+        'musicbrainz_id': mb_id
     }
     send_request_to_spark_cluster(_prepare_query_message('cf_recording.recommendations.recommend', params=params))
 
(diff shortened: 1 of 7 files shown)


listenbrainz/spark/request_manage.py
```diff
@@ -165,12 +165,15 @@ def request_candidate_sets(days, top, similar):
 @cli.command(name='request_recommendations')
 @click.option("--top", type=int, default=200, help="Generate given number of top artist recommendations")
 @click.option("--similar", type=int, default=200, help="Generate given number of similar artist recommendations")
-def request_recommendations(top, similar):
+@click.option("--mb_id", callback=parse_list, default=[], multiple=True, help="Generate recommendations for given users" \
```
mayhem, line 168: I think user-ids might be better; also note that options should use - and not _ to separate words. Also, how are more than one id specified? comma seperated? I think the usage statement should indicate this.

mayhem: And this comment really applies to this whole PR...

vansika: I guess the usage should be updates in the Readme like it is for other commands.

vansika: we have been using user id on server side for some integer. I guess it will be confusing to use user id here. 

We actually want to send musicbrainz ids of users to spark.

listenbrainz_spark/recommendations/tests/test_recommend.py
```diff
@@ -182,34 +182,89 @@ def test_get_users(self):
...
+            'musicbrainz_id': 'vansika',
+            'type': 'cf_recording_recommendations',
+            'top_artist': ["181c4177-f33a-441d-b15d-910acaf18b07"],
+            'similar_artist': ["281c4177-f33a-441d-b15d-910acaf18b07"],
+        }]
+
+        active_users = ['vansika']
+        users = ['vansika', 'vansika_1']
+        messages = recommend.get_message_for_inactive_users(message_arg, active_users, users)
+        print(messages[0])
```
mayhem, line 204: print still needed?

listenbrainz_spark/recommendations/README.md
```diff
@@ -89,12 +89,19 @@ The recommendation generation process has been divided into four stages.
 
   Send a request to spark cluster to generate and return recommendations.
 
-  `./develop.sh manage spark request_recommendations --top=X --similar=Y`
+  `./develop.sh manage spark request_recommendations --top=X --similar=Y --user-id=Z`
```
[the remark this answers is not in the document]
vansika: Okay

listenbrainz_spark/recommendations/tests/test_recommend.py
vansika: Oops🤦‍♀️

PR #936, opened 2020-06-27
LB-639: Send only top 1000 recordings for each user

ishaanshah:
# Problem
All Time recording stats are not getting stored in the DB for some users because they are getting dropped by the request consumer.

# Solution
Most probably this is because of large message size which RMQ has trouble handling. Reducing the message size by limiting the number of entities calculated should fix this issue.

# Action
We should test this on production.

### DIFF
--- a/listenbrainz_spark/request_consumer/request_consumer.py
+++ b/listenbrainz_spark/request_consumer/request_consumer.py
@@ -80,7 +80,12 @@ class RequestConsumer:
                     self.rabbitmq.close()
                     self.connect_to_rabbitmq()
                     self.init_rabbitmq_channels()
-        current_app.logger.debug("Done!")
+
+        avg_size_of_message //= num_of_messages
+
+        current_app.logger.info("Done!")
+        current_app.logger.info("Number of messages sent: {}".format(num_of_messages))
+        current_app.logger.info("Average size of message: {} bytes".format(avg_size_of_message))
 
     def callback(self, channel, method, properties, body):
         request = json.loads(body.decode('utf-8'))
--- a/listenbrainz_spark/request_consumer/request_consumer.py
+++ b/listenbrainz_spark/request_consumer/request_consumer.py
@@ -56,7 +56,8 @@ class RequestConsumer:
         try:
             return query_handler(**params)
         except TypeError as e:
-            current_app.logger.error("TypeError in the query handler for query '%s', maybe bad params. Error: %s", query, str(e), exc_info=True)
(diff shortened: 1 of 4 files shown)


listenbrainz_spark/request_consumer/request_consumer.py
```diff
@@ -80,7 +80,12 @@ def push_to_result_queue(self, messages):
                     self.rabbitmq.close()
                     self.connect_to_rabbitmq()
                     self.init_rabbitmq_channels()
-        current_app.logger.debug("Done!")
+
+        avg_size_of_message //= num_of_messages
```
paramsingh, line 84: Sorry, this is a little confusing, where is avg_size initially defined?

listenbrainz_spark/stats/user/entity.py
```diff
@@ -129,6 +129,11 @@ def create_messages(data, entity: str, stats_range: str, from_ts: int, to_ts: in
     """
     for entry in data:
         _dict = entry.asDict(recursive=True)
+
+        # Clip the recordings to top 1000 so that we don't drop messages
+        if entity == "recordings":
```
paramsingh, line 134: Could we only do this for all time? Because that's what is causing problems rn?

listenbrainz_spark/request_consumer/request_consumer.py
ishaanshah: Sorry for that, incorrect `git add -p`. Fixed it.

listenbrainz_spark/stats/user/entity.py
ishaanshah: Changed

listenbrainz_spark/stats/user/tests/test_entity.py
```diff
@@ -86,3 +86,34 @@ def test_get_entity_all_time(self, mock_create_messages, mock_get_listens, mock_
         mock_df.createOrReplaceTempView.assert_called_with('user_test_all_time')
         mock_create_messages.assert_called_with(data='user_test_all_time_data', entity='test', stats_range='all_time',
                                                 from_ts=from_date.timestamp(), to_ts=to_date.timestamp())
+
+    def test_create_messages_recordings(self):
+        """ Test to check if the number of recordings are clipped to top 1000 """
+        recordings = []
+        for i in range(0, 2000):
```
paramsingh, line 93: This pattern of generating test data instead of saving huge json files in git is great! We should consider doing this on the frontend as well in the future.

PR #937, opened 2020-06-28
get unique recording mbid recommendations

vansika:
# Problem
The cluster is recommending non-unique mbids [LB-650](https://tickets.metabrainz.org/browse/LB-650)

# Solution

- use `msb_recording_name_matchable`  instead of `track_name` in [mapped_listens](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz_spark/recommendations/create_dataframes.py#L183)
- more than one top artists for a user can have the same similar artists which will lead to repeating recording ids. Filter distinct similar artists for each user. 
- Update candidate sets HTML for easy debugging

*Note:* I was able to spot the above-mentioned prob/sol on my local machine. There may be more to it which I can understand once I have HTML files generated in prod.

### DIFF
--- a/listenbrainz_spark/recommendations/candidate_sets.py
+++ b/listenbrainz_spark/recommendations/candidate_sets.py
@@ -1,6 +1,5 @@
 import sys
 import uuid
-import json
 import logging
 from time import time
 from datetime import datetime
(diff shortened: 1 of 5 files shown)


vansika:
@paramsingh ready for merge

PR #940, opened 2020-06-30
Minor fixes in train_models for readability

vansika:


vansika:
I am merging this PR since it only updates the docs.

PR #943, opened 2020-06-30
Fetch next/previous listens in front-end using the API

MonkeyDo:
# Problem
Currently, fetchin gthe next/previous page of listens requires to navigate to another page.
That's costly for everyone, slow and bad UX.

Instead, let's use the existing API endpoints to fetch listens in the web page.

# Action

I still have to

- [x] finalize tests 
- [x] update the browser address bar on navigation
- [x] add buttons to navigate to oldest/newest listens


MonkeyDo:
Rebased and ready to merge!

PR #956, opened 2020-07-09
LB-654: Skip entities with incorrect data while calculating entity statistics

ishaanshah:
# Problem
Some users have incorrect data which leads to validation error while calculating stats for the user.
https://sentry.metabrainz.org/metabrainz/listenbrainz-beta/issues/96963/?query=is%3Aunresolved

# Solution
Iterating over each entry and validating it individually solves the issue. 

# Action
This should be merged after #936


paramsingh:
Let's rebase this now that #936 is merged. :)

PR #957, opened 2020-07-12
Refactor tests for create_dataframes: add testdata

vansika:


vansika:
Not required now when I think of it. Tests are good enough as of now. I might work on it in future.

PR #958, opened 2020-07-12
update steps to reach next month

vansika:
` from_date = datetime(2020, 6, 30)`
`date = from_date  + relativedelta(days=32)`
`print(date) = 2020-08-01 00:00:00`

ideally, it should be `2020-07-01 00:00:00`

I guess the problem is that we are using `32` as steps to reach the next month, it should be `31`.

### DIFF
--- a/listenbrainz_spark/config.py.sample
+++ b/listenbrainz_spark/config.py.sample
@@ -19,7 +19,7 @@ SPARK_RESULT_QUEUE = "spark_result"
 # calculate stats on X months data
 STATS_CALCULATION_WINDOW = 1
 
-STEPS_TO_REACH_NEXT_MONTH = 32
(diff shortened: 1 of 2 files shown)


mayhem:
I don't think 32 or 31 are the right things here. You want to go 1 month into the future, yes? Why not?

>>> from dateutil import relativedelta
>>> from datetime import datetime
>>> from_date = datetime(2020, 6, 30)
>>> date = from_date + relativedelta.relativedelta(months=1)
>>> date
>>> datetime.datetime(2020, 7, 30, 0, 0)

vansika:
@mayhem 🤣🤣🤣

vansika:
Dumb me 🤦‍♀️

PR #959, opened 2020-07-13
LB-659: Add ability to revert back to older dump if the import fails

ishaanshah:
# Problem
The stats pipeline calculates incorrect data if the import is incorrect. We should automatically fall back to older data if the import fails.

# Solution
We can extract the whole `tarfile` first and then process all the files afterwards. This way if the `tarfile` is corrupt we can abort the import and avoid having incorrect data in `HDFS`.

### DIFF
--- a/listenbrainz_spark/hdfs/__init__.py
+++ b/listenbrainz_spark/hdfs/__init__.py
@@ -62,14 +62,14 @@ class ListenbrainzHDFSUploader:
             raise NotImplementedError('Callback to process JSON missing. Aboritng...')
 
         # Extract all files in tmp_dump_dir
-        t = time.time()
+        t0 = time.time()
         current_app.logger.info("Extracting dump in temporary directory")
         try:
             tar.extractall(path=tmp_dump_dir)
         except TarError as err:
-            current_app.logger.error("{} while extracting tarfile, aborting import".format(type(err).__name__))
+            current_app.logger.error("{} while extracting tarfile, aborting import".format(type(err).__name__), exc_info=True)
             return
-        time_taken = time.time() - t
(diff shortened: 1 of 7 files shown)


listenbrainz_spark/hdfs/__init__.py
```diff
@@ -57,31 +59,52 @@ def upload_archive(self, tmp_dump_dir, tar, dest_path, schema, callback=None, fo
...
+                except TarError as err:
+                    current_app.logger.error("{} while extracting {}, aborting import".format(
+                        type(err).__name__, member.name), exc_info=True)
+                    # Cleanup
+                    if utils.path_exists('/temp'):
+                        utils.delete_dir('/temp', recursive=True)
+                    if utils.path_exists(tmp_dump_dir):
+                        utils.delete_dir(tmp_dump_dir, recursive=True)
+                    shutil.rmtree(tmp_dump_dir)
+                    return
```
paramsingh, line 83: If we just return here, the handler on the spark_reader side will not know that the import failed, we should raise an exception here, catch it in the handler and send back an appropriate message.

paramsingh:
While we're in this part of the codebase, could you add a sha256 check before we actually start downloading the dump? Each valid dump file should have a corresponding sha256 file on FTP. We should check that the file exists, download the dump, check the downloaded dump's hash with the hash value in the file.

listenbrainz_spark/hdfs/__init__.py
ishaanshah: We do not send back anything to the `spark_reader` but log the error on sentry for any other request. So I have raised `DumpInvalidException` with appropriate message so that it gets logged in sentry instead of returning.

PR #960, opened 2020-07-15
LB-660: Fix incorrect recording count on charts page

ishaanshah:
# Problem
We are calculating only top 1000 recordings to avoid dropping stats in RMQ. However we should show the actual number of distinct recordings the user has listened to, instead of 1000.

# Solution
Calculating the entity count before clipping it to top 1000 fixes the issue.


paramsingh:
Let's add a test as well

PR #962, opened 2020-07-16
update MSB to MB in text fields in msid->mbid mapping

vansika:
We'd want to do get MB data (instead of MSB data) after lookup, therefore we are using data which comes from MB  in the field names (recording_name, artist_name, release_name). The changes were reflected in the recent mapping uploaded on FTP but the col names were not updated in the codebase. This PR intends to do that.


vansika:
Not a problem. Only minor changes <3

vansika:
I will want to merge this myself some time later.

PR #964, opened 2020-07-16
create html_files directory inside recommendations dir

vansika:
The `html_files` dir might have been accidentally deleted and I wasn't able to track it since it was a part of gitignore. So I have added a test to keep track of the folder.

### DIFF
--- a/listenbrainz_spark/recommendations/utils.py
+++ b/listenbrainz_spark/recommendations/utils.py
@@ -13,6 +13,7 @@ def save_html(filename, context, template):
         html = template_environment.get_template(template).render(context)
         f.write(html)
 
+
 def check_html_files_dir_path():
     dir_exists = os.path.isdir(HTML_FILES_PATH)
     return dir_exists
--- a/listenbrainz_spark/recommendations/utils.py
+++ b/listenbrainz_spark/recommendations/utils.py
@@ -2,11 +2,17 @@ import os
 
 from jinja2 import Environment, FileSystemLoader
 
+HTML_FILES_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'html_files')
+
 def save_html(filename, context, template):
     path = os.path.dirname(os.path.abspath(__file__))
     template_environment = Environment(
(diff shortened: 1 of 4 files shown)


vansika:
The dir is being used in many places, instead of checking everywhere, I'd prefer to add a basic test.

vansika:
> Do we need to keep this in the codebase?

Keep what?

paramsingh:
>Keep what?

The `html_files` directory. 

Generally we put things that are code into the git repository, not the output of the code. I'm not in favor of this. This PR basically adds a conflict between the git ignore and th actual codebase. 

If there are too many places that use the dir, maybe extract the file access into a function that first checks the directory exists, if it doesn't, creates it and then does the work. There is no reason why we should track this folder in git.

I realize the many changes might be frustrating, but following best practices generally helps reduce toil in the future (and makes it easier to do future changes)

vansika:
K.

listenbrainz_spark/recommendations/utils.py
```diff
@@ -2,11 +2,18 @@
...
     template_environment = Environment(
     loader=FileSystemLoader(os.path.join(path, 'templates')))
-    outputfile = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'html_files', filename)
+    outputfile = os.path.join(HTML_FILES_PATH, filename)
     with open(outputfile, 'w') as f:
         html = template_environment.get_template(template).render(context)
         f.write(html)
+
+
+def check_html_files_dir_path():
```
paramsingh, line 17: You could just do this test in the `save_html` function above, and then the users of the save html function don't have to worry about it.

vansika: Ah! I don't know what I was thinking whilst writing it 🙄. Thank you 🌹

vansika:
@brainzbot retest this please.

paramsingh:
Rebasing over master should fix the build, I think. One of the frontend tests in master was flaky.

vansika:
Could you please merge this and send requests for recommendations?

paramsingh:
Sure, could you rebase once, so that we know the build passes

PR #966, opened 2020-07-17
LB-658: Add "Daily Activity" statistics (Spark + DB + API)

ishaanshah:
# Problem

# Solution

# Action


listenbrainz/webserver/views/stats_api.py
```diff
@@ -412,6 +412,82 @@ def get_listening_activity(user_name: str):
     }})
 
 
+@stats_api_bp.route("/user/<user_name>/daily-activity")
+@crossdomain()
+@ratelimit()
+def get_daily_activity(user_name: str):
+    """
+    Get the daily activity for user ``user_name``. The daily activity shows the number of listens
+    submitted by the user per hour over a period of time.
```
paramsingh, line 421:
```suggestion
    submitted by the user for each hour of the day over a period of time. We assume that all listens are in UTC.
```

PR #967, opened 2020-07-17
[wip] Neighbours: User similarity based on Collaborative filtering on artist stats

paramsingh:
I wrote up a small script that uses `implicit` and the top artist stats to calculate top X similar users for each user. And an API endpoint.

This is very similar to the neighbours features that Last.FM used to have. 

It's nearer to production readiness than I expected, need to add a few tests. 

The script and approach for the similarity calculation is brittle and could use some review.


listenbrainz/stats/__init__.py
```diff
@@ -1 +0,0 @@
-STATS_QUEUE_REDIS_KEY_PREFIX = 'listenbrainz.stats_queue.'
```
paramsingh, line 1: Unrelated, i was doing some cleanup.

listenbrainz/stats/exceptions.py
```diff
@@ -1,9 +1,2 @@
-
 class StatisticsException(Exception):
     pass
-
-class NoCredentialsVariableException(StatisticsException):
```
paramsingh, line 5: Unrelated.

admin/sql/create_tables.sql
```diff
@@ -143,4 +143,10 @@ CREATE TABLE recording_feedback (
     created                 TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
 
+CREATE TABLE similar_user (
```
mayhem, line 146: I think we should also put this into the recommendation schema.

data/model/similar_user_model.py
```diff
@@ -0,0 +1,9 @@
+import pydantic
+from typing import List
+
+class SimilarUserRecord(pydantic.BaseModel):
+    user_id: int
+    similarity_score: float
```
mayhem, line 6: what is the range of this float?

listenbrainz/db/user.py
```diff
@@ -1,10 +1,13 @@
 import logging
 import sqlalchemy
 import uuid
+import json
```
mayhem, line 4: lets try to use ujson for speed

listenbrainz/stats/user_similarity.py
```diff
@@ -0,0 +1,48 @@
...
+    and the listen count as the value. This is then used to train the CF model.
+    """
+    artist_ids = {}
+    count = 0
+    data = []
+    row_ind = []
+    col_ind = []
+    for stat_record in user_artist_stats:
+        for artist in stat_record.all_time.artists:
+            if artist.artist_name in artist_ids: #TODO: use artist MSID or MBID
```
mayhem, line 19: I think running the alg on artists_credit_ids is best and then translating them to MBIDs as a post processing step.

alastair:
awesome! This is super close to the artist stuff that I played with a few months back, also using implicit. that work does:
 * Given a user, recommend artists to listen to
 * Given an artist, give other similar artists

I was trying to adapt the existing code that I had to load the LB data, but I think it might be better to take what you have already and add it to here. I'll take a look at this code next week, give some feedback, and see how I can integrate my stuff into it!

paramsingh:
>Given a user, recommend artists to listen to
>Given an artist, give other similar artists

the model trained here can do the same things, I used implicit because of the work you mentioned :)

data/model/similar_user_model.py
paramsingh: Not sure, the docs for implicit weren't very clear.

mayhem:
Closing this to start working in a new direction based on this.

PR #974, opened 2020-07-22
remove beta from navbar and the spotify import description

paramsingh:
# Problem

We're mentioning stuff is in beta when it's been stable for years.

# Solution

Remove the beta tag from the site navbar and from the spotify importer. Also remove a mention from the API docs.

# Action

Deploy!

### DIFF
--- a/listenbrainz/webserver/templates/navbar.html
+++ b/listenbrainz/webserver/templates/navbar.html
@@ -10,7 +10,7 @@
         <span class="icon-bar"></span>
       </button>
       <a class="navbar-brand logo" href="{{ url_for('index.index') }}">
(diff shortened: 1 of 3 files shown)


paramsingh:
Pretty much! (#975)

listenbrainz/webserver/templates/navbar.html
```diff
@@ -10,7 +10,7 @@
         <span class="icon-bar"></span>
       </button>
       <a class="navbar-brand logo" href="{{ url_for('index.index') }}">
-          <img src="{{ url_for('static', filename='img/navbar_logo.svg') }}" alt="ListenBrainz" /><span class="beta">beta</span>
```
alastair, line 13: end of an era...

PR #975, opened 2020-07-22
Improve landing page text

paramsingh:
# Problem

I'd been thinking of rewording our landing page for some time, it isn't very clear on what ListenBrainz is.

This [HN comment](https://news.ycombinator.com/item?id=23918941) is pretty valid feedback.

# Solution

The HN comment pushed me over the edge and I took a shot at rewording the text to make it easier 
for the general non-music technology nerd population to parse.

The first part explains what ListenBrainz does and how to get started. It also points users to discourse in 
case they have more questions.

The second part keeps the info we already had about ListenBrainz (operated by MetaBrainz, commercial users are encouraged to support us, blah blah)

The third part is for contributors. I added a link to https://metabrainz.org/donate.

## Screenshot
image

# Action

In the future, we should probably look at redesigning this page. But a verbiage improvement should do for now.

### DIFF
--- a/listenbrainz/webserver/templates/index/index.html
+++ b/listenbrainz/webserver/templates/index/index.html
@@ -6,7 +6,7 @@
       <img src="{{ url_for('static', filename='img/logo_big.svg') }}" id="big-logo" />
       <h2 class="page-title">Welcome to ListenBrainz!</h2>
       <p>
-        ListenBrainz is an open repository of your music listening history.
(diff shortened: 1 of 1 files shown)


paramsingh:
Here's the text in doc form so that it's easier to collaborate: https://docs.google.com/document/d/1V57QgC-ByNGXKkvEWoe7h0bQgqNYk664lQTeZYltSUU/edit?usp=sharing

PR #976, opened 2020-07-23
get_user_timestamps improvements

mayhem:
# Problem

https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/97474/?environment=production&referrer=alert_email

# Solution

Return 0 if No timestamps are available. add test.


mayhem:
This is related to the heisenbug, but may not fix it. It should however fix the bug in production.

PR #978, opened 2020-07-24
Improve readability of "Daily Activity" API endpoint

ishaanshah:
# Problem
The "Daily Activity" API endpoint is confusing and hard to understand.

# Solution
Grouping hours of each day into a single list solves the issue.

### DIFF
--- a/listenbrainz/webserver/views/stats_api.py
+++ b/listenbrainz/webserver/views/stats_api.py
@@ -1,3 +1,4 @@
+import calendar
 from datetime import datetime
 from enum import Enum
 from typing import List, Union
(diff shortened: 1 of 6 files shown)


listenbrainz/webserver/views/stats_api.py
```diff
@@ -477,7 +474,23 @@ def get_daily_activity(user_name: str):
...
-    daily_activity = [x.dict() for x in getattr(stats, stats_range).daily_activity]
+    daily_activity_unprocessed = [x.dict() for x in getattr(stats, stats_range).daily_activity]
+    daily_activity = {calendar.day_name[day]: [{"hour": hour, "listen_count": 0} for hour in range(0, 24)] for day in range(0, 7)}
+
+    for day, day_data in daily_activity.items():
+        for hour_data in day_data:
+            hour = hour_data["hour"]
+
+            found = False
+            for entry in daily_activity_unprocessed:
```
paramsingh, line 485: you can use a `for...else` here to get rid of the `found` variable. (https://book.pythontips.com/en/latest/for_-_else.html)

listenbrainz/webserver/views/stats_api.py
```diff
@@ -530,6 +543,8 @@ def _get_entity_list(
     entity: str,
     offset: int,
     count: int,
+
```
paramsingh, line 546: could we remove these spaces?

listenbrainz/webserver/views/stats_api.py
ishaanshah: Neat trick!

listenbrainz/webserver/views/stats_api.py
ishaanshah: Fixed it.

PR #980, opened 2020-07-24
don't join on msb_artist_credit_name_matchable to get similar artists

vansika:
The `similar_artist_df` gets the artist names from `artist-relations` which aren't matchable. The artist names in `recordings_df` are matchable. Therefore joining the `similar_artist_df` and `recordings_df` yields nothing at all. 

The simple solution is to join them only on `artist_credit_id`. We need not worry about repetitions since 'artist_credit_id` is unique.


vansika:
@brainzbot retest this please

PR #983, opened 2020-07-24
Improve SHA verification

ishaanshah:
# Problem
`SHA` verification was failing because of trailing whitespaces and presence of filenames.

# Solution
Stripping all the extra whitespaces and removing everything after the first space fixes the issue.

### DIFF
--- a/listenbrainz_spark/ftp/__init__.py
+++ b/listenbrainz_spark/ftp/__init__.py
@@ -105,8 +105,6 @@ class ListenBrainzFTPDownloader:
         """ Reads the SHA file and returns the string stripped of any whitespace and extra characters
         """
         with open(filepath, "r") as f:
-            sha = f.read().replace('\n', '')  # Get rid of new line characters
-            sha = sha.lstrip().rstrip()  # Get rid of any leading or trailing whitespace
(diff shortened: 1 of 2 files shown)


listenbrainz_spark/ftp/__init__.py
```diff
@@ -99,3 +100,13 @@ def _calc_sha256(self, filepath: str) -> str:
                 calculated_sha.update(byte_block)
 
         return calculated_sha.hexdigest()
+
+    def _read_sha_file(self, filepath: str) -> str:
+        """ Reads the SHA file and returns the string stripped of any whitespace and extra characters
+        """
+        with open(filepath, "r") as f:
+            sha = f.read().replace('\n', '')  # Get rid of new line characters
```
mayhem, line 108: I think these three lines can be replaced with:

sha = f.read().split(" ", 1)[0]

no?

mayhem:
Also, thanks!

listenbrainz_spark/ftp/__init__.py
ishaanshah: Makes sense, fixed it.

PR #984, opened 2020-07-24
Change cron jobs for data dumps

paramsingh:
# Problem

Our stats are out of date. It's because we're creating and importing dumps at non-ideal times.

# Solution

Change the timings, create the dumps on monday and thursday. Each full dump takes ~16h, for safety trigger a dump import job 24 hours after the dump creation started.

### DIFF
--- a/docker/dump-crontab
+++ b/docker/dump-crontab
@@ -1,6 +1,14 @@
-# Full Data dumps creation for backup and public use, every Sunday
-00 00 * * 0 /code/listenbrainz/admin/create-dumps.sh full
+# Full dump creation takes the ID of the last incremental
+# dump and creates a full dump for that ID. So, we'll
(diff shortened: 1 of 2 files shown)


docker/dump-crontab
```diff
@@ -1,6 +1,14 @@
-# Full Data dumps creation for backup and public use, every Sunday
-00 00 * * 0 /code/listenbrainz/admin/create-dumps.sh full
+# Full dump creation takes the ID of the last incremental
+# dump and creates a full dump for that ID. So, we'll
+# start creating full dumps and incremental dumps in pairs
+# for now.
```
mayhem, line 4: And sometime later when things are running we do full dumps less often?

paramsingh: yes! I'm planning on using a few weeks of ishaan's gsoc to get incremental dumps support.

PR #985, opened 2020-07-25
LB-671: Add "Artist Origin" statistics (Frontend + Backend)

paramsingh:
# Problem
Not really a problem, but a geography of your artists would be cool.

# Solution
Add a graph using the already calculated artist stats and hosted datasets on bono.metabrainz.org

# Action
- [x] Account for artist MBIDs
- [x] Productionize the API endpoint
- [x] Make it responsive
- [x] Add tests

# Screenshot
2020-08-04-165834_1908x937_scrot


listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
```diff
@@ -0,0 +1,178 @@
...
+                    </span>
+                  </div>
+                )}
+                {!hasError && (
+                  <ResponsiveChoropleth
+                    data={data}
+                    features={features.features}
+                    margin={{ top: 0, right: 0, bottom: 0, left: 0 }}
+                    colors="nivo"
+                    domain={[0, 25]}
```
paramsingh, line 134: This domain needs to be variable,  [0, count for country with max artists]

listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
```diff
@@ -0,0 +1,178 @@
...
+                      {errorMessage}
+                    </span>
+                  </div>
+                )}
+                {!hasError && (
+                  <ResponsiveChoropleth
+                    data={data}
+                    features={features.features}
+                    margin={{ top: 0, right: 0, bottom: 0, left: 0 }}
+                    colors="nivo"
```
paramsingh, line 133: colors need to match our theme

listenbrainz/webserver/views/stats_api.py
```diff
@@ -550,3 +584,29 @@ def _get_entity_list(
...
+    for index in range(0, len(artist_msids), 50):
+        current_batch = artist_msids[index: index + 50]
+        try:
+            artist_credit_id_data = requests.get('http://bono.metabrainz.org:8000/artist-msid-lookup/json', {
+                '[artist_msid]': ','.join(current_batch)
+            }).json()
+            artist_credit_ids = [
+                x['artist_credit_id'] for x in artist_credit_id_data]
+
+            location_data = requests.get('http://bono.metabrainz.org:8000/artist-credit-id-country-code/json', {
```
paramsingh, line 600: The dataset isn't returning country codes for a majority of artists. we need to look into that.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -550,3 +584,29 @@ def _get_entity_list(
     elif entity == 'recording':
         return getattr(stats, stats_range).recordings[offset:count]
     raise APIBadRequest("Unknown entity: %s" % entity)
+
+
+def _get_country_codes_by_msids(artist_msids):
+    country_map = defaultdict(int)
+    for index in range(0, len(artist_msids), 50):
```
paramsingh, line 591: Once the dataset hoster allows POST, we should try to minimize the number of requests we need to send to it.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -498,6 +503,35 @@ def get_daily_activity(user_name: str):
...
+    if not _is_valid_range(stats_range):
+        raise APIBadRequest("Invalid range: {}".format(stats_range))
+
+    stats = db_stats.get_user_artists(user['id'], stats_range)
+    if stats is None or getattr(stats, stats_range) is None:
+        raise APINoContent('')
+
+    artist_msids = [artist.artist_msid for artist in getattr(
+        stats, stats_range).artists if artist.artist_msid is not None]
+    country_code_data = _get_country_codes_by_msids(artist_msids)
```
paramsingh, line 524: We should consider storing this in the database once it's been calculated (and updating every few days or so). This means that the endpoint will only be slow once.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -498,6 +503,35 @@ def get_daily_activity(user_name: str):
     }})
 
 
+@stats_api_bp.route("/user/<user_name>/artist-map")
+@crossdomain()
+@ratelimit()
+def get_artist_map(user_name: str):
```
paramsingh, line 509: Docs and tests.

listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
```diff
@@ -0,0 +1,178 @@
...
+};
+
+export type UserArtistMapState = {
+  loading: boolean;
+  hasError?: boolean;
+  errorMessage?: string;
+  data: Array<{ id: string; value: number }>;
+};
+
+export default class UserArtistMap extends React.Component<
```
paramsingh, line 28: Needs tests

listenbrainz/webserver/views/stats_api.py
```diff
@@ -550,3 +584,29 @@ def _get_entity_list(
...
+            artist_credit_ids = [
+                x['artist_credit_id'] for x in artist_credit_id_data]
+
+            location_data = requests.get('http://bono.metabrainz.org:8000/artist-credit-id-country-code/json', {
+                '[artist_credit_id]': ','.join([str(id) for id in artist_credit_ids])
+            }).json()
+            for location in location_data:
+                country = alpha_2_to_3[location['area_code'].split(',')[0]]
+                country_map[country] += 1
+        except json.decoder.JSONDecodeError:
```
paramsingh, line 606: should probably break this into two, so that we know which request caused the decode error.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -498,7 +508,127 @@ def get_daily_activity(user_name: str):
...
+                stats_range: {
+                    "artist_map": country_code_data,
+                    "from_ts": int(getattr(artist_stats, stats_range).from_ts),
+                    "to_ts": int(getattr(artist_stats, stats_range).to_ts),
+                    "last_updated": int(datetime.now().timestamp())
+                }
+            })
+
+            # Store in DB for future use
+            db_stats.insert_user_artist_map(user['id'], result)
```
paramsingh, line 618: What happens if this insert fails? We should catch an exception here and return the data either ways, given that we've calculated it already.

paramsingh: Also, let's have a test for this case, where we've calculated the data but the insert fails.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -550,3 +680,71 @@ def _get_entity_list(
...
+    ]
+
+
+def _get_mbids_from_msids(artist_msids: list) -> list:
+    """ Get list of artist_mbids corresponding to the input artist_msids
+    """
+    request_data = [{"artist_msid": artist_msid} for artist_msid in artist_msids]
+    artist_mbids = []
+    try:
+        result = requests.post("https://labs.api.listenbrainz.org/artist-credit-from-artist-msid/json", json=request_data)
```
paramsingh, line 717: the labs root url should be in the config file.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -13,7 +17,13 @@
...
 from data.model.user_release_stat import (UserReleaseRecord,
                                           UserReleaseStat)
-from data.model.user_listening_activity import (UserListeningActivityRecord, UserListeningActivityStat)
+from data.model.user_listening_activity import (
+    UserListeningActivityRecord, UserListeningActivityStat)
+from data.model.user_artist_map import (UserArtistMapRecord,
+                                        UserArtistMapStat,
+                                        UserArtistMapStatJson,
+                                        UserArtistMapStatRange)
+from listenbrainz.config import LISTENBRAINZ_LABS_API_URL
```
paramsingh, line 26: This leads to race conditions in prod. Could we use `current_app.config['LISTENBRAINZ_LABS_API_URL']` instead?

listenbrainz/config.py.sample
```diff
@@ -120,6 +120,7 @@ LASTFM_PROXY_URL = 'http://localhost:8080/'
 SERVER_ROOT_URL = 'http://localhost'
 WEBSOCKETS_SERVER_URL = 'http://localhost:8081'
 MUSICBRAINZ_OAUTH_URL = 'https://musicbrainz.org/oauth2/userinfo'
+LISTENBRAINZ_LABS_API_URL = 'https://labs.api.listenbrainz.org'
```
paramsingh, line 123: Could you add this to `consul_config.py.ctmpl`? Exactly this line should be ok, we don't need to make it variable.

paramsingh:
Can't approve a PR I opened, but LGTM. I'll wait for the build to pass.

stragu:
@paramsingh just curious: how does this handle areas that are not countries, for example cities? Does it still check which country the more precise area is in, and use that instead?

Thanks for the great addition, by the way! :smile:

mayhem:
We take the area of the artist and then work out which country that artist is in and then put that on the map. If you really care, the code in question is here:

https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/labs_api/labs/api/artist_country_from_artist_mbid.py

stragu:
Thanks you for the explanation! That makes perfect sense.

I was just thinking of an added functionality that could be fun, further down the line: hovering on a country with zero artists associated to it could reveal a message like "You haven't listened anything from this country. Click to discover new artists!"

Just an idea that would tie into the recommendation engine nicely! Anyway, I know I should go to Jira for this, sorry... :smile:

PR #986, opened 2020-07-26
Send html enable/disable flag for candidate sets with spark request

vansika:
We'd want to control HTML generation for candidate sets with spark requests. In this PR I have added the functionality for the same.

### DIFF
--- a/listenbrainz/spark/request_manage.py
+++ b/listenbrainz/spark/request_manage.py
@@ -156,16 +156,18 @@ def request_model(rank, itr, lmbda, alpha):
 @click.option("--days", type=int, default=7, help="Request recommendations to be generated on history of given number of days")
 @click.option("--top", type=int, default=20, help="Calculate given number of top artist.")
 @click.option("--similar", type=int, default=20, help="Calculate given number of similar artist.")
+@click.option("--html", is_flag=False, help='Enable/disable HTML file generation')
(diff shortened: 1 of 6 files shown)


listenbrainz/spark/request_manage.py
```diff
@@ -156,16 +156,18 @@ def request_model(rank, itr, lmbda, alpha):
 @click.option("--days", type=int, default=7, help="Request recommendations to be generated on history of given number of days")
 @click.option("--top", type=int, default=20, help="Calculate given number of top artist.")
 @click.option("--similar", type=int, default=20, help="Calculate given number of similar artist.")
+@click.option("--html", is_flag=False, help='Enable/disable HTML file generation')
```
paramsingh, line 159: >is_flag=False

Should this be True? This is a flag, right?

mayhem: beat me to it. :)

vansika: I wanted a flag whose default value should be `False`.

Any other way?

vansika: Maybe click.bool

mayhem: is_flag=True, default=False

should do it. See https://click.palletsprojects.com/en/7.x/options/

vansika: Thanks🌹

PR #990, opened 2020-07-28
Send recommendations sorted on rating in spark request.

vansika:
Till now, the recommendations were sent in random order to db. 
With this PR, recs will be sent sorted on rating.


paramsingh:
Not neccessary right this second, but for our stats stuff, we've started keeping a schema for the json in the database and the json being sent over RabbitMQ so that it's easy to comprehend. We use pydantic models for this purpose (https://github.com/metabrainz/listenbrainz-server/blob/1b75cdf5/data/model/user_artist_stat.py#L1). It'd be a good idea to create pydantic models for the jsonb you're storing in the db. That way we'll know exactly what the database contains.

vansika:
@paramsingh thanks for the info :)

PR #994, opened 2020-07-29
develop.sh improvements

alastair:
# Problem

Some small fixes and improvements to develop.sh

# Solution

* Add documentation for some missing items
* Fix some bash-lint suggestions
* Use exec to pass the return code back to the calling terminal
* Simplify webserver running
* Add develop commands for bash and flask-shell
* Add volume for storing bash/flask shell history

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,13 +1,9 @@
 from listenbrainz import db
 from listenbrainz.db import timescale as ts
 from listenbrainz import webserver
-from listenbrainz import stats
 from werkzeug.serving import run_simple
-import subprocess
 import os
 import click
-import subprocess
-from urllib.parse import urlsplit
 
 from listenbrainz.utils import safely_import_config
 safely_import_config()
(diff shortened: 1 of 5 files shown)


manage.py
```diff
@@ -22,24 +18,6 @@ def cli():
...
-                   "'DEBUG' value in the config file.")
-def runserver(host, port, debug=False):
-    application = webserver.create_app()
-    run_simple(
-        hostname=host,
-        port=port,
-        application=application,
-        use_debugger=debug,
-        use_reloader=debug,
-        processes=5
```
alastair, line 39: This is a change compared to what was previously here. Now it's only using 1 process. @mayhem added this in https://github.com/metabrainz/listenbrainz-server/commit/610f0cba83405413b9762367d2faa1b857bedc83, but it's not clear why we needed this back in 2017. I've never seen a problem with needing more than 1 process in a development server, either for LB or AB, but it's possible that there was a clear reason for this. We can change it back if needed

mayhem: I like to have more than one process, so that images and other things don't all wait for one process. It makes the debugging a little smoother overall. The rest? 🗑️

alastair: The `flask run` command that I added here uses werkzeug's `threaded=True` flag, which spawns a new thread for each request. This flag is incompatible with `processes`: https://github.com/pallets/werkzeug/blob/1129879c79f8415d4ea9a67d7c5ce2f2e1d28c78/src/werkzeug/serving.py#L691-L692
it seems like they effectively have the same behaviour, so it looks like this works as you want.

PR #997, opened 2020-07-30
Improve HDFS setup and startup scripts

alastair:
# Problem

* When you stop your spark containers and restart them you get an error: `namenode is running as process 1.  Stop it first.`
* When you format the namenode, the datanode stops connecting

# Solution

When a datanode connects to a namenode for the first time, it obtains the namenode's ID, and stores it locally. This means that it doesn't make sense to format a namenode without also formatting the datanode.
In order to simplify the format command, remove the stored data of both the namenode and datanode before formatting, so that the next time the datanode connects it will get the new namenode ID.

When hdfs starts, it writes a pid file. When you stop but don't remove a container, this file remains in the filesystem.
When hdfs starts, it checks if the pid file exists, and if there is a process running with that pid. The problem is that it doesn't check if the process with that pid is actually hdfs.
This causes an error message on second startup:
>  namenode is running as process 1.  Stop it first.

In our case, it's actually the startup script (bash) which has pid 1, It later execs to hdfs.
In this case it's safe to just remove the pid file before starting hdfs, allowing us to stop and restart the hdfs containers without having to do a `docker-compose down` between the two.

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -37,7 +37,8 @@ function invoke_docker_compose_spark {
 }
 
 function format_namenode {
-    invoke_docker_compose_spark down
+    # run in a subshell to swallow the exec
+    (invoke_docker_compose_spark down)
     docker volume rm -f listenbrainzspark_datanode
     docker volume rm -f listenbrainzspark_namenode
     invoke_docker_compose_spark run --rm hadoop-master \
--- a/develop.sh
+++ b/develop.sh
@@ -37,6 +37,9 @@ function invoke_docker_compose_spark {
 }
 
 function format_namenode {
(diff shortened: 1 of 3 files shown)


develop.sh
```diff
@@ -37,6 +37,9 @@ function invoke_docker_compose_spark {
 }
 
 function format_namenode {
+    invoke_docker_compose_spark down
```
alastair, line 40: ugh, this no longer works because I updated invoke_docker_compose_spark to use `exec` in #994. Will have to work out the best way of doing it

alastair: this is fixed, I use a subshell, and document the syntax

PR #998, opened 2020-07-30
Fix bug in spark import code

ishaanshah:
# Problem
Users get the following error when importing listens for the first time
```
hadoop-master_1     | 2020-07-30 11:53:40,816 WARN hdfs.StateChange: DIR* FSDirectory.unprotectedRenameTo: failed to rename /temp to /data/listenbrainz because destination's parent does not exist
```
This happens because `/data/` directory hasn't been created before.

# Solution
Checking if the above mentioned directory exists and creating it if it doesn't, fixes the issue.

### DIFF
--- a/listenbrainz_spark/hdfs/__init__.py
+++ b/listenbrainz_spark/hdfs/__init__.py
@@ -1,8 +1,9 @@
 import os
+import pathlib
 import sys
-import time
 import shutil
 import subprocess
+import time
 from tarfile import TarError
 
 import listenbrainz_spark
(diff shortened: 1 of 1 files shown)


alastair:
@ishaanshah and I talked a bit about how we should improve the workflow of this script, especially with regard to separation of concerns and which methods should be responsible for uploading/deleting local and remote files. I wonder if we could just merge this now and make a new ticket for additional improvements and test it there?

paramsingh:
Sure, sounds good. Do we have a ticket?

ishaanshah:
I have opened a ticket [LB-782](https://tickets.metabrainz.org/browse/LB-682)

PR #1010, opened 2020-08-01
[API endpoint] Show ratings with recommendations.

vansika:
We'd want to show ratings given by CF to recordings along with the recording MBIDs when querying the API endpoint.


listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -24,7 +24,7 @@ class RecommendationArtistType(Enum):
 @crossdomain()
 @ratelimit()
 def get_recommendations(user_name):
-    """ Get recommendations for user ``user_name``.
+    """ Get recommendations sorted on rating and ratings for user ``user_name``.
```
mayhem, line 27: Here we will need to document what the range of the ratings is. Also, negative ratings are kinda weird -- ideally we could scale the ratings so that they fall on a range from 0.0 -> 1.0. (0 being weak, 1.0 being great).

Is this possible?

vansika: Okay. Scaling should be done on the spark end imo. Also, whenever we hit a rating val which is less than -1 and greater than 1 we should log that to sentry like you said on IRC to know the actual range of ratings in time ( I couldn't find it in the docs :() 

So how should the scaling be done? Let's say after scaling a rating is equivalent to 0.98 and there is one more rating which is actually 0.98, how do we distinguish between the two? I am not very clear on this atm.

mayhem: Just make sure that you only ever post one notification to sentry per run -- because we could possibly find thousands of ratings out of range and that would be an email disaster. :)

The scaling is quite simple -- it is effectively only this:

scaled_rating = (rating / 2.0) + .5
if scaled_rating > 1.0 or scaled_rating < -1.0:
    raise ....

clipped_rating = min(max((scaled_rating, -1.0), 1.0)

Then returned clipped_rating -- and yeah, I guess it would be better to do the scaling in spark...

mayhem:
Ready for merging, Sir @paramsingh

PR #1012, opened 2020-08-03
Change time.time to time.monotonic

AbhinavOhri:
# Problem
LB-677

# Solution
When calculating time taken for something use time.monotonic instead of
time.time to avoid errors due to system clock time shifting.

# Action
code review

### DIFF
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -419,7 +419,7 @@ class TimescaleListenStore(ListenStore):
             start_time and end_time (datetime): the range of time for the listens dump.
             temp_dir (str): the dir to use to write files before adding to archive
         """
-        t0 = time.time()
+        t0 = time.monotonic()
         offset = 0
         listen_count = 0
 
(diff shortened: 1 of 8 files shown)


alastair:
This looks good, thanks for the contribution!
I see a number of other instances where time was imported as `from time import time`, and therefore it's just called as `time()`.
For example, see listenbrainz_spark/ftp/download.py, listenbrainz_spark/recommendations/create_dataframes.py, listenbrainz_spark/recommendations/train_models.py, and others.

shivam-kapila:
Hi. I think there might be some other occurrences too. You can refer [here](http://livegrep.metabrainz.org/search/livegrep?q=time.time&fold_case=auto&regex=false&context=true&repo%5B%5D=metabrainz%2Flistenbrainz-server)

AbhinavOhri:
@shivam-kapila The link u have posted doesn't  contain the changes from this PR.Also at some places  I have not replaced time.time() intentionally as it is being used for calculating the epoch timestamp for which time.monotonic() doesn't provide the correct value. Therefore , I have only changed time.time() to time.monotonic() where it is used to calculate time taken to do something or difference of time .

paramsingh:
Hey @AbhinavOhri, could you rebase your changes over master once?

PR #1013, opened 2020-08-05
Add dataset hoster based labs.api.listenbrainz.org

mayhem:
This PR should come as no surprise -- it is based on the data-set-hoster project.

The test cases for this PR are pretty tricky since they would require a whole MB + mappings DB to work. Instead, I mock psycopg2 and feed it the data that would normally be returned from the SQL query. This tests the code added by this PR, but it does not actually test if the SQL query itself works.

I could attempt to create an additional integration test for this IFF the mapping DB is available, but thats also a giant PITA.


paramsingh:
Not merging today, as the release is relatively big already with the bigquery removal. Will merge tomorrow.

paramsingh:
Actually, the last release was smooth enough, let's try this one too.

PR #1016, opened 2020-08-07
Add count parameter while requesting data from labs API

ishaanshah:
# Problem
The labs API returns 100 artist MBIDs by default if count parameter is not provided. This leads to incomplete data shown on `Artist Origin` map.

# Solution
Adding a count parameter to the URL fixes the issue.

# Action
Test it on Beta and if it is stable deploy to production. We should also truncate the `artist_map` column in PG before deploying.

### DIFF
--- a/listenbrainz/webserver/views/stats_api.py
+++ b/listenbrainz/webserver/views/stats_api.py
@@ -721,9 +721,9 @@ def _get_mbids_from_msids(artist_msids: list) -> list:
     request_data = [{"artist_msid": artist_msid} for artist_msid in artist_msids]
     artist_mbids = []
     try:
-        result = requests.post("{}/artist-credit-from-artist-msid/json?count={}"
-                               .format(current_app.config['LISTENBRAINZ_LABS_API_URL'], len(request_data)),
-                               json=request_data)
+        result = requests.post("{}/artist-credit-from-artist-msid/json"
+                               .format(current_app.config['LISTENBRAINZ_LABS_API_URL']),
+                               json=request_data, params={'count': len(request_data)})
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/stats_api.py
```diff
@@ -721,8 +721,9 @@ def _get_mbids_from_msids(artist_msids: list) -> list:
     request_data = [{"artist_msid": artist_msid} for artist_msid in artist_msids]
     artist_mbids = []
     try:
-        result = requests.post(
-            "{}/artist-credit-from-artist-msid/json".format(current_app.config['LISTENBRAINZ_LABS_API_URL']), json=request_data)
+        result = requests.post("{}/artist-credit-from-artist-msid/json?count={}"
```
paramsingh, line 724: Instead of using string formatting, we should pass count into the params parameter of requests.post.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -754,7 +756,8 @@ def _get_country_code_from_mbids(artist_mbids: set) -> list:
     except requests.RequestException as err:
         current_app.logger.error("Error while getting artist_country_codes, {}".format(err), exc_info=True)
         error_msg = ("An error occurred while calculating artist_map data, "
-                     "try setting 'force_recalculate' to 'false' to get a cached copy if available")
+                     "try setting 'force_recalculate' to 'false' to get a cached copy if available"
+                     "Payload: {}".format(request_data))
```
paramsingh, line 760: I'd like to log what the response from the endpoint was too, if it's >=400, could we log the text?

PR #1017, opened 2020-08-07
Scale the CF generated ratings so that they fall on a range from 0.0 -> 1.0.

vansika:
As discussed, we'd want to scale negative ratings (CF generated) and keep a record of ratings (log to sentry) that don't fall in the expected range i.e. rating > 1 or rating < -1.

### DIFF
--- a/listenbrainz_spark/recommendations/tests/test_recommend.py
+++ b/listenbrainz_spark/recommendations/tests/test_recommend.py
@@ -6,7 +6,8 @@ from listenbrainz_spark.tests import SparkTestCase
 from listenbrainz_spark.recommendations import recommend
 from listenbrainz_spark.recommendations import train_models
 from listenbrainz_spark import schema, utils, config, path, stats
-from listenbrainz_spark.exceptions import RecommendationsNotGeneratedException
+from listenbrainz_spark.exceptions import (RecommendationsNotGeneratedException,
+                                           RatingOutOfRangeException)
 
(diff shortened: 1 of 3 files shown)


listenbrainz_spark/recommendations/tests/test_recommend.py
```diff
@@ -165,6 +166,26 @@ def test_get_recommendations_for_user(self, mock_candidate_set, mock_mbids):
...
+
+        self.assertEqual(mbids_and_ratings[0][1], 0.34)
+        self.assertEqual(mbids_and_ratings[1][1], 0.966)
+
+        mbids_and_ratings = [['xxxx', -0.32], ['yyyy', 1.932], ['zzzz', 2.967]]
+
+        with self.assertRaises(RatingOutOfRangeException) as err:
+            recommend.scale_ratings(mbids_and_ratings)
+
+        expected_err_mesaage = 'The following ratings are beyond the expected range i.e rating > 1 or rating < -1: \n{}' \
```
mayhem, line 182: message is misspelled. :)

vansika:
Whoops. I see we may be bombarded with range err messages. Will fix and push.

vansika:
passing `ratings_beyond_range` from `get_recommendations_for_all` can do the trick of logging once for all users. Though I am not very sure atm about how to test the log message in `get_recommendations_for_all` .

PR #1019, opened 2020-08-07
Handle country not present in pycountry more gracefully

paramsingh:
# Problem

500s on the stats api endpoint

https://sentry.metabrainz.org/metabrainz/listenbrainz-beta/issues/97838/?query=is%3Aunresolved

# Solution

Check if the country is None before trying to get it's alpha 3 code

# Action

Wait for tests to pass, merge and deploy. Add tests in a follow-up.


paramsingh:
Self-approving. @ishaanshah, would you mind adding a test as a follow-up?

PR #1022, opened 2020-08-09
LB-518: Allow users to specify both max_ts and min_ts in the get listens endpoint

AbhinavOhri:
# Problem
LB-518

# Solution
In fetch_listens:
Removed the condition checking if both max_ts and min_ts allowing .
In fetch_listens_from_storage:
Modified the if conditions adjusting the value of from_ts and to_ts depending upon the value of which of  the both variables is given. 
In get_listens:
Added a condition to check if both max_ts and min_ts are given then max_ts should be greater than min_ts ,otherwise raising a respective error .

### DIFF
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -237,9 +237,9 @@ class TimescaleListenStore(ListenStore):
             max_timestamp_window = 432000 * time_range
             if from_ts is not None:
                 to_ts = from_ts + max_timestamp_window
-            else:
+            elif to_ts is not None:
                 from_ts = to_ts - max_timestamp_window
-
+                
         query = """SELECT listened_at, track_name, created, data
                      FROM listen
                     WHERE user_name = :user_name """
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -235,9 +235,9 @@ class TimescaleListenStore(ListenStore):
             max_timestamp_window = -1
         else:
             max_timestamp_window = 432000 * time_range
-            if from_ts is not None:
+            if to_ts is None:
                 to_ts = from_ts + max_timestamp_window
-            elif to_ts is not None:
+            elif from_ts is None:
                 from_ts = to_ts - max_timestamp_window
                 
         query = """SELECT listened_at, track_name, created, data
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -235,7 +235,9 @@ class TimescaleListenStore(ListenStore):
             max_timestamp_window = -1
         else:
             max_timestamp_window = 432000 * time_range
-            if from_ts is not None:
+            if from_ts and to_ts:
+                pass
+            elif from_ts is not None:
                 to_ts = from_ts + max_timestamp_window
             elif to_ts is not None:
                 from_ts = to_ts - max_timestamp_window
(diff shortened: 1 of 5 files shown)


paramsingh:
Hey @AbhinavOhri, thanks for the contribution! In general, it is a good idea to have a PR title that explains exactly what the PR does. For example, I've changed the title here from "Max to min" to something more descriptive.

AbhinavOhri:
I have added the tests @paramsingh

mayhem:
@alastair  : Back during the inaugural LB hacking day I remember there being a reason for not letting both min and max ts be specified. I don't recall why or the circumstances as to why not. I think the big ticket gives a compelling reason and timescale likes things bound, so I can't see any reason not to allow this.

paramsingh:
@AbhinavOhri, can you rebase this over master once?

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -235,17 +235,23 @@ def fetch_listens_from_storage(self, user_name, from_ts, to_ts, limit, order, ti
             max_timestamp_window = -1
         else:
             max_timestamp_window = 432000 * time_range
-            if from_ts is not None:
+            if from_ts and to_ts:
+                pass
```
alastair, line 239: we don't need an if statement which just has a `pass` here, this can be removed.

listenbrainz/webserver/views/api.py
```diff
@@ -101,8 +101,8 @@ def get_listens(user_name):
     min_ts = _parse_int_arg("min_ts")
     # if no max given, use now()
 
-    if max_ts and min_ts:
-        log_raise_400("You may only specify max_ts or min_ts, not both.")
+    if max_ts and min_ts and max_ts < min_ts:
+        log_raise_400("If both are given, max_ts should be greater than min_ts")
```
alastair, line 105:
```suggestion
        log_raise_400("max_ts should be greater than min_ts")
```

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -235,17 +235,23 @@ def fetch_listens_from_storage(self, user_name, from_ts, to_ts, limit, order, ti
...
         query = """SELECT listened_at, track_name, created, data
                      FROM listen
                     WHERE user_name = :user_name """
+                  
 
         if max_timestamp_window < 0:
-            if from_ts is not None:
+            if from_ts and to_ts:
+                query += """AND listened_at > :from_ts
+                            AND listened_at < :to_ts """
```
alastair, line 253: We talked recently about limiting the maximum duration that you can query (to about a year I think?). Does the web API or this method perform any checking of this limit?

alastair:
I found the original design document, and it only mentioned max_id, so I don't think there was a concrete reason for preventing it, unless there was a technical limitation with our tools of the time.

listenbrainz/listenstore/timescale_listenstore.py
mayhem: The API performs this check -- this limit should not exist on the listenstore level, since that would preclude internal processes from accessing larger time ranges.

mayhem:
Then we should resolve the merge conflict and your comments and then I am happy to merge this.

listenbrainz/listenstore/timescale_listenstore.py
alastair: This will have to be rebased to pull the new range checking code in - however, the current code only checks if the `time_range` parameter is less than a set maximum:  https://github.com/metabrainz/listenbrainz-server/blame/c9067520546a77146e992f8e08a4647a9d1c5e1d/listenbrainz/webserver/views/api.py#L107

this PR will also need to check in the API endpoint that the time between from_ts and to_ts is less than this range.

AbhinavOhri:
i have made the changes .Please review again.@paramsingh @alastair @mayhem

PR #1024, opened 2020-08-10
MSB Mapping

mayhem:
This is the first cut at the MSB mapping -- this work should be complete and merge able. For now the mapping is in production on williams and being generated daily (mostly for observation).

There is still some work left to do in the future PR:

https://tickets.metabrainz.org/browse/LB-696


paramsingh:
Maybe this should be in the `messybrainz` sub-project instead of `listenbrainz`? what do you think?

listenbrainz/msb_mapping/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+To inspect the cron log of the container do:
+
+```docker exec -it messybrainz-mapper python manage.py cron-log```
+
+
+Future improvements
+-------------------
+
+Very soon, (in the next few weeks at the latest) this code needs a few more improvements:
```
paramsingh, line 28:
```suggestion
This code needs a few more improvements:
```

A lot of the times we put these time ranges in stuff that people read and real life happens, people forget etc etc. I'd prefer to not have deadlines in READMEs.

mayhem:
> Maybe this should be in the messybrainz sub-project instead of listenbrainz? what do you think?

My thinking has been more along the lines of "do less with MessyBrainz" and more with "ListenBrainz". Part of this is the realization that MSIDs are effectively useless for anything other than referential integrity inside MessyBrainz and for looking up metadata for an MSID that we don't have. My goal therefore has become to resolve to MBIDs as quickly as possible and to prefer working in MBID space. To that affect, the mapping is more an LB project for me, than it is a MSB project.

paramsingh:
Sorry, i've been late on this, have started reading it, it's on my radar for this weekend.

PR #1025, opened 2020-08-10
LB-695: Improve domain for "Artist Origin" graph

ishaanshah:
# Problem
The domain on the artist map graph should be more reasonable.

# Solution
Using log scale fixes the issue


listenbrainz/webserver/static/js/src/stats/Choropleth.tsx
```diff
@@ -70,28 +75,65 @@ export default function CustomChoropleth(props: ChoroplethProps) {
...
+
+    const result = [];
+    for (let i = 0; i < 6; i += 1) {
+      result.push(
+        Math.ceil(Math.E ** ((Math.log(maxArtistCount) / 6) * (i + 1)))
+      );
+    }
+
+    return result;
+  })();
```
paramsingh, line 92: Why is this a function that you call? Could we just do it in the render itself?

PR #1032, opened 2020-08-12
Show number of listens submitted today and yesterday on status page.

paramsingh:
# Problem

I'd like to know how many listens get submitted to ListenBrainz on average per day.

# Solution

Store that data in redis. Show current day's listens AND yesterday's so that it's easy to compare.

**Screenshot**

Screenshot from 2020-08-12 22-40-45

### DIFF
--- a/listenbrainz/webserver/templates/index/current-status.html
+++ b/listenbrainz/webserver/templates/index/current-status.html
@@ -28,6 +28,14 @@
             <td>{{ listen_count }}</td>
           </tr>
         {% endif %}
+        {% if listen_counts_per_day %}
+          {% for data in listen_counts_per_day %}
+          <tr>
+            <td>Number of listens submitted on {{ data['date'] }}</td>
+            <td>{{ data['listen_count'] }}</td>
+          </tr>
+          {% endfor %}
+        {% endif %}
         </tbody>
       </table>
 
--- a/listenbrainz/webserver/templates/index/current-status.html
+++ b/listenbrainz/webserver/templates/index/current-status.html
@@ -31,7 +31,7 @@
         {% if listen_counts_per_day %}
           {% for data in listen_counts_per_day %}
           <tr>
-            <td>Number of listens submitted on {{ data['date'] }}</td>
+            <td>Number of listens submitted {{ data['label'] }} ({{ data['date'] }})</td>
             <td>{{ data['listen_count'] }}</td>
           </tr>
           {% endfor %}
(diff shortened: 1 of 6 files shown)


listenbrainz/webserver/templates/index/current-status.html
```diff
@@ -28,6 +28,14 @@ <h3>ListenBrainz Stats</h3>
             <td>{{ listen_count }}</td>
           </tr>
         {% endif %}
+        {% if listen_counts_per_day %}
```
mayhem, line 31: The screenshot of current status shows today and yesterday's listens, but that isn't reflected in the PR here. Missing commit?

listenbrainz/webserver/views/index.py
```diff
@@ -103,11 +106,26 @@ def current_status():
     except DatabaseException as e:
         user_count = 'Unknown'
 
+    listen_counts_per_day: List[dict] = []
+    for delta in range(2):
```
mayhem, line 110: constant in code?

listenbrainz/webserver/views/index.py
```diff
@@ -103,11 +106,26 @@ def current_status():
...
+        try:
+            day = datetime.utcnow() - relativedelta(days=delta)
+            day_listen_count = _redis.get_listen_count_for_day(day)
+        except:
+            current_app.logger.error("Could not get today's listen count from redis", exc_info=True)
```
mayhem, line 115: This error message could be wrong half of the time. Maybe give the date of the day that can't be fetched instead of saying "today's"?

listenbrainz/webserver/templates/index/current-status.html
paramsingh: `listen_counts_per_day` is a list, and we loop through it later in the template.

listenbrainz/webserver/views/index.py
paramsingh: I'd say let's keep it like this for now, we'll have to change some stuff anyways if we want to show more data.

listenbrainz/webserver/views/index.py
paramsingh: whoops, missed that.

listenbrainz/webserver/templates/index/current-status.html
mayhem: Oh heh, missed the loop.

PR #1033, opened 2020-08-12
LB-431: Save recording mbids from lastfm in a specific field, instead of saving in the recording mbid field

paramsingh:
# Problem

https://community.metabrainz.org/t/lb-431-last-fm-api-returns-track-mbid-instead-of-recording-mbid-for-new-scrobbles/431016/4

People have been complaining about this. 

# Solution

new last fm imported listens will now have the lastfm provided mbid in a custom field, not in the default recording mbid field, which means more accurate data in the future. 

# Action

This doesn't close the ticket, as we'll probably want to clean the old data up at some point.

### DIFF
--- a/listenbrainz/webserver/static/js/src/Scrobble.ts
+++ b/listenbrainz/webserver/static/js/src/Scrobble.ts
@@ -91,7 +91,7 @@ export default class Scrobble {
         release_name: this.releaseName(),
         additional_info: {
           listening_from: "lastfm",
-          recording_mbid: this.trackMBID(),
+          lastfm_recording_mbid: this.trackMBID(),
           lastfm_release_mbid: this.releaseMBID(),
           lastfm_artist_mbid: this.artistMBID(),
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/static/js/src/Scrobble.ts
```diff
@@ -91,7 +91,7 @@ export default class Scrobble {
         release_name: this.releaseName(),
         additional_info: {
           listening_from: "lastfm",
-          recording_mbid: this.trackMBID(),
+          lastfm_recording_mbid: this.trackMBID(),
```
ishaanshah, line 94: Maybe we should rename this field as `lastfm_track_mbid`?

paramsingh: Makes sense.

paramsingh:
ptal @ishaanshah

PR #1039, opened 2020-08-17
Stop request consumer from dying 

vansika:
# Problem
We have been using sys.exit(-1) everywhere after catching an exception to terminate the job which breaks the request consumer.

# Solution
`raise` the error instead to log it to sentry and gracefully terminate the job

### DIFF
--- a/listenbrainz_spark/recommendations/candidate_sets.py
+++ b/listenbrainz_spark/recommendations/candidate_sets.py
@@ -110,7 +110,7 @@ def get_listens_to_fetch_top_artists(mapped_listens_df, from_date, to_date):
     return mapped_listens_subset
 
 
-def get_top_artists(mapped_listens_subset, top_artist_limit, users):
+def get_top_artists(mapped_listens_subset, top_artist_limit, users, from_date, to_date):
     """ Get top artists listened to by users who have a listening history in
         the past X days where X = RECOMMENDATION_GENERATION_WINDOW.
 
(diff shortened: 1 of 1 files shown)


mayhem:
This may need rebasing over master, since the PR contains a bunch of non-exit related stuff.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -149,8 +149,20 @@ def get_top_artists(mapped_listens_subset, top_artist_limit, users):
                                                          'user_name',
                                                          'total_count') \
                                                  .where(top_artist_df.user_name.isin(users))
+        try:
```
paramsingh, line 152: why are we doing this? are we trying to check if the top_artist_df is not empty? is there a more obvious way to do this? We should extract this into a function at least.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -207,6 +219,11 @@ def get_similar_artists(top_artist_df, artist_relation_df, similar_artist_limit)
                                                       'user_name') \
                                               .distinct()
 
+    try:
+        similar_artist_df.take(1)[0]
```
paramsingh, line 223: would prefer to do this in a seperate function

listenbrainz_spark/recommendations/candidate_sets.py
vansika: Yes. I am not sure about any other way. Can dump this into a different function though.

paramsingh: Can we just make the is empty dataframe function return a book and use that, instead of all the try excepts?

paramsingh (changes requested):

vansika:
@paramsingh what? I didn't get you.

PR #1043, opened 2020-08-19
Add feedback controls

shivam-kapila:
Add generic react component to handle listens.
Add feedback components to submit feedback (based on the generic component itself).


listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -14,25 +14,78 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
...
+    };
+
+    this.APIService = new APIService(
+      props.apiUrl || `${window.location.origin}/1`
+    );
+
     this.playListen = props.playListen.bind(this, props.listen);
   }
 
+  handleClickLove = async () => {
```
paramsingh, line 59: This and handleClickHate should be merged into one function with feedback as arg

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -14,25 +14,78 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
...
+      "track_metadata.additional_info.recording_msid"
+    );
+
+    const status = await this.APIService.submitFeedback(
+      user.auth_token,
+      recordingMSID,
+      score
+    );
+
+    if (status === 200) {
```
paramsingh, line 84: What happens if status is not 200? Does the user know somehow?

paramsingh (changes requested):

listenbrainz/db/feedback.py
```diff
@@ -155,3 +156,35 @@ def get_feedback_count_for_recording(recording_msid: str):
...
+    """
+
+    args = {"user_id": user_id, "recording_list": recording_list}
+    query = """ WITH rf AS (
+                  SELECT user_id, recording_msid::text, score
+                    FROM recording_feedback
+                   WHERE recording_feedback.user_id=:user_id
+                )
+                SELECT COALESCE(rf.user_id, :user_id) AS user_id, "user".musicbrainz_id AS user_name,
+                       rec_msid AS recording_msid, COALESCE(rf.score, 0) AS score
```
vansika, line 181: Okay, so I have been reading the feedback (db + API + react) code for some time now and there are a few things that must be changed. You can either work on this
or open a ticket so that these *can be buggy in future features* can catch someone's attention.

(I wasn't able to find all the PRs related to feedback work so I am commenting everything here)

1. There should be separate endpoints for feedback submit and feedback delete. The following json will ideally be interpreted as a submission irrespective of the score. {'recording_msid': 'xxx', score: '0'}
Keeping 1, -1 for submission and 0 for deletion and passing through the same endpoint is not a good idea. Also, you are simply checking in the feedback_api script if the submitted score is in [-1, 0, 1] but there is no constraint in the DB to ensure the same. I would recommend making separate endpoint for feedback submission and deletion and also to have DB constraints for the score submitted.

2. You have allowed -1, 0, 1 as valid score values to be submitted to the API endpoint. But here in this query, you are returning a pseudo score as 0 for recording msids that the user has not rated. Then in react at some places, you have used 0 to delete feedback via the endpoint and at some places, you have used 0 to reflect that the feedback for that recording was not submitted. The whole usage of 0 is very confusing. Even though the code works perfectly rn, we may have bugs in the future which are because of this confusing use of 0 or are difficult to track because of this confusing use of 0.

3. Almost at all the places the previous state has been mutated like this https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/RecentListens.tsx#L520. You should ideally make a copy of the previous state and then do setState.

PR #1050, opened 2020-08-24
Use pointer cursor for listen control buttons

phw:
# Problem
The buttons for listen control (love / unlove / ellipsis menu) do not indicate they are clickable. Having the cursor change to a pointer is such an indication.

# Solution
This makes it clearer for the user they can interact with those elements by clicking.

# Action


paramsingh (approved):
Thanks!

PR #1051, opened 2020-08-24
LB-708: Merge the __init__.py and utils.py in stats directory.

AbhinavOhri:
# Problem

LB-708

# Solution
Merged utils.py and init.py in stats directory


listenbrainz_spark/stats/__init__.py
```diff
@@ -6,98 +6,4 @@
...
-def get_month_end(month: datetime) -> datetime:
-    """ Returns a datetime object denoting the end of the month """
-    _, num_of_days = monthrange(month.year, month.month)
-    return datetime(month.year, month.month, num_of_days, hour=23, minute=59, second=59)
-
-
-def get_year_end(year: int) -> datetime:
-    """ Returns a datetime object denoting the end of the year """
-    return datetime(year, month=12, day=31, hour=23, minute=59, second=59)
+from listenbrainz_spark.stats import utils
```
paramsingh, line 9: This isn't ideal, I'd prefer to move these functions to utils and then change the places where they're imported to utils as well. Right now, if i want to get to the definition of a function, I'd have to go here, figure out that this file imports utils and then go to utils. if we fix the imports, the reader can directly reach utils.

paramsingh (changes requested):

paramsingh (approved):

paramsingh:
FYI: I had to revert this in https://github.com/metabrainz/listenbrainz-server/commit/a12670663b5ed655b3e7299906d6ef7f1cea4e17, looks like the master build failed because of some imports

AbhinavOhri:
@paramsingh  Should I try to fix this and open another PR?

PR #1052, opened 2020-08-25
LB-628: Add support for incremental dumps

ishaanshah:
# Problem
The frequency at which stats are updated is very less.

# Solution
Adding support for incremental dumps will solve the issue.

### DIFF
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -196,7 +196,7 @@ def handle_dump_imported(data):
     dump_name = data['imported_dump']
     import_completion_time = data['time']
     send_mail(
-        subject='A data dump has been imported into the Spark cluster',
+        subject='A {} has been imported into the Spark cluster'.format(' '.join(data['type'].split('_')[1:])),
         text=render_template('emails/dump_import_notification.txt', dump_name=dump_name, time=import_completion_time),
         recipients=['listenbrainz-observability@metabrainz.org'],
         from_name='ListenBrainz',
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -20,7 +20,6 @@ from data.model.user_recording_stat import UserRecordingStatJson
 from data.model.user_missing_musicbrainz_data import UserMissingMusicBrainzDataJson
 
 
-
 TIME_TO_CONSIDER_STATS_AS_OLD = 20  # minutes
 TIME_TO_CONSIDER_RECOMMENDATIONS_AS_OLD = 7  # days
 
@@ -187,7 +186,7 @@ def handle_sitewide_entity(data):
 
 
 def handle_dump_imported(data):
-    """ Process the response that the cluster sends after importing a new full dump
+    """ Process the response that the cluster sends after importing a new dump
 
     We don't really need to _do_ anything, just send an email over for observability.
     """
@@ -197,7 +196,7 @@ def handle_dump_imported(data):
     dump_name = data['imported_dump']
     import_completion_time = data['time']
     send_mail(
-        subject='A full data dump has been imported into the Spark cluster',
+        subject='A data dump has been imported into the Spark cluster',
         text=render_template('emails/dump_import_notification.txt', dump_name=dump_name, time=import_completion_time),
         recipients=['listenbrainz-observability@metabrainz.org'],
         from_name='ListenBrainz',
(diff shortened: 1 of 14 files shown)


listenbrainz_spark/hdfs/__init__.py
```diff
@@ -62,6 +62,17 @@ def upload_archive(self, tmp_dump_dir, tar, dest_path, schema, callback=None, fo
         if callback is None:
             raise NotImplementedError('Callback to process JSON missing. Aborting...')
 
+        # Delete '/temp'if it exists
+        if utils.path_exists("/temp"):
```
paramsingh, line 66: the `/temp` dir here should be a CONSTANT up at the top of the file imo, it's used in a bunch of places.

listenbrainz_spark/utils.py
```diff
@@ -311,3 +311,23 @@ def rename(hdfs_src_path: str, hdfs_dst_path: str):
...
+
+def copy(hdfs_src_path: str, hdfs_dst_path: str, overwrite: bool = False):
+    """ Copy a file or folder in HDFS
+
+        Args:
+            hdfs_src_path – Source path.
+            hdfs_dst_path – Destination path. If the path already exists and is a directory, the source will be copied into it.
+            overwrite - Wether to overwrite the path if it already exists.
+    """
+    print(hdfs_src_path)
```
paramsingh, line 324: this is extraneous.

listenbrainz_spark/utils.py
```diff
@@ -311,3 +311,23 @@ def rename(hdfs_src_path: str, hdfs_dst_path: str):
...
+    walk = hdfs_walk(hdfs_src_path)
+
+    for (root, dirs, files) in walk:
+        for _file in files:
+            src_file_path = os.path.join(root, _file)
+            dst_file_path = os.path.join(hdfs_dst_path, src_file_path.replace(hdfs_src_path, '', 1).strip("/"))
```
paramsingh, line 330: what's the replace here for? Maybe this should be using the os.path functions instead?

listenbrainz_spark/hdfs/upload.py
```diff
@@ -46,43 +47,48 @@ def process_json_listens(self, filename, data_dir, tmp_hdfs_path, schema):
             month = filename.split('/')[-1][0:-5]
             dest_path = os.path.join(data_dir, year, '{}.parquet'.format(str(month)))
 
+        if utils.path_exists(dest_path):
```
paramsingh, line 50: This doesn't make sense in the first read, I figure we should have a parameter passed to the function which tells it to append instead.

listenbrainz_spark/hdfs/upload.py
```diff
@@ -46,43 +47,48 @@ def process_json_listens(self, filename, data_dir, tmp_hdfs_path, schema):
...
+                self.upload_archive(tmp_dump_dir, tar, path.MBID_MSID_MAPPING, schema.msid_mbid_mapping_schema,
+                                    self.process_json, force=True)
 
     def upload_listens(self, archive, force=False):
         """ Decompress archive and upload listens to HDFS.
 
             Args:
                 archive: listens tar file to upload.
-                force: If True deletes dir at path where listens are to be uploaded.
+                force: If True deletes dir at path where listens are to be uploaded,
```
paramsingh, line 75: We should consider renaming force here to `incremental` or something like that

paramsingh (changes requested):

listenbrainz_spark/utils.py
ishaanshah: Removed

listenbrainz_spark/hdfs/__init__.py
ishaanshah: Added it

listenbrainz_spark/hdfs/upload.py
ishaanshah: Added

listenbrainz_spark/hdfs/upload.py
ishaanshah: Renamed it to `overwrite`

listenbrainz_spark/utils.py
ishaanshah: Changed to `os.path.relpath`.

listenbrainz/spark/handlers.py
```diff
@@ -197,7 +196,7 @@ def handle_dump_imported(data):
     dump_name = data['imported_dump']
     import_completion_time = data['time']
     send_mail(
-        subject='A full data dump has been imported into the Spark cluster',
+        subject='A data dump has been imported into the Spark cluster',
```
paramsingh, line 199: can we put in info on what type of dump this was in the subject as well? it's helpful when searching through the inbox.

paramsingh (approved):

PR #1053, opened 2020-08-25
LB-598: Add option to download feedback as json

shivam-kapila:
Add an option on Export page to export feedback.

### DIFF
--- a/listenbrainz/db/model/feedback.py
+++ b/listenbrainz/db/model/feedback.py
@@ -1,6 +1,6 @@
 import uuid
 
-from typing import List
+from datetime import datetime
 from pydantic import BaseModel, ValidationError, validator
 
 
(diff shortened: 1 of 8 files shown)


listenbrainz/db/model/feedback.py
```diff
@@ -10,13 +10,15 @@ class Feedback(BaseModel):
...
-            score: The score associated with the recording (+1/-1 for love/hate respectively)
+            score: the score associated with the recording (+1/-1 for love/hate respectively)
+            created: (Optional)the timestamp when the feedback record was inserted into DB
     """
 
     user_id: int
     user_name: str = None
     recording_msid: str
     score: int
+    created: datetime = None
```
paramsingh, line 21: This should be required, no?

paramsingh: why is it set to none by default

listenbrainz/webserver/views/profile.py
```diff
@@ -168,6 +185,25 @@ def export_data():
...
+    """ Exporting the feedback data to json """
+    filename = current_user.musicbrainz_id + "_lb_feedback-" + datetime.today().strftime('%Y-%m-%d') + ".json"
+
+    # Build a generator that streams the json response. We never load all
+    # feedback into memory at once, and we can start serving the response
+    # immediately.
+    feedback = fetch_feedback(current_user.id)
+    output = stream_json_array(_feedback_to_api(fb=fb) for fb in feedback)
+
+    response = Response(stream_with_context(output))
```
paramsingh, line 200: this code is duplicated from the export listens function right? We should extract it into a seperate function.

paramsingh (changes requested):
The delete your account page also has a button to export your listens, we should add the export feedback button there too

listenbrainz/db/model/feedback.py
shivam-kapila: Because we also validate the feedback submitted via API using this model and it doesnt have a created field

paramsingh (approved):

PR #1054, opened 2020-08-26
Improve spotify imports

shivam-kapila:
Move to native spotipy methods
Import songs only after latest_listen_ts

### DIFF
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -123,7 +123,7 @@ def make_api_request(user, spotipy_call, **kwargs):
 
     Args:
         user (spotify.Spotify): the user whose plays are to be imported.
-        spotipy_call (function call): the Spotipy call which makes request to the required API endpoint
+        spotipy_call (function): the Spotipy function which makes request to the required API endpoint
 
     Returns:
         the response from the spotify API
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -194,7 +194,7 @@ def get_user_recently_played(user):
     """ Get tracks from the current user’s recently played tracks.
     """
     latest_listened_at_ts = int(user.latest_listened_at.timestamp() * 1000)  # latest listen UNIX ts in ms
-    return  user.get_spotipy_client().current_user_recently_played(limit=50)
+    return  user.get_spotipy_client().current_user_recently_played(limit=50, after=latest_listened_at_ts)
 
 
 def get_user_currently_playing(user):
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -17,6 +17,7 @@ from werkzeug.exceptions import BadRequest, InternalServerError, ServiceUnavaila
 from brainzutils.mail import send_mail
 from brainzutils import musicbrainz_db
 from brainzutils.musicbrainz_db import editor as mb_editor
+from datetime import datetime
 
 
 def notify_error(musicbrainz_row_id, error):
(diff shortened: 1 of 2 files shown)


listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -117,12 +118,12 @@ def _convert_spotify_play_to_listen(play, listen_type):
...
 
 
-def make_api_request(user, endpoint, **kwargs):
+def make_api_request(user, spotipy_call, **kwargs):
     """ Make an request to the Spotify API for particular user at specified endpoint with args.
 
     Args:
         user (spotify.Spotify): the user whose plays are to be imported.
-        endpoint (str): the Spotify API endpoint to which the request is to be made
+        spotipy_call (function call): the Spotipy call which makes request to the required API endpoint
```
paramsingh, line 126: function call is not a valid type. this would take a function, right?

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -193,13 +193,14 @@ def make_api_request(user, endpoint, **kwargs):
 def get_user_recently_played(user):
     """ Get tracks from the current user’s recently played tracks.
     """
-    return make_api_request(user, 'me/player/recently-played', limit=50)
+    latest_listened_at_ts = int(user.latest_listened_at.timestamp() * 1000)  # latest listen UNIX ts in ms
+    return  user.get_spotipy_client().current_user_recently_played(limit=50, after=latest_listened_at_ts)
```
paramsingh, line 197: this bypasses the retry logic that is implemented in `make_api_request`. what we should do is pass the current_user_recently_played function into `make_api_request`

```suggestion
    return  make_api_request(user.get_spotipy_client().current_user_recently_played, limit=50, after=latest_listened_at_ts)
```

something like this

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -193,13 +193,14 @@ def make_api_request(user, endpoint, **kwargs):
...
 
 
 def get_user_currently_playing(user):
     """ Get the user's currently playing track.
     """
-    return make_api_request(user, 'me/player/currently-playing')
+    return user.get_spotipy_client().current_user_playing_track()
```
paramsingh, line 203: Same comment here, this should use make_api_request as well.

paramsingh (changes requested):
Let's add a few tests that make sure the `current_user_recently_played` and `current_user_playing_track` are called with the correct parameters

listenbrainz/spotify_updater/spotify_read_listens.py
shivam-kapila: done

listenbrainz/spotify_updater/spotify_read_listens.py
shivam-kapila: done

listenbrainz/spotify_updater/spotify_read_listens.py
shivam-kapila: done

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -193,14 +194,18 @@ def make_api_request(user, endpoint, **kwargs):
 def get_user_recently_played(user):
     """ Get tracks from the current user’s recently played tracks.
     """
-    return make_api_request(user, 'me/player/recently-played', limit=50)
+    latest_listened_at_ts = 0
+    latest_listened_at = user.latest_listened_at
+    if latest_listened_at:
```
paramsingh, line 199: why not just `if user.latest_listened_at` here? the extra variable declaration makes this part of the code confusing, specially because there's no real difference between the names of the two variables.

shivam-kapila: done

paramsingh (changes requested):
Let's bring back the test case that checked the after value when there was no latest_listened_at

paramsingh (approved):

PR #1062, opened 2020-08-29
Don't spam with recommendations error mails

vansika:
# Problem

error for each user is being logged to sentry leading to hundreds of emails. Also, we don't have any mail to inform about the number of users for whom recs have been generated and total time in generating recs.

# Solution
 - store all users_names in a list for whom recs have not been generated and log a single message to sentry.
- send an email informing total user count and total time in generating recs.

### DIFF
--- a/listenbrainz_spark/recommendations/recommend.py
+++ b/listenbrainz_spark/recommendations/recommend.py
@@ -83,7 +83,7 @@ def load_model():
         raise
 
 
-def get_recording_mbids(params, recommendation_df):
+def get_recording_mbids(params: RecommendationParams, recommendation_df):
     """ Get recording mbids corresponding to recommended recording ids sorted on rating.
 
         Args:
@@ -100,7 +100,7 @@ def get_recording_mbids(params, recommendation_df):
     return recording_mbids_df
 
 
-def generate_recommendations(candidate_set, params, limit):
+def generate_recommendations(candidate_set, params: RecommendationParams, limit):
     """
         Args:
             candidate_set (rdd): RDD of user_id and recording_id.
@@ -139,7 +139,7 @@ def get_recommendation_df(recording_ids_and_ratings):
     return df
 
 
-def scale_ratings(mbids_and_ratings, params):
+def scale_ratings(mbids_and_ratings, params: RecommendationParams):
     """ Scale the ratings so that they fall on a range from 0.0 -> 1.0.
 
         Args:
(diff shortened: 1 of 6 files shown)


listenbrainz_spark/recommendations/recommend.py
```diff
@@ -290,7 +291,7 @@ def get_message_for_inactive_users(messages, active_users, users):
     return messages
 
 
-def get_recommendations_for_all(params, users):
+def get_recommendations_for_all(params, users, ti):
```
paramsingh, line 294: should add an explanation in the docstring here of what ti is. Also, we generally use ts for timestamps

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -367,7 +391,8 @@ def main(recommendation_top_artist_limit=None, recommendation_similar_artist_lim
                                   recommendation_top_artist_limit,
                                   recommendation_similar_artist_limit)
 
-    messages = get_recommendations_for_all(params, users)
+    ti = time.monotonic()
+    messages = get_recommendations_for_all(params, users, ti)
```
paramsingh, line 395: Do we actually need to pass the ti from here? Can't we just  calculate it in the get_recommendations function? If we can, we should do that, because the extra parameter makes the function harder to use.

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -134,32 +139,31 @@ def get_recommendation_df(recording_ids_and_ratings):
...
 
 
-def scale_ratings(mbids_and_ratings, ratings_beyond_range):
+def scale_ratings(mbids_and_ratings, params):
     """ Scale the ratings so that they fall on a range from 0.0 -> 1.0.
 
         Args:
             mbids_and_ratings: list of recommended recording mbids and ratings.
-            ratings_beyond_range: list to update with ratings beyond expected range.
+            params: RecommendationParams class object.
```
paramsingh, line 147: We should get into the habit of adding these as type annotations. That way when a person is reading the code in a text editor, they can just control+click to go to the  definition of the class.

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -134,32 +139,31 @@ def get_recommendation_df(recording_ids_and_ratings):
     return df
 
 
-def scale_ratings(mbids_and_ratings, ratings_beyond_range):
+def scale_ratings(mbids_and_ratings, params):
```
paramsingh, line 142:
```suggestion
def scale_ratings(mbids_and_ratings, params: RecommendationParams):
```

paramsingh (changes requested):

listenbrainz_spark/recommendations/recommend.py
paramsingh: Something like https://github.com/metabrainz/listenbrainz-server/pull/1062#discussion_r479773303

paramsingh (approved):

PR #1071, opened 2020-09-02
log recording count of top artist and  similar artist candidate sets

vansika:
log recording count of top artist and similar artist candidate sets. Skipping a test till the time we are testing for runtime.


mayhem (approved):
LGTM

PR #1073, opened 2020-09-03
Generate recommendations for all users at once.

vansika:
# Problem
It took around 16 hours to generate recommendations for around 1700 users. We have to generate recommendations for each user separately since the candidate set for each user is different but the operations which are performed on each set of recommendations can definitely be done for all users at once. I think it will reduce time to some extent.

# Solution
Get mbids corresponding to recording_ids for all users at once.
Scale the ratings for all users at once.

***EDIT***
I have lately realised that recommendations can be generated for all users at once. So in this script, we are performing all operations for all users at once. I am expecting runtime to fall greatly.

### DIFF
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -356,12 +356,15 @@ def notify_cf_recording_recommendations_generation(data):
     if current_app.config['TESTING']:
         return
 
-    user_count = data['user_count']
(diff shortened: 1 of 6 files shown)


mayhem (approved):
as usual hard to follow, but as far as I can tell, it looks ok!

vansika:
@mayhem can we merge, release and test?

PR #1079, opened 2020-09-06
Add functionality so that people are able to follow other users

paramsingh:
# Problem

Users would like to be able to follow other users on ListenBrainz.

More details on the project: https://www.dropbox.com/scl/fi/5ckz83v6goguorpf9uh6p/-%5BProject-Brief%5D-Music-Discovery.paper?dl=0&rlkey=tn38zj49f23ktu1b0lbtndu5n

# Solution

https://drive.google.com/file/d/1nNUh-CAaATZ2_M3q0qZzuQRU6SKr192P/view?usp=sharing

# Action

Right now, the button will only show up for a few users in a list, once we have more functionality I'll open up the button for everyone, for now though, it's ok.

### DIFF
--- a/listenbrainz/db/user_relationship.py
+++ b/listenbrainz/db/user_relationship.py
@@ -21,7 +21,14 @@ from listenbrainz.db.exceptions import DatabaseException
 
 import sqlalchemy
 
+VALID_RELATIONSHIP_TYPES = (
+    'follow',
+)
+
 def insert(user_0: int, user_1: int, relationship_type: str) -> None:
+    if relationship_type not in VALID_RELATIONSHIP_TYPES:
+        raise ValueError(f"Invalid relationship type: {relationship_type}")
+
     with db.engine.connect() as connection:
         connection.execute(sqlalchemy.text("""
             INSERT INTO user_relationship (user_0, user_1, relationship_type)
@@ -49,6 +56,9 @@ def is_following_user(follower: int, followed: int) -> bool:
 
 
 def delete(user_0: int, user_1: int, relationship_type: str) -> None:
+    if relationship_type not in VALID_RELATIONSHIP_TYPES:
+        raise ValueError(f"Invalid relationship type: {relationship_type}")
+
     with db.engine.connect() as connection:
         connection.execute(sqlalchemy.text("""
             DELETE
--- a/listenbrainz/db/user_relationship.py
+++ b/listenbrainz/db/user_relationship.py
@@ -1,0 +1,48 @@
+# listenbrainz-server - Server for the ListenBrainz project.
+#
+# Copyright (C) 2020 Param Singh <iliekcomputers@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+from listenbrainz import db
+from listenbrainz.db.exceptions import DatabaseException
+
+import sqlalchemy
+
+def insert(user_0: int, user_1: int, relationship_type: str):
+    with db.engine.connect() as connection:
+        connection.execute(sqlalchemy.text("""
+            INSERT INTO user_relationship (user_0, user_1, relationship_type)
+                 VALUES (:user_0, :user_1, :relationship_type)
+        """), {
+            "user_0": user_0,
+            "user_1": user_1,
+            "relationship_type": relationship_type,
+        })
+
+
+def already_following_user(follower: int, followed: int) -> bool:
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT COUNT(*)
+              FROM user_relationship
+             WHERE user_0 = :follower
+               AND user_1 = :followed
+               AND relationship_type = 'follow'
+        """), {
+            "follower": follower,
+            "followed": followed,
+        })
+        return result.fetchone() > 0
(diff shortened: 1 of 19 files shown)


listenbrainz/webserver/static/js/src/FollowButton.tsx
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    if (loggedInUserFollowsUser) {
+      if (!hover) {
+        return { buttonClass: "btn btn-sm btn-info", buttonText: "Following" };
+      }
+      return { buttonClass: "btn btn-sm btn-danger", buttonText: "Unfollow" };
+    }
+
+    return { buttonClass: "btn btn-sm btn-warning", buttonText: "Follow" };
```
ishaanshah, line 112: Can we use `btn-primary` instead of `btn-warning` to match the color scheme of the website? Right now the button feels a bit out of place in terms of the color scheme.

listenbrainz/webserver/static/js/src/FollowButton.tsx
```diff
@@ -0,0 +1,137 @@
...
+      };
+    }
+
+    if (justFollowed) {
+      return { buttonClass: "btn btn-sm btn-info", buttonText: "Following" };
+    }
+
+    if (loggedInUserFollowsUser) {
+      if (!hover) {
+        return { buttonClass: "btn btn-sm btn-info", buttonText: "Following" };
```
ishaanshah, line 107: `btn-secondary` instead of `btn-info`?

listenbrainz/webserver/static/js/src/FollowButton.tsx
```diff
@@ -0,0 +1,137 @@
...
+  };
+
+  render() {
+    return (
+      <>
+        <div
+          id="follow-button"
+          onClick={this.handleButtonClick}
+          onKeyPress={this.handleButtonClick}
+          onMouseEnter={this.toggleHover}
```
ishaanshah, line 122: It would be better to pass the hover state in the callback rather than inverting the current state. That way the colors won't end up getting inverted because of some bug in browser code.

listenbrainz/db/user_relationship.py
```diff
@@ -0,0 +1,74 @@
...
+from listenbrainz.db.exceptions import DatabaseException
+
+import sqlalchemy
+
+VALID_RELATIONSHIP_TYPES = (
+    'follow',
+)
+
+
+def insert(user_0: int, user_1: int, relationship_type: str) -> None:
```
mayhem, line 29: Do we have an official policy on docstrings? It is quite clear what the code does and a docstring isn't really needed, IMHO. When do we require them and when not?

mayhem (approved):
All the non-react bits look fine to me.

listenbrainz/webserver/static/js/src/FollowButton.tsx
MonkeyDo: I just tried, and it looks like `btn-primary` is also an orange color.
It also looks like that's what we're using in some other pages for action buttons, so it would be consistent.

However I must say `btn-info` (blue) looks best to me for this action button.

listenbrainz/webserver/static/js/src/FollowButton.tsx
MonkeyDo: Is that the same as `btn-success` (green)?
I would expect green if everything is OK

listenbrainz/webserver/static/js/src/UserPageHeading.tsx
```diff
@@ -0,0 +1,82 @@
...
+  user,
+  loggedInUser,
+  loggedInUserFollowsUser = false,
+}: {
+  user: ListenBrainzUser;
+  loggedInUser: ListenBrainzUser | null;
+  loggedInUserFollowsUser: boolean;
+}) => {
+  return (
+    <>
```
MonkeyDo, line 49: This empty markup shouldn't be necessary.

listenbrainz/webserver/templates/user/base.html
```diff
@@ -4,7 +4,7 @@
 
 {% block content %}
   <div id="user-profile">
-    <h2 class="page-title">{{ user.musicbrainz_id }}</h2>
+    <div id="follow-controls-container"></div>
```
MonkeyDo, line 7: Should maybe be `user-page-heading-container`?

MonkeyDo (commented):
Seems sane to me!

I agree with ishaan about adjusting the button colors.
What I think would be ideal is blue (`btn-info`) for follow action, green (`btn-success`) for following, and orange (`btn-warning`/`btn-primary`) for unfollow action.

I think you could also have icons in the button, which would probably require to have another prop returned by getButtonDetails:
[follow user](https://fontawesome.com/icons/user-plus?style=solid), [following](https://fontawesome.com/icons/user-check?style=solid) and either [this one](https://fontawesome.com/icons/user-minus?style=solid) or [that one](https://fontawesome.com/icons/user-times?style=solid) for unfollowing.

I'm looking forward to this feature, I must say, after our first attempt at 'following'.
Let me know if you want to talk about UX for the rest of the follow feature (like where to show users we're following, etc.)

listenbrainz/webserver/static/js/src/UserPageHeading.tsx
```diff
@@ -0,0 +1,82 @@
...
+// Using this function to limit access to the follow button
+// while it does nothing. Will eventually remove this function.
+// TODO (param, 2020-09-30): remove this function
+const followUsersFeatureEnabled = (currentUser: string): boolean => {
+  const usersAllowedIn = [
+    "rob",
+    "iliekcomputers",
+    "shivam-kapila",
+    "ishaanshah",
+    "Mr_Monkey",
```
MonkeyDo, line 34: I forgot to add: it should be mr_monkey without the uppercase characters.

listenbrainz/webserver/static/js/src/UserPageHeading.tsx
paramsingh: Yep, remnants of some experimentation, removed.

listenbrainz/webserver/templates/user/base.html
paramsingh: Makes total sense. done.

listenbrainz/webserver/static/js/src/UserPageHeading.tsx
paramsingh: Made the feature flag case insensitive :)

listenbrainz/webserver/static/js/src/FollowButton.tsx
paramsingh: Done.

paramsingh:
ALL ABOARD! :steam_locomotive:

PR #1080, opened 2020-09-09
[LB-703] Similar and top artist should be distinct for a user

vansika:
Top artist and similar artists for a user should be distinct so that the recordings in top artists and similar artist playlist don't repeat.


mayhem (approved):
This PR was nice and easy to follow. :)

PR #1082, opened 2020-09-09
Show listen count on Artist Origin map (Backend)

ishaanshah:
This PR adds backend support for coloring  the Artist Origin map by listen counts. Opening a different PR for frontend so that we can safely deploy it in a week from when this PR is deployed.

### DIFF
--- a/listenbrainz/webserver/views/stats_api.py
+++ b/listenbrainz/webserver/views/stats_api.py
@@ -1,3 +1,4 @@
+import bisect
 import calendar
 import json
 from collections import defaultdict
@@ -600,9 +601,16 @@ def get_artist_map(user_name: str):
             top_artists = getattr(artist_stats, stats_range).artists
             for artist in top_artists:
                 if artist.artist_msid is not None:
-                    artist_msids.append(artist.artist_msid)
+                    artist_msids.append({
+                        'artist_msid': artist.artist_msid,
+                        'listen_count': artist.listen_count
+                    })
                 else:
-                    artist_mbids += artist.artist_mbids
+                    for artist_mbid in artist.artist_mbids:
+                        artist_mbids.append({
+                            'artist_mbid': artist_mbid,
+                            'listen_count': artist.listen_count
+                        })
 
             country_code_data = _get_country_codes(artist_msids, artist_mbids)
             result = UserArtistMapStatJson(**{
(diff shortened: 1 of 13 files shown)


listenbrainz/webserver/views/stats_api.py
```diff
@@ -818,7 +826,7 @@ def _get_sitewide_entity_list(
     return sorted(result, key=lambda x: x['from_ts'])
 
 
-def _get_country_codes(artist_msids: list, artist_mbids: list) -> List[UserArtistMapRecord]:
+def _get_country_codes(artist_msids: List[Tuple[str, int]], artist_mbids: List[Tuple[str, int]]) -> List[UserArtistMapRecord]:
```
paramsingh, line 829: the type annotation here seems incorrect, it's a list of dicts.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -861,22 +890,36 @@ def _get_mbids_from_msids(artist_msids: list) -> list:
...
+                    })
         except requests.RequestException as err:
             current_app.logger.error("Error while getting artist_mbids, {}".format(err), exc_info=True)
             error_msg = ("An error occurred while calculating artist_map data, "
                          "try setting 'force_recalculate' to 'false' to get a cached copy if available."
                          "Payload: {}. Response: {}".format(request_data, result.text))
             raise APIInternalServerError(error_msg)
 
+    # Sort the result according to artist_msid and perform binary search to match with appropriate listen_count
+    artist_mbids.sort(key=lambda artist: artist['artist_msid'])
```
paramsingh, line 906: instead of sorting and then applying binary search maybe it's a good idea to construct a dict (or two) which is {artist_msid/mbid: listen_count}, then you can do the lookup easily without the extra logic and bisect stuff.

it'd be much easier to read.

listenbrainz/webserver/views/stats_api.py
ishaanshah: Fixed it.

listenbrainz/webserver/views/stats_api.py
ishaanshah: Yep this makes much more sense, updated.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -595,14 +596,15 @@ def get_artist_map(user_name: str):
...
             # Calculate the data
-            artist_msids = []
-            artist_mbids = []
+            artist_msids = defaultdict(lambda: 0)
+            artist_mbids = defaultdict(lambda: 0)
             top_artists = getattr(artist_stats, stats_range).artists
             for artist in top_artists:
                 if artist.artist_msid is not None:
-                    artist_msids.append(artist.artist_msid)
+                    artist_msids[artist.artist_msid] += artist.listen_count
```
paramsingh, line 604: this would have been cleaner as `artist_msids[artist.artist_msid] = artist.listen_count`, or is there an edge case here that I'm missing?

paramsingh (approved):

listenbrainz/webserver/views/stats_api.py
ishaanshah: We may have multiple entries of the same MSIDs, we should count listens for all of those.

PR #1083, opened 2020-09-10
LB-690: Minor improvements to Listening Activity graph

ishaanshah:
# Problem
- The date shown on hover is not the same as the date on the label, for the listening activity graph in the reports page (month range).
- The weekly listening activity stats are too old, i.e for last to last week and last week.

# Solution
- There was a bug in the code for finding number of days in a month which was causing the issue, the bug has now been fixed.
- Now that incremental dump support has been added, we should compare the stats for current week with last week.


paramsingh (approved):

ishaanshah:
@brainzbot retest this please

PR #1087, opened 2020-09-12
Animate the follow button

MonkeyDo:
I thought the follow icon is quite big next to the name, and with those clear icons and colors I figured we could do something to solve that.
We can show the icon only until the user hovers to reveal the text, with a short animated transition.
Capture d’écran 2020-09-12 à 19 59 08Capture d’écran 2020-09-12 à 19 58 41

Capture d’écran 2020-09-12 à 19 58 59Capture d’écran 2020-09-12 à 19 58 50


paramsingh (approved):
Thank you!

PR #1088, opened 2020-09-13
Update spark config values to try and avoid RPC client error

vansika:
`ERROR TaskSchedulerImpl: Lost executor 2 on 10.0.1.24: Remote RPC client disassociated. Likely due to containers exceeding thresholds, or network issues. Check driver logs for WARN messages.`

I am still not sure what causes the above error in `train_models.py`.  I have tweaked the spark configs.

I also have added a few log messages to understand which process is exactly causing the error.


listenbrainz_spark/recommendations/train_models.py
```diff
@@ -341,31 +345,34 @@ def main(ranks=None, lambdas=None, iterations=None, alpha=None):
 
     # Rdds that are used in model training iterative process are cached to improve performance.
     # Caching large files may cause Out of Memory exception.
-    training_data.persist()
-    validation_data.persist()
+    # commenting out till we are sure that the cached data isn't causing RPC errors.
```
paramsingh, line 348: let's remove the lines, it's easy to get them back from git, and it's generally very easy to forget about things like this.

paramsingh (changes requested):

listenbrainz/webserver/templates/emails/cf_recording_model_upload_notification.txt
```diff
@@ -3,6 +3,6 @@ Hey! 👋
 Data has been successfully trained to create a model.
 
 upload time = {{ time_to_upload }}
-time taken to create model = {{ total_time }}h
+time taken to create model = {{ total_time }}s
```
paramsingh, line 6: unrelated change?

paramsingh (approved):

listenbrainz/webserver/templates/emails/cf_recording_model_upload_notification.txt
vansika: No.

PR #1089, opened 2020-09-14
Make sure incremental dump import is not skipped if RC crashes

ishaanshah:
# Problem
We do not store any information about the current state of listens in HDFS i.e which dumps have been imported. This can easily lead to incorrect state if RC crashes or any other error occurs.

# Solution
Storing information about the dumps imported and their timestamp will solve this issue as we can determine which dumps need to imported.

### DIFF
--- a/listenbrainz_spark/request_consumer/jobs/import_dump.py
+++ b/listenbrainz_spark/request_consumer/jobs/import_dump.py
@@ -45,9 +45,10 @@ def import_full_dump_by_id_handler(id: int):
 def import_newest_incremental_dump_handler():
     imported_dumps = []
     latest_full_dump = utils.get_latest_full_dump()
-    if not latest_full_dump:
+    if latest_full_dump is None:
         # If no prior full dump is present, just import the lates incremental dump
         imported_dumps.append(import_dump_to_hdfs('incremental', overwrite=False))
+        current_app.logger.warn("No previous full dump found, importing latest incremental dump", exc_info=True)
     else:
         # Import all missing dumps from last full dump import
         dump_id = latest_full_dump["dump_id"] + 1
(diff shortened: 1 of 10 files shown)


listenbrainz_spark/request_consumer/jobs/import_dump.py
```diff
@@ -32,16 +35,31 @@ def import_full_dump_by_id_handler(id: int):
...
+        'imported_dump': [dump_name],
         'time': str(datetime.utcnow()),
     }]
 
 
 def import_newest_incremental_dump_handler():
-    dump_name = import_dump_to_hdfs('incremental', overwrite=False)
+    imported_dumps = []
+    latest_full_dump = utils.get_latest_full_dump()
+    if not latest_full_dump:
```
paramsingh, line 46: Just so we're clear, full dump 1 + incremental dump 2 == full dump 2

So what I'm thinking we need to do for all imports is this:
* Figure out which ID the spark cluster is currently at (this doesn't depend on whether the last dump imported was full or incremental), say this ID is x
* If doing an incremental import, import all dumps that are available starting from x + 1
* If doing a full import, import the latest full dump that has value > x

paramsingh (changes requested):

listenbrainz_spark/request_consumer/jobs/import_dump.py
ishaanshah: I have addressed issue number 1 and 2. Solving the 3rd one isn't easy as we don't know the latest dump available to download until we actually start downloading the dump. However this wont be an issue as all incremental dumps after the full dump will be imported on the next call to `import_newest_incremental_dump_handler`.

paramsingh: You can always just list the dumps in the ftp dir and check. I think that should be possible, but I guess we can do it in a future PR.

paramsingh (approved):

PR #1090, opened 2020-09-15
Fix minor bug in artist origin map

ishaanshah:
# Problem
https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/99209/?query=is%3Aunresolved

# Solution
Iterating over items of dictionary fixes the issue.


paramsingh (commented):
Could we add a test to catch this issue next time?

paramsingh (approved):
Merging this to fix the bug, let's add a test in a follow up.

PR #1091, opened 2020-09-16
documentation for recommendation scripts

vansika:
I have added a breif description at the top of every recommendation script to explain the purpose of the script. I have not been very strict about these docs since the scripts are still developing.


listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -1,3 +1,18 @@
+# This script is responsible to create candidate sets for all users. The generated candidate sets
```
paramsingh, line 1: We should make this a docstring, so that editors are able to pick it up.

paramsingh: Also, let's define what a  candidate set is, it's not obvious to a newbie (which admittedly I am)

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -1,3 +1,18 @@
...
+# will be given as input to the recommender to assign ratings to the recordings in candidate sets.
+# The general flow is as follows:
+#
+# Last 7 days listens are filtered from mapped_listens_df and is called the mapped_listens_subset_df.
```
paramsingh, line 5: The listens of the last 7 days are filtered from `mapped_listens_df` and this is called the `mapped_listens_subset_df`.

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -1,3 +1,18 @@
...
+# Top X artists are fetched for each user from the mapped_listens_subset_df. The top_artist_df is joined
```
paramsingh, line 6: what is X here?

listenbrainz_spark/recommendations/candidate_sets.py
```diff
@@ -1,3 +1,18 @@
...
+# with recordings_df to get the dataframe of recordings belonging to the top artists. From this dataframe,
+# recordings listened to by the users in the last 7 days are filtered so that the recommendations don't contain
+# recordings that the user has listened to in the last week. The resultant dataframe is called the top_artists_candidate_set_df.
+#
+# Artists similar to top artists are fetched from the artist_relations_df and the similar_artist_candidate_set_df is generated
+# in a manner similar to the generation of the top artist candidate set.
+#
+# The top artist and similar artist candidate set dataframes are saved to HDFS.
```
paramsingh, line 14: A path here would be helpful

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -1,3 +1,39 @@
+# This script is responsible to process user listening history and create and save dataframes to be used
```
paramsingh, line 1:
```suggestion
# This script is responsible for processing user listening history, creating, and saving the dataframes to be used
```

listenbrainz_spark/recommendations/create_dataframes.py
```diff
@@ -1,3 +1,39 @@
...
+# in training the model. The general flow is as follows:
+#
+# **get_dates_to_train_data: Get timestamp (from_date, to_date) to fetch user listening history.
```
paramsingh, line 4: This seems more like function documentation, I'd prefer if we modeled them as steps in simple english.

listenbrainz_spark/recommendations/create_dataframes.py
paramsingh, line 1: Maybe it's a good idea to follow the standard docstring format here and near the top, list the inputs that it needs and the outputs it generates. the situation, task, action, result framework is helpful when writing docs explaining large scripts. https://en.wikipedia.org/wiki/Situation,_task,_action,_result

listenbrainz_spark/recommendations/recommend.py
```diff
@@ -1,3 +1,15 @@
+# This script is responsible to generate recommendations for the users. The general flow is as follows:
```
paramsingh, line 1:
```suggestion
# This script is responsible for generating recommendations for  users. The general flow is as follows:
```

listenbrainz_spark/recommendations/train_models.py
```diff
@@ -1,3 +1,23 @@
+# This script is responsible to train models and save the best model to HDFS. The general
```
paramsingh, line 1: Again this just seems like a description of the functions, it would be better if the docstring explained in plain english what the script was doing instead of just explaining each function

paramsingh (changes requested):

paramsingh (approved):

vansika:
@brainzbot retest this please

PR #1092, opened 2020-09-16
In case of error 504 the API returns HTML instead of JSON

AbhinavOhri:
# Problem
In case of error 504 the API returns HTML instead of JSON

# Solution
Added error handlers for the respective errors


paramsingh (approved):
I'm happy to merge this, we should have these error handlers.

However, the HTML is actually returned on the gateway level, and flask code isn't even run, as such this problem can't be solved by changing code in LB, it needs gateway level changes which I guess only zas can do.

PR #1097, opened 2020-09-19
Add follow following component

shivam-kapila:
Please look into this. I feel some code might look hacky.

### DIFF
--- a/listenbrainz/webserver/static/js/src/APIService.ts
+++ b/listenbrainz/webserver/static/js/src/APIService.ts
@@ -119,24 +119,24 @@ export default class APIService {
     return { status: response.status };
   };
 
-  getFollowers = async (username: string) => {
+  getFollowersOfUser = async (username: string) => {
     if (!username) {
       throw new SyntaxError("Username missing");
     }
 
-    const url = `/user/${username}/get-followers`;
+    const url = `/user/${username}/followers`;
     const response = await fetch(url);
     this.checkStatus(response);
     const data = response.json();
     return data;
   };
 
-  getFollowing = async (username: string) => {
+  getFollowingForUser = async (username: string) => {
     if (!username) {
       throw new SyntaxError("Username missing");
     }
 
-    const url = `/user/${username}/get-following`;
+    const url = `/user/${username}/following`;
     const response = await fetch(url);
     this.checkStatus(response);
     const data = response.json();
--- a/listenbrainz/webserver/static/js/src/APIService.ts
+++ b/listenbrainz/webserver/static/js/src/APIService.ts
@@ -119,6 +119,30 @@ export default class APIService {
     return { status: response.status };
   };
 
+  getFollowers = async (username: string) => {
+    if (!username) {
+      throw new SyntaxError("Username missing");
+    }
+
+    const url = `/user/${username}/get-followers`;
+    const response = await fetch(url);
+    this.checkStatus(response);
+    const data = response.json();
+    return data;
+  };
+
+  getFollowing = async (username: string) => {
+    if (!username) {
+      throw new SyntaxError("Username missing");
+    }
+
+    const url = `/user/${username}/get-following`;
+    const response = await fetch(url);
+    this.checkStatus(response);
+    const data = response.json();
+    return data;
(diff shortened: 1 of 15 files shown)


paramsingh:
FYI, I've cherrypicked some backend part, specifically the db functions from here, modified a bit and added tests in #1100, thanks!

listenbrainz/webserver/static/js/src/APIService.ts
```diff
@@ -119,6 +119,30 @@ export default class APIService {
     return { status: response.status };
   };
 
+  getFollowers = async (username: string) => {
```
paramsingh, line 122: The name here can be more expressive, I went with `getFollowersOfUser` in the backend code.

listenbrainz/webserver/static/js/src/APIService.ts
```diff
@@ -119,6 +119,30 @@ export default class APIService {
...
+    }
+
+    const url = `/user/${username}/get-followers`;
+    const response = await fetch(url);
+    this.checkStatus(response);
+    const data = response.json();
+    return data;
+  };
+
+  getFollowing = async (username: string) => {
```
paramsingh, line 134: This name could also be more expressive, I used `getFollowingForUser`

listenbrainz/webserver/static/js/src/APIService.ts
```diff
@@ -119,6 +119,30 @@ export default class APIService {
     return { status: response.status };
   };
 
+  getFollowers = async (username: string) => {
+    if (!username) {
+      throw new SyntaxError("Username missing");
```
paramsingh, line 124: this should be something equivalent to valueerror, i guess. SyntaxError seems incorrect.

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -14,6 +14,7 @@ import FollowUsers from "./FollowUsers";
 import APIService from "./APIService";
 import Loader from "./components/Loader";
 import ListenCard from "./listens/ListenCard";
+import FollowerFollowingEntity from "./follow/followerFollowingEntity";
```
paramsingh, line 17: I added a basic UserFeed page in this commit (https://github.com/metabrainz/listenbrainz-server/pull/1109/commits/7d1f5f5e5637ddf021dea063f75bfe3b838759cf), you could cherry pick the commit and then show this component there instead of adding it to the recent listens page.

listenbrainz/webserver/static/js/src/follow/followerFollowingEntity.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    this.state = {
+      mode: "followers",
+      activeModeList: props.followerList,
+    };
+  }
+
+  updateMode = (mode: FollowerFollowingEntityMode) => {
+    const { followerList, followingList } = this.props;
+    this.setState({ mode });
```
paramsingh, line 35: This and the setState in the next line could be merged into one.

listenbrainz/webserver/static/js/src/follow/followerFollowingEntity.tsx
```diff
@@ -0,0 +1,93 @@
...
+          <Pill
+            active={mode === "following"}
+            type="primary"
+            onClick={() => this.updateMode("following")}
+          >
+            {followingList.length} Following
+          </Pill>
+        </div>
+        <div className="follower-following-list">
+          {activeModeList.map((listentry) => {
```
paramsingh, line 68: listentry is not a very expressive name, at the minimum, it should be camelcase listEntry, although i think just user would be ok in this context too.

paramsingh (changes requested):
This is great, thanks for working on it. I think for next steps, we should:

* Integrate this component into the feed page I added in https://github.com/metabrainz/listenbrainz-server/pull/1109/commits/7d1f5f5e5637ddf021dea063f75bfe3b838759cf
* Add tests


Other than that, this should be good to go, imo.

paramsingh (approved):
:ship:

PR #1099, opened 2020-09-20
import mapping into spark cluster before generating recommendations

vansika:
we'd want to import mapping into the spark cluster before generating recommendations

### DIFF
--- a/docker/stats-crontab
+++ b/docker/stats-crontab
@@ -58,5 +58,8 @@
 # user weekly daily_activity
 10 03 * * * /usr/local/bin/python /code/listenbrainz/manage.py spark request_user_stats --type=daily_activity --range=all_time
 
-# Request recommendations every Monday at 23:00
-00 23 * * 1 /code/listenbrainz/docker/cf_recommendation.sh cf
+# Import mapping into the spark cluster every Monday before generating recommendations.
+00 4 * * 1 /usr/local/bin/python /code/listenbrainz/manage.py spark request_import_mapping
+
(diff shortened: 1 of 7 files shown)


docker/stats-crontab
```diff
@@ -58,5 +58,8 @@
 # user weekly daily_activity
 10 03 * * * /usr/local/bin/python /code/listenbrainz/manage.py spark request_user_stats --type=daily_activity --range=all_time
 
-# Request recommendations every Monday at 23:00
-00 23 * * 1 /code/listenbrainz/docker/cf_recommendation.sh cf
+# Import mapping into the spark cluster every Monday before generating recommendations.
+00 4 * * 1 /usr/local/bin/python /code/listenbrainz/manage.py spark request_import_mapping
```
mayhem, line 62: How long does it take to import the mapping? A hour seems like cutting it close.

listenbrainz/spark/handlers.py
```diff
@@ -320,10 +320,13 @@ def notify_mapping_import(data):
         return
 
     mapping_name = data['imported_mapping']
-    import_completion_time = data['time']
+    import_time = data['import_time']
+    time_taken_to_import = data['time_taken_to_import']
```
mayhem, line 324: I guess we'll find out.

mayhem (approved):
:shipit:

docker/stats-crontab
vansika: A few mins. Kept that one hour gap just for safety :p

PR #1102, opened 2020-09-21
Fix anonymous user mixin error that occurs if user isn't logged in

shivam-kapila:
Fixes https://sentry.metabrainz.org/metabrainz/listenbrainz-beta/issues/99811/?environment=production&referrer=alert_email

### DIFF
--- a/listenbrainz/webserver/views/recommendations_cf_recording.py
+++ b/listenbrainz/webserver/views/recommendations_cf_recording.py
@@ -92,13 +92,17 @@ def _get_template(active_section, user):
             error_msg="An error occurred while processing your request. Check back later!"
         )
 
-    spotify_data = spotify.get_user_dict(current_user.id)
+    spotify_data = {}
+    current_user_data = {}
 
-    current_user_data = {
-            "id": current_user.id,
-            "name": current_user.musicbrainz_id,
-            "auth_token": current_user.auth_token,
-    }
+    if current_user:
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/views/recommendations_cf_recording.py
```diff
@@ -92,13 +92,17 @@ def _get_template(active_section, user):
...
-    spotify_data = spotify.get_user_dict(current_user.id)
+    spotify_data = {}
+    current_user_data = {}
 
-    current_user_data = {
-            "id": current_user.id,
-            "name": current_user.musicbrainz_id,
-            "auth_token": current_user.auth_token,
-    }
+    if current_user:
```
vansika, line 98: I think this will work `current_user.is_authenticated`

vansika (changes requested):
Rn, the page still error out if the user is not logged in , the suggested change should work I think.

listenbrainz/webserver/views/recommendations_cf_recording.py
vansika, line 98:
```suggestion
    if current_user.is_authenticated:
```

vansika (approved):

listenbrainz/webserver/views/recommendations_cf_recording.py
shivam-kapila: yep

PR #1108, opened 2020-09-24
pydantic model for recommendations

vansika:
One of the users got API Internal Server Error while trying to get the recommended tracks on listenbrainz.org. I checked the API endpoint and realised that recommendations for that user weren't inserted into the DB in the right format. Defining a pydantic model for recommendations will make sure that data received from spark which isn't in the format as expected is not be inserted into the DB.

### DIFF
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -317,8 +317,9 @@ def handle_recommendations(data):
             })
         )
     except ValidationError:
-        current_app.logger.error("""ValidationError while recommendations for user with musicbrainz_id: {musicbrainz_id}.
-                                 \nData: {data}""".format(musicbrainz_id=musicbrainz_id, data=json.dumps(data, indent=3)))
+        current_app.logger.error("""ValidationError while inserting recommendations for user with musicbrainz_id:
+                                 {musicbrainz_id}. \nData: {data}""".format(musicbrainz_id=musicbrainz_id,
+                                                                            data=json.dumps(data, indent=3)))
 
     current_app.logger.debug("recommendation for {} inserted".format(musicbrainz_id))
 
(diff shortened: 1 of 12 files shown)


listenbrainz_spark/recommendations/recommend.py
```diff
@@ -277,31 +281,78 @@ def create_messages(top_artist_rec_mbid_df, similar_artist_rec_mbid_df, active_u
...
+                current_app.logger.warning("""Invalid entry present in top artist recommendations for user: {user_name},
+                                           skipping""".format(user_name=row.user_name), exc_info=True)
+
             user_rec[row.user_name]['similar_artist'] = []
 
         else:
-            user_rec[row.user_name]['top_artist'].append([row.mb_recording_mbid, row.rating])
+            try:
+                user_rec[row.user_name]['top_artist'].append(
+                    UserRecommendationsRecord(**
```
paramsingh, line 300: why are we expecting an error here? going from dict -> pydantic object -> dict seems redundant. maybe it makes sense to just keep a dict here

listenbrainz/spark/handlers.py
```diff
@@ -304,11 +306,20 @@ def handle_recommendations(data):
...
-    top_artist_recording_mbids = data['top_artist']
-    similar_artist_recording_mbids = data['similar_artist']
+    recommendations = data['recommendations']
 
-    db_recommendations_cf_recording.insert_user_recommendation(user['id'], top_artist_recording_mbids,
-                                                               similar_artist_recording_mbids)
+    try:
+        db_recommendations_cf_recording.insert_user_recommendation(
+            user['id'],
+            UserRecommendationsJson(**{
```
paramsingh, line 314: :tada:

paramsingh (approved):
Changes make sense to me, I'm wondering if we really need the conversion on the spark side before sending the message. it seems a bit redundant, wdyt?

listenbrainz_spark/recommendations/recommend.py
vansika: Hmm. Can you merge and restart request consumer for now? I will update this later.  I guess you won't be available in the morning to restart request consumer, and I want to test stuff for a user whose recommend tracks page raise ISE.

paramsingh: Done! Can you open a ticket (or even better, a PR) addressing the comment?

PR #1112, opened 2020-09-28
update logging level from error to info of rating beyond the range messages

vansika:
We now have an idea about the range of ratings assigned by the recommender to the recordings, therefore the logs in sentry aren't required.


mayhem (approved):
🍰

vansika:
@brainzbot retest this please.

PR #1115, opened 2020-09-30
Move pre-mapping code to utils

ishaanshah:
# Problem
Not a problem per say. The pre-processing that has to be done before using mapping is same for stats as well as recs.

# Solution
We can move the pre-processing code to utils so that we can use it for stats mapping too.


vansika (approved):

paramsingh:
The utils.py file is getting huge. I think it'd be better if instead of adding stuff there we kept stuff in seperate files. For example, the functions here could be in mapping_conversions.py or something. That way, devs don't have to go through 1000 line files.

ishaanshah:
> The utils.py file is getting huge. I think it'd be better if instead of adding stuff there we kept stuff in seperate files. For example, the functions here could be in mapping_conversions.py or something. That way, devs don't have to go through 1000 line files.

I have created an `utils` directory and put current files in `__init__.py` so that we won't have to refactor all the files which use `utils`, further splitting will be done in future PRs, please have a look.

paramsingh (approved):

PR #1127, opened 2020-10-04
Informative user display messages for missing recommendations

vansika:


mayhem (approved):
This may not be complete yet, but it is fine to merge, so we're cleaning it up!

PR #1130, opened 2020-10-07
Add original_recording_mbid

mayhem:
If more than one recording in a lookup was redirected it was impossible to reconcile which one was redirected. Adding the original_recording_mbid column fixes this problem.

### DIFF
--- a/listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
+++ b/listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
@@ -95,7 +95,10 @@ class RecordingFromRecordingMBIDQuery(Query):
                     r['recording_mbid'] = str(r['recording_mbid'])
                     r['[artist_credit_mbids]'] = [str(r) for r in r['artist_credit_mbids']]
                     del r['artist_credit_mbids']
-                    r['original_recording_mbid'] = inverse_redirect_index[str(r['recording_mbid'])]
+                    try:
+                        r['original_recording_mbid'] = inverse_redirect_index[str(r['recording_mbid'])]
+                    except KeyError:
+                        r['original_recording_mbid'] = str(r['recording_mbid'])
                     output.append(r)
 
         return output
--- a/listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
+++ b/listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
@@ -1,3 +1,4 @@
+import copy
 import sys
 import uuid
 
@@ -47,7 +48,6 @@ class RecordingFromRecordingMBIDQuery(Query):
 
                 # Build an index with all redirected recordings
                 redirect_index = {}
-                inverse_redirect_index = {}
                 while True:
                     row = curs.fetchone()
                     if not row:
@@ -55,7 +55,6 @@ class RecordingFromRecordingMBIDQuery(Query):
 
                     r = dict(row)
                     redirect_index[str(r['recording_mbid_old'])] = str(r['recording_mbid_new'])
-                    inverse_redirect_index[str(r['recording_mbid_new'])] = str(r['recording_mbid_old'])
 
                 # Now start looking up actual recordings
                 for i, mbid in enumerate(mbids):
@@ -78,23 +77,19 @@ class RecordingFromRecordingMBIDQuery(Query):
                          ORDER BY r.gid'''
 
                 args = [tuple([psycopg2.extensions.adapt(p) for p in mbids])]
-                if count > 0:
-                    query += " LIMIT %s"
-                    args.append(count)
-                if offset >= 0:
-                    query += " OFFSET %s"
-                    args.append(offset)
-
                 curs.execute(query, tuple(args))
 
-                # Build an index of all the fetched recordings
+                # Build an index of all the fetched recordings and clean up UUID() -> str
                 recording_index = {}
                 while True:
                     row = curs.fetchone()
                     if not row:
                         break
 
-                    recording_index[str(r['recording_mbid'])] = dict(row)
+                    r = dict(row)
+                    mbid = str(r['recording_mbid'])
+                    r['recording_mbid'] = mbid
+                    recording_index[mbid] = dict(row)
 
                 # Finally collate all the results, ensuring that we have one entry with original_recording_mbid for each 
                 # input argument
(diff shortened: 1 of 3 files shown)


listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -54,45 +56,84 @@ def fetch(self, params, offset=-1, count=-1):
...
 
+                # Now start looking up actual recordings
                 for i, mbid in enumerate(mbids):
                     if mbid in redirect_index:
                         mbids[i] = redirect_index[mbid]
 
                 query = '''SELECT r.gid AS recording_mbid, r.name AS recording_name, r.length, r.comment,
                                   ac.id AS artist_credit_id, ac.name AS artist_credit_name,
-                                  array_agg(a.gid) AS artist_credit_mbids
+                                  array_agg(a.gid) AS artist_credit_mbids,
```
alastair, line 66: you said that you were going to unique these, did you fix the issue of the duplicates?

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -45,6 +46,7 @@ def fetch(self, params, offset=-1, count=-1):
                 args = [tuple([psycopg2.extensions.adapt(p) for p in mbids])]
                 curs.execute(query, tuple(args))
 
+                # Build an index with all redirected recordings
                 redirect_index = {}
                 while True:
                     row = curs.fetchone()
```
alastair, line 52: FYI: psycopg2 cursors are iterable, so you don't have to do while True/fetch row/check none. You can just do `for row in curs`

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -54,45 +56,84 @@ def fetch(self, params, offset=-1, count=-1):
                     r = dict(row)
                     redirect_index[str(r['recording_mbid_old'])] = str(r['recording_mbid_new'])
```
alastair, line 57: No strong opinions either way, but in AB when we deal with uuids we often `SELECT x::text` in the sql query to avoid having to call `str` everywhere in python

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -54,45 +56,84 @@ def fetch(self, params, offset=-1, count=-1):
...
 
-                output = []
+                # Build an index of all the fetched recordings and clean up UUID() -> str
+                recording_index = {}
                 while True:
                     row = curs.fetchone()
                     if not row:
                         break
 
                     r = dict(row)
```
alastair, line 100: I believe that if you use the DictCursor cursor_factory, you don't need to cast rows to a dict
Or is this because you need a real dict to be able to serialise the json?

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -54,45 +56,84 @@ def fetch(self, params, offset=-1, count=-1):
...
-                    r['recording_mbid'] = str(r['recording_mbid'])
-                    r['[artist_credit_mbids]'] = [str(r) for r in r['artist_credit_mbids']]
+                    mbid = str(r['recording_mbid'])
+                    r['recording_mbid'] = mbid
+                    recording_index[mbid] = dict(row)
```
alastair, line 103: above, you modify `r`, but then add `row` to the index

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -54,45 +56,84 @@ def fetch(self, params, offset=-1, count=-1):
...
+                for p in params:
+                    mbid = p['[recording_mbid]'] 
+                    try:
+                        r = copy.copy(recording_index[mbid])
+                    except KeyError:
+                        try:
+                            r = copy.copy(recording_index[redirect_index[mbid]])
+                        except KeyError:
+                            out = {} 
+                            for k in self.outputs():
```
alastair, line 117: I noticed that there's nothing that checks that the fields in `r` contain all of the keys in `self.outputs()`. Is this checked higher up in the hoster?

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -54,45 +56,84 @@ def fetch(self, params, offset=-1, count=-1):
...
+                                if k == 'original_recording_mbid':
```
alastair, line 118: Simpler way of doing this might be
```py
out = dict.fromkeys(self.outputs(), None)
out['original_recording_mbid'] = mbid
```

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -54,45 +56,84 @@ def fetch(self, params, offset=-1, count=-1):
...
                     del r['artist_credit_mbids']
+                    r['original_recording_mbid'] = mbid
                     output.append(r)
 
-        return output
+                if offset > 0 and count > 0:
+                    return output[offset:offset+count]
+
+                if offset > 0 and count < 0:
+                    return output[offset:]
```
alastair, line 134: An issue with doing the limit in python and not in postgres is that you have to perform the whole query anyway to get the data. However, because you have to do the redirect lookup, I'm not sure if there's a way around this, right?
Any thoughts on a default count value?

alastair (commented):
Look good, just some small suggestions and questions

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
mayhem: Yes, rather than mess with array_agg(), ugh, I opted for list(set(...)) below

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
mayhem: Ah, yes, thanks for the reminder -- I remember adopting this pattern eons ago when I encountered a bug in psycopg2 when I was a small child...

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
mayhem: The latter.

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
mayhem: Its not checked, but inherent in the query itself. I'm ok with it as is.

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
mayhem: Sadly with the case that you brought up yesterday there is no longer a 1:1 relation between the number of rows requested by the caller and the number of rows requested by this query, making limit/count potentially harmful.

We could probably continue to tweak this function to make it more efficient -- however, I feel that we're in position where we are trying to make things work, rather than making them fast. While I am certain that this Element will be central to Troi in nearly all cases, I don't feel like right now is the time to hone this into perfection. I'll add a note in the comment to this affect and then move on.

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
mayhem: I like that suggestion!

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
mayhem: As for default count value, I'm actually keen on letting the user set count to the number of rows required -- as to avoid paginating right now. Another thing to think about as we move closer to production.

listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
```diff
@@ -90,43 +90,40 @@ def fetch(self, params, offset=-1, count=-1):
...
-                                    out['original_recording_mbid'] = mbid
-                                else:
-                                    out['original_recording_mbid'] = None
+                            out = dict.fromkeys(self.outputs(), None)
+                            out['original_recording_mbid'] = mbid
                             output.append(out)
                             continue
 
-                    r['[artist_credit_mbids]'] = list(set([str(r) for r in r['artist_credit_mbids']]))
+                    r['[artist_credit_mbids]'] = list(set([r for r in r['artist_credit_mbids']]))
```
alastair, line 118: don't need the list comprehension now that we don't need `str`? `list(set(r['artist_credit_mbids']))`

alastair (approved):

PR #1131, opened 2020-10-07
offset should be added in count to fetch recommendations

vansika:
Before:
`count = count`

After:
`count = offset + count`


mayhem:
I haven't committed this fix for this bug yet -- can you please include in this PR?

mayhem:
Superceded by #1132

PR #1132, opened 2020-10-08
Fix recommendation fetch offset bug

mayhem:
This is the same PR as vansika's bug-offset, but with the actual fix included.


mayhem:
This is @vansika's code, approved by me, with the obvious fix added. Self approving and releasing.

vansika:
Okay , so i pushed the updated tests but not the actual updated code 🤷‍♀️

mayhem:
At least the fix was easy.

PR #1133, opened 2020-10-10
Separate component for Recommendations

vansika:
Separate component to render recommendations.


vansika:
@MonkeyDo @shivam-kapila I shared the screenshot on IRC, I don't understand why the recommendation card is not getting rendered, From the first look I understand that all the cards are overlapping each other, hence, we can see only one card, might be an issue of table reference, not sure.

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,87 @@
...
+        > * {
+          display: table-cell;
+          vertical-align: middle;
+        }
+
+        a {
+          word-break: break-word;
+        }
+
+        .recommendation-controls {
```
vansika, line 52: image

Is this because of recommendation-control class?

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,87 @@
...
+
+      &.current-recommendation {
+        background-color: rgba(217, 237, 247, 0.3) !important ;
+      }
+
+       div {
+         padding: 0px;
+       }
+
+      > .rec-col-xs-9, .rec-col-xs-12 {
```
shivam-kapila, line 19: This isnt gonna work. As we discussed on irc

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -0,0 +1,115 @@
...
+  faMeh,
+  faSmileBeam,
+  faLaughBeam,
+} from "@fortawesome/free-solid-svg-icons";
+
+import { getArtistLink, getTrackLink } from "../utils";
+import Card from "../components/Card";
+import RecommendationControl from "./RecommendationControl";
+
+export const DEFAULT_COVER_ART_URL = "/static/img/default_cover_art.png";
```
shivam-kapila, line 16: We can remove this for now

listenbrainz/webserver/static/js/src/recommendations/RecommendationControl.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const { className, action, icon, title } = props;
+  return icon ? (
+    <FontAwesomeIcon
+      icon={icon as IconProp}
+      className={className}
+      title={title}
+      onClick={action}
+    />
+  ) : (
+    <button className={className} title={title} onClick={action} type="button">
```
shivam-kapila, line 24: We can remove the button as we only use fa icons

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,324 @@
...
+            ) || [],
+          currRecPage: currRecPage + 1,
+        },
+        this.afterRecommendationsDisplay
+      );
+      window.history.pushState(null, "", `?page=${currRecPage}`);
+    }
+  };
+
+  recommendationPaginationControl = () => {
```
shivam-kapila, line 156: Why dont we have it main code in render method

shivam-kapila (changes requested):
This is a coarse review. I will review again once classes are over.

vansika:
I have reused @shivam-kapila 's code in this PR at places, Thank you. I have tried to keep everything separate for recommendations (from less file to recommendation controls) so that it is easy to update things in future.

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,86 @@
+
```
shivam-kapila, line 1: maybe we can have a common.less for common styling. Moreover there are quite a few you dont need.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+                  <small
+                    className="text-muted"
+                    title={recommendation.track_metadata.artist_name}
+                  >
+                    {getArtistLink(recommendation)}
+                  </small>
+                </p>
+              </div>
+            </div>
+          </MediaQuery>
```
shivam-kapila, line 85:
```suggestion
            <div className="col-xs-12">
              <div className="track-details">
                <p title={recommendation.track_metadata.track_name}>
                  {getTrackLink(recommendation)}
                </p>
                <p>
                  <small
                    className="text-muted"
                    title={recommendation.track_metadata.artist_name}
                  >
                    {getArtistLink(recommendation)}
                  </small>
                </p>
              </div>
            </div>
```
We dont need Media queries as the layout is same for both Mobile and Desktop view

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+              />
+              <RecommendationControl
+                icon={faMeh}
+                title="This is a bad recommendation!"
+              />
+              <RecommendationControl icon={faSmileBeam} title="I like this!" />
+              <RecommendationControl
+                icon={faLaughBeam}
+                title="I really love this!"
+              />
```
shivam-kapila, line 106: I think we can use the html emoticons that I shared, then we won't need much styling

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+export default class Recommendations extends React.Component<
+  RecommendationsProps,
+  RecommendationsState
+> {
+  private brainzPlayer = React.createRef<BrainzPlayer>();
+  private recommendationsTable = React.createRef<HTMLTableElement>();
+  private APIService: APIService;
+
+  private socket!: SocketIOClient.Socket;
```
shivam-kapila, line 40: Do we need sockets on recs page?

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+    if (idx >= 0) {
+      this.setState({
+        // remove the alert from the array
+        alerts: [...alerts.slice(0, idx), ...alerts.slice(idx + 1)],
+      });
+    }
+  };
+
+  handleClickPreviousRecommendations = () => {
```
shivam-kapila, line 112:
```suggestion
  handleClickPrevious = () => {
```

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,320 @@
...
+            ) || [],
+          currRecPage: currRecPage - 1,
+        },
+        this.afterRecommendationsDisplay
+      );
+      window.history.pushState(null, "", `?page=${currRecPage}`);
+    }
+  };
+
+  handleClickNextRecommendations = () => {
```
shivam-kapila, line 134:
```suggestion
  handleClickNext = () => {
```

shivam-kapila (changes requested):
We need tests

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
vansika: I don't exactly understand the purpose, so I just included them

listenbrainz/webserver/static/css/recommendation-page.less
vansika: I wanted to have a separate less file for recs. It will be easier to only change rec components and styling in the future. 

> Moreover there are quite a few you dont need.

Being more specific will help.

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
shivam-kapila: Sockets are used for real time updates etc. We wont need it here

listenbrainz/webserver/static/css/recommendation-page.less
shivam-kapila: I think if you nuke out the styling for the redundant media query component then it will serve the purpose. I can be specific if you are still stuck

shivam-kapila: ok I will see to it

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+                aria-expanded="true"
+              />
+              <ul
+                className="dropdown-menu dropdown-menu-right"
+                aria-labelledby="recommendationControlsDropdown"
+              >
+                <span
+                  className="angry"
+                  title="I never want to hear this again!"
+                >
```
shivam-kapila, line 80: I think you will need the Recommendation Control component as its gonna be functional and not just spans.

shivam-kapila (commented):
LGTM overall. Tests needed.

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,317 @@
+/* eslint-disable jsx-a11y/anchor-is-valid,camelcase */
+
+import { AlertList } from "react-bs-notifier";
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+import * as _ from "lodash";
```
MonkeyDo, line 6:
```suggestion
import {isEqual} from "lodash";
```

And further down in the same file remove the `_.` :
`_.isEqual(…`-> `isEqual(…`

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+                  className="text-muted"
+                  title={recommendation.track_metadata.artist_name}
+                >
+                  {getArtistLink(recommendation)}
+                </small>
+              </p>
+            </div>
+          </div>
+        </div>
+        <div className="col-xs-3 text-center">
```
MonkeyDo, line 62: It looks really wide to me. I think `col-xs-1` would be enough for this small button, no?
And that removes the very wide margin around it.
Ah, and that will require a change for the element before it `col-xs-9` -> `col-xs-11`


Capture d’écran 2020-10-13 à 21 41 33

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,87 @@
...
+          .smile, .laugh, .angry, .disappoint, .meh {
+              font-size: 18px;
+              margin: 0% 7%;
+          }
+
+          .dropdown-menu {
+            min-width: 400px;
+            vertical-align: middle;
+            height: 40px;
```
MonkeyDo, line 65: I think it looks better without this. There's an uneven space above/below the icons, but it looks nicely centered without the `40px` height.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -0,0 +1,102 @@
...
+              />
+              <ul
+                className="dropdown-menu dropdown-menu-right"
+                aria-labelledby="recommendationControlsDropdown"
+              >
+                <span
+                  className="angry"
+                  title="I never want to hear this again!"
+                >
+                  &#128544;
```
MonkeyDo, line 81: I know it's a question of taste, and as such not anything stopping, but I personally don't like the HTML icons and  do prefer the fontawesome ones.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -46,6 +46,17 @@ declare type Listen = {
   score?: number;
 };
 
+declare type Recommendation = {
```
MonkeyDo, line 49: Unless I'm missing something, this should be equivalent to the Listens type declared above, since the fields you removed (listened_at_iso, playing_now) are marked as optional (the '?' after the variable name).
You should be able to use the Listen type directly without breaking anything.

MonkeyDo (changes requested):
Not sure if I was meant to review the styles of the recommendation card component, I had to modify similar_artist.html to point to recommendations.js instead of main.js in order to see them.

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,87 @@
...
+
+        .recommendation-controls {
+
+          > svg:hover {
+            cursor: pointer;
+          }
+
+          .smile, .laugh, .angry, .disappoint, .meh {
+              font-size: 18px;
+              margin: 0% 7%;
```
MonkeyDo, line 59: This looks really wide on desktop with percentages
```suggestion
              margin: 0 0.4em;
```

listenbrainz/webserver/static/css/recommendation-page.less
MonkeyDo, line 65: This looks really wide to me. I'd suggest this to size to the contents:
```suggestion
            display: flex;
```

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,87 @@
+
+
+#recommendations {
```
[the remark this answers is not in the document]
shivam-kapila: @MonkeyDo exactly my concerns

listenbrainz/webserver/static/js/src/types.d.ts
shivam-kapila: @MonkeyDo I dont think rec will have listened_at. So we went for separate component. @vansika can you remove that

shivam-kapila:
@vansika kindly run lint.sh to fix the build

listenbrainz/webserver/static/css/recommendation-page.less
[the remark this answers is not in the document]
vansika: @MonkeyDo Nice, I understand this now.  One concern I have is #listens has classNames like `listens-control` etc wherein `listens`  prefix/suffix is used. will it not be semantically wrong to use classNames that have `listens` in them for `recommendations` since `listens` and `recommendations` are two separate types. This was my thought process behind duplicating code.

listenbrainz/webserver/static/css/recommendation-page.less
vansika: image

Renders like this if I change the margin.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
vansika: I guess I will keep the spans now, @shivam-kapila can do the fontawsome thing when he will write the css part.

listenbrainz/webserver/static/js/src/types.d.ts
vansika: @MonkeyDo I am trying to keep listens and recommendations as much separate as I can. But BP accepts `Listens` type. Like @shivam-kapila said, recs won't have a listened_at field either but `Listens` type makes it mandatory to have listened at, here https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/types.d.ts#L36

So when I remove `listened_at` from recs I get errors saying `Listened_at missing`, that is why I have kept listenes_at in recs type until I figure out how to solve it. 

Ideally the recs type should be:
`declare type Recommendation = {
  user_name?: string | null;
  track_metadata: {
    artist_name: string;
    release_name?: string | null;
    track_name: string;
    additional_info?: AdditionalInfo;
  };
};`

MonkeyDo: To me that sounds more like an adjustment needed on the Listens type, until there is a clear need to make another type, to avoid code duplication (especially considering the recommendations are based on the listens format).
Namely to make `listened_at` optional: `listened_at?: number` (note the question mark which makes it equal to `number | undefined`).

Inevitably there will be a couple of changes to make to other components, where typescript will complain that `listened_at is possibly 'undefined'.`.
These we should be able to solve relatively easily, for example transforming `new Date(listen.listened_at * 1000).toISOString()` into `(listen.listened_at ? new Date(listen.listened_at * 1000).toISOString() : null)` to ensure listened_at is defined.




There is also another way: [extending](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html) the Listen type, or creating a BaseListen type and extending it for Listen and Recommendation types:
```typescript
declare type Listen = {
  listened_at?: number;
  listened_at_iso?: string | null;
  playing_now?: boolean | null;
  user_name?: string | null;
  track_metadata: {
    artist_name: string;
    release_name?: string | null;
    track_name: string;
    additional_info?: AdditionalInfo;
  };
};

declare type Recommendation = Listen & {
  score?: number;
};
```

Or:
```typescript
declare type BaseListenFormat = {
  user_name?: string | null;
  track_metadata: {
    artist_name: string;
    release_name?: string | null;
    track_name: string;
    additional_info?: AdditionalInfo;
  };
};

declare type Listen = BaseListenFormat & {
  listened_at?: number;
  listened_at_iso?: string | null;
  playing_now?: boolean | null;
};

declare type Recommendation = BaseListenFormat & {
  score?: number;
};
```

vansika: I like the idea of making listened_at optional (there will be tests for `my_listens` etc to ensure that listened_at *must* be passed with Listens so it will work :)). 

is there a way to rename listens -> recommendations when they will be used in Recommendations.tsx? I am trying to avoid Listens name in Recommendation component.

vansika: `declare type Listen = {
  listened_at?: number;
  listened_at_iso?: string | null;
  playing_now?: boolean | null;
  user_name?: string | null;
  track_metadata: {
    artist_name: string;
    release_name?: string | null;
    track_name: string;
    additional_info?: AdditionalInfo;
  };
};

declare type Recommendation = Listen

vansika: How about this? (Score isn't required)

listenbrainz/webserver/static/css/recommendation-page.less
MonkeyDo: Yeah, that's due to  the `min-width: 400px` style you added to the dropdown-menu.
Without it the space is too small for all emoticons to fit on one line, but we can fix that and have the containing element adjust to the size of the icons, and vice-verse:

```
.recommendation-controls.open .dropdown-menu {
    display: flex;
    // width: 180px; // if you want, you can define a width here, and the elements inside will spread out to fill the space
}
.smile, .laugh, .angry, .disappoint, .meh {
    font-size: 18px;
    flex: 1; // Fills out the space evenly
    text-align: center;
}
```

listenbrainz/webserver/static/js/src/types.d.ts
MonkeyDo: That could work.
I think in that case you could just rename the Listen type instead to something that will work with both cases :)

But I think my second suggestion (extend BaseListenFormat into Listen and Recommendation types) would solve that issue and allow for things to change a bit more easily in the future.

listenbrainz/webserver/static/css/recommendation-page.less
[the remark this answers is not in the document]
MonkeyDo: Good question.
I think we don't have to worry too much about that, considering one of two things will happen: 
1) we can rename the styles that will be used for both listens and recs (listen-controls -> controls, listen-card -> card or track-card, or something…)
2) some css will be generated but never used, which isn't too bad:
For example we'll get in the compiled CSS a rule like `#listens > .listen-card.playing-now, #recommendations > .listen-card.playing-now {…}` but we won't have any .playing-now element in the recs, and the css rule will just be ignored while the .listen-card.playing-now element in the #listens element will be styled as you expect.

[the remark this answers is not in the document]
MonkeyDo: In short, what I suggest for now is to extend #listens, and we can clean up any unused css or refactor accordingly once things are a bit less in flux.

listenbrainz/webserver/static/css/recommendation-page.less
vansika: `.recommendation-controls {

        > svg:hover {
          cursor: pointer;
        }

        .recommendation-controls.open .dropdown-menu {
          display: flex;
          width: 1000px;
          
          button {
            width: 100%;
            background: none;
            color: inherit;
            border: none;
            padding: 5px 20px;
            font: inherit;
            cursor: pointer;
            outline: inherit;
            text-align: left;

            &:hover {
              color: #ffffff;
              background-color: #eb743b;
            }
          }
        }

        .smile, .laugh, .angry, .disappoint, .meh {
          font-size: 18px;
          flex: 1;
          text-align: center;
          margin: 0 0.4em;
        }
      }`

vansika: image

vansika: Was just experimenting doing `width: 1000px`, it seems the flex isn't expanding.

MonkeyDo: I see two .recommendation-controls, which means your css will look like:
`.recommendation-controls .recommendation-controls.open .dropdown-menu` and that won't affect your element.

Here's my refactor:
```less
.recommendation-controls {

  &.open .dropdown-menu {
    display: flex;
    width: 1000px;
  }

  .smile, .laugh, .angry, .disappoint, .meh {
    font-size: 18px;
    flex: 1;
    text-align: center;
    margin: 0 0.4em;
  }
}
```

The `&.open` means `{parent selector}.open` which in this case will compile as `.recommendation-controls.open`

listenbrainz/webserver/static/js/src/types.d.ts
shivam-kapila: I dont know what you are trying to achieve by changing RecentListens but I used and all worked fine 

```
  user_name?: string | null;
  track_metadata: {
    artist_name: string;
    release_name?: string | null;
    track_name: string;
    additional_info?: AdditionalInfo;
  };
};

declare type Listen = BaseListenFormat & {
  listened_at: number;
  listened_at_iso?: string | null;
  playing_now?: boolean | null;
};

declare type Recommendation = BaseListenFormat;
```

vansika:
@MonkeyDo I think I am happy with this PR now, I am not touching listened_at for Listens type (people freaking out), that means I will have to supply listened_at with which is okay for me. If you feel the changes are okay, I will want to write tests and merge it.

vansika:
image

@MonkeyDo here is a screenshot for your reference :)

paramsingh (changes requested):
I think we should change BrainzPlayer to not require a listened_at prop instead of changing the recommendation type to keep a listened at there.

vansika:
@paramsingh I will not be doing that in this PR.

vansika:
@paramsingh why not open a ticket :) will help me track

paramsingh:
OK, sounds like a plan.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -32,20 +32,25 @@ interface AdditionalInfo {
...
     release_name?: string | null;
     track_name: string;
     additional_info?: AdditionalInfo;
   };
+};
+
+declare type Listen = BaseListenFormat & {
+  listened_at_iso?: string | null;
+  playing_now?: boolean | null;
   score?: number;
```
MonkeyDo, line 49: I wonder if this is relevant anywhere.
Doesn't need to be addressed in this PR, but maybe have a closer look at it when we tackle listened_at/BP changes.

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -32,20 +32,25 @@ interface AdditionalInfo {
...
 };
 
+declare type Recommendation = BaseListenFormat;
```
MonkeyDo, line 52: That'll work nicely, and means you can easily extend the type with more fields as you need in the future, without having to change anything else( like we're gonna have to do for BP).
Yay future-proofing !

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,61 @@
...
+  &:extend(#listens);
+  padding: 10px;
```
MonkeyDo, line 3: I think this less file will need a good refactor soon (and the listens-page file along with it), as there is still some duplicated css rules and a few things I'd like to change.
However, it ends up as just a few lines of extra unused CSS, which is not a blocker for this PR.
Considering how the rec styles are in flux, let's keep it for another PR.

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,61 @@
...
+    margin-bottom: 7px;
+    max-height: 72px;
+
+    &.current-recommendation {
+      background-color: rgba(217, 237, 247, 0.3) !important ;
+    }
+
+    > .col-xs-9, .col-xs-11 {
+
+      > .col-xs-9, .col-xs-11 {
```
MonkeyDo, line 16: I think you based this styling on what was there before, but a worthy comment:

Among the things that need changing that I mentioned above (for another PR), this use of bootstrap classes that are meant for layout to be used for styling should be replaced.
The reason is that for example if we decide to change the size of the columns but not the styling, we'll have to modify the less rules even though we haven't changed any of the styling.
Instead, we'll want them to be decoupled by using separate classes for styling and layout (like you do below with `.smile, .laugh, .angry, .disappoint, .meh` instead of `> span`).

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,61 @@
+#recommendations {
+  &:extend(#listens);
+  padding: 10px;
+
+  > .recommendation-card {
+    padding: 15px;
+    margin-bottom: 7px;
+    max-height: 72px;
```
MonkeyDo, line 8: We'll have to review this before deploying. Spacing looks off to me:
Capture d’écran 2020-10-15 à 11 37 44

MonkeyDo (approved):
We'll need to go over the CSS, but it's not blocking for merging this PR.

PR #1134, opened 2020-10-11
[WIP]LB-547: Use mapping for calculating artist statistics

ishaanshah:
# Problem
Some artists get entries get repeated because of slight difference in names, different MSIDs because of different metadata. All these entries should be combined into one.

# Solution
Using MSID-MBID mapping to get clean data from MusicBrainz should fix this issue. It also provides benefits such as getting MBIDs for artists and correcting their names.

# Action
I have tested the code using subset of mapping on my local machine but we should test it with the full mapping on production server to make sure that we don't run into `OOM` issues and the results are actually better than before.

# Note
Should be merged after #1115


ishaanshah:
Closing this as we are moving away from ACRP matching in spark cluster to mapping listens using Typesense.

PR #1135, opened 2020-10-13
Add prop based styling for follow button

shivam-kapila:
Since we needed different styling for two types of follow button, this PR adds the feature:

**DEMO**
Screencast from 25-11-20 03 36 00 PM IST


MonkeyDo (changes requested):
The approach works well.

As suggested in the chat (reposting here for posterity), I wonder if we can try to unify the looks of these two buttons.
Here's the sort of thing I have in mind, keeping both the blue rounded button style and the full color background expandable icon:
follow_button_test


A separate issue: this is what results if there's a network issue trying to follow a user in the feed:
Capture d’écran 2020-10-15 à 12 55 56
We either want a button that can be a lot wider (and maybe with an option to be red when there's an error?), or we don't want that error message in the button.
I'm guessing it would be ideal to use the same Alert cards component we use in the RecentListens component for example.

MonkeyDo:
Here's the resulting less for that example above. I reconstituted it from my experimenting in chrome console, so don't consider it a replacement for existing less file.

```less
#follow-button {
    margin-left: 10px;
    font-family: "Sintony";
    color: white;
    background: #353070;
    border: 2px solid #353070;
    font-weight: bold;
    border-radius: 4px;
    padding: 5px 8px;
    text-align: center;
    transition: color 0.42s, background 0.42s;
    min-width: 3em;
	
	&:hover, &:focus {
		background: white;
		color: #353070;
	}

	&.icon-only {
		 > .text {
			margin-left: 0px;
			width: 0;
			display: inherit;
			vertical-align: inherit;
			transition: width 0.42s, margin-left 0.42s;
			overflow-x: hidden;
		}
		&:hover > .text, &:focus > .text {
			width: 4.5em;
			margin-left: 5px;
		}
	}
}
```

listenbrainz/webserver/static/css/follow.less
```diff
@@ -0,0 +1,43 @@
+@purple: #353070;
```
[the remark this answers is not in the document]
shivam-kapila: we don't have lobes theme in LB as of now

shivam-kapila:
@MonkeyDo unification done. Please take a look

listenbrainz/webserver/static/css/follow.less
```diff
@@ -0,0 +1,50 @@
...
+      transition: width 0.42s, margin-left 0.42s;
+      overflow-x: hidden;
+    }
+
+    &:hover, &:focus {
+      background: white;
+      color: #353070;
+
+      > .text {
+        min-width: 4.75em;
```
MonkeyDo, line 32: This should probably be `width` instead, since it is the property used in the `transition` (on line 23).
Otherwise you get a sort of jumpy animation.

listenbrainz/webserver/static/css/follow.less
```diff
@@ -0,0 +1,50 @@
...
+        margin-left: 5px;
+      }
+    }
+  }
+
+  &.block {
+    width: 85px;
+
+    &.following {
```
MonkeyDo, line 41: I think we should also add `&:hover` here, since otherwise the button does not show any indication of hovering (apart from the cursor change)

MonkeyDo (changes requested):
Last couple of details, it's looking quite nice !

I think having similar style buttons on different pages is definitely looking a lot more coherent.

listenbrainz/webserver/static/css/follow.less
shivam-kapila: It works to change following to unfollwo. Do we need to change styling for follow button too?

listenbrainz/webserver/static/css/follow.less
shivam-kapila: done

listenbrainz/webserver/static/css/follow.less
MonkeyDo: I don't understand your comment. I don't see any `unfollow` class anywhere.

What I'm suggesting is a simple addition to match the `icon-only` style's
```less
&:hover, &:focus {
      background: white;
      color: #353070;
```
rule (line 27), that are missing on the `block` style.
I'm just suggesting to add `&:hover, &:focus` after `&.following` on this line (41) to have the same `background: white; color: #353070;` style for `block`.

listenbrainz/webserver/static/css/follow.less
```diff
@@ -0,0 +1,50 @@
...
+      margin-left: 0px;
+      display: inherit;
+      vertical-align: inherit;
+      transition: width 0.42s, margin-left 0.42s;
+      overflow-x: hidden;
+    }
+
+    &:hover, &:focus {
+      background: white;
+      color: #353070;
```
MonkeyDo, line 29: This line and line 43 should use `@purple`.

shivam-kapila: Thanks. done

listenbrainz/webserver/static/css/follow.less
shivam-kapila: sorry for the confusion. done

shivam-kapila:
@MonkeyDo 
Screencast from 25-11-20 03 36 00 PM IST

MonkeyDo (approved):
LGTM ! 🚀

PR #1139, opened 2020-10-15
Fetch up to 1000 recommendations in one request

vansika:
For the purpose of experimenting with recommendations to generate daily jams, enable fetching up to 1000 recs using the endpoint. This is a temporary feature, may change in future.


listenbrainz/webserver/views/recommendations_cf_recording_api.py
```diff
@@ -144,7 +144,9 @@ def _process_recommendations(recommendations, count, artist_type, user_name, off
         err_msg = 'No recommendations for user {}, please try again later.'.format(user_name)
         raise APINoContent(err_msg, payload={'last_updated': int(getattr(recommendations, 'created').timestamp())})
 
-    count = min(count, MAX_ITEMS_PER_GET)
+    # For the purpose of experimenting with recommendations, we're allowing to fetch at most
+    # 1K recommendations.
+    count = min(count, 1000)
```
mayhem, line 149: This should really be a defined constant, tied the number of recommendations we generate. For expediency sake, maybe open a ticket for improving that so we can merge this faster?

mayhem (approved):
Approved, with one caveat. Let's open a ticket or push a fix -- your call. Thanks!

listenbrainz/webserver/views/recommendations_cf_recording_api.py
vansika: Cool. Will do!

vansika:
@brainzbot retest this please

listenbrainz/webserver/views/recommendations_cf_recording_api.py
vansika: Okay, so i was thinking about it. I would prefer to not have a single constant for passing to cron job and FETCH_MAX. that ways offset will have lil meaning. Also, fetching 1k at once is quite a good number, I am not sure if we will want to fetch 2k at once, even if we want to offset will help. 
Though I am not very rigid here, just wanted to discuss.

mayhem: Two separate ones are fine by me, as long as we use constants.

vansika: Sounds good. If you need #1138 and #1139 merged ASAP, feel free to. I have noted this, I will open a PR by tomorrow :)

PR #1142, opened 2020-10-15
Add test interface to BU metrics storage

alastair:
# Problem

This is a test branch to integrate numerical metrics storage in redis and make it available to the metabrainz granfa stats system.

# Solution

Add two internal api methods, `/internal/metrics/increment/<metricname>` and `/internal/metrics/stats/<metricname>` to store and retrieve stats.

I wasn't sure if we should add the stats endpoint to the existing API url structure, as it's not something that needs to be used by end-users. Should we stay with a separate prefix like `/internal/`?

# Action

Deploy this branch on test or beta, and check that we can read the stats from telegraf and show on granfa. If it works as expected, release a new version of BU and use it to add stats to LB

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -236,6 +236,7 @@ def _register_blueprints(app):
     from listenbrainz.webserver.views.login import login_bp
     from listenbrainz.webserver.views.api import api_bp
     from listenbrainz.webserver.views.api_compat import api_bp as api_bp_compat
+    from listenbrainz.webserver.views.api_metrics import api_metrics_bp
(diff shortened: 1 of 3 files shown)


paramsingh:
Really excited for this, do you need any help with deploying this?

alastair:
We deployed this as a test with zas: https://stats.metabrainz.org/d/Ftiq-ptMz/listenbrainz-stats?orgId=1&refresh=1m
The feedback was that my proposal in BU was far too complex, and we should rely on functionality in influx to store this data instead of grouping it ourselves. I'll redo the bu pr based on this discussion

alastair:
This was only a test and doesn't need to be merged, closing.

PR #1143, opened 2020-10-16
Recommendation feedback [DB + API]

vansika:
Database script to store recommendation feedback + Feedback API script to submit, delete, get recommendation feedback!

### DIFF
--- a/listenbrainz/db/model/recommendation_feedback.py
+++ b/listenbrainz/db/model/recommendation_feedback.py
@@ -7,7 +7,7 @@ from pydantic import BaseModel, ValidationError, validator
 def get_allowed_ratings():
     """ Get rating values that can be submitted corresponding to a recommendation.
     """
-    return [ 'like', 'love', 'dislike', 'hate', 'bad_recommendation']
+    return ['like', 'love', 'dislike', 'hate', 'bad_recommendation']
 
 
 def check_recording_mbid_is_valid_uuid(rec_mbid):
--- a/listenbrainz/db/model/recommendation_feedback.py
+++ b/listenbrainz/db/model/recommendation_feedback.py
@@ -4,35 +4,52 @@ from datetime import datetime
 from pydantic import BaseModel, ValidationError, validator
 
 
-class RecommendationFeedback(BaseModel):
-    """ Represents a recommendation feedback object
+def get_allowed_ratings():
+    """ Get rating values that can be submitted corresponding to a recommendation.
+    """
+    return [ 'like', 'love', 'dislike', 'hate', 'bad_recommendation']
+
+
+def check_recording_mbid_is_valid_uuid(rec_mbid):
+    try:
+        rec_mbid = uuid.UUID(rec_mbid)
+        return str(rec_mbid)
+    except (AttributeError, ValueError):
+        raise ValueError('Recording MBID must be a valid UUID.')
+
+
+class RecommendationFeedbackSubmit(BaseModel):
+    """ Represents a recommendation feedback submit object.
         Args:
             user_id: the row id of the user in the DB
-            user_name: (Optional) the MusicBrainz ID of the user
             recording_mbid: the MusicBrainz ID of the recording
-            feedback: the feedback associated with the recording
+            rating: the feedback associated with the recommendation
             created: (Optional)the timestamp when the feedback record was inserted into DB
     """
 
     user_id: int
-    user_name: str = None
     recording_mbid: str
-    feedback: str
+    rating: str
     created: datetime = None
 
-    @validator('feedback')
-    def check_feedback_is_valid(cls, feedback):
-        expected_feedback = [
-            'like', 'love', 'dislike', 'hate', 'bad_recommendation'
-        ]
-        if feedback not in expected_feedback:
-            raise ValueError('Feedback can only have a value in {}'.format(expected_feedback))
-        return feedback
+    @validator('rating')
+    def check_feedback_is_valid(cls, rating):
+        expected_rating = get_allowed_ratings()
+        if rating not in expected_rating:
+            raise ValueError('Feedback can only have a value in {}'.format(expected_rating))
+        return rating
 
-    @validator('recording_mbid')
-    def check_recording_mbid_is_valid_uuid(cls, rec_mbid):
-        try:
-            rec_mbid = uuid.UUID(rec_mbid)
-            return str(rec_mbid)
-        except (AttributeError, ValueError):
-            raise ValueError('Recording MBID must be a valid UUID.')
+    _is_recording_mbid_valid: classmethod = validator("recording_mbid", allow_reuse=True)(check_recording_mbid_is_valid_uuid)
+
+
+class RecommendationFeedbackDelete(BaseModel):
+    """ Represents a recommendation feedback delete object.
+        Args:
+            user_id: the row id of the user in the DB
+            recording_mbid: the MusicBrainz ID of the recommendation
+    """
+
+    user_id: int
+    recording_mbid: str
+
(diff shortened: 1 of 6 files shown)


listenbrainz/db/model/recommendation_feedback.py
```diff
@@ -0,0 +1,55 @@
...
+    except (AttributeError, ValueError):
+        raise ValueError('Recording MBID must be a valid UUID.')
+
+
+class RecommendationFeedbackSubmit(BaseModel):
+    """ Represents a recommendation feedback submit object.
+        Args:
+            user_id: the row id of the user in the DB
+            recording_mbid: the MusicBrainz ID of the recording
+            rating: the feedback associated with the recommendation
```
mayhem, line 26: This doesn't tell me the type or expected values that this argument might take. We should mention that needs to be an enum value.

listenbrainz/db/recommendations_cf_recording_feedback.py
```diff
@@ -0,0 +1,106 @@
...
+               AND recording_mbid = :recording_mbid
+            """), {
+                'user_id': feedback_delete.user_id,
+                'recording_mbid': feedback_delete.recording_mbid,
+            }
+        )
+
+
+def get_feedback_for_user(user_id: int, limit: int, offset: int, rating: str = None) -> List[RecommendationFeedbackSubmit]:
+    """ Get a list of recommendation feedback given by the user in descending order of their creation
```
mayhem, line 53: I think you should mention that feedback can be filtered here.

listenbrainz/webserver/views/recommendations_cf_recording_feedback_api.py
```diff
@@ -0,0 +1,185 @@
...
+            ],
+            offset: 0,
+            total_count: 1,
+            user_name: 'Vansika'
+        }
+
+    If the optional argument ``rating`` is not given, this endpoint will return all the feedback submitted by the user.
+    Otherwise filters the feedback to be returned by rating.
+
+    :param rating: Optional, refer to db/model/recommendation_feedback.py for allowed rating values.
```
mayhem, line 142: This comment is great and you should copy this line to where I commented above.

listenbrainz/webserver/views/recommendations_cf_recording_feedback_api.py
```diff
@@ -0,0 +1,185 @@
...
+    return jsonify({
+        "recommendation-feedback": feedback,
+        "count": len(feedback),
+        "total_count": total_count,
+        "offset": offset,
+        "user_name": user["musicbrainz_id"]
+    })
+
+
+def _feedback_to_api(fb):
```
mayhem, line 181: This function is mysterious. It could use a docstring and an improved name.

mayhem (commented):
This looks very good -- only minor comments. Now ready for tests!

listenbrainz/db/model/recommendation_feedback.py
```diff
@@ -0,0 +1,55 @@
...
+    created: datetime = None
+
+    @validator('rating')
+    def check_feedback_is_valid(cls, rating):
+        expected_rating = get_allowed_ratings()
+        if rating not in expected_rating:
+            raise ValueError('Feedback can only have a value in {}'.format(expected_rating))
+        return rating
+
+    _is_recording_mbid_valid: classmethod = validator("recording_mbid", allow_reuse=True)(check_recording_mbid_is_valid_uuid)
```
shivam-kapila, line 42: I think we can use validator as it is like [this](https://github.com/metabrainz/listenbrainz-server/blob/06231b355ad3ce18885b82873c5dd784e7aa70e8/listenbrainz/db/model/feedback.py#L29). If it doesnt work let me know too :p.

listenbrainz/db/model/recommendation_feedback.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class RecommendationFeedbackDelete(BaseModel):
+    """ Represents a recommendation feedback delete object.
+        Args:
+            user_id: the row id of the user in the DB
+            recording_mbid: the MusicBrainz ID of the recommendation
+    """
+
+    user_id: int
```
shivam-kapila, line 52: Can these two models be comined with rating and created as optional?

listenbrainz/db/model/recommendation_feedback.py
```diff
@@ -0,0 +1,55 @@
...
+        Args:
+            user_id: the row id of the user in the DB
+            recording_mbid: the MusicBrainz ID of the recording
+            rating: the feedback associated with the recommendation
+            created: (Optional)the timestamp when the feedback record was inserted into DB
+    """
+
+    user_id: int
+    recording_mbid: str
+    rating: str
```
shivam-kapila, line 32: This isn't actually a rating. Wondering if we call it feedback only. But no strong opinions.

shivam-kapila (commented):
The PR looks awesome. Was just wondering if the Pydantic model can be simplified.

listenbrainz/db/model/recommendation_feedback.py
vansika: I wanted to use the same validator function for two classes, this the way for it ig

listenbrainz/db/model/recommendation_feedback.py
vansika: Yeah, I thought about it, but given the the very nature of submit feedback (without rating you cannot submit feedback) , I wanted to keep rating as mandatory so that it looks good symantically. That is why made two classes.

listenbrainz/db/model/recommendation_feedback.py
vansika: And if you want, you can use the same validation func in feedback. py. It is exactly same func in two places.

listenbrainz/db/model/recommendation_feedback.py
shivam-kapila: Lets ask @alastair I think he will have a good point for the clash.

listenbrainz/db/model/recommendation_feedback.py
shivam-kapila: Lets ask @alastair I think he will have a good point for the clash.

vansika:
@mayhem pushed the requested changes and tests. Have a look, I want to merge it today, possibly :)

mayhem:
I doubt I'll have time until Tuesday, sadly.

listenbrainz/db/tests/test_recommendations_cf_recording_feedback.py
```diff
@@ -112,54 +111,99 @@ def test_get_feedback_for_user(self):
...
                 RecommendationFeedbackSubmit(
                     user_id=self.user2['id'],
-                    recording_mbid=str(uuid.uuid4()),
+                    recording_mbid=rec_mbid,
+                    rating='love'
+                )
+            )
+            # prepended to the list since ``get_feedback_for_users`` returns data in descending
+            # order of creation.
+            feedback_love.insert(
```
mayhem, line 126: Is there a reason why you save the mbid and create the object twice? Why not create the object once and save it for adding to the feedback_love[] ?

mayhem (approved):
Looks good to me, even though I find it weird to create the objects twice..

listenbrainz/db/tests/test_recommendations_cf_recording_feedback.py
vansika: No, I don't know why I did that 🙄. 
Pushing the changes.

mayhem (approved):
:shipit:

PR #1144, opened 2020-10-16
Disable cronjob for incremental dump import

ishaanshah:
# Problem
There is a bug in code for incremental dump import which causes it to import dumps infinitely.

# Solution
We need to pause cronjobs for incremental dump import temporarily to make sure that data in cluster is in a valid state.


mayhem (approved):
Thanks!

PR #1146, opened 2020-10-19
Refactor RecentListens - Remove recommendations code

vansika:
I have refactored recentListens to remove all the code which is associated with recommendations since we now have separate components for recommendations.


MonkeyDo (changes requested):
I still see one item: `expectedRecommendationsPerPage`.
Apart from that, everything looks fine to me!

MonkeyDo (approved):
🚀

PR #1147, opened 2020-10-19
Fix bug in code for tracking imported data into spark cluster

ishaanshah:
# Problem
There was a bug in the code that inserted information about latest imported dump which lead to multiple imports of the same dump.

# Solution
The above mentioned bug has been fixed and the tests have been modified to ensure that it doesn't occur in the future again.

# Action
We should test this on the cluster once and then re-enable the cron-job for importing data daily.


ishaanshah:
@brainzbot retest this please

paramsingh (approved):
Discussed over IRC, looks good but need a few tickets to be opened.

PR #1148, opened 2020-10-20
Fix incremental time range fetching mechanism

MonkeyDo:
# Problem
Vansika reported an issue (in fact there are multiple) when navigating from her listens first page to the second, then back to the first.
1. 'Newer' button is not disabled, although we are on the first page
2. The "No more listens to show…" message is shown, although we've arrived at the latest listen.
3. There is a typo in the aforementioned message

Capture d’écran 2020-10-20 à 20 38 13

# Solution

The pagination buttons rely on a few timestamps that are updated when fetching listens, but weren't in the case of listens fetched in the incremental time range retry.
Also stop retry mechanism if we're at the very first page.


MonkeyDo:
So I don't know how or why, but one specific test is passing when I run the tests locally but not with CI.
I'm in sync with the master branch.
The CI error message:
Capture d’écran 2020-11-06 à 17 36 18

And when I try to correct the test as suggested, here's what I get when I run the tests locally:
Capture d’écran 2020-11-06 à 17 35 18

These two timestamps correspond to the timestamps of the first two elements of the `listens` array in `recentListenProps`, the mock react props we use in tests.
The code being tested sets `previousListenTs` to the the timestamp of the first element of the array, and that should be `1586513524`.
I cannot understand why the timestamp of the second listen item (`1586513255`) gets set instead, when the code specifically uses `previousListenTs: newListens?.[0]?.listened_at`.
I can't understand why in the CI tests the second item's timestamp is used instead, and I can't understand why I get two different results depending on where I'm running the test.
That suggests to me that either the test fixtures are different or the order of the listens is altered, but why only in CI?

Help! @paramsingh , @shivam-kapila do you have any idea

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -978,13 +978,14 @@ describe("Pagination", () => {
...
+        // We're not expecting to see this ts as it will be updated by checkListensRange
         previousListenTs: 1234567881,
         listens: [],
       });
-
+      const sortedListens = sortBy(listens, "listened_at");
       const getListensForUserSpy = jest
         .fn()
-        .mockImplementation(() => Promise.resolve(listens));
+        .mockImplementation(() => Promise.resolve(sortedListens.slice(0, 25)));
```
MonkeyDo, line 988: I've had to return only 25 items (as expected from the API) because on Node 10, somehow the order of sortedListens changes during execution of the test.
See https://tickets.metabrainz.org/browse/LB-756

shivam-kapila (approved):
LGTM overall. I agree we should update node soon

MonkeyDo:
Thanks for the review @shivam-kapila !

Do you think you could help me find the SO item you mentioned that might be related to this weird node version issue?
https://tickets.metabrainz.org/browse/LB-756

shivam-kapila:
@MonkeyDo I am looking for it. I will post in the ticket when I find it

PR #1149, opened 2020-10-21
Sync recommendation feedback with emoticons

vansika:
This PR syncs recommendation feedback with emoticons i.e submit/delete feedback as the user clicks on an emoji


listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -1,36 +1,129 @@
...
+    }
+    newAlert(
+      "danger",
+      title || "Error",
+      typeof error === "object" ? error.message : error
+    );
+  };
+
+  handleClick(e) {
+    e.stopPropagation();
```
vansika, line 121: @MonkeyDo this does not stop the drop-down from closing on click

vansika:
@MonkeyDo also what do you feel about the general flow of it?

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
MonkeyDo: I'll have to look into this

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -45,16 +45,109 @@
...
           display: flex;
+
+          > svg:hover {
+            cursor: pointer;
+          }
         }
 
-        .smile, .laugh, .angry, .disappoint, .meh {
-          font-size: 18px;
+        .fa-angry, .fa-frown, .fa-meh, .fa-smile-beam, .fa-grin-stars, .fa-ellipsis-v {
```
MonkeyDo, line 60: We shouldn't be using the fontawesome classnames here, because it makes it hard to replace them with something else in the future (or with userscripts I guess).

Instead, each feedback item should have its own classname that you can target, the way you have it below like `&.angry{…`.
So instead of `.fa-angry { &.angry …` I think it would be better to have something like `.angry { &.active …` or `.angry { &.selected …` or some other word to denote that that particular feedback has been selected.

MonkeyDo (changes requested):
Just a quick first pass review (sorry, I have to run out!) which I can complete later.
At least I commented on the main things that would require the biggest changes.

listenbrainz/db/recommendations_cf_recording_feedback.py
```diff
@@ -105,3 +106,32 @@ def get_feedback_count_for_user(user_id: int) -> int:
...
+            user_id: the row ID of the user in the DB
+            recording_list: list of recording_mbid for which feedback records are to be obtained
+                            - if record is present then return it
+                            - if record is not present then return rating = None
+
+        Returns:
+            A list of Feedback objects
+    """
+
+    args = {"user_id": user_id, "recording_list": tuple(recording_list)}
```
vansika, line 124: @mayhem okay so the query errored out because I didn't convert the list to tuple. phew

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -62,6 +65,66 @@ export default class Recommendations extends React.Component<
...
+    let recordings = "";
+
+    if (recommendations) {
+      recommendations.forEach((recommendation) => {
+        const recordingMbid = get(
+          recommendation,
+          "track_metadata.additional_info.recording_mbid"
+        );
+        if (recordingMbid) {
+          recordings += `${recordingMbid},`;
```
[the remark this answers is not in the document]
vansika: but why do we want to declare an array and not an empty string?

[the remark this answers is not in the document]
vansika: `Variable 'recordings' implicitly has type 'any[]' in some locations where its type cannot be determined.` is what I get. maybe because APIService.getFeedbackForUserForRecommendations expect recordings to be of type string

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -54,6 +56,7 @@ export default class Recommendations extends React.Component<
             props.recommendations.length / this.expectedRecommendationsPerPage
           )
         : 0,
+      recommendationFeedbackMap: {},
```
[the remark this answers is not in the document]
vansika: umm...we basically need to do a lookup for an mbid to find the corresponding rating. I am not sure how can we store recording_mbid and corresponding rating. list of lists, something like [['mbid A', 'rating A']]? but that's too nested ig.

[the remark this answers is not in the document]
vansika: or maybe have two arrays `recommendationFeedbackMbids` and `recommendationFeedbackRating` 
`recommendationFeedbackMbids`  = [mbid 1, mbid 2]
`recommendationFeedbackRating`  = [rating1, rating 2]
 where
`recommendationFeedbackMbids[0]`  corresponds to `recommendationFeedbackRating[0]`
and so on...

but this can be messy to maintain and error-prone ig

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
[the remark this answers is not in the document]
vansika: `Argument of type 'any[]' is not assignable to parameter of type 'string'.ts(2345)
Variable 'recordings' implicitly has an 'any[]' type.`

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
vansika: thanks :)

listenbrainz/webserver/static/css/recommendation-page.less
vansika: right, I agree :)

listenbrainz/db/recommendations_cf_recording_feedback.py
mayhem: I had suspected as much -- this can be quite tricky and has caught me off guard a number of times.

mayhem: Looks much much better!

mayhem (approved):
The python/SQL bits look sane to me.

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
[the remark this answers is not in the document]
MonkeyDo: This works for me:
```
getFeedback = async () => {
    const { user, recommendations } = this.props;
    const recordings: string[] = [];

    if (recommendations) {
      recommendations.forEach((recommendation) => {
        const recordingMbid = get(
          recommendation,
          "track_metadata.additional_info.recording_mbid"
        );
        if (recordingMbid) {
          recordings.push(recordingMbid);
        }
      });
      try {
        const data = await this.APIService.getFeedbackForUserForRecommendations(
          user.name,
          recordings.join(",")
        );
        return data.feedback;
      } catch (error) {
        this.newAlert(
          "danger",
          "Playback error",
          typeof error === "object" ? error.message : error
        );
      }
    }
    return {};
  };
```

The three changes are:
- `const recordings: string[] = [];` we defined an empty array of strings with the proper type
- `recordings.push(recordingMbid);`  we're pushing the MBID to the array, faster and easier to manipulate than concatenating to a string. If we needed (or will in the future), we can then use array functions to modify or filter the array, which would be lot easier than if we had a string !
- `recordings.join(",")` we call this only once at the end to generate a string of MBIDs separated by commas (again, faster that modifying the string every time in the forEach loop.

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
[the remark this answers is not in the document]
MonkeyDo: You're right, I wrote too fast yesterday, forget what I said about arrays, it would indeed look awful :)

There are however a couple of things to keep in mind with nested properties in the state, but in short we want to make sure that 1. we make a copy of and modify the map (so that react detects a change)  and 2. We make sure we're not overwriting parts of the state inadvertantly (the calls to setstate are asynchronous).
The TL;DR is this syntax:
`this.setState(currentState => { recommendationFeedbackMap: { ...currentState. recommendationFeedbackMap, 'some-bbid': 0} });`
We give setState a callback function that is called with the previous state (this ensures we're not overwriting other things), make a  copy of the feedbackmap (the object spread syntax "…"), and modify the one element we want.

Some reading material:
https://medium.com/@sydbails/how-i-created-nested-properties-in-my-react-js-state-and-why-you-probably-shouldnt-7d9453340e04
https://reactjs.org/docs/optimizing-performance.html#the-power-of-not-mutating-data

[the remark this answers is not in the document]
vansika: https://youtu.be/k5HR9gJh1rw 
Have taken help from this too. 

So what I understand is, since js/react is asynchronous and also react encourages to not mutate states, we want to copy previous state, mutate the copy and then set it as the current state. 

Since the lang is asynchronous, it might happen that when we mutate the previous state to assign it to current state, the call reverts or something happens and we want to render the previous state but it would lead to wrong results since we have already mutated the previous state. That is why it is encouraged to not mutate the state as such but make a copy. The way I imagine is by pointers, mutating the states without copy is like every state in the hierarchy is pointing to same location in the memory though what we ideally want is to have a separate location (copy of data) for every state. 


I hope I make sense (though I feel I have used the words "previous", " Current", "state" In a confusing manner)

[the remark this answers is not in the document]
vansika: It would have been nice if typescript or the compiler points out that "hey, you are mutating states, be careful", something like that , will help noobs at least.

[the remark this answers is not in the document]
MonkeyDo: >  it might happen that when we mutate the previous state to assign it to current state, the call reverts or something happens

Yes, that's one of the potential issues, nicely summarized :)
In our case, it might be unlikely that a user would click feedback icons so fast that we would run into that issue (click on 'love', quickly click on 'hate', if the async 'hate' setState finishes before the 'love' one finished, the 'love' feedback would be the last applied; confusing for the user…), but let's not code something that allows that.

The second potential issue is that is a bit harder to detect is that when React compares the state  `{foo: {bar:0, baz:1})` and `{foo: {bar:0, baz:2}}` it won't see any difference; for React, that's the same object with the same structure, so don't update anything; it won't make a deep comparison of each nested value because that can be really slow.
That's why we create a *new* `foo` object (using for example `foo2 = {…foo}` or `foo2 = Object.assign({}, foo)`) to ensure React knows something has changed, which will trigger a re-render.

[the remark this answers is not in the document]
vansika: Right, so should we avoid stuff like this too? 
https://github.com/metabrainz/listenbrainz-server/blob/rec-component-feedback/listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx#L117

I have used it at many places to get the recs for the current page, and now I think it isn't the right way, maybe!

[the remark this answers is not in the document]
MonkeyDo: Ahaaa, very interesting question!
In this case, you should be just fine. The reason is that array.slice returns a new array, so no chance for Readct to get confused. 
From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice :
>The slice() method returns a shallow copy of a portion of an array into a new array object

The other thing is that we're not dealing with a nested object here but instead an array. That should be absolutely fine.
It's if you had in your state an javascript object with nested properties (`setState({loading: true, recommendations: {nestedProperty1: true, nestedProperty2: false}})`) that you would have the same problem.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
MonkeyDo: So for some reason I'm having a hard time to run the PR locally, but at first glance, the `e.stopPropagation();` bit should be inside the function called when clicking the buttons `inside` the modal (the `RecommendationControl` components), not on the ellipsis button.

So you'll have something like 
```
<RecommendationControl
  icon={faAngry}
  title="I never want to hear this again!"
  action={(event: React.SyntheticEvent) =>
    this.submitFeedback(
      event,
      feedback === "hate" ? "remove_feedback" : "hate"
    )
  }
  className={`${feedback === "hate" ? " angry" : ""}`}
/>
```

Note that we're receiving the onClick event from RecommendationControl in the bound function (`action={(event: React.SyntheticEvent) => …`) and passing it to submitFeedback as the first argument (` this.submitFeedback(event,…`), so we need to modify submitFeedback like this:

```
submitFeedback = async (
    event: React.SyntheticEvent,
    rating: RecommendationFeedBack
  ) => {
    /* Stop click propagation so that the Bootstrap dropdown does not close on feedback click */
    event.stopPropagation();
    […rest of the function]
```

and the `action` in RecommendationControlProps type like this:
`action?: (event: React.SyntheticEvent) => void;`

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -112,17 +112,16 @@ export default class Recommendations extends React.Component<
     this.setState({ recommendationFeedbackMap });
   };
 
-  updateFeedback = (recordingMbid: string, rating: RecommendationFeedBack) => {
-    const { recommendationFeedbackMap } = this.state;
-    recommendationFeedbackMap[recordingMbid] = rating;
+  updateFeedback = (recordingMbid: string, rating: RecommendationFeedBack | null) => {
+    let recommendationFeedbackMap = {...this.state.recommendationFeedbackMap, [recordingMbid]: rating};
```
vansika, line 116: @MonkeyDo  I think this is the only place where I was mutating the state?

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
vansika: Can you tell me what issue you faced when trying to run it locally?

vansika: and the dropdown still closes :(

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -112,16 +112,25 @@ export default class Recommendations extends React.Component<
...
   };
 
-  updateFeedback = (recordingMbid: string, rating: RecommendationFeedBack | null) => {
-    let recommendationFeedbackMap = {...this.state.recommendationFeedbackMap, [recordingMbid]: rating};
-    this.setState({ recommendationFeedbackMap });
+  updateFeedback = (
+    recordingMbid: string,
+    rating: RecommendationFeedBack | null
+  ) => {
+    this.setState((state) => ({
```
vansika, line 119: @MonkeyDo I think this is the only place where I was mutating the state?

shivam-kapila (commented):
I also wonder where we can show the emoticon itself instead of the ellipsis. Maybe grayed out (or very light grey shade) of the love rec emoticon in case feedback is not given. The ellipsis is not clear about its purpose.

vansika:
@shivam-kapila I agree on showing the emoticons. As far as I remember @MonkeyDo voted for the drop down. Let's have a discussion with @MonkeyDo and then we can move forward accordingly.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
vansika: As decided on IRC, we don't want this feature now.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -177,45 +219,47 @@ export default class RecommendationCard extends React.Component<
...
+              <button
+                className="btn"
                 id="recommendationControlsDropdown"
                 data-toggle="dropdown"
                 aria-haspopup="true"
                 aria-expanded="true"
-              />
+                type="button"
+              >
+                <FontAwesomeIcon icon={icon as IconProp} /> {text}
```
vansika, line 230: @MonkeyDo I haven't used Font Awesome regular icons in this commit. Once we are okay with the general flow of it then we can move to the styling part. Can you please try and run this on your machine and play with the feedback buttons and let me know if we are on the same page regarding the general flow?

vansika: Posting a video on IRC for your reference!

MonkeyDo: I think the general flow works well. It's easy enough for users to understand without any room for doubt.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -234,56 +240,64 @@ export default class RecommendationCard extends React.Component<
...
+                  iconHover={faAngry}
+                  iconNonHover={faAngryRegular}
                   title="I never want to hear this again!"
                   action={() =>
                     feedback === "hate"
                       ? this.deleteFeedback("Like", faThumbsUpRegular)
                       : this.submitFeedback("hate", faAngry, "hate")
                   }
-                  className={`${feedback === "hate" ? " angry" : ""}`}
+                  classNameHover="on"
```
vansika, line 251: @MonkeyDo regular/solid icons work with hover/non-hover state like we wanted to but there is one issue I am stuck at:

If I click on let us say angry face, the next time when I click on `Like` to get the dropdown the angry face is still regular and is not solid. in other words, the icons are rendered solid only on hover and not onclick. I understand what the problem is; I am not managing the classes accordingly, but I still don't understand how can I get around this. I am posting a video on IRC for your reference.

MonkeyDo (changes requested):
Here's the trick for showing the icon on hover and when that feedback is selected:

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -57,36 +225,80 @@ export default class RecommendationCard extends React.Component<
...
-                </span>
-                <span className="smile" title="I like this!">
-                  &#128578;
-                </span>
-                <span className="laugh" title="I really love this!">
-                  &#128516;
-                </span>
+                  action={() =>
+                    feedback === "hate"
+                      ? this.deleteFeedback("Like", faThumbsUpRegular)
```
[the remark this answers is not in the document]
vansika: right, makes sense

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -1,36 +1,207 @@
...
       props.recommendation
     );
   }
 
+  componentDidUpdate(
+    prevProps: RecommendationCardProps,
+    prevState: RecommendationCardState
+  ) {
+    const { currentFeedback } = this.props;
+    if (currentFeedback !== prevProps.currentFeedback) {
```
vansika, line 82: @MonkeyDo I have changed all the bits as per the comments and everything works as expected.  But I noticed one thing, on refresh or fresh loading of the page, the feedback is rendered as expected, for example: if a user has hated a track it will be rendered as `faAngry Hate`, good enough as expected.

But

If I do `next` to move to the next page and do `previous` to move to the previous page the feedback changes from `faAngry Hate`  => `faThumbsUpRegular Like` which isn't expected. I checked the state and realised that the `icon` and `text` states change to the one initialized in constructor. 

```
this.state = {
      feedback: props.currentFeedback || null,
      icon: faThumbsUpRegular,
      text: "Like",
    };
```

The only solution I can think of is do `if (icon !== prevState.icon && text != prevState.text)` and if the condition evaluates to `True` i should update the icon and text accordingly.

Before moving forward, I wanted to ask if this solution makes sense to you and are there any drawbacks to this? I am not really sure about the downsides of `componentDidUpdate`

Posting a video on IRC for your reference.

MonkeyDo: Two changes are needed to get the loading feedback to work with pagination:

1. Call `this.loadFeedback();` in `afterRecommendationsDisplay` (which is called after switching pages each time)
2. Change `getFeedback` to fetch feedback for the recordings in the *state* rather than in the props (i.e the current page the user sees instead of for all recs):
```
getFeedback = async () => {
    const { recommendations } = this.state;
    const { user } = this.props;
```

vansika: @MonkeyDo I understand why we want to do the first step but I don't really understand the purpose of step 2, I mean I don't think step 2 is the reason for feedback not getting updated with `next` and `previous`. Are we doing it to make the rendering process faster (fetching feedback for 25 recs is faster than fetching feedback for 100 recs, though I don't think it really matters)?

Since if we fetch feedback for recording in the props (all recs), the current page recs feedback will also be updated since current page resc are a subset of all recs.

MonkeyDo: You're absolutely right, in fact there should be no need to make a new request each time we change page.

Instead, there's something fishy happening in `componentDidUpdate` in the `RecommendationCard` component, and the icon is not updated. I thought it could be because `setState` is called twice in a row, but that didn't end up being the case. I'm not entirely sure why, but `currentFeedback` is always equal to `prevProps.currentFeedback` and the icon+text aren't updated.

Here's what I've tested and propose:

1. Get rid of `componentDidUpdate` entirely
2. We can get rid of all the items in the state. We don't need to copy the props into state, instead we can use the props directly in the render function
3. Place this snippet in the render function:
```
    const { currentFeedback, recommendation, className } = this.props;
    let icon: any;
    let text: string;
    switch (currentFeedback) {
      case "hate":
        icon = faAngry;
        text = "hate";
        break;
      case "dislike":
        icon = faFrown;
        text = "dislike";
        break;
      case "bad_recommendation":
        icon = faMeh;
        text = "bad";
        break;
      case "like":
        icon = faSmileBeam;
        text = "like";
        break;
      case "love":
        icon = faGrinStars;
        text = "love";
        break;
      default:
        icon = faThumbsUpRegular;
        text = "Like";
        break;
    }
```

This switch statement is equivalent to if/else statements and sets icon and text according to the currentFeedback prop (the prop is always up-to-date in the render function).
Note that I'm not using the state but rather the props, and you'll need to replace `feedback` with `currentFeedback` where you instantiate the RecommendationControl components.

With this the cards are updated when the props change and all looks fine :)

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -57,36 +228,87 @@ export default class RecommendationCard extends React.Component<
...
+                    : icon === faFrown
+                    ? "dislike"
+                    : icon === faMeh
+                    ? "bad"
+                    : icon === faSmileBeam
+                    ? "like"
+                    : icon === faGrinStars
+                    ? "love"
+                    : ""
+                }`}
```
MonkeyDo, line 244: How about using ``className={`btn ${currentFeedback}`}`` here instead?
The only css class you would have to redefine is `bad`-> `bad_recommendation`.
With the added advantage of not requiring specifically these FA icons so they're easier to replace if we ever want to.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
vansika: > currentFeedback is always equal to prevProps.currentFeedback

isn't this something to worry about? Ideally they should not be equal. Can the code break in future if we don't compare `currentFeedback` with `prevProps.currentFeedback`? isn't it an alarm if the state is not changing?

vansika: Should we investigate where the value is coming from in `currentFeedback`?

MonkeyDo: >Should we investigate where the value is coming from in currentFeedback?

I found the root issue: `componentDidUpdate` is not called when the component is initially instantiated, but only on subsequent re-renders

From the docs: https://reactjs.org/docs/react-component.html#componentdidupdate
>componentDidUpdate() is invoked immediately after updating occurs. This method is not called for the initial render.

I think this point out to the fact that we don't need `componentDidUpdate` here, and using such mechanisms to copy the props into state is a bug-prone anti-pattern (for this component).

There's a good explanation why [here](https://daveceddia.com/where-initialize-state-react/)
>Initializing State from Props
In most cases, this is an antipattern. Don’t “copy props into state.” It creates a second source of truth for your data, which usually leads to bugs. One source of truth is best.
Components will already re-render when their props change, so there’s no need to duplicate the props as state and then try to keep it up to date.

>Think of it this way: it’s fine if the state needs a starting value which the component will then control. Ask yourself: does this component “own” the data? Does it just need a jump-start from a prop? Those are good reasons to initialize state from a prop.

For example, in the `Recommendations` component you do use `this.props.recommendations` as a source to set a slice of it in the state (https://github.com/metabrainz/listenbrainz-server/blob/rec-component-feedback/listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx#L47).
It makes sense there to initialize the state from props, as we don't expect the props to change.

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -159,6 +235,7 @@ export default class Recommendations extends React.Component<
   };
 
   afterRecommendationsDisplay() {
+    this.loadFeedback();
```
MonkeyDo, line 238: We're not going to need this, it was a bad suggestion from my part !

vansika: Change getFeedback to fetch feedback for the recordings in the state rather than in the props (i.e the current page the user sees instead of for all recs):

```
getFeedback = async () => {
    const { recommendations } = this.state;
    const { user } = this.props;
```

should revert this too?

MonkeyDo: Yes, that also needs to be reverted; You were absolutely right the first time and we only need to get feedback once, and for all the recommendations.

listenbrainz/webserver/static/css/recommendation-page.less
vansika: So a couple of things here, I guess I am a bit confused:
 ```{`btn ${currentFeedback}`}``` is the button class which changes with feedback
The CSS linked:
```
.btn {
          font-size: 18px;
          color: @blue;
          background-color: white;

          &.hate {
            color: #C62828;
          }

          &.dislike {
            color: #f5873d;
          }

          &.bad_recommendation {
            color: #ebbf0b;
          }

          &.like {
            color: #b0e111;
          }

          &.love {
            color: #32cd32;
          }

          &:hover {
            background-color: @blue;
            color: white;
          }
        }
```
so if feedback == hate, the class will be `className='btn hate`, the colour of the icon will change to red and so on

The CSS for the 5 emoticons. This is not linked with any class as such. It kinds defines the base property of the emotions. 

```
.fa-angry, .fa-frown, .fa-meh, .fa-smile-beam, .fa-grin-stars, .fa-thumbs-up {
          stroke-width: 40px;
          font-size: 18px;
          flex: 1;
          text-align: center;
          margin: 0 0.4em;

        }
        .fa-angry {
          color: #C62828;

        }

        .fa-frown {
          color: #f5873d;

        }

        .fa-meh {
          color: #ebbf0b;

        }

        .fa-smile-beam {
          color: #b0e111;

        }

        .fa-grin-stars {
          color: #32cd32;

        }
```
Our goal is to not use the icon names as it is or `.fa-angry, .fa-frown, .fa-meh, .fa-smile-beam, .fa-grin-stars, .fa-thumbs-up`
if we have to use something else other than the names of the icons, we will need to link that to some class since we aren't using the default class, I guess. I just understand the problem statement here, I am not really clear about the solution though.

MonkeyDo: Backtracking on what I said above (again!), sorry about the flip-flopping. As the components change so do the solutions :)

I suggested using the `currentFeedback` as a css class for styling.
That makes styling the feedback button (not the dropdown) pretty easy, but we need to extend it to styling the emoticons in the dropdown.

For the common styling (what affects all emoticons like size, margins, etc.) we can use the recommendation-icon class we hardcoded in `RecommendationControl`.
For styling each emoticon separately (for colours), I propose replacing the `selected` property with a `cssClasses` property and setting it like this in `RecommendationCard`:

```
<RecommendationControl
	iconHover={faGrinStars}
	icon={faGrinStarsRegular}
	title="I really love this!"
	action={() =>
	currentFeedback === "love"
		? this.deleteFeedback()
		: this.submitFeedback("love", faGrinStars, "love")
	}
	cssClass={`love ${currentFeedback === "love" ? "selected" : ""}`}
/>
```

You'll notice we are moving the logic of deciding whether to have the 'selected' class applied from the RecControl to the RecCard component.
In RecControl, we'll simply have this:

```
const { cssClass, iconHover, icon, action, title } = props;
  return (
    <div
      className={`recommendation-icon ${cssClass}`}
```

We'll end up with something like this in the page:
Capture d’écran 2020-11-02 à 12 19 23

From here it should be pretty easy to style the buttons accordingly:
Here's the entirety of my `.recommendation-control` in the LESS file:

```
.recommendation-controls {

        &.open .dropdown-menu {
          display: flex;
        }

        .btn {
          font-size: 18px;
          color: @blue;
          background-color: white;

          &:hover {
            background-color: @blue;
            color: white;
          }
        }
        
        .hate {
          color: #C62828;
        }

        .dislike {
          color: #f5873d;
        }

        .bad_recommendation {
          color: #ebbf0b;
        }

        .like {
          color: #b0e111;
        }

        .love {
          color: #32cd32;
        }
        
        .recommendation-icon > span {
          stroke-width: 40px;
          font-size: 18px;
          flex: 1;
          text-align: center;
          margin: 0 0.4em;
        }

        .recommendation-icon:hover .off,
        .recommendation-icon.selected .off,
        .recommendation-icon .on {
          display: none;
        }
        .recommendation-icon:hover .on,
        .recommendation-icon.selected .on {
          display: inline;
        }

      }
```

Note that we're only defining the combination [feedback string class -> color property] once, and they apply to both the 'like' button states and the colors of the emoticons in the dropdown.

vansika:
@MonkeyDo Things are in place now, for the next step, I think we should work on getting the mobile view to render properly.  Is MediaQuery the solution for this? I have seen the usage of MediaQuery in  ListenCard.tsx for the mobile view. Here is a screenshot for your reference.
I think recommendation control component is not getting enough space on the card, maybe. the col-xs-9 specifically.

image

MonkeyDo:
> I think we should work on getting the mobile view to render properly.

Agreed!

I think flexbox is the right solution here, and can replace the bootstrap columns (`col-xs-…`) entirely.
For simplicity's sake, I'll paste below my the full less file and jsx template because It'll be hard to follow otherwise.

I've removed the mentions of bootstrap columns and use flexbox to automatically expand the track-details (with an ellipsis cutoff) while the like button stays aligned to the right.

The LESS file already contains the changes I proposed in my previous comment today, but the JSX doesn't.  (https://github.com/metabrainz/listenbrainz-server/pull/1149#discussion_r515911026).

LESS:
```less
@import "theme/theme.less";

#recommendations {
  &:extend(#listens);
  padding: 10px;

  > .recommendation-card {
    padding: 15px;
    margin-bottom: 7px;
    max-height: 72px;
    display: flex;

    &.current-recommendation {
      background-color: rgba(217, 237, 247, 0.3) !important ;
    }

    .track-details {
      flex: 1;
      overflow: hidden;

      > * {
        overflow-x: hidden;
        text-overflow: ellipsis;
        white-space:nowrap;
        margin-bottom: 0px;
      }
    }

    a {
      word-break: break-word;
    }

    .recommendation-controls {

      &.open .dropdown-menu {
        display: flex;
      }

      .btn {
        font-size: 18px;
        color: @blue;
        background-color: white;

        &:hover {
          background-color: @blue;
          color: white;
        }
      }
      
      .hate {
        color: #C62828;
      }

      .dislike {
        color: #f5873d;
      }

      .bad_recommendation {
        color: #ebbf0b;
      }

      .like {
        color: #b0e111;
      }

      .love {
        color: #32cd32;
      }
      
      .recommendation-icon > span {
        stroke-width: 40px;
        font-size: 18px;
        flex: 1;
        text-align: center;
        margin: 0 0.4em;
      }

      .recommendation-icon:hover .off,
      .recommendation-icon.selected .off,
      .recommendation-icon .on {
        display: none;
      }
      .recommendation-icon:hover .on,
      .recommendation-icon.selected .on {
        display: inline;
      }

    }
  }
}
```

JSX:
```jsx
    return (
      <Card
        onDoubleClick={this.playRecommendation}
        className={`recommendation-card row ${className}`}
      >
          <div title={recommendation.track_metadata.track_name}>
            {getTrackLink(recommendation)}
          </div>
          <small
            className="text-muted"
            title={recommendation.track_metadata.artist_name}
          >
            {getArtistLink(recommendation)}
          </small>
        </div>
        <div className="recommendation-controls">
          <button
            className={`btn ${currentFeedback}`}
            id="recommendationControlsDropdown"
            data-toggle="dropdown"
            aria-haspopup="true"
            aria-expanded="true"
            type="button"
          >
            <FontAwesomeIcon icon={icon as IconProp} /> {text}
          </button>
          <ul
            className="dropdown-menu dropdown-menu-right"
            aria-labelledby="recommendationControlsDropdown"
          >
            <RecommendationControl
              iconHover={faAngry}
              icon={faAngryRegular}
              title="I never want to hear this again!"
              action={() =>
                currentFeedback === "hate"
                  ? this.deleteFeedback()
                  : this.submitFeedback("hate", faAngry, "hate")
              }
              selected={currentFeedback === "hate"}
            />
            <RecommendationControl
              iconHover={faFrown}
              icon={faFrownRegular}
              title="I don't like this!"
              action={() =>
                currentFeedback === "dislike"
                  ? this.deleteFeedback()
                  : this.submitFeedback("dislike", faFrown, "dislike")
              }
              selected={currentFeedback === "dislike"}
            />
            <RecommendationControl
              iconHover={faMeh}
              icon={faMehRegular}
              title="This is a bad recommendation!"
              action={() =>
                currentFeedback === "bad_recommendation"
                  ? this.deleteFeedback()
                  : this.submitFeedback("bad_recommendation", faMeh, "bad")
              }
              selected={currentFeedback === "bad_recommendation"}
            />
            <RecommendationControl
              iconHover={faSmileBeam}
              icon={faSmileBeamRegular}
              title="I like this!"
              action={() =>
                currentFeedback === "like"
                  ? this.deleteFeedback()
                  : this.submitFeedback("like", faSmileBeam, "like")
              }
              selected={currentFeedback === "like"}
            />
            <RecommendationControl
              iconHover={faGrinStars}
              icon={faGrinStarsRegular}
              title="I really love this!"
              action={() =>
                currentFeedback === "love"
                  ? this.deleteFeedback()
                  : this.submitFeedback("love", faGrinStars, "love")
              }
              selected={currentFeedback === "love"}
            />
          </ul>
        </div>
      </Card>
    );
```

MonkeyDo:
I forgot one last thing in the LESS file above, and edited the comment to update it.

listenbrainz/webserver/static/css/recommendation-page.less
vansika: This takes away the drop-down emoticon colours. looking into it.

vansika:
@MonkeyDo  The changes we have made from the above comment, work okay, but I don't really like how it renders. I think the reason is that we have right-aligned the feedback, ideally, it should be left-aligned.  I think if we left-align than the text will come out of the card from the right boundary like it was happening before. I think the flexbox is shifting the text inwards if it is coming out of the card which makes the alignment really off. Also, the artist and track name are too much shifted to the left and the icons/feedback is too much shifted towards right. I personally like the alignment like how it was before. Is there a way to keep that alignment intact when dealing with mobile view. Here is a screenshot for your reference.

Before:
image

After:
image

MonkeyDo:
Yes we can adjust it to align like before:

First things first, the easy one: the space between the song titles/artists and the card you can adjust with `.recommendation-card`'s `padding` properties.

For aligning the like buttons left, we can add a small css rule to `.recommendation-controls` : `flex: 0 130px;'.
This instructs a flex item to not automatically expand and have a base width of 130px (the width of the longest button, "dislike") so that all the .recommendation-controls elements will be the same width, and we're back to the previous alignment.

For example:
Capture d’écran 2020-11-02 à 17 24 08

I would suggest to also add css media queries so that we can save a bit of space in mobile view, probably first with reducing the left-padding in small sizes. For example:
Capture d’écran 2020-11-02 à 17 35 51

vansika:
Cool, I think things are in place now. A few nitpicks, I want your feedback on them:
1) I personally like the hover box colour (blue), do you?
2) I don't have a strong opinion on the text and icon colour on hover (it is white) but should we consider changing it to the actual feedback colour. For eg `Red` color for Hate etc
3) I only like the hate icon colour, I guess the other feedback icon colors need some refactoring, especially the `Like` and `Bad`, the colour makes them difficult to read.

Here is an image for your reference.
image

vansika:
Also, Can you run it and have a look at the mobile view? I think it looks great, just want a thumbs up from you :)

vansika:
This is how it looks for user != currentUser

image

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -62,6 +65,76 @@ export default class Recommendations extends React.Component<
     );
   }
 
+  componentDidMount(): void {
+    const { user, currentUser } = this.props;
+    const { currRecPage } = this.state;
+    if (currentUser?.name === user?.name) {
+      this.loadFeedback();
+    }
+    window.history.replaceState(null, "", `?page=${currRecPage}`);
```
MonkeyDo, line 74: The browser history part is not fully working; I'll note it down here for posterity but I think it should be fixed in another PR.
I don't mind doing that PR if you want.

`currRecPage` is set in the state in the constructor to `1`, so if I visit the recs page with `?page=2` in my URL (suppose for example I want to send you the second page of my recs), I'll land on page 1 regardless of my URL parameter.

MonkeyDo (changes requested):
This is looking quite good now. I think this should be the last round of review and I consider it ready to be unleashed 🚀 

To answer your questions:

1. I don't mind the blue hover color for buttons, and it is consistent with other buttons on the website.
2. I think we should keep the default behavior and colors of hovered buttons, it work well as it is and I don't see a good reason to want tot change it.
3. Interestingly, I might not be getting the same colors as you do. Here's a screenshot comparing the screenshot you sent and the listenbrainz logo from the website:
Capture d’écran 2020-11-03 à 11 41 47
The blue on your screenshot appears a lot more purple for some reason, and I do agree the emoticon colors are also harder to read on your screenshot. I agree we might need to adjust those colors a bit, maybe after some user feedback?

Out of curiosity, if I screenshot the page on my computer, do the colors look better for you?
Capture d’écran 2020-11-03 à 12 05 42

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
vansika: Yes, we have a ticket to implement that i.e. take `page` arg from URL and jump to that page, I think we need to slice the `recommendations` prop using `page` and update the page to that. Will do in a separate PR. 

> I don't think using the browser back button to go to the next page is working either.

But I am not sure what is causing this and how to fix this? Is there a fix in RecentListens for this?

vansika:
I really think that colour for `Like` (light green) should be changed but I am not very rigid here. As you said, we can wait for user/community feedback.  I think we can leave it for now :)

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
MonkeyDo: Yep! https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/RecentListens.tsx#L115-L116

The `popstate` event triggers when the browser history changes (for example clicking the browser's 'previous' button).

The `handleURLChange` mechanism will not be 100% the same, because in RecentListens each time we call the api to load listens, while in recs page the recommendations will already be there in props and we just need to set a new slice in the state instead, but that gives you an idea of the implementation

MonkeyDo (approved):
🚀

vansika:
I think everything is in place and I can move forward to write tests!

listenbrainz/webserver/static/js/src/recommendations/Recommendations.test.tsx
```diff
@@ -101,6 +355,7 @@ describe("handleClickNext", () => {
 
     await instance.handleClickNext();
 
+    expect(wrapper.state("loading")).toBeFalsy();
```
vansika, line 358: This fails which is mysterious to me because `afterRecommendationsDisplay` does sets the state=`loading` to false
Can you have a look?

MonkeyDo: So on line 349 we're overwriting the  `afterRecommendationsDisplay` method entirely, so `setState({loading:false})` is never called.

Instead, we want to spy on the existing method without replacing it, like this:
```js
const afterRecommendationsDisplaySpy = jest.spyOn(
      instance,
      "afterRecommendationsDisplay"
    );
[…]
    expect(afterRecommendationsDisplaySpy).toHaveBeenCalledTimes(1);
```

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -62,6 +65,76 @@ export default class Recommendations extends React.Component<
...
+        return data.feedback;
+      } catch (error) {
+        this.newAlert(
+          "danger",
+          "Playback error",
+          typeof error === "object" ? error.message : error
+        );
+      }
+    }
+    return {};
```
MonkeyDo, line 105: Found another small bug:
```suggestion
    return [];
```
Since the API returns an array, and we're using `feedback.forEach(` in the `loadFeedback` function (`forEach` is a method of array, no of objects), when there is no feedback currently an error will be thrown: `TypeError: feedback.forEach is not a function`

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -1,96 +1,275 @@
...
+  faFrown as faFrownRegular,
+  faMeh as faMehRegular,
+  faSmileBeam as faSmileBeamRegular,
+  faGrinStars as faGrinStarsRegular,
+} from "@fortawesome/free-regular-svg-icons";
+
 import { get as _get } from "lodash";
-import { faEllipsisV } from "@fortawesome/free-solid-svg-icons";
 import { IconProp } from "@fortawesome/fontawesome-svg-core";
+import { IconDefinition } from "@fortawesome/fontawesome-common-types";
```
vansika, line 21: This fails even though the same import is used in RecommendationControl here 

```
> listenbrainz@1.0.0 format /code
> eslint ./static/js/src --ignore-path .gitignore --ext .js,jsx,ts,tsx --fix --quiet


/code/static/js/src/recommendations/RecommendationCard.tsx
  21:32  error  Unable to resolve path to module '@fortawesome/fontawesome-common-types'  import/no-unresolved

✖ 1 problem (1 error, 0 warnings)

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! listenbrainz@1.0.0 format: `eslint ./static/js/src --ignore-path .gitignore --ext .js,jsx,ts,tsx --fix --quiet`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the listenbrainz@1.0.0 format script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-11-05T15_32_35_543Z-debug.log
```

MonkeyDo: If you look more closely at the code line you linked, you'll notice it has a `// eslint-disable-line import/no-unresolved` commet which instructs eslint to ignore this issue.
That's why it's working there, at least, although I can't tell you what the actual issue is
Considering everything is working and the types are only used for development/compilation, I would just slap the same comment as in the other file and be done with it :)

vansika: Oooo.... I didn't know comments can be that useful ;)

vansika:
@MonkeyDo though I have tried to cover everything in the tests (fe), can you sanity check once if everything is in place? (given tests are really imp so that things don't break :))

vansika:
@alastair @mayhem can you have a look at the python bits (db+api+tests)?

MonkeyDo (changes requested):
A couple of corrections on the tests, and a main suggestion to throw in a check of the html elements wherever you can (everywhere you're expecting visual changes).

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -16,25 +16,249 @@ const recommendation: Recommendation = {
...
+    const spy = jest.spyOn(instance.APIService, "submitRecommendationFeedback");
+    spy.mockImplementation(() => Promise.resolve(200));
+
+    await instance.submitFeedback("hate");
+
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalledWith("lalala", "yyyy", "hate");
+
+    expect(instance.props.updateFeedback).toHaveBeenCalledTimes(1);
+    expect(instance.props.updateFeedback).toHaveBeenCalledWith("yyyy", "hate");
```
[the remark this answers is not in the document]
vansika:
```
it("calls API, calls updateFeedback correctly", async () => {
    const wrapper = shallow<RecommendationCard>(
      <RecommendationCard {...props} />
    );
    const instance = wrapper.instance();

    const wrapperRec = shallow<Recommendations>(
      <Recommendations
        {...(JSON.parse(
          JSON.stringify(recommendationprops)
        ) as RecommendationsProps)}
      />
    );
    const instanceRec = wrapperRec.instance();
    const updateFeedbackSpy = jest.spyOn(instanceRec,"updateFeedback");

    const spy = jest.spyOn(instance.APIService, "submitRecommendationFeedback");
    spy.mockImplementation(() => Promise.resolve(200));

    await instance.submitFeedback("hate");

    expect(spy).toHaveBeenCalledTimes(1);
    expect(spy).toHaveBeenCalledWith("lalala", "yyyy", "hate");
    expect(updateFeedbackSpy).toHaveBeenCalledTimes(1);
    expect(updateFeedbackSpy).toHaveBeenCalledWith("yyyy", "hate");
```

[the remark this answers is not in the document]
vansika: something like this?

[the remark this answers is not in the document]
vansika: The error

 static/js/src/recommendations/RecommendationCard.test.tsx
  ● submitFeedback › calls API, calls updateFeedback correctly

    expect(jest.fn()).toHaveBeenCalledTimes(expected)

    Expected number of calls: 1
    Received number of calls: 0

       98 |     expect(spy).toHaveBeenCalledTimes(1);
       99 |     expect(spy).toHaveBeenCalledWith("lalala", "yyyy", "hate");
    > 100 |     expect(updateFeedbackSpy).toHaveBeenCalledTimes(1);
          |                               ^
      101 |     expect(updateFeedbackSpy).toHaveBeenCalledWith("yyyy", "hate");
      102 |     expect(wrapper.find('.recommendation-controls').childAt(0).childAt(2)).toEqual('Like');
      103 |   });

      at static/js/src/recommendations/RecommendationCard.test.tsx:100:31
      at step (static/js/src/recommendations/RecommendationCard.test.tsx:44:23)
      at Object.next (static/js/src/recommendations/RecommendationCard.test.tsx:25:53)
      at fulfilled (static/js/src/recommendations/RecommendationCard.test.tsx:16:58)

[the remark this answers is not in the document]
MonkeyDo: I see the error you mentioned in IRC. Basically I spoke to quickly again :)
`updateFeedback` is not a method of RecommendationCard, but instead a method passed as a prop from its parent component, so what I suggested isn't going to work like that.

Here's how we'll do it: considering we don't have a real `updateFeedback` method (because we don't have a parent `Recommendations` component or its state), we can fake what is supposed to happen by updating the `RecommendationCard`'s *props*:
```javascript
it("calls API, calls updateFeedback correctly", async () => {
    const wrapper = shallow<RecommendationCard>(
      <RecommendationCard {...{ ...props, updateFeedback: jest.fn() }} />
    );
    const instance = wrapper.instance();

    const spy = jest.spyOn(instance.APIService, "submitRecommendationFeedback");
    spy.mockImplementation(() => Promise.resolve(200));

    // Check initial values of HTML elements (I've done just this one for proof-of-concept)
    expect(
      wrapper.find(".recommendation-controls").childAt(0).hasClass("love")
    ).toEqual(true);

    await instance.submitFeedback("hate");

    expect(spy).toHaveBeenCalledTimes(1);
    expect(spy).toHaveBeenCalledWith("lalala", "yyyy", "hate");

    expect(instance.props.updateFeedback).toHaveBeenCalledTimes(1);
    expect(instance.props.updateFeedback).toHaveBeenCalledWith("yyyy", "hate");

    // Emulate updating the props after calling Recommendations component's updateFeedback
    wrapper.setProps({ currentFeedback: "hate" });
    
    // Check that HTML elements have changed accordingly
    expect(
      wrapper.find(".recommendation-controls").childAt(0).hasClass("hate")
    ).toEqual(true);
  });
```

This remains a valid test; since we're making sure `updateFeedback` is called with the right arguments and we're not trying to test `updateFeedback`.
Notice also that I'm first checking the HTML elements before calling `submitFeedback` so that I can compare before/after.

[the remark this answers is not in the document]
vansika: So just to make sure if I understand this correctly:
1. `submitRecommendationFeedback` calls `updateFeedback`
2. `updateFeedback` is basically a map where Feedback states of all the recs are stored and updated as and when they change.
3. `currentFeedback` gets the value from the map (recommendationFeedbackMap) which is maintained by `updateFeedback` 
4. the value of `currentFeedback` is directly responsible for the values of HTML elements (text, icon etc)
5. So we are doing two things here, first, checking if `updateFeedback` is called with the right arguments, second, tweaking value of `currentFeedback` to check if values of HTML elements are being updated. 

Though the two things mentioned in the point 5 are not dependent on each other i.e we are explicitly changing the `currentFeddback` value and it is not being changed on just calling `submitFeedback`, so we could have separate tests for those but I think it is semantically correct to test them in one function :)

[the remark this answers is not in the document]
vansika: also, I had a feeling that spying on `updateFeedback` is not a good idea because it is not a function of this script, it will be checked anyway in Recommendations.test.tsx. It is better to check its effects. Thanks for confirming that :)

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -16,25 +16,249 @@ const recommendation: Recommendation = {
...
+  it("renders correctly if isCurrentUser is true and CurrentUser.authtoken is set", () => {
     const wrapper = mount<RecommendationCard>(
       <RecommendationCard {...props} />
     );
 
     expect(wrapper).toMatchSnapshot();
   });
+  it("renders correctly if isCurrentUser is False", () => {
+    const wrapper = mount<RecommendationCard>(
+      <RecommendationCard {...{ ...props, isCurrentUser: false }} />
```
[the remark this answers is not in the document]
vansika: removing the extra braces gives me `Left side of comma operator is unused and has no side effects` and `Cannot find name 'isCurrentUser'`. Should we just keep the double braces for now?

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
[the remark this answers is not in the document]
vansika: > wrapper.find('.recommendation-controls').childAt(0).childAt(0).prop("icon").iconName -> this would be the FA icon name, not 100% certain we want to test on that, but I don't have a better idea

this is how it will be:
`expect(wrapper.find('.recommendation-controls').childAt(0).childAt(0).prop("icon").iconName).toEqual("grin-stars");`
I think we should test the icon names also, to ensure correct emoticon is being rendered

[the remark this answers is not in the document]
vansika:
```it("does nothing if isCurrentUser is false", async () => {
    const wrapper = shallow<RecommendationCard>(
      <RecommendationCard
        {...{ ...props, isCurrentUser: false, updateFeedback: jest.fn() }}
      />
    );
    const instance = wrapper.instance();

    const spy = jest.spyOn(instance.APIService, "submitRecommendationFeedback");
    spy.mockImplementation(() => Promise.resolve(200));

    instance.submitFeedback("dislike");
    expect(spy).toHaveBeenCalledTimes(0);
    expect(instance.props.updateFeedback).toHaveBeenCalledTimes(0);
    expect(
      wrapper.find(".recommendation-controls").childAt(0).hasClass("love")
    ).toEqual(true);
    expect(wrapper.find('.recommendation-controls').childAt(0).childAt(0).prop("icon").iconName).toEqual("grin-stars");
    expect(wrapper.find('.recommendation-controls').childAt(0).childAt(2).text()).toEqual("Love");
  });
```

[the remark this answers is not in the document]
vansika: The above fails with following error which is fishy

submitFeedback › does nothing if isCurrentUser is false

    Method “childAt” is meant to be run on 1 node. 0 found instead.

      116 |     expect(instance.props.updateFeedback).toHaveBeenCalledTimes(0);
      117 |     expect(
    > 118 |       wrapper.find(".recommendation-controls").childAt(0).hasClass("love")
          |                                                ^
      119 |     ).toEqual(true);
      120 |     expect(wrapper.find('.recommendation-controls').childAt(0).childAt(0).prop("icon").iconName).toEqual("grin-stars");
      121 |     expect(wrapper.find('.recommendation-controls').childAt(0).childAt(2).text()).toEqual("Love");

      at ShallowWrapper.single (node_modules/enzyme/src/ShallowWrapper.js:1652:13)
      at ShallowWrapper.childAt (node_modules/enzyme/src/ShallowWrapper.js:1248:17)
      at static/js/src/recommendations/RecommendationCard.test.tsx:118:48
      at step (static/js/src/recommendations/RecommendationCard.test.tsx:44:23)
      at Object.next (static/js/src/recommendations/RecommendationCard.test.tsx:25:53)
      at static/js/src/recommendations/RecommendationCard.test.tsx:19:71
      at Object.<anonymous>.__awaiter (static/js/src/recommendations/RecommendationCard.test.tsx:15:12)
      at Object.<anonymous> (static/js/src/recommendations/RecommendationCard.test.tsx:103:48)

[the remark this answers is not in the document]
vansika: Also, I think we should test how button text/icon change with solid emoticons in the dropdown, basically we should test the interdependency of dropdown and button

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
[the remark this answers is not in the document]
MonkeyDo: Well, looks like I was wrong, so indeed let's not break the code because monkey thought he knew better :p

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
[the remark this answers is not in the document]
MonkeyDo: > The above fails with following error which is fishy

If I'm not mistaken, we don't show that feedback component if `isCurrentUser` is false, so actually that should be expected.
You could ensure that in your test with something like:
`expect(wrapper.exists(".recommendation-controls")).toEqual(false);`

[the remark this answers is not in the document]
MonkeyDo: > Also, I think we should test how button text/icon change with solid emoticons in the dropdown, basically we should test the interdependency of dropdown and button

I think we should wait for that one until we have revamped the feedback component.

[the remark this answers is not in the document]
vansika: > If I'm not mistaken, we don't show that feedback component if isCurrentUser is false, so actually that should be expected.
You could ensure that in your test with something like:
expect(wrapper.exists(".recommendation-controls")).toEqual(false);

Oh right. My bad :p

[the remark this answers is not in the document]
vansika: > I think we should wait for that one until we have revamped the feedback component.

I see. ummm....should we open a ticket for this? Though opening a ticket at this point for tests on this specific feature is not really ideal since we the feedback component will change a lot after this PR but also I want to make sure that we don't forget this. The dependency between `button` and `dropdown` is the soul of this PR. What do you think?

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -16,25 +22,297 @@ const recommendation: Recommendation = {
...
+    spy.mockImplementation(() => Promise.resolve(201));
+
+    instance.submitFeedback("bad_recommendation");
+
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalledWith("lalala", "yyyy", "bad_recommendation");
+
+    expect(props.updateFeedback).toHaveBeenCalledTimes(0);
+    expect(
+      wrapper.find(".recommendation-controls").childAt(0).hasClass("love")
```
vansika, line 158: I think we don't need to check HTML elements at places where updateFeedback is not called at all. What do you think?

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -16,25 +22,297 @@ const recommendation: Recommendation = {
...
+    instance.submitFeedback("dislike");
+    expect(instance.handleError).toHaveBeenCalledTimes(1);
+    expect(instance.handleError).toHaveBeenCalledWith(
+      "Error while submitting recommendation feedback - error"
+    );
+    expect(props.updateFeedback).toHaveBeenCalledTimes(0);
+    expect(
+      wrapper.find(".recommendation-controls").childAt(0).hasClass("love")
+    ).toEqual(true);
+    expect(wrapper.find('.recommendation-controls').childAt(0).childAt(0).prop("icon").iconName).toEqual("grin-stars");
```
vansika, line 186: Same thing here, no need to check. In cases where updateFeedback should be called, we are checking the HTML elements after explicitly changing current Feedback (since it is a func of some other script). But in cases where updateFeedback is not called or is not behaving as expected, we need not check the elements,  I think.

vansika:
Also, > expect(wrapper.props('cssClass')).toEqual("bad_recommendation"); this makes sense to me, do you think there is something nuggy about it as it the test fails.  `Expected 0 arguments, but got 1.`

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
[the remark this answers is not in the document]
MonkeyDo: You're absolutely right, let's do it properly and add the necessary tests (initial state with and without feedback, and a test [or expect(…)] each for after setting and after removing a feedback).

And when we get to modifying the component, we'll have ensured we haven't broken anything.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
MonkeyDo: I guess it's a good way to ensure that the feedback hasn't changed, but perhaps it is better done by checking the value of `instance.props.currentFeedback` instead, especially if you are making separate tests for the soul of the PR, as you said :)

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -16,25 +22,297 @@ const recommendation: Recommendation = {
...
+    expect(instance.props.updateFeedback).toHaveBeenCalledTimes(0);
+    expect(wrapper.exists(".recommendation-controls")).toEqual(false);
+  });
+
+  it("doesn't call updateFeedback if status code is not 200", async () => {
+    const wrapper = shallow<RecommendationCard>(
+      <RecommendationCard {...props} />
+    );
+    const instance = wrapper.instance();
+    props.updateFeedback = jest.fn();
```
MonkeyDo, line 146: Something I hadn't caught earlier that I think negates all the tests that have this structure (`props.updateFeedback = jest.fn();` followed by `expect(props.updateFeedback)…`).

I believe it's replacing the function in the `props` object created at the top of the test file, not modifying the props of the component being tested. The wrapper or instance won't automatically update its props like that, especially since we're passing a clone of the props (with the spread syntax to copy the object `{…props}`) to start with.

Instead it should be something like 
```
    const updateFeedbackSpy = jest.spyOn(instance.props, "updateFeedback");
[...]
    expect(updateFeedbackSpy).toHaveBeenCalledTimes(0);
```

I spotted 6 cases in this test file, but it's worth having another look and making sure we don't do that anywhere else.

MonkeyDo:
> Expected 0 arguments, but got 1

That means the `wrapper.props()` method can only be called as is, without any argument.
Here's what my code editor's intellisense says:  "Returns the props hash for the current node of the wrapper."

So supposedly the syntax will more likely be something like `wrapper.props().cssClass`

listenbrainz/webserver/static/js/src/recommendations/Recommendations.test.tsx
```diff
@@ -31,6 +32,256 @@ const props = {
...
+          JSON.stringify(recommendationPropsOne)
+        ) as RecommendationsProps)}
+      />
+    );
+    const instance = wrapper.instance();
+
+    const recommendationFeedbackMap: RecommendationFeedbackMap = {
+      "973e5620-829d-46dd-89a8-760d87076287": "bad_recommendation",
+    };
+    wrapper.setState(JSON.parse(JSON.stringify(recommendationFeedbackMap)));
```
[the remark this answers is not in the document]
vansika: makes sense. Thank you :)

vansika:
@MonkeyDo I think the PR looks good to me now. Have a look!

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -128,22 +140,23 @@ describe("submitFeedback", () => {
...
   });
 
   it("doesn't call updateFeedback if status code is not 200", async () => {
     const wrapper = shallow<RecommendationCard>(
-      <RecommendationCard {...props} />
+      <RecommendationCard {...{ ...props, updateFeedback: jest.fn() }}/>
     );
     const instance = wrapper.instance();
-    props.updateFeedback = jest.fn();
+    const updateFeedbackSpy = jest.spyOn(instance.props, "updateFeedback");
```
MonkeyDo: This only needs one jest function:
```javascript
<RecommendationCard {...props} />
…
const updateFeedbackSpy = jest.spyOn(instance.props, "updateFeedback");
```

There's no need to replace updateFeedback before spying on it.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -153,20 +166,16 @@ describe("submitFeedback", () => {
...
-    expect(wrapper.find('.recommendation-controls').childAt(0).childAt(0).prop("icon").iconName).toEqual("grin-stars");
-    expect(wrapper.find('.recommendation-controls').childAt(0).childAt(2).text()).toEqual("Love");
+    expect(updateFeedbackSpy).toHaveBeenCalledTimes(0);
+    expect(instance.props.currentFeedback).toEqual('love');
   });
 
   it("calls handleError if error is returned", async () => {
     const wrapper = shallow<RecommendationCard>(
-      <RecommendationCard {...props} />
+      <RecommendationCard {...{ ...props, updateFeedback: jest.fn() }} />
```
MonkeyDo, line 191: Same comment as above, no need to replace updateFeedback here.

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -245,44 +255,44 @@ describe("deleteFeedback", () => {
 
   it("does nothing if isCurrentUser is false", async () => {
     const wrapper = shallow<RecommendationCard>(
-      <RecommendationCard {...{ ...props, isCurrentUser: false }} />
+      <RecommendationCard {...{ ...props, isCurrentUser: false, updateFeedback: jest.fn() }} />
```
MonkeyDo: Ditto, no need to replace updateFeedback

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -245,44 +255,44 @@ describe("deleteFeedback", () => {
...
-    expect(instance.props.updateFeedback).toHaveBeenCalledTimes(0);
+    expect(updateFeedbackSpy).toHaveBeenCalledTimes(0);
     expect(wrapper.exists(".recommendation-controls")).toEqual(false);
   });
 
   it("does nothing if CurrentUser.authtoken is not set", async () => {
     const wrapper = shallow<RecommendationCard>(
       <RecommendationCard
-        {...{ ...props, currentUser: { auth_token: undefined, name: "test" } }}
+        {...{ ...props, currentUser: { auth_token: undefined, name: "test"}, updateFeedback: jest.fn() }}
```
MonkeyDo: Here too

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -245,44 +255,44 @@ describe("deleteFeedback", () => {
...
     expect(spy).toHaveBeenCalledTimes(0);
-    expect(instance.props.updateFeedback).toHaveBeenCalledTimes(0);
+    expect(updateFeedbackSpy).toHaveBeenCalledTimes(0);
     expect(wrapper.exists(".recommendation-controls")).toEqual(false);
   });
 
   it("doesn't call updateFeedback if status code is not 200", async () => {
     const wrapper = shallow<RecommendationCard>(
-      <RecommendationCard {...props} />
+      <RecommendationCard {...{ ...props, updateFeedback: jest.fn() }} />
```
MonkeyDo, line 338: Here too

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -292,16 +302,16 @@ describe("deleteFeedback", () => {
...
     expect(spy).toHaveBeenCalledWith("lalala", "yyyy");
 
-    expect(props.updateFeedback).toHaveBeenCalledTimes(0);
+    expect(updateFeedbackSpy).toHaveBeenCalledTimes(0);
   });
 
   it("calls handleError if error is returned", async () => {
     const wrapper = shallow<RecommendationCard>(
-      <RecommendationCard {...props} />
+      <RecommendationCard {...{ ...props, updateFeedback: jest.fn() }} />
```
MonkeyDo, line 356: And here too :)

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
```diff
@@ -313,6 +323,168 @@ describe("deleteFeedback", () => {
...
+    expect(myComponents.get(1).props.cssClass).toEqual('dislike ');
+
+    // validate RecommendationComponent props for 'bad_recommendation' emoticon
+    expect(myComponents.get(2).props.cssClass).toEqual('bad_recommendation ');
+
+    // validate RecommendationComponent props for 'love' emoticon
+    expect(myComponents.get(4).props.cssClass).toEqual('love ');
+  });
+
+  it("check button and dropdown values when currentFeedback == 'Hate' ", async () => {
```
MonkeyDo: Hate -> Love

MonkeyDo (commented):
Final nitpicks!
The tests look solid and now everything is covered nicely!

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.test.tsx
vansika: I think we need it

```
 static/js/src/recommendations/RecommendationCard.test.tsx (84.149s)
  ● submitFeedback › calls API, calls updateFeedback correctly

    TypeError: Cannot assign to read only property 'updateFeedback' of object '[object Object]'

      82 |     );
      83 |     const instance = wrapper.instance();
    > 84 |     const updateFeedbackSpy = jest.spyOn(instance.props, "updateFeedback");
         |                                    ^
      85 | 
      86 |     const spy = jest.spyOn(instance.APIService, "submitRecommendationFeedback");
      87 |     spy.mockImplementation(() => Promise.resolve(200));

      at ModuleMockerClass.spyOn (node_modules/jest-mock/build/index.js:800:26)
      at static/js/src/recommendations/RecommendationCard.test.tsx:84:36
      at step (static/js/src/recommendations/RecommendationCard.test.tsx:44:23)
      at Object.next (static/js/src/recommendations/RecommendationCard.test.tsx:25:53)
      at static/js/src/recommendations/RecommendationCard.test.tsx:19:71
      at Object.<anonymous>.__awaiter (static/js/src/recommendations/RecommendationCard.test.tsx:15:12)
      at Object.<anonymous> (static/js/src/recommendations/RecommendationCard.test.tsx:79:51)
```

MonkeyDo: You're absolutely right.
`read only property 'updateFeedback'` Hm, that's very surprising to me, but hey.

After all, we define `updateFeedback` at the beginning of the file to be a mock function, so it doesn't matter.
When I commented last I thought we were replacing an actual functional method and could possibly interfere with the tests, but that's not the case.

MonkeyDo (approved):
Good to go ! :rocket:

vansika:
Okay, so I think I will test this giant PR one more time (final shot) and then we can merge it!

vansika:
Okay so I have tested the feedback stuff, and it works as expected. @MonkeyDo let's merge? I want to get this merged into the master as the rec-component-feedback branch has grown biggggggggggg

PR #1151, opened 2020-10-21
Improve error management in BrainzPlayer

MonkeyDo:
We want to be a bit more permissive with what type of error we're receiving to display in alerts.
We want to be able to receive regular node Errors, but also to receive HTTP errors which are in another format.

Subsequently, we can take advantage of these changes to display more information from SpotifyPlayer SDK so users have better error messages/codes to quote when running into a bug.

### DIFF
--- a/listenbrainz/webserver/static/js/src/types.d.ts
+++ b/listenbrainz/webserver/static/js/src/types.d.ts
@@ -126,6 +126,11 @@ declare type Alert = {
   message: string | JSX.Element;
 };
 
+// Expect either a string or an Error or an html Response object
+declare type ErrorForAlert =
+  | string
+  | { message?: string; status?: number; statusText?: string };
+
 declare type FollowUsersPlayingNow = {
   [user: string]: Listen;
 };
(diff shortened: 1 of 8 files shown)


listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -126,6 +126,11 @@ declare type Alert = {
   message: string | JSX.Element;
 };
 
+// Expect either a string or an Error or an html Response object
+declare type ErrorForAlert =
```
shivam-kapila, line 130: ErrorForAlert confuses me (IDK why?). I cant think of a better name too. Sorry.

shivam-kapila (approved):
This looks very good especially that that the alerts are more meaningful (and I don't have to jump to console every time :p). One little naming concern but no strong opinions on that too.

listenbrainz/webserver/static/js/src/types.d.ts
MonkeyDo: I agree, not the best naming. I'll think of something.

PR #1153, opened 2020-10-22
Utils for shareable dataframe functions

vansika:
I am working on artist recommendations. The general flow for it will be the same as recording recommendations. The functions used in create_dataframes for recording and artist recommendations are dumped into a single utils files so that they can be shared.


mayhem (approved):
Looks good!

PR #1154, opened 2020-10-23
make separate dirs to store recording and artist recommendation scripts

vansika:
As we are moving on to generate *artists users might like*, we want to cleanly maintain scripts for recording and artist recommendations. In this PR, I have moved recommendation scripts generating *recordings that a user might like* to a separate dir and have renamed files, queries so that devs can understand and differentiate queries that generate recording recs and artist recs.


mayhem (approved):
I'm glad for these refactoring PRs to not contain other changes. Thank you!

PR #1155, opened 2020-10-26
Remove unused devserver and werkzeug profiler code from server startup

alastair:
# Problem

The `werkzeug.contrib.profiler` package was moved to `werkzeug.middleware.profiler` in werkzeug 1.0. 
Newer versions of brainzutils use a newer version of werkzeug, and therefore was causing startup to fail. 
Note that we weren't actually using the profiler - the instantiation was commented out. It was just the import that was causing the startup to fail.

# Solution

Remove unused import.

We also no longer need a dev server startup script here - this has been replaced by `flask run` in docker-compose.yml.

### DIFF
--- a/listenbrainz/server.py
+++ b/listenbrainz/server.py
@@ -1,22 +1,3 @@
-#!/usr/bin/env python
 from listenbrainz.webserver import create_app
-import argparse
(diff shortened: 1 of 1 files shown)


mayhem (approved):
I love deleted code!

PR #1157, opened 2020-10-29
Fix incremental dump creation

ishaanshah:
# Problem
Incremental dumps were getting created using the `listened_at` field instead of `created`. This lead to missing data for people whole imported or submitted listens in bulk.

# Solution
Modifying the query to use `created` fixes the problem.

# Action
We should test this once on the production server and then recreate incremental dumps since the last full dump.

### DIFF
--- a/listenbrainz/listenstore/tests/test_timescalelistenstore.py
+++ b/listenbrainz/listenstore/tests/test_timescalelistenstore.py
@@ -74,9 +74,7 @@ class TestTimescaleListenStore(DatabaseTestCase):
         return len(test_data)
 
     def _insert_with_created(self, listens):
-        """
-            Insert a batch of listens. Returns a list of (listened_at, track_name, user_name) that indicates
-            which rows were inserted into the DB. If the row is not listed in the return values, it was a duplicate.
+        """ Insert a batch of listens with 'created' field.
         """
         submit = []
         for listen in listens:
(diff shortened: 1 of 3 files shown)


listenbrainz/listenstore/tests/test_timescalelistenstore.py
```diff
@@ -72,6 +73,27 @@ def _create_test_data(self, user_name, test_data_file_name=None):
         self.logstore.insert(test_data)
         return len(test_data)
 
+    def _insert_with_created(self, listens):
+        """
+            Insert a batch of listens. Returns a list of (listened_at, track_name, user_name) that indicates
```
mayhem, line 78: This comment isn't right since this function does not return any data.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -342,29 +342,24 @@ def get_listens_query_for_dump(self, start_time, end_time):
...
+    def get_incremental_listens_query(self, start_time, end_time):
         """
             Get a query for a batch of listens for an incremental listen dump.
             This uses the `created` column to fetch listens.
         """
 
         query = """SELECT listened_at, track_name, user_name, created, data
                      FROM listen
-                    WHERE created > :start_ts
+                    WHERE created >= :start_ts
```
mayhem, line 353: Are you sure about this? This change looks like it leads to the oldest listen to be included in two replication packets.

mayhem (commented):
Just some questions for now...

listenbrainz/listenstore/tests/test_timescalelistenstore.py
ishaanshah: Fixed it.

listenbrainz/listenstore/timescale_listenstore.py
ishaanshah: Oops, makes sense, changed it.

mayhem (approved):
Looks good now!

PR #1165, opened 2020-11-02
Fix monthly daily activity stats calculation

ishaanshah:
# Problem
The time range for monthly daily activity was incorrect leading to listens for 0 days to get selected on 1st of each month.

# Solution
This PR changes the time range calculation to make sure at least a single day's listens are included and also catches `ZeroDivisionError` in request consumer.


mayhem (changes requested):
OK, one tiny change, the rest looks good.

PR #1168, opened 2020-11-03
update dataframe id prefix 

vansika:
I have updated the dataframe id prefix in so that it reflects that the dataframes will be used in `recording recommendations`. The metadata saved using the previous prefix will still be there in the parquet!

### DIFF
--- a/listenbrainz_spark/recommendations/recording/create_dataframes.py
+++ b/listenbrainz_spark/recommendations/recording/create_dataframes.py
@@ -122,9 +122,8 @@ from pyspark.sql.functions import rank, col, row_number
 #   ]
 
 
-# Dataframe id is made up of two parts.
-# String + UUID
 # The following var defines the string of which the dataframe id is made up of.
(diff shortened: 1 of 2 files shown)


listenbrainz_spark/recommendations/recording/create_dataframes.py
```diff
@@ -122,6 +122,12 @@
 #   ]
 
 
+# Dataframe id is made up of two parts.
+# String + UUID
```
mayhem, line 126: I find this comment confusing, since there are no UUIDs shown at all. Perhaps explain that it will be added later?

mayhem (approved):
Maybe touch up that comments, but feel free to merge when tests pass if you fix up the comment.

listenbrainz_spark/recommendations/recording/create_dataframes.py
vansika: I completely agree. I will do it in this PR before merging. Thanks :)

PR #1171, opened 2020-11-07
LB-755: Solve 500 in feedback api

shivam-kapila:
# Problem
The `/feedback/user/<user_name>/get-feedback-for-recordings` returns 500 in prod if `recordings` arg is not pased

# Solution
Raise 400 in such cases.

### DIFF
--- a/listenbrainz/tests/integration/test_feedback_api.py
+++ b/listenbrainz/tests/integration/test_feedback_api.py
@@ -666,7 +666,7 @@ class FeedbackAPITestCase(IntegrationTestCase):
         self.assertEqual(response.json["error"], "Cannot find user: nouser")
 
     def test_get_feedback_for_recordings_for_user_no_recordings(self):
-        """ Test to make sure that the API sends 400 if params recordings is not paased or has nothing. """
+        """ Test to make sure that the API sends 400 if param recordings is not passed or is empty. """
         response = self.client.get(url_for("feedback_api_v1.get_feedback_for_recordings_for_user",
                                            user_name=self.user["musicbrainz_id"]))  # missing recordings param
         self.assert400(response)
--- a/listenbrainz/tests/integration/test_feedback_api.py
+++ b/listenbrainz/tests/integration/test_feedback_api.py
@@ -666,12 +666,17 @@ class FeedbackAPITestCase(IntegrationTestCase):
         self.assertEqual(response.json["error"], "Cannot find user: nouser")
 
     def test_get_feedback_for_recordings_for_user_no_recordings(self):
-        """ Test to make sure that the API sends 400 if params recordings has nothing. """
+        """ Test to make sure that the API sends 400 if params recordings is not paased or has nothing. """
+        response = self.client.get(url_for("feedback_api_v1.get_feedback_for_recordings_for_user",
+                                           user_name=self.user["musicbrainz_id"]))  # missing recordings param
+        self.assert400(response)
+        self.assertEqual(response.json["error"], "'recordings' has no valid recording MSID.")
+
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/feedback_api.py
```diff
@@ -180,9 +180,12 @@ def get_feedback_for_recordings_for_user(user_name):
 
     recordings = request.args.get('recordings')
 
+    if not recordings:
```
vansika, line 183: You should also check if passed recording msids are valid uuids.

vansika (changes requested):

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: Thats actually checked in pydantic model right now. e.g. https://api.listenbrainz.org/1/feedback/user/shivam-kapila/get-feedback-for-recordings?recordings=foo

shivam-kapila: Please let me know if this needs change

vansika: Ah cool!

vansika: I will have a second pass tomorrow :)

vansika: https://api.listenbrainz.org/1/feedback/user/Vansika%20Pareek/get-feedback-for-recordings?recording=foo returns 500?

vansika: I am not really sure about the SQL query you are using here, but I think the model will check DB/SQL returned `results` and the not the values/args passed to the SQL query which are directly taken from the URL and may not be in the desired format. have a look :)

vansika: I wrote `recording` in place of `recordings`. Nevamind. Looks good.

vansika (approved):
The changes look fine but ig we should wait till tomorrow, you might want/need to add/delete stuff after the scheduled meeting on feedback.

listenbrainz/tests/integration/test_feedback_api.py
```diff
@@ -666,12 +666,17 @@ def test_get_feedback_for_recordings_for_user_invalid_user(self):
         self.assertEqual(response.json["error"], "Cannot find user: nouser")
 
     def test_get_feedback_for_recordings_for_user_no_recordings(self):
-        """ Test to make sure that the API sends 400 if params recordings has nothing. """
+        """ Test to make sure that the API sends 400 if params recordings is not paased or has nothing. """
```
vansika, line 669: Passed?

vansika: ".. If param `recordings` is not passed or is empty"

listenbrainz/webserver/views/feedback_api.py
shivam-kapila: This 500 will vanish after the PR is moved to prod. But the other one is still 400.

listenbrainz/tests/integration/test_feedback_api.py
shivam-kapila: Makes sense

listenbrainz/webserver/views/feedback_api.py
vansika: Yeah, my bad. I missed the `s` in the spelling :p

shivam-kapila:
@mayhem since the error is still in prod, I think we might want to deploy this fix. Please take a look.

mayhem:
Given that this 500 only occurs when the user uses the API wrong, it doesn't really warrant a separate release.

PR #1173, opened 2020-11-10
LB-717: Return array instead of object when no feedback

MonkeyDo:
Like it says on the tin.
Returning an empty array instead of an object, since we call `.forEach(…)` on it afterwards (`{}.forEach` throws an error).
https://tickets.metabrainz.org/browse/LB-717


shivam-kapila (approved):
🚢... Such changes are a bit controversial and should be brought to the notice of the board XD

MonkeyDo:
> 🚢... Such changes are a but controversial and should be brought to the notice of the board XD

Right?

PR #1177, opened 2020-11-17
Improve formatting of Spotify errors

MonkeyDo:
# Problem
I was seeing some `false` and `undefined` in my error messages.

# Solution
Don't do it dumb.


shivam-kapila:
I was also gonna ask you same. :)

PR #1178, opened 2020-11-19
Critical LB dump fixes

mayhem:
When the listenbrainz-cron-prod container is killed while dumps are being sync'ed to the FTP server, empty dump files may be left in the container which may end up being sync'ed to the FTP server. This has caused 0B *private* dump files to appear on the FTP server. While no private data has been leaked as part of this, we need to ensure that the dump creation process is more robust than it is now. This PR aims to accomplish this.

### DIFF
--- a/listenbrainz/db/dump_manager.py
+++ b/listenbrainz/db/dump_manager.py
@@ -188,7 +188,7 @@ def import_dump(private_archive, public_archive, listen_archive, threads):
             threads (int): the number of threads to use during decompression, defaults to 1
     """
     if not private_archive and not public_archive and not listen_archive:
-        print('You need to enter a path to the archive(s) to import!')
+        current_app.logger.critical('You need to enter a path to the archive(s) to import!')
         sys.exit(1)
 
     app = create_app()
@@ -235,7 +235,7 @@ def _cleanup_dumps(location):
     dump_files = [x for x in os.listdir(location) if full_dump_re.match(x)]
     full_dumps = [x for x in sorted(dump_files, key=get_dump_id, reverse=True)]
     if not full_dumps:
-        print('No full dumps present in specified directory!')
+        current_app.logger.critical('No full dumps present in specified directory!')
     else:
         remove_dumps(location, full_dumps, NUMBER_OF_FULL_DUMPS_TO_KEEP)
 
@@ -243,7 +243,7 @@ def _cleanup_dumps(location):
     dump_files = [x for x in os.listdir(location) if incremental_dump_re.match(x)]
     incremental_dumps = [x for x in sorted(dump_files, key=get_dump_id, reverse=True)]
     if not incremental_dumps:
-        print('No full dumps present in specified directory!')
+        current_app.logger.critical('No full dumps present in specified directory!')
     else:
         remove_dumps(location, incremental_dumps, NUMBER_OF_INCREMENTAL_DUMPS_TO_KEEP)
 
@@ -252,17 +252,17 @@ def remove_dumps(location, dumps, remaining_count):
     keep = dumps[0:remaining_count]
     keep_count = 0
     for dump in keep:
-        print('Keeping %s...' % dump)
+        current_app.logger.info('Keeping %s...' % dump)
         keep_count += 1
 
     remove = dumps[remaining_count:]
     remove_count = 0
     for dump in remove:
-        print('Removing %s...' % dump)
+        current_app.logger.info('Removing %s...' % dump)
         shutil.rmtree(os.path.join(location, dump))
         remove_count += 1
 
-    print('Deleted %d old exports, kept %d exports!' % (remove_count, keep_count))
+    current_app.logger.info('Deleted %d old exports, kept %d exports!' % (remove_count, keep_count))
     return keep_count, remove_count
 
 
@@ -304,7 +304,7 @@ def transmogrify_dump_file_to_spark_import_format(in_file, out_file, threads):
                         if member.name.endswith(".listens"):
                             filename = member.name.replace(".listens", ".json")
                             filename = filename.replace("-full", "-spark-full")
-                            print("mogrify: ", filename)
+                            current_app.logger.info("mogrify: ", filename)
                             tmp_file = tempfile.mkstemp()
                             with os.fdopen(tmp_file[0], "w") as out_f:
                                 with tarf.extractfile(member) as f:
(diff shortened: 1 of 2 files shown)


listenbrainz/db/dump_manager.py
```diff
@@ -157,12 +168,24 @@ def create_incremental(location, threads, dump_id):
...
+        except IOError as e:
+            sys.exit(-1)
 
         # if in production, send an email to interested people for observability
         send_dump_creation_notification(dump_name, 'incremental')
 
+        # Write the DUMP_ID file so that the FTP sync scripts can be more robust
+        print(os.path.join(dump_path, "DUMP_ID.txt"))
+        with open(os.path.join(dump_path, "DUMP_ID.txt"), "w") as f:
+            f.write("%s %s incremental\n" % (end_time.strftime('%Y%m%d-%H%M%S'), dump_id))
```
zas, line 185: Why a newline here and not in full? https://github.com/metabrainz/listenbrainz-server/pull/1178/files#diff-e4f7644db199bd1cedfa7296a724b02679b34ea30b8920fdbae2eff15153cd9cR129

zas (changes requested):

listenbrainz/db/dump_manager.py
mayhem: Good catch. fixing.

mayhem: Fixed

listenbrainz/db/dump_manager.py
```diff
@@ -157,12 +168,24 @@ def create_incremental(location, threads, dump_id):
...
+            if not sanity_check_dumps(dump_path, 6):
+                return sys.exit(-1)
+        except IOError as e:
+            sys.exit(-1)
 
         # if in production, send an email to interested people for observability
         send_dump_creation_notification(dump_name, 'incremental')
 
+        # Write the DUMP_ID file so that the FTP sync scripts can be more robust
+        print(os.path.join(dump_path, "DUMP_ID.txt"))
```
zas, line 183: Is this `print()` statement useful?

listenbrainz/db/dump_manager.py
```diff
@@ -280,6 +306,37 @@ def write_hashes(location):
...
+        expected_count (int): the number of files that are expected to be present
+    Return:
+        boolean: true if the dump passes the sanity check
+    """
+
+    count = 0
+    for file in os.listdir(location):
+        try:
+            dump_file = os.path.join(location, file)
+            print(dump_file)
```
zas, line 323: No use of logger here?

listenbrainz/db/dump_manager.py
```diff
@@ -280,6 +306,37 @@ def write_hashes(location):
...
+            if os.path.getsize(dump_file) == 0:
+                print("Dump file %s is empty!" % dump_file)
+                return False
+            count += 1
+        except IOError as e:
```
zas, line 328: `os.path.getsize()` may raise `OSError`.

In Python3, `os` module always raise `OSError` (but IOError is a deprecated alias), see https://www.python.org/dev/peps/pep-3151/#confusing-set-of-os-related-exceptions about this subject.

https://docs.python.org/3/library/os.path.html#os.path.getsize

listenbrainz/db/dump_manager.py
```diff
@@ -280,6 +306,37 @@ def write_hashes(location):
             raise
 
 
+def sanity_check_dumps(location, expected_count):
+    """ Sanity check the generated dummps to ensure that none are empty
```
zas, line 310:
```suggestion
    """ Sanity check the generated dumps to ensure that none are empty and the expected count was created.
```

listenbrainz/db/dump_manager.py
```diff
@@ -157,12 +168,24 @@ def create_incremental(location, threads, dump_id):
             write_hashes(dump_path)
         except IOError as e:
             current_app.logger.error('Unable to create hash files! Error: %s', str(e), exc_info=True)
-            return
+            sys.exit(-1)
+
+        try:
+            if not sanity_check_dumps(dump_path, 6):
+                return sys.exit(-1)
+        except IOError as e:
```
zas, line 176: See my comment about `OSError` vs `IOError` below.

listenbrainz/db/dump_manager.py
```diff
@@ -111,14 +111,25 @@ def create_full(location, threads, dump_id, last_dump_id):
...
+            sys.exit(-1)
 
         # if in production, send an email to interested people for observability
         send_dump_creation_notification(dump_name, 'fullexport')
 
         current_app.logger.info('Dumps created and hashes written at %s' % dump_path)
 
+        # Write the DUMP_ID file so that the FTP sync scripts can be more robust
+        with open(os.path.join(dump_path, "DUMP_ID.txt"), "w") as f:
+            f.write("%s %s full\n" % (end_time.strftime('%Y%m%d-%H%M%S'), dump_id))
```
zas, line 129: `end_time.strftime('%Y%m%d-%H%M%S')` is done 3 times, perhaps store in a variable to reduce code redundancy.

zas (changes requested):

listenbrainz/db/dump_manager.py
mayhem: The loggers are not available in this context, so I would have to create another logger for some of the code, some of the time. Not what I really want to be doing right now.

listenbrainz/db/dump_manager.py
```diff
@@ -157,12 +169,23 @@ def create_incremental(location, threads, dump_id):
             write_hashes(dump_path)
         except IOError as e:
             current_app.logger.error('Unable to create hash files! Error: %s', str(e), exc_info=True)
-            return
+            sys.exit(-1)
+
+        try:
+            if not sanity_check_dumps(dump_path, 6):
+                return sys.exit(-1)
+        except IOError as e:
```
zas, line 177: `os.listdir()` raises `OSError` (in fact, all `os.` methods do), think about changing those, those are aliases now, but they may not in the future.

listenbrainz/db/dump_manager.py
```diff
@@ -275,11 +301,39 @@ def write_hashes(location):
...
                 sha256sum = subprocess.check_output(['sha256sum', os.path.join(location, file)]).decode('utf-8').split()[0]
                 f.write(sha256sum)
-        except IOError as e:
+        except OSError as e:
             current_app.logger.error('IOError while trying to write hash files for file %s: %s', file, str(e), exc_info=True)
             raise
 
 
+def sanity_check_dumps(location, expected_count):
+    """ Sanity check the generated dumps to ensure that none are empty
```
zas, line 310: You discarded part of my proposed change, but the method doesn't only check if none of the files are empty, but also check if the expected count matches.

zas (approved):

PR #1179, opened 2020-11-19
Make create-dumps.sh script safer

zas:
My changes are doing following: 

- dumps are stored in an unique temp dir, if script exits normally (=not kill -9 or power off), directory is cleaned at exit whatever happens
- it checks for DUMP_ID.txt file existence and parses it (else bail out)
- it checks for empty files or directories (none should exist), bail out if found
- proper quoting of strings


mayhem (approved):
Looks sane.

PR #1184, opened 2020-11-30
Make timescale test setup part of generic testing code

alastair:
# Problem

We need to run unit tests against the timescale database

# Solution
Move timescale database setup to the generic testing module, rather than making it a part of an integration class testcase.

### DIFF
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -17,6 +17,7 @@ services:
 
   timescale:
     image: timescale/timescaledb:latest-pg12
+    command: postgres -F
     environment:
       POSTGRES_PASSWORD: 'postgres'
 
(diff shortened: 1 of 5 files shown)


docker/docker-compose.test.yml
```diff
@@ -17,6 +17,7 @@ services:
 
   timescale:
     image: timescale/timescaledb:latest-pg12
+    command: postgres -F
```
mayhem, line 20: Cute. Can you actually tell it being faster?

mayhem (approved):
Thanks!

docker/docker-compose.test.yml
alastair: yes, when I first did this with AB you could visibly see the tests running faster.

PR #1185, opened 2020-12-01
Add run script for jenkins integration tests, fix jenkins unit tests

alastair:
# Problem

Travis is slooooow

# Solution

Set up unit + integration tests on jenkins again


alastair:
@brainzbot test this please

alastair:
Integration tests fail with
 > python3: can't open file 'manage.py': [Errno 2] No such file or directory

I think this is because the docker-compose file for jenkins integration tests mounts the wrong directory into `/code`

Unit tests fail with
> Cannot import config.py. Waiting and retrying...

We should make sure we copy a config file like we do in the travis tests

alastair:
Frontend tests are run in travis but are missing here

mayhem:
Well, I moved it along, but now I have no idea why all of these tests fail. Can't solely be based on timing, can it?

shivam-kapila:
Not sure if it will help much but the integration tests are failing due to Redis (Not properly setup?). The unit tests are passing but build fails at junit step

listenbrainz/tests/integration/test_api.py
```diff
@@ -133,16 +133,25 @@ def test_get_listens_with_time_range(self):
...
+            # This sleep allows for the timescale subscriber to take its time in getting
+            # the listen submitted from redis and writing it to timescale.
+            # Removing it causes an empty list of listens to be returned.
+            # We try a few times (up to a maximum of 5) to see if the data is present,
+            # but then fail if the data still hasn't arrived
+            time.sleep(2)
+
+            response = self.client.get(url)
+            data = json.loads(response.data)['payload']
+            if data['count'] == expected_count:
```
MonkeyDo, line 150: Is there a possibility that this would result in a false positive, for example if I'm expecting 3 listens but in the end 4 listens arrive, will this test just pass as soon as it gets 3, missing the erroneous 4th?

alastair: 2 things to prevent this from happening:
- we only submit 3 items (the `for i in range(3):....` above)
- we only use this check to break out of the sleep loop quickly. A few lines down we still have an assert to check that the count value is expected. If there are 4, the assert on l154 will fail.

alastair:
Integration tests have one failure, but this is the error
> sqlalchemy.exc.OperationalError: (psycopg2.errors.ObjectInUse) source database "template1" is being accessed by other users
> E       DETAIL:  There is 1 other session using the database.

which I thought I fixed in https://github.com/metabrainz/listenbrainz-server/pull/1185/commits/921034725721a2414927f8b8e47c6b95817406d1

I advocate that we merge now and deal with this if it comes up again

mayhem (approved):
Fantastic! Once the merge conflict is merged, happy to see this rolled out!

listenbrainz/tests/integration/test_api.py
MonkeyDo: Right, thanks :)

MonkeyDo (approved):
Approving the JS / NPM bits

PR #1141, opened 2020-10-15
Upgrade pip in to avoid random failing builds

amCap1712:
Travis builds are failing randomly with the error.
```
Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/pip/_internal/cli/base_command.py", line 176, in main
    status = self.run(options, args)
  File "/usr/local/lib/python3.7/site-packages/pip/_internal/commands/install.py", line 346, in run
    session=session, autobuilding=True
  File "/usr/local/lib/python3.7/site-packages/pip/_internal/wheel.py", line 886, in build
    assert have_directory_for_build
AssertionError
You are using pip version 19.0.1,
```
According to https://github.com/pypa/pip/issues/6197, this is fixed in 19.0.2. Hence, upgrading `pip`.

PR #1192, opened 2020-12-01
Upgrade to a new point release of python 3.7 and pip 20.2

alastair:
# Problem
PR #1141 upgraded pip unconditionally, but this has since caused other problems with the release of pip 20.3 (https://github.com/metabrainz/brainzutils-python/pull/45).

# Solution
Upgrade to a new version of our upstream python images that have a pinned version of pip 20.2, which should fix the errors reported in #1141, but will not cause new errors with pip 20.3

# Action
- Wait for BU 1.16 to be released and upgrade dependency here
- Build + release to test to ensure that all services start up and run properly
- Merge + deploy


alastair:
Replaced by #1271

PR #1193, opened 2020-12-02
Pin an exact version of metabrainz base image and pip

alastair:
# Problem

As described in #1192, LB no longer build with pip 20.3

# Solution
Explicitly install pip 20.2


mayhem (approved):
:+1:

PR #1197, opened 2020-12-03
Merge branch rec-component-feedback to master 

vansika:
I had opened branch *rec-component-feedback* a few months ago to add the feature of [Recommendation Feedback](https://docs.google.com/document/d/1loj903RXCYZgdL9IFGk7SXE54WFkZyN-4Jvr1wJ3WWM/edit?usp=sharing) to ListenBrainz. 
The branch looks good and contains all the features mentioned in the google doc. I think it looks good now and should be merged in master.


mayhem:
I see some tests failing, but I don't think they are related to your changes, are they?

vansika:
@mayhem  yeah, some tests in test_api are failing. Even though the failed build is apparently not related to this PR, but I still want the build to pass before we merge this. It is a big feature addition, just trying to be careful!

@MonkeyDo @ishaanshah can you please have a look at the failing build. https://travis-ci.org/github/metabrainz/listenbrainz-server/builds/747484369

shivam-kapila:
The same tests are failing in other PRs too. Maybe it started after #1184

vansika:
@shivam-kapila ooo, thanks for the update?

@mayhem @MonkeyDo should we merge if it looks good to you?

MonkeyDo (changes requested):
There is one last modification required before we publish this feature, that we talked about a couple of weeks ago ([relevant logs](https://chatlogs.metabrainz.org/brainzbot/metabrainz/2020-11-11/?msg=4699760&page=1): separating "bad_recommendation" from the emoticons feedback.
I think it's important to do this change now since we already know the change needs to happen, and we haven't introduced this feedback component to users yet.

I think the easiest way forward would be to implement the "bad recommendation" UI feature  at a later time, and modify the existing `RecommendationCard` component to replace or remove the neutral emoticon (faMeh) icon.

Removing it would simply leave us with 4 emoticons to choose from, and that would be the quickest way to get the PR merged, but I think there is a better solution:

Currently to ‘un-like' we need to click again on the emoticon we previously selected (if feedback = "like", click on the "like" emoticon to delete the feedback).
What I've heard so far is that this isn't very intuitive for the people who have tried it.
We could instead use the neutral emoticon as the "delete my feedback" button. Semantically for the user this is pretty consistent ("i feel neutral about this track") and with the right hover title should be more obvious to users than having to click again on the previously selected emoticon.

vansika:
> We could instead use the neutral emoticon as the "delete my feedback" button. Semantically for the user this is pretty consistent ("i feel neutral about this track") and with the right hover title should be more obvious to users than having to click again on the previously selected emoticon.

If we want the neutral emoticon to delete the feedback, it shouldn't be displayed in the drop down with other emoticons because it isn't intuitive as well. In my opinion, delete feedback can be a different option (like the trash bin icon) on the card altogether that doesn't looks like an emoticon at all so that there is no room for confusion (emojis like these generally mean give feedback). From what I have seen generally, double click means negating, so I think we can remove the bad_recommendation from this PR, merge and open a ticket about separating the component. One effect from this merge (if released) would be that bad_recommendations enum in db won't be used at all, but that's just fine!

MonkeyDo:
> delete feedback can be a different option (like the trash bin icon) on the card altogether that doesn't looks like an emoticon at all so that there is no room for confusion 

You've got a point. Clearly this needs more thought and time, so I agree:
>I think we can remove the bad_recommendation from this PR, merge and open a ticket about separating the component.

with the understanding that you'll just be removing the neutral `RecommendationControl` component and fixing associated tests.

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,100 @@
...
+
+      .hate {
+        color: #C62828;
+      }
+
+      .dislike {
+        color: #f5873d;
+      }
+
+      .bad_recommendation {
```
MonkeyDo, line 68: As per the main thread discussion, this bit can be removed.

shivam-kapila:
>  delete feedback can be a different option (like the trash bin icon) on the card altogether that doesn't looks like an emoticon at all so that there is no room for confusion
definitely valid point

alastair:
@brainzbot add to whitelist

vansika:
image

the spacing :(

listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx
```diff
@@ -0,0 +1,258 @@
...
+              id="recommendationControlsDropdown"
+              data-toggle="dropdown"
+              aria-haspopup="true"
+              aria-expanded="true"
+              type="button"
+            >
+              <FontAwesomeIcon icon={icon as IconProp} /> {text}
+            </button>
+            <ul
+              className="dropdown-menu dropdown-menu-right"
```
shivam-kapila, line 197: try adding text-center class here to solve spacing issue.

MonkeyDo:
> image
> 
> the spacing :(

This one is pretty easy: the Bootstrap styles for the dropdown has a min width of 160px which is why there is extra space.

Two options to solve it:
1) [preferable I think]: Since we already set the dropdown-menu to use flex, we can simply take advantage of that and tell any children inside the dropdown to take the max space available:
```css:
#recommendations > .recommendation-card .recommendation-controls > * {
    flex: 1;
}
```
Note the `> *` with means "any element that is a direct child of a recommendation-controls element".

2) Overwrite the Bootstrap style for the dropdown-menu element to have a smaller `min-width`.

listenbrainz/webserver/static/css/recommendation-page.less
```diff
@@ -0,0 +1,96 @@
...
+      word-break: break-word;
+    }
+
+    .recommendation-controls {
+
+      flex: 0 140px;
+
+      &.open .dropdown-menu {
+        display: flex;
+        top: auto;
```
MonkeyDo, line 46: Relating to my comment https://github.com/metabrainz/listenbrainz-server/pull/1197#issuecomment-757916336

You can simply add 
```less:
> * {
    flex: 1;
}
``` inside this block.

**Comment by vansika** [comment]
@MonkeyDo The PR looks good and mergeable to me. Though Idk why integration tests fails,  looks unrelated to the PR, have a look :)

**Review by MonkeyDo — APPROVED** [review]
Good to go ! :rocket:

File: listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx [hunk 4]
```diff
@@ -0,0 +1,397 @@
...
+          "track_metadata.additional_info.recording_mbid"
+        );
+        if (recordingMbid) {
+          recordings.push(recordingMbid);
+        }
+      });
+      try {
+        const data = await this.APIService.getFeedbackForUserForRecommendations(
+          user.name,
+          recordings.join(",")
```
MonkeyDo, line 94: Currently on beta, this recordings array contains 1000 items, and the server doesn't like this 1000-mbids-long in that URL. So we'll have to call loadFeedback or getFeedback only for the items in the page the user is currently viewing instead.

listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx
```diff
@@ -0,0 +1,397 @@
...
+    const { user, currentUser } = this.props;
+    const { currRecPage } = this.state;
+    if (currentUser?.name === user?.name) {
+      this.loadFeedback();
+    }
+    window.history.replaceState(null, "", `?page=${currRecPage}`);
+  }
+
+  getFeedback = async () => {
+    const { user, recommendations } = this.props;
```
MonkeyDo, line 78: Follow-up to my previous comment, this should be :
```
const { user } = this.props;
const { recommendations } = this.state;
```
Since the recommendations in the state at any given time are the recs on the page the user is on.

Then we'll need to call `loadFeedback` on each page change as well.

PR #1200, opened 2020-12-04
Fix double encoding of spotify query

MonkeyDo:
# Problem
I introduced a regression with PR #1176 
We're double-encoding the query string which breaks the spotify track search:

We encode for example `hello world` into `hello%20world`, and then the `%` is encoded again into `%25` for a result of `hello%2520world`.

# Solution

Don't call encodeURI after already encoding each search input safely with `encodeURIComponent`.
The rest of the query parameters (`?track=` etc.)we write ourselves so we know it is safe and doesn't need encoding.

### DIFF
--- a/listenbrainz/webserver/static/js/src/utils.tsx
+++ b/listenbrainz/webserver/static/js/src/utils.tsx
@@ -32,7 +32,7 @@ const searchForSpotifyTrack = async (
   }
 
   const response = await fetch(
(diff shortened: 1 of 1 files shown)


shivam-kapila:
The same tests are failing in other PRs too. Maybe it started after #1184

MonkeyDo:
I am merging despite the CI failure; the two failing tests are unrelated to the PR and a know issue with our CI setup

PR #1201, opened 2020-12-08
Remove travis

mayhem:
Thanks to alastair's work we now have tests in jenkins and can drop the sooper slow travis tests.


alastair (approved):
fine by me!

PR #1203, opened 2020-12-10
Webpack typescript checking

MonkeyDo:
Turns out `babel-loader` strips away the typescript types and does no validation, which is the reason we have a separate `type_checker` container.

This PR adds the [fork-ts-checker webpack plugin](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin) which handles type checking—in its own thread so as not to slow compilation down—and removes the `type_checker` container.

In dev, type errors will appear clearly formatted (much more visibly than with `type_checker`) and in prod the webpack build will fail on type errors.

One less container ! 🎉 🍾 🥂


alastair (commented):
Looks good to me, although there is an error on build now that we should fix before merging this

PR #1208, opened 2020-12-23
Do not check for public field while adding tracks

amCap1712:
While adding a new track to the playlist, we do not need to check for the public field. This PR goes a bit further to create separate validators for checking the type of the fields and the data of the playlist.


alastair:
@brainzbot add to whitelist

alastair:
I tested this and still saw a problem when performing an edit with no `public` field set. I reworked it a bit in https://github.com/metabrainz/listenbrainz-server/pull/1209/commits/bc141c1725c9d75c9bbcfa22e572e488ac96d1e0 and applied it there. Thanks!

PR #1209, opened 2020-12-24
Create a URL namespace to redirect users to their profile page

alastair:
# Problem
The user-facing informational pages contain a username in the url, and
so we don't have a specific URL where we can send everyone to with a
link on twitter, etc.

# Solution
 Make a new namespace /my/[page] that redirects
to /user/[username]/page if the user is logged in.


amCap1712 (approved):
Looks great!

PR #1212, opened 2020-12-27
Broadcast playlist edits in real time using websockets

amCap1712:
Edits made by the owner or collaborators on a playlist are sent instantly to all users who are have the playlist opened at that time. This is an initial attempt that just works. 

Things like login authentication, using a queue and fetching playlist in the server itself instead of sending multiple requests from the client so on can be explored and included if needed.

This PR also contains removing the follow server part.  I needed that for easy local testing but for merging that can probably be split into a different PR. I also removed the UNIQUE_EXCHANGE queue from `timescale_writer.py`, as I was getting some errors. I think that was only being used by the follow server. I can try to explore that further if it is required.

## Tasks left

- Integration and unit tests are also required but I am currently not sure as to how to create those.
- Review websockets setup once we remove the follow-server mechanism


docker/docker-compose.yml
```diff
@@ -27,7 +27,7 @@ services:
       - redis:/data:z
 
   timescale:
-    image: timescale/timescaledb:latest-pg12
+    image: timescale/timescaledb:1.7.4-pg12
```
shivam-kapila, line 30: We are not sure about this. Let's wait to hear from ruaok

docs/dev/devel-env.rst
```diff
@@ -181,7 +181,7 @@ different services. We provide a small description of each container here:
 * ``rabbitmq``: Used for passing listens between different services
 * ``web``: This is the main ListenBrainz server
 * ``api_compat``: A Last.fm-compatible API server
-* ``follow_server``: A helper server used for the user-following component of ListenBrainz
+* ``websockets``: A helper server used for the user-following component of ListenBrainz
```
shivam-kapila, line 184:
```suggestion
* ``websockets``: A helper server used for the playlist component of ListenBrainz
```

listenbrainz/websockets/webserver.py
```diff
@@ -0,0 +1,60 @@
...
+    return identifier[idx + 1:]
+
+
+@socketio.on('change_playlist')
+def dispatch_playlist_updates(data):
+    emit('playlist_changed', data, room=get_playlist_id(data))
+
+
+@socketio.on('joined')
+def joined(data):
```
shivam-kapila, line 41: Should we also notify the other users in cases like the collab playlists?

shivam-kapila (commented):
Overall looks so good. Excited to see sockets in action :)

docker/docker-compose.yml
amCap1712: Yup! This is one of the makeshift fixes I needed to make for local dev.

listenbrainz/websockets/webserver.py
amCap1712: That seems to be a different feature (sounds more like notifications). Here, we are just showing updates to anyone who has the playlist opened in their browser.

shivam-kapila: Makes sense

amCap1712:
PR's cleaned up :)

listenbrainz/config.py.sample
```diff
@@ -119,6 +119,7 @@ API_URL = 'https://api.listenbrainz.org'
 LASTFM_PROXY_URL = 'http://localhost:8080/'
 SERVER_ROOT_URL = 'http://localhost'
 WEBSOCKETS_SERVER_URL = 'http://localhost:8081'
+PLAYLIST_WEBSOCKETS_URL = 'http://localhost:8082'
```
alastair, line 122: what's the difference between websockets_server and playlist_websockets? We should be using the same server for both playlists and follow if possible

amCap1712: I think it is possible but I tried to keep everything separate for easier testing and the follow server is going to be removed soon so there will be only one server.

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
```diff
@@ -111,8 +111,9 @@ export default class PlaylistPage extends React.Component<
     });
   }
 
-  async componentDidMount() {
-    // this.connectWebsockets();
+  componentDidMount(): void {
+    this.connectWebsockets();
```
MonkeyDo, line 115: Let's be super clean and add a `componentWillUnmount` method and call `if(this.socket?.connected){this.socket.disconnect()}` in it to make sure we terminate the connection and free up the resources.

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
```diff
@@ -568,6 +579,7 @@ export default class PlaylistPage extends React.Component<
         currentUser.auth_token
       );
       this.setState({ playlist: JSPFObject.playlist });
+      this.emitPlaylistChanged();
```
MonkeyDo, line 582: We should move this line up to right after we get confirmation that the API call succeeded, Line 571 right after `this.newAlert("success", "Saved playlist", "");`.

That way we correctly emit the change even if something goes wrong in this trycatch block.

And behold! Magic happens: we don't actually need this trycatch block at all and we can remove it!
Sicne we get updates through websockets, this whole `Fetch the newly editd playlist and save it to state` is redundant.

MonkeyDo (changes requested):
It was really nice to see everything working as soon as I fired it up, from the WS container to the front-end updating automatically. Nicely done !

A couple of comments below but apart from that it's all working on the front-end side!

MonkeyDo:
I'll make a note to point out that we (alastair, rob and I) discussed a future improvement where the API would directly react when a playlist is modified and have an update sent, rather than require that `emitPlaylistChanged` call from the front-end.

It won't be for right now anyway, and I think the feature already works well.

amCap1712:
>the API would directly react when a playlist is modified and have an update sent, rather than require that emitPlaylistChanged call from the front-end

I too had thought about this. The way I thought we could get this to work is by setting up an update queue similar to what the TimescaleListenWriter does after successful inserts. It would be a bit more work, hence I decided to go with the current simpler implementation first.

MonkeyDo:
> I too had thought about this. The way I thought we could get this to work is by setting up an update queue similar to what the TimescaleListenWriter does after successful inserts. It would be a bit more work, hence I decided to go with the current simpler implementation first.

I completely agree with the approach of Incremental improvements.
I think alastair and rob already have a plan to get the API to communicate with other parts of the project, but I honestly can't remember. 😊

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
MonkeyDo: I see you removed the call to `emitPlaylistChanged`, but that's not what I meant, sorry if I was not clear.

You should indeed call `emitPlaylistChanged` right after `this.newAlert("success", "Saved playlist", "")` and remove the block of code that comes after that fetches the playlist through the API: https://github.com/metabrainz/listenbrainz-server/pull/1212/files#diff-4a293c4781b3b1e784e1cad81c20cd70c3c01f8d7ef9526aab4294126eeeea9eR580-R589

We don't need to call the API anymore since we receive the updated playlist via WS

MonkeyDo (approved):

MonkeyDo:
@brainzbot retest this please

MonkeyDo (approved):
Looks good, thanks !

PR #1214, opened 2020-12-29
LB-530: Refactor the LastFm Modal code

kshitijk83:
# Problem
Currently the function for showing messages in LastFm Importer Modal is present in other components. However this is against React's design principles and makes it difficult to add new code or test the existing code. Refactoring the code so that the content shown by Modal is controlled by the component itself will solve the issue.

For reference this is the code which has to be refactored

https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/LastFMImporterModal.tsx

https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/Importer.tsx

JIRA ticket: https://tickets.metabrainz.org/browse/LB-530

# Solution

Refactored the `LastFmImporter.tsx` code to include the importing functionality which was delegated to `Importer.tsx` file. I renamed `updateMessage` function which was from Importer.tsx to `updateMessageImporter`, kept the same private variables and initialized them in the constructor function.


listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -1,6 +1,13 @@
 import * as ReactDOM from "react-dom";
 import * as React from "react";
 import Importer from "./Importer";
+// Importer.tsx
```
shivam-kapila, line 4: don't think these comments are needed

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -26,8 +33,36 @@ export type ImporterState = {
...
+
+  private numCompleted = 0; // number of pages completed till now
+
+  // Variables used to honor LB's rate limit
+  private rlRemain = -1;
+  private rlReset = -1;
+  private rlOrigin = -1;
+
+  public msg?: React.ReactElement; // Message to be displayed in modal
+  public canClose = true;
```
shivam-kapila, line 65: we should pull out the state as a type just like the props

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -1,6 +1,13 @@
 import * as ReactDOM from "react-dom";
 import * as React from "react";
 import Importer from "./Importer";
```
shivam-kapila, line 3: Is this import needed now?

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -48,10 +91,10 @@ export default class LastFmImporter extends React.Component<
     const { lastfmUsername } = this.state;
     this.toggleModal();
     event.preventDefault();
-    this.importer = new Importer(lastfmUsername, this.props);
+    // this.importer = new Importer(lastfmUsername, this.props);
```
shivam-kapila, line 94: You can remove this if not needed

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -68,6 +111,264 @@ export default class LastFmImporter extends React.Component<
     this.setState({ msg: this.importer.msg });
   };
 
+  // Importer.tsx
```
shivam-kapila, line 114: here too

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -659,7 +659,7 @@ export default class RecentListens extends React.Component<
       <ul className="pager" style={{ display: "flex" }}>
         <li
           className={`previous ${
-            currRecPage && currRecPage <= 1 ? "hidden" : ""
+            currRecPage && currRecPage <= 1 ? "disabled" : ""
```
shivam-kapila, line 662: rebase over master please. I think by mistake you made a branch from the other PR branch

shivam-kapila (changes requested):
Thanks for the contribution Looks good. Maybe @ishaanshah has some other things in mind too. Would like his review. Other than this please remove the duplicated code from `Importer.tsx` and not to forget the tests.

shivam-kapila:
I checked this PR out and saw a lot of linting related issues, especially `Must use destructuring props assignment`. You can observe these issues (and correct some of them automatically) by running `lint.sh`. Please look into this too

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -60,14 +225,149 @@ export default class LastFmImporter extends React.Component<
...
+      this.userToken,
+      "import",
+      payload
+    );
+    this.updateRateLimitParameters(response);
+  }
+
+  async startImport() {
+    this.updateModalAction(<p>Your import from Last.fm is starting!</p>, false);
+    this.playCount = await this.getTotalNumberOfScrobbles();
```
shivam-kapila, line 264: this playCount is used only when it's not an incremental import. Imo we can add a check to only fetch the count from last.fm if it's not an incremental import. Saves an extra API call :)

shivam-kapila (changes requested):
this looks good. I think we can remove the redundant code from Importer..tsx and fix the tests (cut paste IG)

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -60,14 +225,150 @@ export default class LastFmImporter extends React.Component<
...
+                Note: This import will stop at the starting point of your last
+                import. :)
+                <br />
+              </span>
+            )}
+            <span>Please don&apos;t close this page while this is running</span>
+          </span>
+        </p>
+      );
+      this.updateMessage(msg);
```
ishaanshah, line 301: As this function doesn't do anything much now, we can just replace it with `this.setState({ msg })`

ishaanshah (changes requested):
Hey, thanks a lot for opening the PR.
I tested PR, however I found out couple of regressions
- The importer doesn't report an error if the user is invalid.
- The importer also doesn't report an error if there's an network error and the page freezes.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
kshitijk83: Thanks for the review @shivam-kapila. This component need to fetch the api at this point.. otherwise it is showing modal with success at an invalid username too

kshitijk83:
> Hey, thanks a lot for opening the PR.
> I tested PR, however I found out couple of regressions
> 
>     * The importer doesn't report an error if the user is invalid.
> 
>     * The importer also doesn't report an error if there's an network error and the page freezes.
> 
> 
> Also can you add regression tests for the above errors to ensure that it doesn't happen in the future.

Hey @ishaanshah .. regarding the page freeze issue, i noticed it is freezing only when the server is not running, i guess that's bound to happen if i stop the server while clicking submit. Can you explain more about this issue, as in my system it is working fine and is not freezing on the network error given that the server is running.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
shivam-kapila: Hm. Makes sense

ishaanshah (approved):
> Hey @ishaanshah .. regarding the page freeze issue, i noticed it is freezing only when the server is not running.

Hi, thanks for making the changes. You are correct about the freezing issue, I accidentally disabled all networks (including docker) which made the server unreachable and caused the free. On turning of internet access the importer worked as expected. The PR looks good to me, thanks again for opening it. I'll let @MonkeyDo have a look at it and merge it.

shivam-kapila:
@kshitijk83 I think there must be some redundant code left in the original files after you moved these codes. Can you remove that please?

shivam-kapila (changes requested):
@kshitijk83 I think there must be some redundant code left in the original files after you moved the functions. Can you remove that please?

shivam-kapila:
@brainzbot  retest this please

shivam-kapila (approved):
Thanks for your contribution. This looks ready. The failing JS build is failing due to a known other is (fixed in #1228), related to this issue. The changes here are good to go :tada:

MonkeyDo (approved):
I'm not very familiar with the LastFM importer code, but the PR looks fine to me, thanks for opening it ! :)

PR #1215, opened 2020-12-30
Document that Content-Type header is required for submit-listens

ruuda:
# Problem

If the client does not provide a `Content-Type` header in the `/1/submit-listens` request, then the response is

    {
      "code": 400,
      "error": "Cannot parse JSON document: Expected object or value"
    }

even if the body does contain a valid json document. It is not clear from the docs that a `Content-Type` header is needed.

# Solution

Mention the header in the docs.

# Action

It would be nice to re-deploy the docs to help future readers.

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -47,6 +47,7 @@ def submit_listen():
     For complete details on the format of the JSON to be POSTed to this endpoint, see :ref:`json-doc`.
 
     :reqheader Authorization: Token <user token>
+    :reqheader Content-Type: *application/json*
     :statuscode 200: listen(s) accepted.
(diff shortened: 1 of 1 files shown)


alastair:
@brainzbot add to whitelist

alastair:
@ruuda thanks for this! I don't know why we were missing it. 
Almost all of our endpoints require the json content type, is this the only one that was missing it? Could you do a quick review to make sure that we're not missing it anywhere else?

ruuda:
Thanks for having a look! I checked all POST endpoints at https://listenbrainz.readthedocs.io/en/production/dev/api/, there were two more aside from `submit-listens` and both accept a json body, so I also added the header there.

alastair:
Thanks!

PR #1224, opened 2021-01-04
Fix ignoring some files when building Docker images

yvanzo:
# Problem

Patterns in the `.dockerignore` file are incorrectly following `.gitignore` format.
The main difference is that root directory `/` is implicit in `.dockerignore` format.
See https://docs.docker.com/engine/reference/builder/#dockerignore-file for reference.

This leads to not ignoring a bunch of files when building Docker images.
It can make Docker images to be larger than necessary or to contain unintended data.

# Solution

* Do not rely on defaults which are different from `.gitignore` syntax by explicitly prepending with `/` to match from the root directory only (default) or with `**/` to match from any subdirectory (and from the root directory too).
* Append an ending `/` to specify when ignoring directories
* Extend some patterns to match more actual and potential files: `.gitattributes`, `venv-py39`…

That pull request follows commit https://github.com/metabrainz/sir/commit/c03d56865917d9fddec54dd7a432268c04afa5ff whihc addresses the same issue in SIR.

# Action

* Rebuild, check it still works for all build targets, and compare images size.
* Possibly add more patterns to ignore more files. (I just fixed and extended existing patterns, I did not add any new pattern.)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,9 +5,9 @@
 **/*.py[cod]
 
 # Virtual environment
-/env
-/venv
-/build
+/env/
+/venv/
+/build/
 
 # Logs
 **/*.log
@@ -15,11 +15,11 @@
 **/pip-delete-this-directory.txt
 
 # Test results
-**/htmlcov
+**/htmlcov/
 **/.coverage
 
 # Private credentials
 /credentials
 
 # Javascript packages
-/node_modules
+/node_modules/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,25 +1,25 @@
-.git
+/.git
 
 # Byte-compiled / optimized / DLL files
-__pycache__/
-*.py[cod]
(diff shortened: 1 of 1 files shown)


alastair:
@brainzbot add to whitelist

alastair:
So that I understand this better - the main issue is that something like 
* `__pycache__/`

only refers to a directory ./__pycache__ in the root directory, and not in any other subdirectory? For this reason you've explicitly prefixed a `/` to the beginning of rules that refer to files in the root, and added `**/` to rules that should apply in all subdirectories?

yvanzo:
@alastair: Exactly, see also commits message and linked documentation.

It makes a difference mainly when building from a developer’s working local copy.

yvanzo:
Note that `**/` matches `/` too.

alastair (approved):

PR #1227, opened 2021-01-04
LB-765: In development, use a single postgres server for all databases

alastair:
# Problem

Timescale is just a postgres database, so instead of starting up 2 servers we can just start one and put all data there

# Solution

Configure a single database server and create 3 databases.

# Action

Local developers will lose all of their user content, as it was in the old database volumes. You'll have to recreate it


alastair:
still some `depends_on: timescale` in some docker-compose configs to remove

paramsingh (approved):
thanks, this is really cool!

alastair:
rebased and force-pushed to rerun tests.

PR #1228, opened 2021-01-05
LB-781: Fix failure in UserListeningActivity.test.tsx due to new year

ishaanshah:
# Problem
Arrival of new year causes a test in UserListeningActivity.test.tsx to fail as the function uses current date.

# Solution
Mocking the function which returns the current date fixes the issue.

PR #1231, opened 2021-01-12
Mock call to fetch_playlist_recording_metadata when getting playlists

alastair:
# Problem

Due to configuration issues, we currently can't connect to external hosts from tests in jenkins (LB-782).
The playlist get method makes a call to labs.api.listenbrainz.org to get metadata for the playlist given its mbids.
Because this is a unit test, we don't really want the result of this lookup to be determined by an external service anyway, so we should disable it when running tests.

# Solution
mock the external api calls

### DIFF
--- a/listenbrainz/tests/integration/test_playlist_api.py
+++ b/listenbrainz/tests/integration/test_playlist_api.py
@@ -1,3 +1,4 @@
+from unittest import mock
 from uuid import UUID
 
 import dateutil.parser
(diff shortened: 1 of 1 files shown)


mayhem (approved):
OK, makes sense.

alastair:
The only remaining failing integration tests are due to sporadic issues that we're still trying to track down.
JS tests fixed by #1228

PR #1234, opened 2021-01-12
Add eslint with  autofix to Webpack dev process

MonkeyDo:
This PR adds eslint (on a separate process) to Webpack build process, but only in dev mode.
This ensures the code gets a chance to be reviewed by a human prior to committing (as opposed to a precommit git hook for example)

Also committed the resulting two files with autofixable eslint issues and removing the `lint.sh` file which becomes obsolete.

One opinionated choice as well: turning warnings off for the `camelcase` ESLint rule.
Cooperation between python, API structure and JS means we'll have lots of fields not in camelcase, and I don't think this is a real issue we need to worry about.


webpack.config.js
```diff
@@ -16,23 +16,27 @@ module.exports = function (env) {
...
     entry: {
-      main: "/static/js/src/RecentListens.tsx",
-      import: "/static/js/src/LastFMImporter.tsx",
-      userEntityChart: "/static/js/src/stats/UserEntityChart.tsx",
-      userReports: "/static/js/src/stats/UserReports.tsx",
-      userPageHeading: "/static/js/src/UserPageHeading.tsx",
-      userFeed: "/static/js/src/UserFeed.tsx",
-      playlist: "/static/js/src/playlists/Playlist.tsx",
-      playlists: "/static/js/src/playlists/Playlists.tsx",
+      main: "./static/js/src/RecentListens.tsx",
```
paramsingh, line 28: I might be wrong here, but I think that the absolute paths here were because the frontend code dir was being mounted in a docker container (i think, `static_builder`) in /static for the automatic builds on code changes during dev. Would that process be affected by these changes?

docs/dev/devel-env.rst
```diff
@@ -255,15 +255,18 @@ are a helpful way to validate new changes without a lot of work.
 Lint your code
 --------------
 
-ListenBrainz uses ESLint to lint the frontend codebase, please make sure you lint
-all new frontend code using
+ListenBrainz uses ESLint to lint the frontend codebase as part of the development process, in Webpack.
 
-.. code-block:: bash
+This should fix auto-fixable issues and list all other issues in your terminal.
```
paramsingh, line 260:
```suggestion
This should automatically fix trivial auto-fixable issues and list all other issues in your terminal.
```

webpack.config.js
MonkeyDo: Strangely, the code was mounted both to `/code/static` and `/static` in the docker-compose file:
https://github.com/metabrainz/listenbrainz-server/blob/master/docker/docker-compose.yml#L113-L114

I'm just removing the `/static` mount and pointing any path to the equivalent `/code/static` folder.
I'm choosing to keep `/code/static` rather than `/static/` because we already have other necessary files in `/code` (https://github.com/metabrainz/listenbrainz-server/blob/master/docker/Dockerfile.webpack#L6)
It also makes sure the setup it more in line with the CI setup where we copy to `/code/static`: https://github.com/metabrainz/listenbrainz-server/blob/master/docker/Dockerfile.webpack#L13


I can confirm that static_builder does its job properly and recompiles (and reruns eslint) on code changes.

paramsingh: Sounds good, thanks for confirming.

paramsingh (approved):

webpack.config.js
MonkeyDo: Thanks for reviewing! I feel better having had to re-re-confirm I'm not breaking anything :)

PR #1235, opened 2021-01-12
Return json in the error handler in the API

alastair:
# Problem

When an API endpoint raises one of our APIExceptions, it gets serialised as json, but if the error handling comes from flask (e.g. 404 or 500) the regular flask error handler is triggered, which returns html. Any error that happens inside the `/1/...` API path should return json.

# Solution

Return json

On a web page http500 errors look like this:
Screenshot 2021-03-18 at 3 30 33 PM

and on the api:
Screenshot 2021-03-18 at 3 30 40 PM

### DIFF
--- a/listenbrainz/webserver/errors.py
+++ b/listenbrainz/webserver/errors.py
@@ -107,7 +107,6 @@ class CompatError(object):
 
 
 def init_error_handlers(app):
-
     def error_wrapper(template, error, code):
         hide_navbar_user_menu = False
         if code == 500:
@@ -136,7 +135,7 @@ def init_error_handlers(app):
                 A Response which will be a json error if request was made to the LB api and an html page
                 otherwise
         """
-        if current_app.config.get('IS_API_COMPAT_APP'):
+        if current_app.config.get('IS_API_COMPAT_APP') or request.path.startswith(API_PREFIX):
             return jsonify({'code': code, 'error': error.description}), code
         return error_wrapper('errors/{code}.html'.format(code=code), error, code)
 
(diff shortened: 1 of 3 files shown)


alastair:
@brainzbot retest this please

listenbrainz/webserver/errors.py
```diff
@@ -160,20 +160,20 @@ def not_found(error):
     def file_size_too_large(error):
         return handle_error(error, 413)
 
-    @app.errorhandler(500)
+    @app.errorhandler(InternalServerError)
```
amCap1712, line 163: Why do we use `InternalServerError` here and the error codes elsewhere? Like why we don't use `@app.errorhandler(ServiceUnavailable)` instead of ` @app.errorhandler(503)` ?

amCap1712 (approved):

listenbrainz/webserver/errors.py
amCap1712: Is it because the unhandled error can only be an `InternalServerError`?

alastair: You're right, we could make this more consistent.
I _believe_ that the other items use numbers instead of exceptions because that's how it always was. When I originally opened this PR with flask 0.12, you had to use an exception here instead of a numerical code, so I changed it. 
However, looking at the new documentation for flask 1.1 (https://flask.palletsprojects.com/en/1.1.x/errorhandling/#unhandled-exceptions, https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.handle_exception) it seems that you can register it with a number. Let me test it and if it works I'll change it back for consistency

alastair: Yes, in the end flask runs this errorhandler on unhandled exceptions even if we use 500, I changed it back for consistency, and added a comment explaining how this handler is different

PR #1236, opened 2021-01-12
LB-519: Show the timestamp imported in the Last.FM importer

jdaok:
# Problem

JIRA ticket [LB-519](https://tickets.metabrainz.org/browse/LB-519)
The Last.FM import page doesn't show the timestamp of the last imported track, so users can't gauge the progress.

# Solution

I retrieved the first song's listened_to from the payload, which is the latest timestamp imported on that page.
I converted this to an epoch to display it as a string and add it to the message.

image

### DIFF
--- a/listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
+++ b/listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
@@ -312,3 +312,22 @@ describe("LastFmImporter Page", () => {
     expect(wrapper.find('input[type="submit"]').props().disabled).toBe(true);
   });
 });
+
+describe("LastFmImporter Page", () => {
+  it("should properly convert latest imported timestamp to string", () => {
+    // Check getlastImportedString() and formatting
+    let testDate = Number(page["recenttracks"]["track"][0]["date"]["uts"]);
+    let lastImportedDate = new Date(testDate * 1000);
+    let msg = lastImportedDate.toLocaleString("en-US", {
+        month: "short",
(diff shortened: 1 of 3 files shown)


paramsingh:
@brainzbot add to whitelist

paramsingh (approved):
Thanks for doing this! Looks good to me, is there any chance you could add a test for this?

jdaok:
> Thanks for doing this! Looks good to me, is there any chance you could add a test for this?

Could you clarify what you mean by adding a test? I ran ./test.sh and the frontend tests and I see that the github bot added the 4 checks. (sorry this is my first pull request!)

paramsingh:
No worries, what I meant was whether it'd be possible to add a test to https://github.com/metabrainz/listenbrainz-server/blob/6379eb3d/listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx#L1 checking that the date is rendered correctly? just an assertion that the date rendered is in the correct format or something like that?

I'm happy to merge this without the test if it's non-trivial, but I figure we should look into it anyways.

jdaok:
Thanks for the clarification. I added the test and separated my edits into a function, please let me know if it looks good.

shivam-kapila:
Hey, thanks for the contribution. This looks nice. I just had a lil suggestion. Should we make the text as "Latest import timestamp". Last timestamp imported is a lil confusing imo.

MonkeyDo (approved):

MonkeyDo:
@brainzbot retest this please

PR #1237, opened 2021-01-14
Use consul-template 0.18.5 base image and update startup scripts

alastair:
# Problem
This is the first prototype of improving the LB startup scripts to use a newer version of the `metabrainz/consul-template-base` docker image, with the aim of getting LB ready to use the new version of consul when we release it.

The python base images are still based on the 0.16 version of the consul-template-base image, which has a single consul config file. MusicBrainz updated to 0.18.5 years ago, and we left the python apps behind. Now with the plans to upgrade consul and consul-template we need to upgrade to the new syntax.

Additionally, there is some confusion in the naming of some startup scripts, especially the `uwsgi` service which is actually a general wrapper around all apps that we run. This is the LB version of https://github.com/metabrainz/acousticbrainz-server/pull/329

# Solution

We create a separate runit service for each listenbrainz service that we want to run. By default, all services are disabled (presence of a `down` file in the service directory). `rc.local` is executed before runit, this looks at the environment variables and removes the `down` file. Instead of executing consul-template in one process and the actual app in another process, we make consul-template exec the child process. This means that we're guaranteed to always have a config file rendered when services start up, removing the need for our sleep/check loops when loading config files in python.

# Action
Currently this has only been tested with the web image. Some of the changes to docker-compose.yml are only required for testing, the final version of this PR shouldn't include them.

Still a bunch of pending items to do
- [x] Deploy new version of python image using new consul-template, and switch LB to use this
- [x] Decide if we have separate startup scripts for prod and beta, or if we combine them together like we did in AB. Currently these changes have only been made in the `docker/beta` folder.
- [x] See if we can remove duplication in consul-template config files
- [x] We need to render config files when running cron. Need to decide if we execute cron directly from the consul config file, or do something different
- [x] Configure working directory for jobs that use `python -m` or `python manage.py`
- [x] Determine what happens when services disappear from consul and the config file is rewritten. Do the processes restart? Stop?
- [x] we certainly have a problem with the LB dumps -- when in a dump the consul config gets rerendered, the dumps fail. far from optimal.
- [x] Check that signals from consul are propagated to child processes correctly


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM metabrainz/python:3.7-20190302 as listenbrainz-base
+FROM metabrainz/python:newconsul as listenbrainz-base
```
alastair, line 1: I just built this image locally for testing - https://github.com/metabrainz/docker-python/blob/master/3.7/Dockerfile#L1
Change to `docker/beta/timescale_writer/consul-template-timescale-writeFROM metabrainz/consul-template-base:v0.18.5-2` and `docker build -t metabrainz/python:newconsul .`

Dockerfile
```diff
@@ -67,25 +67,63 @@ RUN useradd --create-home --shell /bin/bash --uid 900 --gid 900 lbdumps
...
 RUN touch /etc/service/cron/down
 
-# Consul Template service is already set up with the base image.
-# Just need to copy the configuration.
-COPY ./docker/consul-template.conf /etc/consul-template.conf
+# API Compat (last.fm) server
+COPY ./docker/$deploy_env/api_compat/uwsgi-api-compat.ini /etc/uwsgi/uwsgi-api-compat.ini
+COPY ./docker/$deploy_env/api_compat/consul-template-api-compat.conf /etc/consul-template-api-compat.conf
+COPY ./docker/$deploy_env/api_compat/api_compat.service /etc/service/api_compat/run
+RUN chmod +x /etc/service/api_compat/run
```
alastair, line 88: chmod may not be required if we set the permission on the file in git

docker/beta/api_compat/consul-template-api-compat.conf
```diff
@@ -0,0 +1,12 @@
+template {
+    source = "/code/listenbrainz/consul_config.py.ctmpl"
+    destination = "/code/listenbrainz/listenbrainz/config.py"
+}
+
+exec {
+    command = "uwsgi /etc/uwsgi/uwsgi-api-compat.ini"
+    splay = "60s"
+    reload_signal = "SIGHUP"
```
alastair, line 9: we need to double-check all of these signals to determine that the child process does the right thing. I just copied these from the musicbrainz config file

docker/docker-compose.yml
```diff
@@ -42,11 +42,14 @@ services:
     build:
       context: ..
       dockerfile: Dockerfile
```
alastair, line 44: changes in this file are temporary and only for testing

docker/beta/spark_reader/consul-template-spark-reader.conf
```diff
@@ -0,0 +1,17 @@
+template {
+    source = "/code/listenbrainz/consul_config.py.ctmpl"
+    destination = "/code/listenbrainz/listenbrainz/config.py"
+}
+
+template {
+    source = "/code/listenbrainz/admin/config.sh.ctmpl"
+    destination = "/code/listenbrainz/admin/config.sh"
```
alastair, line 8: I'm not sure which services require config.sh. It seems like at least some of the spark code requires it, as well as the cron jobs. See other note about rendering templates for cron.

paramsingh: I'm pretty sure config.sh is only required by the cron jobs, no spark code runs with consul, so it should be okay.

alastair: I found a reference to it in https://github.com/metabrainz/listenbrainz-server/blob/a464312a452375a79a3693a49011d7b13e71a4a9/docker/start-request-consumer-container.sh#L5, but I don't know what this file is used for

docker/beta/api_compat/consul-template-api-compat.conf
alastair: According to https://uwsgi-docs.readthedocs.io/en/latest/Management.html we either use `SIGINT` to kill uwsgi, or run it as `uwsgi --die-on-term` to have it accept `SIGTERM` as the shutdown command

docker/beta/follow_dispatcher/consul-template-follow-dispatcher.conf
```diff
@@ -0,0 +1,13 @@
+template {
+    source = "/code/listenbrainz/consul_config.py.ctmpl"
+    destination = "/code/listenbrainz/listenbrainz/config.py"
+}
+
+exec {
+    # TODO: Working directory
+    command = "python3 manage.py run_follow_server -h 0.0.0.0 -p 3031"
```
alastair, line 8: To run in a working directory, use cd && python, but consider consul-template documentation about shells and signals: https://github.com/hashicorp/consul-template#exec-mode

It's possible that we're going to need a separate shell script in /usr/local/bin that does cd and then uses exec

alastair:
After a discussion with @mwiencek:

- MBS restarts if consul services change and the template is re-rendered (with SIGHUP). This is mitigated by having multiple mbs instances, and the splay config option. It's possible that we could prevent this happening by configuring consul to send a signal to uwsgi that it ignores, this needs more testing.
- For cron, MBS uses a separate service that renders the config file and doesn't `exec{}` cron. This looks like it should solve the concern of a dump being killed if the config file is regenerated
- we should use a shell script for `python` commands that cd to the right directory, optionally change user, and then exec the process so that signals are sent properly
- consul-template can take multiple `-config` flags, which may allow us to have one config file for the template definitions, and a separate config file for the exec definitions, preventing us from duplicating template definitions. Look into it

alastair:
After discussion with @mayhem:
* There is value in being able to turn off a service in consul and cause the website to restart with no connections to external services (e.g. to be able to restart postgres with no active connections). Support this
* At the same time, we should allow the website to come up even if it's missing config for a service and then show a nice error message which explains the problem, rather than a generic error causing us to have to go and look through server logs to find the issue
* Not worth trying to deduplicate `template{}` blocks in multiple config files
* We want to add a way of tracking when services are restarted. Add a hook in runit before the downstream app is executed to send a message to sentry. Track unclean shutdowns and report to sentry when this happens. Create a separate project in sentry to track these messages over different apps
* @mayhem has feeling that processes sometimes quit for unknown reasons, and it may be related to consul. See if we can look into this.

docker/beta/spark_reader/consul-template-spark-reader.conf
paramsingh: that is a different config.sh, sadly 🙈

PR #1238, opened 2021-01-16
LB-535: Split RecentLists component into smaller components

kshitijk83:
# Problem

The RecentListens component is used on the follow page, /recent and on the profile pages.
As a result, there is a lot of if-else logic based on the page the component is being rendered on.
This makes it hard to work with. We should break off the common stuff into a smaller component and have different components for each of the pages using the common stuff.

ticket link: https://tickets.metabrainz.org/browse/LB-535

# Solution

refactored recent listens page's song list into `RecentListen,tsx` and made `ListenPager.tsx` for pagination code.


paramsingh:
hey @kshitijk83, it might be a good idea for us to chat about this task over IRC once. While the changes here look sane, I think there is probably more work involved here.

kshitijk83:
Sure @paramsingh

MonkeyDo:
Hi @kshitijk83 I hope you don't mind but I'm going to close this PR.
As the files are getting increasingly stale and needing rebasing, it's going to become difficult to keep it up to date.
I also think while separating the pagination could be useful it wasn't the main issue with the component. There's a lot of work needed to break it down but to be honest I'm not even sure where to start 😅

PR #1241, opened 2021-01-18
LB-796: Correct link to playlist exported to Spotify

MonkeyDo:
The alert created after successfully exporting a ListenBrainz playlist shows a link like https://api.spotify.com/v1/playlists/<id>. This is an API link and visiting it without proper auth returns 401.

For the playlist to open in the Spotify player the playlist must be like https://open.spotify.com/playlist/<id> where <id> in both the cases is the same.

### DIFF
--- a/listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
+++ b/listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
@@ -631,13 +631,14 @@ export default class PlaylistPage extends React.Component<
         newPlaylist.id,
         spotifyURIs
       );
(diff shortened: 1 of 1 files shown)


alastair (approved):
thank you!

shivam-kapila:
> The alert created after successfully exporting a ListenBrainz playlist shows a link like https://api.spotify.com/v1/playlists/<id>. This is an API link and visiting it without proper auth returns 401.
> 
> For the playlist to open in the Spotify player the playlist must be like https://open.spotify.com/playlist/<id> where <id> in both the cases is the same.

Looks like copy paste didnt work for monospace text. Lol. Thanks for the fix :)

PR #1242, opened 2021-01-19
[LB-516] Rewrite the APIService retries to be iterative

jdaok:
[Jira Ticket LB-516](https://tickets.metabrainz.org/browse/LB-516)
Rewrite the APIService retries to be iterative + updated tests

# Problem

the retry logic in the APIService file for submit-listens recursively calls the same function if there's a rate limit error or something similar.
This is risky as it can lead to heap overflows in the real world. We should change this logic to be iterative instead.

# Solution

- Removed recursive function calls and added a while loop to make function iterative

- Because the function can't return a response and continue the loop at the same time, the 429 error test was updated to retain it's functionality.


paramsingh (approved):
Looks good to me, would you mind confirming that you tested this manually once by importing on your local environment too?

jdaok:
Yes my import was successful. I didn't have any rate limiting or network errors but I think the tests should cover that.

listenbrainz/webserver/static/js/src/APIService.ts
```diff
@@ -161,31 +161,33 @@ export default class APIService {
...
+              Authorization: `Token ${userToken}`,
+              "Content-Type": "application/json;charset=UTF-8",
+            },
+            body: JSON.stringify(struct),
+          });
+  
+          // we skip listens if we get an error code that's not a rate limit
+          if (response.status === 429) {
+            // Rate limit error, this should never happen, but if it does, try again in 3 seconds.
+            setTimeout(
```
ishaanshah, line 178: This won't cause the JavaScript to sleep for 3 seconds here, it will simply setup a time to call the given callback after 3 seconds and continue. If you want it to sleep you can wait on a `Promise` which gets resolved in 3 seconds like this.
```js
await new Promise((resolve) => setTimeout(resolve, 3000))
```

ishaanshah (changes requested):
Hi thanks for opening the PR, I went through the code and had a couple of comments.

ishaanshah:
There also seem to be some linting issues which are causing the JS CI tests to fail, you can fix them by running `./lint.sh`.

jdaok:
Thanks for the review. I added your suggestions, updated the tests, and the linting issues should be fixed now.

ishaanshah (approved):
Looks good to me, thanks again for opening the PR :)

MonkeyDo (commented):
Thanks for this PR @jdaok !

MonkeyDo (approved):
This change seem to be working as expected and the retry mechanism is working nicely, well done :)
Thanks for improving this part of the codebase!

I did find upon manual testing that the retrying mechanism works as expected, but some other part of the lastfm import mechanism does not recover from network disconnections.
To be clear, the issue is in some other part of the code and is not related to this PR, so I'm approving and merging this now.

If you are interested in figuring out the other issue with the importer, I created this ticket:  https://tickets.metabrainz.org/browse/LB-806

jdaok:
No problem guys :) and I was able to reproduce that issue with the importer as well, I'll get to work on the ticket.

PR #1243, opened 2021-01-19
Remove unused nginx docker configuration

alastair:
# Problem

Some files that aren't used anywhere in the LB docker setup.
Note that these files are distinct from those mentioned in LB-799, I left those files in place.

# Solution

Remove unused files


mayhem (approved):
Time to skip that pesky test... But otherwise fine....

PR #1244, opened 2021-01-19
LB-509: Instructions to run on port other than 80

r0han22:
Add instructions to run on port other than 80

# Problem

   The current version of the docs makes no mention of how to run listenbrainz on a port other than port 80.

# Solution
 Added an additional note with instructions on how to run the server on another port if port 80 is being used.

### DIFF
--- a/docs/dev/devel-env.rst
+++ b/docs/dev/devel-env.rst
@@ -154,6 +154,14 @@ environment by running ``develop.sh up``.
 .. code-block:: bash
 
     ./develop.sh up
+    
+ 
+.. note::
+
+    ``./develop.sh`` up does not work if another application is already rinning on port 80, 
+    because listenbrainz_web_1 wants to bind to port 80. This can be handled by editing 
+    "80:80" to "9000:80" in the web section of ``docker/docker-compose.yml`` and rerunning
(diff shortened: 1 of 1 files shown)


alastair:
@brainzbot add to whitelist

alastair:
Hi @r0han22, thanks for the patch! Please check your text for spelling before you commit it, I see an error in "running" on the first line.

I think that we can simplify your note a bit, I don't think it's necessary to describe `listenbrainz_web_1`. Instead, I would describe it in terms of the service defined in the file. You don't need to explicitly specify a different port, such as 9000. The user can choose their own port. I would write something like:

By default, the `web` service listens on port 80. If you already have a service listening on this port then you can change it by updating the `ports` section of `docker/docker-compose.yml`:

    ports:
      - "80:80"

To change the listening port, change only the value before the `:`.

r0han22:
Hi @alastair , thanks for the feedback. 
Made the required changes to the docs,  should I submit a fresh PR or is the updated commit fine?

alastair (approved):
Thanks for making the update. This looks fine. No need to open a new PR.

MonkeyDo (approved):
Thanks !

PR #1247, opened 2021-01-19
Don't allow playlist owner to be collaborator

MonkeyDo:
This PR fixes two issues related to having  a playlist owner also be in the collaborators:
[LB-787](https://tickets.metabrainz.org/browse/LB-787) and [LB-788](https://tickets.metabrainz.org/browse/LB-788).

We simply remove the playlist owner if it is in collaborators upon playlist creation, edition and duplication.

# Cleanup

We probably want to run an SQL script to clean up existing playlist where creator_id is in collaborator_ids

### DIFF
--- a/listenbrainz/tests/integration/test_playlist_api.py
+++ b/listenbrainz/tests/integration/test_playlist_api.py
@@ -335,7 +335,9 @@ class PlaylistAPITestCase(IntegrationTestCase):
         """ Test to ensure creating a playlist with collaborators works """
 
         playlist = get_test_data()
-        playlist["playlist"]["extension"][PLAYLIST_EXTENSION_URI]["collaborators"] = [self.user2["musicbrainz_id"]]
+        # Owner is in collaborators, should be filtered out
+        playlist["playlist"]["extension"][PLAYLIST_EXTENSION_URI]["collaborators"] = [self.user["musicbrainz_id"],
+                                                                                      self.user2["musicbrainz_id"]]
 
         response = self.client.post(
             url_for("playlist_api_v1.create_playlist"),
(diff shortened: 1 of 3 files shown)


listenbrainz/tests/integration/test_playlist_api.py
```diff
@@ -335,7 +335,9 @@ def test_playlist_create_with_collaborators(self, mock_fetch_playlist_recording_
         """ Test to ensure creating a playlist with collaborators works """
 
         playlist = get_test_data()
-        playlist["playlist"]["extension"][PLAYLIST_EXTENSION_URI]["collaborators"] = [self.user2["musicbrainz_id"]]
+        # Owner is in collaborators, should be filtered out
```
alastair, line 338: is this a TODO, or a description of this test? If it's a description, can you change it to "If the owner is in collaborators, it should be filtered out"

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -286,6 +286,10 @@ def create_playlist():
         get("extension", {}).get(PLAYLIST_EXTENSION_URI, {}).\
         get("collaborators", [])
 
+    # Don't allow creator to also be a collaborator
+    if user["musicbrainz_id"] in collaborators:
```
alastair, line 290: If someone puts the a collaborator name in the list more than once this will only remove one of them. We should also make this list unique before filtering and before passing it to the db method

MonkeyDo: Good catch, thanks

alastair: the python way of doing this is `collaborators = list(set(collaborators))`, instead of going through dict.

mayhem (approved):
Looks good, lets get this out too!

mayhem:
Boo, the integration test looks unhappy. Will skip this PR for now.

listenbrainz/webserver/views/playlist_api.py
MonkeyDo: > the python way of doing this is `collaborators = list(set(collaborators))`, instead of going through dict.

I just realized `list(set(…` doesn't conserve the original list order while `list(dict.fromkeys(…` does.
I don't think that's of much importance considering we always overwrite the collaborators, but just wanted to check if I could be missing some other reason why order might be important…

mayhem: Why don't we make a habit of keeping the list sorted? In which case, we just add a sorted() call on the outside and you're done.

alastair: in that case we should probably sort when pulling out of the database rather than when inserting - we have no ordering on the SELECT statement so it's not guaranteed that this would come out ordered anyway.

MonkeyDo: Created a ticket: https://tickets.metabrainz.org/projects/LB/issues/LB-807

PR #1249, opened 2021-01-20
Only get recommendation feedback for current page of recs

MonkeyDo:
Currently deployed recommendation feedback page on beta doesn't show submitted feedback when a user refreshes the page.
The issue is that we request feedback for all 1000 items, making the URL far too long and the request fails.

This PR changes the mechanism and only fetches feedback for the items in the page the user is on.


vansika (approved):
Don't know why the test fails, otherwise looks good!

MonkeyDo:
I confirm tests are passing despite the status.

PR #1250, opened 2021-01-21
Suppress javascript CI tests errors

MonkeyDo:
Allow each task to complete without throwing an error and breaking the build.
Instead we user report files for after each task


alastair (approved):
👍 nice

docker/jenkins/javascript.sh
```diff
@@ -40,13 +40,13 @@ function run_tests {
     docker ps -a | grep $COMPOSE_PROJECT_NAME | awk '{print $0}'
 
     docker-compose -f $COMPOSE_FILE_LOC -p $COMPOSE_PROJECT_NAME run --name ${TEST_CONTAINER_REF}_test \
-                frontend_tester npm run test:ci
+                frontend_tester npm run test:ci || true
```
MonkeyDo, line 43: @alastair I guess we'll find out when the tests run, but are we 100% sure this is correct for this particular task (npm run test:ci), meaning if the report shows errors Jenkins will pick that up and consider the job failed?

alastair: the main question here is "do you want to run the linter if a test fails?". With your change that you've made, all 3 tasks will always run. We can double-check in jenkins that it correctly marks the job as failed if the test result xml file contains a failure. Then, we can make a decision what an error in the linter means (fail the test, mark it as degraded, ignore it).

paramsingh, line 43: Should we add comments explaining why the `|| true` are needed? It isn't immediately obvious.

paramsingh, line 43: Should we add comments explaining why the `|| true` are needed? It isn't immediately obvious.

MonkeyDo: I think with the use of Jenkins' [warnings-ng-plugin](https://github.com/jenkinsci/warnings-ng-plugin) we'll want to allow each task to run and let it report all warnings and errors in the build details.

That'll be less back-and-forth if for example there is an error in the tests but also a linting error. Better to see both reported the first time.

PR #1251, opened 2021-01-22
LB-802: Last.fm importer user friendliness improvements

jdaok:
# Problem
[LB-802](https://tickets.metabrainz.org/browse/LB-802)

1. https://community.metabrainz.org/t/request-clarify-importer-restart-text/506204
The last.fm importer restart text is vague and should be clarified

2. https://community.metabrainz.org/t/problem-with-import-from-last-fm/454546
When the user's last.fm is private, the error message is unrelated and unhelpful ("The number submitted listens is different from the that Last.fm reports due to an inconsistency in their API, sorry!")

3. https://tickets.metabrainz.org/browse/LB-258
The last.fm username input should remember the username so it doesn't have to be typed repeatedly.

# Solution

1. Updated general restart text
image

2. Added catching for last.fm privacy error and changed error message
image

3. HTML input box should now suggest previous inputted usernames on most browsers
image

- fixed eslint error from my previous commit  (Expected 'this' to be used by class method)


listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -151,6 +155,20 @@ export default class LastFmImporter extends React.Component<
     }
   }
 
+  async getUserPrivacy() {
+    // fetch page from Last.fm
+    const { lastfmUsername } = this.state;
+    const url = `${this.lastfmURL}?method=user.getrecenttracks&user=${lastfmUsername}&api_key=${this.lastfmKey}&format=json`;
+    const response = await fetch(encodeURI(url));
```
MonkeyDo, line 162: Let's add a try-catch block around this await to catch any errors with that fetch call.

What should happen if the fetch fails ? (for example a temporary network issue)
Once that is decided, you can add a test with a simulated failure of the fetch with something like `window.fetch.mockImplementationOnce(() => Promise.reject("Some error"));`

MonkeyDo (changes requested):
Looks good so far, thanks for opening this PR and continuing to improve the importer :)

MonkeyDo:
Hi @jdaok !
Can I help you with anything on this PR? I'd love to see those improvements merged :)
Let me know if you have any question !

jdaok:
Moved to https://github.com/metabrainz/listenbrainz-server/pull/1417

PR #1252, opened 2021-01-22
Typesense index deploy

mayhem:
This PR adds atomic typesense index rotation, stable MBID mapping calculation, stable year lookup calculation, with each being re-created atomically every 24 hours. All services remain fully in operation while indexes are being updated.

Most of the code in this PR is cleanup, except for the typesense index rotation code, which is actually new.


listenbrainz/mbid_mapping/docker/crontab
```diff
@@ -1,2 +1,2 @@
-# Create the recording pairs and mapping each monday and thursday
-0 4 * * 1,5 listenbrainz /usr/local/bin/python /code/mapper/manage.py create-all >> /code/mapper/lb-cron.log 2>&1
+# Create the typesend index once a day
```
alastair, line 1: typesense

mayhem:
The usual heisenbugs hanging about....

listenbrainz/mbid_mapping/mapping/typesense_index.py
```diff
@@ -80,11 +117,15 @@ def build_index():
...
-                    print(i)
+                if i and i % 1000000 == 0:
+                    log("typesense index: Indexed %d rows" % i)
 
             if documents:
-                client.collections[COLLECTION_NAME].documents.import_(documents)
+                client.collections[collection_name].documents.import_(documents)
+
+    log("typesense index: indexing complete. waiting for background tasks to finish.")
+    sys.stdout.flush()
```
alastair, line 130: probably don't need the flush if you're already doing it in `log`

listenbrainz/mbid_mapping/mapping/typesense_index.py
```diff
@@ -80,11 +117,15 @@ def build_index():
...
-                if i and i % 100000 == 0:
-                    print(i)
+                if i and i % 1000000 == 0:
+                    log("typesense index: Indexed %d rows" % i)
 
             if documents:
-                client.collections[COLLECTION_NAME].documents.import_(documents)
+                client.collections[collection_name].documents.import_(documents)
+
+    log("typesense index: indexing complete. waiting for background tasks to finish.")
```
alastair, line 129: just out of interest, what background tasks are happening at this point?

alastair (changes requested):
Looks mostly OK. I think we can remove some duplication in the Dockerfile

listenbrainz/mbid_mapping/mapping/typesense_index.py
mayhem: Dunno, its not specified in the docs. But if I terminate right away, I get nasty sounding error messages.

alastair (approved):
thank you!

PR #1253, opened 2021-01-25
This weeks warning noise removal...

mayhem:
See ticket LB-805 for keeping track of issues related to the muting of these warnings.

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -112,6 +112,9 @@ def _send_listens_to_queue(listen_type, listens):
 def validate_listen(listen, listen_type):
     """Make sure that required keys are present, filled out and not too large."""
 
+    if listen is None:
+        raise APIBadRequest("Listen is empty and cannot be validated.")
+
     if listen_type in (LISTEN_TYPE_SINGLE, LISTEN_TYPE_IMPORT):
         if 'listened_at' not in listen:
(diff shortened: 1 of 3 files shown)


mayhem:
Failing tests are all heisentests. :(

paramsingh (approved):

listenbrainz/webserver/views/api_tools.py
```diff
@@ -112,6 +112,9 @@ def _send_listens_to_queue(listen_type, listens):
 def validate_listen(listen, listen_type):
     """Make sure that required keys are present, filled out and not too large."""
 
+    if listen is None:
```
alastair, line 115: When you made this change last week to at the `if listen is not None` check, was there a specific exception in sentry that you were fixing?
https://github.com/metabrainz/listenbrainz-server/commit/ed250a374694fc770bfefbe0f53b4ff8d23f6287#diff-fc2ff3e6c304f407ce5cdfcf56803a0be49f73b44e9e6d5621fc20c7b214e06bR135

I'm thinking out loud about this a bit, but some things come to mind:
 * If we're checking that it's "emptyish", perhaps we should just check `if not listen` instead, which will also check against `{}`
 * Where is this method called from, and why is it passing empty listens in? Is the correct response to an empty listen to skip it, or to return an error? Even in a bulk import - should 1 bad listen prevent 99 good ones from being imported?

PR #1254, opened 2021-01-25
LB-778: Don't clear add-to-playlist search input on blur

MonkeyDo:
As [LB-778] mentions, currently in the playlist page when a user loses focus of the add-a-track search input, whatever query they had typed gets cleared.
The reason is that we're using the React-Select library with a bit of an edge use case and the default behavior is to allow searching within the options in the dropdown, rather than fetch API results.

The solution is to cache everything in the state and take manual control of the React-Select component.
Cache the search query and prevent it from being cleared, and cache the search results for that query too.

We also prevent the dropdown from closing when a user selects a result, so that they can easily add multiple songs from those same results.


paramsingh:
Is it possible to add tests for this behaviour?

MonkeyDo:
> Is it possible to add tests for this behaviour?

I'll give it a go

MonkeyDo:
@brainzbot retest this please my dear

MonkeyDo:
The Playlist component didn't have any tests. I added a standard "page loads OK" test with snapshot comparison and a test for the search input behavior that this PR changes.
I'll open a ticket about creating more tests.

In the meantime, JS tests are passing (I know it looks like they're not, but that's because we're working on the Jenkins config).
As for Unit and Integration tests randomly failing… ¯\_(ツ)_/¯ 
Let's ask @brainzbot retest this please

MonkeyDo:
@brainzbot retest this please

PR #1256, opened 2021-01-28
Jenkins image cleanup

alastair:
# Problem

When running tests on jenkins, we leave behind images from each build, using up more and more disk space and causing our sysadmin to be sad.

# Solution

Introduce labels to images to be able to filter them.
After running a test, clean up all listenbrainz images that were built before this test run. We do this by filtering based on the label `org.label-schema.name=ListenBrainz`, and use the current `$COMPOSE_PROJECT_NAME` plus known image name (`listenbrainz).
Don't remove the image for this test run, because we want to keep around intermediate image layers. Intermediate layers mean that we don't have to do the apt/pip installs every single test run.

I removed the `docker ps -a --no-trunc | xargs` pipeline because it doesn't seem to be required. In my testing, the `docker-compose down` command in the line above correctly stopped and removed all containers associated with the test run. What's more, the xargs flag `--no-run-if-empty` flag isn't available in macos, causing problems when running these test scripts locally during development/testing. I think it's fine to remove these commands.

# TODO

Double-check that the test script doesn't quit early if the `docker image rm` command fails because there are no matching images

This change has only been made to unit tests for now. We also need to apply it to integration tests, js tests, and spark tests.
In addition to removing images, I noticed that tests also don't remove named volumes. This is the case especially for the spark tests, which create 2 volumes on each run. We should also remove these.

The `docker image rm` step will only untag and remove the final built image. In the case that intermediate layers are no longer used, they will stay around. We need to run a `docker image prune` periodically. I don't know if we should do this in the test run script, or if we should set up a periodic cron job on the server to do this.


docker/jenkins/unit.sh
```diff
@@ -23,10 +23,8 @@ function cleanup {
     docker-compose -f $COMPOSE_FILE_LOC \
                    -p $COMPOSE_PROJECT_NAME \
                    down --remove-orphans
-    docker ps -a --no-trunc  | grep $COMPOSE_PROJECT_NAME \
-        | awk '{print $1}' | xargs --no-run-if-empty docker stop
-    docker ps -a --no-trunc  | grep $COMPOSE_PROJECT_NAME \
-        | awk '{print $1}' | xargs --no-run-if-empty docker rm
+    # Untag LB images that were built before this test run
+    docker image rm $(docker images --filter="before=$COMPOSE_PROJECT_NAME_listenbrainz" --filter "label=org.label-schema.name=ListenBrainz" --format '{{.Repository}}:{{.Tag}}')
```
mayhem, line 27: This seems like a clever approach, but I worry that it is a bit too simplistic. From what I am reading, two successive test runs of the same project will have the same filter key, which could cause it to not remove enough containers. Which is far far better than what we had before and eventually will stabilize itself. But if we were to only ever test one project, I am not sure this would work.

mayhem (commented):
I think this is Good Enough, thoughts?

docker/jenkins/unit.sh
alastair: `COMPOSE_PROJECT_NAME` includes `${BUILD_TAG}`, which is the incrementing build number in jenkins. Every successive test run will have a different argument to `before`.

This means that if we are running job 3, this will select items with label `name=ListenBrainz` (the listenbrainz image) and built before job 3, this would include job 2 (and job 1 if it hasn't yet been deleted).

Note that we're not using this to say "delete every other project in jenkins that was previously built", we're only saying "delete old versions of the LB run".

docker/jenkins/unit.sh
```diff
@@ -23,10 +23,8 @@ function cleanup {
     docker-compose -f $COMPOSE_FILE_LOC \
                    -p $COMPOSE_PROJECT_NAME \
                    down --remove-orphans
-    docker ps -a --no-trunc  | grep $COMPOSE_PROJECT_NAME \
-        | awk '{print $1}' | xargs --no-run-if-empty docker stop
-    docker ps -a --no-trunc  | grep $COMPOSE_PROJECT_NAME \
-        | awk '{print $1}' | xargs --no-run-if-empty docker rm
+    # Untag LB images that were built before this test run
```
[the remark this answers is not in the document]
alastair: yeah, I specifically added "Untag", because I understand that if we have an image with 10 layers, `docker image rm` will only remove the image tag from the most recent layer, and not delete other layers if they're also used by other images. I will double-check this assumption and update the text if needed.
For simplicity's sake, perhaps "remove" is clearer.

docker/jenkins/unit.sh
mayhem: OK, I didn't understand the nuances of this, let's give it a shot.

paramsingh (approved):
Looks good to me!

amCap1712:
As we are discussing to test GH Actions, should we put this on hold for the time being?

alastair:
We decided to add an additional filter to include the exact docker-compose project name. I think we should add that and merge it to get this work out of the way in case we need it in the future.

alastair (commented):
cool, let's merge this and get it into the git history!

PR #1270, opened 2021-02-03
LB-806: Importer connection

jdaok:
Jira ticket [[LB-806]](https://tickets.metabrainz.org/browse/LB-806)

I separated the loop inside startImport to make it easier to test. importLoop will always receive a valid payload from getPage (it's retries are already thoroughly tested). The importer can only fail to recover and stop running if importLoop encounters an unhandled exception, which the new test should take care of.

### DIFF
--- a/listenbrainz/webserver/static/js/src/LastFMImporter.tsx
+++ b/listenbrainz/webserver/static/js/src/LastFMImporter.tsx
@@ -1,6 +1,6 @@
 import * as ReactDOM from "react-dom";
 import * as React from "react";
-import { faSpinner, faCheck } from "@fortawesome/free-solid-svg-icons";
+import { faSpinner, faCheck, faTimes } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { IconProp } from "@fortawesome/fontawesome-svg-core";
 import APIService from "./APIService";
@@ -171,7 +171,9 @@ export default class LastFmImporter extends React.Component<
     const { lastfmUsername } = this.state;
 
     const retry = (reason: string) => {
-      // console.warn(`${reason} while fetching last.fm page=${page}, retrying in 3s`);
+      console.warn(
+        `${reason} while fetching last.fm page=${page}, retrying in 3s`
+      );
       setTimeout(() => this.getPage(page), 3000);
     };
 
@@ -272,6 +274,7 @@ export default class LastFmImporter extends React.Component<
   async importLoop() {
     while (this.page > 0) {
       // Fixing no-await-in-loop will require significant changes to the code, ignoring for now
+      this.lastImportedString = "...";
       const payload = await this.getPage(this.page); // eslint-disable-line
       if (payload) {
         // Submit only if response is valid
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -314,8 +304,21 @@ export default class LastFmImporter extends React.Component<
...
+    this.updateModalAction(<p>Your import from Last.fm is starting!</p>, false);
+    this.latestImportTime = await this.APIService.getLatestImport(
+      this.userName
+    );
+    this.incrementalImport = this.latestImportTime > 0;
+    this.playCount = await this.getTotalNumberOfScrobbles();
+    this.totalPages = await this.getNumberOfPages();
+    this.page = this.totalPages; // Start from the last page so that oldest scrobbles are imported first
+
+    await this.importLoop(); // import pages
```
MonkeyDo, line 319: It would be wise to have a try-catch block around this call to `importLoop` to make sure any error in the promise won't be swallowed (.

In the catch block we have an opportunity to show the user a friendly error message letting them know the import failed (and why—by showing the error on screen).

That also means that the test you wrote ("should not contain any uncaught exceptions") won't be testing uncaught errors thrown by `importLoop` anymore, and we should instead rely on another testing method.
[you can keep the test which will catch other possible uncaught errors]

That can be for example checking instead the final message the user sees, with a test ensuring that in normal circumstances the user sees a success message, and another test to deliberately fail `importLoop` somehow and ensure the error is caught and shown to the user.

Don't hesitate to ask me questions if any part of this is unclear or if you'd like pointers on how to achieve it :)

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -194,14 +194,14 @@ export default class LastFmImporter extends React.Component<
       }
     } catch {
       // Retry if there is a network error
-      retry("Error");
+      retry("Network error");
```
MonkeyDo, line 197: Yes! That's a much much much better error message, nice catch!
Unfortunately, it looks like the error message isn't used at all. Maybe worth having a look at using it, in another PR?

MonkeyDo (changes requested):
Thanks for tackling this issue, it's going to make the importer a lot more robust !

jdaok:
@MonkeyDo Thanks for the clarification on the ticket! I added the two tests to verify that the correct message is being shown depending on importLoop's success/fail. I updated the uncaught exceptions test and I think it should be kept, without it errors like the call to the undefined payload may slip through the testing.

The importLoop should never fail now (getPage will keep retrying on null payload, no unhandled exceptions). If it does here is what the message looks like: 
image

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
jdaok: It should appear in the console as part of a warning now.

MonkeyDo:
Thanks for the changes ! That looks quite a lot better now !

I took the liberty to change a couple of things myself, as I wanted to merge the PR before the LB release we're doing today.
The only changes were: rebase from master (there was a merge conflict) and fix a couple if linting issues.

PR #1271, opened 2021-02-08
Pin package versions and update brainzutils

amCap1712:
Once https://github.com/metabrainz/brainzutils-python/pull/45 is merged, BU will use open ended versions. To make sure we use desired versions, pin the versions in downstream projects.

The BU PR must be merged before this PR and the `requirements.txt` must be updated to use the new release of BU. The same is applicable for the dataset hoster.


requirements.txt
```diff
@@ -21,9 +21,9 @@ yattag == 1.14.0
 xmltodict == 0.12.0
 oauth2client == 4.1.3
 pika == 0.13.0
-git+https://github.com/metabrainz/brainzutils-python.git@v1.14.1
+git+https://github.com/metabrainz/brainzutils-python.git@version-ranges
 spotipy == 2.16.1
-git+https://github.com/metabrainz/data-set-hoster.git@v-2021-01-12.0#egg=datasethoster
+git+https://github.com/amCap1712/data-set-hoster.git#egg=datasethoster
```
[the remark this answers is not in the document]
amCap1712: It was intentional till the dataset hoster was released. Update now.

requirements_spark.txt
```diff
@@ -8,7 +7,18 @@ numpy==1.19.4
...
+werkzeug==1.0.1
+Flask-DebugToolbar==0.11.0
+Flask-UUID==0.2
+raven[flask]==6.10.0
+redis==3.5.3
+msgpack-python==0.5.6
+requests==2.23.0
+SQLAlchemy==1.3.16
+mbdata==25.0.4
+sqlalchemy-dst==1.0.1
```
alastair, line 24: Double-check these values, the spark setup doesn't run a flask webserver, so most of these are probably not needed

requirements_spark.txt
```diff
@@ -8,7 +7,18 @@ numpy==1.19.4
 git+https://github.com/metabrainz/brainzutils-python.git@v1.14.1
```
alastair, line 7: should also be upgraded

requirements.txt
```diff
@@ -21,9 +21,9 @@ yattag == 1.14.0
 xmltodict == 0.12.0
 oauth2client == 4.1.3
 pika == 0.13.0
-git+https://github.com/metabrainz/brainzutils-python.git@v1.14.1
+git+https://github.com/metabrainz/brainzutils-python.git@v1.16.0
```
alastair, line 24: upgrade to 1.16.1, which removes psycopg2-binary from the BU dependencies

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM metabrainz/python:3.7-20190302 as listenbrainz-base
+FROM metabrainz/python:3.7-20201201 as listenbrainz-base
```
alastair, line 1: as we discussed with CB, ensure that we force install pip 21 in this dockerfile (and others) so that we can test the new version resolver

alastair (commented):
This needs to be cleaned up a bit to match the changes that we made with CB

requirements_spark.txt
alastair: spark does use the manage.py cli, which means that we do need both BU, Flask, and all of the related dependencies. This is fine to leave here

alastair (approved):

PR #1272, opened 2021-02-08
Fix Webpack build

MonkeyDo:
https://github.com/metabrainz/listenbrainz-server/commit/f7dec5 introduced a regression with the build setup.
The static folder is oddly mounted twice for the static_builder service in our docker-compose file.
I tried cleaning that up but consequently broke the production build.
@paramsingh tried to warn me, but I didn't listen ! 😨 

I had to make a couple more adjustments with the Webpack config to work with different code paths between dev and build setup.

While in there I took the opportunity to reduce image build time by removing an unused postinstall script being run in frontend_builder image build.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -98,9 +98,9 @@ WORKDIR /static
 # Compile static files
 RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
     apt-get install -y nodejs && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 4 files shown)


MonkeyDo:
I re-ran the Jenkins unit tests (https://ci.metabrainz.org/job/listenbrainz-unit/316/) with a "success" state, but github checks didn't like that and currently show as failed, can safely be ignored.
In any case this PR only touches JS setup, and Jenkins Javascript tests passed fined.

PR #1275, opened 2021-02-10
Clean up: Correction of websockets port in config template

MonkeyDo:
Use the correct default port for Websockets server after recent changes in PR #1212
Updates test snapshots accordingly.

### DIFF
--- a/listenbrainz/config.py.sample
+++ b/listenbrainz/config.py.sample
@@ -110,7 +110,7 @@ UPLOAD_FOLDER = "/tmp/lastfm-backup-upload"
 API_URL = 'https://api.listenbrainz.org'
 LASTFM_PROXY_URL = 'http://localhost:8080/'
 SERVER_ROOT_URL = 'http://localhost'
-WEBSOCKETS_SERVER_URL = 'http://localhost:8081'
+WEBSOCKETS_SERVER_URL = 'http://localhost:8082'
 MUSICBRAINZ_OAUTH_URL = 'https://musicbrainz.org/oauth2/userinfo'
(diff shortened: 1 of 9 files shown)


amCap1712:
How about changing the follow dispatcher container names to websockets in production as well  to avoid future confusion?

MonkeyDo:
@brainzbot retest this please

MonkeyDo:
> How about changing the follow dispatcher container names to websockets in production as well to avoid future confusion?

That's a good idea, but will require more work, and on another repo. I'll open a ticket.

amCap1712:
Great, thanks!
To clarify I was referring I was to making changes here, https://github.com/metabrainz/listenbrainz-server/blob/53e2d91f6eaf4b7ae7e88077a0cc45d588ab0364/docker/beta/uwsgi/uwsgi.service#L25
https://github.com/metabrainz/listenbrainz-server/blob/53e2d91f6eaf4b7ae7e88077a0cc45d588ab0364/docker/prod/uwsgi/uwsgi.service#L19
https://github.com/metabrainz/listenbrainz-server/blob/53e2d91f6eaf4b7ae7e88077a0cc45d588ab0364/docker/test/uwsgi/uwsgi.service#L25 . But I do not know how beta/test/prod works so that might entail changes elsewhere as well.

MonkeyDo:
Yes, exactly :)
it would definitely be cleaner to change the names of the services, but that does require changing the services themselves and to be honest I'm afraid to touch it :)
I'll get someone to hold my hand and do it, one of these days.

PR #1276, opened 2021-02-10
Allow playlist collaborators to access private playlist

MonkeyDo:
We forgot to implement access to playlist for that playlist's collaborators.

There are a couple of different types of access for collaborators:
- use API endpoints to add/move/delete tracks in the playlist
- use API endpoint to copy the playlist
- view the web page for a playlist

# Left to do
We don't allow collaborators to edit a playlist details (title, private/public, collaborators, etc.), and currently have no provisions to allow a collaborator to remove themselves as collaborator from a playlist. Instead the owner needs to remove them.
A possible avenue for spam?

### DIFF
--- a/listenbrainz/db/model/playlist.py
+++ b/listenbrainz/db/model/playlist.py
@@ -79,7 +79,7 @@ class Playlist(BaseModel):
     recordings: List[PlaylistRecording]
     # mbid of the playlist referred to in copied_from_id
     copied_from_mbid: Optional[uuid.UUID]
-    
+
     def is_visible_by(self, user_id: int):
         if self.public:
             return True
--- a/listenbrainz/db/model/playlist.py
+++ b/listenbrainz/db/model/playlist.py
@@ -97,9 +97,9 @@ class Playlist(BaseModel):
 
     def is_modifiable_by(self, user_id: int):
         """Check if user can modify a playlist
-        
+
         Check if a user is allowed to add/move/delete items in a playlist.
-        
+
         Args:
             user_id : row id of the user.
         """
--- a/listenbrainz/db/model/playlist.py
+++ b/listenbrainz/db/model/playlist.py
@@ -80,7 +80,12 @@ class Playlist(BaseModel):
     # mbid of the playlist referred to in copied_from_id
     copied_from_mbid: Optional[uuid.UUID]
 
-    def is_visible_by(self, user_id: int):
+    def is_visible_by(self, user_id: Optional[int]):
+         """Check if user is allowed to access a playlist
+        
+        Args:
+            user_id : (Optional) row id of the user.
+        """
         if self.public:
             return True
         if user_id:
@@ -88,8 +93,6 @@ class Playlist(BaseModel):
                 return True
             elif user_id in self.collaborator_ids:
                 return True
-            else:
-                return False
         return False
 
 
--- a/listenbrainz/db/model/playlist.py
+++ b/listenbrainz/db/model/playlist.py
@@ -94,6 +94,18 @@ class Playlist(BaseModel):
             elif user_id in self.collaborator_ids:
                 return True
         return False
+        
+    def is_modifiable_by(self, user_id: int):
+        """Check if user can modify a playlist
+        
+        Check if a user is allowed to add/move/delete items in a playlist.
+        
+        Args:
+            user_id : row id of the user.
+        """
+        if user_id == self.creator_id or user_id in self.collaborator_ids:
+            return True
+        return False
 
 
 class WritablePlaylist(Playlist):
--- a/listenbrainz/db/model/playlist.py
+++ b/listenbrainz/db/model/playlist.py
@@ -83,6 +83,8 @@ class Playlist(BaseModel):
     def is_visible_by(self, user_id: Optional[int]):
         """Check if user is allowed to access a playlist
 
+        user_id may be None, for example if the user is not logged in.
+
         Args:
             user_id : (Optional) row id of the user.
         """
@@ -99,6 +101,7 @@ class Playlist(BaseModel):
         """Check if user can modify a playlist
 
         Check if a user is allowed to add/move/delete items in a playlist.
+        user_id is required, since playlist modifications require a logged in user
 
         Args:
             user_id : row id of the user.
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/playlist.py
```diff
@@ -19,9 +19,9 @@ def load_playlist(playlist_mbid: str):
         raise BadRequest("Provided playlist ID is invalid: %s" % playlist_mbid)
 
     playlist = db_playlist.get_by_mbid(playlist_mbid, True)
-    # TODO: Allow playlist collaborators to access private playlist
-    if playlist is None or not playlist.public and (not current_user.is_authenticated or playlist.creator_id != current_user.id):
-        raise NotFound("Cannot find playlist: %s" % playlist_mbid)
+    if playlist is None or not playlist.public and not current_user.is_authenticated \
```
alastair, line 22: because we actually have an object here, let's take advantage of it in order to ensure we don't mess this up in multiples 

in `class Playlist`:

```py
def is_visible_by(user):
    if self.public:
        return True
    if user.is_authenticated:
        if is owner
        elif is collaborator
        else: return False
```

However, I fear that this may be more difficult than expected, because it seems that `current_user` in the website is an object, and `user` in the API, based on the api token is a dictionary. This should be normalised somehow too, not in this fix

alastair (changes requested):

alastair:
We decided that a collaborator shouldn't be able to edit the metadata, just tracks.

I think that it's important that a person can remove themselves as a collaborator from a playlist.

listenbrainz/webserver/views/playlist.py
MonkeyDo: Perhaps we can use the user ID instead since we have a way to get that both from the web and API views?

alastair:
@brainzbot retest this please

alastair:
@brainzbot retest this please

listenbrainz/webserver/views/playlist.py
alastair: yes, using the user id is fine, but we'll have to work out where to put the anonymous user check. Off the top of my head I don't know what `validate_auth_header` returns in the playlist get api method if you don't authenticate

mayhem (changes requested):
Overall this looks good -- I suppose the "allow a user to remove themselves from a playlist" can be done in a separate PR, but I think it is important to update the tests to reflect the changes in this PR. Thanks for implementing this!

alastair (changes requested):

listenbrainz/db/model/playlist.py
```diff
@@ -80,6 +80,18 @@ class Playlist(BaseModel):
...
+            return True
+        if user_id:
+            if user_id == self.creator_id:
+                return True
+            elif user_id in self.collaborator_ids:
+                return True
+            else:
+                return False
+        return False
+
```
shivam-kapila, line 94:
```suggestion
    def is_visible_to(self, user_id: int):
        if self.public:
            return True
        if user_id and (user_id == self.creator_id or user_id in self.collaborator_ids):
                return True
        return False

```

shivam-kapila: Being a grammar nazi here
Plus a thought for user id test, can it be zero?

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -444,10 +444,12 @@ def get_playlist(playlist_mbid):
...
 
-    if not playlist.public:
-        user = validate_auth_header()
-        if playlist.creator_id != user["id"]:
-            raise APINotFound("Cannot find playlist: %s" % playlist_mbid)
+    user = validate_auth_header(True)
+    user_id = None
+    if user:
+        user_id = user["id"]
+    if not playlist.is_visible_by(user_id):
```
shivam-kapila, line 451:
```suggestion
    user = validate_auth_header(True)
    if not playlist.is_visible_by(user["id"]):
```

shivam-kapila (changes requested):
A few simplifications :)

listenbrainz/db/model/playlist.py
alastair: One thing that I like about the original version is that it's very explicit about what it's doing - 1: check if it's public, 2: check who the creator is, 3: check who the collaborators are. I don't think this change brings any value that makes it significantly better than what was originally there

listenbrainz/webserver/views/playlist_api.py
alastair: `validate_auth_header` returns None if the argument `optional=True` is set and no token is provided, so this will break.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -444,10 +444,12 @@ def get_playlist(playlist_mbid):
...
         raise APINotFound("Cannot find playlist: %s" % playlist_mbid)
 
-    if not playlist.public:
-        user = validate_auth_header()
-        if playlist.creator_id != user["id"]:
-            raise APINotFound("Cannot find playlist: %s" % playlist_mbid)
+    user = validate_auth_header(True)
+    user_id = None
+    if user:
+        user_id = user["id"]
```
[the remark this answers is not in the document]
MonkeyDo: The user is optional here but required in the other endpoints, and I wanted to ensure we call `is_visible_by` with an existing user id, but I suppose it can be optional (as you suggested in another review comment).
Will change it accordingly.

listenbrainz/db/model/playlist.py
MonkeyDo: I agree, I think legibility comes first if there is no significant drawback.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -371,8 +371,8 @@ def edit_playlist(playlist_mbid):
         log_raise_400("Provided playlist ID is invalid.")
 
     playlist = db_playlist.get_by_mbid(playlist_mbid, False)
-    if playlist is None or (not playlist.public and playlist.creator_id != user["id"]):
-        raise APINotFound("Cannot find playlist: %s" % playlist_mbid)
+    if playlist is None or not playlist.is_visible_by(user["id"]):
+        raise NotFound("Cannot find playlist: %s" % playlist_mbid)
```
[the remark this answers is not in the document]
MonkeyDo: Woops 🤦

mayhem:
@brainzbot Ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn

mayhem:
@brainzbot please be so kind and remove your head from your anus and test this again. thanks!

paramsingh:
@brainzbot retest this please

mayhem (approved):
OK, if tests finally pass, lets get this moving!

PR #1278, opened 2021-02-11
Sort playlist collaborators

MonkeyDo:
We are getting random [test failures](https://ci.metabrainz.org/job/listenbrainz-integration/274/console) because we don't sort our lists of playlist collaborators:

```
        self.assertEqual(response.json["playlist"]["extension"]
                         [PLAYLIST_EXTENSION_URI]["collaborators"], [self.user2["musicbrainz_id"],
>                                                                    self.user3["musicbrainz_id"]])
E       AssertionError: Lists differ: ['troi-bot', 'anothertestuserpleaseignore'] != ['anothertestuserpleaseignore', 'troi-bot']
E       
E       First differing element 0:
E       'troi-bot'
E       'anothertestuserpleaseignore'
E       
E       - ['troi-bot', 'anothertestuserpleaseignore']
E       + ['anothertestuserpleaseignore', 'troi-bot']
```

I tried using a getter/setter on the `Playlist` model so as to implement this only once on the model, but that doesn't work because pydantic doesn't serialize `@property` fields: https://github.com/samuelcolvin/pydantic/issues/935
We get errors: `'Playlist' object has no attribute 'collaborators'`

Couldn't find a good way to make it work so I went for sorting everywhere we set collaborators, but I'm open to suggestions.

### DIFF
--- a/listenbrainz/db/playlist.py
+++ b/listenbrainz/db/playlist.py
@@ -67,6 +67,7 @@ def get_by_mbid(playlist_id: str, load_recordings: bool = True) -> Optional[mode
             user = db_user.get(user_id)
             if user:
                 collaborators.append(user["musicbrainz_id"])
+            collaborators.sort()
             obj['collaborators'] = collaborators
         return model_playlist.Playlist.parse_obj(obj)
 
@@ -180,6 +181,7 @@ def _playlist_resultset_to_model(connection, result, load_recordings):
                 user = db_user.get(user_id)
                 if user:
                     collaborators.append(user["musicbrainz_id"])
+            collaborators.sort()
             p.collaborators = collaborators
 
     return playlists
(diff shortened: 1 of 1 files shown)


listenbrainz/db/playlist.py
```diff
@@ -67,6 +67,7 @@ def get_by_mbid(playlist_id: str, load_recordings: bool = True) -> Optional[mode
             user = db_user.get(user_id)
             if user:
                 collaborators.append(user["musicbrainz_id"])
+            collaborators.sort()
```
mayhem, line 70: I think these two lines should be outside the for loop. And given that this code is duplicated 3 times, we should make a helper function and call that 3 times.

mayhem (changes requested):

listenbrainz/db/playlist.py
MonkeyDo: 4 times actually !

mayhem (approved):
🐔 🇨🇱 🐿️

MonkeyDo:
@brainzbot retest this please

PR #1279, opened 2021-02-11
Dump fixes

mayhem:
Periodically we would have empty dump files be synced to the FTP server causing all sorts of unpleasant problems.

To address these problems, I've:

* Add permanent logs files in a volume that allows us to review what dumps are doing, beyond the superficial "a dump was created" emails.
* Moved cron jobs around to less likely conflict with each other. This includes using lock files to prevent two dump instances from running at the same time.
* Updated the create dump scripts to clean up the TMP_DIR if an error happens during dumping. This act of not cleaning up left dead dumps kicking around and on the next successful dump those files would be synced to the FTP site causing the havoc.
* Crontab now has MAILTO="" in order to prevent one more log error message about not MTA being installed.


Dockerfile
```diff
@@ -66,6 +66,7 @@ RUN useradd --create-home --shell /bin/bash --uid 900 --gid 900 lbdumps
 
 RUN groupadd --gid 901 listenbrainz_stats_cron
 RUN useradd --create-home --shell /bin/bash --uid 901 --gid 901 listenbrainz_stats_cron
+RUN mkdir /logs && chown lbdumps:lbdumps /logs
```
alastair, line 69: This will need a matching `--volume` flag in docker-server-configs. is that on your list?

admin/create-dumps.sh
```diff
@@ -63,11 +63,13 @@ trap finish EXIT
 if [ "$DUMP_TYPE" == "full" ]; then
     if ! /usr/local/bin/python manage.py dump create_full -l "$TMPDIR" -t $DUMP_THREADS --last-dump-id; then
 	echo "Full dump failed, exiting!"
+        finish
```
alastair, line 66: the `trap finish EXIT` above will cause `finish` to call any time the script exits, even when it's finished with `exit 1`. These additions aren't needed

docker/cron/dump-crontab
```diff
@@ -1,10 +1,11 @@
+MAILTO=""
 
-## Trigger a full dump on 1st and 15th of every month
-00 20 1,15 * * lbdumps /code/listenbrainz/admin/create-dumps.sh full
-## Around 24 hours later, trigger a full import into the spark cluster
-00 21 2,16 * * lbdumps /usr/local/bin/python /code/listenbrainz/manage.py spark request_import_full
+## Trigger a full dump on 1st and 15th of every month, in the middle of the night, do not block to wait for lock.
+0 0 1,15 * * lbdumps flock -x -n /var/lock/lb-dumps.lock /code/listenbrainz/admin/create-dumps.sh full >> /logs/dumps.log 2>&1
```
alastair, line 4: just to confirm that you're aware that this will only lock within the currently running container.
The issue that you found previously when cron was running in a separate container won't be fixed by this lock, as the lock files aren't visible by all containers

mayhem: Yes, that is not the problem I am trying to solve here -- these locks are for preventing cron jobs in the same container from stepping on each other.

Dockerfile
mayhem: yep, I have a branch for that as well.

admin/create-dumps.sh
mayhem: Then how is it possible that this script leaves a dirty dump around when it fails? That is the whole point of this exercise -- how can a failed dump be let after this script terminates. do you see any way?

Dockerfile
mayhem: In fact that PR is merged already: https://github.com/metabrainz/docker-server-configs/blob/master/scripts/services.sh#L1520

mayhem:
Replaced by PR 1282. Closing.

PR #1280, opened 2021-02-11
Rename the follow server to websockets

mayhem:


MonkeyDo (approved):
Provided the service name matches the consul config (and I don't doubt that'll be the case), 🚀

PR #1281, opened 2021-02-11
Safely get track metadata in React components

MonkeyDo:
While using the follow page on LB, the page broke trying to access a property of track_metadata while track_metadata did not exist on that listen (maybe a listening_now type of listen?).

In any case this is a fragile property access and I found a few in the code that we want to fix.

### DIFF
--- a/listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
+++ b/listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
@@ -163,13 +163,13 @@ export default class ListenCard extends React.Component<
           <MediaQuery minWidth={768}>
             <div className="col-xs-9">
               <div className="track-details">
-                <p title={listen.track_metadata.track_name}>
+                <p title={listen.track_metadata?.track_name}>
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -163,13 +163,13 @@ export default class ListenCard extends React.Component<
           <MediaQuery minWidth={768}>
             <div className="col-xs-9">
               <div className="track-details">
-                <p title={listen.track_metadata.track_name}>
+                <p title={listen.track_metadata?.track_name}>
                   {getTrackLink(listen)}
```
alastair, line 167: does the call to `_.get` in getTrackLink also deal with this situation correctly?

MonkeyDo: That is correct.

alastair (approved):
not my area of expertise, but looks good

PR #1282, opened 2021-02-12
Improved dump logging and dump hardening (lol), try II

mayhem:
Periodically we would have empty dump files be synced to the FTP server causing all sorts of unpleasant problems.

To address these problems, I've:

* Add permanent logs files in a volume that allows us to review what dumps are doing, beyond the superficial "a dump was created" emails.
* Moved cron jobs around to less likely conflict with each other. This includes using lock files to prevent two dump instances from running at the same time.
* Updated the create dump script to check to see if it is running inside the production cron container, otherwise exit.
* Crontab now has MAILTO="" in order to prevent one more log error message about not MTA being installed.


docker/cron/stats-crontab
```diff
@@ -57,9 +59,3 @@
 
 # user weekly daily_activity
 10 03 * * * listenbrainz_stats_cron /usr/local/bin/python /code/listenbrainz/manage.py spark request_user_stats --type=daily_activity --range=all_time
-
```
mayhem, line 60: This process is no longer running -- it keeps importing the same old data, so stop it for now.

docker/cron/dump-crontab
```diff
@@ -1,10 +1,11 @@
+MAILTO=""
```
paramsingh, line 1: Could we make this `listenbrainz-observability@metabrainz.org`?

paramsingh (approved):

docker/cron/dump-crontab
mayhem: We could. But the data it would give us would be nearly useless since all the output is re-directed to the log file. Doing so would be an effective way to monitor that cron is running and not much else.

alastair: BTW, I ran into something similar in AB cron last week, where I got this warning after running a cron job which failed:

> Feb 21 16:52:03 2288dacacfda CRON[64]: (CRON) info (No MTA installed, discarding output)

I'm not sure if there's an MTA installed in the LB base image at the moment, so I think that redirecting output and disabling mail is the right thing to do right at this moment

admin/config.sh.ctmpl
```diff
@@ -30,3 +30,8 @@ RSYNC_FULLEXPORT_DIR="$FTP_DIR/fullexport"
 RSYNC_INCREMENTAL_DIR="$FTP_DIR/incremental"
 RSYNC_FULLEXPORT_KEY='/home/{{template "KEY" "user"}}/.ssh/rsync-listenbrainz-dumps-full'
 RSYNC_INCREMENTAL_KEY='/home/{{template "KEY" "user"}}/.ssh/rsync-listenbrainz-dumps-incremental'
+
+# Check to make sure that this container is the prod cron container, otherwise never rsync anything!
+PROD="{{ (env "DEPLOY_ENV") }}"
+CONTAINER_NAME="{{ (env "CONTAINER_NAME") }}"
+DO_NOT_RSYNC=1
```
alastair, line 37: where is this variable used?

admin/create-dumps.sh
```diff
@@ -32,6 +32,13 @@ function add_rsync_include_rule {
     echo "$SHA256_FILE_RULE" >> "$RULE_FILE"
 }
 
+if [[ "${CONTAINER_NAME}" = "listenbrainz-cron-prod" && "${PROD}" = "beta" ]]
```
alastair, line 35: what does the `"${PROD}" = "beta"` check do here?


when we talked about this, I was considering installing separate configs in consul - e.g. instead of `docker-server-configs/LB/create-dumps-config.json` have `docker-server-configs/LB/create-dumps-config.prod.json` and `.beta.json`. This would let us ensure that we never load prod settings in test or beta, even if we accidentally start the cron container in those environments.


One issue with this check is that it makes it difficult to run this script either locally for testing, or in the one-off case where we might want to run it in test. It involves specifically setting some environment variables only in this situation. Whereas having separate constants in consul/config.sh would allow us to just enable a flag to allow us to run the dump

alastair (changes requested):
some questions about the variables/checks that you've added here

admin/config.sh.ctmpl
mayhem: No longer used. removed.

admin/create-dumps.sh
mayhem: Whops, it should be "prod" -- leftover from testing.

I agree that we need to have better testing for this -- let's add a ticket for that, but continue on with this for the time being.

alastair:
I think I pushed to the wrong branch and screwed up the commits here. Will try and unwind everything later this afternoon

admin/create-dumps.sh
alastair: https://tickets.metabrainz.org/browse/LB-819

alastair:
I cherry-picked @mayhem's changes into https://github.com/metabrainz/listenbrainz-server/pull/1237 and they applied cleanly. next time we'll try not to break things with random rebases.

PR #1283, opened 2021-02-12
create-dumps.sh: Various minor fixes and improvements

atj:
* Enable shell errexit option so that unhandled failures result in
  script termination.
* Consolidate exit traps into a single on_exit function.
* Log how long the script took to run.
* Improve argument and variable validation.
* Quote all arguments to commands to prevent globbing and word
  splitting.
* Use "-n" to check if variables have a value instead of "! -z"
* Use rsync to copy files to the backup and FTP directories to prevent
  truncated files on failure or concurrent runs of the script.

---

The output from shellcheck is as follows:

```
In create-dumps.sh line 60:
source "admin/config.sh"
       ^---------------^ SC1091: Not following: admin/config.sh was not specified as input (see shellcheck -x).


In create-dumps.sh line 61:
source "admin/functions.sh"
       ^------------------^ SC1091: Not following: admin/functions.sh was not specified as input (see shellcheck -x).


In create-dumps.sh line 115:
mkdir -m "$BACKUP_DIR_MODE" -p \
      ^-- SC2174: When used with -p, -m only applies to the deepest directory.


In create-dumps.sh line 136:
mkdir -m "$FTP_DIR_MODE" -p "$FTP_CURRENT_DUMP_DIR"
      ^-- SC2174: When used with -p, -m only applies to the deepest directory.

For more information:
  https://www.shellcheck.net/wiki/SC2174 -- When used with -p, -m only applie...
  https://www.shellcheck.net/wiki/SC1091 -- Not following: admin/config.sh wa...
```

The mkdir warnings are not relevant as far as I can tell because the parent directory is specified as a separate argument in both cases.


alastair:
@brainzbot add to whitelist

atj:
I realised that the retry() function wasn't compatible with "set -e" so I've added another commit to fix it. Shellcheck return no errors or warnings.

alastair:
hi @atj - sorry for the drama, can you re-open this PR against master?

PR #1284, opened 2021-02-13
Fix websockets data format to avoid breaking listens page

amCap1712:
Currently, the data received from the listenstores is directly sent to clients through websocket messages. This works for `playing_now` listens because their format matches that expected by the frontend. However, this is not the case for the `single` or `import` listens. The `timescale_writer` modifies the format of the inserted listens. In timescale database, we use the `data` column but elsewhere we refer to it as `track_metadata`. This modified data is received by the websockets server and is then sent to the client. The client code is unable to parse this data and causing rendering issues.

The fix is simple to rename the `data` key to `track_metadata` of each listen received from `timescale_listenstore`.

### DIFF
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -235,9 +235,15 @@ export default class RecentListens extends React.Component<
 
   receiveNewListen = (newListen: string): void => {
     const json = JSON.parse(newListen);
+    // the websocket message received may not contain the expected track_metadata and listened_at fields
+    // therefore, we look for their alias as well.
     if (!("track_metadata" in json)) {
       json.track_metadata = json.data;
       delete json.data;
+    }
+    if (!("listened_at" in json)) {
+      json.listened_at = json.timestamp;
+      delete json.timestamp;
     }
     const listen = json as Listen;
     this.setState((prevState) => {
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -239,25 +239,29 @@ export default class RecentListens extends React.Component<
       json = JSON.parse(newListen);
       // the websocket message received may not contain the expected track_metadata and listened_at fields
       // therefore, we look for their alias as well.
-      if (!("track_metadata" in json) && "data" in json) {
-        json.track_metadata = json.data;
-        delete json.data;
-      } else {
-        // eslint-disable-next-line no-console
-        console.debug(
-          `Could not find track_metadata and data in following json: ${json}`
-        );
-        return;
+      if (!("track_metadata" in json)) {
+        if ("data" in json) {
+          json.track_metadata = json.data;
+          delete json.data;
+        } else {
+          // eslint-disable-next-line no-console
+          console.debug(
+            `Could not find track_metadata and data in following json: ${json}`
+          );
+          return;
+        }
       }
-      if (!("listened_at" in json) && "timestamp" in json) {
-        json.listened_at = json.timestamp;
-        delete json.timestamp;
-      } else {
-        // eslint-disable-next-line no-console
-        console.debug(
-          `Could not find listened_at and timestamp in following json: ${json}`
-        );
-        return;
+      if (!("listened_at" in json)) {
+        if ("timestamp" in json) {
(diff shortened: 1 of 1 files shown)


listenbrainz/websockets/follow_server_dispatcher.py
```diff
@@ -22,9 +23,12 @@ def send_listens(self, listens, listen_type):
...
             self.socketio.emit(event_name, json.dumps(listen), room=listen['user_name'])
 
-
     def callback_listen(self, channel, method, properties, body):
-        listens = json.loads(body)
+        ts_listens = json.loads(body)
+        for ts_listen in ts_listens:
+            ts_listen["track_metadata"] = ts_listen.pop("data")
+
+        listens = [Listen.from_json(ts_listen).to_api() for ts_listen in ts_listens]
```
shivam-kapila, line 31:
```suggestion
        for ts_listen in ts_listens:
            ts_listen["track_metadata"] = ts_listen.pop("data")

        ts_listen = Listen.from_json(ts_listen).to_api()
```

shivam-kapila (changes requested):
I dont think we need two loops. Rest looks good. Plus can we add a test?

amCap1712:
Instead of making the changes to the listen format server-side, we let the client it handle it. The rationale here is to avoid possible expensive serialization / deserialization costs.

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -234,7 +234,18 @@ export default class RecentListens extends React.Component<
   };
 
   receiveNewListen = (newListen: string): void => {
-    const listen = JSON.parse(newListen) as Listen;
+    const json = JSON.parse(newListen);
+    // the websocket message received may not contain the expected track_metadata and listened_at fields
+    // therefore, we look for their alias as well.
+    if (!("track_metadata" in json)) {
```
MonkeyDo, line 240: I this this could be made a bit more reliable by
1. checking if data is present in the object :`if (!("track_metadata" in json) &&  "data" in json)`. Same goes for `listened_at/timestamp`.
2. add an `else` statement to handle cases where there is neither `track_metadata` nor `data`. We probably want to write a helpful message to `console.debug()` and return
3. Wrap the whole JSON parsing and keys checking in a try-catch block and log any caught error to `console.error()` and return

Just some future proofing to make sure we never get the page breaking for failing at some websocket message processing :)

MonkeyDo (changes requested):
Looking pretty good :)

I've noted down some suggestions to make this process more resilient, considering we're here fixing similar errors along with #1281

listenbrainz/webserver/static/js/src/RecentListens.tsx
amCap1712: Agreed :). I have updated the PR with the changes.

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -234,7 +234,36 @@ export default class RecentListens extends React.Component<
...
+      // the websocket message received may not contain the expected track_metadata and listened_at fields
+      // therefore, we look for their alias as well.
+      if (!("track_metadata" in json) && "data" in json) {
+        json.track_metadata = json.data;
+        delete json.data;
+      } else {
+        // eslint-disable-next-line no-console
+        console.debug(
+          `Could not find track_metadata and data in following json: ${json}`
+        );
```
MonkeyDo, line 249: Needs a `return` statement to make sure we don't proceed

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -234,7 +234,36 @@ export default class RecentListens extends React.Component<
...
+      }
+      if (!("listened_at" in json) && "timestamp" in json) {
+        json.listened_at = json.timestamp;
+        delete json.timestamp;
+      } else {
+        // eslint-disable-next-line no-console
+        console.debug(
+          `Could not find listened_at and timestamp in following json: ${json}`
+        );
```
MonkeyDo, line 258: Same as above, needs a return statement

MonkeyDo (changes requested):
Final things, apart form that it's good to go

MonkeyDo (approved):
GTG !

amCap1712:
Thanks for the reviews!

PR #1285, opened 2021-02-14
use recommendation feedback in BP player

vansika:
# Problem
The BP player uses RecentLsitens component, due to which we get `51 years ago` in the listened_at field. 

# Solution
Use recommendations feedback component.

### DIFF
--- a/listenbrainz/webserver/views/player.py
+++ b/listenbrainz/webserver/views/player.py
@@ -43,17 +43,21 @@ def load():
             error_msg="'Listens' array must have one or more items."
         )
 
-    user_data = {
+    current_user_data = {
         "id": current_user.id,
         "name": current_user.musicbrainz_id,
         "auth_token": current_user.auth_token,
     }
     spotify_data = spotify.get_user_dict(current_user.id)
     props = {
-        "user": user_data,
+        "user": {
+            "id": current_user.id,
+            "name": current_user.musicbrainz_id,
+        },
+        "current_user": current_user_data,
         "spotify": spotify_data,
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/player.py
```diff
@@ -43,17 +43,21 @@ def load():
...
         "name": current_user.musicbrainz_id,
         "auth_token": current_user.auth_token,
     }
     spotify_data = spotify.get_user_dict(current_user.id)
     props = {
-        "user": user_data,
+        "user": {
+            "id": current_user.id,
+            "name": current_user.musicbrainz_id,
+        },
```
MonkeyDo, line 56: This feels like a bit of a hack (the Recommendations component is supposed to show recs *for a user*, whereas the player isn't), but I don't see an actual issue anywhere and there's no easy way to make feedback work for the user on this page without refactoring the Recommendations component.

I think it would be good to add a comment here in player.py explaining why we are manually making user equal to current_user, so that the code doesn't get copied in another page or"fixed" without warning.

MonkeyDo (changes requested):
This looks like it's working fine, and fixing the two main problems with the player page (timestamp issues on the ListenCard and lack of feedback feature).

Good to ship after a small comment:

listenbrainz/webserver/views/player.py
vansika: hmmm, yeah, the recommendation component expects `user` and `current user` as keys for its working, it is indeed a hack, adding comment makes sense, thanks for pointing out :)

PR #1289, opened 2021-02-17
Remove flask and brainzutils dependency from listenbrainz_spark

amCap1712:
Flask and in turn brainzutils was being used in listenbrainz_spark only for the purpose of loading configuration values and logging. We instead use the logging module. Further, we add PysparkIntegration and PysparkWorkerIntegraton to the sentry setup in listenbrainz_spark. This may come handy in the production setup.

### TODO before merging
1) #1271 should be merged prior to this
2) The PysparkWorkerIntegration needs to be tested to make sure that at the least it does not break production setup.

### DIFF
--- a/listenbrainz_spark/config.py.sample
+++ b/listenbrainz_spark/config.py.sample
@@ -19,10 +19,10 @@ SPARK_RESULT_QUEUE = "spark_result"
 # calculate stats on X months data
 STATS_CALCULATION_WINDOW = 1
 
-LOG_SENTRY = {
-    'dsn':'',
-    'environment': 'development',
-}
+#LOG_SENTRY = {
+#    'dsn':'',
+#    'environment': 'development',
+#}
 SENTRY_DSN_PUBLIC = ''
 
 # Model id is made up of two parts.
--- a/listenbrainz_spark/ftp/__init__.py
+++ b/listenbrainz_spark/ftp/__init__.py
@@ -1,11 +1,10 @@
 import os
 import ftplib
 import hashlib
+import logging
 
 from listenbrainz_spark import config
 from listenbrainz_spark.exceptions import DumpInvalidException
-
-from flask import current_app
 
 
 class ListenBrainzFTPDownloader:
@@ -20,7 +19,7 @@ class ListenBrainzFTPDownloader:
             self.connection = ftplib.FTP(config.FTP_SERVER_URI)
             self.connection.login()
         except ftplib.error_perm:
-            current_app.logger.critical("Couldn't connect to FTP Server, try again...")
+            logging.critical("Couldn't connect to FTP Server, try again...")
             raise SystemExit
 
     def list_dir(self, path=None, verbose=False):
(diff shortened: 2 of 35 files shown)


mayhem (commented):
Ok, looks pretty good to me. There is one spark test that needs fixing. Once you're done with that, coordinate with me about doing a test deployment to ensure that everything is cool before we merge this.

listenbrainz_spark/config.py.sample
```diff
@@ -19,10 +19,10 @@ SPARK_RESULT_QUEUE = "spark_result"
 # calculate stats on X months data
 STATS_CALCULATION_WINDOW = 1
 
-LOG_SENTRY = {
-    'dsn':'',
-    'environment': 'development',
-}
+#LOG_SENTRY = {
+#    'dsn':'',
+#    'environment': 'development',
```
alastair, line 24: how do we set config.py in production? In that case, we need to make sure that we set environment properly there

listenbrainz_spark/ftp/__init__.py
```diff
@@ -20,7 +19,7 @@ def connect(self):
             self.connection = ftplib.FTP(config.FTP_SERVER_URI)
             self.connection.login()
         except ftplib.error_perm:
-            current_app.logger.critical("Couldn't connect to FTP Server, try again...")
+            logging.critical("Couldn't connect to FTP Server, try again...")
```
alastair, line 22: if we are going to use python logging outside of flask then we should make some effort to configure it.
It'd be good to be able to set a level in config.py. Additionally, we should have a logging setup method that allows us to configure the handlers that we want (and the ability to add a file handler if needed). We could put this in a separate py file (or listenbrainz_spark/init) and use it in all places that we do logging

alastair (changes requested):

Dockerfile
```diff
@@ -31,7 +31,7 @@ WORKDIR /code
 
 RUN mkdir /code/listenbrainz
 WORKDIR /code/listenbrainz
-RUN pip3 install pip==20.2.4
+RUN pip3 install pip==21.0.1
```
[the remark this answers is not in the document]
alastair: oh, I see that it includes the commits for that PR too, no problem!

listenbrainz_spark/ftp/__init__.py
amCap1712: That's a great idea! I'll get to it.

listenbrainz_spark/config.py.sample
```diff
@@ -19,10 +19,10 @@ SPARK_RESULT_QUEUE = "spark_result"
...
+#}
 SENTRY_DSN_PUBLIC = ''
```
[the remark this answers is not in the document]
amCap1712: SENTRY_DSN_PUBLIC is being set in production. I checked found that the LOG_SENTRY.dsn value contains the secret part and the SENTRY_DSN_PUBLIC does not. However, the secret part is optional and deprecated in newer versions of sentry. This is explained in the [sentry docs](https://docs.sentry.io/product/sentry-basics/dsn-explainer/#what-the-dsn-does). We can remove the SENTRY_PUBLIC_DSN variable.

During testing this and other projects with `bu-sentrysdk-upgrade`,  we used the DSN without the secret part. So, it should be safe to remove secret part from the production configurations as well.

listenbrainz_spark/config.py.sample
amCap1712: We do set the environment in production to `production`.

alastair:
@amCap1712 can you rebase this?

PR #1290, opened 2021-02-18
Use BU ratelimiting module

amCap1712:
LB is currently using its own ratelimiter module. BU provides a ratelimiting module that is used in other Brainz projects. Here, we migrate to use the BU module.


mayhem (approved):
This PR makes me happy since it cleans up bits and bobs that have needed attention for some time. Thank you!

alastair:
let's merge this at the same time that we upgrade to a version of BU with https://github.com/metabrainz/brainzutils-python/pull/55. This means that we'll have to re-add the limits to redis because the keys will change. We should add instructions to read the current values and to set the new ones

PR #1291, opened 2021-02-22
Replace RABBITMQ_SERVICE environment variable with Consul key-value pair

yvanzo:
# Problem

Setting ListenBrainz to use a different RabbitMQ service for upgrade or maintenance operations currently require to recreate ListenBrainz containers, which is not convenient.

# Solution

Define RabbitMQ service name using key-value pair in consul template, which configuration can be updated without recreating ListenBrainz containers.

# Action

`rabbitmq_service` key must be set in Consul config prior to deploying this patch.


mayhem (approved):
Much nicer!

PR #1292, opened 2021-02-24
Add mbid mapping to labs api

mayhem:
This PR adds the new MBID lookup endpoint that will be used for the MessyBrainz mapping and for troi as well as the year lookup endpoint that is specifically geared towards year lookups in troi. The year endpoint is not very useful unless you have exact metadata as it is in MusicBrainz.

Also tucked into this PR is yet another template1 heisentest error placebo, but it works, so...


listenbrainz/db/timescale.py
```diff
@@ -42,16 +42,25 @@ def run_sql_script_without_transaction(sql_file_path):
...
-                # the exception sqlalchemy.exc.ProgrammingError "can't execute an empty query"
-                if line and not line.startswith("--"):
-                    connection.execute(line)
-        except sqlalchemy.exc.ProgrammingError as e:
-            print("Error: {}".format(e))
-            return False
-        finally:
-            connection.connection.set_isolation_level(1)
-            connection.close()
+        while True:
```
mayhem, line 45: I've added this check here too since the random failure was occurring here as well. I realize that adding these checks causes them to never trip, effectively making them placebo checks, but I don't mind placebos. Right?

listenbrainz/labs_api/labs/api/mbid_mapping.py
```diff
@@ -0,0 +1,230 @@
...
+            attempt to detune it and try again for detuned artist and detuned recording.
+            If the hit is good enough, return it, otherwise return None.
+        """
+        ac_hit = hit['document']['artist_credit_name']
+        r_hit = hit['document']['recording_name']
+
+        ac_detuned = self.detune_query_string(ac_hit)
+        r_detuned = self.detune_query_string(r_hit)
+
+        is_alt = False
```
amCap1712, line 129: `is_alt` is set here and in the loop but does not seem to be used anywhere.

listenbrainz/labs_api/labs/api/mbid_mapping.py
```diff
@@ -0,0 +1,230 @@
...
+        recording_name_p = prepare_query(recording_name)
+
+        ac_detuned = prepare_query(
+            self.detune_query_string(artist_credit_name_p))
+        r_detuned = prepare_query(self.detune_query_string(recording_name_p))
+
+        tries = 0
+        while True:
+
+            tries += 1
```
amCap1712, line 197: `tries` is updated but not being used anywhere.

listenbrainz/db/timescale.py
amCap1712, line 45:
```suggestion
        retries = 0
        while True:
```
We use `retries` in `except` clause but do not set it which can cause `NameError`.

listenbrainz/db/timescale.py
```diff
@@ -42,16 +42,25 @@ def run_sql_script_without_transaction(sql_file_path):
...
+                break
+            except sqlalchemy.exc.ProgrammingError as e:
+                print("Error: {}".format(e))
+                return False
+            except sqlalchemy.exc.OperationalError:
+                print("Trapped template1 access error, FFS! Sleeping, trying again.")
+                retries += 1
+                if retries == 5:
+                    raise
+                sleep(1)
```
amCap1712, line 61: Need to use as `time.sleep` or modify the import.

amCap1712 (commented):
Overall, looks good! I have left a few comments. I'll also try it out locally later and add other suggestions if any.

listenbrainz/labs_api/labs/api/mbid_mapping.py
mayhem: This is for something I had before and no longer used, but then when thinking about this, I will bring it back in more detail and another PR. :)

listenbrainz/db/timescale.py
mayhem: oops. bad copypasta.

listenbrainz/db/timescale.py
mayhem: fixed.

listenbrainz/labs_api/labs/tests/test_year_from_artist_credit_recording.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class MainTestCase(flask_testing.TestCase):
+
+    def create_app(self):
+        app = create_app()
+        app.config['MB_DATABASE_URI'] = 'yermom'
+        return app
+
+    def setUp(self):
```
amCap1712, line 44: I think you can get rid of the `setUp` and `tearDown` methods because they only call the super class methods.

amCap1712 (approved):
Looks good!

PR #1293, opened 2021-02-24
Add user facing recording search

mayhem:
Carry out a recording search suitable for end-users to enter a mix of artist and recording name in hopes of finding the right track. Since this endpoint is end-user facing, it supports only one query per call.

This function is currently residing in bono-data-sets and is relied upon by the production server, so this should be merged ASAP. Took only 2 months to get this done! 🤦


listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
```diff
@@ -228,7 +228,7 @@ export default class PlaylistPage extends React.Component<
   searchForTrack = async (inputValue: string): Promise<OptionType[]> => {
     try {
       const response = await fetch(
-        "https://datasets.listenbrainz.org/acrm-search/json",
+        "https://labs.api.listenbrainz.org/recording-search/json",
```
MonkeyDo, line 231: For pages that use the regular LB API, we send the API root url to the client in the props:
https://github.com/metabrainz/listenbrainz-server/blob/98d6c11fc8ee60106ca4f90c143fc67ba694e064/listenbrainz/webserver/views/user.py#L145
and we use in on the front-end like this:
https://github.com/metabrainz/listenbrainz-server/blob/98d6c11fc8ee60106ca4f90c143fc67ba694e064/listenbrainz/webserver/static/js/src/RecentListens.tsx#L899


I think it would be better to do the same here and send the config value for `LISTENBRAINZ_LABS_API_URL` to the front-end and use it here instead of hardcoding the `https://labs.api.listenbrainz.org` part.

Happy to help with the front-end part if you need it.

MonkeyDo (commented):
Future-proofing suggestion

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
mayhem: Oh, good suggestion, yes please for help!

MonkeyDo (approved):

amCap1712 (approved):
Looks good!

PR #1294, opened 2021-02-24
LB-821: Fix the position of recording_msid in WS messages

shivam-kapila:
# Problem

The listens received via web sockets contain the `recording_msid` as a top-level key rather than in `listen.track_metadata.additional_info`. This borks features like listen to feedback as they can't find the `recording_msid` in `additional_info`

# Solution

Moving the `recoding_msid` down into `additional_info` should fix this.
Also added a test to see that the keys in the WS message are replaced/moved as desired.

### DIFF
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -263,6 +263,23 @@ export default class RecentListens extends React.Component<
           return;
         }
       }
+      // The websocket message received contains the recording_msod as a top level key.
+      // Therefore, we need to shift it json.track_metadata.additional_info.
+      if (!("recording_msid" in json.track_metadata.additional_info)) {
+        if ("recording_msid" in json) {
+          json.track_metadata.additional_info.recording_msid =
+            json.recording_msid;
+          delete json.recording_msid;
+        } else {
+          // eslint-disable-next-line no-console
+          console.debug(
+            `Could not find recording_msid in following json: ${json}`
+          );
+          return;
+        }
+      }
+      // the websocket message received contain some keys which are are either duplicates or are not required in the frontend.
+      // Ideally this should be handled server-side and this will probably be fixed with protobuf move.
     } catch (error) {
       // eslint-disable-next-line no-console
       console.error(error);
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -263,7 +263,7 @@ export default class RecentListens extends React.Component<
           return;
         }
       }
-      // The websocket message received contains the recording_msod as a top level key.
+      // The websocket message received contains the recording_msid as a top level key.
       // Therefore, we need to shift it json.track_metadata.additional_info.
       if (!("recording_msid" in json.track_metadata.additional_info)) {
         if ("recording_msid" in json) {
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -240,7 +240,7 @@ export default class RecentListens extends React.Component<
       json = JSON.parse(newListen);
     } catch (error) {
       // eslint-disable-next-line no-console
-      console.error("Coudn't parse the new listen as JSON: ", json);
+      console.error("Coudn't parse the new listen as JSON: ", error);
       return;
     }
     const listen = formatWSMessageToListen(json);
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -235,7 +235,15 @@ export default class RecentListens extends React.Component<
   };
 
   receiveNewListen = (newListen: string): void => {
-    const listen = formatWSMessageToListen(JSON.parse(newListen));
+    let json;
+    try {
+      json = JSON.parse(newListen);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error("Coudn't parse the new listen as JSON: ", json);
+      return;
+    }
+    const listen = formatWSMessageToListen(json);
 
     if (listen) {
       this.setState((prevState) => {
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -14,6 +14,7 @@ import FollowUsers from "./FollowUsers";
 import APIService from "./APIService";
 import Loader from "./components/Loader";
 import ListenCard from "./listens/ListenCard";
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -263,6 +263,23 @@ export default class RecentListens extends React.Component<
           return;
         }
       }
+      // The websocket message received contains the recording_msid as a top level key.
+      // Therefore, we need to shift it json.track_metadata.additional_info.
+      if (!("recording_msid" in json.track_metadata.additional_info)) {
```
MonkeyDo, line 268: This can throw error is a variety of cases if the structure isn't exactly `json.track_metadata.additional_info`:
- cannot access property of undefined if track_metadata isn't there
- cannot use 'in' operator if right-side is undefined


Instead, you can use a safe method like [lodash's `has`](https://lodash.com/docs/#has):
`if(!has(json, "track_metadata.additional_info.recording_msid")){…`

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -263,6 +263,23 @@ export default class RecentListens extends React.Component<
...
+        if ("recording_msid" in json) {
+          json.track_metadata.additional_info.recording_msid =
+            json.recording_msid;
+          delete json.recording_msid;
+        } else {
+          // eslint-disable-next-line no-console
+          console.debug(
+            `Could not find recording_msid in following json: ${json}`
```
MonkeyDo, line 276: This will result in `Could not find recording_msid in following json: [object Object]` in the console.
You can pass the json object as a second argument for it to be printed with correct formatting:

```suggestion
            "Could not find recording_msid in following json:", json
```

MonkeyDo (changes requested):
Looks pretty good!

In JS, deep property access in an object that you suspect is not formatted as expected = danger of throwing errors!
Some suggestions below for safe access methods.

I would also like to see this whole formatting bit separated into its own method (something like `formatWSMessageToListen`) and put in `listenbrainz/webserver/static/js/src/utils.tsx` so that it's:
- abstracted from the rest of the page
- reusable if we need it on other pages
- easier to remove if the formatting gets fixed on the server-side
- can be unit tested (here also it will be easier to remove after a server-side fix)

listenbrainz/webserver/static/js/src/RecentListens.tsx
shivam-kapila: done

listenbrainz/webserver/static/js/src/RecentListens.tsx
shivam-kapila: done

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -102,4 +102,67 @@ const getPlayButton = (listen: any, onClickFunction: () => void) => {
...
+          json
+        );
+        return null;
+      }
+    }
+    // The websocket message received contains the recording_msid as a top level key.
+    // Therefore, we need to shift it json.track_metadata.additional_info.
+    if (!_.has(json, "track_metadata.additional_info.recording_msid")) {
+      if ("recording_msid" in json) {
+        json.track_metadata.additional_info.recording_msid =
```
MonkeyDo, line 140: This line also has the potential to break. I thought I commented on it with the last review but I must have lost my comment.

Javascript won't like it if the json object doesn't have the structure we expect.
Say for example there's no `additional_info`,  this line will throw an error trying to set a chid property of `undefined`.

However like above there's an elegant solution with [lodash's merge method](http://lodash.com/docs/#merge):

```
merge(json, {
  track_metadata: { additional_info: { recording_msid: json.recording_msid } },
});
```
That should ensure the structure.

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -102,4 +102,67 @@ const getPlayButton = (listen: any, onClickFunction: () => void) => {
...
+        console.debug(
+          "Could not find recording_msid in following json: ",
+          json
+        );
+        return null;
+      }
+    }
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error(error);
```
MonkeyDo, line 154: I don't think we should allow the function to return `json` if there is an error caught here, because it probably means that our json has a structure sufficiently broken that our attempts to tiptoe around it with `has` and `merge` have failed.
There's something seriously wrong with that format…

If that's the case, the rest of the UI is probably not going to like that poisoned listen either and could potentially throw errors too.

What I suggest is adding that same `return null;` here as you did in the other error cases.

MonkeyDo (changes requested):
Thanks, that's a lot cleaner now !
Some last bits to ensure nothing breaks in the future:

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -234,60 +235,28 @@ export default class RecentListens extends React.Component<
...
-        if ("timestamp" in json) {
-          json.listened_at = json.timestamp;
-          delete json.timestamp;
-        } else {
-          // eslint-disable-next-line no-console
-          console.debug(
-            `Could not find listened_at and timestamp in following json: ${json}`
-          );
-          return;
+    const listen = formatWSMessageToListen(JSON.parse(newListen));
```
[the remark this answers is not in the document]
shivam-kapila: understood

listenbrainz/webserver/static/js/src/utils.tsx
shivam-kapila: This came to my mind but I missed it.Thanks

listenbrainz/webserver/static/js/src/utils.tsx
shivam-kapila: Whoops. Thanks for catching this.

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -235,7 +235,15 @@ export default class RecentListens extends React.Component<
   };
 
   receiveNewListen = (newListen: string): void => {
-    const listen = formatWSMessageToListen(JSON.parse(newListen));
+    let json;
+    try {
+      json = JSON.parse(newListen);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error("Coudn't parse the new listen as JSON: ", json);
```
MonkeyDo, line 243: `json` is always going to be undefined in this context.

More useful would be to print the `error`  instead

MonkeyDo (commented):
One last one :)

listenbrainz/webserver/static/js/src/RecentListens.tsx
shivam-kapila: silly mistake :/

MonkeyDo (approved):

PR #1295, opened 2021-02-25
Check if user has revoked authorization if refreshing token fails

amCap1712:
Sentry is filled with Spotify errors relating to failure if refreshing API tokens. I tried to reproduce the problem locally. It seems that one of the possible causes is that the user revoked LB's authorization to Spotify through Spotify UI but didn't unlink their account from LB. This [official post](https://developer.spotify.com/community/news/2016/07/25/app-ready-token-revoke/) explains how to handle this case. Unfortunately, `spotipy` eats up the error response body in case of any unsuccessful request.

We already fetch the access token manually because of another `spotipy` bug. It makes to modify it to handle refreshing tokens as well.

## TODO:
1) Test by linking a user and revoking authorization from Spotify UI.
2) Log response body for unhandled 400 errors from Spotify.
3) Add a unit/integration test case.

### DIFF
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -84,7 +84,7 @@ class Spotify:
         return "<Spotify(user:%s): %s>" % (self.user_id, self.musicbrainz_id)
 
 
-def refresh_user_token(spotify_user):
+def refresh_user_token(spotify_user: Spotify):
     """ Refreshes the user token for the given spotify user.
 
     Args:
@@ -214,7 +214,7 @@ def update_latest_listened_at(user_id, timestamp):
     db_spotify.update_latest_listened_at(user_id, timestamp)
 
 
-def get_access_token(code):
+def get_access_token(code: str):
     """ Get a valid Spotify Access token given the code.
 
     Returns:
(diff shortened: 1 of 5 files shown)


listenbrainz/domain/spotify.py
```diff
@@ -211,23 +228,28 @@ def get_access_token(code):
     is a bug in the spotipy code which leads to loss of the scope received from the
     Spotify API.
     """
+    r = _get_spotify_token("authorization_code", code)
+    if r.status_code != 200:
+        raise SpotifyListenBrainzError(r.reason)
+    return r.json()
+
+
+def _get_spotify_token(grant_type, token):
```
paramsingh, line 237: Would like docstrings and type annotations on this function.

paramsingh (changes requested):
Looks reasonable to me, I had a few comments about docstrings and type annotations. We should also add tests for these changes.

listenbrainz/domain/tests/test_spotify.py
```diff
@@ -51,6 +52,26 @@ def test_refresh_user_token(self, mock_update_token, mock_get_spotify_oauth, moc
         )
         mock_get_user.assert_called_with(self.spotify_user.user_id)
 
+    @requests_mock.Mocker()
+    def test_refresh_user_token_bad(self, mock_requests):
```
paramsingh, line 56: both this and the other test have the same name, which means that only one of them is being run :(

listenbrainz/webserver/views/profile.py
```diff
@@ -311,11 +311,15 @@ def refresh_spotify_token():
     spotify_user = spotify.get_user(current_user.id)
     if not spotify_user:
         raise APINotFound("User has not authenticated to Spotify")
+
     if spotify_user.token_expired:
         try:
             spotify_user = spotify.refresh_user_token(spotify_user)
         except spotify.SpotifyAPIError:
             raise APIServiceUnavailable("Cannot refresh Spotify token right now")
+        else:
```
paramsingh, line 320: Can we add tests for this branch of code as well?

listenbrainz/domain/tests/test_spotify.py
amCap1712: :facepalm: . Fixed now.

listenbrainz/webserver/views/profile.py
amCap1712: Added.

paramsingh (approved):

PR #1296, opened 2021-03-01
Remove old follow feature

amCap1712:
The old follow feature never really gained traction. The user feed timeline and the playlists features supersede and replace it. Hence, remove the feature.


mayhem (approved):
The python changes look fine, and I've glanced at the JS ones, but I think @MonkeyDo should have a look too.

shivam-kapila (changes requested):
I had tested the changes and everything was working perfectly. Apart from this, there's some clean up left in [lIsten card component](https://github.com/metabrainz/listenbrainz-server/blob/d2755856e9f80529eaad5088034a8f3ff44840a2/listenbrainz/webserver/static/js/src/listens/ListenCard.tsx#L158)

MonkeyDo (changes requested):
Good riddance :)

Overall good job, I barely see any traces left of it. Just these small three:

This type definition: https://github.com/metabrainz/listenbrainz-server/blob/870f7a7c6473a5ebfbda780fea65e9788ce479a7/listenbrainz/webserver/static/js/src/types.d.ts#L195-L197

And as Shivam mentioned, there are a couple of occurrences in the [ListenCard.tsx](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/listens/ListenCard.tsx) component, as well as in the associated test file and snapshot.

Finally, I see a few `FOLLOW_LIST_QUEUE` in the config sample and consul_config template, as well as its usage in `listenbrainz/websockets/listens_dispatcher.py`. Should those be there? If so they should probably be renamed?

amCap1712:
I'll remove the remaining js parts.

I also think that the `FOLLOW_LIST_QUEUE` should be renamed. While doing that should I change the actual name (`follow_list`) of the queue as well because I am not sure if changing the queue name will require other changes in production as well?

MonkeyDo:
> I'll remove the remaining js parts.
> 
> I also think that the `FOLLOW_LIST_QUEUE` should be renamed. While doing that should I change the actual name (`follow_list`) of the queue as well because I am not sure if changing the queue name will require other changes in production as well?

I'll be honest, I have no idea what that queue is or does.
@mayhem might have a more useful opinion?

mayhem:
The FOLLOW_LIST_QUEUE is the name of the rabbitmq queue that distributes the events. It should now be called something like WEBSOCKETS_QUEUE or some-such. However, the coordination of renaming a queue requires a bit of finesse when releasing. I would prefer to do this in a separate step PR.

paramsingh:
We'll need a schema change to delete the follow_list table from production as well.

MonkeyDo (approved):

MonkeyDo (approved):
Good to go ! :rocket:

PR #1309, opened 2021-03-02
Add new table for user recommendation events

paramsingh:
# Problem

We want users to be able to recommend tracks to their followers.

# Solution

This PR adds a table that would be used to save those recommendation events.

# Action

We can discuss whether we want to do things this way or if we want a more generic table to store feed events.

### DIFF
--- a/admin/sql/updates/2021-03-02-add-user-recommendation-event-table.sql
+++ b/admin/sql/updates/2021-03-02-add-user-recommendation-event-table.sql
@@ -1,20 +1,20 @@
 BEGIN;
 
-CREATE TYPE user_recommendation_event_type_enum AS ENUM('recording');
+CREATE TYPE user_timeline_event_type_enum AS ENUM('recording');
 
-CREATE TABLE recommendation.user_recommendation_event (
+CREATE TABLE user_timeline_event (
   id                    SERIAL, -- PK
   user_id               INTEGER, -- FK to "user"
-  recommendation_type   user_recommendation_event_type_enum,
+  event_type            user_timeline_event_type_enum,
   metadata              JSONB,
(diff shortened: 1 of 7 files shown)


admin/sql/updates/2021-03-02-add-user-recommendation-event-table.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+CREATE TYPE user_recommendation_event_type_enum AS ENUM('recording');
+
+CREATE TABLE recommendation.user_recommendation_event (
+  id                    SERIAL, -- PK
+  user_id               INTEGER, -- FK to "user"
+  recommendation_type   user_recommendation_event_type_enum,
+  metadata              JSONB,
```
vansika, line 9: What do you plan to store in here?

mayhem (commented):
I think this is pretty good. I like that the data in JSONB and and that there is an enum type defining the event -- that covers all I need. Two things I think we should change/add:

1. I'm not a fan of the name user_recommendation_event since there will be events that won't specifically be about recommendations, but other "events". How about user_timeline_event?
2. This table needs indexes -- we need to carefully consider the index type that we would create on the recommendation_type field since its an enum with few distinct values.

admin/sql/updates/2021-03-02-add-user-recommendation-event-table.sql
paramsingh: I haven't defined a spec yet, but for the track recommendation feature it'd store track metadata (artist name, track name, spotify ID, MBIDs) etc.

mayhem (approved):
As per our IRC convo, the spec for what goes into the metadata field will be provided in the next PR, so I'm ok to merge this.

PR #1310, opened 2021-03-02
User similarity, approach 2

mayhem:
WIP, this PR is rehashed version of Param's work, but with a focus on having the processing be done by Spark.


mayhem:
OK, done for now, will open when closer to finishing time.

PR #1314, opened 2021-03-04
LB-531: Adjust timestamp format for older listens on listens page

jdaok:
Jira ticket [LB-529:](https://tickets.metabrainz.org/browse/LB-529)

# Problem

The timestamps on the recent listens page can be vague when viewing older listens.

# Solution

preciseTimestamp will determine how old the listened_at date is and format it accordingly on the ListenCard:
image (<24 hours)
image (<1 year)
image (>1 year)


listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -46,6 +46,42 @@ export default class ListenCard extends React.Component<
...
+    }
+    if (
+      // one year to yesterday : show without year
+      msDifference / (1000 * 3600 * 24 * 1) >
+      1
+    ) {
+      return `${listenDate.toLocaleString("en-US", {
+        day: "2-digit",
+        month: "short",
+        // year: "numeric",
```
MonkeyDo, line 75: Better to remove the comment or explicitly set year to `undefined` for clarity.

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -46,6 +46,42 @@ export default class ListenCard extends React.Component<
   ListenCardProps,
   ListenCardState
 > {
+  static preciseTimestamp = (listen: Listen): string => {
```
MonkeyDo, line 49: We use formats other than Listen in other parts fo the website, for which we'll also want to use this method (once it's in utils.tsx).

In order to make it work everywhere, we'll want to pass just a timestamp argument rather than the whole listen object.

This `listen.listened_at_iso || listen.listened_at` business should be outside of this method consequently.

MonkeyDo (changes requested):
Thanks for opening this PR, that's a great improvement !

A few of us have been talking about these annoyingly vague timestamps for a little while now, and I just started to try and fix it this week in a WIP feature.
I was going for a similar implementation, but yours is more precise with the optional year, that's great!

We will use this method anywhere we display timestamps; consequently, I'll ask that you move the method to [static/js/src/utils.tsx](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/utils.tsx) so it can be imported in other components/pages.

alastair:
This is super cool, thanks for contributing it! I updated the title so that it references the ticket that we have open to track this task.

As @MonkeyDo said, I really like the year split too.

This is a topic that everyone has an idea about, and there might be some other opinions here, but I think that we should only show "x minutes ago" for listens less than an hour old. Anything else should have a timestamp.

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -46,6 +46,42 @@ export default class ListenCard extends React.Component<
...
+    const listened_at = listen.listened_at_iso || listen.listened_at * 1000;
+    const listenDate: Date = new Date(listened_at);
+    const msDifference = new Date().getTime() - listenDate.getTime();
+    if (
+      // over one year old : show with year
+      msDifference / (1000 * 3600 * 24 * 365) >
```
alastair, line 55: I'm just opening a suggestion here - should it be a listen over 365 days old, or should it be any listen from last year? e.g. in Feb 2021, a listen from Dec 2020 will include the year

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -46,6 +46,42 @@ export default class ListenCard extends React.Component<
   ListenCardProps,
   ListenCardState
 > {
+  static preciseTimestamp = (listen: Listen): string => {
+    const listened_at = listen.listened_at_iso || listen.listened_at * 1000;
+    const listenDate: Date = new Date(listened_at);
```
alastair, line 51: I don't know how the JS Date object works here, but listened_at is going to be in the UTC timezone. Is this object in the local tz, or UTC, and how does it compare with the timezone in the next line?

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -46,6 +46,42 @@ export default class ListenCard extends React.Component<
...
+    const msDifference = new Date().getTime() - listenDate.getTime();
+    if (
+      // over one year old : show with year
+      msDifference / (1000 * 3600 * 24 * 365) >
+      1
+    ) {
+      return `${listenDate.toLocaleString("en-US", {
```
alastair, line 58: Just asking a question here, should we hard-code this locale, or is there a way of using the browser's default locale?

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
MonkeyDo: I think if I'm in Feb 2021 and I see a date in March, I'll assume it's that of the past year. 
I think this assumption works well and we end up with a slightly less formal looking date.

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
MonkeyDo: Date expects a UTC timestamp and toLocaleString will transform it according to the current TZ (unless `timeZone` is specified in the options).

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
MonkeyDo: Nice catch, I missed that! 

We should use `.toLocaleString(undefined, {…options})` so that the user's browser's locale is automatically used (while still passing options as a second argument).

jdaok: Thank you for the review! :)

> This is a topic that everyone has an idea about, and there might be some other opinions here, but I think that we should only show "x minutes ago" for listens less than an hour old. Anything else should have a timestamp.

@alastair I considered this too. Personally since I listen and submit less frequently, they are older and so no "x minutes ago" formatted timestamps would appear on my my recent listens page most of the time. 

image

^ Users are still always able to view the exact timestamp for any listen by hovering. (I could also update this to be more similar to the >1 year format for more readability).

I think allowing listens up to 24 hours old (instead of 1 hour old) to be formatted this way makes the page look less formal for users who listen and submit less frequently, but I'd like to hear more opinions as well.

jdaok:
> This is a topic that everyone has an idea about, and there might be some other opinions here, but I think that we should only show "x minutes ago" for listens less than an hour old. Anything else should have a timestamp.

@alastair I considered this too. Personally since I listen and submit less frequently, they are older and so no "x minutes ago" formatted timestamps would appear on my my recent listens page most of the time. 

image

^ Users are still always able to view the exact timestamp for any listen by hovering. (I could also update this to be more similar to the >1 year format for more readability).

I think allowing listens up to 24 hours old (instead of 1 hour old) to be formatted this way makes the page look less formal for users who listen and submit less frequently, but I'd like to hear more opinions as well.

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -102,4 +103,44 @@ const getPlayButton = (listen: any, onClickFunction: () => void) => {
   );
 };
 
-export { searchForSpotifyTrack, getArtistLink, getTrackLink, getPlayButton };
+const preciseTimestamp = (listened_at: number): string => {
+  const listenDate: Date = new Date(listened_at);
```
MonkeyDo, line 107: With the comment above came the realization that we should probably add one last protection in case the expected  ISO date string or unix timestamp is not what we expect, or if an unexpected `undefined` is passed:

We can easily check if the Date object was able to create a new valid date like this:
```
if(isNaN(listenDate)){
    return '—';
}
```

Here I'm suggesting we return a placeholder rather than nothing if the date is invalid. How does that sound? Or should we `return listened_at;` instead?

MonkeyDo (changes requested):
I'm finding more things to tweak as I keep looking at the code, but it's starting to look really nice and *much* more robust than before ! 🎉 

It would be nice to see unit tests for `preciseTimestamp`  where you try throwing curveballs at it to see if you end up with a nice result anyway and nothing breaks.
A unit test file `utils.test.tsx` was [just created yesterday](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/utils.test.tsx), so you'll need to pull the recent changes on master into your branch.

Keep up the good work, this is an invaluable utility!

listenbrainz/webserver/static/js/src/utils.tsx
jdaok: I went with returning listened_at: all numbers will translate into valid outputs (even negative numbers for some reason) so I imagine that if the date is invalid it would be from an incorrectly formatted ISO string that might be more helpful than just a placeholder.

MonkeyDo (approved):
Awesome, thanks a lot !
We're going to use this utility everywhere!

PR #1316, opened 2021-03-05
Allow users to recommend recordings from the listens page

paramsingh:
# Problem

People should be able to share tracks to their followers.

# Solution

Added API call client method to the client and a dropdown option on the user's listens page.
Screenshot from 2021-03-05 13-53-08

# Action

Not sure why the spark build is failing, i haven't changed anything related to that.


paramsingh:
@brainzbot retest this please

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -130,6 +130,47 @@ export default class ListenCard extends React.Component<
...
+          metadata
+        );
+        if (status === 200) {
+          newAlert(
+            "success",
+            `You recommended a track to your followers!`,
+            `${metadata.artist_name} - ${metadata.track_name}`
+          );
+        }
+      } catch (error) {
```
MonkeyDo, line 165: This won't catch successful requests with an error code.
Currently if I receive a 500 error (error in the action, but successful request) nothing happens, no error is thrown by checkStatus in APIService.ts
An issue with checkStatus I think more than with your code, but until that's solved better to throw a helpful error yourself after `if (status === 200) {…}`.

MonkeyDo (commented):
Looks good !

Eventually we'll want to refactor the multiple card components we have so that this sort of feature works everywhere on the site without the need to implement it separately for each page. 
We currently have separate PlaylistItemCard and RecommendationCard and possibly the new TimelineEventCard.
Anyway, that's for another PR.

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
MonkeyDo: Scratch that ! The issue is in `APIService.recommendTrackToFollowers` where `this.checkStatus(response);` should be `await this.checkStatus(response);`

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -285,6 +326,10 @@ export default class ListenCard extends React.Component<
                       title="Delete Listen"
                       action={this.deleteListen}
                     />
+                    <ListenControl
+                      title="Recommend recording"
```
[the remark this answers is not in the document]
paramsingh: I'm happy with "recommend to my followers"

paramsingh:
@brainzbot retest this please

shivam-kapila:
This looks good. We might increase the width of the drop down menu so  that Recommend Recording comes in one line. Plus I would prefer it be above Delete action, positive above negative :).

MonkeyDo (commented):
I agree with Shivam for the dropdown menu width.
You can this bit of css to [this class](https://github.com/metabrainz/listenbrainz-server/blob/aad1ebba6ae7c8e4a48092cd1c8645aaba2feea8/listenbrainz/webserver/static/css/listens-page.less#L289):
`width: max-content;`

That should do the trick.

Apart from that, looks good to go !

MonkeyDo (approved):

paramsingh:
Javascript tests pass, I'm happy to merge.

PR #1317, opened 2021-03-05
Revert numpy version to 1.19.4

shivam-kapila:
# Problem
Numpy has dropped support for Python 3.6.x from version 1.20.0. [(reference).](https://github.com/numpy/numpy/blob/v1.20.0/setup.py#L29)
This is failing the Spark builds after #1303 was merged.

# Solution

Using version 1.19.5 fixes the problem. It still supports Python 3.6.x [(reference).](https://github.com/numpy/numpy/blob/v1.19.5/setup.py#L31)


paramsingh (approved):
lgtm, assuming build passes

PR #1315, opened 2021-03-04
Create timeline view for UserFeed page

MonkeyDo:
See https://docs.google.com/document/d/1_rVY55eHksV0GuydG3J1Ooo7Ud_ypXeFGHnDOP1_On0

WIP: Waiting for API endpoints and view to be merged for the feed events. In the meantime I'm using a `fake-user-feed.json` with fake data. See PR #1318.

This PR also removes the temporary hardcoded feature flag ( a list of users allowed to view feed and follow buttons)

With the UserFeed component moved to its own directory in the tree, you will need to run `./develop.sh build static_builder` to refresh the static builder image.

preview

PR #1318, opened 2021-03-05
Better feed endpoint

paramsingh:
# Problem

#1315 needs an API endpoint to show real data.

# Solution

This PR adds the API endpoint, tests and documentation.

I also removed the old feed/listens endpoint as we changed our approach quite a bit since.

### DIFF
--- a/listenbrainz/db/user_relationship.py
+++ b/listenbrainz/db/user_relationship.py
@@ -122,7 +122,7 @@ def get_follow_events(user_ids: Tuple[int], min_ts: int, max_ts: int, count: int
              WHERE ur.user_0 IN :user_ids
                AND ur.created >= :min_ts
                AND ur.created <= :max_ts
-          ORDER BY created
+          ORDER BY created DESC
              LIMIT :count
         """), {
             "user_ids": tuple(user_ids),
--- a/listenbrainz/db/user_relationship.py
+++ b/listenbrainz/db/user_relationship.py
@@ -131,8 +131,8 @@ def get_follow_events(user_ids: Tuple[int], min_ts: int, max_ts: int, count: int
               JOIN "user" follower ON ur.user_0 = follower.id
               JOIN "user" followed ON ur.user_1 = followed.id
              WHERE ur.user_0 IN :user_ids
-               AND ur.created >= :min_ts
-               AND ur.created <= :max_ts
+               AND ur.created > :min_ts
+               AND ur.created < :max_ts
           ORDER BY created DESC
              LIMIT :count
         """), {
--- a/listenbrainz/db/user_relationship.py
+++ b/listenbrainz/db/user_relationship.py
@@ -16,7 +16,7 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
-from typing import List
+from typing import List, Tuple
 
 from listenbrainz import db
 from listenbrainz.db.exceptions import DatabaseException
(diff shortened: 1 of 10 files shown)


paramsingh:
@brainzbot retest this please

listenbrainz/db/user_relationship.py
```diff
@@ -110,3 +111,35 @@ def get_following_for_user(user: int) -> List[dict]:
             "user": user,
         })
         return [dict(row) for row in result.fetchall()]
+
+def get_follow_events(user_ids: Tuple[int], min_ts: int, max_ts: int, count: int) -> List[dict]:
+    """ Gets a list of follow events for specified users.
+
+    user_ids is a tuple of user row IDs.
+
+    Returns a dict of the following format:
```
mayhem, line 120: Returns an array of dicts

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -85,3 +96,196 @@ def create_user_recording_recommendation_event(user_name):
...
+    user = validate_auth_header()
+    if user_name != user['musicbrainz_id']:
+        raise APIUnauthorized("You don't have permissions to view this user's timeline.")
+
+    db_conn = webserver.create_timescale(current_app)
+    min_ts, max_ts, count, time_range = _validate_get_endpoint_params(db_conn, user_name)
+    if min_ts is None and max_ts is None:
+        max_ts = int(time.time())
+
+    users_following = db_user_relationship.get_following_for_user(user['id'])
```
mayhem, line 128: Making n queries, where n is the number of event types, for events from the same DB table is a bad pattern. We should rework this to accomplish all this in one query. This approach is really not future proof.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -85,3 +96,196 @@ def create_user_recording_recommendation_event(user_name):
...
+    users_for_feed_events = users_following + [user]
+    follow_events = get_follow_events(
+        user_ids=tuple(user['id'] for user in users_for_feed_events),
+        min_ts=min_ts or 0,
+        max_ts=max_ts or int(time.time()),
+        count=count,
+    )
+
+    recording_recommendation_events = get_recording_recommendation_events(
+        users_for_events=users_for_feed_events,
```
mayhem, line 151: These blocks that return empty data bug me. Duplicate code; can we just return None or at least a helper function?

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -85,3 +96,196 @@ def create_user_recording_recommendation_event(user_name):
...
+        min_ts=min_ts or 0,
+        max_ts=max_ts or int(time.time()),
+        count=count,
+    )
+
+    # TODO: add playlist event and like event
+    all_events = sorted(listen_events + follow_events + recording_recommendation_events, key=lambda event: -event.created)
```
mayhem, line 158: The DB should do this.

mayhem (changes requested):
I think we need to reduce the number of DB queries for fetching the timeline data.

listenbrainz/webserver/views/user_timeline_event_api.py
paramsingh: Not completely sure I understand, this code doesn't return empty data, it returns recording recommendation events.

listenbrainz/webserver/views/user_timeline_event_api.py
paramsingh: the events live in a number of different tables.

like -> recording_feedback
listen -> listen table in timescale
followers and follow events -> user_relationship
recording_recommendation -> user_timeline_event

I don't think it's realistic to expect to get all events in just the one query.

listenbrainz/webserver/views/user_timeline_event_api.py
mayhem: ok, may not be possible..

paramsingh:
As per discussion on IRC, ticket about caching created: https://tickets.metabrainz.org/browse/LB-829

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -85,3 +96,196 @@ def create_user_recording_recommendation_event(user_name):
...
+
+@user_timeline_event_api_bp.route('/user/<user_name>/feed/events', methods=['OPTIONS', 'GET'])
+@crossdomain(headers="Authorization, Content-Type")
+@ratelimit()
+def user_feed(user_name: str):
+    """ Get feed events for a user's timeline.
+
+    :param user_name: The MusicBrainz ID of the user whose timeline is being requested.
+    :type user_name: ``str``
+    :param max_ts: If you specify a ``max_ts`` timestamp, events with timestamps less than or equal to it will be returned
```
mayhem, line 109: Our get_listens endpoint uses non-inclusive timestamps and this endpoint uses inclusive-timestamps. Is there a compelling reason to use inclusive timestamps here? It would be nice to keep consistency across API calls.

mayhem (commented):
OK, given our discussion and the addition of LB-829 this looks good now, except for my question on inclusive vs non-inclusive timestamps.

paramsingh:
The only unit test that failed is a timescale test that is unrelated and it seems flaky. I'm ok with merging this.

PR #1319, opened 2021-03-05
Add similarity score component

shivam-kapila:
This PR aims to add a similarityScore react component. The mockup looks as follows against various similarityScore values:

### < 0.3
image

### > 0.3 and < 0.7
image

### > 0.7
image

### Compact mode
image

P.S. No better name for the component came to my mind. Suggestions please.

### DIFF
--- a/listenbrainz/webserver/static/js/src/SimilarityScore.tsx
+++ b/listenbrainz/webserver/static/js/src/SimilarityScore.tsx
@@ -9,11 +9,11 @@ export type SimilarityScoreProps = {
 const getclassName = (similarityScore: number): string => {
   let className = "";
   if (similarityScore <= 0.3) {
-    className = "progress-bar-danger";
+    className = "red";
   } else if (similarityScore <= 0.7) {
-    className = "progress-bar-warning";
+    className = "orange";
   } else {
-    className = "progress-bar-success";
+    className = "purple";
   }
   return className;
 };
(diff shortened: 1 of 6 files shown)


listenbrainz/webserver/static/js/src/SimilarityScore.tsx
```diff
@@ -0,0 +1,49 @@
...
+      >
+        <div
+          className={`progress-bar ${className}`}
+          style={{
+            width: `${similarityScore * 100}%`,
+          }}
+        />
+      </div>
+      <p>
+        Your compatibility with {user.name} is {similarityScore * 10}/10!
```
MonkeyDo, line 43:
```suggestion
        Your compatibility with {user.name} is {similarityScore * 10}/10
```

I like the enthusiasm, but I'd like to suggest a more toned-down version :)

listenbrainz/webserver/static/js/src/SimilarityScore.tsx
```diff
@@ -0,0 +1,49 @@
...
+        tabIndex={0}
+      >
+        <div
+          className={`progress-bar ${className}`}
+          style={{
+            width: `${similarityScore * 100}%`,
+          }}
+        />
+      </div>
+      <p>
```
MonkeyDo, line 42: We might want a flag like a `showText` boolean in the props, and only return the text if it's set.
That way we can (as you suggested) show the similarity score in the "People similar to you" list.

Maybe the version without the "Your compatibility with…" phrase should have just `{similarityScore * 10}/10` next to or on the progress bar ?

MonkeyDo (commented):
Haven't tried to render the component out anywhere but the code looks good.
Liking what I see on the screenshots!

listenbrainz/webserver/static/js/src/SimilarityScore.tsx
shivam-kapila: Your wish has been granted :p

paramsingh (commented):
Code looks good to me, is there a way to possibly make the colors match up the LB theme more? Like purple instead of green or something like that?

paramsingh (approved):

MonkeyDo (approved):

PR #1321, opened 2021-03-08
Refresh token might not always be present in response

amCap1712:


paramsingh (approved):
lgtm, let's release this asap as it is causing errors for users

listenbrainz/domain/spotify.py
```diff
@@ -110,9 +110,9 @@ def refresh_user_token(spotify_user: Spotify):
             break
         elif response.status_code == 400:
             error_body = response.json()
-            if "error" in error_body and error_body["error"] == "invalid_grant" and \
-                    "error_description" in error_body and error_body["error_description"] == "Refresh token revoked":
-                # user has revoked authorization through spotify ui, delete from our spotify db as well.
+            if "error" in error_body and error_body["error"] == "invalid_grant":
+                # user has revoked authorization through spotify ui or deleted their spotify account etc.
+                # in any of these case, we should delete usr from our spotify db as well.
```
shivam-kapila, line 115:
```suggestion
                # in any of these cases, we should delete user from our spotify db as well.
```

shivam-kapila (changes requested):
Oops I noticed approval afterwards. Feel free to ignore this for now.

PR #1325, opened 2021-03-08
Spotify reader do not process deleted users

amCap1712:
At one place, the code does not whether the user is `None` before trying to process a user. Instead of returning None, raise an Exception in case of invalid grant received from Spotify API. An exception will stop the execution before and we can also know the reason why it happened.

### DIFF
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -280,6 +280,8 @@ def get_user_dict(user_id):
 
 
 class SpotifyInvalidGrantError(Exception):
+    """ Raised if spotify API returns invalid_grant during authorization. This usually means that the user has revoked
+    authorization to the ListenBrainz application through Spotify UI."""
     pass
 
 
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -111,9 +111,6 @@ def refresh_user_token(spotify_user: Spotify):
         elif response.status_code == 400:
             error_body = response.json()
             if "error" in error_body and error_body["error"] == "invalid_grant":
-                # user has revoked authorization through spotify ui or deleted their spotify account etc.
-                # in any of these cases, we should delete user from our spotify db as well.
-                db_spotify.delete_spotify(spotify_user.user_id)
                 raise SpotifyInvalidGrantError(error_body)
 
         response = None  # some other error occurred
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -92,10 +92,10 @@ def refresh_user_token(spotify_user: Spotify):
 
     Returns:
         user (domain.spotify.Spotify): the same user with updated tokens
-        None: if the user has revoked authorization to spotify
 
     Raises:
         SpotifyAPIError: if unable to refresh spotify user token
+        SpotifyInvalidGrantError: if the user has revoked authorization to spotify
 
     Note: spotipy eats up the json body in case of error but we need it for checking
     whether the user has revoked our authorization. hence, we use our own
@@ -114,10 +114,9 @@ def refresh_user_token(spotify_user: Spotify):
                 # user has revoked authorization through spotify ui or deleted their spotify account etc.
                 # in any of these cases, we should delete user from our spotify db as well.
                 db_spotify.delete_spotify(spotify_user.user_id)
-                return None
+                raise SpotifyInvalidGrantError(error_body)
 
-            response = None  # some other error during request
-
+        response = None  # some other error occurred
         retries -= 1
 
     if response is None:
@@ -281,6 +280,10 @@ def get_user_dict(user_id):
         'access_token': user.user_token,
         'permission': user.permission,
     }
+
+
+class SpotifyInvalidGrantError(Exception):
+    pass
 
 
 class SpotifyImporterException(Exception):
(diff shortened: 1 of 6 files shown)


listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -277,6 +277,7 @@ def process_one_user(user):
         else:
             if user is None:
                 current_app.logger.debug("%s has revoked spotify authorization", str(user))
+                return
```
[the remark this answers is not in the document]
amCap1712: Done

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -171,12 +171,8 @@ def make_api_request(user, spotipy_call, **kwargs):
                 if not tried_to_refresh_token:
                     try:
                         user = spotify.refresh_user_token(user)
-                    except SpotifyException as err:
+                    except (SpotifyException, spotify.SpotifyInvalidGrantError) as err:
```
alastair, line 174: This feels a bit funny catching both an exception from Spotify and an exception raised from `refresh_user_token` in the same line. Do we think that the error message in both of these cases should be the same?
In the case where we know (?) that the user has removed permission, we report it as "please unlink and link your account again". Perhaps in this case we can explicitly say something like "It seems that we no longer have permission to read your account"

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -274,9 +270,9 @@ def process_one_user(user):
         except spotify.SpotifyAPIError:
             current_app.logger.error('Could not refresh user token from spotify', exc_info=True)
             raise
-        else:
-            if user is None:
-                current_app.logger.debug("%s has revoked spotify authorization", str(user))
+        except spotify.SpotifyInvalidGrantError:
+            current_app.logger.error("%s has revoked spotify authorization", str(user), exc_info=True)
```
alastair, line 274: I'm not sure that we should log this as an error. Consider that an error should probably be actionable by us, and in this case we can't do anything.
Because we have sentry, we'll get an email every time this happens. If we want to use sentry as a logger here, I think it'd be better to use info or similar. Additionally, it'd be great to have this grouped as a single error message - we could use parameters to store the username on each event.

listenbrainz/domain/spotify.py
```diff
@@ -283,6 +282,10 @@ def get_user_dict(user_id):
     }
 
 
+class SpotifyInvalidGrantError(Exception):
```
alastair, line 285: comment describing what this exception is used for

listenbrainz/spotify_updater/spotify_read_listens.py
amCap1712: Now that I go through this part again, I don't think we need to catch these exceptions here. These will be handled in `process_one_user`.

listenbrainz/spotify_updater/spotify_read_listens.py
amCap1712: Agreed.

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -146,7 +147,7 @@ def make_api_request(user, spotipy_call, **kwargs):
                 # Rate Limit Problems -- the client handles these, but it can still give up
                 # after a certain number of retries, so we look at the header and try the
                 # request again, if the error is raised
-                time_to_sleep = e.headers.get('Retry-After', delay)
+                time_to_sleep = int(e.headers.get('Retry-After', delay))
```
alastair, line 150: this could raise ValueError if for some reason it wasn't an int. add try/catch with a reasonable default. maybe 1 second?

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -267,12 +264,23 @@ def process_one_user(user):
...
-            raise
+        except spotify.SpotifyAPIError as e:
+            # if it is an error from the Spotify API, show the error message to the user
+            spotify.update_last_updated(
+                user_id=user.user_id,
+                success=False,
+                error_message=str(e),
+            )
+            if not current_app.config['TESTING']:
+                notify_error(user.musicbrainz_row_id, str(e))
```
alastair: I don't think it's a great idea to send the contents of an exception to a user. Do you know if these exception messages from spotipy are nice to read? Can a user get more detailed information if they read this error, or is the result of all of them just going to be "you should re-authenticate"?

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -267,12 +264,23 @@ def process_one_user(user):
...
+            raise spotify.SpotifyListenBrainzError("Could not refresh user token from spotify")
         except spotify.SpotifyInvalidGrantError:
-            current_app.logger.error("%s has revoked spotify authorization", str(user), exc_info=True)
-            raise
+            if not current_app.config['TESTING']:
+                notify_error(user.musicbrainz_row_id, "User has revoked authorization through spotify")
```
alastair: similar to the above comment - this message goes to a user, so it should be something that they can action - "It seems like you've revoked permission for us to read your spotify account"

alastair (approved):
some small items mentioned as individual comments, but after that this is good to go

listenbrainz/spotify_updater/spotify_read_listens.py
amCap1712: Since we are defaulting to delay if the header is missing, let's use that in case of an invalid header as well.

listenbrainz/spotify_updater/spotify_read_listens.py
amCap1712: This will send "Could not authenticate with Spotify, please unlink and link your account again." to the user.

PR #1326, opened 2021-03-09
Improve startup processes when consul goes wonky

amCap1712:
This PR adds more visibility to the startup of LB services and adds reporting to inform us when something goes wrong.

Current situation:
 * consul-template hangs while waiting for a config key if one doesn't exist, causing no config file to be generated, and the process in the container doesn't get started
 * Sometimes it's not clear what is being started in a container
 * Sometimes services in a container restart and we don't know why. We want more visibility into when this happens

We have the following startup changes:
 * Add some more explicit logging in rc.local to indicate what is happening
 * Use `keyOrDefault` in consul-template instead of `key`. This means that consul-template will render the config file even if the value is missing in consul (default behaviour with key is to hang until the value appears)
 * Always use run-lb-command, even when running uwsgi (see api-compat example)
 * In run-lb-command, grep for sentinel default settings. If these exist, log a message to the console and exit. Otherwise, startup
 * when processes start up, print a small message saying which service is running
 * when starting uwsgi, exit if the python app was unable to be loaded. Previous behaviour was to stay running but always return http500, which was a bit vague. This helps us to report this case to sentry
 * When consul-template gives uwsgi SIGHUP to reload config, completely shut down uwsgi. This allows us to report this case to sentry
 * If a process quits for any reason, report to sentry that this has happened. The two main cases where this is going to happen is if consul-template gets a changed value, or we shut down a container for redeploy

Questions:
* Should we also log to sentry the missing items if config items are missing? (it means that we'd get 2 notifications, one for the missing items and one for the restart)
* Should we make separate events in sentry? If there are multiple messages with the same subject they get bundled into a single issue. This means that we won't get emails the second time that it happens.
* If we have missing config items, we currently sleep for 10 seconds and then restart. Do we want to sleep forever (or 9999) to make it clear what the error is, or should we keep a short-ish sleep so that we can push a config item into consul and then have the service recover nicely?

Still to do:
* [ ] Check other config files if they're rendered (config.sh in cron)
* [x] Need to apply the same pattern to all services
* [x] Need to move common bash patterns into a util file
* [ ] Need to add the sentry DSN to docker-startup-configs

cc @atj, maybe you have some bash tips!


consul_config.py.ctmpl
```diff
@@ -1,9 +1,9 @@
 {{- define "KEY" -}}
-    {{ key (printf "docker-server-configs/LB/config.%s.json/%s" (env "DEPLOY_ENV") .) }}
+    {{ keyOrDefault (printf "docker-server-configs/LB/config.%s.json/%s" (env "DEPLOY_ENV") .)  "KEYDOESNOTEXIST" }}
```
alastair, line 2: can we make this value include the key itself? so we end up with something like `KEYDOESNOTEXIST_ incoming_exchange`. This allows us to include information about the missing key directly in an error message

consul_config.py.ctmpl
```diff
@@ -18,6 +18,8 @@ SQLALCHEMY_DATABASE_URI = "postgresql://listenbrainz:listenbrainz@{{.Address}}:{
 MESSYBRAINZ_SQLALCHEMY_DATABASE_URI = "postgresql://messybrainz:messybrainz@{{.Address}}:{{.Port}}/messybrainz"
 POSTGRES_ADMIN_URI="postgresql://postgres@{{.Address}}:{{.Port}}/postgres"
 {{end}}
+{{else}}
+KEYDOESNOTEXIST = "pgbouncer-master is down"
```
alastair, line 22: I'd prefer that this was something like
`SQLALCHEMY_DATABASE_URI = "SERVICEDOESNTEXIST_pgbouncer_master"`
or similar. Add it for all config items inside the `if`. This is the same motivation as adding the key name to the KEYDOESNTEXIST message.

By adding a field for each config item, we actually allow the service to load a config if we want to.
Currently when apps start up, if a service is missing they try and read a config item but fail. I think it's better to be able to read the value, but then fail at the next stage.
I'm happy to discuss this, as I'm not 100% convinced that it's the right thing to do.

alastair (changes requested):

consul_config.py.ctmpl
amCap1712: Makes sense to me. One possible issue could be whether the service is able to handle an incorrect config. For API keys I think this could be an issue. But for database connections I think the connection would just fail logging an exception which is fine.

alastair: yes, that's a valid concern. The hope here is that we'll get a notification before the service starts anyway, so this might not be as much of a problem

consul_config.py.ctmpl
```diff
@@ -1,9 +1,9 @@
 {{- define "KEY" -}}
-    {{ key (printf "docker-server-configs/LB/config.%s.json/%s" (env "DEPLOY_ENV") .) }}
+    {{ keyOrDefault (printf "docker-server-configs/LB/config.%s.json/%s" (env "DEPLOY_ENV") .)  "KEYDOESNOTEXIST_{{.}}" }}
```
alastair, line 2: this `{{.}}` syntax isn't correct, it just gets output as a . and not the parameter

amCap1712: Looking into it.

docker/services/api_compat/api_compat.finish
```diff
@@ -0,0 +1,32 @@
...
+  echo $(date '+%Y-%m-%d %H:%M:%S %Z') "$@"
+}
+
+message="api compat terminated. "
+if [ $# -gt 0 ] && [ $1 -eq 0 ]; then
+  message+="Exit status 0 (consul triggered reload or service was stopped)"
+elif [ $# -gt 0 ] && [ $1 -eq 22 ]; then
+  message+="Exit status 22 (uwsgi failed to start app, check container logs)"
+elif [ $# -gt 0 ]; then
+  message+="Exit status $1"
```
alastair, line 13: I'm also using status 10 in `run-lb-command` to indicate that there was a missing config key. We could also add this check ehre

docker/services/api_compat/consul-template-api-compat.conf
```diff
@@ -4,8 +4,8 @@ template {
 }
 
 exec {
-    command = "uwsgi --die-on-term /etc/uwsgi/uwsgi-api-compat.ini"
-    splay = "60s"
+    command = "run-lb-command uwsgi --die-on-term /etc/uwsgi/uwsgi-api-compat.ini"
+    splay = "5s"
```
alastair, line 8: 60s was too long - it meant that a signal from consul-template could have taken up to 60s to get to our app

alastair:
Example sentry log:
https://sentry.metabrainz.org/metabrainz/service-startup-errors/issues/106570/?query=is%3Aunresolved

Here's some examples of the container output now:


missing config item
```
web_1               | *** Running /etc/rc.local...
-->  explicit messages from rc.local about what is happening
web_1               | 2021-03-09 17:51:50 UTC Hello, this is rc.local. CONTAINER_NAME is listenbrainz-api-compat-prod, and DEPLOY_ENV is prod
web_1               | 2021-03-09 17:51:50 UTC rc.local: Enabling api compat
web_1               | *** Booting runit daemon...
web_1               | *** Runit started as PID 11
web_1               | Mar  9 17:51:50 3dc28afe9878 syslog-ng[26]: syslog-ng starting up; version='3.5.6'
web_1               | 2021/03/09 17:51:55 [DEBUG] (logging) enabling syslog on LOCAL0
-->  explicit message about the command being run before it is executed
web_1               | 2021-03-09 17:51:55 UTC run-lb-command uwsgi --die-on-term /etc/uwsgi/uwsgi-api-compat.ini
-->  explicit message about a consul error
web_1               | 2021-03-09 17:51:55 UTC Found keys that are missing from consul
web_1               | SECRET_KEY = '''KEYDOESNOTEXIST_docker-server-configs/LB/config.prod.json/secret_key'''
web_1               | Consul Template returned errors:
web_1               | child process died with exit code 10
-->  this message is also sent to sentry
web_1               | 2021-03-09 17:51:55 UTC api compat terminated. Exit status 10
web_1               | 2021-03-09 17:51:55 UTC Exited with non-0 status, sleeping 10 seconds
```

consul value changes and config file gets re-rendered:
```
web_1               | *** uWSGI is running in multiple interpreter mode ***
web_1               | spawned uWSGI master process (pid: 43)
web_1               | spawned uWSGI worker 1 (pid: 49, cores: 1)
--> config file changes at this moment
web_1               | ...gracefully killing workers...
web_1               | Gracefully killing worker 1 (pid: 49)...
web_1               | worker 1 buried after 1 seconds
web_1               | uWSGI: GAME OVER (insert coin)
web_1               | Consul Template returned errors:
web_1               | child process died with exit code 0
--> This message reported to sentry
web_1               | 2021-03-09 18:00:29 UTC api compat terminated. Exit status 0 (consul triggered reload or service was stopped)
--> service starts again
web_1               | 2021/03/09 18:00:44 [DEBUG] (logging) enabling syslog on LOCAL0
web_1               | 2021-03-09 18:00:44 UTC run-lb-command uwsgi --die-on-term /etc/uwsgi/uwsgi-api-compat.ini
web_1               | [uWSGI] getting INI configuration from /etc/uwsgi/uwsgi-api-compat.ini
```

some error occurs when loading the app
```
web_1               |   File "/usr/local/lib/python3.7/site-packages/redis/connection.py", line 509, in __init__
web_1               |     self.port = int(port)
web_1               | ValueError: invalid literal for int() with base 10: 'SERVICEDOESNOTEXIST_listenbrainz-redis'
web_1               | unable to load app 0 (mountpoint='') (callable not found or import error)
web_1               | *** no app loaded. GAME OVER ***
web_1               | Consul Template returned errors:
web_1               | child process died with exit code 22
--> specific log message. this is sent to sentry too
web_1               | 2021-03-09 17:49:42 UTC api compat terminated. Exit status 22 (uwsgi failed to start app, check container logs)
web_1               | 2021-03-09 17:49:42 UTC Exited with non-0 status, sleeping 10 seconds
```

docker/services/api_compat/api_compat.finish
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
alastair, line 1: this file needs to be duplicated for each service. we should have an abstracted function that contains all duplicate code, and call that from each service's `finish` file.

alastair:
one more thing that came to mind. In the case of the cron container we have a consul-template instance to generate config files, but this is separate from the runit service that starts cron.
This means we have no way of logging something if a config item in cron changes, and we don't have a hook to log if the cron service quits. Do we think it's worth adding one?

docker/services/api_compat/api_compat.finish
atj: You create a generic log function that is included by all your finish files:

```shell
function log {
        printf "%(%Y-%m-%d %H:%M:%S %Z)T ${LOGTAG:-$0} "
        echo $@
}
```

Then at the top of each file:

```shell
export LOGTAG="api_compat"
. common.sh
```

atj: A generic log function is a good idea:

```shell
function log {
        printf "%(%Y-%m-%d %H:%M:%S %Z)T ${LOGTAG:-$0} "
        echo $@
}
```

Then at the top of each finish file:

```shell
export LOGTAG="api_compat"

. /etc/functions.sh
```

consul_config.py.ctmpl
amCap1712: I think the issue was that this statement is inside another `{{}}`. I have replaced this with a `printf`.

docker/services/api_compat/uwsgi-api-compat.ini
```diff
@@ -10,3 +10,7 @@ enable-threads = true
 processes = 50
 log-x-forwarded-for=true
 disable-logging = true
+; quit uwsgi if the python app fails to load
+need-app = true
```
mayhem, line 14: 👏

listenbrainz/spark/spark_reader.py
```diff
@@ -103,7 +103,7 @@ def start(self):
         """
 
         with self.app.app_context():
-
+            current_app.logger.info('Spark consumer has started!')
```
mayhem, line 106: I like this standard "container has started" message. Should we have a "container exited because FOO" message as well?

mayhem (commented):
Overall this looks like a drastic improvement. I'll leave some more comments to the answer the questions.

mayhem:
Right now we print the contents of the config object at startup, which is really nice if something goes wrong. But 99% of the time it is noise. Would it make sense to print this only when the container fails to start?

listenbrainz/spark/spark_reader.py
alastair: We have this message, printed by the runit `finish` script:

`web_1               | 2021-03-09 18:00:29 UTC api compat terminated. Exit status 0 (consul triggered reload or service was stopped)`

It's difficult to get a hook in the right place in python to print such a message.

mayhem:
> Should we also log to sentry the missing items if config items are missing? (it means that we'd get 2 notifications, one for the missing items and one for the restart)

For dire events like containers not starting, more messages is better.

> Should we make separate events in sentry? If there are multiple messages with the same subject they get bundled into a single issue. This means that we won't get emails the second time that it happens.

What if we add a timestamp to the message itself? That will make each message unique and generate a unique notification.

> If we have missing config items, we currently sleep for 10 seconds and then restart. Do we want to sleep forever (or 9999) to make it clear what the error is, or should we keep a short-ish sleep so that we can push a config item into consul and then have the service recover nicely?

Without question, the latter. This and only this is why LB recovers automatically when something critical like PG goes away and then comes back.

alastair:
> Without question, the latter. This and only this is why LB recovers automatically when something critical like PG goes away and then comes back.

That's a really good point. Note that we're using two different sentinels here - one for a missing key and one for a missing service, so we could restart quickly if a service is down, but more slowly if a key is missing. What I'm trying to avoid here is spamming sentry with a million messages before we get to fixing the problem.

> What if we add a timestamp to the message itself? That will make each message unique and generate a unique notification.

That will definitely work. I'm having a look at sentry "fingerprints", which is how it does this automatic grouping. If we just set a new fingerprint to a new uuid, it should store each event separately.

alastair (approved):
I've made some final changes to this, I'm happy to merge and test it now

PR #1331, opened 2021-03-09
LB-830: Make all buttons on listen page equal width

phw:
# Problem

The three dots icon on the listens page is very narrow, making it hard to click. It is even difficult with a mouse, and trying to hit this on my mobile completely drives me nuts

The clickable area on those icons should be expanded. The three dots is exactly as wide as the dots, making it wider would solve the issue.

See LB-830

# Solution
This fixes the three dots button to be very narrow and hard to hit. 1em is the height and the width of the other rectangular buttons. The three dots button ends up centered in this width.


paramsingh:
@brainzbot add to whitelist

PR #1335, opened 2021-03-11
Fix removing playlist collaborators

MonkeyDo:
The current playlist page does not allow to delete collaborators properly because of these two problems:

1. On the front-end a utility is used to duplicate a playlist before sending it to be saved, but that utility merges the existing collaborators array rather than replacing it. That means if you remove all the collaborators and save, the playlist will still have the original collaborators sent to be saved.

I copied the working code we use for the Playlists.tsx page. Ironically that object destructuring method is what was used in Playlist.tsx before being replaced with `defaultsDeep`.

2. When the playlist owner is the only collaborator added to a playlist, we do a check for the length of that list *before*  removing the owner from the list of collaborators, leaving us to fetch an empty list of user IDs which throws an error.

### DIFF
--- a/listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
+++ b/listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
@@ -569,20 +569,17 @@ export default class PlaylistPage extends React.Component<
       return;
     }
     try {
-      // Replace keys that have changed but keep all other attributres
-      const editedPlaylist: JSPFPlaylist = defaultsDeep(
-        {
-          annotation: description,
-          title: name,
-          extension: {
-            [MUSICBRAINZ_JSPF_PLAYLIST_EXTENSION]: {
-              public: isPublic,
-              collaborators,
-            },
+      const editedPlaylist: JSPFPlaylist = {
+        ...playlist,
+        annotation: description,
+        title: name,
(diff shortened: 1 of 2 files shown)


mayhem (approved):
Python side is fine.

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
```diff
@@ -569,20 +569,17 @@ export default class PlaylistPage extends React.Component<
...
-      const editedPlaylist: JSPFPlaylist = defaultsDeep(
-        {
-          annotation: description,
-          title: name,
-          extension: {
-            [MUSICBRAINZ_JSPF_PLAYLIST_EXTENSION]: {
-              public: isPublic,
-              collaborators,
-            },
+      const editedPlaylist: JSPFPlaylist = {
```
MonkeyDo, line 572: This code I copied form the [Playlists.tsx component](https://github.com/metabrainz/listenbrainz-server/blob/5b5ad5d11500ee8b4f64caf65b3a5c3f8fd67624/listenbrainz/webserver/static/js/src/playlists/Playlists.tsx#L339-L349), which is working contrarily to the `defaultsDeep` replacement which turned out to be a bad idea.

PR #1336, opened 2021-03-12
UI improvements

jdaok:
This PR makes some minor tweaks to the UI.

- Make follow user button more responsive
Before
gif33
After
gif12

- Change visited link styling
image
=>
image

- Progress bar on importer


MonkeyDo:
Hi @jdaok ,

Thanks for lookig into this. I'll ask you to open separate pull requests for each improvement, it will makes reviews a lot easier :)

shivam-kapila:
Hi, I had a couple of comments before you work on these as separate PRs.

> Make follow user button more responsive
  I think that was intended so that the button doesn't lose focus immediately and gives time to read. Even if we change it, I noticed an unnecessary orange transition. Please remove that.

> Change visited link styling
I think the visited link should have some difference at least, maybe some other color if you want to change?

>  Progress bar on importer
Yay!!

MonkeyDo:
If you ever find time to go over this again @jdaok I loved the LastFM importer progress bar idea, and I think there's value in the changes you did to the follow button too.
As mentioned above please open separate PRs for those two improvements.

The link styling is not something I see us doing anytime soon. Those are standard browser mechanisms, and I find that removing that visited link indication makes navigation more difficult.

PR #1338, opened 2021-03-12
LB-826: changed config.sh to spark_config.sh

r0han22:
# Problem
There are two files named config.sh

# Solution
Renaming the config.sh file with spark settings to spark_config.sh


amCap1712 (changes requested):
The `listenbrainz_spark.zip` file should not be committed. I think it makes to sense to add it to `.gitignore` to avoid this in the future.

shivam-kapila:
@alastair isn't the file also used in `start-request-consumer.sh` or its the other config.sh
http://livegrep.metabrainz.org/search/livegrep?q=config.sh&fold_case=auto&regex=false&context=true&repo%5B%5D=metabrainz%2Flistenbrainz-server

alastair:
@shivam-kapila good catch, yes. this also needs to be updated in start-request-consumer.sh

r0han22:
Make the required changes. That was a little sloppy of me, sorry about that

alastair (approved):
thank you @r0han22, no problem! this is a big code base.

alastair:
I rebased this PR to delete the commit that added the zip file, this way we won't have it lying around in the history. thanks for the contribution!

PR #1339, opened 2021-03-13
Set retry limit for LastFMImporter

amCap1712:
The `LastFMImporter` _used to retry_ indefinitely in case of errors (e.g. _50x_). Further, when a page import failed, we _used to_ move to the next page, while the previous pages continued to retry.

We are now restricting the number of retries to _3_. We also _do not_ move to the next page while we are still retrying the current page.

A few tests have also been added to ensure that the retry logic works as expected.

### DIFF
--- a/listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
+++ b/listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { mount, shallow } from "enzyme";
-import LastFmImporter from "./LastFMImporter";
+import LastFmImporter, {LASTFM_RETRIES} from "./LastFMImporter";
 // Mock data to test functions
 import * as page from "./__mocks__/page.json";
 import * as getInfo from "./__mocks__/getInfo.json";
@@ -133,7 +133,6 @@ describe("getTotalNumberOfScrobbles", () => {
 });
 
 describe("getPage", () => {
-  const RETRIES = 3;
   beforeEach(() => {
     const wrapper = shallow<LastFmImporter>(<LastFmImporter {...props} />);
     instance = wrapper.instance();
@@ -148,7 +147,7 @@ describe("getPage", () => {
   });
 
   it("should call with the correct url", () => {
-    instance.getPage(1, RETRIES);
+    instance.getPage(1, LASTFM_RETRIES);
 
     expect(window.fetch).toHaveBeenCalledWith(
       `${props.lastfmApiUrl}?method=user.getrecenttracks&user=${instance.state.lastfmUsername}&api_key=${props.lastfmApiKey}&from=1&page=1&format=json`
@@ -159,7 +158,7 @@ describe("getPage", () => {
     // Mock function for encodeScrobbles
     LastFmImporter.encodeScrobbles = jest.fn(() => ["foo", "bar"]);
 
-    const data = await instance.getPage(1, RETRIES);
+    const data = await instance.getPage(1, LASTFM_RETRIES);
     expect(LastFmImporter.encodeScrobbles).toHaveBeenCalledTimes(1);
     expect(data).toEqual(["foo", "bar"]);
   });
@@ -174,10 +173,10 @@ describe("getPage", () => {
     });
 
     const getPageSpy = jest.spyOn(instance, "getPage");
-    await instance.getPage(1, RETRIES);
+    await instance.getPage(1, LASTFM_RETRIES);
 
-    // we run the timer sufficient number of timers to ensure retries are not exceeded or undetected due to timeouts
-    for (let i = 0; i < 10; i += 1) {
+    // we run the timer sufficient number of times to ensure retries are not exceeded or undetected due to timeouts
+    for (let i = 0; i < LASTFM_RETRIES + 2; i += 1) {
       jest.runAllTimers();
       // make timers and promises play well with jest
       // eslint-disable-next-line no-await-in-loop
@@ -196,10 +195,10 @@ describe("getPage", () => {
       });
     });
     const getPageSpy = jest.spyOn(instance, "getPage");
-    await instance.getPage(1, RETRIES);
+    await instance.getPage(1, LASTFM_RETRIES);
 
-    // we run the timer sufficient number of timers to ensure retries are not exceeded or undetected due to timeouts
-    for (let i = 0; i < 10; i += 1) {
+    // we run the timer sufficient number of times to ensure retries are not exceeded or undetected due to timeouts
+    for (let i = 0; i < LASTFM_RETRIES + 2; i += 1) {
       jest.runAllTimers();
       // make timers and promises play well with jest
       // eslint-disable-next-line no-await-in-loop
@@ -220,9 +219,9 @@ describe("getPage", () => {
 
     const getPageSpy = jest.spyOn(instance, "getPage");
 
-    await instance.getPage(1, RETRIES);
-    // we run the timer sufficient number of timers to ensure retries are not exceeded or undetected due to timeouts
-    for (let i = 0; i < 10; i += 1) {
+    await instance.getPage(1, LASTFM_RETRIES);
+    // we run the timer sufficient number of times to ensure retries are not exceeded or undetected due to timeouts
+    for (let i = 0; i < LASTFM_RETRIES + 2; i += 1) {
       jest.runAllTimers();
       // make timers and promises play well with jest
       // eslint-disable-next-line no-await-in-loop
--- a/listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
+++ b/listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
@@ -183,7 +183,7 @@ describe("getPage", () => {
       await Promise.resolve();
     }
 
-    expect(getPageSpy).toHaveBeenCalledTimes(4);
+    expect(getPageSpy).toHaveBeenCalledTimes(1 + LASTFM_RETRIES);
   });
 
   it("should skip the page if 40x is recieved", async () => {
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
```diff
@@ -172,11 +173,18 @@ describe("getPage", () => {
...
     });
 
-    await instance.getPage(1);
-    // There is no direct way to check if retry has been called
-    expect(setTimeout).toHaveBeenCalledTimes(1);
+    const getPageSpy = jest.spyOn(instance, "getPage");
+    await instance.getPage(1, RETRIES);
 
-    jest.runAllTimers();
+    // we run the timer sufficient number of timers to ensure retries are not exceeded or undetected due to timeouts
```
shivam-kapila, line 179:
```suggestion
    // we run the timer sufficient number of times to ensure retries are not exceeded or undetected due to timeouts
```

listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
```diff
@@ -133,6 +133,7 @@ describe("getTotalNumberOfScrobbles", () => {
 });
 
 describe("getPage", () => {
+  const RETRIES = 3;
```
shivam-kapila, line 136: I think we can export this variable from the Importer file and then use it here. In case we change the value in the future, this might prevent tests from breaking.

listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
```diff
@@ -187,9 +195,18 @@ describe("getPage", () => {
         status: 404,
       });
     });
+    const getPageSpy = jest.spyOn(instance, "getPage");
+    await instance.getPage(1, RETRIES);
+
+    // we run the timer sufficient number of timers to ensure retries are not exceeded or undetected due to timeouts
+    for (let i = 0; i < 10; i += 1) {
```
shivam-kapila, line 202: If we import the RETRIES, then we can also update i < RETRIES + 2.

shivam-kapila (commented):
LGTM overall. No changes required as such. Just a few observations I think can solidify the tests.

listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
```diff
@@ -172,11 +172,18 @@ describe("getPage", () => {
...
-    jest.runAllTimers();
+    // we run the timer sufficient number of times to ensure retries are not exceeded or undetected due to timeouts
+    for (let i = 0; i < LASTFM_RETRIES + 2; i += 1) {
+      jest.runAllTimers();
+      // make timers and promises play well with jest
+      // eslint-disable-next-line no-await-in-loop
+      await Promise.resolve();
+    }
+
+    expect(getPageSpy).toHaveBeenCalledTimes(4);
```
alastair, line 186: what's this 4? Is it LASTFM_RETRIES + 2 ? either add a comment explaining how it's computed, or use a constant

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -151,7 +150,7 @@ export default class LastFmImporter extends React.Component<
     }
   }
 
-  async getPage(page: number) {
+  async getPage(page: number, retries: number) {
     /*
      * Fetch page from Last.fm
```
alastair, line 155: arg documentation, what is page and retries.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -162,7 +161,9 @@ export default class LastFmImporter extends React.Component<
       console.warn(
         `${reason} while fetching last.fm page=${page}, retrying in 3s`
```
alastair, line 162: would be nice to make this comment and the 3000ms timer in the next line read from the same variable so that we can adjust it as needed

alastair (approved):
Functionality seems sane, though I've not tested it myself.

To confirm the expected functionality:
Before this PR, we just repeatedly retried on http 5xx error. We never retried on 4xx error.
With this change, we now only retry LASTFM_RETRIES times if there is a 5xx page

listenbrainz/webserver/static/js/src/LastFMImporter.test.tsx
amCap1712: `1 + LASTFM_RETRIES`, fixed.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
amCap1712: Done.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -151,18 +150,24 @@ export default class LastFmImporter extends React.Component<
...
+      const timeout = 3000;
       console.warn(
-        `${reason} while fetching last.fm page=${page}, retrying in 3s`
+        `${reason} while fetching last.fm page=${page}, retrying in 
+        ${timeout / 1000}s`
       );
-      setTimeout(() => this.getPage(page), 3000);
+      if (retries > 0) {
+        setTimeout(() => this.getPage(page, retries - 1), timeout);
+      }
```
MonkeyDo, line 170: Is there any indication to the user that we stopped retrying and that it's just not working, if we get to the max number of retries?

MonkeyDo (approved):
The logic looks good, approving for speedy deployment.

One comment about the user-facing side of things, though.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
amCap1712: There is some logic here https://github.com/metabrainz/listenbrainz-server/blob/a85eb2c7c003f91d14e12e016533f73cdde9def7/listenbrainz/webserver/static/js/src/LastFMImporter.tsx#L335, but I do not fully understand how this works. 

I think adding an else to this which raises an error could work?

MonkeyDo: That would probably work.

The thrown error would be caught [here](https://github.com/metabrainz/listenbrainz-server/blob/a85eb2c7c003f91d14e12e016533f73cdde9def7/listenbrainz/webserver/static/js/src/LastFMImporter.tsx#L200), so you may need to modify the logic of that catch block to 
1. ensure we're not just retrying over and over and 
2. pass a clear error message that can be displayed to the end user

And finally, you'll want to wrap [the initial `getPage` call](https://github.com/metabrainz/listenbrainz-server/blob/a85eb2c7c003f91d14e12e016533f73cdde9def7/listenbrainz/webserver/static/js/src/LastFMImporter.tsx#L279) in a try-catch block and get the message to display in the catch.

Does that make sense to you?
I propose doing all this in another PR, to keep this current one from getting bigger.

amCap1712: That makes sense. Sure, let's take that up in a different PR.

MonkeyDo (changes requested):
I missed a spot!

Unless I'm mistaken, whenever the `retry` function is called, it should be `await`ed and we should be returning its content.
For example:
`retry(`Got ${response.status}`);` -> `return await retry(`Got ${response.status}`);`

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -196,16 +188,36 @@ export default class LastFmImporter extends React.Component<
         this.countReceived += payload.length;
         return payload;
       }
-      if (/^5/.test(response.status.toString())) {
-        return await retry(`Got ${response.status}`);
+      // ignore 40x errors
+      if (!/^4/.test(response.status.toString())) {
```
alastair, line 192: this test is a little broader than what it used to be. From what I understand, `response.ok` covers 2xx codes. We used to trigger on 5xx and skip everything else. Now we skip on 4xx and trigger on everything else (which could maybe include 3xx? 
I guess there's no 6xx, so it's not really an issue, but it's a bit weird to see a check using `4` when really it's a handler for errors starting with `5`

MonkeyDo: I adapted it to follow the description of one of the tests: "it should retry if there is any other error".
But that is indeed a different mechanism and possibly not what we want to do.
Let me take a closer look at the LFM API.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -196,16 +188,36 @@ export default class LastFmImporter extends React.Component<
...
     } catch {
       // Retry if there is a network error
-      await retry("Network error");
+      // eslint-disable-next-line no-console
+      console.warn(
+        `Network error while fetching last.fm page=${page}, retrying in
+        ${timeout / 1000}s`
+      );
+      if (retries <= 0) {
+        return null;
```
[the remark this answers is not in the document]
MonkeyDo: Not a super common thing, and sub-ideal, but I couldn't make throwing an error work with this particular setup.
I was a bit tired so let me try again and see if I can do it cleaner.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -196,16 +188,36 @@ export default class LastFmImporter extends React.Component<
...
+      }
+      await new Promise((resolve) => setTimeout(resolve, timeout));
```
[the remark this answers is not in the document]
MonkeyDo: Correct. Sort of an inline setTimeout for the async/await age.

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
MonkeyDo: I did it in here anyway since I was revamping the whole code anyway…

MonkeyDo (approved):
Silly to approve myself, but I've done some extra testing and it all seems to work, and the code is a lot neater now, plus extra tests.

PR #1340, opened 2021-03-13
LB-774: Fix Sphinx Documentation Warnings

dpmittal:
# Problem

Many Sphinx warnings were coming up while building the documentation locally which needed to be fixed as shown in https://tickets.metabrainz.org/browse/LB-774

# Solution

The warnings were coming up due to improper indentation in JSON code-block examples and some other bugs in documentation.

# Action

1. In this file `docs/dev/spark-devel-env.rst`:- The `info` wasn't a proper directive so change it to `seealso` as per my comprehension but yes directives like `note` can be used.
2.  In the toctree in file `docs/index.rst`:- I have just added the left out rst files at the end but might have to change the ordering of the index.

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -256,10 +256,12 @@ def get_similar_users(user_name):
     Get list of users who have similar music tastes (based on their listen history)
     for a given user. Returns an array of dicts like these:
 
-    {
-      "user_name": "hwnrwx",
-      "similarity": 0.1938480256
-    }
+    .. code-block:: json
+
+        {
+            "user_name": "hwnrwx",
+            "similarity": 0.1938480256
+        }
 
     :param user_name: the MusicBrainz ID of the user whose similar users are being requested.
     :statuscode 200: Yay, you have data!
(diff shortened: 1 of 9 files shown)


listenbrainz/webserver/views/api.py
```diff
@@ -256,10 +256,12 @@ def get_similar_users(user_name):
...
-    {
-      "user_name": "hwnrwx",
-      "similarity": 0.1938480256
-    }
+    .. code-block:: json
+
+        {
+        "user_name": "hwnrwx",
+        "similarity": 0.1938480256
+        }
```
shivam-kapila, line 264:
```suggestion
        {
            "user_name": "hwnrwx",
            "similarity": 0.1938480256
        }
```
Please update the indentation for all other jsos too.

shivam-kapila (changes requested):

alastair:
@brainzbot ok to test

alastair:
thanks for the changes @dpmittal, it's great to see no warnings!

I think that we should use `..note` in spark-devel-env instead of seealso.

Is it possible to make develop-sh.rst appear as a sub-section of `devel-env.rst` or would that involve moving the contents from one file to the other? At the very least we should group them together so that "Using develop.sh" comes after "Set up ListenBrainz Server development environment"

alastair:
I made the changes that I suggested. I was able to move develop-sh into the page for setting up the development environment. thanks again

dpmittal:
Thanks a lot, @alastair for reviewing and merging the changes :D Apologies, actually I just saw your suggestions but till then you had committed them

PR #1343, opened 2021-03-15
Add sentry reporting to react

amCap1712:
Adding the sentry integration to react and frontend will provide more visibility into errors that occur on the frontend.

This means the DSN for the associated project will become public as it is passed to `react-props`. According to [sentry docs](https://docs.sentry.io/product/sentry-basics/dsn-explainer/#dsn-utilization), this should be fine.

We might also want to add `Sentry.catchException` blocks manually at places where we just catch and log the error.


listenbrainz/config.py.sample
```diff
@@ -89,11 +89,11 @@ PLAYING_NOW_MAX_DURATION = 10 * 60
 #    'level': 'ERROR',
 # }
 
-# LOG_SENTRY = {
-#    'dsn':'',
-#    'environment': 'development',
-#    'level': 'ERROR',
-# }
+LOG_SENTRY = {
+    'dsn':'https://examplePublicKey@o0.ingest.sentry.io/0',
```
alastair, line 93: Is this a real test dsn that people can use? 
I'd prefer that during local development we don't configure a DSN, and only enable it in js if the value is non-null.

listenbrainz/webserver/views/index.py
```diff
@@ -155,6 +155,7 @@ def recent_listens():
         "mode": "recent",
         "spotify": spotify_user,
         "api_url": current_app.config["API_URL"],
+        "sentry_dsn": current_app.config["LOG_SENTRY"]["dsn"]
```
alastair, line 158: We need to add this to all pages, so let's use a flask context processor. We already have one for some static files processing, so we should make it more generic: https://github.com/metabrainz/listenbrainz-server/blob/f6b02d9c63/listenbrainz/webserver/__init__.py#L154

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -530,7 +531,10 @@ document.addEventListener("DOMContentLoaded", () => {
   const domContainer = document.querySelector("#react-container");
   const propsElement = document.getElementById("react-props");
   const reactProps = JSON.parse(propsElement!.innerHTML);
-  const { api_url, current_user, spotify, events } = reactProps;
+  const { api_url, current_user, spotify, events, sentry_dsn } = reactProps;
```
alastair, line 534: do we have types configured for these reactProps? @MonkeyDo

listenbrainz/webserver/views/index.py
alastair: ah, I realised that these are set in props instead of the page template variables. Any reason to do it here instead? I guess in the js we don't have the template processor, so we can't substitute the value in, is that right?
Perhaps we could render a global app props in addition to local page props.

amCap1712: As far as I know, we don't have the template processor in the react side. The props are passed like [this](https://github.com/metabrainz/listenbrainz-server/blob/afb3d43aeaf916aaca059c758571498e0893d497/listenbrainz/webserver/templates/base-react.html#L10) to the React. We could perhaps create a `react-global-props` here and pass some things like the dsn (maybe the api url as well) using the context processor. In the react side, we then use that variable instead of `reactProps`.

listenbrainz/config.py.sample
amCap1712: I copied this DSN from sentry docs but guarding the configuration seems fine to me will do that.

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
MonkeyDo: Negative.
I guess it comes right back to our experiment with exporting TS types from Pydantic models…

listenbrainz/webserver/views/index.py
MonkeyDo: 100% agree with the need for global props. Might be better to refactor that in another PR though.

MonkeyDo (commented):
Looking at the sentry/react docs, (that package looks pretty amazing, loving the easy set-up), it looks like we could be using their provided ErrorBoundary to replace the custom one we currently use. Did you have a look at that at all?
https://docs.sentry.io/platforms/javascript/guides/react/components/errorboundary/

```
import * as Sentry from "@sentry/react";
...
<Sentry.ErrorBoundary fallback={"An error has occurred"}>
  <Example />
</Sentry.ErrorBoundary>;
```

amCap1712:
Yes, I did. I considered to add that but first wanted to confirm with you whether its a drop-in replacement.

MonkeyDo:
> Yes, I did. I considered to add that but first wanted to confirm with you whether its a drop-in replacement.

I think we might need to change a thing or two in our existing [ErrorBoundary](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/ErrorBoundary.tsx) custom component, such as removing the `componentDidCatch` and instead getting the error message from Sentry.ErrorBoundary props passed to the fallback.

Probably not a huge amount of modifications; I'm not clear on what the benefits are though, so it could also be that it's not needed or doesn't bring us anything more.
Maybe scope for another PR, once we have this PR deployed and have a point of comparison?

amCap1712:
We also need to look into unbundling the sentry library from the .js files or perhaps lazy-loading it.

alastair:
@amCap1712 what's the state of this?
 - we need to add global application-level props, but we'll do this as a separate PR
 - we need to add error boundaries, but we'll do this as a separate PR
 - we have to work out how to unbundle sentry.js from the main bundle (this PR, or another one?)
 - other than that, it's good?

MonkeyDo:
The previous commit is just solving merging conflicts in preparation for merging.

Instead of lazy-loading, I'll look into loading separate dependency bundles with Webpack (in a separate PR)

PR #1344, opened 2021-03-15
Add a notification that similar users imported or failed

mayhem:
Add a simple email letting the folks on the LB observability email about similar users being imported or the import having failed. Output should look like:

Similar users were received by the spark consumer
and are being written into the database.

6511 users had similar users calculated with an average
of 48.4 being generated per user.

### DIFF
--- a/listenbrainz/db/similar_users.py
+++ b/listenbrainz/db/similar_users.py
@@ -15,10 +15,10 @@ ROWS_PER_BATCH = 1000
 
 def import_user_similarities(data):
     """ Import the user similarities into the DB by inserting the data into a new table
-        and then rotating the table into place atomically. 
+        and then rotating the table into place atomically.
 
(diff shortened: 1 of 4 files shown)


listenbrainz/db/similar_users.py
```diff
@@ -75,7 +79,7 @@ def import_user_similarities(data):
         conn.rollback()
         current_app.logger.error(
             "Error: Cannot correlate user similarity user name: %s" % str(err))
-        return
+        return (0, 0.0, "Error: Cannot correlate user similarity user name: %s" % str(err))
```
paramsingh, line 82: Can we add a docstring (or type annotations) for this return type.

paramsingh (commented):
lgtm, one small comment.

amCap1712 (approved):

PR #1345, opened 2021-03-16
LB-838: Limit Error Messages

Solai-Adithya:
# Problem
Presently there are issues with YouTube playback and that generates a series of errors. This spams the user with error messages. These messages can be limited. [LB838 - Ticket link](https://tickets.metabrainz.org/projects/LB/issues/LB-838?filter=myopenissues)

# Solution
After discussing with others over IRC, I have implemented a basic change. If another error message with same type, title and message is encountered, the original error message is concatenated with the count of the error message. 

### Screen Recording:

[Screen Recording](https://user-images.githubusercontent.com/56514792/111285255-f4fdec00-8666-11eb-982c-cfa8e413b2a0.mp4)

# Some Clarifications
- I maintain count to be an optional argument. So that existing code at many places wouldn't have to be changed (The places where we trigger errors). Would this be fine?
- What I wanted to implement is that if it times out, then decrease error count by 1, if dismissed by user then close entire message but this seemed complex to implement because we are using a library for this notification. 
- I was not sure if variable name 'alertsList' is appropriate. The list temporarily stores the lists of alerts, I edit this array and use this to set the state. If there are any suggestions for this variable name, I'll implement it.

### DIFF
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -225,16 +225,37 @@ export default class RecentListens extends React.Component<
   newAlert = (
     type: AlertType,
     title: string,
-    message: string | JSX.Element
+    message: string | JSX.Element,
+    count?: number
   ): void => {
     const newAlert: Alert = {
       id: new Date().getTime(),
       type,
       headline: title,
       message,
+      count,
     };
 
     this.setState((prevState) => {
+      const alertsList = prevState.alerts;
+      for (let i = 0; i < alertsList.length; i += 1) {
+        const item = alertsList[i];
+        if (
+          item.type === newAlert.type &&
+          item.headline.includes(newAlert.headline) &&
+          item.message === newAlert.message
+        ) {
+          if (!alertsList[i].count) {
+            // If the count attribute is undefined, then Initializing it as 2
+            alertsList[i].count = 2;
+          } else {
+            alertsList[i].count! += 1;
+          }
+          alertsList[i].headline = `${newAlert.headline} (${alertsList[i]
+            .count!})`;
+          return { alerts: alertsList };
+        }
+      }
       return {
         alerts: [...prevState.alerts, newAlert],
       };
--- a/listenbrainz/webserver/static/js/src/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/RecentListens.tsx
@@ -242,7 +242,7 @@ export default class RecentListens extends React.Component<
         const item = alertsList[i];
         if (
           item.type === newAlert.type &&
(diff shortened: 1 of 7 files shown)


MonkeyDo:
@brainzbot add to whitelist

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -225,16 +225,37 @@ export default class RecentListens extends React.Component<
...
+      count,
     };
 
     this.setState((prevState) => {
+      const alertsList = prevState.alerts;
+      for (let i = 0; i < alertsList.length; i += 1) {
+        const item = alertsList[i];
+        if (
+          item.type === newAlert.type &&
+          item.headline.includes(newAlert.headline) &&
```
MonkeyDo, line 245: Instead of using `Array.include` here (and in the other components) I suggest `String.startsWith` instead like so:
`item.headline.startsWith(newAlert.headline)`

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -225,16 +225,37 @@ export default class RecentListens extends React.Component<
...
+          item.message === newAlert.message
+        ) {
+          if (!alertsList[i].count) {
+            // If the count attribute is undefined, then Initializing it as 2
+            alertsList[i].count = 2;
+          } else {
+            alertsList[i].count! += 1;
```
MonkeyDo, line 252: Nice use of the non-null operator here ! 👏

MonkeyDo (changes requested):
Thanks for opening this PR !
Looks quite good; I've got one small comment.

This issue has been bugging me for a while now, and I'm really happy to see it resolved !

It's a shame we are copying this alerts code all over, but I've got plans to refactor it into a separate component. 
But that'll be for another PR.

listenbrainz/webserver/static/js/src/RecentListens.tsx
Solai-Adithya: Yeah, I got an ESLint error here. I googled a bit and found this, felt satisfying.

listenbrainz/webserver/static/js/src/RecentListens.tsx
Solai-Adithya: Great suggestion, I'll implement this.

paramsingh:
@brainzbot add to whitelist

MonkeyDo (approved):
That looks great, thanks a lot !

MonkeyDo:
No-one should have to modify code like that across 5 files, I created a ticket so that we end up refactoring all this and avoiding the duplicate code: https://tickets.metabrainz.org/browse/LB-843

PR #1347, opened 2021-03-16
Dump user feedback

mayhem:
The spark cluster needs to get period user and recommendation feedback data and this PR adds the functionality to dump a feedback dump and the rsync it to the FTP server. I anticipate that we'll add more spark data dumps in the future, so its good to have separate FTP dir for it.

Also in this PR is the change to keep 30 incremental feedback items. This gives users a complete month to catch up on incremental dumps -- 6 dumps was not enough for this.


mayhem:
@brainzbot test again please

paramsingh:
Thinking out loud, would the recording_recommendation events in the user_timeline_event table be useful wherever this dump is gonna be used?

mayhem:
Eventually yes. But realistically we're not there yet -- but with this in mind I created the spark FTP dir so that we can create more "out of replication sequence" dumps timed to be fresh for spark ingestion.

listenbrainz/db/dump.py
```diff
@@ -301,6 +349,96 @@ def create_public_dump(location, dump_time, threads=DUMP_DEFAULT_THREAD_COUNT):
...
+                   EXTRACT(DAY FROM r.created) AS day
+              FROM recording_feedback r
+              JOIN "user"
+                ON r.user_id = "user".id
+          ORDER BY created"""))
+
+        last_day = ()
+        todays_items = []
+
+        while True:
```
amCap1712, line 384: There's some scope of de-duplicating the code here. The `while` loop can probably be extraced into a separate method of its own receiving `key` names like `recording_msid` or `mb_recording_id` so on.

mayhem: Yeah, I see what you mean, but doing so would complicate the code and make it less readable in order to avoid duplicating a few lines of code. I would much prefer to keep it as is.

listenbrainz/db/dump.py
```diff
@@ -226,38 +269,43 @@ def _create_dump(location, dump_type, tables, dump_time, threads=DUMP_DEFAULT_TH
...
-                            current_app.logger.error('Error while copying table %s: %s', table, str(e), exc_info=True)
-                            raise
-                    transaction.rollback()
-
-
-            tar.add(archive_tables_dir, arcname=os.path.join(archive_name, 'lbdump'.format(dump_type)))
+                if dump_type == "feedback":
+                    dump_user_feedback(connection, location=archive_tables_dir)
+                else:
+                    with connection.begin() as transaction:
```
alastair, line 287: if we get one more dump type here I'd like to abstract this section out into a separate file too like you've done with `dump_user_feedback`, to aid in readability. For now it's probably OK though

listenbrainz/db/dump.py
```diff
@@ -301,6 +349,96 @@ def create_public_dump(location, dump_time, threads=DUMP_DEFAULT_THREAD_COUNT):
...
+        todays_items = []
+
+        while True:
+            row = result.fetchone()
+            today = (row[4], row[5], row[6]) if row else ()
+            if (not row or today != last_day) and len(todays_items) > 0:
+                full_path = os.path.join(location, "feedback", "listens", "%02d" % int(last_day[0]),
+                                         "%02d" % int(last_day[1]), "%02d" % int(last_day[2]))
+                os.makedirs(full_path)
+                with open(os.path.join(full_path, "data.json"), "wb") as f:
```
alastair, line 391: small thing - apparently this concept of writing multiple lines of json to a single file has started to be called "jsonl": https://jsonlines.org/, perhaps for clarity we could name this file `data.jsonl` instead

listenbrainz/db/dump.py
```diff
@@ -301,6 +349,96 @@ def create_public_dump(location, dump_time, threads=DUMP_DEFAULT_THREAD_COUNT):
...
+              JOIN "user"
+                ON r.user_id = "user".id
+          ORDER BY created"""))
+
+        last_day = ()
+        todays_items = []
+
+        while True:
+            row = result.fetchone()
+            today = (row[4], row[5], row[6]) if row else ()
```
alastair, line 386: sqlalchemy should turn a date field in postgres into a python datetime (or date) here, allowing us to use `thedate.year` in order to understand what this is doing more clearly. Are you doing this to avoid the conversion on the assumption that it'll be faster?

If it's significantly faster and you'd prefer to keep it like this, please use dictionary access into the row: `row['month']` in order to make it clearer what rows you're using and to avoid bugs if we update the select statement

listenbrainz/db/dump_manager.py
```diff
@@ -260,13 +335,32 @@ def _cleanup_dumps(location):
...
-    dump_files = [x for x in os.listdir(location) if incremental_dump_re.match(x)]
-    incremental_dumps = [x for x in sorted(dump_files, key=get_dump_id, reverse=True)]
+    # Clean up incremental dumps
+    incremental_dump_re = re.compile(
+        'listenbrainz-dump-[0-9]*-[0-9]*-[0-9]*-incremental')
+    dump_files = [x for x in os.listdir(
+        location) if incremental_dump_re.match(x)]
+    incremental_dumps = [x for x in sorted(
+        dump_files, key=get_dump_id, reverse=True)]
+    print(incremental_dumps)
```
alastair, line 345: is this print superfluous?

listenbrainz/db/dump_manager.py
```diff
@@ -260,13 +335,32 @@ def _cleanup_dumps(location):
     else:
         remove_dumps(location, full_dumps, NUMBER_OF_FULL_DUMPS_TO_KEEP)
 
-    incremental_dump_re = re.compile('listenbrainz-dump-[0-9]*-[0-9]*-[0-9]*-incremental')
-    dump_files = [x for x in os.listdir(location) if incremental_dump_re.match(x)]
-    incremental_dumps = [x for x in sorted(dump_files, key=get_dump_id, reverse=True)]
+    # Clean up incremental dumps
+    incremental_dump_re = re.compile(
+        'listenbrainz-dump-[0-9]*-[0-9]*-[0-9]*-incremental')
+    dump_files = [x for x in os.listdir(
```
alastair, line 341: could do `os.listdir` once at the beginning of this cleanup, but probably not computationally expensive given the number of items in this directory

listenbrainz/db/dump_manager.py
```diff
@@ -182,9 +200,57 @@ def create_incremental(location, threads, dump_id):
...
+        try:
+            write_hashes(dump_path)
+        except IOError as e:
+            current_app.logger.error(
+                'Unable to create hash files! Error: %s', str(e), exc_info=True)
+            sys.exit(-1)
+
+        try:
+            if not sanity_check_dumps(dump_path, 3):
+                return sys.exit(-1)
```
alastair, line 240: sys.exit doesn't return anything

listenbrainz/db/dump_manager.py
```diff
@@ -182,9 +200,57 @@ def create_incremental(location, threads, dump_id):
...
+
+        try:
+            write_hashes(dump_path)
+        except IOError as e:
+            current_app.logger.error(
+                'Unable to create hash files! Error: %s', str(e), exc_info=True)
+            sys.exit(-1)
+
+        try:
+            if not sanity_check_dumps(dump_path, 3):
```
alastair, line 239: it's difficult to understand what the value of the second argument is here, can we rename it to something like `check_expected_number_of_dumps` or make the expeted_count argument required, so that when reading this code we can understand its behavour?

listenbrainz/db/dump.py
mayhem: then I think we should update all of our dumps in on go, not just one being in consistent from another. I would say we do this when we get rid of MSIDs from spark, which will basically be a complete refactor of the CF spark code. That would be a good time to introduce this change.

listenbrainz/db/dump_manager.py
mayhem: Its trivial and everything is going to be in disk cache anyways.

listenbrainz/db/dump_manager.py
mayhem: expected_count is required and documented. I don't think it needs a change.

admin/create-dumps.sh
```diff
@@ -58,7 +58,11 @@ DUMP_TYPE="${1:-full}"
 if [ "$DUMP_TYPE" == "full" ]; then
     SUB_DIR="fullexport"
 else
-    SUB_DIR="incremental"
+    if [ "$DUMP_TYPE" == "incremental" ]; then
```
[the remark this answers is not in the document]
mayhem: Fixed

admin/create-dumps.sh
```diff
@@ -58,7 +58,11 @@ DUMP_TYPE="${1:-full}"
...
+        SUB_DIR="incremental"
+    else
+        SUB_DIR="spark"
```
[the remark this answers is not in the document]
mayhem: Fixed.

alastair (approved):
thanks

PR #1349, opened 2021-03-17
Fix charts and reports pages for old browsers

MonkeyDo:
Some of our dependencies of our graph libraries do not generate ES2015 code, therefore breaking the page completely on older versions of Firefox (< v.45) and IE.
Reference:
https://github.com/d3/d3-scale/releases/tag/v3.0.0
https://github.com/d3/d3-scale/issues/225
(related, although from another library altogether) https://github.com/swimlane/ngx-charts/issues/1369

This commit adds those libraries I found that need to be transpiled (there may be others) to our babel workflow in Webpack.
Dependencies are usually expected to be ES2015 and the node_modules folder is ignored by babel by default.

One other option was to use this utility that automatically detects libraries that are compiled to ES2015 and adds them to the workflow, but that didn't work so well for me when I tried: https://andersdjohnson.github.io/webpack-babel-env-deps/

I also added core-js as a dependency and set it up to transpile other parts of our code that were breaking on older browsers. A recent version of core-js fixes a specific issue with iterators on FF<45: https://github.com/zloirock/core-js/releases/tag/v3.8.3

This only affects the compiled javascript output, nothing in the codebase itself.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4247,9 +4247,9 @@
       "dev": true
     },
     "core-js": {
-      "version": "1.2.7",
-      "resolved": "https://registry.npmjs.org/core-js/-/core-js-1.2.7.tgz",
-      "integrity": "sha1-ZSKUwUZR2yj6k70tX/KYOk8IxjY="
+      "version": "3.9.1",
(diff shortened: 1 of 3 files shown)


MonkeyDo:
@brainzbot retest this please

PR #1350, opened 2021-03-17
Delete items from .gitignore from previous tech stacks

alastair:
We've removed vagrant, kitchen, and bigquery so remove their config
files from gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,10 @@
-.kitchen
-.vagrant
 .DS_Store
 *.pyc
 *.swp
 config.py
-custom_config.py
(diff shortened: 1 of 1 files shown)


mayhem (approved):
I can't possibly approve this -- so many further changes require to bring this up to scratch!

mayhem:
Oh and we have to wait for those tests to wait -- we'll have to see how many break because of this far reaching change!

alastair:
what we'll find is that someone still has a custom_config.py lying around from years ago and when they do `git add .` they manage to add it

PR #1351, opened 2021-03-18
Add missing api docs

amCap1712:
Some of the api endpoints are absent from the api docs at [listenbrainz.readthedocs.io](https://listenbrainz.readthedocs.io/en/production/dev/api/). This is because we need to manually add new api blueprints to [api.rst](https://github.com/metabrainz/listenbrainz-server/blob/master/docs/dev/api.rst). This is only required when we use a new blueprint. Adding a new endpoint to a blueprint that is already documented (present in **api.rst**) does not need any extra changes.

We have also decided to use `code-block` directives for specifying json blocks in the documentation.

I think these details are generally useful in coding/reviewing apis and will add these to the _checklist_ (to be created in the wiki).


docs/dev/api.rst
```diff
@@ -66,6 +69,46 @@ The playlists API allows for the creation and editing of lists of recordings
    :include-empty-docstring:
    :undoc-static:
 
+Feedback API Endpoints
+^^^^^^^^^^^^^^^^^^^^^^
+These API endpoints allow to submit and retrieve feedback for a user's recordings
```
alastair, line 74:
```suggestion
These API endpoints allow to submit and retrieve feedback recordings that a user has listened to
```

docs/dev/api.rst
```diff
@@ -66,6 +69,46 @@ The playlists API allows for the creation and editing of lists of recordings
...
+
+.. autoflask:: listenbrainz.webserver:create_app_rtfd()
+   :blueprints: feedback_api_v1
+   :include-empty-docstring:
+   :undoc-static:
+
+Recording Recommendation API Endpoints
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+ListenBrainz uses collaborative filtering to generate recording recommendations,
+which may be further processed to generate playlists for users. These api endpoints
```
alastair, line 84: I think we shouldn't specifically say that we use collaborative filtering here, or that we "might" generate playlists. Just say that we have a recommendation system that can generate similar recordings to those that a user has listened to. If we want to in the future we could add some pages to the developers section of the docs that explain how this works

docs/dev/api.rst
```diff
@@ -66,6 +69,46 @@ The playlists API allows for the creation and editing of lists of recordings
...
+
+.. autoflask:: listenbrainz.webserver:create_app_rtfd()
+   :blueprints: recommendation_feedback_api_v1
+   :include-empty-docstring:
+   :undoc-static:
+
+Missing MusicBrainz Data API Endpoints
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+These api endpoints allow to retrieve data which the user has submitted to
+ListenBrainz but is missing from MusicBrainz.
```
alastair, line 106: I don't understand what this API does. Is it things that are _missing_ from MusicBrainz, or things that we've not successfully managed to link to MusicBrainz? What is the way that this is added, and why would a user want to access this data?

alastair (commented):
Thanks for cleaning this up. I have a few questions about some of the new endpoints. It's not clear if we need to document these for users, or if they should stay internal.

docs/dev/api.rst
amCap1712: This is generated during matching the listens with the MBID mapping. The listens could not be matched that are used to create this data. The intent to add this data was for displaying it in some form on LB website so as to nudge users to add this data to MB.

Speaking in general regardless of whether the API is useful to users or not, I think putting it on the docs adds a little to the convenience of developers. I say this because docs are rendered in a more human readable form on readthedocs than the source.

In case, these endpoints adds noise for users, we should probably consider adding another different section to docs.

docs/dev/api.rst
amCap1712: The endpoint uses collaborative filtering in its name so I think the intent might be to use different endpoints for different sources of recommendations in the future. I think In that case, it would be clearer if we mention CF here. also, @mayhem might have suggestions regarding this.

alastair: good point, if the endpoint has CF in the name then explaining its use here is fine.

docs/dev/api.rst
alastair: yeah, I'd consider adding this to a separate section not directly on the core submit/get documentation.

alastair (approved):
other than the comment about splitting the "Missing MusicBrainz Data" section to another page, this is fine

PR #1352, opened 2021-03-19
LB-843: Refactor alert notifications as higher-order component

MonkeyDo:
Ticket: LB-843
The code for alert notifications is repeated on each fo the 5 main pages we currently have, making it every developer's nightmare.

This PR adds a new [higher-order component](https://reactjs.org/docs/higher-order-components.html) (in short a function that takes a component, enhances it and wraps it before returning it) to have a more modular approach.

The alerts/notifications state, logic and rendering now lives in one place and we can easily enhance a page to add notifications with a simple `const MyPageWithAlertNotifications = withAlertNotifications(MyPage);`

That also allows us to move the alert tests to their own separate file, and while we're at it add a bit more testing.

While I was in there refactoring everything, I also took the liberty of adding missing `ErrorBoundary` components around two of our pages.


listenbrainz/webserver/static/js/src/AlertNotificationsHOC.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  describe("newAlert", () => {
+    it("creates a new alert", () => {
+      const WrappedComponent = withAlertNotifications(fakeComponent);
+      const wrapper = shallow<React.ComponentType>(<WrappedComponent />);
+      const instance = wrapper.instance();
+
+      expect(wrapper.state().alerts).toEqual([]);
+
+      (instance as any).newAlert("warning", "Test", "foobar");
```
MonkeyDo, line 73: I can't seem to make typescript recognize that this component implements `AlertNotificationsI` and thus the `newAlert` and `onAlertDismissed` methods.
Ah well.

MonkeyDo:
@brainzbot retest this please

amCap1712:
I think LB-526 refers to this issue. If that's the case, let's update the title to link the PR in JIRA.

alastair (approved):
not sure I can give too much feedback on this... it removes a lot of code and makes it simpler to show alerts, so I guess that's a good thing?

MonkeyDo:
There are fantom failures of the unit and integration tests, but the PR doesn't change any of the code it test, so I'm merging before we open more PRs that will need rebasing.

PR #1356, opened 2021-03-24
LB-793: validate-token endpoint takes token as an argument instead of header

amCap1712:
LB-793

In all our other endpoints, we accept the token as a part of the `Authorization` header. However, the `validate-token` endpoint accepts it as a query parameter. To remove this inconsistency, the `validate-token` endpoint is now updated to accept the token as in the `Authorization` Header. If the header is missing, the query parameters are checked for the token for backward compatibility.

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -414,7 +414,8 @@ def validate_token():
     header = request.headers.get('Authorization')
     if header:
         auth_token = header.split(" ")[1]
-    else:  # for backward compatibility, check for auth token in query parameters as well
+    else:
+        # for backwards compatibility, check for auth token in query parameters as well
         auth_token = request.args.get('token', '')
 
     if not auth_token:
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -375,13 +375,19 @@ def latest_import():
 
 
 @api_bp.route('/validate-token', methods=['GET'])
+@crossdomain(headers='Authorization')
 @ratelimit()
 def validate_token():
     """
     Check whether a User Token is a valid entry in the database.
 
-    In order to query this endpoint, send a GET request with the token to check
-    as the `token` argument (example: /validate-token?token=token-to-check)
+    In order to query this endpoint, send a GET request with the token in
+    the Authorization header.
+
+    .. note::
+        - This endpoint also checks for `token` argument in query
+        params (example: /validate-token?token=token-to-check) if the
+        Authorization header is missing for backward compatibility.
 
     A JSON response, with the following format, will be returned.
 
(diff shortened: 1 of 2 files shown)


mayhem:
Can you please provide a description with each PR? The motivation for this PR is very important as we provide a public way for someone to brute force attack to find valid tokens. Not that this makes sense, but I would expect people to still abuse this.

amCap1712:
I understand the concern about a brute force attack. However, this endpoint is already in production for some time now. I do not know the initial motivation behind adding the endpoint. It can also be argued that if a request fails with a 401, the client would know that its because of an invalid token making this endpoint redundant.

amCap1712:
I found #444 and LB-377 under which the endpoint was initially added.

mayhem:
@alastair : This endpoint (which I didn't know existed and is in production) makes me really nervous. The bug mentioned was reported by a community member and implemented by a volunteer. I personally think it is a bad idea and that we should remove it. If someone needs to validate a token, they should do it on their own web page. We can always update error messages to indicate that users may need to check their auth token.

mayhem:
@alastair says:
>I don't see how this can facilitate a brute force attack any more than the submit listens endpoint can
> bad token: returns 403, good token: returns 200

Very good point, this endpoint does not increase our risk.

listenbrainz/webserver/views/api.py
```diff
@@ -375,13 +375,19 @@ def latest_import():
...
+@crossdomain(headers='Authorization')
 @ratelimit()
 def validate_token():
     """
     Check whether a User Token is a valid entry in the database.
 
-    In order to query this endpoint, send a GET request with the token to check
-    as the `token` argument (example: /validate-token?token=token-to-check)
+    In order to query this endpoint, send a GET request with the token in
+    the Authorization header.
```
alastair, line 385: Can we open a ticket to improve the documentation for authentication? We currently have this text in most endpoints:
>   In order to update the timestamp of a user, you'll have to provide a user token in the Authorization
>   Header. User tokens can be found on https://listenbrainz.org/profile/ .

It'd be nice to have a separate "authorization" section in the docs and link all endpoints to this.

listenbrainz/webserver/views/api.py
```diff
@@ -405,7 +411,12 @@ def validate_token():
     :statuscode 200: The user token is valid/invalid.
     :statuscode 400: No token was sent to the endpoint.
     """
-    auth_token = request.args.get('token', '')
+    header = request.headers.get('Authorization')
+    if header:
+        auth_token = header.split(" ")[1]
+    else:  # for backward compatibility, check for auth token in query parameters as well
```
alastair, line 417: nit: move the comment on to the next line.

alastair (approved):
fine by me, thanks for keeping backwards compatibility.

listenbrainz/webserver/views/api.py
amCap1712: Done. LB-861

PR #1357, opened 2021-03-24
Reorganize Dockerfile sequence of events

MonkeyDo:
We currently optimize our build for image size rather than build time.
After some discussion ruaok alastair and I agreed that we would rather optimize for build time by changing the order in which we run or Dockerfile operations.
In particular, installing NodeJS, copying and building front-end code was done in one command and we would remove the JS dependencies (node_modules) at the end of the command, saving the space of these dependencies in the final image, at the cost of having to reinstall Node and all the dependencies every time we build for prod, regardless of whether they have changed.

Instead, we move the installation of Node and of the dependencies much sooner in the image to have better caching and thus faster build times.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,19 +64,26 @@ RUN groupadd --gid 901 listenbrainz_stats_cron
 RUN useradd --create-home --shell /bin/bash --uid 901 --gid 901 listenbrainz_stats_cron
 RUN mkdir /logs && chown lbdumps:lbdumps /logs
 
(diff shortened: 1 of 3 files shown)


MonkeyDo:
I built and deployed to test.LB using this PR's Dockerfile, and everything seems to be built and running correctly.

mayhem (approved):
Looks as discussed -- thanks!

alastair (approved):
We worked on this in person in the office, looks as I remember it.

PR #1358, opened 2021-03-24
Upgrade brainzutils and integrate metrics endpoints

alastair:
# Problem

We sometimes want to count things

# Solution

Integrate the BU metrics module to let us count things!

In code, we can use these two methods to make metrics:
for things that count up:

    from brainzutils import metrics
    metrics.increment('foo')

for things that you want to set a value:

    metrics.set_count('messybrainz_match', artists=200, recordings=4432)

Then there are API endpoints available: 
`/internal/statistics` for _all_ metrics created by `increment`
Screenshot 2021-03-24 at 6 18 00 PM

`/internal/counter/[counter_name]` for metrics created by `set_count`
Screenshot 2021-03-24 at 6 19 11 PM

# Action

Set up telegraf/logging/monitoring/etc to show these stats somewhere

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -4,6 +4,7 @@ import sys
 from time import sleep
 from shutil import copyfile
 
+from brainzutils import metrics
 from brainzutils.flask import CustomFlask
 from flask import request, url_for, redirect
 from flask_login import current_user
@@ -101,6 +102,9 @@ def gen_app(config_path=None, debug=None):
 
     # Timescale connection
     create_timescale(app)
+
+    # Metrics
+    metrics.init('listenbrainz')
 
     # RabbitMQ connection
     try:
@@ -240,6 +244,9 @@ def _register_blueprints(app):
 
     from listenbrainz.webserver.views.api import api_bp
     app.register_blueprint(api_bp, url_prefix=API_PREFIX)
+
+    from listenbrainz.webserver.views.api_internal import api_internal
+    app.register_blueprint(api_internal, url_prefix='/internal')
 
     from listenbrainz.webserver.views.api_compat import api_bp as api_bp_compat
     app.register_blueprint(api_bp_compat)
--- a/listenbrainz/webserver/views/api_internal.py
+++ b/listenbrainz/webserver/views/api_internal.py
@@ -1,0 +1,18 @@
+from brainzutils import metrics
+from flask import Blueprint, jsonify
+
+# These views expose internal statistics for use in the MetaBrainz logging sytem
+# you're welcome to look at them youselves, but they are not supported outside
+# of our infrastructure
+
+api_internal = Blueprint('api_internal', __name__)
+
+
+@api_internal.route('/statistics')
+def statistics():
+    return jsonify(metrics.stats())
+
+
+@api_internal.route('/counter/<metric_name>')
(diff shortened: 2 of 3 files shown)


amCap1712 (approved):

mayhem (approved):
OK, given the context of our discussion yesterday, this makes sense. Let me chat with @zas about next steps.

mayhem:
As per convo with @zas, we should merge and deploy this, then point him to the working endpoint so he can configure telegraf.

mayhem:
As per our IRC discussion before you needed to leave, here is a suggested data structure:

https://gist.github.com/mayhem/4cc28dac0a66afdda7cd054b5eb6e327

listenbrainz/webserver/__init__.py
```diff
@@ -102,6 +103,9 @@ def gen_app(config_path=None, debug=None):
     # Timescale connection
     create_timescale(app)
 
+    # Metrics
+    metrics.init('listenbrainz')
```
alastair, line 107: Currently this is what we set `key` to, it'll the same for all containers. If we want this to be different for each service, we should init it here to the env variable CONTAINER_NAME, but we should also include an indicator that it's from listenbrainz: something like `init('listenbrainz', os.environ['CONTAINER_NAME'])` This would internally store that the second argument is a metric of listenbrainz, and the `/internal/stats` endpoint can return a list of all keys under the listenbrainz namespace.

Do we also want to add the hostname to this init?

listenbrainz/webserver/views/api_internal.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@api_internal.route('/statistics')
+def statistics():
+    return jsonify(metrics.stats())
+
+
+@api_internal.route('/counter/<metric_name>')
+def statistics_counter(metric_name):
+    return jsonify(metrics.stats_count(metric_name))
```
alastair, line 18: we should return incrementing counters (`metrics stats`) and groups of metrics (`stats_count`) in a single endpoint as a list of json, according to the [telegraf json arrays parsing](https://github.com/influxdata/telegraf/tree/master/plugins/parsers/json#arrays).
This means that we only need to configure one telegraf ingester for all LB statistics.

alastair (commented):
small proposed modification to the json format, based on my proposed changes to `metrics.init`:
```json
[
{"key": "listenbrainz:mbid-mapping-writer",
 "host": "lemmy",
 "num_users":  532,
 "mbid_matching:artists": 90,
 "mbid_matching:recordings": 360
},
{"key": "listenbrainz:web-prod",
 "host": "prince",
 "listen_count": 421,
 "user_count": 13
},
{"key": "listenbrainz:spotify-reader",
 "spotify_items_imported":  42321,
 "user_playlists_generated": 43
}
]

listenbrainz/webserver/__init__.py
mayhem: The host key should be set in each metric as per our discussion in IRC today. If this is the place to add it, then yes.

listenbrainz/webserver/views/api_internal.py
mayhem: makes sense.

alastair:
Updated proposed output, along with some informative comments at: https://gist.github.com/alastair/772a2fcd167fd373fb0a114147f38c16

PR #1342, opened 2021-03-15
Add api endpoint for approved bots to post on user's timeline

amCap1712:

PR #1359, opened 2021-03-25
Add support for 'notification' events on timeline UI

MonkeyDo:
Hand in hand with PR #1342 , which adds an endpoint to add notification events for a user, this PR adds support for displaying the notification events (with a bell icon) and sanitizing and displaying the html message.


amCap1712 (approved):
Awesome. I tested locally and it works great!

MonkeyDo:
I realized I forgot to choose a color for notifications (under the icon). Let me add that.

PR #1360, opened 2021-03-26
User Timeline: send notification events and potential bugfix

amCap1712:
In #1342, I had forgotten to update the `user_feed` endpoint to send notification events as well to the frontend. I realised this while testing #1359 locally. This PR in conjunction with the other two successfully displays the notification events in the feed.

Screenshot from 2021-03-26 13-41-12

During testing, I also found a potential bug. If I do not follow a user, no events are sent for my timeline. For example, I do not follow anyone and recommend a track to my followers, this event does not show up in my timeline.

(In the attached image, it does because it was taken after the fix but the problem is easily reproducible in production just unfollow all users and open your timeline.)

### DIFF
--- a/data/model/user_timeline_event.py
+++ b/data/model/user_timeline_event.py
@@ -42,7 +42,7 @@ class RecordingRecommendationMetadata(pydantic.BaseModel):
 
 
 class NotificationMetadata(pydantic.BaseModel):
-    creator_id: int
+    creator: str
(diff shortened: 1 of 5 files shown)


amCap1712:
@brainzbot retest this please

mayhem (approved):
Looks good. Thanks for the detailed description!

PR #1364, opened 2021-03-30
Pass API URL to FollowButton component

amCap1712:
FollowButton is not using the API URL to call follow and unfollow endpoints. As we have moved those endpoints to API now, we need to modify it to use the API URL. Instead of reading the props again in the FollowButton component, we pass it down the hierarchy of components.

### DIFF
--- a/listenbrainz/webserver/static/js/src/FollowButton.tsx
+++ b/listenbrainz/webserver/static/js/src/FollowButton.tsx
@@ -65,7 +65,9 @@ class FollowButton extends React.Component<
 
     const { apiUrl } = this.props;
 
-    this.APIService = new APIService(apiUrl);
+    this.APIService = new APIService(
+      props.apiUrl || `${window.location.origin}/1`
+    );
   }
 
   componentDidUpdate(prevProps: FollowButtonProps) {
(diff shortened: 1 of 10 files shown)


listenbrainz/webserver/static/js/src/FollowButton.tsx
```diff
@@ -62,7 +63,9 @@ class FollowButton extends React.Component<
       error: false,
     };
 
-    this.APIService = new APIService(`${window.location.origin}/1`);
+    const { apiUrl } = this.props;
+
+    this.APIService = new APIService(apiUrl);
```
shivam-kapila, line 68:
```suggestion
    this.APIService = new APIService(
      props.apiUrl || `${window.location.origin}/1`
    );
```

shivam-kapila: Just for the sake of consistency with other components.

shivam-kapila (changes requested):
LGTM. Just a small change for consistency among the components

listenbrainz/webserver/static/js/src/FollowButton.tsx
shivam-kapila: Ref: http://livegrep.metabrainz.org/search/livegrep?q=this.APIService%20%3D%20new%20APIService&fold_case=auto&regex=false&context=true&repo%5B%5D=metabrainz%2Flistenbrainz-server

shivam-kapila (approved):
The build passed, so ready to ship

PR #1365, opened 2021-03-30
Add not None checks at a few places

amCap1712:
Sentry has multiple different errors that occur due to performing some operation on `None`. This can easily be mitigated by adding checks for None as required.

### DIFF
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -31,6 +31,9 @@ def notify_error(musicbrainz_row_id, error):
         error (str): a description of the error encountered.
     """
     user_email = mb_editor.get_editor_by_id(musicbrainz_row_id)['email']
+    if not user_email:
+        return
+
     spotify_url = current_app.config['SERVER_ROOT_URL'] + '/profile/connect-spotify'
     text = render_template('emails/spotify_import_error.txt', error=error, link=spotify_url)
     send_mail(
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -5,9 +5,9 @@ import json
 
 from datetime import datetime
 from listenbrainz.utils import safely_import_config
-from listenbrainz.webserver.errors import APIBadRequest
-
 safely_import_config()
+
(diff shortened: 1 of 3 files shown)


listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -271,7 +274,7 @@ def process_one_user(user):
         listenbrainz_user = db_user.get(user.user_id)
 
         currently_playing = get_user_currently_playing(user)
-        if currently_playing is not None and 'item' in currently_playing:
+        if currently_playing is not None and 'item' in currently_playing and currently_playing['item'] is not None:
```
alastair, line 277: is this structure of `object['item']` something that comes from spotipy or from us?
If spotipy returns it, what is the situation where it returns just None instead of a dictionary? I wonder why we have such a complex check here. If spotipy (or our function) always returns a dictionary, perhaps we don't need all of the conditions. Also consider `currently_playing.get('item')` to return None if it's not there either. Let's look carefully into what happens if there's no currently playing instead of just adding more checks to this line

amCap1712: `currently_playing['item'` comes directly from Spotify. spotipy just forwards the response JSON to us. [Spotify docs](https://developer.spotify.com/documentation/web-api/reference/#category-player) mention that there are two playing types, `track` and `episode`. The default is `track` which we use. When the user's playback type is not `track`, that is `episode`. Spotify will still send an `item` field in the response but set it to null. Parsing this gives us the `None` here.

alastair: ok, therefore this check seems fine. In which case would currently_playing be None and not a dict? would be good to add a comment describing basically what you said here so that we understand why the check is like this.

amCap1712: No, that means `currently_playing['item'] can be `None` so checking for `item` in the `dict` is inadequate. Can `currently_playing` be `None`, I *think* yes. If no playback is found, Spotify returns an empty response which is odd. because it should be 204 in that case not 200. Perhaps, I didn't understand the docs for this case quite right. For now, let's get this fix out, I'll do some real testing with the Spotify API and figure this out.

amCap1712:
@brainzbot retest this please

PR #1366, opened 2021-03-31
LB-851: Calculate similar users only for users who have a minimum number of listens

amCap1712:
LB-851


mayhem (approved):
Looks good -- thank you!

PR #1367, opened 2021-03-31
Upgrade brainzutils and sentry-sdk

amCap1712:


alastair:
I still see a bunch of config items that start with `LOG_EMAIL`, especially the old style config items `LOG_EMAIL_ENABLED`

alastair:
consul config template defines `SENTRY_DSN_PUBLIC`, but this isn't used anywhere. Remove it, and change the consul key `sentry/dsn_private` to just `sentry/dsn` (we use the same public and private dsn in both server, and will use it in prod with #1343)

PR #1370, opened 2021-03-31
Decrease logging level when stats for deleted user are calculated

amCap1712:
The spark cluster can take upto 2 weeks to reflect deleted users. In the meantime, its not useful to log this error at critical since we can do nothing about it.


mayhem (approved):

mayhem:
@brainzbot please retest

PR #1371, opened 2021-04-02
Raise APIBadRequest if json invalid according to postgresql

amCap1712:
If the json submitted by users violates in invalid according the PostgreSQL database, it will raise a DataError. We should report this to the user as a BadRequest. I tried to add a test for this but I couldn't create a JSON file that reproduces this error locally.

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -13,6 +13,8 @@ import ujson
 import uuid
 
 from flask import current_app, request
+from sqlalchemy.exc import DataError
+
 from listenbrainz.listen import Listen
 from listenbrainz.webserver import API_LISTENED_AT_ALLOWED_SKEW
 from listenbrainz.webserver.external import messybrainz
@@ -40,6 +42,7 @@ LISTEN_TYPE_SINGLE = 1
 LISTEN_TYPE_IMPORT = 2
 LISTEN_TYPE_PLAYING_NOW = 3
 
+
 def insert_payload(payload, user, listen_type=LISTEN_TYPE_IMPORT):
     """ Convert the payload into augmented listens then submit them.
         Returns: augmented_listens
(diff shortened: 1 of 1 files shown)


mayhem (approved):

listenbrainz/webserver/views/api_tools.py
```diff
@@ -49,6 +52,8 @@ def insert_payload(payload, user, listen_type=LISTEN_TYPE_IMPORT):
         _send_listens_to_queue(listen_type, augmented_listens)
     except (APIInternalServerError, APIServiceUnavailable) as e:
```
alastair, line 53: picky: why do we catch this and then immediately raise it again without making a change? This is no different to removing these two lines

listenbrainz/webserver/views/api_tools.py
```diff
@@ -49,6 +52,8 @@ def insert_payload(payload, user, listen_type=LISTEN_TYPE_IMPORT):
...
         raise
+    except DataError:
+        raise APIBadRequest("Listen submission contains invalid characters.")
```
alastair, line 56: sorry to be picky here - are we sure that DataError is raised only if data is invalid? what does "invalid" mean? 
I see that we made a similar wishy-washy statement in AB, so maybe it's not a huge issue.

We should explicitly say in the documentation that null characters are not allowed. An alternate way to do this check would be to additionally have a check the moment that we load the POST data, and look for null characters, and if so return APIBadRequest immediately with an error message that explicitly mentions these bad characters

alastair (commented):
Good fix in general, but I think we can make a few small improvements

listenbrainz/webserver/views/api_tools.py
amCap1712: I would prefer to handle this after the insert fails. Sentry had ~50k of these messages even if we assume those were all unique (no retries) its hardly 0.1% of the submissions. Performing a check on all data to catch less than 0.1% sounds a bit overkill to me especially since the bad data is harmless.

I do agree with the DataError and message thing. I'll look further into it and see if we can improve it further.

listenbrainz/webserver/views/api_tools.py
amCap1712: Agreed. will remove.

PR #1372, opened 2021-04-02
Catch APIBadRequest and continue in Spotify Reader

amCap1712:
We do not need to report APIBadRequest to sentry as we cannot do anything about it. Just catch it and continue.

### DIFF
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -248,8 +248,8 @@ def parse_and_validate_spotify_plays(plays, listen_type):
         try:
             validate_listen(listen, listen_type)
             listens.append(listen)
-        except APIBadRequest as e:
-            current_app.logger.info(str(e))
+        except APIBadRequest:
+            pass
     return listens
 
 
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -5,6 +5,8 @@ import json
 
 from datetime import datetime
 from listenbrainz.utils import safely_import_config
+from listenbrainz.webserver.errors import APIBadRequest
+
 safely_import_config()
 
 from dateutil import parser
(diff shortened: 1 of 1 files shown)


listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -246,9 +248,8 @@ def parse_and_validate_spotify_plays(plays, listen_type):
         try:
             validate_listen(listen, listen_type)
             listens.append(listen)
-        except BadRequest as e:
-            current_app.logger.error(str(e))
-            raise
+        except APIBadRequest as e:
+            current_app.logger.info(str(e))
```
mayhem, line 252: this still logs the error and the raise is discarded. Should it be the other way round?

amCap1712: Its intended to be this way. We do not want to report APIBadRequest to sentry hence the logging level. I removed the raise because we want to ignore invalid listens. 

We do so because these listens are received from Spotify not users. As discussed in IRC, removed the logging as well.

mayhem (approved):
Fine after tests pass.

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -246,9 +248,8 @@ def parse_and_validate_spotify_plays(plays, listen_type):
         try:
             validate_listen(listen, listen_type)
             listens.append(listen)
-        except BadRequest as e:
-            current_app.logger.error(str(e))
-            raise
+        except APIBadRequest:
```
alastair, line 251: one thing about no longer logging this is that we have no way of tracking what bad listen data we're getting from spotify, in order to try and fix our parser.
I agree that putting it in sentry is not the right place for it, otherwise we end up with thousands of messages. However, I'm not sure if logging it to file is the right place either, because it could get very large very quickly.
It could be interesting to have a "probability logger", which has a 5% or 10% chance of logging an item. This allows us to glance at these bad items and try and fix them occasionally, without filling up our logs. Thoughts?

amCap1712: I agree that we may occasionally want to poke at the bad listen data to see if there are issues on our end. FWIW, I had looked into some of the errors which have currently occurred to see if there was some issue on our side. Currently, I don't think there is any but of course in future this might change. It might be useful to employ a probability based logger here. 

Also, in case a probability logger doesn't work out we could always report the error to sentry and just mark it as ignored. Ignored errors do not show up in the default view nor do they trigger an email. I am not sure if this is a good idea but it can be used if probability logger plan does not pan out. We do have a couple of such errors already in LB but I think those are that way way because there was no other way to discard errors (now we have `ignore_errors` in configuration).

PR #1373, opened 2021-04-02
Add similar users overview

mayhem:
This new view shows the top similar users who are basically spammers and causing the similarities between legit users to be quite low. This view should be instrumental for fighting spammers and having healthy user similarities.


amCap1712 (approved):

mayhem:
Thank you!

PR #1374, opened 2021-04-04
LB-827: Previous button on first page of Top Artists page disabled

Solai-Adithya:
# Problem
Minor bug fix: LB-827

# Solution
As suggested in the ticket, I have added a check for positive integers and have greyed out the previous button.
Screenshot from 2021-04-04 13-51-33


paramsingh (changes requested):
Can we add a test for this in https://github.com/metabrainz/listenbrainz-server/blob/792e3d68/listenbrainz/webserver/static/js/src/stats/UserEntityChart.test.tsx#L1 ?

listenbrainz/webserver/static/css/theme/boostrap/pager.less
```diff
@@ -46,7 +46,7 @@
     > a:hover,
     > a:focus,
     > span {
-      color: @pager-disabled-color;
+      color: @pager-disabled-color !important;
```
MonkeyDo, line 49: This won't be needed in the end, and `!important` rules in CSS make me feel a bit uncomfortable :)
I see the issue you were solving with this, and found the root cause of it that we can fix: the `<a>` tags in our navigation have an empty `href` attribute, which causes the `.disabled` style not to be applied.

We recently corrected similar links in #796 so what I suggest is changing those anchor with empty `href` in UserEntityChart.tsx into:

```
<a
  href={`?page=${Math.max(prevPage,1)}&range=${range}&entity=${entity}`}
  role="button"
  onClick={(event) => this.changePage(prevPage, event)}
>
```
                      
and
```
<a
  href={`?page=${Math.min(nextPage,totalPages)}&range=${range}&entity=${entity}`}
  role="button"
  onClick={(event) => this.changePage(nextPage, event)}
>
```

This will fix the css styling issue and as a bonus also allows users to right-click on the navigation buttons and open in a new tab, for example.

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
```diff
@@ -95,6 +95,8 @@ export default class UserEntityChart extends React.Component<
       event.preventDefault();
     }
 
+    if (newPage <= 0) return; // Checking if newPage in query is a positive integer.
```
MonkeyDo, line 98:
```suggestion
    const { totalPages } = this.state;
    if (newPage <= 0 || newPage > totalPages) return; // Checking if newPage in query is a positive integer.
```

That way we guard against going over the total number of pages as well (which does not break anything currently but results in empty pages).

Please also add a small test for this as you did for newPage <= 0.

MonkeyDo (changes requested):
Thanks for tackling this @Solai-Adithya  :)
I had a good look and have a couple of suggestions :

MonkeyDo:
Hi @Solai-Adithya !
Can I do anything to help you reopen and finish this PR?

Solai-Adithya:
Hi @MonkeyDo, I closed the Pull Request because I was facing various issues with setting up the project (I was not engaged with the project for some time, when I came back, I had to setup the project again and it kept failing). I'll try setting it up now, If I am successful with the setup, I'll commit and recreate the PR. If I face issues now, I'll report on IRC

Solai-Adithya:
Hi @MonkeyDo I was trying to do something else and accidentally closed this and lost commits and I couldn't undo it. Later, while redoing it, I was running into issues with the localhost login and authentication. So after a long gap, I pulled it up and fixed the login issue today but I see that the bug is fixed already, So I'll close the ticket in JIRA as well?

MonkeyDo:
Hi @Solai-Adithya ! Thanks a lot for coming back to this!
It does look like the issue is fixed. Having a closer look, it seems to be my doing, although I don't particularly remember it :p 
https://github.com/metabrainz/listenbrainz-server/commit/d9ab0db84aa507fcc20b34cfe27a68803cb90b06

I think indeed you can close that ticket as fixed. One less issue !

If you're still interested in improving navigation links, we've got a few tickets open under a "navigation-and-links" label: https://tickets.metabrainz.org/issues/?jql=labels%20%3D%20navigation-and-links

PR #1377, opened 2021-04-07
LB-510: Minor text issue in error response for submit-listens

vardansaini:
# Problem
JIRA ticket LB-510 "Minor text issue in error response for submit-listens"

# Solution
I changed the error message under try for empty fields. I also fixed typo in except segment of code. 

# Action
The problem was not described clearly in the ticket. I believe some more changes might be needed but not exactly sure what those changes should be. It would be really great if I could get a little more guidance on what exactly is needed.


alastair:
@brainzbot add to whitelist

mayhem:
Replaced by #1401

PR #1380, opened 2021-04-08
Add artist credit recording lookup to labs api & MBID mapping improvements

mayhem:
While working on the MBID mapping writer it became clear that we're getting a lot more exact matches than I expected now and that the artist credit recording dataset is perfect for this. Postgres can handle these queries much faster than typesense, so this new endpoint allows me to send the majority of traffic to postgres and not to typesense.

### DIFF
--- a/listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
+++ b/listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+
+import psycopg2
+import psycopg2.extras
+from datasethoster import Query
+from unidecode import unidecode
+
+import config
+
+
+class ArtistCreditRecordingLookupQuery(Query):
+
+    def names(self):
+        return ("acr-lookup", "MusicBrainz Artist Credit Recording lookup")
+
+    def inputs(self):
+        return ['artist_credit_name', 'recording_name']
+
+    def introduction(self):
+        return """This lookup performs an semi-exact string match on Artist Credit and Recording. The given parameters will have non-word
+                  characters removed, unaccted and lower cased before being looked up in the database."""
+
+    def outputs(self):
+        return ['artist_credit_name', 'release_name', 'recording_name', 
+                'artist_credit_id', 'release_mbid', 'recording_mbid']
+
+    def fetch(self, params, offset=-1, count=-1):
+        artists = []
+        recordings = []
+        for param in params:
+            artists.append("".join(unidecode(param['artist_credit_name'].lower()).split()))
+            recordings.append("".join(unidecode(param['recording_name'].lower()).split()))
+        artists = tuple(artists)
+        recordings = tuple(recordings)
+
+        with psycopg2.connect(config.DB_CONNECT_MB) as conn:
+            with conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as curs:
+                curs.execute("""SELECT DISTINCT ac.name AS artist_credit_name, 
+                                       rl.name AS release_name, 
+                                       r.name AS recording_name,
+                                       rl.gid AS release_mbid,
+                                       r.gid AS recording_mbid,
+                                       artist_credit_id
+                                  FROM mapping.recording_artist_credit_pairs
+                                  JOIN recording r
+                                    ON r.id = recording_id
+                                  JOIN release rl
+                                    ON rl.id = release_id
+                                  JOIN artist_credit ac
+                                    ON r.artist_credit = ac.id
+                                 WHERE artist_credit_name IN %s
+                                   AND recording_name IN %s""", (artists, recordings))
+
+                results = []
+                while True:
+                    data = curs.fetchone()
+                    if not data:
+                        break
+
+                    results.append(dict(data))
+
+                return results
--- a/listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
+++ b/listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 
+import re
+
 import psycopg2
 import psycopg2.extras
 from datasethoster import Query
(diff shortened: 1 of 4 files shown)


amCap1712 (approved):

listenbrainz/mbid_mapping/mapping/mbid_mapping.py
```diff
@@ -23,15 +26,17 @@ def create_tables(mb_conn):
         with mb_conn.cursor() as curs:
             curs.execute("DROP TABLE IF EXISTS mapping.tmp_mbid_mapping")
             curs.execute("""CREATE TABLE mapping.tmp_mbid_mapping (
+                                         id                        SERIAL,
```
alastair, line 29: SERIAL sets up a sequence and uses it to fill in this value if you don't provide it. You're generating your own ids and inserting them with `insert_rows(mb_curs2, "mapping.tmp_mbid_mapping", rows)`, which means that the sequence won't be advanced. I can't see you creating an index or PK on this field, it might be better to make it explicitly `INTEGER NOT NULL` so that we don't run into any unexpected problems later

alastair: in fact, I see you doing `DELETE FROM mapping.tmp_mbid_mapping WHERE id IN (` so maybe an index is a good idea here too (unless I can't see where you're doing it)

alastair (commented):
Looks good, just one query about the new id column

listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
mayhem: No, this is somewhat new -- the semantics of a nearly-exact match (remove whitespace, non word and unaccent) between typesense and PG data is a bit different that I made this newish endpoint necessary. But overall, the bono setup is working -- I upgrade dataset-hoster yesterday and all is well.

listenbrainz/mbid_mapping/mapping/mbid_mapping.py
mayhem: Hehehhehe, yes this part was a bit... odd. When inserting using execute_values, it isn't exactly clear (supported?) on how to define DEFAULT as one of the values, which should be passed instead of an INT for serial. But since I am building the table by hand and can easily add an increment, I'll just manually specify them.

As for actually needing an index on this column and the delete in question -- this is done before indexes exist -- its being done FOR index creation. I want a UNIQUE index on the table and thus need to remove dups that resolve to the same combined_lookup value -- dups are not needed and complicate querying.

I don't see needing that column/index post table creation/population, since all the lookups happen via combined_lookup.

listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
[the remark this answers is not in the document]
alastair: ah, now I see the subtle differences between the two. Is it a good idea to either 1) delete the one in bono, or 2) give this one a different name given that they have different semantics? I can see us changing the base url and keeping the input data the same and being confused as to why it's not working

listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
```diff
@@ -0,0 +1,65 @@
...
+                                  FROM mapping.mbid_mapping
+                                 WHERE combined_lookup IN %s""", (lookup_strings,))
+
+                results = []
+                while True:
+                    data = curs.fetchone()
+                    if not data:
+                        break
+
+                    data = dict(data)
```
alastair, line 58: as you're already using DictCursor you probably don't need this line in order to do `data["combined_lookup"]`

however the later line when you add it to results is needed to cast it to a real dict

listenbrainz/mbid_mapping/mapping/mbid_mapping.py
alastair: Yeah, I guessed that the reason for doing the serial manually was because you can't set a default without writing out all of the columns in the INSERT statement (`INSERT INTO table (recording_name, recording_mbid) VALUES (1, 2)` etc).

if the deduplicate process is fast enough, you probably don't need to bother with the index then. I guess that this will do a table scan instead of index lookup for the `WHERE id IN (...)` clause, though? Does this have the potential to slow down in the future if this table gets bigger?

My comment about SERIAL was that there is a potential data inconsistency error:
 - set id SERIAL, default value becomes `nextval(id sequence)`
 - do a bulk insert, setting your own id values
 - at some point in time, try and insert a row into the table, relying on the sequence to give you an id number
 - the next value from the sequence already exists in the table and fails

of course, the insert will only fail if you have a unique index on id, which you don't. Also,  you're only ever creating this table and its contents automatically, so you'll probably never run into this situation.

My only suggestion here was to recommend to set the column type explicitly to INTEGER instead of SERIAL to make it explicit that you're not relying on any magic sequences.

listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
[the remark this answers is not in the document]
mayhem: Yeah, I'm still struggling with that a bit. The things in bono are half baked and in the process of baking them for deployment to labs, things always change. In particular naming of the queries and often adding multiple row support. My theory was to remove old one from bono once they are merged on deployed onto labs, but that hasn't been happening.

listenbrainz/labs_api/labs/api/artist_credit_recording_lookup.py
mayhem: I'll remove the second cast and be done with it.

alastair: oh yeah, doing it that way works too.

listenbrainz/mbid_mapping/mapping/mbid_mapping.py
mayhem: ok, that last point is a good suggestion, I like it.

alastair (approved):

PR #1381, opened 2021-04-08
LB-656: Calendar-based navigation in listens page

MonkeyDo:
This PR aims to add a date picker to navigate to the listens at a specific date.

This PR is deployed to test.listenbrainz.org

I decided on https://github.com/wojtekmaj/react-date-picker as a library, which required minimal styling to fit our page and had all the features I needed.

The date picker is also updated whenever the user navigates through other means.

Here is a look at what the date picker looks like:
image


MonkeyDo:
@brainzbot retest this please

listenbrainz/webserver/static/css/listens-page.less
```diff
@@ -384,3 +384,17 @@
...
+#navigation.pager{
+  display: flex;
+  flex-wrap: wrap;
+  & > .date-time-picker {
+   flex: 1;
+   order: 0;
+   
+   @media (max-width: @screen-phone) {
+     // on phones, put the datepicker below the arrow nav
+     order: 1;
```
shivam-kapila, line 397: Can we have some margin above it so that it doesn't glue to the pager?

listenbrainz/webserver/static/css/DatePicker.less
```diff
@@ -0,0 +1,342 @@
...
+// 		text-align: center;
+// 	  }
+// 	}
+  
+// 	&__second-hand {
+// 	  &__body {
+// 		background-color: @orange;
+// 	  }
+// 	}
+//   }
```
amCap1712, line 176: Are these needed any more? If not let's remove these. We can always view the Git history in case we need them again.

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -447,12 +463,61 @@ export default class RecentListens extends React.Component<
...
+        loading: false,
+      });
+      return;
+    }
+    this.setState(
+      {
+        listens: newListens,
+        nextListenTs: newListens[newListens.length - 1].listened_at,
+        previousListenTs: newListens[0].listened_at,
+        lastFetchedDirection: "newer",
```
amCap1712, line 507: It seems we are not using `lastFetchedDirection` anywhere now. If that's the case, let's remove it.

amCap1712 (approved):
Looks good! Awesome work :100:

listenbrainz/webserver/static/css/DatePicker.less
MonkeyDo: Totally forgot about those 😓

MonkeyDo:
@brainzbot retest this please

PR #1382, opened 2021-04-09
Changes needed to deploy the mbid-mapping for the new LB startup service

mayhem:
As it says on the tin -- tested and working in production now.

### DIFF
--- a/listenbrainz/mbid_mapping/Dockerfile
+++ b/listenbrainz/mbid_mapping/Dockerfile
@@ -18,9 +18,9 @@ COPY . /code/mapper
 FROM mbid-mapping-base as mbid-mapping-prod
 
 # service start up scripts
-COPY docker/consul-template.conf /etc/consul-template.conf
 RUN mkdir /etc/service/mapper
-COPY docker/mapper.service /etc/service/mapper/run
+COPY ./docker/consul-template.conf /etc/consul-template-mbid-mapping.conf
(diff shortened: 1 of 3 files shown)


listenbrainz/mbid_mapping/Dockerfile
```diff
@@ -18,9 +18,9 @@ COPY . /code/mapper
 FROM mbid-mapping-base as mbid-mapping-prod
 
 # service start up scripts
-COPY docker/consul-template.conf /etc/consul-template.conf
 RUN mkdir /etc/service/mapper
```
amCap1712, line 21: While we are at it, we can probably remove this line. The documentation for [COPY](https://docs.docker.com/engine/reference/builder/#copy) says
> If <dest> doesn’t exist, it is created along with all missing directories in its path.

listenbrainz/mbid_mapping/docker/mapper.service
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
```
amCap1712, line 1: The permissions for this file seem to have changed. Just to confirm, that's intended or harmless?

mayhem: Intended.

mayhem:
I fixed the suggested change -- merging so I can deploy a labs update.

PR #1383, opened 2021-04-12
LB-864: Fix html links used for pagination

MonkeyDo:
LB-864
Before this PR, users won't be able to use the pagination buttons on the Recent Listens page as they would regular html links.
In particular, they can't choose to open in a new tab or window.

Considering we're already using `<a>` html elements inside our bootstrap pagination buttons, it's trivial to add an `href` attribute to fix those issues and have actual valid html links.

For use in that current page, we prevent the default link behavior (and thus navigating away) and refresh the page with the API, just as before.


listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -566,6 +578,13 @@ export default class RecentListens extends React.Component<
       currentUser,
     } = this.props;
 
+    const isNewestButtonDisabled = listens?.[0]?.listened_at >= latestListenTs;
+    const isNewerButtonDisabled =
+      !previousListenTs || previousListenTs >= latestListenTs;
+    const isOlderButtonDisabled =
+      !nextListenTs || nextListenTs <= oldestListenTs;
+    const isOldestButtonDisabled =
+      listens?.[listens?.length - 1]?.listened_at <= oldestListenTs;
```
alastair, line 587: if listens is null, will `listens?.length - 1` return a number or some other placeholder? or will the first `listens?.` ensure that this doesn't happen?

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -688,15 +708,18 @@ export default class RecentListens extends React.Component<
                           if (e.key === "Enter") this.handleClickNewer();
                         }}
                         tabIndex={0}
+                        href={
+                          isNewerButtonDisabled
+                            ? undefined
+                            : `${window.location.origin}${window.location.pathname}?min_ts=${previousListenTs}`
```
alastair, line 714: would this be clearer if it was just `<a href="?min_ts=xxxxxxx">` ? I think the browser will sort out all of the relative path calculation properly.

MonkeyDo: Mm, yes I suppose it should. Whatever I had in mind for deciding against it at first doesn't make sense. I'll change it now.

listenbrainz/webserver/static/js/src/RecentListens.tsx
MonkeyDo: If listens is null `listens?` will evaluate as `undefined` (same for the other `?.` operators in that line)

amCap1712 (approved):

PR #1361, opened 2021-03-28
Youtube Player Revamp

amCap1712:
Implementation for Youtube Player and music services revamp. The server side implementation is more or less done. The major parts missing from it is a migration plan for moving data from existing spotify users to the new `external_auth` table. Also, I need to add tests for the server side Youtube Playe code.

Things on the front end side need some more work. There are a few known bugs in BrainzPlayer UI and I am in the process of fixing them.

The current implementation is working and can be tested locally.
Screenshot from 2021-03-29 01-37-11

To test locally, add a `YOUTUBE_CONFIG` variable to your `config.py ` with entire contents of `client_secrets.json` file available from after configuring Google OAuth API, another `YOUTUBE_API_KEY` with the a key from Google APIs with Youtube Data v3 enabled and finally a `YOUTUBE_CALLBACK_URL` with the redirect uri.

PR #1384, opened 2021-04-12
Add SQL scripts for connecting new music services and migrating user data

amCap1712:
The complete PR is at  https://github.com/metabrainz/listenbrainz-server/pull/1361. I am trying to break it down into smaller chunks to aid easier review. Plus, it would be nice to get the DB changes in before the rest of PR so that we can test the player on test or beta.


amCap1712:
I noticed we also have an index for the `spotify_auth` table. In this case, we'll need to convert it to a partial index. We can add that later if needed.

admin/sql/updates/2021-04-12-migrate-existing-spotify-users-to-new-schema.sql
```diff
@@ -0,0 +1,26 @@
...
+     service_details
+    )
+SELECT user_id,
+       'spotify' as service,
+       user_token as access_token,
+       refresh_token,
+       token_expires,
+       last_updated,
+       record_listens,
+       jsonb_build_object(
```
alastair, line 20: 🎉

admin/sql/updates/2021-04-12-new-music-services-table.sql
```diff
@@ -0,0 +1,25 @@
+BEGIN;
```
alastair, line 1: maybe add a suffix to the dates to ensure that we know to apply this one before the migrate one

admin/sql/updates/2021-04-12-migrate-existing-spotify-users-to-new-schema.sql
```diff
@@ -0,0 +1,26 @@
+BEGIN;
```
alastair, line 1: what will the process be when we deploy this? I suspect we'll need to re-run this script in order to add all of the new users that have since configured spotify with the old system, right? Perhaps then it might be a good idea to delete from this table and reset the sequence before doing the insert? then we can just reuse this script again when we do the final deploy

admin/sql/updates/2021-04-12-new-music-services-table.sql
```diff
@@ -0,0 +1,25 @@
...
+    service                 external_auth_service NOT NULL,
+    access_token            TEXT NOT NULL,
+    refresh_token           TEXT,
+    token_expires           TIMESTAMP WITH TIME ZONE,
+    last_updated            TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+    record_listens          BOOLEAN NOT NULL,
+    service_details         JSONB
+);
+
+ALTER TABLE external_auth ADD CONSTRAINT external_auth_pkey PRIMARY KEY (id);
```
alastair, line 17: I see in your original proposal in the document that you were going to use (user_id, service) as the primary key. If you still want to ensure that we can have only 1 user,service row you could add a unique index over these two fields.

We'll also need other generic indexes - I'm not 100% about the kinds of queries that you'll need to do, but it sounds like one on user_id, one on service, and one on (user_id, service) is probably useful.

amCap1712: Yes, @mayhem suggested that we add a id SERIAL as well so I think it makes sense to use it as the PK. I'll add the unique index on \ (user_id, service). I think adding an index on user_id might be useful but on service might not be. My crude understanding (maybe wrong) of an index is to allow using binary search on the contents. But for service there are going to very few values and large number of occurences of the same value. The other two indices will probably cover it well enough to not create this one. We can add remaining indices later as the need arises.

admin/sql/updates/2021-04-12-new-music-services-table.sql
```diff
@@ -0,0 +1,25 @@
+BEGIN;
+
+CREATE TYPE external_auth_service AS ENUM ('spotify', 'youtube');
+
+CREATE TABLE external_auth (
```
alastair, line 5: Did I glance over the design document and miss that this was the proposed name? :)
how about `external_service_oauth` to make it a bit more explicit about what kind of external it is?

amCap1712: Makes sense. `external_service_oauth` is clearer.

admin/sql/updates/2021-04-12-new-music-services-table.sql
alastair: A good rule of thumb is simply "if you want to `select WHERE <this field>=something` then you should have an index".
So if you need to select all users for a service we should have an index on service. You're right that with the current setup we may have few enough rows that it doesn't make a huge difference, but imo it's always better safe than sorry here.

amCap1712: Makes sense. Agreed, will add.

admin/sql/drop_tables.sql
```diff
@@ -3,6 +3,7 @@ BEGIN;
 DROP TABLE IF EXISTS "user"                         CASCADE;
 DROP TABLE IF EXISTS data_dump                      CASCADE;
 DROP TABLE IF EXISTS spotify_auth                   CASCADE;
+DROP TABLE IF EXISTS external_service_oauth                  CASCADE;
```
mayhem, line 6: nit: indent.

mayhem (approved):

PR #1386, opened 2021-04-13
Automatically build and push image to Docker Hub when new tag is pushed

amCap1712:
Our deployment workflow is a bit convoluted. One step in simplifying the deployment is to automate building of docker images. Hence, adding an actions workflow for it.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -3,7 +3,7 @@ name: Build image and publish to Docker Hub
 on:
   push:
     tags:
-      - '*'
+      - 'v-[0-9]{4}-[0-9]{2}-[0-9]{1,2}.[0-9]+'
 
 jobs:
 
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -1,0 +1,23 @@
+name: Build image and publish to Docker Hub
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+
+  deploy:
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v2
+    
+    - name: Login to Docker Hub
+      run: echo ${{ secrets.DOCKER_HUB_PASSWORD }} | docker login -u ${{ secrets.DOCKER_HUB_USERNAME }} --password-stdin
+    
+    - name: Build and push image
+      # Github stores the current tag in an enviroment variable (GITHUB_REF) in the format /refs/tags/TAG_NAME. 
+      # Using shell parameter expansion, we extract the TAG_NAME.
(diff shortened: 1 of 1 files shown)


.github/workflows/build-image.yml
```diff
@@ -0,0 +1,23 @@
+name: Build image and publish to Docker Hub
+
+on:
+  push:
+    tags:
+      - '*'
```
alastair, line 6: does this mean that it'll only build if we make a tag? What about making it a bit more specific? Can we ensure that the tag is formatted correctly?

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    
+    - name: Login to Docker Hub
+      run: echo ${{ secrets.DOCKER_HUB_PASSWORD }} | docker login -u ${{ secrets.DOCKER_HUB_USERNAME }} --password-stdin
+    
+    - name: Build and push image
+      # Github stores the current tag in an enviroment variable (GITHUB_REF) in the format /refs/tags/TAG_NAME. 
+      # Using shell parameter expansion, we extract the TAG_NAME.
+      run: ./docker/push.sh 'prod' "${GITHUB_REF/refs\/tags\//}"
```
alastair, line 23: I opened a ticket about improving push.sh to make it fail properly if any intermediate process fails (e.g. don't push if the build fails). Maybe we could include that here? (although I guess the push will fail also if the build fails, because the tagged version won't exist)

alastair:
this is super cool!

.github/workflows/build-image.yml
amCap1712: Yes, it'll only build and push the image on a tag. Yes, I think it supports regex so we can restrict the build to properly formatted tags.

amCap1712: We can do something like `v-[0-9]{4}-[0-9]{2}-[0-9]{1,2}.[0-9]+` instead of `*` to trigger the workflow only if the tag matches the regex.

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,23 @@
+name: Build image and publish to Docker Hub
+
+on:
+  push:
+    tags:
+      - 'v-[0-9]{4}-[0-9]{2}-[0-9]{1,2}.[0-9]+'
```
amCap1712, line 6: I would think it should be `\.` instead of `.` but the [semver example](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags) in the docs says otherwise,

alastair: it looks like they're patterns, not regexes, so the guidelines are probably correct.

alastair (commented):
🎆

alastair (approved):

PR #1387, opened 2021-04-13
Fix user similarity thresholding

mayhem:
The absolute threshold method of limiting max number of similar users wasn't working well. Changing it from an absolute value threshold to a max number of users per user threshold works much better and is more future proof.

This PR also commits a number of tweaked settings (reduce dataframe days to 1 year and minimum listens to 50) that give good results for a broad number of users.


mayhem:
@brainzbot please retest

amCap1712 (approved):

PR #1388, opened 2021-04-14
Setup request consumer for new cluster

amCap1712:
Modify the request consumer setup to work with the new cluster setup. Since we have now installed spark on the host directly, we do not need to include it in the docker image. 

We'll also need to change the configuration parameters according to the resources available on the host and yarn cluster manager.

### DIFF
--- a/Dockerfile.spark.newcluster
+++ b/Dockerfile.spark.newcluster
@@ -1,5 +1,17 @@
 FROM metabrainz/python:3.7-20210115
 
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+    wget \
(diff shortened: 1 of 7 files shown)


mayhem (approved):
Cursory glance, looks sane for our testing purposes now.

mayhem:
Dockerfile.spark.newcluster will be nenamed in a subsequent PR.

amCap1712:
This is the new setup we are currently using in production. The changes here are only one half of the story and not enough to reproduce the cluster on another set of machines. I hope to put up some document explaining the steps to setup a cluster soon.

PR #1390, opened 2021-04-16
Remove time ranges and refactor listen fetching

mayhem:
TBC


listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -34,8 +34,17 @@
...
 DUMP_FILE_SIZE_LIMIT = 1024 * 1024 * 1024  # 1 GB
 DATA_START_YEAR = 2005
-SECONDS_IN_TIME_RANGE = 432000
+DATA_START_YEAR_IN_SECONDS = 1104537600
 
+# This value MUST match the time_bucket definition of the listen_count view.
+# See admin/timescale/create_views.sql for more details.
+LISTEN_COUNT_BUCKET_WIDTH = 86400 * 5
+
+# How many listens to fetch on the first attempt. If we don't fetch enough, doublt it, try again.
```
MonkeyDo, line 43: "doublt it"
typo aside, if I understand correctly we're multiplying by WINDOW_SIZE_MULTIPLIER rather than doubling.

mayhem:
please test

MonkeyDo (commented):
Hard for me to review the python part and knowing next to nothing of timescale, but I didn't see anything obviously wrong.
Had another pass over the JS bits, looks good.

admin/timescale/create_views.sql
```diff
@@ -1,18 +1,8 @@
...
-CREATE VIEW listened_at_max
-       WITH (timescaledb.continuous, timescaledb.refresh_lag=43200, timescaledb.refresh_interval=3600)
-         AS SELECT time_bucket(bigint '86400', listened_at) AS listened_at_bucket, user_name, max(listened_at) AS max_value
-            FROM listen group by time_bucket(bigint '86400', listened_at), user_name;
-
-CREATE VIEW listened_at_min
-       WITH (timescaledb.continuous, timescaledb.refresh_lag=43200, timescaledb.refresh_interval=3600)
-         AS SELECT time_bucket(bigint '86400', listened_at) AS listened_at_bucket, user_name, min(listened_at) AS min_value
-            FROM listen group by time_bucket(bigint '86400', listened_at), user_name;
+CREATE VIEW listen_count_30day
```
amCap1712, line 3: Given that we are on Timescale 2 in prod now, we'll probably need to rewrite this query. The breaking changes in creating continuous aggregates are listed at https://docs.timescale.com/timescaledb/latest/overview/release-notes/changes-in-timescaledb-2/#continuous-aggregates.

amCap1712 (commented):
The PR looks good but we should take a look at the timescale bits again as the prod is now on 2.2. However, the local development image is still on 1.7. Therefore, I expect that the tests would begin to fail. To migrate the local development image to v2.*, we will also need to update the pre-existing timescale bits.

admin/timescale/create_views.sql
mayhem: Yep, for sure. I'll try and work on this later today.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -25,59 +25,75 @@
...
+        self.dump_temp_dir_root = conf.get(
+            'LISTEN_DUMP_TEMP_DIR_ROOT', tempfile.mkdtemp())
+
+    def set_empty_cache_values_for_user(self, user_name):
+        """When a user is created, set the listen_count and timestamp keys so that we
+           can avoid the expensive lookup for a brand new user."""
 
-    def get_listen_count_for_user(self, user_name, need_exact=False):
+        cache.set(REDIS_USER_LISTEN_COUNT + user_name, 0, time=0, encode=False)
+        cache.set(REDIS_USER_TIMESTAMPS + user_name, "0,0", time=0)
```
amCap1712, line 72: We can probably use hset here.

admin/timescale/create_views.sql
```diff
@@ -1,18 +1,4 @@
...
-
-CREATE VIEW listened_at_min
-       WITH (timescaledb.continuous, timescaledb.refresh_lag=43200, timescaledb.refresh_interval=3600)
-         AS SELECT time_bucket(bigint '86400', listened_at) AS listened_at_bucket, user_name, min(listened_at) AS min_value
-            FROM listen group by time_bucket(bigint '86400', listened_at), user_name;
-
-COMMIT;
+CREATE MATERIALIZED VIEW listen_count_30day
+       WITH (timescaledb.continuous)
+         AS SELECT time_bucket(bigint '2592000', listened_at) AS listened_at_bucket, user_name, count(listen)
```
amCap1712, line 3: I think timescale supports intervals like 30 days. Would be more readable to use that if possible.

mayhem: I sounds like it should be supported, but in reality if you use anything other than seconds, it fails. I really wanted to use month, rather than 30 days... :(

admin/timescale/create_views.sql
```diff
@@ -1,4 +1 @@
-CREATE MATERIALIZED VIEW listen_count_30day
-       WITH (timescaledb.continuous)
-         AS SELECT time_bucket(bigint '2592000', listened_at) AS listened_at_bucket, user_name, count(listen)
-            FROM listen group by time_bucket(bigint '2592000', listened_at), user_name;
+CREATE MATERIALIZED VIEW listen_count_30day WITH (timescaledb.continuous) AS SELECT time_bucket(bigint '2592000', listened_at) AS listened_at_bucket, user_name, count(listen) FROM listen GROUP BY time_bucket(bigint '2592000', listened_at), user_name;
```
amCap1712, line 1: Also, We can use `listened_at_bucket` directly in GROUP BY clause. I tried locally and it seemed to work.

admin/timescale/create_views.sql
amCap1712: Ah! ok. makes sense. Intervals like 30 days only work if the column is a date or timestamp.

listenbrainz/listenstore/redis_listenstore.py
```diff
@@ -1,28 +1,31 @@
...
 
 class RedisListenStore(ListenStore):
 
-    RECENT_LISTENS_KEY = "lb_recent_sorted"
+    RECENT_LISTENS_KEY = "rl-"
     RECENT_LISTENS_MAX = 100
+    PLAYING_NOW_KEY = "pn."
     LISTEN_COUNT_PER_DAY_EXPIRY_TIME = 3 * 24 * 60 * 60  # 3 days in seconds
-    LISTEN_COUNT_PER_DAY_KEY_FORMAT = "lb_listen_count_for_day_{}"
+    LISTEN_COUNT_PER_DAY_KEY_FORMAT = "lc-day-"
```
alastair, line 19: I guess this was called `_FORMAT` because it contains a format string that we used `.format()` with. As you're just using it as a prefix with string concatenation, perhaps we could call it `LISTEN_COUNT_PER_DAY_KEY` or `LISTEN_COUNT_PER_DAY_KEY_PREFIX`?

listenbrainz/listenstore/redis_listenstore.py
```diff
@@ -1,28 +1,31 @@
...
 
 
     def __init__(self, log, conf):
         super(RedisListenStore, self).__init__(log)
-        self.redis = Redis(host=conf['REDIS_HOST'], port=conf['REDIS_PORT'], decode_responses=True)
-        self.ns = conf['REDIS_NAMESPACE']
+
+        # Initialize brainzutils cache
+        init_cache(host=conf['REDIS_HOST'], port=conf['REDIS_PORT'],
```
alastair, line 26: Is this the only place where we init_cache? If possible we should only set this once in `create_app`, perhaps [after we create the redis connection](https://github.com/metabrainz/listenbrainz-server/blob/bbf6a832b30a7e3fa1b35f808299646d75719c32/listenbrainz/webserver/__init__.py#L97). By this point, `cache._r` should be set so that you can read it here.

listenbrainz/listenstore/redis_listenstore.py
```diff
@@ -34,7 +37,7 @@ def get_playing_now(self, user_id):
                 Listen object which is the currently playing song of the user
 
         """
-        data = self.redis.get(self.ns + 'playing_now:{}'.format(user_id))
+        data = cache.get(self.PLAYING_NOW_KEY + str(user_id))
```
alastair, line 40: in `__init__` you create `self.redis`, but here you just use the cache. Later in this file (check_connection) you use `cache._r` instead of `self.redis`. Could be made more consistent.

listenbrainz/tests/integration/test_profile_views.py
```diff
@@ -47,6 +47,7 @@ def test_export(self):
         # now export data and check if contains all the listens we just sent
         resp = self.client.post(url_for('profile.export_data'))
         self.assert200(resp)
+        current_app.logger.warn(str(resp.data))
```
alastair, line 50: not needed?

alastair (commented):
Preliminary review containing some comments. Looking at listenstore implementation now

listenbrainz/listenstore/redis_listenstore.py
mayhem: Sadly it isn't ready. Because of the way we have our imports calling create_redis() imports the redis_listenstore and then tries to use the cache object that is not ready. Let's fix this later if you feel strongly about it.

mayhem:
All items, except one (I commented) have been taken care of.

listenbrainz/listenstore/redis_listenstore.py
alastair: this ties in with LB-879, which means that we can do it there. OK to leave here.

listenbrainz/listenstore/timescale_listenstore.py
alastair: would this be a hash where the key is REDIS_USER_TIMESTAMPS, the field is user_name, and then the value is the 0 or "0,0" etc?

amCap1712: I was thinking REDIS_USER_TIMESTAMPS + user_name to be the hash and min and max as fields in it.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -208,94 +246,158 @@ def insert(self, listens):
...
 
         return inserted_rows
 
-    def fetch_listens_from_storage(self, user_name, from_ts, to_ts, limit, order, time_range):
+    def fetch_listens_from_storage(self, user_name, from_ts, to_ts, limit, order):
         """ The timestamps are stored as UTC in the postgres datebase while on retrieving
             the value they are converted to the local server's timezone. So to compare
             datetime object we need to create a object in the same timezone as the server.
 
+            If neither from_ts nor to_ts is provided, the latest listens for the user are returned.
```
alastair, line 278: how many is "the latest" ?

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -208,94 +246,158 @@ def insert(self, listens):
...
+            min_ts, max_ts = self.get_timestamps_for_user(user_name)
+            if min_user_ts < 0:
+                min_user_ts = min_ts
+                max_user_ts = max_ts
+            if min_ts and min_ts < min_user_ts:
+                min_user_ts = min_ts
+            if max_ts and max_ts > max_user_ts:
+                max_user_ts = min_ts
+
+        if to_ts == None and from_ts == None:
```
alastair, line 314: use `is` to compare against None

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -208,94 +246,158 @@ def insert(self, listens):
...
-        else:
-            max_timestamp_window = SECONDS_IN_TIME_RANGE * time_range
-            if to_ts is None:
-                to_ts = from_ts + max_timestamp_window
-            elif from_ts is None:
-                from_ts = to_ts - max_timestamp_window
-
+        min_user_ts = max_user_ts = -1
+        for user_name in user_names:
+            min_ts, max_ts = self.get_timestamps_for_user(user_name)
```
alastair, line 305: Is this going to come from redis the majority of the time? (I believe it is, just checking). If it's coming from the db, then this is 1 query per username, which could potentially be slow (faster to do it for all usernames in a single query).
How many usernames are typically used here? Again - if there are always many usernames it may be faster to get all keys from redis at once. But if there's normally only one then it's probably not worth optimising this lookup

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -208,94 +246,158 @@ def insert(self, listens):
...
+            if min_user_ts < 0:
+                min_user_ts = min_ts
+                max_user_ts = max_ts
+            if min_ts and min_ts < min_user_ts:
```
alastair, line 309: Could this loop/check be improved by doing something like

    min_user_ts = min(min_ts, min_user_ts)
    max_user_ts = max(max_ts, max_user_ts)

I would set the sentinels to None before the loop (to make it more explicit that this is an empty value), and have an additional check like you do on line 306.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -208,94 +246,158 @@ def insert(self, listens):
...
+
+                        if from_dynamic:
+                            to_ts -= window_size
+                            window_size *= WINDOW_SIZE_MULTIPLIER
+                            from_ts -= window_size
+
+                        break
+
+                    listens.append(Listen.from_timescale(
+                        result[0], result[1], result[4], result[2], result[3]))
```
alastair, line 384: I see that this is copied from an existing line, but seeing rows added in a different order to what they were selected from seems like a recipe for disaster. We should add it to an issue to fix at a later stage

alastair (commented):
listenstore reviews

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -25,59 +25,75 @@
...
-        query = "SELECT SUM(count) FROM listen_count WHERE user_name = :user_name"
+    def reset_listen_count(self, user_name):
+        """ Reset the listen count of a user from cache and put in a new calculated value.
+            returns the re-calculated listen count.
 
+            Args:
+                user_name: the musicbrainz id of user whose listen count needs to be reset
+        """
+        query = "SELECT SUM(count) FROM listen_count_30day WHERE user_name = :user_name"
+        t0 = time.time()
```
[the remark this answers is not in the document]
mayhem: And then the stats for one query will be off that likely no one will ever see.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -110,20 +160,32 @@ def _select_single_timestamp(self, select_min_timestamp, user_name):
...
-            query = "SELECT min(min_value) AS value FROM listened_at_min WHERE user_name = :user_name"
-        else:
-            query = "SELECT max(max_value) AS value FROM listened_at_max WHERE user_name = :user_name"
+            sort_clause = ""
+            function = "min"
 
+        query = """SELECT %s(listened_at) AS ts
+                     FROM listen
+                     WHERE user_name = :user_name
+                  GROUP BY listened_at
```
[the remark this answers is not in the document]
mayhem: Probably a leftover from previous times. Let me try without.

listenbrainz/listenstore/timescale_listenstore.py
mayhem: They are mostly coming from redis, yes. I see your points, but they all existed as this before this PR and I really don't want to fix this stuff now. Please make a ticket.

listenbrainz/listenstore/timescale_listenstore.py
mayhem: Reordered.

listenbrainz/listenstore/timescale_listenstore.py
mayhem: That depends on what you pass in for limit. I don't see your question here.

listenbrainz/listenstore/timescale_listenstore.py
mayhem: Reworked to be more concise.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -86,21 +102,55 @@ def get_listen_count_for_user(self, user_name, need_exact=False):
...
+            max_ts = self._select_single_timestamp(False, user_name)
+            # intended for production monitoring
+            self.log.info("timestamps %s %.2fs" % (user_name, time.time() - t0))
+            if min_ts and max_ts:
+                cache.set(REDIS_USER_TIMESTAMPS + user_name,
+                          "%d,%d" % (min_ts, max_ts), time=0)
+
+        return min_ts, max_ts
 
     def _select_single_timestamp(self, select_min_timestamp, user_name):
```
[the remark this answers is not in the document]
mayhem: I'd prefer to leave this for when we improve this code to be more efficient.

PR #1393, opened 2021-04-19
Check for CI env variable

amCap1712:
GH Actions is reporting passed workflow even when tests fail. This is probably because the script continues running even when the tests fail. And then the final status reported by the script is due to  the `exit 0`. Therefore, store the exit code returned by the command executing the tests and return that at the end of the script. 

Unrelated change, instead of using `-ci` check if the CI environment variable is set to `true`. This environment variable is automatically set by GH Actions when running the workflows.

### DIFF
--- a/listenbrainz/db/tests/test_dump.py
+++ b/listenbrainz/db/tests/test_dump.py
@@ -47,6 +47,7 @@ class DumpTestCase(DatabaseTestCase):
         shutil.rmtree(self.tempdir)
 
     def test_create_private_dump(self):
+        assert False
         time_now = datetime.today()
         dump_location = db_dump.create_private_dump(self.tempdir, time_now)
         self.assertTrue(os.path.isfile(dump_location))
--- a/listenbrainz/db/tests/test_dump.py
+++ b/listenbrainz/db/tests/test_dump.py
@@ -47,7 +47,6 @@ class DumpTestCase(DatabaseTestCase):
         shutil.rmtree(self.tempdir)
 
     def test_create_private_dump(self):
-        assert False
         time_now = datetime.today()
         dump_location = db_dump.create_private_dump(self.tempdir, time_now)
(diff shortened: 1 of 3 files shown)


listenbrainz/db/tests/test_dump.py
```diff
@@ -47,6 +47,7 @@ def tearDown(self):
         shutil.rmtree(self.tempdir)
 
     def test_create_private_dump(self):
+        assert False
```
alastair, line 50: oops

alastair (commented):
can you run shellcheck on test.sh? I think there are some recommendations that it might give. Don't worry about fixing things that you didn't touch in this PR.

What sets the $CI env variable?

For the if/else check, can we instead use the check to change just the command that we run? e.g.
```
if [ "$CI" == "true" ] ; then
  command="format:ci"
else
  command="format"
fi

docker-compose ...... $command

```

listenbrainz/db/tests/test_dump.py
amCap1712: I added this intentionally to ensure the fix works. Will remove this before merging.

amCap1712:
The CI environment variable is set by [Github Actions](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables) itself.

amCap1712:
I ran shellcheck, it proposed many improvements but none of them are related to the changes in this PR.

alastair:
interesting, I would have thought shellcheck would suggest to use `"${CI}"` for reading variables, but no problem. Thanks for proactively running it!
Let's add a comment somewhere that explains where $CI comes from, so that we're not confused by it.

amCap1712:
I have added a check at the top of the script that prints if we are running inside CI along with the comment. This may be helpful in diagnosing issues in case something goes wrong.

alastair (approved):
thanks

PR #1394, opened 2021-04-19
Fix docker image workflow

amCap1712:
Fix pattern in docker image workflow and add caching.


alastair (approved):
does this work? If so, go for it!

PR #1395, opened 2021-04-20
Spotify player fixes and improvements

MonkeyDo:
A few things clumped in the same PR:

- Fix broken Spotify player due to a recent update in their API that borked it
- Revert to loading the Spotify SDK via a CDN rather than having to update the saved script every time they update it (no channels where that is communicated makes it tricky)
- Remove annoying verbose "We found a matching track in Spotify" message
- Stop fetching the currently paying spotify track on page load. This was used to display some album art in the player, but it's confusing to show something without user interaction and that might not be related to anything in the page)
- As a consequence, we needed to make the player a bit more interesting when no content is loaded:
image


alastair (commented):
Looks generally good to me, though I don't know anything about this part of the codebase. would be nice to test it live to make sure that it works as expected

alastair (approved):
I used it on test. looks good.

As I mentioned on IRC, one problem where I was unable to refresh tokens, and wasn't kicked to a "reconnect" workflow, but that seems unrelated. Also had to hard-refresh to reload css, but that's also unrelated.

PR #1397, opened 2021-04-21
Migrate spotify integration to new tables

amCap1712:
There are now two tables for integrating with external music services. `external_service_oauth` stores details like access_token etc and other authentication details. The `listens_importer` table only stores details related to importing listens. All services will use the `external_service_oauth` table but only services that support import listening history will need to use `listens_importer` table.

No new functionality is introduced in this PR, only a rewrite of the existing spotify integration is being done. The intent is to simplify integrating new services in the future.


listenbrainz/db/external_service_oauth.py
```diff
@@ -0,0 +1,112 @@
...
+
+from listenbrainz import db, utils
+import sqlalchemy
+
+
+def save_token(user_id, service, access_token, refresh_token, token_expires_ts, record_listens, service_details):
+    """ Add a row to the external_auth table for specified user with corresponding tokens and information.
+
+    Args:
+        user_id (int): the ListenBrainz row ID of the user
```
alastairporter, line 11: let's start migrating types to the function definition for new methods

listenbrainz/db/external_service_oauth.py
```diff
@@ -0,0 +1,112 @@
...
+            VALUES 
+            (:user_id, :service, :access_token, :refresh_token, :token_expires, :record_listens, :service_details)
+            """), {
+                "user_id": user_id,
+                "service": service.value,
+                "access_token": access_token,
+                "refresh_token": refresh_token,
+                "token_expires": token_expires,
+                "record_listens": record_listens,
+                "service_details": json.dumps(service_details),
```
alastairporter, line 33: not needed any more?

alastairporter (commented):
few quick comments before I look at it in more detail

mayhem (approved):
I see nothing sticking out as a problem here. You've run this code on test/beta, yes?

amCap1712:
I have tested this on beta and the spotify account linking seems to work fine but I do not use Spotify Premium so would be nice if someone could confirm streaming works for them. I have also tested the spotify reader locally but that's probably not enough.

Also, this PR alone is incomplete. #1404 is builds on this. Currently, the two PRs are separate because I think that may help ease reviewing. To deploy, #1404 should be merged into this PR and then this PR should be merged into the master.

amCap1712:
Superseded by #1361

PR #1398, opened 2021-04-21
Add build step for front-end tests

MonkeyDo:
Otherwise, new dependencies added to package.json won't be installed.
See this action  for an example: https://github.com/metabrainz/listenbrainz-server/pull/1343/checks?check_run_id=2399818752

### DIFF
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -25,6 +25,9 @@ jobs:
     - uses: satackey/action-docker-layer-caching@v0.0.11
       continue-on-error: true
 
+    - name: Build frontend tests
+      run: ./test.sh fe -b
+
     - name: Run frontend tests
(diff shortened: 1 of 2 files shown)


alastair:
for some reason `test.sh fe -b` builds the listenrbainz (python) image. It looks like we don't need it to run fe tests, so we can remove that.

amCap1712:
Is this the case in local development as well? I mean do we have to run `./test.sh fe -b` after adding a new frontend dependency so that tests run correctly locally.

alastair:
@amCap1712 Yes, you need to run fe -b to rebuild the local test container if dependencies change. I think that changing behaviour of the `test fe` command is for another issue...

amCap1712:
Agreed. Makes sense in that case.

alastair (approved):

PR #1399, opened 2021-04-21
Startup dump fixes, with merge conflict fixed

mayhem:
This PR replaces #1308.


mayhem:
This should be ready to merge. Creating dumps works, but once we merge this I will trigger a feedback dump in production, just to make sure.

admin/create-dumps.sh
```diff
@@ -18,6 +18,15 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
+set -e
+
+if [[ "${CONTAINER_NAME}" == "listenbrainz-cron-prod" && "${PROD}" == "prod" ]]
```
alastair, line 23: @mayhem You may have missed this in the merge - in f5ed0b7f0982fd54acb99230a662045ba1ab5cd2 this was moved to the top of the file and deleted from further down, but now the check also exists later in this file on line 71

admin/create-dumps.sh
```diff
@@ -32,12 +41,26 @@ function add_rsync_include_rule {
     echo "$SHA256_FILE_RULE" >> "$RULE_FILE"
 }
 
-echo "This script is being run by the following user: "; whoami
+# all cleanup code should be placed within this function
+function on_exit {
+    echo "Disk space when create-dumps ends:"; df -m
+
+    if [ -n "$TMPDIR" ]; then
+        rm -rf "$TMPDIR"
```
alastair, line 49: @atj is this dangerous if someone sets TMPDIR to `/` ? (in our containers we run things as root). Might be good to have some more validation here

alastair (commented):
two small comments

admin/create-dumps.sh
mayhem: Fixed.

alastair (approved):
Thansk!

admin/create-dumps.sh
atj: Good spot. Yes this is dangerous because there is no guarantee that `TMPDIR` was set by the script before `on_exit` is called. Unfortunately the mitigation added here can be trivially overcome because it doesn't take relative paths into account (e.g. `/../`).

My suggestion would be to explicitly set TMPDIR to the empty string at the top of the script (with an explanatory comment), so that even if an attacker has injected it into the environment it is guaranteed to be set to a safe value whenever `on_exit` is called.

PR #1401, opened 2021-04-21
Fix LB-510 correctly

mayhem:
This PR replaces #1377 with a better fix. Fixes LB-510.

### DIFF
--- a/listenbrainz/tests/integration/test_api.py
+++ b/listenbrainz/tests/integration/test_api.py
@@ -392,6 +392,11 @@ class APITestCase(ListenAPIIntegrationTestCase):
         self.assert400(response)
         self.assertEqual(response.json['code'], 400)
 
+        del payload["payload"][0]["track_metadata"]["track_name"]
+        response = self.send_data(payload)
+        self.assert400(response)
+        self.assertEqual(response.json['code'], 400)
+
     def test_empty_artist_name(self):
         """ Test for invalid submission in which a listen contains an empty track name
         """
@@ -400,6 +405,12 @@ class APITestCase(ListenAPIIntegrationTestCase):
         response = self.send_data(payload)
         self.assert400(response)
         self.assertEqual(response.json['code'], 400)
+
+        del payload["payload"][0]["track_metadata"]["artist_name"]
+        response = self.send_data(payload)
+        self.assert400(response)
+        self.assertEqual(response.json['code'], 400)
+
 
     def test_too_many_tags_in_listen(self):
         """ Test for invalid submission in which a listen contains more than the allowed
--- a/listenbrainz/tests/integration/test_api.py
+++ b/listenbrainz/tests/integration/test_api.py
@@ -397,8 +397,22 @@ class APITestCase(ListenAPIIntegrationTestCase):
         self.assert400(response)
         self.assertEqual(response.json['code'], 400)
 
+    def test_bad_track_name_format(self):
+        """Test for invalid submission in which a listen has a track_name field but it's not a string"""
+        with open(self.path_to_data_file('empty_track_name.json'), 'r') as f:
+            payload = json.load(f)
+            payload["payload"][0]["track_metadata"]["track_name"] = []
+        response = self.send_data(payload)
+        self.assert400(response)
+        self.assertEqual(response.json['code'], 400)
+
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/api_tools.py
```diff
@@ -154,12 +154,14 @@ def validate_listen(listen, listen_type):
 
     # Basic metadata
     try:
+        if len(listen['track_metadata']['track_name']) == 0:
+            raise APIBadRequest("required field track_metadata.track_name is empty.", listen)
         if not listen['track_metadata']['track_name']:
-            raise APIBadRequest("JSON document does not contain required "
-                                "track_metadata.track_name.", listen)
```
alastair, line 159: I'm not sure this fix will do exactly what we want. if `['track_name']` exists but is None (null in json) then the `len()` check will fail: `TypeError: object of type 'NoneType' has no len()`. I think for checking an empty value we're still better off with just the original check: `if not listen['track_metadata']['track_name']:`, which will trigger if the item is both None and an empty string (among other things). There is a further test lower down which checks if the artist name is a string... maybe we should also have a check for the track title.

We check for the absense of a track_name or artist_name key implicitly by use of the `except KeyError` at the end of this try block.

listenbrainz/tests/integration/test_api.py
```diff
@@ -383,6 +383,24 @@ def test_too_large_listen(self):
...
+
+    def test_empty_artist_name(self):
+        """ Test for invalid submission in which a listen contains an empty track name
+        """
+        with open(self.path_to_data_file('empty_artist_name.json'), 'r') as f:
+            payload = json.load(f)
+        response = self.send_data(payload)
+        self.assert400(response)
+        self.assertEqual(response.json['code'], 400)
+
```
alastair, line 403: it'd be good to also include tests for missing artist_name and track_name fields. We could use the same json files and just `del` the field before sending it.

I had a quick look and It seems like there are no tests anywhere for `validate_listen` (other than these two new ones), which seems like an oversight. We could add some more to a new `listenbrainz/webserver/views/test/test_api_tools.py`, which just calls this method without having to pass a listen through the whole webserver.

alastair (commented):
I think there are still some cases which we get wrong in the validation depending on the value of these fields

listenbrainz/tests/integration/test_api.py
mayhem: I'll do the former here, but lets open a ticket for adding better tests and get this PR out of the way.

listenbrainz/webserver/views/api_tools.py
mayhem: I reworked the whole section.

listenbrainz/webserver/views/api_tools.py
```diff
@@ -153,18 +153,24 @@ def validate_listen(listen, listen_type):
...
+    if 'artist_name' in listen['track_metadata']:
         if not isinstance(listen['track_metadata']['artist_name'], str):
             raise APIBadRequest("artist_name must be a single string.", listen)
-    except KeyError:
-        raise APIBadRequest("JSON document does not contain a valid metadata.track_name "
-                       "and/or track_metadata.artist_name.", listen)
+
+        if len(listen['track_metadata']['artist_name']) == 0:
+            raise APIBadRequest("required field track_metadata.artist_name is empty.", listen)
+        listen['track_metadata']['artist_name'] = listen['track_metadata']['artist_name'].strip()
```
MonkeyDo, line 170: Should we call strip *before* checking its length for the artist name?
Otherwise an artist name with only whitespace characters will be stripped to an empty string.
I thought you mentioned the artist name should never be empty.

mayhem: Good catch. Move the tests to after the strip. As far as artist names being empty, I'm now confused as to what it means for an artist to be empty. lol. I think the code is fine as it is now.

alastair (approved):
I added one more check to the track_name condition to verify that it's a string (this test already existed in artist_name) and added a few more tests to check this situation. 
Take a look over it, and if you're happy with those changes then this is fine by me to merge

mayhem:
Works for me. Thanks!

PR #1402, opened 2021-04-21
Implement global application context (API service injection & global app props)

MonkeyDo:
This PR is a trial run at using [React context](https://reactjs.org/docs/context.html) as a  sort of dependency injection for global app props (and an API service instance).

That way, we avoid passing down the API URL and other props down every level of the component hierarchy, and avoid each child component creating their own instance of the API service too. (Hadn't thought about that, but we've currently got ~30 instances of the API service on each of our pages !).

Instead, we use React context to initialize an instance of the API service once in the main page component (recentlistens.tsx, recommendations.tsx, playlist.tsx, etc.) that is then provided to children anywhere in the hierarchy.

This will also be useful for "global" props such as current logged in user, music services auth tokens, etc.
We should end up with a context like so:

```
{
  "APIService": -an instance of the APIService class-,
  "webSocketsServerUrl": "https://sockets.listenbrainz.org",
  "spotify": {
     "token": "blablabla"
   },
  "currentUser": {
    "id": 3641,
    "name": "mr_monkey",
    "auth_token": "fnord"
  }
}
```

Once we have better access to the API service wherever we want, we can solve issues such as LB-803, LB-728


alastair:
any reason why we don't just go straight to an object for the context so that it makes adding future items to the context easier? It would also mean that we don't have to do this.context for the api service, as you mentioned.

Having said that, the [react docs](https://reactjs.org/docs/context.html#consuming-multiple-contexts) say that in order to prevent unnecessary re-renders when a context changes, we should have multiple ones. Does that make sense? I can't remember if you planned to use this for the implementation of our "page-level props" or not? I suspect that these items will never change on page load, so likely not a problem here, right?

listenbrainz/webserver/static/js/src/FollowButton.tsx
```diff
@@ -187,5 +181,6 @@ class FollowButton extends React.Component<
     );
   }
 }
+FollowButton.contextType = APIContext;
```
alastair, line 184: the react docs say that this can be added as [static field of the class](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties), which in my opinion is a bit easier to follow. Do we have this plugin?

MonkeyDo: Not currently, I tried and got some errors. I can try adding the plugin though, I agree the static field syntax is better.

listenbrainz/webserver/static/js/src/FollowButton.tsx
```diff
@@ -93,37 +87,37 @@ class FollowButton extends React.Component<
...
-          });
-          if (updateFollowingList) {
-            updateFollowingList(user, "unfollow");
-          }
-        } else {
-          this.setState({ error: true });
+    const { unfollowUser } = this.context;
+
+    unfollowUser(user.name, loggedInUser?.auth_token!).then(({ status }) => {
+      if (status === 200) {
```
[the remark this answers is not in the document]
MonkeyDo: Yeah, not sure who wrote this, but implementation is not correct on the component side nor on the API side (which should do the status checking

amCap1712 (approved):
Looks awesome! :100:

PR #880, opened 2020-05-28
Spotify permissions: use array of strings instead of one long string

MonkeyDo:
# Problem

SpotifyPermission type allows one of a fixed number of strings.
 However, the permission string passed down to the react components is one concatenated string with the various permissions

# Solution
Instead, pass an array of permissions down to react components.

PR #1404, opened 2021-04-22
Improve the Spotify Integration code

amCap1712:
Now that we are moving to new schema structure which supports integrating multiple services, it makes sense to use this opportunity to rewrite existing code to promote reuse between different services and make adding new integrations easier.

### DIFF
--- a/listenbrainz/webserver/static/js/src/SpotifyPlayer.tsx
+++ b/listenbrainz/webserver/static/js/src/SpotifyPlayer.tsx
@@ -212,7 +212,7 @@ export default class SpotifyPlayer
         }
       })
       .catch((error) => {
-        handleError(error.message);
+        handleError(error);
       });
   };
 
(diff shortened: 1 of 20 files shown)


amCap1712:
@MonkeyDo I have applied the relevant commits from #880 to this PR. I also made some additional changes on the frontend handling of spotify permissions as an array as necessary.

listenbrainz/webserver/static/js/src/SpotifyPlayer.tsx
```diff
@@ -212,7 +212,7 @@ export default class SpotifyPlayer
         }
       })
       .catch((error) => {
-        handleError(error);
+        handleError(error.message);
```
MonkeyDo, line 215: Looking at the implementation of handleError, this shouldn't be needed:
https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/BrainzPlayer.tsx#L164-L178

MonkeyDo (commented):
The Javascript bits look good !

listenbrainz/webserver/static/js/src/SpotifyPlayer.tsx
amCap1712: Will fix.

amCap1712:
Cool. Let's close #880 in favour this PR then.

PR #1405, opened 2021-04-23
Add source maps to Webpack production build

MonkeyDo:
Publishing these source maps will give Sentry access to them to be able to display error in an un-minified context.

This is the easy solution, and since all our code is publicly hosted anyway… 🤷 

Reference: https://docs.sentry.io/platforms/javascript/guides/react/sourcemaps/hosting-publicly/
and https://webpack.js.org/configuration/devtool/


amCap1712 (approved):
:100:

PR #1406, opened 2021-04-23
Youtube player bug

MonkeyDo:
Sentry reported issue with the Youtube player: `Cannot read property 'video_id' of undefined`
https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/115090

We have some unsafe property access, easy to fix.

It also turns out the `onError` handler wasn't tied to the youtube player.
Fixing this, along with canceling the (admitedly hacky) `checkVideoLoaded` function we run with a timeout, should fix some of the issues with the player and at the very least let users know something went wrong.

### DIFF
--- a/listenbrainz/webserver/static/js/src/YoutubePlayer.tsx
+++ b/listenbrainz/webserver/static/js/src/YoutubePlayer.tsx
@@ -139,10 +139,9 @@ export default class YoutubePlayer
     const { onTrackNotFound } = this.props;
     // We use cueVideoById("") as a means to clear any playlist.
     // If search or loadVideoByID yield no results we can detect that nothing was found
-    if (
-      this.youtubePlayer.getVideoData &&
-      !this.youtubePlayer.getVideoData().video_id
-    ) {
+    const videoData =
+      this.youtubePlayer.getVideoData && this.youtubePlayer.getVideoData();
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/static/js/src/YoutubePlayer.tsx
```diff
@@ -139,10 +142,9 @@ export default class YoutubePlayer
     const { onTrackNotFound } = this.props;
     // We use cueVideoById("") as a means to clear any playlist.
     // If search or loadVideoByID yield no results we can detect that nothing was found
-    if (
-      this.youtubePlayer.getVideoData &&
-      !this.youtubePlayer.getVideoData().video_id
-    ) {
+    const videoData =
+      this.youtubePlayer.getVideoData && this.youtubePlayer.getVideoData();
```
amCap1712, line 146: I am a bit confused what this does.

amCap1712 (approved):
Looks Good!

listenbrainz/webserver/static/js/src/YoutubePlayer.tsx
MonkeyDo: I tried to adress it and make it more declarative, and improved the comment. Let me know if that's clearer

PR #1407, opened 2021-04-23
Fix bugs in stats pages

MonkeyDo:
If user stats have not been calculated, API will return 204 and an empty response.
The code that then processes the response currently tries to access a non-existing payload and throws an error.
I added some checks to abort early if there is no payload and return an appropriate empty  array.
Added some tests for this behavior.

Errors reported in Sentry:
https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/115123
and
https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/115127

### DIFF
--- a/listenbrainz/webserver/static/js/src/stats/UserListeningActivity.tsx
+++ b/listenbrainz/webserver/static/js/src/stats/UserListeningActivity.tsx
@@ -144,16 +144,16 @@ export default class UserListeningActivity extends React.Component<
     if (!data?.payload) {
       return result;
     }
-    if (range === "week") {
+    else if (range === "week") {
       result = this.processWeek(data);
     }
-    if (range === "month") {
+    else if (range === "month") {
       result = this.processMonth(data);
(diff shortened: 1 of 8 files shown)


listenbrainz/webserver/static/js/src/stats/UserListeningActivity.tsx
```diff
@@ -140,19 +140,23 @@ export default class UserListeningActivity extends React.Component<
     data: UserListeningActivityResponse
   ): UserListeningActivityData => {
     const { range } = this.props;
+    let result = [] as UserListeningActivityData;
+    if (!data?.payload) {
+      return result;
+    }
     if (range === "week") {
-      return this.processWeek(data);
+      result = this.processWeek(data);
```
amCap1712, line 148: Nit: Can use `else if`s here.

amCap1712 (approved):
Looks Good!

PR #1408, opened 2021-04-23
Catch unhandled promise rejections in ErrorBoundary

MonkeyDo:
React ErrorBoundaries do not catch errors thrown in async code (such as event listeners and promises).
That's…not great.
With this PR, we add a listener in ErrorBoundary listening to [the unhandledrejection event](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event), and setting the error to be displayed in the rendered component.

This allows us to get rid of some ugly hack (suggested in the React ticket[1]) we have peppered in the code, for example:
https://github.com/metabrainz/listenbrainz-server/blob/22c365c82fb75d692dbd8cafb91fd4c54462e868/listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx#L331-L336

This nifty unhandledrejection trick is courtesy of https://eddiewould.com/2021/28/28/handling-rejected-promises-error-boundary-react/

[1]: Some extra context: https://github.com/facebook/react/issues/11409

### DIFF
--- a/listenbrainz/webserver/static/js/src/ErrorBoundary.tsx
+++ b/listenbrainz/webserver/static/js/src/ErrorBoundary.tsx
@@ -15,9 +15,31 @@ export default class ErrorBoundary extends React.Component<
     this.state = { hasError: false, error: null };
   }
 
+  componentDidMount() {
+    // Add an event listener to the window to catch unhandled promise rejections & stash the error in the state
+    window.addEventListener(
+      "unhandledrejection",
+      this.promiseRejectionHandler.bind(this)
+    );
+  }
+
   componentDidCatch(error: Error) {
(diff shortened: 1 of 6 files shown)


alastair (commented):
This is beyond my skill level to review, but if it works then it should be fine. Any news on this mythical "concurrent mode" which the comments speak of in hushed tones?

MonkeyDo:
> This is beyond my skill level to review, but if it works then it should be fine. Any news on this mythical "concurrent mode" which the comments speak of in hushed tones?

Still in experimental phase.
But to be honest I failed to understand how concurrent mode relates to this issue at all, and suspect that was copied from a vague GH issue.

PR #1412, opened 2021-04-26
Remove BU from spark

amCap1712:


requirements_spark.txt
```diff
@@ -5,7 +5,6 @@ pika==0.13.0
 python-dateutil==2.8.0
 # Numpy has dropped support for Python 3.6.x from version 1.20.0. So we use version 1.19.5 till we upgrade Python version in Spark images.
 numpy==1.19.5
-git+https://github.com/metabrainz/brainzutils-python.git@v1.16.1
```
alastair, line 8: if we're going to remove this should we also remove flask, werkzeug, etc, etc?

amCap1712: my bad :sweat:

alastair (approved):
I grepped through the listenbrainz_spark directory for all of the removed dependencies and didn't find any of them there, looks good!

PR #1415, opened 2021-04-28
Upgrade redis to 6.2.2

mayhem:
As it says on the tin -- this will be needed for BU metrics writing. A docker-server-configs equivalent PR will be opened shortly.


alastair (approved):
Fine by me

PR #1418, opened 2021-04-29
Importer Progress Bar

jdaok:
110920110-dba91880-82d1-11eb-8a1c-0b8623d75d76
This PR adds a fun progress bar to the Last.fm importer.
(from https://github.com/metabrainz/listenbrainz-server/pull/1336)


listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -317,6 +323,21 @@ export default class LastFmImporter extends React.Component<
               Please don&apos;t close this page while this is running.
             </span>{" "}
             <br /> <br />
+            <div
+              className="progress"
+              style={{
+                height: `${7}px`,
```
MonkeyDo, line 329:
```suggestion
                height: "7px",
```

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
```diff
@@ -268,6 +268,12 @@ export default class LastFmImporter extends React.Component<
     this.startImport();
   };
 
+  progressBarPercentage() {
+    if (this.totalPages > this.numCompleted)
+      return (this.numCompleted / this.totalPages) * 100;
+    return 50;
```
MonkeyDo, line 274: Could you explain this 50 to me? Is it some sort of "unknown state"?

Also, should it return 100 if this.totalPages === this.numCompleted ?

MonkeyDo (commented):
That's a great looking addition, thanks !

listenbrainz/webserver/static/js/src/LastFMImporter.tsx
shivam-kapila: IMO this 50 will lead to the progress bar start from half itself?

MonkeyDo:
This doesn't seem to be working for me (on Brave browser 1.22.72).
I see no progress bar.
When I inspect the element, I can see the progress bar is there and the `aria-valuenow` property is updating as expected.
However the bar has no width at all; that's expected as the `aria-*` properties are only used for assistive technology (i.e. screen reader) and not for anything else. So there's currently nothing setting the width of the progressbar appropriately.

Manually setting a `width` percentage attribute in the styles of the progress bar element does work and show the bar with appropriate width, so setting the width style dynamically should work (with `<div … style={{width: myPercentageValue}}`.
While we're at it, I think it would be good for progressBarPercentage to return a rounded percentage value. That should work better with screen readers.

In the screenshot below you can see the absence of orange progress bar and on the right, the `aria-valuenow` attribute being updated.
image

MonkeyDo:
You can ignore the failing front-end test, that was fixed in #1483 
If you want you can merge master into this branch so that it goes away.

jdaok:
Thanks @MonkeyDo, with just the `aria-valuenow` property it still worked properly on my Chrome browser for some reason, but this is the right way to do it and its fixed now, and I checked on Brave too :)

MonkeyDo:
> Thanks @MonkeyDo, with just the `aria-valuenow` property it still worked properly on my Chrome browser for some reason, but this is the right way to do it and its fixed now, and I checked on Brave too :)

I figured it would be something along those lines. Thanks for fixing it !

MonkeyDo (approved):
Thanks for the fine tuning ! :)

PR #1419, opened 2021-05-02
Change tag pattern to trigger build image workflow

amCap1712:
Turns out @alastair was right all along, Github Actions indeed does not support `{}` in the tag pattern.


alastair (approved):
it's kind of weird that they use things that look almost exactly like regular expressions but aren't... anyway, looks good

PR #1420, opened 2021-05-02
Pass some common props using flask context processor

amCap1712:
#1402 adds supports for global app props in React using contexts. Similarly, on the server side, we can inject some commonly used properties in multiple places using a flask context processor. This way we do not forget to pass a needed prop and also get to deduplicate some logic.

We'll need to decide what props should be injected as global props. The api url and the sentry dsn are needed on every so it makes sense to include those in global props. The spotify user data is needed on most pages but not all so it can probably be injected either way. Props or global props can both be  used with or without React context however it can be used as a baseline to decide global props.


listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
```diff
@@ -586,21 +586,25 @@ export default class UserEntityChart extends React.Component<
...
   const propsElement = document.getElementById("react-props");
+  const globalPropsElement = document.getElementById("global-react-props");
   let reactProps;
+  let globalReactProps;
   try {
     reactProps = JSON.parse(propsElement!.innerHTML);
+    globalReactProps = JSON.parse(globalPropsElement!.innerHTML);
   } catch (err) {
-    // Show error to the user and ask to reload page
+    // TODO: Show error to the user and ask to reload page
```
amCap1712, line 596: This block of code is being repeated multiple files, I'll try to extract this into a separate function and reuse it.

listenbrainz/webserver/utils.py
```diff
@@ -25,3 +31,22 @@ def reformat_date(value, fmt="%b %d, %Y"):
...
+    spotify_data = {}
+    if current_user.is_authenticated:
+        current_user_data = {
+            "id": current_user.id,
+            "name": current_user.musicbrainz_id,
+            "auth_token": current_user.auth_token,
+        }
+        spotify_data = spotify.get_user_dict(current_user.id)
+    props = {
+        "api_url": current_app.config["API_URL"],
```
MonkeyDo, line 47: There's also the websockets url we can pass.
It's not used everywhere but seems like the sort of thing we would want in global props

amCap1712: Makes sense, I'll add it to global props.

listenbrainz/webserver/__init__.py
```diff
@@ -147,9 +148,14 @@ def create_app(debug=None):
     # Static files
     import listenbrainz.webserver.static_manager
     static_manager.read_manifest()
-    app.context_processor(lambda: dict(get_static_path=static_manager.get_static_path))
     app.static_folder = '/static'
 
+    from listenbrainz.webserver.utils import inject_global_props
```
alastair, line 153: any reason why this item (and the static manager) are imported in this method and not at the top of the file? I could believe that it's because of circular imports, but if that's not the case then let's put them at the top

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -731,29 +731,21 @@ export default class RecentListens extends React.Component<
   }
 }
 
+// RecentListens.contextType = GlobalAppContext;
```
alastair, line 734: ?

alastair (changes requested):
This is really cool! nice to see a huge reduction in duplicated code

listenbrainz/webserver/static/js/src/RecentListens.tsx
amCap1712: A merge gone wrong. Will fix.

listenbrainz/webserver/templates/base-react.html
```diff
@@ -7,5 +7,6 @@
 
 {% block scripts %}
   {{ super() }}
+  <script id="global-react-props" type="application/json">{{ inject_global_props()|safe }}</script>
```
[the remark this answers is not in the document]
alastair: The [example on the flask documentation](https://flask.palletsprojects.com/en/1.1.x/templating/#context-processors) shows a context processor which includes changing information (user object) on each request, so it must be possible, maybe we need to update the way that we call `@app.context_processor`

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -215,11 +215,27 @@ export function loadScriptAsync(document: any, scriptSrc: string): void {
   container.appendChild(el);
 }
 
+const parsePageProps = () => {
+  const domContainer = document.querySelector("#react-container");
+  const propsElement = document.getElementById("react-props");
```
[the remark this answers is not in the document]
MonkeyDo: Yes, it would indeed be clearer to have `react-page-props` as opposed to `react-global-props`

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -215,11 +215,27 @@ export function loadScriptAsync(document: any, scriptSrc: string): void {
...
+  const globalPropsElement = document.getElementById("global-react-props");
+  let reactProps;
+  let globalReactProps;
+  try {
+    reactProps = JSON.parse(propsElement!.innerHTML);
+    globalReactProps = JSON.parse(globalPropsElement!.innerHTML);
+  } catch (err) {
+    // Show error to the user and ask to reload page
```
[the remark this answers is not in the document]
MonkeyDo: Alastair and I discussed this a bit, and we both agree that's it should be impossible for the server to send malformed JSON when using `ujson.dumps`.

So any error caught here would really be an edge case, and I don't think we need a contingency plan for that. More likely than not a page refresh would sort whatever weird issue happened.
I think what should happen here is:

That being said, there are two things we can do to make sure we don't end up with a silent killer error:
- create a user-friendly error `There was a problem when loading the page: ${err.message}. Please refresh the page and contact us if the problem persists.` or something along those lines.
- print the error to the console (`console.error(…`)
- return an optional error in the object returned  by `parsePageProps`, formatted as [an Alert type](https://github.com/metabrainz/listenbrainz-server/blob/63826071c819aba376018e68985298a8af8bcb29/listenbrainz/webserver/static/js/src/types.d.ts#L10) 


We can then display that error by passing it in an array  as a prop `initialAlerts`. (see https://github.com/metabrainz/listenbrainz-server/blob/63826071c819aba376018e68985298a8af8bcb29/listenbrainz/webserver/static/js/src/AlertNotificationsHOC.tsx#L39)
The AlertNotificationsHOC component will take that alert array and display it to the user.

listenbrainz/webserver/utils.py
MonkeyDo: I think there might be a useful side-effect with the naming convention difference, in that we can easily identify whether the value is coming directly from the parsed props text or if it has already been passed as a component property.

MonkeyDo (approved):
All front-end bits working !
Currently deployed on beta.listenbrainz.org (as of June 1st)

alastair:
looks fine by me. let's merge it tomorrow?!

PR #1421, opened 2021-05-03
Timescale offline support

mayhem:
Gaga's fan is failing and we will need to replace it. This will take at least 15 minutes, so this PR adds support for having timescale be offline for while the remainder of the services remain up. In theory our web-containers should start up without having the DB connect string defined and give correct error messages. 

API calls will get a 503 error with the correct error message. Web users will be redirected to this error page if they attempt to load a page that requires the timescale listenstore:

Screen Shot 2021-05-03 at 4 30 41 PM

### DIFF
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -32,6 +32,10 @@ MB_DATABASE_URI = 'postgresql://musicbrainz_ro@{{.Address}}:{{.Port}}/musicbrain
 {{end}}
 {{end}}
 
+SQLALCHEMY_TIMESCALE_URI = ""
+TIMESCALE_ADMIN_URI=""
+TIMESCALE_ADMIN_LB_URI=""
+
 {{if service "timescale-listenbrainz"}}
 {{with index (service "timescale-listenbrainz") 0}}
 SQLALCHEMY_TIMESCALE_URI = "postgresql://listenbrainz_ts:listenbrainz_ts@{{.Address}}:{{.Port}}/listenbrainz_ts"
@@ -39,7 +43,6 @@ TIMESCALE_ADMIN_URI="postgresql://postgres:postgres@{{.Address}}:{{.Port}}/postg
 TIMESCALE_ADMIN_LB_URI="postgresql://postgres:postgres@{{.Address}}:{{.Port}}/listenbrainz_ts"
 {{end}}
 {{end}}
-SQLALCHEMY_TIMESCALE_URI = ""
 
 {{if service "pgbouncer-williams"}}
 {{with index (service "pgbouncer-williams") 0}}
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -39,6 +39,7 @@ TIMESCALE_ADMIN_URI="postgresql://postgres:postgres@{{.Address}}:{{.Port}}/postg
 TIMESCALE_ADMIN_LB_URI="postgresql://postgres:postgres@{{.Address}}:{{.Port}}/listenbrainz_ts"
 {{end}}
 {{end}}
+SQLALCHEMY_TIMESCALE_URI = ""
 
 {{if service "pgbouncer-williams"}}
 {{with index (service "pgbouncer-williams") 0}}
--- a/listenbrainz/db/timescale.py
+++ b/listenbrainz/db/timescale.py
@@ -21,6 +21,9 @@ def init_db_connection(connect_str):
     for more info.
     """
     global engine
+    if not connect_str:
+        return
+
     while True:
         try:
             engine = create_engine(connect_str, poolclass=NullPool)
--- a/listenbrainz/webserver/decorators.py
+++ b/listenbrainz/webserver/decorators.py
@@ -3,6 +3,7 @@ from datetime import timedelta
 from flask import request, current_app, make_response, redirect, url_for
 from six import string_types
 
+from listenbrainz.webserver import timescale_connection
 
 
 def crossdomain(origin='*', methods=None, headers=None,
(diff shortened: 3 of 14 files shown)


amCap1712 (approved):
The changes look good. What about the `timescale_writer` container? I think it handles connection errors but wanted to confirm

mayhem:
For the time that gaga is offline, we'll just take the timescale_writer offline and let listens pile up in RMQ.

listenbrainz/webserver/decorators.py
```diff
@@ -45,3 +46,34 @@ def wrapped_function(*args, **kwargs):
...
+
+def api_listenstore_needed(func):
+    """
+        This API decorator checks to see if timescale is online (by having
+        a DB URI) and if not, it raises APIServiceUnavailable.
+    """
+    @wraps(func)
+    def decorator(*args, **kwargs):
+        from listenbrainz.webserver.errors import APIServiceUnavailable
+        if not current_app.config["SQLALCHEMY_TIMESCALE_URI"]:
```
alastair, line 59: we don't need to make this change for today's release, but I think that this check would be better to verify if the timescale listenstore is None instead of looking at the config option. that way we can support bringing up LB with these error messages even if there is a URI set but the host is down (for example). I'll open a ticket after the review.

listenbrainz/webserver/templates/index/listens_offline.html
```diff
@@ -0,0 +1,17 @@
+{%- extends 'base.html' -%}
+{%- block title -%}Listens currently unavailable - ListenBrainz{%- endblock -%}
+{%- block content -%}
+  <h2 class="page-title">Listens currently unavailable</h2>
+
+  <p>
+     The database that contains listens for our users is currently offline
+     for maintenance. Please try again in a few minutes.
```
alastair, line 8: message saying that incoming listens are still being stored?

listenbrainz/webserver/timescale_connection.py
```diff
@@ -6,6 +6,10 @@
 
 def init_timescale_connection(logger, conf):
     global _ts
+
+    if not conf["SQLALCHEMY_TIMESCALE_URI"]:
+        return None
```
alastair, line 11: None not needed, can just be `return`. not important for this release

listenbrainz/db/timescale.py
```diff
@@ -21,6 +21,9 @@ def init_db_connection(connect_str):
     for more info.
     """
     global engine
+    if not connect_str:
```
alastair, line 24: Is this guard also needed? In [`init_timescale_connection`](https://github.com/metabrainz/listenbrainz-server/blob/8677e14353ab9efecd361d5ad4dff4b3db4cfc01/listenbrainz/webserver/timescale_connection.py#L7) we also have this guard.

It's odd to see that we appear to call init_db_connection twice - once in [`gen_app`](https://github.com/metabrainz/listenbrainz-server/blob/bbf6a832b30a7e3fa1b35f808299646d75719c32/listenbrainz/webserver/__init__.py#L112) and again in the [constructor for `TimescaleListenStore`](https://github.com/metabrainz/listenbrainz-server/blob/e3c0d66b9dede36bee02a7429d645f52b2b422c7/listenbrainz/listenstore/timescale_listenstore.py#L52). I don't think that both of these calls have to be here - if we didn't call it in gen_app then this guard wouldn't be needed.

Not important to fix for this release, I'll open a ticket

listenbrainz/webserver/views/api_compat.py
```diff
@@ -79,6 +80,7 @@ def api_auth_approve():
 
 @api_bp.route('/2.0/', methods=['POST', 'GET'])
 @ratelimit()
+@api_listenstore_needed
```
alastair, line 83: future improvement - some methods in this endpoint need ts and some don't. If we plan to support this API into the future then we should allow submissions to work.

This endpoint returns XML on failure not json

alastair (approved):
OK to release, but we should re-address some points here when we're not under pressure to get a fan replaced

listenbrainz/webserver/templates/index/listens_offline.html
mayhem: On it.

listenbrainz/webserver/timescale_connection.py
mayhem: Fixed.

listenbrainz/webserver/views/api_compat.py
mayhem: On the some bits vs others -- yeah, I saw this, but the API sees little use, so I think its ok for now.

listenbrainz/db/timescale.py
mayhem: I believe its better to keep this guard as well, for non-web timescale connections that might be attempted. As for the double init, lets do a ticket.

listenbrainz/webserver/views/api_compat.py
alastair: I updated this to only return an error message on user.getInfo (using the lastfm error messages), although this is broken anyway: LB-877

listenbrainz/webserver/decorators.py
alastair: done

consul_config.py.ctmpl
```diff
@@ -39,6 +39,7 @@ TIMESCALE_ADMIN_URI="postgresql://postgres:postgres@{{.Address}}:{{.Port}}/postg
 TIMESCALE_ADMIN_LB_URI="postgresql://postgres:postgres@{{.Address}}:{{.Port}}/listenbrainz_ts"
 {{end}}
 {{end}}
+SQLALCHEMY_TIMESCALE_URI = ""
```
amCap1712, line 42: If we move this line before the service block, we should be able to stop and start the containers without rebuilding images once the db is back up. Plus, if sometime the timescale db goes down we get a nice error page on the website itself rather than a hung up container.

alastair: interesting idea - this ties in with my future plans about making LB able to start up if it's missing its services.

One unknown here is that I don't know when the service will disappear from consul (or when it'll come back up). I understand that it'll happen when the docker container starts up again. One nice thing about an explicit value here is that we can turn it on when we want to, rather than it depending on the startup of another service... 

I did actually originally think that the value in consul was an explicit key that was pasted directly into this string, but now I see that it's not as easy as that.

PR #1422, opened 2021-05-03
LB-849: Require users to have a confirmed email before creating an LB account

amCap1712:
We have decided that users should have a verified email with MB in order to create a LB account. Existing users without an email will be shown a banner warning them to add an email to MB.

From 1 November, 2021, any listens from a user without an email associated to their account will not be accepted.


listenbrainz/webserver/views/api.py
```diff
@@ -51,6 +51,11 @@ def submit_listen():
     :resheader Content-Type: *application/json*
     """
     user = validate_auth_header()
+    if not current_app.config["DEBUG"] and \
+            current_app.config["REJECT_LISTENS_WITHOUT_USER_EMAIL"] and not user["email"]:
+        log_raise_400('The listens were rejected because the user does not has not provided an email.'
+                      'Please provide an <a href="https://musicbrainz.org/account/edit">email address</a> to submit'
```
mayhem, line 57: Very good error message. :)

listenbrainz/webserver/views/login.py
```diff
@@ -29,11 +27,21 @@ def musicbrainz_post():
...
     if provider.validate_post_login():
         user = provider.get_user()
-        db_user.update_last_login(user.musicbrainz_id)
-        login_user(user, remember=True, duration=datetime.timedelta(current_app.config['SESSION_REMEMBER_ME_DURATION']))
-        next = session.get('next')
-        if next:
-            return redirect(next)
+        if user:
+            if not user.email:
+                flash.warning(Markup('You have not provided an email address. Please provide an '
```
mayhem, line 32: It might be nice to not duplicate this string in the code.

mayhem (approved):
One nit, but fine to merge after that. We also need a docker server configs PR for the REJECT_LISTENS_WITHOUT_USER_EMAIL setting.

listenbrainz/webserver/login/provider.py
```diff
@@ -32,13 +33,27 @@ def get_user():
     data = s.get('oauth2/userinfo').json()
     musicbrainz_id = data.get('sub')
     musicbrainz_row_id = data.get('metabrainz_user_id')
-    user = db_user.get_or_create(musicbrainz_row_id, musicbrainz_id)
-    if user:
-        if not user['musicbrainz_row_id']:
-            db_user.update_musicbrainz_row_id(musicbrainz_id, data['metabrainz_user_id'])
-        return User.from_dbrow(user)
+
+    if not current_app.config["DEBUG"]:
```
alastair, line 37: I don't think that we should use this as the guard for identifying if we connect to the musicbrainz db or not. Let's instead use a connection/engine object, which gets set if the connection string is available in the config.

It'd be great to be able to connect to a real mb database to be able to test this during development, too. Can we have a system using a docker-compose overlay file that adds a new session? I guess the test database won't be good for this because it won't have user emails, right?

listenbrainz/webserver/views/login.py
```diff
@@ -29,11 +27,21 @@ def musicbrainz_post():
...
+                                     '<a href="https://musicbrainz.org/account/edit">email address</a> before '
+                                     '1 November 2021, or we will stop recording your submitted listens.'))
```
alastair, line 34: , or you will be unable to submit listens

listenbrainz/webserver/views/api.py
alastair: does this come out as a json message from the API? Not sure if html is a good idea

listenbrainz/webserver/login/provider.py
```diff
@@ -32,13 +33,27 @@ def get_user():
...
+
+    if user is None and user_email is None:  # new user without email in MB
         return None
+    elif user is None and user_email is not None:  # new user with email in MB
+        db_user.create(musicbrainz_row_id, musicbrainz_id, email=user_email)
+        user = db_user.get_by_mb_row_id(musicbrainz_row_id, musicbrainz_id)
+    else:  # old user, always update email from MB on login
+        user = dict(user)
+        user["email"] = user_email
+        db_user.update_user_email(musicbrainz_id, user_email)
```
alastair, line 52: should we also use this email address to send notifications to the user at the spotify importer, or is that for later?

alastair: we talked about having a daily job that updates emails if they've changed, can we add that too?

alastair (changes requested):

listenbrainz/webserver/login/provider.py
amCap1712: Yes, I have added the cron job now.

listenbrainz/webserver/login/provider.py
```diff
@@ -33,16 +36,23 @@ def get_user():
     data = s.get('oauth2/userinfo').json()
     musicbrainz_id = data.get('sub')
     musicbrainz_row_id = data.get('metabrainz_user_id')
-    user_exists_check = db_user.get_by_mb_id(musicbrainz_id)
-    user = db_user.get_or_create(musicbrainz_row_id, musicbrainz_id)
-    if user:
-        if user_exists_check is None:
-            ts.set_empty_cache_values_for_user(musicbrainz_id)
-        if not user['musicbrainz_row_id']:
-            db_user.update_musicbrainz_row_id(musicbrainz_id, data['metabrainz_user_id'])
```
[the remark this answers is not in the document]
amCap1712: I figured today that this check is useful in supporting the workflow if a user deletes their MB account and then creates it again with the same username. Will put the check back with a comment.

amCap1712:
We dump the `"user"` table in public dumps and private dumps. We should make the needed amendments to the columns dumped in each case so that email are excluded in public dumps but present in private dumps,
https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/db/dump.py#L58-L68
https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/db/dump.py#L115-L124

amCap1712:
The public dump import is currently broken, there seem to be multiple reason around this. We use COPY FROM in importing but the list of columns [here ](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/db/dump.py#L58-L68) is not complete. Also, this list needs to the name of columns not their value during import.

listenbrainz/config.py.sample
```diff
@@ -157,3 +157,6 @@ SESSION_REMEMBER_ME_DURATION = 365
 
 # listen dumps creation dir
 LISTEN_DUMP_TEMP_DIR_ROOT = os.path.join(os.path.dirname(os.path.realpath(__file__)), '..', 'tmp')
+
+REJECT_LISTENS_WITHOUT_USER_EMAIL = False
+REJECT_USERS_WITHOUT_EMAIL = False
```
alastair, line 162: These items should have documentation explaining what they do.

listenbrainz/webserver/login/provider.py
```diff
@@ -33,16 +36,23 @@ def get_user():
...
-        return User.from_dbrow(user)
+
+    user = db_user.get_by_mb_row_id(musicbrainz_row_id, musicbrainz_id)
+    user_email = None
+    if mb_engine:
+        user_email = mb_editor.get_editor_by_id(musicbrainz_row_id)['email']
```
alastair, line 43: we should also check that this is a _valid_ email address (column `email_confirm_date`) in the musicbrainz editor row

listenbrainz/webserver/login/provider.py
```diff
@@ -33,16 +36,23 @@ def get_user():
...
+    if user is None:
+        if current_app.config["REJECT_USERS_WITHOUT_EMAIL"] and user_email is None:
+            return None
+        db_user.create(musicbrainz_row_id, musicbrainz_id, email=user_email)
+        user = db_user.get_by_mb_row_id(musicbrainz_row_id, musicbrainz_id)
+        ts.set_empty_cache_values_for_user(musicbrainz_id)
     else:
-        return None
+        user["email"] = user_email
+        db_user.update_user_email(musicbrainz_id, user_email)
```
alastair, line 53: we're missing case here where the user doesn't have an email configured in musicbrainz, but already has an account in LB and signs in again.

listenbrainz/config.py.sample
alastair: Setting this item to True will prevent a user from logging in. I wonder if we should have an intermediate step here where from release date we always show a warning to users who have no email, but don't prevent them from logging in until the date that we set `REJECT_LISTENS_WITHOUT_USER_EMAIL` to True

alastair (changes requested):
I added some comments about the workflow. I'm testing dumps now.

listenbrainz/webserver/login/provider.py
alastair: after looking at the musicbrainz server email validation process, this might not be necessary. There appears to never be a situation where an email is set but email_confirm_date is null.

alastair: email is null and email_confirm_date is non-null on a public data dump. I'm not sure this is a usecase that we need to support

amCap1712: We don't use email_confirm_date in LB and do not store, so I don't think we need it in our dumps anyways.

alastair: I meant to say here that LB _should_ check email confirm date, in order to know that the email address that we're using is a valid email.
However, it turns out that musicbrainz always sets these in parallel. If the email is null then  email_confirm_date, and if email is set, there's always a email_confirm_date. The only situation where this doesn't happen is on a public data dump, but I don't think we need to support this case. We can probably ignore this comment.

listenbrainz/webserver/login/provider.py
amCap1712: fixed now.

listenbrainz/webserver/login/provider.py
amCap1712: Its intended that existing users without email be able to login. The warnings are not shown by this part of the code. Its done in https://github.com/metabrainz/listenbrainz-server/pull/1422/files#diff-7271692c5f71146001b6530c09733f46e8fa24e6183b335adbeb7e71428dca7bR31-R39

listenbrainz/webserver/views/login.py
```diff
@@ -29,8 +28,20 @@ def musicbrainz_post():
...
-        login_user(user, remember=True, duration=datetime.timedelta(current_app.config['SESSION_REMEMBER_ME_DURATION']))
+        no_email_warning = Markup('You have not provided an email address. Please provide an '
+                                  '<a href="https://musicbrainz.org/account/edit">email address</a> ')
+
+        if not user:
+            flash.error(no_email_warning + 'before creating a ListenBrainz account.')
+            return redirect(url_for('index.index'))
+
+        if current_app.config["REJECT_USERS_WITHOUT_EMAIL"] and not user["email"]:
+            flash.warning(no_email_warning + 'before 1 November 2021, or you will be unable to submit listens.')
```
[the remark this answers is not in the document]
amCap1712: A blog post is a good idea. I don't we have one yet though :/. We can add it here once we have one.

listenbrainz/config.py.sample
amCap1712: Added docs on the options work. As discussed elsewhere, existing users without email will always be able to login.

listenbrainz/db/timescale.py
```diff
@@ -8,7 +8,7 @@
 from listenbrainz import config
 
 # This value must be incremented after schema changes on replicated tables!
-SCHEMA_VERSION = 5
+SCHEMA_VERSION = 6
```
amCap1712, line 11: I think we also need to add an empty `''` for email like we do for auth_token to https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/db/dump.py#L58-L68

listenbrainz/db/dump.py
```diff
@@ -121,6 +121,7 @@
         'latest_import',
         'musicbrainz_row_id',
         'gdpr_agreed',
+        'email',
```
amCap1712, line 124: I think we also need to add an empty '' for email like we do for auth_token to https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/db/dump.py#L58-L68

amCap1712:
Pending Tasks
- Figure out to prevent email leaks (only send email when specific email is passed, use a class ?)
- Check email while fetching users for spotify import

alastair (approved):
I think we've iterated on this enough, looks good to me! Thanks for the work 🎉

PR #1424, opened 2021-05-04
Don't add GIT_COMMIT_SHA build to the label at the beginning of the build

alastair:
# Problem

This causes all build to invalidate the entire cache as the
GIT_COMMIT_SHA changes every time that we make a new image. 

# Solution

Add the version label at the end after all other items have changed

# Action

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 ARG PYTHON_BASE_IMAGE_VERSION=3.7-20210115
 FROM metabrainz/python:$PYTHON_BASE_IMAGE_VERSION as listenbrainz-base
 
+ARG PYTHON_BASE_IMAGE_VERSION
 ARG deploy_env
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
LGTM

alastair:
I set the vcs hash to empty at the beginning to override the base image version, and also added an additional ARG command necessary for setting the version in the based-on-image label

Dockerfile
```diff
@@ -1,11 +1,11 @@
 ARG PYTHON_BASE_IMAGE_VERSION=3.7-20210115
 FROM metabrainz/python:$PYTHON_BASE_IMAGE_VERSION as listenbrainz-base
 
+ARG PYTHON_BASE_IMAGE_VERSION
```
MonkeyDo, line 4: That so weird.

MonkeyDo (approved):
Reapproved!

PR #1425, opened 2021-05-04
LB-858 (I): Add button to report a user and an admin view to view all reports

amCap1712:
Users can easily report spammers with this feature and the admins can view a list of all such reports and act on those as needed. For details, see LB-858.

Screenshot from 2021-05-04 22-43-45
image

Screenshot from 2021-05-04 22-40-53


listenbrainz/webserver/templates/user/profile.html
```diff
@@ -2,9 +2,15 @@
...
 {% block profile_content %}
 
-    <div style="margin-top:20px">
+    <div style="margin-top: 20px; margin-bottom: 20px">
     	<img src="{{ url_for('static', filename='img/musicbrainz-16.svg') }}" />
     	<b><a href="https://musicbrainz.org/user/{{ user.musicbrainz_id }}">See profile on MusicBrainz</a></b>
+        {% if current_user.is_authenticated and current_user.musicbrainz_id != user.musicbrainz_id %}
+            <form method="POST" action="{{ url_for("user.report_abuse", user_name=user.musicbrainz_id) }}" style="float: right">
+                <input type="hidden" name="report" value="yes" />
+                <input class="btn btn-danger" type="submit" value="Report User" />
```
alastair, line 11: it'd be nice to have some kind of help here, e.g. a (?) that shows something on hover or click? explaining what report is for and what it does

listenbrainz/webserver/templates/user/profile.html
```diff
@@ -2,9 +2,15 @@
 
 {% block profile_content %}
 
-    <div style="margin-top:20px">
+    <div style="margin-top: 20px; margin-bottom: 20px">
     	<img src="{{ url_for('static', filename='img/musicbrainz-16.svg') }}" />
     	<b><a href="https://musicbrainz.org/user/{{ user.musicbrainz_id }}">See profile on MusicBrainz</a></b>
+        {% if current_user.is_authenticated and current_user.musicbrainz_id != user.musicbrainz_id %}
+            <form method="POST" action="{{ url_for("user.report_abuse", user_name=user.musicbrainz_id) }}" style="float: right">
```
alastair, line 9: can we make this ajaxy?

MonkeyDo:
As discussed on IRC, we're going to want to move this "report user" button to a react component so that we can jazz it up a bit, by which I mean:

- show a modal on click
- in the modal have a small help text to explain what this is for
- add an optional message input
- should we also send the id/name of the user that is doing the reporting?
- submitting or cancelling closes the modal and allows the user to return to their activities

The message and reporter id aren't currently part of the SQL schema, so let's figure out if it's something we would want.

MonkeyDo:
All ready now !

@amCap1712 can you please review too when you find some time?

admin/sql/updates/2021-05-04-add-report-users-table.sql
```diff
@@ -0,0 +1,27 @@
+BEGIN;
+
+CREATE TABLE reported_users (
+    id                  SERIAL,
+    reporter_user_id    INTEGER NOT NULL, -- FK to "user".id of the user who reported
+    reported_user_id    INTEGER NOT NULL, -- FK to "user".id of the user who was reported
+    reported_at         TIMESTAMP WITH TIME ZONE DEFAULT NOW(),
+    reason              TEXT
```
amCap1712, line 8: This migration has already been run on the database so that we could test this on beta. But at that time we did not have a `reason` column, there are two ways to about this.
1. Add another migration script to add the `reason` column using `ALTER TABLE`.
2. Add a `DROP TABLE IF EXISTS` at the beginning of this script.

listenbrainz/webserver/static/js/src/UserPageHeading.tsx
```diff
@@ -24,29 +24,35 @@ import * as Sentry from "@sentry/react";
...
+        {loggedInUser && user.name !== loggedInUser.name && (
+          <FollowButton
+            type="icon-only"
+            user={user}
+            loggedInUser={loggedInUser}
+            loggedInUserFollowsUser={loggedInUserFollowsUser}
+          />
+        )}
+      </h2>
+      <ReportUserButton user={user} alreadyReported={alreadyReportedUser} />
```
amCap1712, line 54: Why do we not check whether the user is logged in before showing the Report button as we do for Follow button? (I assume we do not want to show the report button to not logged in users.)

amCap1712:
@MonkeyDo Thanks for taking over this. Awesome Work! A couple of questions otherwise looks good to merge.

listenbrainz/webserver/static/js/src/UserPageHeading.tsx
MonkeyDo: No reason, I couldn't make up my mind :) But you're right, we just shouldn't show it if the user isn't logged in, all the more because we add another component (the modal) that would be pretty much useless.

admin/sql/updates/2021-05-04-add-report-users-table.sql
```diff
@@ -0,0 +1,31 @@
+BEGIN;
+
+-- We added a column 'reason' after initially running this update script on the prod database,
+-- so we first drop the table before recreating it.
+DROP TABLE IF EXISTS reported_users
```
amCap1712, line 5: Missing a `;` . :wink:

MonkeyDo: Eek !

listenbrainz/webserver/static/js/src/UserPageHeading.tsx
```diff
@@ -24,29 +24,37 @@ import * as Sentry from "@sentry/react";
...
+        {loggedInUser && user.name !== loggedInUser.name && (
+          <FollowButton
+            type="icon-only"
+            user={user}
+            loggedInUser={loggedInUser}
+            loggedInUserFollowsUser={loggedInUserFollowsUser}
+          />
+        )}
+      </h2>
+      {loggedInUser && (
```
amCap1712, line 54: We should probably have the `user.name !== loggedInUser.name` here as well.

MonkeyDo: I don't think that's an issue, since the API does that check.
I'd rather the button be visible in the same place at all times so that users know where to find it (it's likely that the most time they will spend is on their own page; seeing the button there cements it in place)

amCap1712: Makes sense. :+1:

PR #1426, opened 2021-05-05
Update timescale image used in local development

amCap1712:
The prod image is running 2.2. Updating the local image to match the one used in prod.


mayhem (changes requested):
I think we should pin exact versions for both production and for dev, so that we remain in lockstep.

mayhem (approved):

PR #1429, opened 2021-05-05
Fix typescript compilation

MonkeyDo:
My previous PR #1402 inadvertently broke production build due to a missing configuration option for Typescript transpilation.

Not exactly 100% sure why this didn't show up during dev builds at all, but I have some theories (post processing steps like minification or sourcemaps generation).

I also moved the babel configuration to `babel.config.js` (which already existed) to avoid confusion of having two separate configs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -148,7 +148,7 @@ COPY ./docker/services/cron/dump-crontab /etc/cron.d/dump-crontab
 RUN chmod 0644 /etc/cron.d/dump-crontab
 
 # Compile front-end (static) files
-COPY webpack.config.js .eslintrc.js tsconfig.json ./listenbrainz/webserver/static /static/
+COPY webpack.config.js babel.config.js .eslintrc.js tsconfig.json ./listenbrainz/webserver/static /static/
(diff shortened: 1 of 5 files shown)


amCap1712:
Applying 63e191b6e2a5be3d302ae4bc8c526cb7b8bedc2f on this branch seems to fix issues for me.

MonkeyDo:
For posterity, the issue revolves around a bleeding-edge typescript feature (type-only class fields e.g. [here](https://github.com/metabrainz/listenbrainz-server/blob/670bf582702890c8011a5cb0b738ed527b7b7479/listenbrainz/webserver/static/js/src/recommendations/RecommendationCard.tsx#L50)) and the interaction of the various presets and plugins we use when compiling.

The feature seems to be bit in flux at the moment, so the simple solution of this commit will be enough until Babel v8 comes out, which will support all this a lot better.

In short, this solution changes the order in which plugins are run and adds allowDeclareFields setting (which will be true by default in Babel v8).

I tried updating our various packages but found some more caveats.
Let's revisit this when Babel v8 comes out and it's more stable.

PR #1430, opened 2021-05-05
LB-882: Build production image in Github actions

amCap1712:
Sometimes a change may break the prod image but not the development workflow. We won't know about before merging the change and attempting to deploy from it. Hence, build the image in CI to catch such issues beforehand.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -30,3 +30,22 @@ jobs:
 
     - name: Run tests
       run: ./test.sh
+      
+  prod:
+      
+      runs-on: ubuntu-latest
+      
+      needs: test
+      
+      steps:
+      - uses: actions/checkout@v2
+      
+      - name: Login to Docker Hub
+        run: echo ${{ secrets.DOCKER_HUB_PASSWORD }} | docker login -u ${{ secrets.DOCKER_HUB_USERNAME }} --password-stdin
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -30,3 +30,22 @@ jobs:
...
+      
+      - name: Login to Docker Hub
+        run: echo ${{ secrets.DOCKER_HUB_PASSWORD }} | docker login -u ${{ secrets.DOCKER_HUB_USERNAME }} --password-stdin
+        continue-on-error: true
+
+      - uses: satackey/action-docker-layer-caching@v0.0.11
+        continue-on-error: true
+        
+      - name: Build production image
+        run: docker build --target listenbrainz-prod --build-arg deploy_env=test --build-arg GIT_COMMIT_SHA=test .
```
alastair, line 51: by setting the build args to these values, it means that we can't reuse the docker cache from the test build on the base image. Consider setting them to the same value as we do in docker-compose.yml

amCap1712: I have removed the deploy env variable and fixed some shellcheck warnings as well.

alastair:
I rebased against master and tested this - test.sh builds the base layer, and the build command in this job correctly uses the cached base layer and then build the prod layer.

alastair (approved):

PR #1436, opened 2021-05-07
Upgrade NodeJS to version 16

MonkeyDo:
We are currently using Node v10, which was deprecated a week ago on April 30th.

Settled for v16, which is not yet LTS but will be in October.

The main change starting Node v15 is that uncaught promise exceptions now interrupt the process.
The rest of the changes in this PR concern the way we run our tests and error handling.

When we usually run in a browser, our pages are wrapped in an ErrorBoundary component that catches unhandled rejections, so these changes only concern our testing framework, where we don't wrap the components and have to handle the uncaught errors in another way (in the jest setup file)

Similarly, when running in Node, `fetch` isn't available, so I set up `jest-fetch-mock` globally (we were using it but only in certain test files).


alastair (approved):
I don't understand node stuff, but I'm super happy to see us going all the way to 16.
Not sure I can give any feedback on the specifics here - if we deploy on test and it appears to work will that be enough?

MonkeyDo:
> I don't understand node stuff, but I'm super happy to see us going all the way to 16.
> Not sure I can give any feedback on the specifics here - if we deploy on test and it appears to work will that be enough?

I deployed on test, and checked that everything works fine.
I also took a minute to check on old browser versions (FF 33) and everything is working fine there too (ensuring the babel updates didn't break our long-trail-aware setup).

MonkeyDo:
All tests are passing, prod image build works, and manual testing after deploying on test.listenbrainz.org is working fine, so I'm going to merge this!

PR #1441, opened 2021-05-10
LB-783: Add URL to ListenBrainz dumps

alastair:
# Problem
There was no place in the documentation that linked to the URL where LB dumps are available

# Solution

Add a link

Screenshot 2021-05-10 at 10 49 33 AM

### DIFF
--- a/docs/dev/listenbrainz-dumps.rst
+++ b/docs/dev/listenbrainz-dumps.rst
@@ -9,6 +9,9 @@ and the incremental data dumps twice a week.
 Each dump contains a number of different files. Depending on your use cases,
 you may or may not require all of them.
 
+Dump mirrors
+============
+See the `ListenBrainz data page <https://listenbrainz.org/data>`_ for information about where to download the data dumps from.
 
 File Descriptions
 =================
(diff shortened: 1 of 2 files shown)


amCap1712:
While we are at it, let's add data.mb.org to https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/templates/index/data.html as well.

alastair:
hah, I didn't even know that there was a page on listenbrainz.org! Maybe we shouldn't have it in 2 different places. Thoughts?

amCap1712:
Neither did I till a few minutes ago. I think it makes sense to have it in one place only so as to avoid confusion. Let's just data.mb to the webpage then? If we want, we can link to the webpage from the documentation instead of directly adding the links there.

alastair:
Updated to make the data page more detailed, and docs just links to the data page:

Screenshot 2021-05-10 at 1 05 12 PM

Screenshot 2021-05-10 at 1 05 24 PM

amCap1712 (approved):
Looks good! Thanks.

PR #1442, opened 2021-05-10
Upgrade to BU 2.0.1

alastair:
# Problem

New release of BU 2.0, which includes some changes to the cache API.

# Action

I ran a full spotify import cycle locally and things seem to work (counts changing as expected - this was a core bit of what used the redis integration), but we should double-check that there are no other bits of code that I didn't touch that have also changed.


listenbrainz/listenstore/redis_listenstore.py
```diff
@@ -77,20 +77,20 @@ def update_recent_listens(self, unique):
 
         # Don't take this very seriously -- if it fails, really no big deal. Let is go.
         if recent:
-            cache._r.zadd(cache._prep_key(self.RECENT_LISTENS_KEY), recent, nx=True)
+            cache._r.zadd(cache._prep_key(self.RECENT_LISTENS_KEY, cache._glob_namespace), recent, nx=True)
```
alastair, line 80: these changes are wrong - prep_key adds the global namespace to the beginning of the key always, this argument should be None

alastair:
Just re-pushed to add BU 2.0.1 which allows the `namespace` argument to `_prep_key` to be None. 

I updated some calls to `cache._r.incr` to `cache.increment` because we also added the `amount` parameter. Tests now pass (including some which previously failed), but I want to double-check these cases a bit more carefully to ensure that cache values are incrementing as expected.

mayhem (approved):

mayhem:
I fixed the merge conflict and merge master in, lets see if that fixes the tests. It would be nice to merge this so I can add more stats to other parts.

PR #1443, opened 2021-05-10
Add ftp dump date check

mayhem:
I've moved the daily incremental dump creation to the morning (from mid-day) EU and added a check to ensure that full dumps at at most 16 days old and incremental dumps at most 26 hours old. If they are older than that, send mail to the observability list.


mayhem:
We missed more incremental dumps from being processed and not being reported because this remains unmerged. PLZ review?

alastair (changes requested):
This is fine, but can we move the logic of the ftp checker out of dump_manager?

amCap1712 (approved):
Looks Good!

PR #1444, opened 2021-05-10
Upgrade spotipy

alastair:
# Problem

dependabot is sleeping on the job and not rebasing https://github.com/metabrainz/listenbrainz-server/pull/1304 properly.

New version of spotipy is out. It depends on a newer version of requests than the one that we have.

# Solution

Upgrade spotipy and requests

# Action

I tested spotify integration on my local environment. Now playing, playback history, and the player work as expected.

I had a look at the release notes for requests and there doesn't seem to be any major things that would affect our few uses.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,7 +37,7 @@ werkzeug==1.0.1
 Flask-DebugToolbar==0.11.0
 Flask-UUID==0.2
 msgpack-python==0.5.6
-requests==2.23.0
+requests==2.25.1
 SQLAlchemy==1.3.23
 mbdata==25.0.4
 sqlalchemy-dst==1.0.1


requirements.txt
```diff
@@ -21,7 +21,7 @@ xmltodict == 0.12.0
 oauth2client == 4.1.3
 pika == 0.13.0
 git+https://github.com/metabrainz/brainzutils-python.git@v1.18.1
-spotipy == 2.16.1
+spotipy == 2.17.1
```
amCap1712, line 24: I see https://github.com/plamere/spotipy/releases `2.18.0` is out. Let's upgrade directly to that.

amCap1712: Also, relook at https://github.com/metabrainz/listenbrainz-server/pull/1361#discussion_r630037327

alastair:
We decided that because there's some complexity in making changes here along with #1361 that we'll just merge this one now and deal with the 2.18 upgrade later.

PR #1445, opened 2021-05-11
Run build image action on publishing release

amCap1712:
We are having some issues with action triggering on pushing a tag. Now attempting, to trigger the action on publishing a release.

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -1,9 +1,8 @@
 name: Build image and publish to Docker Hub
 
 on:
-  push:
-    tags:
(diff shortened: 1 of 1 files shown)


amCap1712:
Yes, https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release mentions that the GITHUB_SHA is of the last commit in the release and that's what the checkout action uses by default so we should be fine.

alastair:
ah, I see. I didn't quite understand how to read that table. what it is saying is "We will set the environment variable GITHUB_SHA to 'Last commit in the tagged release' "?

alastair (approved):

amCap1712:
Right.

PR #1446, opened 2021-05-11
Timescale ca policies

mayhem:
Now that I have some experience up continuous aggregate update policies, I've added the one that we have in production.


mayhem:
This will be replaced by a new PR.

PR #1447, opened 2021-05-11
Handle listen deletion correctly.

mayhem:
Deleting all listens did the wrong thing to delete the cache key, when they should've been NULLed. Deleting single listen was also doing stupid shit. This is worth a hot fix.

### DIFF
--- a/listenbrainz/listenstore/tests/test_timescalelistenstore.py
+++ b/listenbrainz/listenstore/tests/test_timescalelistenstore.py
@@ -426,8 +426,11 @@ class TestTimescaleListenStore(DatabaseTestCase):
         self.assertEqual(count + 1, int(cache.get(user_key, decode=False) or 0))
 
     def test_delete_listens(self):
-        self._create_test_data(self.testuser_name)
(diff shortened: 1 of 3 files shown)


listenbrainz/listenstore/tests/test_timescalelistenstore.py
```diff
@@ -426,30 +426,39 @@ def test_listen_counts_in_cache(self):
         self.assertEqual(count + 1, int(cache.get(user_key, decode=False) or 0))
 
     def test_delete_listens(self):
-        self._create_test_data(self.testuser_name)
-        listens, min_ts, max_ts = self.logstore.fetch_listens(user_name=self.testuser_name, to_ts=1400000300)
+        uid = random.randint(2000, 1 << 31)
```
amCap1712, line 429: I think it would make sense to fetch the cache values after deleting the listens and making sure they are what we expect.

mayhem: Good idea. Done.

amCap1712 (approved):

PR #1448, opened 2021-05-11
Add management command to see when next cron job is going to be

alastair:
# Problem

When we make a deployment, we don't always know if it's safe to update cron, because something might be running or about to run

# Solution

Make a management command that tells us when the next cron job will run

# Action

A few things still to check:

1. read both cron files
2. not sure if we should run this directly in the cron container using the _installed_ crontabs, or just the ones in the build tree. Theoretically they should be the same so I don't think that's a problem
3. double-check the timezone of the container that is running cron to make sure that we're using timezones/UTC correctly
4. add a "Most recently run" list - for example we don't want to restart if dumps just started 5 minutes ago
5. consider looking for the cron process and checking if it has any child processes, this is a clear indication that something is running

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,3 +1,5 @@
+import datetime
+
 import listenbrainz.db.dump_manager as dump_manager
 import listenbrainz.spark.request_manage as spark_request_manage
 from listenbrainz.listenstore.timescale_utils import recalculate_all_user_data as ts_recalculate_all_user_data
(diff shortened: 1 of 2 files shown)


amCap1712 (approved):
Looks good! Thanks!

alastair:
After a discussion today, we realised that it'd be good to also test that the username column is correct - we can do this by just checking if col 6 is one of our expected cron usernames.

PR #1449, opened 2021-05-11
Timestamp fixes

mayhem:
I'm continuing to debug the timescale listenstore and found a case where we never saved zero timestamps, which actually became quite important. This exposed a buggy test that should've been an integration test, so I moved that.


listenbrainz/listenstore/tests/test_timescalelistenstore.py
```diff
@@ -467,4 +451,12 @@ def test_000_delete_single_listen(self):
         min_ts, max_ts = self.logstore.get_timestamps_for_user(testuser_name)
         self.assertEqual(min_ts, 1400000000)
         self.assertEqual(max_ts, 1400000200)
-       
+
+    def test_for_empty_timestamps(self):
```
alastair, line 455: add a docstring describing this test to remind us why we have a special case here:
`"""Even if a user has no listens they should have the sentinel timestamps of 0,0 stored in the cache to avoid continually recomputing them"""`
or similar

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -144,11 +144,9 @@ def get_timestamps_for_user(self, user_name):
             t0 = time.monotonic()
             min_ts = self._select_single_timestamp(True, user_name)
             max_ts = self._select_single_timestamp(False, user_name)
```
alastair, line 146: can we add to the docstring of select_single_timestamp to say that it returns a sentinel 0 in the case that they user has no listens?

alastair (approved):
This looks good, I'd just like to see 2 small comments added to make it easier to understand how some methods work without having to read them through.

PR #1450, opened 2021-05-12
Delete listens as well when user is deleted using admin console

amCap1712:
When a user is deleted using the admin console, the current and default mechanism is to delete it from the users table. However, we also want the the listens of the deleted user to go away. Hence, override the default behaviour. We already have a `delete_user` method used when a user deletes their account, we can reuse the same here.

### DIFF
--- a/listenbrainz/model/user.py
+++ b/listenbrainz/model/user.py
@@ -1,4 +1,7 @@
 from datetime import datetime
+
+from flask import current_app
+from psycopg2 import OperationalError, DatabaseError
 from listenbrainz.model import db
 from listenbrainz.webserver.admin import AdminModelView
 from listenbrainz.webserver.views.user import delete_user
@@ -51,4 +54,9 @@ class UserAdminView(AdminModelView):
     ]
 
     def delete_model(self, model):
-        delete_user(model.musicbrainz_id)
+        try:
+            delete_user(model.musicbrainz_id)
+            return True
+        except OperationalError or DatabaseError as err:
+            current_app.logger.error(err, exc_info=True)
+            return False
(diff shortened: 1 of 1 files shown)


mayhem (approved):
Happy to merge when tests pass!

listenbrainz/model/user.py
```diff
@@ -47,3 +49,6 @@ class UserAdminView(AdminModelView):
         'musicbrainz_id',
         'musicbrainz_row_id',
     ]
+
+    def delete_model(self, model):
+        delete_user(model.musicbrainz_id)
```
alastair, line 54: the `base implementation of `delete_model`](https://flask-admin.readthedocs.io/en/latest/_modules/flask_admin/contrib/sqla/view/#ModelView.delete_model) returns `True` when it succeeds. It's possible that this value is used somewhere else so maybe we should also do it here.
Just checking - I guess that `after_model_delete` isn't useful for us because this is called after the db row itself is deleted, which means that delete_user won't be able to run?

amCap1712: We could use `after_model_delete`, I would expect delete_user to continue normally as deleting something that doesn't exist usually behaves. However, it seems counter intuitive to execute the delete twice and I did not want to have different logic to delete a user in different places.

alastair (approved):
not tested, but it looks fine

listenbrainz/model/user.py
```diff
@@ -47,3 +52,11 @@ class UserAdminView(AdminModelView):
         'musicbrainz_id',
         'musicbrainz_row_id',
     ]
+
+    def delete_model(self, model):
+        try:
+            delete_user(model.musicbrainz_id)
```
alastair, line 58: I'm a bit surprised that all of these methods use a username instead of a lb row id or musicbrainz row id... might be worth adding a ticket to clean this up. I guess it also relies on LB-866

mayhem: user_name are in use *everywhere*. :(

amCap1712: Yeah, probably makes sense to use an id. Will open a ticket.

PR #1452, opened 2021-05-12
Use docker cache-from to use caching in building production image

amCap1712:
Github Actions caching does not seem to work for tagged revisions due to cache scoping. Long story short, we workaround this by pulling the latest built image from docker hub and using that as a source of cache for next build.

### DIFF
--- a/docker/push.sh
+++ b/docker/push.sh
@@ -11,15 +11,23 @@ set -e
 
 cd "$(dirname "${BASH_SOURCE[0]}")/../"
 
-git describe --tags --dirty --always > .git-version
+GIT_COMMIT_SHA="$(git describe --tags --dirty --always)"
+echo "$GIT_COMMIT_SHA" > .git-version
 
 TAG=${1:-beta}
-echo "building for tag $TAG"
-docker build \
+
+function build_and_push_image {
+    echo "building for tag $1"
+    docker build \
         --cache-from metabrainz/listenbrainz:latest \
-        --tag metabrainz/listenbrainz:"$TAG" \
-        --tag metabrainz/listenbrainz:latest \
+        --tag metabrainz/listenbrainz:"$1" \
         --target listenbrainz-prod \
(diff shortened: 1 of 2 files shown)


alastair:
the documentation at https://docs.docker.com/engine/reference/commandline/build/#specifying-external-cache-sources says
> To use an image as a cache source, cache metadata needs to be written into the image on creation. This can be done by setting --build-arg BUILDKIT_INLINE_CACHE=1 when building the image.

do you know if this is necessary in our case? I guess your experiments showed that it was working as expected anyway, right?

docker/push.sh
```diff
@@ -15,7 +15,11 @@ git describe --tags --dirty --always > .git-version
 
 TAG=${1:-beta}
 echo "building for tag $TAG"
-docker build -t metabrainz/listenbrainz:"$TAG" \
+docker build \
+        --cache-from metabrainz/listenbrainz:latest \
```
alastair, line 19: should we only tag `:latest` if TAG is a v-[date], compared to beta/test/etc?

amCap1712:
>do you know if this is necessary in our case?
Right, I didn't add this because it worked without it.

docker/push.sh
amCap1712: I think that makes sense to do.

amCap1712:
@atj Can you please review the changes in the docker/push.sh script ?

atj (changes requested):
LGTM, but I would recommend you anchored the regexp.

docker/push.sh
```diff
@@ -28,6 +28,8 @@ function build_and_push_image {
 
 build_and_push_image "$TAG"
 
+# Only build and push the latest image if the TAG matches the date
+# pattern (v-YYYY-MM-DD.N) we use for production releases.
 if [[ $TAG =~ v-[0-9]{4}-[0-9]{2}-[0-9]{2}\.[0-9]+  ]]; then
```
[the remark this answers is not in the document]
amCap1712: Thanks! Fixed.

PR #1454, opened 2021-05-12
Rearrange imports in cosnul template to render config correctly

amCap1712:


consul_config.py.ctmpl
```diff
@@ -11,6 +8,10 @@ import json
     {{- end -}}
 {{- end -}}
 
+import os
```
alastair, line 11: maybe worth a comment here to remind us not to add things before the `define`?

alastair (approved):
great!

consul_config.py.ctmpl
amCap1712: Done

PR #1455, opened 2021-05-12
Improve continuous aggregate updates

mayhem:
This PR changes the default continuous aggregate policy to cover only the last year. Refreshes of the rest of the time is done in 3 year chunks one at a time in order to not eat huge amounts of disk space as the default continuous aggregate policy for all times does.

The downside is of this setup is that if someone's listen counts go out of whack during delete/re-import it will take up to 24 hours to get fixed. The expiry time of the post import listens counts have been adjusted to reflect this.


listenbrainz/listenstore/timescale_utils.py
```diff
@@ -110,3 +112,54 @@ def recalculate_all_user_data():
...
+
+        Example:
+
+           Assuming today is 2022-01-01 and this function is called for year_offset 1 and
+           year_count 1 then all of 2021 will be refreshed.
+    """
+
+    timescale.init_db_connection(config.SQLALCHEMY_TIMESCALE_URI)
+
+    end_ts = int(datetime.now().timestamp()) - 31536000
```
amCap1712, line 135: extract 31536000 in a variable like SECONDS_IN_A_YEAR?

amCap1712 (approved):

listenbrainz/listenstore/timescale_utils.py
mayhem: Fixed. Waiting for tests to pass, then will merge.

PR #1456, opened 2021-05-12
Fix bugs after Youtube Player release

amCap1712:
I have found two bugs that were introduced by the Youtube Player PR. One was trying to delete a user that doesn't exist. Added some context in the code comment. Second was that after refresh expired token in spotify updater, we didn't get the latest_listened_ts field for the user. I have added another method to fetch that field as well.

### DIFF
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -35,7 +35,7 @@ def notify_error(musicbrainz_row_id, error):
     if not user_email:
         return
 
-    spotify_url = url_for("profile.music_services_details")
+    spotify_url = current_app.config['SERVER_ROOT_URL'] + '/profile/music-services/details/'
     text = render_template('emails/spotify_import_error.txt', error=error, link=spotify_url)
     send_mail(
         subject='ListenBrainz Spotify Importer Error',
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -149,8 +149,8 @@ def make_api_request(user: dict, endpoint: str, **kwargs):
             spotipy_call = getattr(spotipy_client, endpoint)
             recently_played = spotipy_call(**kwargs)
             break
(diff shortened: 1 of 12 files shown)


amCap1712:
Also, added a missing prop at one place. Spotify reader is currently using this hotfix. beta.lb is also using this and youtube player seems to work fine there. I'll look into why tests are failing.

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -125,26 +125,28 @@ def _convert_spotify_play_to_listen(play, listen_type):
...
     """
     retries = 10
     delay = 1
     tried_to_refresh_token = False
 
     while retries > 0:
         try:
+            spotipy_client = spotipy.Spotify(auth=user['access_token'])
+            spotipy_call = getattr(spotipy_client, endpoint)
             recently_played = spotipy_call(**kwargs)
```
alastair, line 150: these could fail if the endpoint doesn't exist on the client, or if the kwargs aren't valid for that call. We should catch these specific cases and log them internally (e.g. imagine if spotipy changes the api at some point)

listenbrainz/spotify_updater/spotify_read_listens.py
```diff
@@ -149,6 +149,9 @@ def make_api_request(user: dict, endpoint: str, **kwargs):
             spotipy_call = getattr(spotipy_client, endpoint)
             recently_played = spotipy_call(**kwargs)
             break
+        except (AttributeError, TypeError) as err:
+            current_app.logger.critical("Invalid spotipy endpoint or arguments:", err, exc_info=True)
```
alastair, line 153: arguments to logger functions don't work like print. If you want to give an argument you need a % formatter placeholder in the string. this should be something like:

> current_app.logger.critical("Invalid spotipy endpoint or arguments: %s", err, exc_info=True)

However, do we even need to pass the exception if we also pass `exc_info=True`?

amCap1712: Makes sense. Fixing.

PR #1457, opened 2021-05-13
Update Webpack to v5 and transpile less to css

MonkeyDo:
### This was deployed on May 13th to test.LB.org for testing, everything looking fine !

The original aim was to compile our Less files to Css with Webpack to benefit from the cache-busting strategy we have in place (a manifest file and content hashed in the file names).

With Webpack 4 that was a bit problematic as WP is not really made to separately process and export files but rather to have the content inlined or the file name injected in an html file.

Webpack 5 brings new asset management modules that simplify the process and allows us to generate a single css file from a less file import, all without generating an unused javascript file along with it.

That also required a couple of updates (Less version and a few of webpack plugins) and a small refactor of our manifest file/ asset directory structure.

I followed the upgrade guide [here](https://webpack.js.org/migrate/5/) but to be honest the WP upgrade itself was smooth sailing as we don't currently do much processing in webpack itself (like chunking or asset optimization strategies)

Oh, and removing a css file that was there and unused from the very beginning of the project! (highlight.css)


listenbrainz/webserver/static/.gitignore
```diff
@@ -1,2 +1,3 @@
 !js/*
-js/dist/*
\ No newline at end of file
+js/dist/*
```
alastair, line 2: can we remove this if we no longer output anything in this location?

listenbrainz/webserver/static_manager.py
```diff
@@ -16,4 +16,4 @@ def read_manifest():
 def get_static_path(resource_name):
     if resource_name not in manifest_content:
         return "/static/%s" % resource_name
-    return "/static/js/dist/%s" % manifest_content[resource_name]
+    return manifest_content[resource_name]
```
alastair, line 19: before, the manifest file only contained the basename, but now it includes the full path?

listenbrainz/webserver/templates/base.html
```diff
@@ -17,7 +17,8 @@
       <link href="{{ url_for('static', filename='css/main.less') }}" rel="stylesheet/less" type="text/css" />
       <script src="{{ url_for('static', filename='js/lib/less.min.js') }}"></script>
     {% else %}
-      <link href="{{ url_for('static', filename='css/main.css') }}" rel="stylesheet" media="screen" />
+      <!-- The css file has a .less extension in the manifest file entry (due to its original name in wbepack entry) -->
```
alastair, line 20: use a flask comment: 
`{# The css file has a ...  #}`
so that we don't render it in the final template

webpack.config.js
```diff
@@ -26,21 +30,30 @@ module.exports = function (env) {
...
-      import: "/static/js/src/LastFMImporter.tsx",
-      userEntityChart: "/static/js/src/stats/UserEntityChart.tsx",
-      userReports: "/static/js/src/stats/UserReports.tsx",
-      userPageHeading: "/static/js/src/UserPageHeading.tsx",
-      userFeed: "/static/js/src/user-feed/UserFeed.tsx",
-      playlist: "/static/js/src/playlists/Playlist.tsx",
-      playlists: "/static/js/src/playlists/Playlists.tsx",
-      recommendations: "/static/js/src/recommendations/Recommendations.tsx",
+      // Importing main.less file here so that it gets compiled.
+      // All the Less/CSS will be exported separately to a main.css file
```
alastair, line 35: maybe one more comment saying that we don't have a standalone entrypoint for the less file so that it doesn't generate a superfluous js file

alastair (commented):
nice!

listenbrainz/webserver/templates/base.html
alastair: spelling: wbepack -> webpack

listenbrainz/webserver/static_manager.py
MonkeyDo: Correct. That's from the addition of output.publicPath in the webpack config.

Looks like with this new version you have to have a public path defined (maybe it would work with an empty string, haven't trie) or it defaults to "auto", so you get something like `"recentListens.js":"autorecentListens.5ab3956377725756fbf5.js"`, note the "auto" at the beginning of the file path.

It also seemed to me that it's better to defined it in one spot rather than two, and the webpack config is that place for that.

PR #1459, opened 2021-05-13
LB-895: Conditionally run tests based on paths modified

amCap1712:
For spark and frontend tests, I have added paths for which tests should run. For unit and integration tests, I have added paths to not run tests for. I have tested the paths for spark tests extensively at my own fork (https://github.com/amCap1712/listenbrainz-server/pull/15). The overall changes in a PR are considered while deciding whether the action should be run.

There is scope for refining the paths further but I think this is a good starting point. It is intentional that these patterns are only applicable to pull requests. Direct push to master and merge will trigger all actions.


amCap1712:
The valid file path patterns are documented here https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-file-paths.

.github/workflows/spark-tests.yml
```diff
@@ -5,6 +5,13 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ '*' ]
+    paths:
+      - 'listenbrainz_spark/**'
+      - 'requirements_spark.txt'
+      - 'requirements_development.txt'
+      - 'Dockerfile.spark'
+      - 'docker/docker-compose.spark*.yml'
+      - 'test.sh'
```
amCap1712, line 14: Spark code is also affected by the request consumer but currently we do not have any integrated testing so I haven't included that in the paths.

.github/workflows/frontend-tests.yml
```diff
@@ -5,6 +5,12 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ '*' ]
+    paths:
+      - '**.jsx?'
+      - '**.tsx?'
+      - '**.json'
```
amCap1712, line 11: There are some false positives, like the .json files used by integration tests and spark tests. Can exclude those by adding some negation patterns but not sure if its worth it.

.github/workflows/frontend-tests.yml
```diff
@@ -5,6 +5,13 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ '*' ]
+    paths:
```
alastair, line 8: building the production image is important in both js and python tests. Because we need it in both places, maybe we should pull it out into its own workflow file after all?

amCap1712: Yeah, makes sense.

MonkeyDo (approved):
LGTM, from what I can tell.
I'm in favor of merging it and adjusting what needs to be adjusted after we see how it performs

PR #1460, opened 2021-05-13
LB-897: Import of incremental dump fails successfully

amCap1712:
Send an email to listenbrainz-observability if dump imported successfully otherwise notify listenbrainz-exceptions.


listenbrainz/spark/handlers.py
```diff
@@ -197,10 +197,20 @@ def handle_dump_imported(data):
...
     dump_name = data['imported_dump']
     import_completion_time = data['time']
+
+    if dump_name:
+        template = 'emails/dump_import_notification.txt'
+        subject = 'A {} has been imported into the Spark cluster'.format(' '.join(data['type'].split('_')[1:]))
+        recipients = ['listenbrainz-observability@metabrainz.org']
+    else:
+        template = 'emails/dump_import_failure.txt'
+        subject = 'No dump imported into Spark cluster!'
```
mayhem, line 207: If there is any way we can make this statement more useful (download failed, no file available, hash failed), that would be excellent.

mayhem (changes requested):
Thanks for knocking this out, but lets take it one step further and suss out why it failed and include the in the message.

listenbrainz/webserver/templates/emails/dump_import_failure.txt
```diff
@@ -0,0 +1,18 @@
...
+
+These are some of the errors that occurred:
+
+{% for error in errors %}
+    {{ error }}
+{% endfor %}
+
+Go fix your shit!
+
+Your Friendly Neighbourhood Brainzbot 🤖
```
mayhem, line 17: Brainzbot doesn't sound all that friendly. :)

listenbrainz/spark/handlers.py
```diff
@@ -190,20 +190,22 @@ def handle_sitewide_entity(data):
...
     """
     if current_app.config['TESTING']:
         return
 
-    dump_name = data['imported_dump']
+    errors = data['errors']
     import_completion_time = data['time']
-    send_mail(
-        subject='A {} has been imported into the Spark cluster'.format(' '.join(data['type'].split('_')[1:])),
-        text=render_template('emails/dump_import_notification.txt', dump_name=", ".join(dump_name), time=import_completion_time),
```
mayhem, line 202: Is this template file still in use? If not, we should remove it in this PR.

mayhem (approved):
LGTM.

listenbrainz/spark/handlers.py
amCap1712: Removed.

PR #1464, opened 2021-05-17
Do not use cache-from for now

amCap1712:
Due a weird issue with incorrect CMD in images, revert using cache for now.


alastair (commented):
as we're not using `listenbrainz:latest` any more, how about we also remove the check where we build and tag this image, it'll make the action complete faster

alastair (approved):
good for now

PR #1465, opened 2021-05-17
Don't output any progress while compiling prod js files

alastair:
It takes up thousands of lines and gives nothing useful

# Problem

# Solution

# Action


MonkeyDo (approved):
:rocket:

PR #1467, opened 2021-05-17
Stop BrainzPlayer in other tabs/windows

MonkeyDo:
When a user has multiple tabs open, they can have sounds coming from these multiple sources.

This PR uses localStorage and associated events to communicate that command across browser windows to focus on the one you are currently using.

The decision to use localstorage events over [other methods](https://blog.bitsrc.io/4-ways-to-communicate-across-browser-tabs-in-realtime-e4f5f6cbedca), even though it looks like a hack, comes from simplicity of implementation and browser support.
For our simple use case it does everything we want it to and should work everywhere™

I already see one issue: when you are playing with spotify (player 1), and start a song with spotify too in another tab (player 2), that'll pause player 1. So far so good.
However if you return to player 1 and try to unpause, you'll be greeted with a spotify error. That's because playing in player 2 somehow unloaded the song being played in player 1.
Will have to investigate, but regardless it's not a result of this PR but something that is already happening in prod AFAICT.


alastair (approved):
This looks cool. As we mentioned in person, the only thing that I'd like to see is a comment that explains how events on localstorage only get triggered on other tabs for the same site, and not the tab that stored the value

shivam-kapila (approved):
This is amazing. Tested this and it works flawlessly. :tada:

PR #1471, opened 2021-05-18
Enhance and fix some "docblocks" for apis

mhor:
# Problem

Some params types are missing on api documentation.

# Solution

Enhance and fix some "docblocks" for apis.

# Action

~

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -327,6 +327,7 @@ def latest_import():
         }
 
     :param user_name: the MusicBrainz ID of the user whose data is needed
+    :type user_name: ``str``
     :statuscode 200: Yay, you have data!
     :resheader Content-Type: *application/json*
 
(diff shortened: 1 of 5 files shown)


listenbrainz/webserver/views/api.py
```diff
@@ -327,6 +327,7 @@ def latest_import():
         }
 
     :param user_name: the MusicBrainz ID of the user whose data is needed
+    :type user_name: ``str``
```
alastair, line 330: It took me a while to work out why this wasn't showing up properly. It turns out that we manually construct the doc items for this endpoint because it supports both GET and POST: https://github.com/metabrainz/listenbrainz-server/blob/master/docs/dev/api.rst
I updated it here, though we should do something about making sure that we don't duplicate documentation in two places.

While looking through the [sphinx flask documentation](https://sphinxcontrib-httpdomain.readthedocs.io/en/stable/#resource-fields) I also found that we should be using `:query` here instead of `:param` because it's a GET query parameter instead of a function parameter. I changed this in the doc file, but we also should slowly make this change everywhere else that query params are used.

alastair (approved):
Thanks for the additional improvements!

PR #1474, opened 2021-05-19
Replace mktemp with something we can control better.

mayhem:
I just don't understand how that old code was supposed to work. Also clear up confusing var names of TMP_DIR and TMPDIR.

### DIFF
--- a/admin/create-dumps.sh
+++ b/admin/create-dumps.sh
@@ -28,7 +28,7 @@ source "admin/functions.sh"
 
 # This variable contains the name of a directory that is deleted when the script
 # exits, so we sanitise it here in case it was included in the environment.
-TMPDIR=""
+DUMP_TEMP_DIR=""
 
 if [ "$CONTAINER_NAME" == "listenbrainz-cron-prod" ] && [ "$PROD" == "prod" ]
 then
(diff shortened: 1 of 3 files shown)


admin/create-dumps.sh
```diff
@@ -85,20 +85,23 @@ else
     exit
 fi
 
-TMPDIR=$(mktemp --tmpdir="$TEMP_DIR" -d -t "$SUB_DIR.XXXXXXXXXX")
+DUMP_TEMP_DIR="$TEMP_DIR/$SUB_DIR.$$"
+echo "TEMP_DIR is $TEMP_DIR"
```
mayhem, line 89: Lets keep these for the time being so we can see WTF is going on.

admin/create-dumps.sh
```diff
@@ -85,20 +85,23 @@ else
     exit
 fi
 
-TMPDIR=$(mktemp --tmpdir="$TEMP_DIR" -d -t "$SUB_DIR.XXXXXXXXXX")
+DUMP_TEMP_DIR="$TEMP_DIR/$SUB_DIR.$$"
```
amCap1712, line 88: Let's add a comment for what `$$` means.

amCap1712: I see it refers to the shell's `PID`. What are the chances of a name conflict? I do not know how cron launches shell but I assume it will launch a new one always so I guess the question is do `PID`s get reused and if yes what is the possibility of the conflict?

mayhem: The use of $$ is an old skool trick to avoid conflicts -- generally in this context we don't need to worry about the PIDs rolling over since they would have to make an entire roll-over pass while dumps are being generated. That isn't going to happen. 
Alternatively we could use use a UUID, but then the UUID program might not be found and it dumps break yet again.

amCap1712 (approved):

amCap1712 (approved):

PR #1475, opened 2021-05-19
Add timescale_writer listen submission metrics

mayhem:
Remove the old non functioning timescale-writer stats stuff and replace it with simple metric submission.


listenbrainz/timescale_writer/timescale_writer.py
```diff
@@ -126,6 +139,9 @@ def start(self):
                 sleep(self.ERROR_RETRY_DELAY)
                 sys.exit(-1)
 
+            init_cache(host=current_app.config['REDIS_HOST'], port=current_app.config['REDIS_PORT'], namespace=current_app.config['REDIS_NAMESPACE'])
+            metrics.init("listenbrainz")
```
alastair, line 143: this feels a bit weird to me. we're already running this in a flask app context (`with app.app_context()`, above). This means that we should probably init the cache in `gen_app` instead. I see that there's this really weird set of interdependencies:

gen_app calls create_redis: https://github.com/metabrainz/listenbrainz-server/blob/b31d54346a2a600c43ef629cf0662fe3fd1d296f/listenbrainz/webserver/__init__.py#L97

create_redis calls init_redis_connection: https://github.com/metabrainz/listenbrainz-server/blob/b31d54346a2a600c43ef629cf0662fe3fd1d296f/listenbrainz/webserver/__init__.py#L33

init_redis_connection makes a RedisListenStore: https://github.com/metabrainz/listenbrainz-server/blob/56ff715d6ec882fdd077c7fbcb17f14a73706fc5/listenbrainz/webserver/redis_connection.py#L15

RedisListenStore calls init_cache, which calls brainzutils.cache.init: https://github.com/metabrainz/listenbrainz-server/blob/86d67b8fd99da44ca1a3ccbe7dc4889118ae7371/listenbrainz/listenstore/redis_listenstore.py#L27

it's confusing that setting up the cache is done as a side-effect of creating a RedisListenStore. To make it easier to follow, we should instead call `brainzutils.cache.init` (or listenbrainz.utils.init_cache) from gen_app. This makes it clear that the cache can be used elsewhere in LB, not just related to RedisListenStore.

If RedisListenStore needs a pyredis object (i.e. `cache._r`), then we could pass that in as a constructor. However, I see that RedisListenStore uses `brainzutils.cache` for storing stuff anyway, so maybe this isn't even needed??

In any case, all of these points tie in with LB-879, and could potentially be fixed there. As a temporary solution, we could add the `metrics.init()` call to gen_app after it sets up redis.

alastair (commented):
the counting itself looks fine to me. However, I think we should add metrics.init to a more global location so that it's present for all processes that use a flask app.

amCap1712 (approved):
:100:

PR #1477, opened 2021-05-23
Document click scripts using sphinx-click 

amCap1712:
Some days ago we were looking for a way to import dumps in a development instance. After some time we found the `import_dump` command. Hence, adding the documentation for these commands to ListenBrainz documentation with the hope that these might come in handy sometime in the future. 

Screenshot from 2021-05-23 23-02-41


manage.py
```diff
@@ -1,280 +1,10 @@
-from datetime import datetime
-
+from listenbrainz.manage import cli
 import listenbrainz.db.dump_manager as dump_manager
 import listenbrainz.spark.request_manage as spark_request_manage
-from listenbrainz.listenstore.timescale_utils import recalculate_all_user_data as ts_recalculate_all_user_data, \
```
amCap1712, line 5: sphinx-click has a bug due to which it errors while trying to generate docs for multicommand groups. Therefore, I have moved actual commands to listenbrainz/manage.py and kept add dump and spark commands in root manage.py.

alastair:
The documentation already has this split where it contains two types of data - information for people who want to access the LB API, and information for people who want to develop the LB application. The documentation of these scripts kind of ties into the second part, but it's perhaps even a 3rd second ("information for us, as the people who run lb.org"). Should we try and improve the separation of items in the documentation (and especially in the menu bar on the left) in order to make it easier to understand what parts people need to pay attention to?

amCap1712:
I think separating documentation into further sections according to this makes sense. We should definitely do it.

alastair:
What are we going to do about this? Merge as-is, or spend some time reorganising the documentation first?

amCap1712:
This does not look urgent to merge, so let's reorganise first methinks.

PR #1478, opened 2021-05-24
Set permission of FTP dumps directory after copying into it

alastair:
In f5ed0b7f we changed from copying the contents of the FTP directory
with cp to using rsync, which also copies the permissions of the
source directory. This means that the permissions of $FTP_CURRENT_DUMP_DIR
were set to the permissions of $DUMP_DIR (700) instead of 755,
preventing world-read permissions and breaking ftp/http download.
In order to be explicit about what permissions are required, manually
chmod the directory at the same time that the files are chmod'd

# Problem

# Solution

# Action


amCap1712 (approved):
:100:

PR #1479, opened 2021-05-24
Log recommendations without users to debug only

amCap1712:
We cannot do anything about this. It can take upto a month for the deleted user to be removed from the dumps. Hence, lowering the log level to debug only to reduce noise in sentry.


alastair (approved):
fine, but we should consider making our logger calls consistent (%-formatting)

amCap1712:
Logging level and format made consistent with other similar log calls in file.

PR #1480, opened 2021-05-24
LB-767: stop 'sanitising' docker compose container names

amCap1712:
LB-767


test.sh
```diff
@@ -226,9 +216,6 @@ if [ "$1" == "spark" ]; then
 fi
 
 if [ "$1" == "int" ]; then
-    # Project name is sanitized by Compose, so we need to do the same thing.
-    # See https://github.com/docker/compose/issues/2119.
-    INT_COMPOSE_PROJECT_NAME=$(echo $INT_COMPOSE_PROJECT_NAME_ORIGINAL | awk '{print tolower($0)}' | sed 's/[^a-z0-9]*//g')
     INT_TEST_CONTAINER_NAME=listenbrainz
     TEST_CONTAINER_REF="${INT_COMPOSE_PROJECT_NAME}_${INT_TEST_CONTAINER_NAME}_1"
```
alastair, line 220: I don't see TEST_CONTAINER_REF used anywhere in the file, can we clear this up while we're at it? Maybe there are some other variables that aren't used either

alastair (approved):
Looks good. A few unused variables that we can get rid of.
Because the project name will change, this will have the effect of leaving behind some unused images and volumes, but it should only happen during tests, so I think it's fine to just leave it as-is. Developers can docker prune in their own time to remove them.

msb-develop.sh
```diff
@@ -9,13 +9,7 @@ fi
 
 
 COMPOSE_FILE_LOC=docker/docker-compose.yml
```
alastair, line 11: this is the wrong compose file for messybrainz. In fact, I can't even see a compose file any more, I suspect it disappeared when the repos were merged. Is there a reason to start up messybrainz as a standalone server?

I recommend that we delete this file entirely, and optionally add back docker-compose.yml into the messybrainz/ directory. There's no reason to have a develop.sh file if docker-compose.yml is named the default name and in a directory that can be automatically used as the project name

msb-test.sh
```diff
@@ -8,11 +8,8 @@
 # ./test.sh -d             clean test containers
```
alastair, line 8: not needed? Do msb tests get run while running regular test.sh?

alastair (commented):
Some more cleanups that we can make

msb-develop.sh
amCap1712: Indeed, this is wrong. I am not sure whether we need messybrainz as separate server. Maybe @mayhem has some thoughts about it?

msb-test.sh
amCap1712: MSB tests do run when running test.sh but msb-test.sh is not used anywhere.

amCap1712: Removed.

msb-develop.sh
amCap1712: Removed. It is still possible to run a local development server using `msb_manage.py` in case needed.

alastair: MessyBrainz used to run as an external server and LB connected over http to submit data, but this no longer happens. I'm happy to remove this functionality and add it back later if we need it.

amCap1712: Removed that as well.

alastair (approved):
tests work for me, looks good.

PR #1481, opened 2021-05-25
Log all requests received in spark

amCap1712:
Spark only logs `Received a request` message. Add some more info to the logs to identify the requests as well.

### DIFF
--- a/listenbrainz_spark/request_consumer/request_consumer.py
+++ b/listenbrainz_spark/request_consumer/request_consumer.py
@@ -44,6 +44,9 @@ class RequestConsumer:
             logger.error('Bad query sent to spark request consumer: %s', json.dumps(request), exc_info=True)
             return None
 
+        logger.info('Query: %s' % query)
+        logger.info('Params: %s' % str(params))
+
         try:
             query_handler = listenbrainz_spark.query_map.get_query_handler(query)
         except KeyError:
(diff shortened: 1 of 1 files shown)


listenbrainz_spark/request_consumer/request_consumer.py
```diff
@@ -44,6 +44,9 @@ def get_result(self, request):
             logger.error('Bad query sent to spark request consumer: %s', json.dumps(request), exc_info=True)
             return None
 
+        logger.info('Query: %s' % query)
+        logger.info('Params: %s' % str(params))
```
alastair, line 48: let's try and be consistent with our use of loggers. Have a read of the [logging documentation](https://docs.python.org/3/library/logging.html#logging.Logger.debug):

> The msg is the message format string, and the args are the arguments which are merged into msg using the string formatting operator.

This means that these lines should be `logger.info('Query: %s', query)`.

The main reason for this is so that the logger can skip doing string interpolation if the correct logging level isn't set.
I don't mind too much which way we do this - string interpolation should be pretty fast either way, however we _should_ be consistent in the way that we do it. I see other logger lines in this file which use the parameter format.

alastair (changes requested):
some comments on consistency

listenbrainz_spark/request_consumer/request_consumer.py
amCap1712: Makes sense. Fixed now.

alastair (approved):
Much better, thanks!

PR #1482, opened 2021-05-25
Make user reports JS component more robust

MonkeyDo:
Currently, if we don't have any data for the current time range (let's say the dafault 'week' range), the API will return data for the past week only.
The reports React component expects some data for the current week, and will throw an error trying to access a property of an non-existing object:
https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/119844

With this PR, we make sure that we can safely access those nested properties, and default to `undefined` otherwise.
This means that the chart legend will show a colour square for the current week (see top right in screenshot), but with no information next to it (where we would expect a time range for the current week), and the listen count and per day count will be 0 (which might not be true information: I might have listens this week but the stats haven't been calculated):

#### After:
image
#### Before:
image

### DIFF
--- a/listenbrainz/webserver/static/js/src/stats/UserListeningActivity.tsx
+++ b/listenbrainz/webserver/static/js/src/stats/UserListeningActivity.tsx
@@ -188,14 +188,14 @@ export default class UserListeningActivity extends React.Component<
     });
 
     this.setState({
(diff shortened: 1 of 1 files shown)


amCap1712 (approved):
:100:

MonkeyDo:
Ignoring failing front-end test which is unrelated and will be fixed in another PR.

PR #1484, opened 2021-05-25
Simplify importing the latest incremental dump into spark

amCap1712:
The current code to import incremental dumps into spark is non-trivial. I do not see the need for us to maintain it, given we can always import using the id. The current code tries to import all incremental dumps since the last full dump. When it does not find a dump, it assumes there are no further dumps and stops importing. Its possible to maintain the current functionality in the current state but it will increase the complexity further. I suggest we just import the latest dump. In case, there are dumps missing we can always import them manually by id. I'll open tickets to simplify the import code and add back this functionality at that point if it makes sense.

### DIFF
--- a/listenbrainz_spark/ftp/__init__.py
+++ b/listenbrainz_spark/ftp/__init__.py
@@ -2,9 +2,8 @@ import os
 import ftplib
 import hashlib
 import logging
-from collections import namedtuple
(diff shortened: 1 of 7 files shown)


mayhem (approved):

amCap1712:
Added back functionality to import all missing dumps since last full dump.

PR #1485, opened 2021-05-26
Startup bugfixes

alastair:
# Problem

Some bugs found when deploying the new startup changes

# Solution

# Action

### DIFF
--- a/docker/lb-startup-common.sh
+++ b/docker/lb-startup-common.sh
@@ -31,20 +31,17 @@ function send_sentry_message() {
   # Send a message to sentry
   # required globals:
   #   $SENTRY_SERVICE_ERROR_DSN: the DSN to report errors to
-  #   $PROJECT: the name of the project that is running (e.g. 'listenbrainz')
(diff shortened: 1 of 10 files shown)


docker/lb-startup-common.sh
```diff
@@ -31,20 +31,18 @@ function send_sentry_message() {
   # Send a message to sentry
   # required globals:
   #   $SENTRY_SERVICE_ERROR_DSN: the DSN to report errors to
+  #   $SENTRY_SERVICE_ERROR_ENVIRONMENT: the name of this project, used to set a sentry environment
+  #   $DEPLOY_ENV: the deploy environment, used to set a sentry environment
   #   $PROJECT: the name of the project that is running (e.g. 'listenbrainz')
```
amCap1712, line 36: Seems to be no longer used.

amCap1712 (approved):

docker/lb-startup-common.sh
alastair: wait, I thought I deleted that 😒

PR #1489, opened 2021-05-28
Remove lbdumps user from Dockerfile and crontab

amCap1712:
The lbdumps user was only needed because we used a storage box in production. With the server upgrade, we know have enough disk space on the server itself and do not need the storage box. Hence, we are getting rid of the storage box and related workarounds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,8 +70,11 @@ COPY . /code/listenbrainz
 ###########################################
 FROM listenbrainz-base as listenbrainz-prod
 
-# Create directories for cron logs, dump backups and FTP syncs
-RUN mkdir /logs /mnt/backup /mnt/ftp
+# Create directories for cron logs and dumps
+# /mnt/dumps: Temporary working space for dumps
+# /mnt/backup: All dumps
+# /mnt/ftp: Subset of all dumps that are uploaded to
+RUN mkdir /logs /mnt/dumps /mnt/backup /mnt/ftp
 
 # Install NodeJS and front-end dependencies
 RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,18 +70,8 @@ COPY . /code/listenbrainz
 ###########################################
 FROM listenbrainz-base as listenbrainz-prod
 
-
-# production sidenote: We create a `lbdumps` user to create data dumps because
-# the ListenBrainz servers on prod (etc. lemmy) use storage boxes [0] which
-# are owned by lbdumps on the host too.
-# [0]: https://github.com/metabrainz/syswiki/blob/master/ListenBrainzStorageBox.md
-RUN groupadd --gid 900 lbdumps
-RUN useradd --create-home --shell /bin/bash --uid 900 --gid 900 lbdumps
-RUN mkdir /logs && chown lbdumps:lbdumps /logs
-
-# Create directories for backups and FTP syncs
-RUN mkdir /home/lbdumps/backup /home/lbdumps/ftp
-RUN chown -R lbdumps:lbdumps /home/lbdumps/backup /home/lbdumps/ftp
+# Create directories for cron logs, dump backups and FTP syncs
+RUN mkdir /logs /mnt/backup /mnt/ftp
 
 # Install NodeJS and front-end dependencies
 RUN curl -sL https://deb.nodesource.com/setup_16.x | bash - && \
--- a/admin/create-dumps.sh
+++ b/admin/create-dumps.sh
@@ -134,7 +134,6 @@ DUMP_DIR=$(dirname "$DUMP_ID_FILE")
 DUMP_NAME=$(basename "$DUMP_DIR")
 
 # Backup dumps to the backup volume
-# Create backup directories owned by user "listenbrainz"
 echo "Creating Backup directories..."
 mkdir -m "$BACKUP_DIR_MODE" -p \
     "$BACKUP_DIR/$SUB_DIR" \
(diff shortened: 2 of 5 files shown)


admin/create-dumps.sh
```diff
@@ -136,12 +136,7 @@ DUMP_NAME=$(basename "$DUMP_DIR")
 # Backup dumps to the backup volume
 # Create backup directories owned by user "listenbrainz"
 echo "Creating Backup directories..."
-mkdir -m "$BACKUP_DIR_MODE" -p \
-         "$BACKUP_DIR/$SUB_DIR/" \
-         "$BACKUP_DIR/$SUB_DIR/$DUMP_NAME"
-chown "$BACKUP_USER:$BACKUP_GROUP" \
-      "$BACKUP_DIR/$SUB_DIR/" \
-      "$BACKUP_DIR/$SUB_DIR/$DUMP_NAME"
+mkdir -p "$BACKUP_DIR/$SUB_DIR/$DUMP_NAME"
```
alastair, line 139: this skips the `-m "$BACKUP_DIR_MODE"` option to set the mode. Was that intended?
Note the behaviour of `-p` and `-m` here, -m only sets the permissions on the last directory:

```
alastair@bono:~/t$ mkdir -m 700 -p foo/bar/baz
alastair@bono:~/t$ ls -l
total 4
drwxrwxr-x 3 alastair alastair 4096 May 31 13:39 foo
alastair@bono:~/t$ ls -l foo/
total 4
drwxrwxr-x 3 alastair alastair 4096 May 31 13:39 bar
alastair@bono:~/t$ ls -l foo/bar/
total 4
drwx------ 2 alastair alastair 4096 May 31 13:39 baz
```

Dockerfile
```diff
@@ -70,18 +70,8 @@ COPY . /code/listenbrainz
...
-# [0]: https://github.com/metabrainz/syswiki/blob/master/ListenBrainzStorageBox.md
-RUN groupadd --gid 900 lbdumps
-RUN useradd --create-home --shell /bin/bash --uid 900 --gid 900 lbdumps
-RUN mkdir /logs && chown lbdumps:lbdumps /logs
-
-# Create directories for backups and FTP syncs
-RUN mkdir /home/lbdumps/backup /home/lbdumps/ftp
-RUN chown -R lbdumps:lbdumps /home/lbdumps/backup /home/lbdumps/ftp
+# Create directories for cron logs, dump backups and FTP syncs
+RUN mkdir /logs /mnt/backup /mnt/ftp
```
alastair, line 74: I want to look into this in more detail - from what I understand, we don't need to make a directory if we want to mount a volume into it, however there are a few exceptions:

* if we define a directory as a volume with the docker `VOLUME` command, then docker will automatically create  an anonymous volume at runtime and copy the contents of this directory in the image to this volume the first time it's used. This is useful to do things like set permissions and add initial data (e.g. imagine that we want to add subdirectories for the incremental/full/spark folders automatically)
* If we attach a volume to this location when we `docker run`, it'll use this named volume as a location. Because we use the shared docker image, if we declare these locations with `VOLUME` then docker will start 3 new anonymous volumes for _every_ container that we start (web, timescale writer, spotify reader, etc). This seems a bit silly to me as well. We could prevent it from doing so by mounting the named volumes into all of these containers, but that also seems like the wrong decision.
* Keep in mind that if we don't `mkdir` here, and if we don't declare a location with `VOLUME` and we forget to mount a volume or if we accidentally run a dump command in another container then the dump will happen inside the context of the container and the files will disappear if the container is shut down. I think this is no worse than the situation we had before this PR

I think the best solution here may be to remove these specific mkdirs if we can just mount a volume into the location. I'll test it.

alastair (commented):
We also need to update docker-server-configs to replace the location that the volume is mounted to in `services.sh`, and also the locations/expected chown users in the consul config.

admin/create-dumps.sh
amCap1712: Right, I removed this because I thought the `chmod` a few lines down would handle it.
https://github.com/metabrainz/listenbrainz-server/pull/1489/files#diff-44f9cbe734a7b984e40c16e76ad9d48859088a6efb8501f5a01c91f4e5f335e2R145

admin/create-dumps.sh
```diff
@@ -136,12 +136,7 @@ DUMP_NAME=$(basename "$DUMP_DIR")
 # Backup dumps to the backup volume
 # Create backup directories owned by user "listenbrainz"
```
amCap1712, line 137: Also, this comment looks wrong. We don't have a `listenbrainz` user. Should I remove it?

admin/create-dumps.sh
alastair: here there is only a chmod for the files in the directory ($BACKUP_FILE_MODE) not for the directory itself ($BACKUP_DIR_MODE), however I'm not sure what the final goal of these specific permissions is, perhaps we can just leave it with the defaults in this case.

admin/create-dumps.sh
alastair: yes

alastair:
I also found references to `/home/listenbrainz` in `admin/config.sh.sample`, which should be updated to use the locations in /mnt

alastair:
`config.sh.ctmpl` has an item:
> RSYNC_INCREMENTAL_KEY='/home/{{template "KEY" "user"}}/.ssh/rsync-listenbrainz-dumps-incremental'

However, if we set the user to root, then the home directory of root will now be `/root`, not `/home/root`. In order to make this able to support both kinds of locations, perhaps we should add another consul setting `user_home` or similar, so that we can set it to `/root` or `/home/lbdumps` in consul depending on how we want to set up the users.

admin/create-dumps.sh
alastair: looking at this more, [we _do_ have a listenbrainz user](https://github.com/metabrainz/listenbrainz-server/blob/d540cdbabe2fcd1fae9d84b105fe2373c039a177/Dockerfile#L53), though we don't use it for dumps, and this comment is out of date, so it needs to be fixed anyway

PR #1490, opened 2021-05-28
Use global app props in React components

MonkeyDo:
Once  #1420 is merged, we will have global app props passed to the front-end.
With this PR the global props replace the previous mechanism of passing some props down the entire component tree, the continuation of #1402 .

The tests required some modifications along with it, mainly to do with providing the React context.


amCap1712 (approved):
:rocket:

PR #1492, opened 2021-06-01
Implement pinned recording table

jdaok:
This PR aims to adds the pinned recording table to the database and contains the corresponding SQL update script.

After the script is run the table looks like this:
image


mayhem:
We had a quick chat about this feature and sorted out the smaller details of how we want this to function. See our logs here: https://chatlogs.metabrainz.org/libera/metabrainz/2021-06-01/?msg=4814022&page=2

In summary, we'd like:

- Pinned tracks to stay pinned for a week, then "unpin" themselves.
- We'd like the user to be able to unpin a track, if they wish.
- We'd like the user to be able to delete pinned tracks.
- We'd like to avoid adding a periodic job to unpin tracks.

To accomplish all of the above, one change will be necessary to the schema. I'll leave a detailed comment in the file in question.

admin/sql/create_tables.sql
```diff
@@ -207,6 +207,15 @@ CREATE TABLE user_relationship (
     created             TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
 
+CREATE TABLE pinned_track (
+    id                      SERIAL, -- PK
+    user_id                 INTEGER NOT NULL, -- FK to "user".id
+    recording_mbid          UUID NOT NULL,
+    blurb_content           VARCHAR,
+    currently_pinned        BOOLEAN NOT NULL DEFAULT TRUE,
```
mayhem, line 215: Let's replace this with "pinned_until TIMESTAMP WITH TIME ZONE NOT NULL". This way when a track is pinned, we set this value to now() + 1 week. Then in the view that displays the user's pinned tracks we can look if the pin has expired and adjust the text accordingly. (e.g. This user has this pinned track X OR This user's last pinned track was X)

admin/sql/updates/2021-05-30-add-pinned-track-table.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+CREATE TABLE pinned_track(
+    id                      SERIAL, -- PK
+    user_id                 INTEGER NOT NULL, -- FK to "user".id
+    recording_mbid          UUID NOT NULL,
+    blurb_content           VARCHAR,
+    currently_pinned        BOOLEAN NOT NULL DEFAULT TRUE,
```
mayhem, line 8: Fix this too, obvs. :)

admin/sql/create_indexes.sql
```diff
@@ -42,4 +42,6 @@ CREATE INDEX user_id_ndx_user_timeline_event ON user_timeline_event (user_id);
 CREATE INDEX event_type_ndx_user_timeline_event ON user_timeline_event (event_type);
 CREATE INDEX user_id_event_type_ndx_user_timeline_event ON user_timeline_event (user_id, event_type);
 
+CREATE UNIQUE INDEX user_id_rec_mbid_ndx_pinned_track ON pinned_track (user_id, recording_mbid);
```
mayhem, line 45: We're unlikely to want to fetch data from this table using both user_id and recording_mbid. The normal pattern I expect is to ask "what recordings does this user have pinned"? So, this index should be on user_id only.

admin/sql/create_tables.sql
```diff
@@ -207,6 +207,15 @@ CREATE TABLE user_relationship (
     created             TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
 
+CREATE TABLE pinned_track (
+    id                      SERIAL, -- PK
+    user_id                 INTEGER NOT NULL, -- FK to "user".id
+    recording_mbid          UUID NOT NULL,
+    blurb_content           VARCHAR,
```
mayhem, line 214: In other places we would normally use TEXT here.

admin/sql/create_tables.sql
```diff
@@ -207,6 +207,15 @@ CREATE TABLE user_relationship (
     created             TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW()
 );
 
+CREATE TABLE pinned_track (
```
mayhem, line 210: Let's call this pinned_recording, since we're pinning recordings and not tracks (the MB db actually has tracks, which are distinct from recordings).

mayhem (changes requested):
This looks quite good, contrary to what you might think, regardless of the numerous comments on this PR. :) Some minor changes and this should be good to go.

admin/sql/create_indexes.sql
mayhem: Also, this should not be a UNIQUE index, otherwise a user can never pin the same track twice.

jdaok:
Thank you @mayhem, I've made the necessary changes. 
I read the discussion on IRC and my understanding for how the table will function is this (please correct me if I'm wrong)
- Instead of using the currently_pinned column, we will determine the active pin by comparing the current time (now()) to pinned_until.
- When unpinning a recording, we will update the pinned_until to the current time and it won't render as pinned anymore.

mayhem:
Yup. that sounds right!

mayhem (approved):
Looking good!

alastair (approved):
looks good, let's merge it!

PR #1493, opened 2021-06-02
Use sets to represent Spotify Permissions in backend

amCap1712:
Makes the check more robust as disucssed here:
https://github.com/metabrainz/listenbrainz-server/pull/1361#discussion_r630037327


alastair (approved):
cool, thanks!

PR #1494, opened 2021-06-02
Add retrying to revoking Youtube access tokens

amCap1712:
Try to revoke token with Google Auth API otherwise Google will consider the account to be still connected and will not send a refresh_token next time the user tries to connect again. We retry multiple times to reduce chances of failure. If it doesn't succeed proceed normally and just delete from our database.

Further, when a user connects again and the response from Google OAuth API is missing the refresh_token, we try to invalidate the access_token received in that response and ask the user to authenticate again.

### DIFF
--- a/listenbrainz/domain/youtube.py
+++ b/listenbrainz/domain/youtube.py
@@ -30,14 +30,23 @@ class YoutubeService(ExternalService):
         self.client_config = current_app.config["YOUTUBE_CONFIG"]
         self.redirect_uri = current_app.config["YOUTUBE_REDIRECT_URI"]
 
-    def add_new_user(self, user_id: int, token: dict):
-        external_service_oauth.save_token(user_id=user_id,
-                                          service=self.service,
-                                          access_token=token["access_token"],
-                                          refresh_token=token["refresh_token"],
-                                          token_expires_ts=int(token["expires_at"]),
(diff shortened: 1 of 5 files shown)


listenbrainz/domain/youtube.py
```diff
@@ -79,15 +90,29 @@ def refresh_access_token(self, user_id: int, refresh_token: str):
...
+            session = requests.Session()
+            session.mount("https://",
+                          HTTPAdapter(max_retries=Retry(total=3, backoff_factor=1, method_whitelist=["POST"])))
+            response = session.post(OAUTH_REVOKE_URL,
+                                    params={'token': access_token},
+                                    headers={'content-type': 'application/x-www-form-urlencoded'})
+            response.raise_for_status()
+        except RequestException:
+            error_msg = response.text if response else None
+            current_app.logger.error("Error while trying to revoke token: %s", error_msg, exc_info=True)
```
alastair, line 115: It might be useful to include a user id here, so at least we could go back through logs if we ever get a support request about this.

alastair (approved):
Looks good!

listenbrainz/domain/youtube.py
amCap1712: Done.

PR #1496, opened 2021-06-03
Remove browser Less(CSS) compilation

shivam-kapila:
Since we use webpack to generate CSS we can get rid of in-browser LESS compilation


MonkeyDo (approved):
I struggle to come up with a reason why we would want to keep that functionality. 

What's more, although we had that `COMPILE_LESS` config value it is in no way used in the webpack config, so we're always compiling the Less files.

There are some instances of `COMPILE_LESS` left in the messybrainz folder, but I'm not sure what we're supposed to do with that.
There's no JS files to compile there, so I assume we don't use the webpack build in any way, so we should probably keep the current messybrainz setup as-is

MonkeyDo:
Paging Dr.@alastair to be the Devil's advocate

shivam-kapila:
> I struggle to come up with a reason why we would want to keep that functionality. 
> 
> What's more, although we had that `COMPILE_LESS` config value it is in no way used in the webpack config, so we're always compiling the Less files.
> 
> There are some instances of `COMPILE_LESS` left in the messybrainz folder, but I'm not sure what we're supposed to do with that.
> There's no JS files to compile there, so I assume we don't use the webpack build in any way, so we should probably keep the current messybrainz setup as-is

Yeah. I didn't remove the MessyBrainz config for the same reason.

PR #1497, opened 2021-06-04
LB-875: Github action linting doesn't report errors

amCap1712:
1) Fix the extensions passed to ESLint. According to docs, dot should be appended in front of each comma separated extension.
2) Make ESLint output to stdout in human readable format.
3) Add --color to force emitting colours in CI logs.
4) Always error on warning. When we run LB locally, static builder container will try to fix all warnings and errors. This causes unnecessary changes in diffs and commits.
5) Register a problem matcher to annotate eslint warnings and errors.

Sample Action Run: https://github.com/amCap1712/listenbrainz-server/actions/runs/905685879


.github/eslint.json
```diff
@@ -0,0 +1,22 @@
+{
```
amCap1712, line 1: Copied from https://github.com/actions/setup-node/blob/master/.github/eslint-stylish.json . Do we need to add a copyright notice and if yes where? I would have added a comment in the file but json doesn't support comments.

MonkeyDo (approved):
Looks pretty good to me, and the [tests on your fork](https://github.com/amCap1712/listenbrainz-server/actions/runs/905685879) were very convincing.

amCap1712:
I have updated the PR to use the official `setup-node` action (from which I had taken the .json file) to register problem matcher. The action will not install Node but only register problem matchers. According to its [docs](https://github.com/actions/setup-node/tree/master#v2-beta) this is a supported use case.

.github/workflows/frontend-tests.yml
```diff
@@ -32,14 +32,27 @@ jobs:
...
+    # We run eslint inside a docker container, the frontend sources are mounted
+    # as a volume at /code/static. Therefore, the file name in the eslint output
+    # has file paths beginning with /code/static. As a result, Github Actions is
+    # unable to display those inline with code. To fix, this the output of eslint
+    # is piped to sed which replaces /code/static with listenbrainz/webserver/static .
+    # Finally, piping eats up the exit status so we exit manually with the correct status.
     - name: Run linter
-      run: ./test.sh fe -f
+      run: |
+        ./test.sh fe -f | sed 's/\/code\/static/listenbrainz\/webserver\/static/g'
```
MonkeyDo, line 54: Feels very black-magic-ey, and not in a comfortable way, but I don't have any good alternative to offer (short of changing all the paths we use in the docker mounts).

MonkeyDo (approved):
I think we should go for it and fix any issues that might pop up.
In any case this is already (looking at your demo on your fork) a hundred times better than the absence of reporting we had !

.github/workflows/frontend-tests.yml
amCap1712: Yeah, that was the only alternative I could think of and it seemed not to be much more far reaching.

PR #1500, opened 2021-06-08
LB-786: Upgrade SocketIO client and server libraries

amCap1712:
Upgrade Flask-SocketIO to 5.x, it requires the frontend socket-io.client library to be also upgraded to 3.x at least. Things seemed to work, hence socket-io.client has been upgraded to 4.x. Types are also provided by the socket-io.client, hence remove the deprecated package @types/docket-io.client package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "react-select": "^3.1.0",
     "react-sortablejs": "^6.0.0",
     "react-youtube": "^7.12.0",
-    "socket.io-client": "^2.2.0",
+    "socket.io-client": "^4.1.2",
     "sortablejs": "^1.12.0",
     "time-ago": "^0.2.1"
   },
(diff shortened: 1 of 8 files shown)


alastair (commented):
looks like there's still a failing frontend test, but otherwise if you tested the backend implementation and it works then that's fine by me!

amCap1712:
Right, I tested on test.lb and it worked. I was unable to fix the test though, I hope @MonkeyDo can help :)

MonkeyDo (approved):
Tests fixed !

package.json
```diff
@@ -63,7 +63,7 @@
     "react-select": "^3.1.0",
     "react-sortablejs": "^6.0.0",
     "react-youtube": "^7.12.0",
-    "socket.io-client": "^2.2.0",
+    "socket.io-client": "^4.1.2",
```
MonkeyDo, line 66: This introduced a regression for older browsers —such as CatQuest's FF44 — either the newer version of this library or one of its dependencies isn't compiled to ES5.
I'll open a PR to fix it.

amCap1712: Oh! Thanks!

MonkeyDo: For posterity -> #1524

PR #1508, opened 2021-06-15
Improve pydantic model validation

jdaok:
This PR aims to improve pydantic model validation by making them stricter, and implements validator reuse across models.

Summary of changes made:
- Renamed `check_rec_mbid_msid_is_valid_uuid()` to `check_valid_uuid()`
- Updated some invalid UUID's that caused tests to fail
- Updated int type to stricter NonNegativeInt type in most models
- Updated str type to stricter constr(min_length=1) type in most models


amCap1712:
Thanks for adding validators in other models as well. We also have a bunch of pydantic models in https://github.com/metabrainz/listenbrainz-server/tree/master/data/model . Can you please take a look and update those as well ? 

IIRC, we just use `str` as type for `mbid`s in these models at some places which is not ideal. Changing the type from `str` to `uuid.UUID` might here need some extensive changes, (some of these models are used in spark as well) so feel free to leave those out for now.

listenbrainz/db/model/utils.py
```diff
@@ -0,0 +1,42 @@
+import uuid
```
amCap1712, line 1: I see this file has already been merged in #1503. However, I think `validators.py` is a better name for this file. Can you please rename it?

jdaok:
@amCap1712 Done, I opened [LB-917](https://tickets.metabrainz.org/browse/LB-917) to track the task you mentioned.

amCap1712:
Updated the PR to work with new models.

PR #1510, opened 2021-06-16
Improve pinnedRecording database queries + models

jdaok:
This PR aims to make several improvements to the database models and queries in [PR 1503:](https://github.com/metabrainz/listenbrainz-server/pull/1503)

**1) Separate PinnedRecording and WritablePinnedRecording models**
  
  - When pinning a recording there wouldn't be a reason to set 'created' to anything other than the current datetime, but the query functions return a validated object and we wouldn't want the 'created' value to be replaced in the validator. 
  - From now on all pinned recordings would have a 'created' time set automatically to now(). (WritablePinnedRecording model)

**2) Unpin and Delete queries return True/False**: improves the responses and error messages at the API level when unpinning and deleting tracks.

**3) Get_pins_for_user_following()**

### DIFF
--- a/listenbrainz/db/pinned_recording.py
+++ b/listenbrainz/db/pinned_recording.py
@@ -2,7 +2,7 @@ import sqlalchemy
 import json
 
 from listenbrainz import db
-from listenbrainz.db.model.pinned_recording import PinnedRecording, PinnedRecordingSubmit
+from listenbrainz.db.model.pinned_recording import PinnedRecording, WritablePinnedRecording
 from typing import List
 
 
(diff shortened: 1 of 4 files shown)


mayhem (changes requested):
Other than our convo in IRC about having one class derive from the other, this looks pretty good.

mayhem (approved):
Looks good to me now, thanks!

listenbrainz/db/pinned_recording.py
```diff
@@ -40,28 +41,37 @@ def unpin(user_id: int):
...
-        connection.execute(sqlalchemy.text("""
+        result = connection.execute(sqlalchemy.text("""
             UPDATE pinned_recording
                SET pinned_until = NOW()
              WHERE (user_id = :user_id AND pinned_until >= NOW())
             """), {
             'user_id': user_id,
             }
         )
+        return result.rowcount == 1
```
amCap1712, line 59: While its unlikely the user will have more than one pin at a time (unless something is wrong on our end), I think its still better do this check better as `result.rowcount > 0`.

amCap1712 (approved):
LGTM

PR #1511, opened 2021-06-16
Validate mbids of listens submitted through AudioScrobbler 1.2 API

amCap1712:
We discovered that some of the clients using AudioScrobbler 1.2 API are submitting invalid values as mbids. In the AudioScrobbler 2.0 and ListenBrainz API, we reject such listens but AS 1.2 is a legacy api and many clients out there are probably unmaintained. Considering this, we have decided to drop the invalid mbid fields but accept the listens.


mayhem (approved):
Thanks!

amCap1712:
We already some invalid data in the DB both messybrainz and listens table possibly elsewhere, we'll need to come up with a way to fix that data.

PR #1512, opened 2021-06-17
Remove --last-dump-id flag to simplify dumps code

amCap1712:
We have decided to simplify the dumps code, by default a dump will always add an entry to data_dump table. In case something goes wrong, we will pass in the `--dump-id` manually.

### DIFF
--- a/admin/create-dumps.sh
+++ b/admin/create-dumps.sh
@@ -77,6 +77,9 @@ if [ -z $DUMP_BASE_DIR ]; then
 fi
 
 DUMP_TYPE="${1:-full}"
+# consume dump type argument so that we can pass the remaining arguments to
+# the python dump manager script
+shift
 
 if [ "$DUMP_TYPE" == "full" ]; then
     SUB_DIR="fullexport"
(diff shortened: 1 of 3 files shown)


admin/create-dumps.sh
```diff
@@ -77,6 +77,9 @@ if [ -z $DUMP_BASE_DIR ]; then
 fi
 
 DUMP_TYPE="${1:-full}"
+# consume dump type argument so that we can pass the remaining arguments to
+# the python dump manager script
```
alastair, line 81: in this case it'd be good to have some kind of documentation for this script that explains that we're able to do this. Perhaps just comments at the top of the file like we do in develop.sh and test.,sh is enough for now

alastair (commented):
I don't understand the behaviour of this flag, but it sounds like you've gone through it a few times and worked out what was going on, so I'm happy to go with your suggestion 👍

mayhem (approved):
Yeah, for simplification!

admin/create-dumps.sh
amCap1712: Added a few line on script usage at top.

PR #1513, opened 2021-06-17
Terminate cron

mayhem:
This PR is the sister PR to https://github.com/metabrainz/docker-server-configs/pull/173 .
This add the admin/cron_lock.py script that manages the cron lock file and is used by the docker-server-configs script/terminate-lb-cron.sh script.

### DIFF
--- a/admin/create-dumps.sh
+++ b/admin/create-dumps.sh
@@ -63,6 +63,9 @@ function on_exit {
         local duration=$(( $(date +%s) - START_TIME ))
         echo "create-dumps took ${duration}s to run"
     fi
+
+    # Remove the cron lock
+    admin/cron_lock.py unlock-cron
 }
 
 trap on_exit EXIT
@@ -88,6 +91,9 @@ else
     echo "Dump type must be one of 'full', 'incremental' or 'feedback'"
     exit
 fi
+
+# Lock cron, so it cannot be accidentally terminated.
+admin/cron_lock.py lock-cron "Creating $DUMP_TYPE dump."
 
 DUMP_TEMP_DIR="$DUMP_BASE_DIR/$SUB_DIR.$$"
 echo "DUMP_BASE_DIR is $DUMP_BASE_DIR"
--- a/admin/cron_lock.py
+++ b/admin/cron_lock.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+
+import click
+
+
+CRON_LOGS_DIR = "/logs"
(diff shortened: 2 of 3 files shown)


admin/cron_lock.py
```diff
@@ -0,0 +1,89 @@
...
+CRON_LOGS_DIR = "/logs"
+CRON_LOCK_FILE = os.path.join(CRON_LOGS_DIR, "cron.lock")
+
+
+def sanity_check():
+    """ Check to make sure the logs dir exists and that the CONTAINER_NAME env
+        var is set correctly for the cron container. If the PROD_ENV var is not
+        set, assume that we are not running in production and exist with success. """
+
+    if "PROD_ENV" not in os.environ:
```
alastair, line 18: where is this environment variable set? did you mean to check `DEPLOY_ENV=prod` ?

listenbrainz/listenstore/timescale_utils.py
```diff
@@ -114,6 +115,16 @@ def recalculate_all_user_data():
             pass
 
 
+def unlock_cron():
+    """ Unlock the cron container """
+
+    # Unlock the cron container
+    try:
+        subprocess.run(["admin/cron_lock.py", "unlock-cron"])
```
alastair, line 123: What will happen if we start the dumps at 4am on the 15th and then an hour later (5am according to crontab) we try and run refresh_listen_count_aggregate? From what I can tell here we'll fail to make the lock file (because it already exists) but this code won't fail due to the try/except. When the aggregate process finishes, we'll delete the dump lock.

admin/create-dumps.sh
```diff
@@ -63,6 +63,9 @@ function on_exit {
         local duration=$(( $(date +%s) - START_TIME ))
         echo "create-dumps took ${duration}s to run"
     fi
+
+    # Remove the cron lock
+    admin/cron_lock.py unlock-cron
```
alastair, line 68: In all other places in this file where we call python scripts, we call `python scriptname.py` rather than using +x. In order to maintain consistency I'd prefer to explicitly run python here.

alastair (changes requested):
Some questions about the interaction of multiple lock files, but otherwise fine

admin/cron_lock.py
mayhem: I really fucking hate dyslexia. :facepalm:

admin/create-dumps.sh
mayhem: Good call!

listenbrainz/listenstore/timescale_utils.py
mayhem: Also a good catch!

PR #1515, opened 2021-06-18
Silently drop empty or null mbids in listen submissions

amCap1712:
The existing checks are insufficient as they allow "", null, [] values for various mbid fields to pass through. We have decided to silently drop such empty fields without erroring.

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -186,13 +186,13 @@ def validate_listen(listen, listen_type):
                 if len(tag) > MAX_TAG_SIZE:
                     raise APIBadRequest("JSON document may not contain track_metadata.additional_info.tags "
                                         "longer than %d characters." % MAX_TAG_SIZE, listen)
-        # MBIDs
+        # MBIDs, both of the mbid validation methods mutate the listen payload if needed.
         single_mbid_keys = ['release_mbid', 'recording_mbid', 'release_group_mbid', 'track_mbid']
         for key in single_mbid_keys:
-            verify_mbid_validity(listen, key, multi = False)
+            validate_single_mbid_field(listen, key)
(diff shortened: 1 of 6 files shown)


mayhem (approved):
:shipit:

listenbrainz/webserver/views/api_tools.py
```diff
@@ -186,13 +186,13 @@ def validate_listen(listen, listen_type):
...
-        # MBIDs
+        # MBIDs, both of the mbid validation methods mutate the listen payload if needed.
         single_mbid_keys = ['release_mbid', 'recording_mbid', 'release_group_mbid', 'track_mbid']
         for key in single_mbid_keys:
-            verify_mbid_validity(listen, key, multi = False)
+            validate_single_mbid_field(listen, key)
         multiple_mbid_keys = ['artist_mbids', 'work_mbids']
         for key in multiple_mbid_keys:
-            verify_mbid_validity(listen, key, multi = True)
+            validate_multiple_mbids_field(listen, key)
```
alastair, line 195: because the validate_listen method now mutates a listen in-place, I'd like to make it clear that this is the case. I'm not sure the best way to do this
At the very least, update the documentation to include this step.
Optionally, make this method _return_ the modified dictionary so that we can do
`listens = validate_listens(listens)`
This means that reading the code it will be clear to us that some modification is going on. Otherwise I can see us encountering something in the future where we forget that this is happening and make some assumption based on it.

alastair (commented):
Fine by me, just one comment about documentation/clarity on the behaviour of this method.

listenbrainz/webserver/views/api_tools.py
amCap1712: Agreed. Fixed now.

alastair (approved):
looks better now, thanks!

PR #1517, opened 2021-06-19
Speed up python tests in CI with sharding and tmpfs mounts

amCap1712:
The python tests take about 8-10 mins to execute usually. Using sharding, we spilt the tests into batches and run them parallelly using Github Actions' strategy matrix. One issue with the current approach is that the image is built for each batch. It would nice if we could split the image building and test running into different steps but that requires us to upload the image somewhere.

Another thing I was looking into is mounting a `tmpfs` volume in the database image. I tried that but didn't get noticeable performance gains, further that is only supported on Linux according to my understanding. I do not know if that means it will error on Docker Mac or just ignored. @alastair can you try running tests with this PR locally? If it just gets ignored, we can add it anyways.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -22,6 +22,10 @@ jobs:
 
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        shard: [ 0, 1 ]
+
     steps:
     - uses: actions/checkout@v2
 
@@ -39,4 +43,4 @@ jobs:
       continue-on-error: true
 
     - name: Run tests
-      run: ./test.sh
+      run: ./test.sh --shard-id=${{ matrix.shard }} --num-shards=2
--- a/test.sh
+++ b/test.sh
@@ -210,8 +210,9 @@ if [ "$1" == "spark" ]; then
     echo "Running tests"
     docker-compose -f $SPARK_COMPOSE_FILE_LOC \
                    -p $SPARK_COMPOSE_PROJECT_NAME \
-                   -e PYTHONDONTWRITEBYTECODE=1 \
-                run --rm request_consumer \
+                run \
+                -e PYTHONDONTWRITEBYTECODE=1 \
+                --rm request_consumer \
                 dockerize -wait tcp://namenode:9000 -timeout 60s \
                 python -m pytest -c pytest.spark.ini "$@"
     RET=$?
--- a/test.sh
+++ b/test.sh
@@ -210,9 +210,10 @@ if [ "$1" == "spark" ]; then
     echo "Running tests"
     docker-compose -f $SPARK_COMPOSE_FILE_LOC \
                    -p $SPARK_COMPOSE_PROJECT_NAME \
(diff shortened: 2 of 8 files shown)


amCap1712:
I tried the uploading/downloading stuff I mentioned above. Github provides [upload-artifact](https://github.com/actions/upload-artifact) and [download-artifact](https://github.com/actions/download-artifact) actions. According to my understanding, these artifacts are separate from cache. Further, these artifacts count against the Actions storage quota. The settings shows that each organization has 2GB of that. But during tests on my repo, no usage shows up there. So, presumably public artifacts do not count against storage quota.

That said, the upload times need to considered for doing this. The spark image is ~1G after compression and takes <1 min to compress and ~2min to upload. This is on par with the best case build time. However, spark image builds usually take much more time. Its definitely a good contender for doing the artifact dance.

The LB python image used in unit tests and integration tests is ~250 MB. Assuming similar build speeds, expect a ~30s upload time and a similar compression time.

Anyways, its a bit funny we use the cache action to cache built docker image layers and also add some scripts to do the same steps to export the final image as an artifact.

amCap1712:
#1518 and #1519 together improve the running time of spark tests very well. Even in the event of cache-miss, the build takes 1.5mins and tests run 3x faster. All in all no need to do the artifact dance for spark tests now.

test.sh
```diff
@@ -205,22 +209,28 @@ if [ "$1" == "spark" ]; then
...
 fi
 
 if [ "$1" == "int" ]; then
+    if [ "$2" == "-b" ]; then
+        int_build
+        exit $?
+    fi
     echo "Taking down old containers"
     int_dcdown
-    echo "Building current setup"
```
alastair, line 221: one nice thing about the original code is that you can just run `./test.sh int` and it'll take care of everything - running a build if necessary and then doing tests. By moving it out into `-b` we have to remember to periodically run build if we change dependencies, etc. If you don't want to run a build each time during CI tests, I think my preference here would be to have a specific workflow for CI which is different to the developer interface

.github/workflows/unit-tests.yml
```diff
@@ -39,4 +43,4 @@ jobs:
       continue-on-error: true
 
     - name: Run tests
-      run: ./test.sh
+      run: ./test.sh --shard-id=${{ matrix.shard }} --num-shards=2
```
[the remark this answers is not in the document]
amCap1712: I didn't update unit tests so that we could compare unit and integration test runs side by side,

test.sh
amCap1712: Yeah, I was thinking this too. Currently, the sharding and artifact dance does not seem to improve tests run as much as I would have thought. I wondering if its worth to pursue this PR currently. This will probably be useful when we merge unit/integration tests but as things stand currently the gain in tests speed probably do not justify adding more stuff around this.

alastair:
@amCap1712 given your comments here, what do you think about this PR? Do you want to close it now and come back to it if we find that it's needed when we merge test runs together?

amCap1712:
Yes, happy to close for now.

PR #1518, opened 2021-06-20
Modify spark test configuration to make test execution faster

amCap1712:
Spark's defaults are optimized for large distributed datasets. Our test datasets are small and therefore altering the configuration to make test execution faster. This reduces the test execution time from ~9 mins to ~3 mins.

### DIFF
--- a/listenbrainz_spark/__init__.py
+++ b/listenbrainz_spark/__init__.py
@@ -50,6 +50,14 @@ def init_test_session(app_name):
                 .builder \
                 .master('local') \
                 .appName(app_name) \
+                .config("spark.sql.shuffle.partitions", "1") \
+                .config("spark.default.parallelism", "1") \
+                .config("spark.executor.cores", "1") \
+                .config("spark.executor.instances", "1") \
+                .config("spark.shuffle.compress", "false") \
+                .config("spark.rdd.compress", "false") \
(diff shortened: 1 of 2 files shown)


listenbrainz_spark/__init__.py
```diff
@@ -50,6 +50,14 @@ def init_test_session(app_name):
...
                 .master('local') \
                 .appName(app_name) \
+                .config("spark.sql.shuffle.partitions", "1") \
+                .config("spark.default.parallelism", "1") \
+                .config("spark.executor.cores", "1") \
+                .config("spark.executor.instances", "1") \
+                .config("spark.shuffle.compress", "false") \
+                .config("spark.rdd.compress", "false") \
+                .config("spark.dynamicAllocation.enabled", "false") \
+                .config("spark.io.compression.codec", "lz4") \
```
alastair, line 60: It'd be good to have a comment either here in this file or in the commit message explaining what these do, and optionally a link to a website that explains what these options mean.
You did a good description in the PR description, but putting it in comments or the commit message would mean less hoops to jump through to find a description (we could go from git log -> github commit page -> associated PR, but it'd be faster to just read a comment!)

alastair (approved):
I added a comment here explaining where the config items came from. OK to merge now.

PR #1519, opened 2021-06-20
Split spark docker image into base and dev/test

amCap1712:
The spark dockerfile can be broadly classified into two parts. The first part is downloading Java, Hadoop and Spark. This part is LB agnostic and will change infrequently, maybe a couple of times a year. Hadoop and Spark are downloaded from Apache mirrors and interestingly are often the bottleneck in the build.  I think it might be better if we split this not often changing part into a different dockerfile and push that to DockerHub. The LB specific code or anything that may change more frequently can remain in the dev/test image.


alastair (approved):
Looks good, thanks!

PR #1520, opened 2021-06-21
Adhoc fix for thresholding listens

amCap1712:
There are two issues with the current thresholding code:
1) threshold is applied on individual (recording mbid, user_id) pairs.
This is wrong because the threshold is for total listens of the user.
2) threshold is applied after assigning ranks to recordings and users.
This means if the thresholding removes a user or recording from the
listens, the matrix for generated during later processing will have
rows and columns filled with 0s.

This patch fixes both of these issues by performing thresholding on
total listen count for user just after mapping listens. The current
fix is not the best possible but the quickest one I could come up with.

### DIFF
--- a/listenbrainz_spark/recommendations/dataframe_utils.py
+++ b/listenbrainz_spark/recommendations/dataframe_utils.py
@@ -50,14 +50,13 @@ def get_dates_to_train_data(train_model_window):
     return to_date, from_date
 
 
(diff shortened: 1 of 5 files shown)


amCap1712:
Do not merge yet, I'll deploy and test after approval. If it works, will update documentation.

mayhem (approved):
Lets give it a try!

alastair:
I'm not going to review this because I haven't done much of the recommendation stuff. Ping me explicitly if you want me to look at it.

PR #1523, opened 2021-06-22
LB-865: Move messybrainz lookup to after listen data is added to the queue

amCap1712:
LB-865 describes the rationale to do so.  This means either we add another queue to do messybrainz lookup for now playing
listens or stop doing msb lookup for them. Since, the now playing listens are only transient and not stored in the database, we have believe it is safe to stop doing msb lookups for now playing listens. This creates some differences between the now playing Listens and submit listens, hence creating a separate class.

### DIFF
--- a/listenbrainz/listen.py
+++ b/listenbrainz/listen.py
@@ -123,17 +123,14 @@ class Listen(object):
     def from_json(cls, j):
         """Factory to make Listen() objects from a dict"""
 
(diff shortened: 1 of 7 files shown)


mayhem (approved):
Looks good to me -- glad to see things getting streamlined!

amCap1712:
- [ ] Recheck that frontend code does not use msids of now playing listens.

PR #1525, opened 2021-06-23
Do not block startup if typesense is down

amCap1712:
Typesense was acting up today and needed to be shutdown. This brought down LB along with it. In case we need to shutdown typesense again, disabling startup improvements for it. The SERVICEDOESNOTEXIST sentinel is grepped by the startup script to detect issues and prevent container from starting. Setting it to any other value so that container starts even in absence of typesense.

### DIFF
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -72,8 +72,8 @@ TYPESENSE_HOST = "{{.Address}}"
 TYPESENSE_PORT = {{.Port}}
 {{end}}
 {{else}}
-TYPESENSE_HOST = "SERVICEDOESNOTEXIST_typesense-listenbrainz"
-TYPESENSE_PORT = "SERVICEDOESNOTEXIST_typesense-listenbrainz"
+TYPESENSE_HOST = "localhost"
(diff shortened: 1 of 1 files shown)


amCap1712:
No need to merge this, I have pushed an image based on this branch `metabrainz/listenbrainz:no-typesense` in case we need to deploy this.

amCap1712:
Typesense is fixed now, closing.

amCap1712:
@alastair investigated using docker health checks and stuff to fix this properly but he found that the consul stuff may change soon so its probably not worth doing just presently. We can relook using health checks in future again and fixing this properly at that time.

alastair (approved):
Looks fine. I see that the only thing that uses this value is labs api and the mapping, so it doesn't matter what the value is within the LB container.

As a future work, I wonder if something like this should be moved out of the main config file. How would we do this? Flask can load multiple config files - we could say that labs-api loads the main config and then an additional one that contains labs-specific config items? Not sure if this is a good idea, or if it will cause confusions.

amCap1712:
That sounds like a neat idea and we already do it for cron. But yeah, it may also cause confusion.

PR #1514, opened 2021-06-18
Add Pinned Recording API endpoint

jdaok:
TO DO:
- [x] Add tests

This PR aims to add the Pinned Recording API endpoint which contains routes to 

- pin, 
https://api.listenbrainz.org/1/pin
- unpin, 
https://api.listenbrainz.org/1/pin/unpin
- delete pin, 
https://api.listenbrainz.org/1/pin/delete/(pinned-id)
- get pin history for a user, 
https://api.listenbrainz.org/1/(user-name)/pins
- and get pin for a user's following list.
https://api.listenbrainz.org/1/(user-name)/pins/following

PR #1527, opened 2021-06-26
Add optional recordingMBID to pinned recordings

jdaok:
This PR aims to add recording MBID's to pinned recordings. (builds upon [#1514 ](https://github.com/metabrainz/listenbrainz-server/pull/1514))

As part of ListenBrainz' migration to MBID's, we should store optional MBID's in addition to required MSID's in the `pinned_recordings `table, which allows us to easily switch to MBID's in the future once mapping improves.

Changes specific to this PR:
- update create_tables.sql
- adds admin sql file to recreate the pinned_recording table
- update PinnedRecording model + validator
- update SQLAlchemy queries + tests
- update API + tests


admin/sql/updates/2021-06-25-add-optional-mbid-to-pinned-recording-table.sql
```diff
@@ -0,0 +1,30 @@
+BEGIN;
+
+-- this table is currently empty, we can drop it and recreate it with the new changes
+
+DROP TABLE IF EXISTS pinned_recording;
```
mayhem, line 5: I see your logic here -- but I would really prefer to see this redone using ALTER TABLE rather than DROP TABLE. This is more proof against future disasters -- imagine if someone fat-fingers a script name and wants to run a simple update script, but instead gets this script instead by accident. BAM, all of our pins are gone. If this is ALTER TABLE, then the statement will fail (since the column will already exist in this case) and nothing bad happens.

listenbrainz/db/tests/test_pinned_recording.py
```diff
@@ -19,10 +19,10 @@ def setUp(self):
         self.followed_user_2 = db_user.get_or_create(3, "followed_user_2")
 
         self.pinned_rec_samples = [
-            {"recording_mbid": "7f3d82ee-3817-4367-9eec-f33a312247a1", "blurb_content": "Amazing first recording"},
-            {"recording_mbid": "7f3d82ee-3817-4367-9eec-f33a312247a1", "blurb_content": "Wonderful second recording"},
-            {"recording_mbid": "7f3d82ee-3817-4367-9eec-f33a312247a1", "blurb_content": "Incredible third recording"},
-            {"recording_mbid": "67c4697d-d956-4257-8cc9-198e5cb67479", "blurb_content": "Great fourth recording"},
+            {"recording_msid": "7f3d82ee-3817-4367-9eec-f33a312247a1", "blurb_content": "Amazing first recording"},
```
mayhem, line 22: Shouldn't we also test recording_mbids here as well?

mayhem (changes requested):
Overall I think we should be testing the MBID and MSID support equally, but this is very much on the right track!

admin/sql/updates/2021-06-25-add-optional-mbid-to-pinned-recording-table.sql
jdaok: I didn't think about this, it makes sense to be extra cautious about using DROP TABLE. :o Done.

mayhem (approved):
Ok, looks good!

PR #1528, opened 2021-06-28
Add consul_template KEY_JSON to use when the value of key needs to parsed before use in Python

amCap1712:
Json object, arrays, booleans stored in consul config need to be parsed using `json.loads` before using in Python. Add a nested template to avoid duplicating json.loads multiple times.

### DIFF
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -8,6 +8,10 @@
     {{- end -}}
 {{- end -}}
 
+{{- define "KEY_JSON" -}}
+    json.loads('''{{template "KEY" .}}''')
+{{- end -}}
+
 # Put everything after the defines otherwise the rendered config might eat up some newlines
 import os
 import json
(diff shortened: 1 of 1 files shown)


mayhem (approved):
This looks ok, but consult templates are really not my strong suit.

alastair:
As I mentioned in IRC, we have to decide what the behavior will be if the value that we get from consul isn't a valid json string. In this case, json.loads will raise an exception, uwsgi will fail to load, and then we'll send a startup failure to sentry. Are we happy with this?

One other option would be to make this data a valid python dict/list as needed. However, I suspect that this might be even more complex than this proposed solution. As I understand we get a json config file from google, right? so it should just be a matter of copying it into the consul config file with the appropriate escaping, right?

amCap1712:
I think not loading uwsgi on invalid config is fine. We would want to fix it quickly anyways. Yes, we get a json file from google so we just need to escape it and put that in a string.

alastair (approved):

PR #1529, opened 2021-06-28
Move to per-user scaling

mayhem:
As per discussion on IRC -- I've not run this code. I'm happy to do the deploy and relevant fixes for this PR.


mayhem:
We just tested this and the user similarities look much nicer now. I think we should merge this for now and see what other improvements we want to make later.

amCap1712 (approved):
LGTM

PR #1530, opened 2021-06-28
Terminate cron

mayhem:
Attempt #2 (#3?) -- this version add support for multiple locks in order avoid different tasks from clobbering each other's locks.

### DIFF
--- a/admin/create-dumps.sh
+++ b/admin/create-dumps.sh
@@ -65,7 +65,7 @@ function on_exit {
     fi
 
     # Remove the cron lock
-    admin/cron_lock.py unlock-cron create-dumps
+    python3 admin/cron_lock.py unlock-cron create-dumps
 }
 
 START_TIME=$(date +%s)
@@ -91,12 +91,10 @@ else
 fi
 
 # Lock cron, so it cannot be accidentally terminated.
-admin/cron_lock.py lock-cron create-dumps "Creating $DUMP_TYPE dump."
+python3 admin/cron_lock.py lock-cron create-dumps "Creating $DUMP_TYPE dump."
 
 # Trap should not be called before we lock cron to avoid wiping out an existing lock file
 trap on_exit EXIT
-
-sleep 9999
 
 DUMP_TEMP_DIR="$DUMP_BASE_DIR/$SUB_DIR.$$"
 echo "DUMP_BASE_DIR is $DUMP_BASE_DIR"
--- a/admin/create-dumps.sh
+++ b/admin/create-dumps.sh
@@ -65,10 +65,8 @@ function on_exit {
     fi
 
     # Remove the cron lock
-    admin/cron_lock.py unlock-cron
+    admin/cron_lock.py unlock-cron create-dumps
 }
-
-trap on_exit EXIT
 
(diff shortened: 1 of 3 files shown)


admin/create-dumps.sh
```diff
@@ -89,6 +90,14 @@ else
     exit
 fi
 
+# Lock cron, so it cannot be accidentally terminated.
+admin/cron_lock.py lock-cron create-dumps "Creating $DUMP_TYPE dump."
+
+# Trap should not be called before we lock cron to avoid wiping out an existing lock file
+trap on_exit EXIT
+
+sleep 9999
```
alastair, line 99: is this from testing?

admin/create-dumps.sh
```diff
@@ -89,6 +90,14 @@ else
     exit
 fi
 
+# Lock cron, so it cannot be accidentally terminated.
+admin/cron_lock.py lock-cron create-dumps "Creating $DUMP_TYPE dump."
```
alastair, line 94: In the previous PR we decided to prefix these with `python [file.py]`, I see it's still there in the python code, so I guess this is an oversight

admin/create-dumps.sh
```diff
@@ -89,6 +90,14 @@ else
...
+
+# Trap should not be called before we lock cron to avoid wiping out an existing lock file
+trap on_exit EXIT
```
alastair, line 97: To confirm what you mean by this comment - if `lock-cron create-dumps` fails (because the lock file already exists, and you want to quit the script), we shouldn't run `unlock-cron`, because that'd delete the lock file for the already running job, not this one that we just did?

admin/create-dumps.sh
```diff
@@ -89,6 +90,14 @@ else
     exit
 fi
 
+# Lock cron, so it cannot be accidentally terminated.
```
alastair, line 93: We still have an additional lock file [listed in the crontab](https://github.com/metabrainz/listenbrainz-server/blob/master/docker/services/cron/crontab#L6). Do we want to keep both of them, or do you think that we can remove the crontab one and just keep this one?

alastair (changes requested):
Looks like one small item remaining from testing.

I'll give this a test myself now, but I think it looks fine.

admin/create-dumps.sh
mayhem: Sigh, yes. Testing this is... a bit contrived. :(

admin/create-dumps.sh
mayhem: Keep both -- they serve very different functions. The crontab ones are actual exclusions that prevent multiple things from running at once and this PR is about a cooperative lock to prevent cron container termination at the wrong time.

admin/create-dumps.sh
mayhem: Ah, yes, I only fixed the ones from the python context, not the shell context.

admin/create-dumps.sh
mayhem: Correct.

alastair (approved):
looks good now, thanks!

PR #1531, opened 2021-06-28
Add SQL script to logout all users

amCap1712:
Need to logout all users so that all users see a warning when they login again.


mayhem (approved):

alastair (approved):
I tested on a local database with 2 users. each user got a new different login_id.

alastair:
let's add a comment to this file explaining what it does and when to use it - can link to the flask-login doc page.

PR #1536, opened 2021-06-29
Add global scaling to user similarities

mayhem:
As discussed in IRC.


amCap1712 (approved):
I think there are no tests for global_similarity. Would be nice to add 1 or 2 tests to cover it. Other than that, LGTM.

mayhem:
Done!

PR #1537, opened 2021-06-29
Fix missing email bug in sign up process preventing new users from signing up

amCap1712:
The musicbrainz_post method in login view checks for the user's email to determine whether a warning should be shown. Add email in new user's dict so that the check can happen but they are not shown the warning. Currently, we get a NoSuchColumnError preventing new users from signing up.


alastair (approved):
👍

PR #1535, opened 2021-06-29
Add blurb content limit to pin recordings

jdaok:
This PR aims to add a validator to the `blurb_content` attribute in the `PinnedRecording` model. 

(builds upon #1527, Diff link [here](https://github.com/jdaok/listenbrainz-server/compare/add-MBID-to-pinned-recording...jdaok:add-blurb-content-limit-to-pin-recordings))

PR #1538, opened 2021-07-01
Add Pinned Recording Modal

jdaok:
This PR aims to add the PinRecordingModal component to the frontend to allow users to pin recordings.

Changes:
- adds PinRecordingModal component + tests + snapshot
- adds APIService function + tests
- adds modal to RecentListens page + tests + snapshot
- adds new ListenControl to ListenCard + tests + snapshot
- adds new optional attributes to ListenControl
- updates preciseTimestamp in utils to take optional display setting

image


listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
```diff
@@ -0,0 +1,174 @@
...
+  PinRecordingModalProps,
+  PinRecordingModalState
+> {
+  static contextType = GlobalAppContext;
+  declare context: React.ContextType<typeof GlobalAppContext>;
+  readonly maxBlurbContentLength = 280;
+
+  constructor(props: PinRecordingModalProps) {
+    super(props);
+    this.state = { blurbContent: undefined };
```
paramsingh, line 29: this seems weird? Could we initialize it as an empty string `''` and then remove the optionality from the `PinRecordingModalState` type as well? It removes us having to work with undefined, less code is always good.

what do you think?

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
```diff
@@ -0,0 +1,174 @@
...
+              >
+                <span aria-hidden="true">&times;</span>
+              </button>
+              <h4 className="modal-title" id="PinRecordingModalLabel">
+                Pin This Recording to Your Profile
+              </h4>
+            </div>
+            <div className="modal-body">
+              <p>
+                Why are you pinning{" "}
```
paramsingh, line 120: (nit)

This sentence reads a bit weird to me, it can also be taken in a negative sense. Could we make the sentence more obviously positive?

"Why do you love this song?" or something similar? I also notice that the placeholder is along these lines, so perhaps we're okay with this.

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
```diff
@@ -0,0 +1,174 @@
...
+              <small style={{ display: "block", textAlign: "right" }}>
+                {blurbContent?.length} / {this.maxBlurbContentLength}
+              </small>
+              <small>
+                Pinning this recording will replace any recording currently
+                pinned. <br />
+                <b>
+                  {track_name} by {artist_name}
+                </b>{" "}
+                will be unpinned from your profile in one week.
```
paramsingh, line 149: (optional, not blocking the PR)

Do we know the actual time for this? Maybe we could show it on hover or something, it might be a good user experience.

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
```diff
@@ -0,0 +1,174 @@
...
+              </small>
+            </div>
+            <div className="modal-footer">
+              <button
```
paramsingh, line 153: This button doesn't have an onClick, is that intentional? Does bootstrap handle the closing of the modal instead?

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
```diff
@@ -0,0 +1,174 @@
...
+        }
+      } catch (error) {
+        this.handleError(error, "Error while pinning recording");
+      }
+    }
+  };
+
+  handleBlurbInputChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+    event.preventDefault();
+    const input = event.target.value.replace(/\s\s+/g, " "); // remove line breaks and excessive spaces
```
paramsingh, line 70: Is there a reason we don't allow line breaks?

mayhem:
#1535 is merged, please merge master into this branch!

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -813,3 +830,33 @@ describe("Pagination", () => {
...
+      <GlobalAppContext.Provider value={mountOptions.context}>
+        <RecentListens {...props} />
+      </GlobalAppContext.Provider>
+    );
+    const instance = wrapper.instance();
+    const recordingToPin = props.listens[0];
+    let pinRecordingModal = wrapper.find(PinRecordingModal).first();
+
+    // recordingToPin is initially undefined
+    expect(pinRecordingModal.props()).toEqual({
```
paramsingh, line 846: For my understanding, why are we supporting an undefined `recordingToPin` in the modal? Ideally if we do not have a `recordingToPin` we wouldn't render the modal at all. This would help because then we don't need to care about weird edgecases where recordingToPin is undefined and we're rendering the modal without track names or something. Ideally, we should make the types as strict as possible.

paramsingh (commented):
Code looks good to me! I left a few questions about the possibility of avoiding null types, but don't really have other comments.

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
jdaok: Done. I agree, string `''` and undefined string are the same but this lets us remove the optionality.

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
jdaok: Sure, I appended the timestamp to the text, that way it's viewed more easily as opposed to having to hover.

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
jdaok: Yes, when clicked, bootstrap uses the data-dismiss attribute in the button to hide elements with the class "modal" on the page iirc.

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
jdaok: It just makes the cards look uniform when they eventually get rendered back to the user, otherwise we could get a blurb content with nothing but line breaks. Restricting it in the input rather is better imo because the user can see how it will look when it gets rendered.

Twitter allows one line break at a time in tweets, maybe we could do that if this is too restrictive?

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
jdaok: The `PinnedRecordingModal ` would have to be rendered before it gets shown: if we only render `PinnedRecordingModal `when `recordingToPin `is set the dataTarget in `ListenControl` isn't able to locate the element with `id=#PinRecordingModal` to show the first time it gets clicked since setState doesn't update `recordingToPin` in the state immediately. 

I changed the component so recordingToPin is initialized to a random listen so we can avoid the weird undefined edgecases like you said.

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
jdaok: Changed the message to make the purpose of pinning recordings a bit clearer.

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -231,44 +231,54 @@ const formatWSMessageToListen = (wsMsg: any): Listen | null => {
 };
 
 // recieves or unix epoch timestamp int or ISO datetime string
-const preciseTimestamp = (listened_at: number | string): string => {
+const preciseTimestamp = (
+  listened_at: number | string,
+  displaySetting?: "timeAgo" | "includeYear" | "excludeYear"
```
jdaok, line 236: Updated this function so we can explicitly choose a setting to display timestamps (was needed to show timestamp in modal).

paramsingh (approved):
Looks good to me, thanks for making those changes!

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
paramsingh: No I think it's okay for now, let's iterate on it later if we need to.

PR #1539, opened 2021-07-01
BrainzPlayer datasources improvements

MonkeyDo:
This PR changes a few things with how we play music with BrainzPlayer:

- ~~Instead of trying to play a track with each datasource until one works, we check if any of the datasources will be able to play the track, and if not we skip it.~~
That was a dumb idea and broke our fallback strategy of trying to play on any of the datasources it could.
- The previous functionality of determining the source of a listen and finding a usable datasource is improved and partly moved from the BrainzPlayer component to two new methods in the DataSource interface: `isListenFromThisService` and `canSearchAndPlayTracks`

- we also reduce the verbosity of warnings, removing the "We couldn't find a matching song on any music service we tried" messages
- while adding a new toast notification with the song title when we change track
- we ensure that nothing happens with the player until a direct user interaction to play a song:
  - don't take any action until a flag is set by a user action, including if there is any error with the datasources (previously an error while setting up datasources could trigger playing the next track in the list even if the user didn't ask to play in the first place)
  - ~~don't render anything if a datasource hasn't been selected to play (rendering iframes before a track is played sometimes creates problems)~~ This broke the Youtube player whose iframe needs to be initialized first even if it is not yet in use
- finally, we add a helpful message when no service is registered to search for and play songs.
Suggestions welcome on the wording !
This may or may not be relevant depending on how the whole Youtube auth things goes.
image


listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -244,26 +256,29 @@ export default class BrainzPlayer extends React.Component<
...
     /** If no matching datasource was found, revert to the default bahaviour
-     * (play from source 0 or if called from failedToFindTrack, try next source)
+     * (try playing from source 0 or try next source)
      */
     if (selectedDatasourceIndex === -1) {
       selectedDatasourceIndex = datasourceIndex;
     }
 
-    const { onCurrentListenChange } = this.props;
-    onCurrentListenChange(listen);
```
MonkeyDo, line 266: This moved up to the beginning of the function

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -280,44 +295,6 @@ export default class BrainzPlayer extends React.Component<
     });
   };
 
-  getSourceIndexByListenData = (listen: Listen | JSPFTrack): number => {
```
MonkeyDo, line 283: This was moved to the datasources themselves that now implement an `isListenFromThisService` method

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -369,22 +350,18 @@ export default class BrainzPlayer extends React.Component<
...
     }
-    const { currentDataSourceIndex } = this.state;
+    const { currentListen } = this.props;
 
-    if (currentDataSourceIndex < this.dataSources.length - 1) {
+    if (currentListen && currentDataSourceIndex < this.dataSources.length - 1) {
       // Try playing the listen with the next dataSource
       this.playListen(currentListen, currentDataSourceIndex + 1);
     } else {
-      this.handleWarning(
```
MonkeyDo, line 384: This was just too annoying.

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -409,8 +386,22 @@ export default class BrainzPlayer extends React.Component<
...
   trackInfoChange = (title: string, artist?: string): void => {
     this.setState({ currentTrackName: title, currentTrackArtist: artist });
+    const message = (
+      <>
+        <FontAwesomeIcon icon={faPlayCircle as IconProp} />
+        &emsp;{title}
+        {artist && ` — ${artist}`}
+      </>
+    );
+    this.handleInfoMessage(message);
```
MonkeyDo, line 396: Show a notification with the song title on a blue background:
image

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -409,8 +386,22 @@ export default class BrainzPlayer extends React.Component<
...
   };
 
+  // eslint-disable-next-line react/sort-comp
+  throttledTrackInfoChange = _throttle(this.trackInfoChange, 2000, {
```
MonkeyDo, line 400: The Youtube player was calling this method twice in short succession when loading a song.

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -448,14 +439,22 @@ export default class BrainzPlayer extends React.Component<
...
       progressMs,
       durationMs,
+      isActivated,
     } = this.state;
     const { APIService, youtubeAuth, spotifyAuth } = this.context;
+    // Determine if the user is authenticated to search & play tracks with any of the datasources
+    const hasDatasourceToSearch =
+      this.dataSources.findIndex((ds) =>
+        ds.current?.canSearchAndPlayTracks()
+      ) !== -1;
```
MonkeyDo, line 449: This (and the associated message line 468) might or might not be useful depending on wether we have a default player (like youtube using API keys) or if they all require auth. It does appear for a flash we the page loads, but suppose it doesn't hurt much to have clear indications in any case.

PR #1542, opened 2021-07-08
Query Youtube API using api key only

amCap1712:
Found that this works by hit and try and some chance. All documentation points to need of Oauth to access the Youtube but somehow this works 😄 . We tested this on beta. This also renders most of the Youtube Oauth stuff we wrote useless 😭 , will remove that in a subsequent PR.

### DIFF
--- a/listenbrainz/webserver/templates/user/music_services.html
+++ b/listenbrainz/webserver/templates/user/music_services.html
@@ -72,32 +72,9 @@
             <div class="panel-body">
                 <p>
                     ListenBrainz integrates with Youtube to let you play music tracks from ListenBrainz pages.
-                    ListenBrainz has a number of music discovery features that require authentication to search
-                    and play tracks from Youtube.
+                    You do not need to do anything to enable this. ListenBrainz will automatically search for
+                    tracks on Youtube and play one if it finds a match.
                 </p>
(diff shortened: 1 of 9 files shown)


listenbrainz/webserver/templates/user/music_services.html
```diff
@@ -72,32 +72,9 @@ <h3 class="panel-title">Youtube</h3>
             <div class="panel-body">
                 <p>
```
mayhem, line 73: I wonder if we should allow people to disable playing from YT as a preference.

mayhem (approved):
:crossed_fingers: :crossed_fingers: :crossed_fingers: :crossed_fingers: :crossed_fingers: :crossed_fingers: :crossed_fingers: :crossed_fingers: :crossed_fingers: :crossed_fingers:

listenbrainz/webserver/templates/user/music_services.html
amCap1712: @MonkeyDo had some plans about implementing preferences in BrainzPlayer. Maybe revisit when we do that?

mayhem: Yeah, perfect.

PR #1544, opened 2021-07-08
LB-662: Fix public dump import

alastair:
LB-662

# Problem
Performing an import of just public dump + listen data would fail, due to `PUBLIC_TABLES['"user"']` contain constant values that were dumped into the file. When we used this list to import the dump file, it failed because it should only contain column names and not constant values. 

# Solution
Make a copy of the `PUBLIC_TABLES` dictionary which is used for importing, and set the `"user"` key to a list of column names.

The resulting public dump files are slightly different (they contain the string `\N` to represent null, instead of an empty string `''`), but we decided that we wouldn't increase the schema value, because the public dumps never worked anyway.

Update the import method to allow the private dump file to be optional.

Change the dump method to send an email after all dump steps have finished (writing DUMP_ID.txt) so that in the case that sending an email fails, we still have all of the outputs of the dump created.

### DIFF
--- a/listenbrainz/db/dump.py
+++ b/listenbrainz/db/dump.py
@@ -54,7 +54,7 @@ INCREMENTAL_MAX_AGE = 26  # hours
 
 # this dict contains the tables dumped in public dump as keys
 # and a tuple of columns that should be dumped as values
-PUBLIC_TABLES = {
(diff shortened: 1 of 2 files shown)


amCap1712 (approved):
Tried on bono. Successfully imports the dump.

PR #1545, opened 2021-07-08
Parquet spark dumps

mayhem:
Remove the old transmogrification spark dumps and build new dumps that build parquet files that are easier for spark to import. This should also be faster than transmogrifying the dumps.


listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -666,6 +677,155 @@ def dump_listens(self, location, dump_id, start_time=datetime.utcfromtimestamp(0
...
+                      AND recording_mbid IS NOT NULL
+                 ORDER BY created ASC"""
+
+        args = (start_time, end_time)
+
+        listen_count = 0
+
+        current_created = None
+        conn = timescale.engine.raw_connection()
+        with conn.cursor(cursor_factory=psycopg2.extras.RealDictCursor) as curs:
```
amCap1712, line 728: Now that we are using Pandas, we might as well use its `read_sql`(https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html#pandas-read-sql) utility. The only issue I see with this is how to estimate sizes if we do this.

mayhem: Speaks for keeping it as is, doesn't it. :)

mayhem:
Oh, also I should point out that I haven't actually done a full dump with this. We should do that manually after we merge.

listenbrainz/listenstore/timescale_listenstore.py
amCap1712: Yeah but I think if we read directly into dataframe and then manipulate it instead of doing in it a loop ourself it'll probably be faster.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -666,6 +677,155 @@ def dump_listens(self, location, dump_id, start_time=datetime.utcfromtimestamp(0
...
+                    'created': []
+                }
+                while True:
+                    result = curs.fetchone()
+                    if not result:
+                        break
+
+                    for col in data:
+                        if col == 'artist_credit_id':
+                            data[col].append(int(result[col]))
```
amCap1712, line 752: Can we cast the the column in the sql query instead?

mayhem: The column is an int -- I am not even certain why this is necessary; by default it goes to double as you saw.

listenbrainz/listenstore/timescale_listenstore.py
mayhem: Let continue as is for now and keep in mind that we can use this to speed up the dumps at some point in the future. Like I said, we haven't done a full dump yet, so no idea how long this takes. Faster than before, I'm nearly certain....

listenbrainz/listenstore/timescale_listenstore.py
amCap1712: Right, that's why its weird.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -666,6 +681,227 @@ def dump_listens(self, location, dump_id, start_time=datetime.utcfromtimestamp(0
...
+                        data['recording_mbid'][i] = None
+                        data['artist_credit_id'][i] = None
+                    else:
+                        data['artist_credit_mbids'].append(ac_mapping[row])
+
+                filename = os.path.join(temp_dir, "%d.parquet" % parquet_file_id)
+
+                # Create a pandas dataframe, then write that to a parquet files
+                df = pd.DataFrame(data, dtype=object)
+                table = pa.Table.from_pandas(df)
```
amCap1712, line 829: I encountered a strange issue while generating some parquet files for tests. Due to some reason, pandas was adding an extra index column causing a schema mismatch in Spark. The issue got fixed when I added, `preserve_index=False`. (https://github.com/apache/arrow/issues/2244)

I haven't faced this issue when working with the parquet files generated by the dumps code but should we add this kwarg here as well to be on the safer side?

mayhem: Done.

mayhem:
Master merged in to resolve merge conflict.

PR #1546, opened 2021-07-09
Deduplicate listens submitted by spotify reader

amCap1712:
Spotify provides the "played_at" time in resolution of milliseconds but during parsing we round it off to the nearest lower second because in LB we use  timestamps having resolution of seconds. However when querying spotify we again  have to convert seconds to milliseconds, this actually makes us fetch the last listen again.

Following example demonstrates how this happens:
 
We receive a listen a played_at of `2021-07-09T17:09:00.381Z` from spotify.  This is equal to a milliseconds timestamp of `1625850540381,` we round this off to nearest lower seconds timestamp. So we put latest_listened_at as `1625850540` in our database. Now when we query spotify again, we need to convert this back into milliseconds which becomes `1625850540000`. This is lower than the actual timestamp we received hence spotify sends us the last listen again.

To solve this issue, we store the `latest_listened_at` timestamp in listens_importer table with precision of milliseconds.

### DIFF
--- a/listenbrainz/db/listens_importer.py
+++ b/listenbrainz/db/listens_importer.py
@@ -1,3 +1,5 @@
+from typing import Union
+
 from data.model.external_service import ExternalServiceType
 from listenbrainz import db, utils
 import sqlalchemy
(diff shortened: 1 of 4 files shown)


mayhem (approved):
OK, I think this is a good solution -- lets hope the orange and blue lines move a little closer together!

PR #1547, opened 2021-07-11
Add helper functions for retrieving PinnedRecording track_metadata 

jdaok:
This PR aims to:
- rename `load_recording()` to `load_recording_from_msid()`
- add `load_recording_from_mbid()` to helper function messybrainz
- add `track_metadata` attribute to PinnedRecording
- add `fetch_track_metadata_for_pin()` helper function for PinnedRecordings

### DIFF
--- a/messybrainz/__init__.py
+++ b/messybrainz/__init__.py
@@ -28,9 +28,11 @@ def submit_listens_and_sing_me_a_sweet_song(recordings):
     else:
         raise exceptions.ErrorAddingException("Failed to add data")
 
+
 def load_recording_from_msid(msid):
     with db.engine.begin() as connection:
         return data.load_recording_from_msid(connection, msid)
+
 
 def load_recording_from_mbid(mbid):
     with db.engine.begin() as connection:
--- a/messybrainz/__init__.py
+++ b/messybrainz/__init__.py
@@ -6,7 +6,7 @@ from messybrainz import db
 
 def submit_listens_and_sing_me_a_sweet_song(recordings):
     """ Inserts a list of recordings into MessyBrainz.
-    
+
     Args:
         recordings (list): a list of recordings to be inserted
     Returns:
@@ -51,7 +51,7 @@ def load_recording_from_msid(msid):
 
 def load_recording_from_mbid(mbid):
     """ Returns data for a recording with specified MusicBrainz ID.
-    
+
     Args:
         mbid (uuid): the MusicBrainz ID of the recording
     Returns:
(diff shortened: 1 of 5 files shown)


messybrainz/__init__.py
```diff
@@ -28,16 +28,21 @@ def submit_listens_and_sing_me_a_sweet_song(recordings):
     else:
         raise exceptions.ErrorAddingException("Failed to add data")
 
-def load_recording(mbid):
+
+def load_recording_from_msid(msid):
```
mayhem, line 32: Please add docstrings, even if the function is trivial.

messybrainz/db/data.py
```diff
@@ -219,6 +220,43 @@ def load_recording(connection, messybrainz_id):
...
+    """
+    query = text("""SELECT rj.data
+                         , d.recording_mbid
+                         , r.artist
+                         , r.release
+                         , r.gid
+                      FROM recording_json rj
+                 LEFT JOIN recording r
+                        ON rj.id = r.data
+                 LEFT JOIN recording_cluster rc
```
mayhem, line 241: You shouldn't need to join on recording_cluster -- those tables are leftover from a failed GSoC project we haven't cleaned up. Did you download this messybrainz data? http://data.metabrainz.org/pub/musicbrainz/messybrainz/ If you don't want to do that, (I understand), I could run some queries for you to make sure they work right.

mayhem (changes requested):
The SQL queries will need some updating -- see comments.

messybrainz/__init__.py
jdaok: Adding docstrings for the functions in this file.

messybrainz/db/data.py
jdaok: Sure I removed the joins to `recording_cluster` and `recording_redirect` from the queries because neither the MBID/MSID functions returned anything that used data from the joins. (I didn't have 12gb for the dump, but please test for me :))

messybrainz/db/data.py
```diff
@@ -212,13 +208,46 @@ def load_recording(connection, messybrainz_id):
...
 
+def load_recording_from_mbid(connection, musicbrainz_id):
+    """ Return data for a recording with specified MusicBrainz ID.
+
+    Args:
+        connection: sqlalchemy connection to execute db queries with
+        messybrainz_id (uuid): the MusicBrainz ID of the recording
+
+    Returns:
+        dict: the recording data for the recording with specified MessyBrainz ID
```
mayhem, line 227: MusicBrainz ID

messybrainz/db/data.py
```diff
@@ -212,13 +208,46 @@ def load_recording(connection, messybrainz_id):
...
+    """
+    query = text("""SELECT rj.data
```
mayhem, line 229: This query works fine as well.

mayhem (approved):
Looks good to me -- just one tiny comment change before merge.

PR #1549, opened 2021-07-11
Add PinnedRecordingCard component

jdaok:
This PR aims to implement a PinnedRecordingCard component and renders it on the Recent Listens page.
chrome-capture


paramsingh:
This PR is huge! 😅 

Could we possibly split this into smaller parts? A reasonable thing to do might be to do the backend in one PR and the frontend in another?

Smaller PRs are easier to review, and they're also safer because it's easier to pin down any bugs that might be missed with larger PRs. 

ptal @jdaok

jdaok:
@paramsingh Sure I split it into 3 PRs. I suggest reviewing in this order (#1550, #1551, finally this one #1549.) However now that they are split the tests started failing, I will merge from master as soon as the changes are accepted so that the tests can pass. If there is a way that makes it easier to review, please let me know!

listenbrainz/webserver/static/js/src/PinRecordingModal.tsx
```diff
@@ -59,6 +59,7 @@ export default class PinRecordingModal extends React.Component<
             `${recordingToPin.track_metadata.artist_name} - ${recordingToPin.track_metadata.track_name}`
           );
           this.setState({ blurbContent: "" });
+          window.location.reload();
```
paramsingh, line 62: let's add a comment explaining why we reload here, it's not immediately obvious to someone who doesn't have much context on the code

listenbrainz/webserver/static/js/src/PinnedRecordingCard.test.tsx
```diff
@@ -0,0 +1,327 @@
...
+  });
+});
+
+describe("determineIfCurrentlyPinned", () => {
+  it("returns true when pinned_until > now", () => {
+    const wrapper = mount<PinnedRecordingCard>(
+      <PinnedRecordingCard {...props} />
+    );
+    const instance = wrapper.instance();
+    expect(instance.determineIfCurrentlyPinned()).toBeTruthy();
```
paramsingh, line 68: nit
```suggestion
    expect(instance.determineIfCurrentlyPinned()).toBe(true);
```

we know that this function always returns a bool , so tightening up the assertion would be a good idea.

listenbrainz/webserver/static/js/src/PinnedRecordingCard.test.tsx
```diff
@@ -0,0 +1,327 @@
...
+  });
+
+  it("returns false when pinned_until < now", () => {
+    const wrapper = mount<PinnedRecordingCard>(
+      <PinnedRecordingCard
+        {...{ ...props, PinnedRecording: expiredPinnedRecording }}
+      />
+    );
+    const instance = wrapper.instance();
+    expect(instance.determineIfCurrentlyPinned()).toBeFalsy();
```
paramsingh, line 78: same nit

```suggestion
    expect(instance.determineIfCurrentlyPinned()).toBe(false);
```

listenbrainz/webserver/static/js/src/PinnedRecordingCard.tsx
```diff
@@ -0,0 +1,240 @@
...
+    }
+  };
+
+  render() {
+    const { PinnedRecording, isCurrentUser, userName } = this.props;
+    const { currentlyPinned, isDeleted } = this.state;
+    const { track_name } = PinnedRecording.track_metadata;
+    const { artist_name } = PinnedRecording.track_metadata;
+
+    return (
```
paramsingh, line 123: this component is huge, hundreds of lines of jsx in one component is not ideal. It's not immediately obvious to me if this could be broken into smaller sub-components, but if it could, we should definitely do it. In general small components are easier to reason about and test.

Would you mind taking a look and seeing if this can be broken into smaller subcomponents?

listenbrainz/webserver/static/js/src/PinnedRecordingCard.tsx
```diff
@@ -0,0 +1,240 @@
...
+      <Card
+        className={`pinned-recording-card row${
+          currentlyPinned ? " currently-pinned" : ""
+        }
+        ${isDeleted ? " deleted" : ""}`}
+      >
+        <div className={`${isCurrentUser ? " col-xs-9" : " col-xs-12"}`}>
+          <MediaQuery minWidth={768}>
+            <div className="col-xs-8">
```
paramsingh, line 132: maybe we could create separate components for browser and phone? That would make this file quite a lot smaller and easier to read.

listenbrainz/webserver/static/js/src/PinnedRecordingCard.tsx
```diff
@@ -0,0 +1,240 @@
...
+              <div className="pin-details">
+                <span
+                  className={`${
+                    currentlyPinned ? " pin-title text-muted" : "hidden"
+                  }`}
+                >
+                  <FontAwesomeIcon icon={faThumbtack as IconProp} />{" "}
+                  {isCurrentUser ? "My " : `${userName}'s `} Pinned Recording
```
paramsingh, line 140: This code is duplicated later as well, is this in other places as well? Could we make it a separate component, or at least a function so that we don't need to duplicate code?

paramsingh (commented):
Code looks great, thanks for separating out the PRs. I've left a few comments, ptal @jdaok

listenbrainz/webserver/static/js/src/PinnedRecordingCard.tsx
jdaok: The browser and phone components are very similar and only have a few differing lines in render() to shift some html around. I broke up the similarities into JSX-element-returning-functions to make the component more readable. 

Please let me know if this looks better (not sure if you wanted the two split up into different files), if it still looks too big, some other large components are written similarly and I could update those as well.

listenbrainz/webserver/static/js/src/PinnedRecordingCard.tsx
jdaok: I separated it into a function `renderPinTitle` and also did the same for `renderPinDate` and `renderBlurbContent`.

jdaok:
Merged approved changes #1550 and #1551 so tests pass :)

paramsingh (approved):

PR #1552, opened 2021-07-13
Migrate lastfm import timestamps from users table to listens_importer table

amCap1712:
This aligns storing Last.FM timestamps with Spotify and makes adding future services easier (for eg: libre.fm).


mayhem (commented):
SQL bits look fine.

mayhem (approved):
:ship:

PR #1553, opened 2021-07-14
Add Libre.fm import support

amCap1712:
Add support for importing listens from Libre.fm. Most of the code from Last.fm importer is reused. The importer defaults to importing from Last.fm.

Pending
- [ ] Add UI for resetting Libre.fm timestamps
- [ ] Fix text everywhere to make it neutral to last.fm/libre.fm


mayhem (commented):
SQL bits look fine here too. Proceed to test!

MonkeyDo:
See how you like it now @amCap1712 !
image

admin/sql/updates/2021-07-27-add-critiquebrainz-as-external-service.sql
```diff
@@ -1,3 +1 @@
-BEGIN;
 ALTER TYPE external_service_oauth_type ADD VALUE 'critiquebrainz';
-COMMIT;
```
amCap1712, line 3: Not related to this PR but @alastair reported this script was failing when he tried locally hence fixing it as well here. (Can't run alter type ... add value inside transaction).

amCap1712:
@MonkeyDo mentioned that the pending UI tasks as in description will be done in a different PR.

mayhem (approved):
The python and SQL bits are fine, not sure about the bulk of the react stuff. I'm happy with you merging this since it isn't a critical feature.

PR #1555, opened 2021-07-18
Add feature flag for Pin UI 

jdaok:
This PR adds a feature flag for the pin feature that will be removed once the feature is complete.

- Adds PinFeatureEnabled() to utils.tsx
- Temporarily removes some tests
- updates 2 snapshots


paramsingh (approved):

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -832,32 +834,34 @@ describe("Pagination", () => {
...
-    expect(pinRecordingModal.props()).toEqual({
-      isCurrentUser: true,
-      recordingToPin,
-      newAlert: props.newAlert,
-    });
-  });
-});
+// Will re-add this test when feature flag is removed
+
+// describe("pinRecordingModal", () => {
```
amCap1712, line 839: Is it possible to use `skip` here instead?

jdaok: Changed to skip

amCap1712 (approved):

listenbrainz/webserver/static/js/src/RecentListens.test.tsx
amCap1712: Thanks!

PR #1557, opened 2021-07-20
LB-571: Add trailing slash to all user endpoints

alastair:
# Problem
LB-571

URLs such as https://listenbrainz.org/user/alastairp/ currently result in a 404 page.

# Solution
Add a trailing / to `@route` definitions in our views. This configures flask so that a redirect is added for the same route without a / which then redirects to this specific route (with HTTP 308), and then the redirected URL serves the view.

This means that https://listenbrainz.org/user/alastairp -> 308 -> https://listenbrainz.org/user/alastairp/ , and https://listenbrainz.org/user/alastairp/  will load normally.

I made a specific decision not to add this to any API endpoints, because:
* I don't know how many API clients support 307/308 redirects instead of 301/302 (a POST on 301/2 will retry as GET, which is something that we don't want, but maybe a 308 is ignored as unknown, which is just as bad)
* We explicitly document the API endpoints in readthedocs, and they don't have any trailing slashes, so I don't think that there are any clients which are erroneously adding a trailing /

I had a look into how we could support both trailing / and no / in the API endpoints with no redirect, and the only way I could find was to define each `@route` twice. To me this feels like additional work that isn't really necessary, so I'm recommending that we leave it for now.

### DIFF
--- a/listenbrainz/webserver/test/test_routes.py
+++ b/listenbrainz/webserver/test/test_routes.py
@@ -1,0 +1,28 @@
+import flask
+
+from listenbrainz.db.testing import DatabaseTestCase
+from listenbrainz.webserver import API_PREFIX
+from listenbrainz.webserver.testing import ServerTestCase
+
+
+class RoutesTestCase(DatabaseTestCase, ServerTestCase):
+
+    def setUp(self):
+        ServerTestCase.setUp(self)
+        DatabaseTestCase.setUp(self)
+
+    def test_routes_have_trailing_slash(self):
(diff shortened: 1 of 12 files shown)


alastair:
Urgh, not sure about the failing test when debug toolbar is active... We could add it to the exclude list, or disable debug toolbar in tests...

amCap1712 (approved):

amCap1712:
Disabling debug toolbar in tests sounds fine.

PR #1558, opened 2021-07-20
Add support for getting countries form aritst credit ids

mayhem:
Mostly copied from the MBID version.

### DIFF
--- a/listenbrainz/labs_api/labs/api/artist_country_from_artist_credit_id.py
+++ b/listenbrainz/labs_api/labs/api/artist_country_from_artist_credit_id.py
@@ -1,0 +1,107 @@
+from operator import itemgetter
+
+from datasethoster import Query
+import psycopg2
+import psycopg2.extras
+from werkzeug.exceptions import NotFound
+from listenbrainz import config
+
+
+class ArtistCountryFromArtistCreditIdQuery(Query):
+
+    def names(self):
+        return ("artist-country-code-from-artist-credit-id", "MusicBrainz Artist Country From Artist Credit Id")
+
+    def inputs(self):
+        return ['[artist_credit_id]']
+
+    def introduction(self):
+        return """Given artist credit id look up countries for those artists. Any artist_mbids
+                  not found in the database will be omitted from the results. If none are
+                  found a 404 error is returned."""
+
+    def outputs(self):
+        return ['artist_credit_id', 'artist_mbid', 'country_code']
+
+    def fetch(self, params, count=-1, offset=-1):
+
+        with psycopg2.connect(config.MB_DATABASE_URI) as conn:
+            with conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as curs:
+
+                acs = tuple([r['[artist_credit_id]'] for r in params])
+                query = """ SELECT a.gid AS artist_mbid,
+                                    ar.id AS area_id,
+                                    code AS country_code,
+                                    ac.id AS artist_credit_id
+                               FROM artist a
+                               JOIN artist_credit_name acn
+                                 ON acn.artist = a.id
+                               JOIN artist_credit ac
+                                 ON acn.artist_credit = ac.id
+                               JOIN area ar
+                                 ON a.area = ar.id
+                    FULL OUTER JOIN iso_3166_1 iso
+                                 ON iso.area = ar.id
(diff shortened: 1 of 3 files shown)


listenbrainz/labs_api/labs/api/artist_country_from_artist_credit_id.py
```diff
@@ -0,0 +1,107 @@
...
+                                 ON a.area = ar.id
+                    FULL OUTER JOIN iso_3166_1 iso
+                                 ON iso.area = ar.id
+                              WHERE ac.id IN %s
+                           ORDER BY artist_credit_id"""
+
+                args = [acs]
+                if count > 0:
+                    query += " LIMIT %s"
+                    args.append(count)
```
alastair, line 51: If an AC has multiple artists, and you pass in multiple AC ids then this limit/offset is not going to be directly related to the number of items that were passed in. e.g.

```
artist1, ac1
artist2, ac1
artist3, ac1
artist3, ac2
artist4, ac2
```

a limit of 2 is only going to return some of the items from ac1. Does this make sense from a practical point of view? Not sure if it's expected that these arguments are going to return `count` artists regardless of how many ACs were added, or if it should relate to the number of ACs instead? I'm not convinced that this is actually a problem, and it could be OK.

listenbrainz/labs_api/labs/api/artist_country_from_artist_credit_id.py
```diff
@@ -0,0 +1,107 @@
...
+                               JOIN artist_credit_name acn
+                                 ON acn.artist = a.id
+                               JOIN artist_credit ac
+                                 ON acn.artist_credit = ac.id
+                               JOIN area ar
+                                 ON a.area = ar.id
+                    FULL OUTER JOIN iso_3166_1 iso
+                                 ON iso.area = ar.id
+                              WHERE ac.id IN %s
+                           ORDER BY artist_credit_id"""
```
alastair, line 46: if we have LIMIT and COUNT, then this ORDER BY needs something else to order by when the AC is the same - `ORDER BY artist_credit_id, artist.id` perhaps?

listenbrainz/labs_api/labs/api/artist_country_from_artist_credit_id.py
```diff
@@ -0,0 +1,107 @@
...
+                curs.execute(query, tuple(args))
+                areas = []
+                mapping = []
+                while True:
+                    row = curs.fetchone()
+                    if not row:
+                        break
+
+                    r = dict(row)
+                    areas.append(r['area_id'])
```
alastair, line 65: if there are multiple artists from the same area then this list will have duplicates. Perhaps a set?

listenbrainz/labs_api/labs/api/artist_country_from_artist_credit_id.py
```diff
@@ -0,0 +1,107 @@
...
+                    mapping.append(dict(row))
+
+                if not areas:
+                    return []
+
+                areas = tuple(areas)
+                curs.execute("""WITH RECURSIVE area_descendants AS (
```
alastair, line 72: I ran these queries with some test data. It looks like the first query will return a country if the artist's area is a country. If it's more specific than a country, this query loops through the area table finding the parent until it gets a country. Is that right? If so, it'd be good to have comment here saying that it only returns rows for areas that aren't countries

listenbrainz/labs_api/labs/api/artist_country_from_artist_credit_id.py
```diff
@@ -0,0 +1,107 @@
...
+class ArtistCountryFromArtistCreditIdQuery(Query):
+
+    def names(self):
+        return ("artist-country-code-from-artist-credit-id", "MusicBrainz Artist Country From Artist Credit Id")
+
+    def inputs(self):
+        return ['[artist_credit_id]']
+
+    def introduction(self):
+        return """Given artist credit id look up countries for those artists. Any artist_mbids
```
alastair, line 19: artist_mbids - should it say artist credit ids?

alastair (changes requested):
one question about limit/offset, and a few suggestions for documentation updates

mayhem:
No longer needed once the MBID mapping dump containers MBIDs for artists.

PR #1559, opened 2021-07-21
Add pin events to user timeline (backend)

jdaok:
(this PR doesn't depend on any currently opened PR's and can be merged)
This PR contains the backend changes to add pin events to the user timeline.

Changes:
- adds get_pins_for_feed() to pinned_recording.py to retrieve pins for list of users
- adds new class APIPinEvent to APIEventMetadata
- adds get_recording_pin_events() to user_timeline_event_api.py to fetch pins and convert to APITimelineEvents

frontend changes -> #1560


listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -360,3 +371,44 @@ def get_recording_recommendation_events(users_for_events: List[dict], min_ts: in
...
+        user_ids=(user['id'] for user in users_for_events),
+        min_ts=min_ts,
+        max_ts=max_ts,
+        count=count,
+    )
+
+    events = []
+    for pin in recording_pin_events_db:
+        try:
+            pin = fetch_track_metadata_for_pin(pin)
```
mayhem, line 390: This is a bit of a problem here -- given that the fetch_track_metadata_for_pin is in a loop, it means that we're going to do 1 DB query for each pin event, which is far from ideal. We need to support doing this in one query -- one possibility is to refactor the fetch_track_metadata_for_pin to accept a list of pins, rather than just one.

mayhem (changes requested):
Looks very good overall, we just need to refactor that one (two?) function(s).

listenbrainz/webserver/views/user_timeline_event_api.py
jdaok: Thanks, I opened #1570 to refactor the functions and fix this, after that gets merged we should be able to merge this PR.

mayhem: OK, 1570 is merged.

mayhem: Fix the merge conflict on 1560 and lets get this merged!

PR #1560, opened 2021-07-21
Add pin event to TimelineEvent component

jdaok:
(Based on changes in #1559, diff link [here](https://github.com/jdaok/listenbrainz-server/compare/pin-timeline-backend...jdaok:pin-timeline-frontend?expand=1))

This PR aims to add pin events to the user timeline. (frontend changes)
image

An `additionalDetails` prop is added to the TimelineEvent component which is used to render the pin's "blurb" under the track details.


paramsingh (approved):
Looks really good, excited to ship this!

PR #1561, opened 2021-07-21
BrainzPlayer: media notifications and media keys

MonkeyDo:
This PR adds two features (where supported):
- Notification API: on track change use native OS notifications
  - if notifications are allowed for this browser tab, uses the Notification API; otherwise fallback to a toast notification
  - where available, will display album artwork
  - displays in the notification tray on mobile
- MediaSession API allows us to use native browser/OS media player info
  - fixes usage of prev/next keyboard media keys
  - on Chromium browsers, styles automatically adapt to the album cover colors (see screenshots below)
  - requires a sort of hack to overwrite the MediaSession created by the embedded players' iframes
- while in there I fixed LB-625

Native notifications:
image

Fallback toast notification (bottom right):
image

MediaSession:
image
image
image


listenbrainz/webserver/static/js/src/SoundcloudPlayer.tsx
```diff
@@ -58,7 +58,6 @@ export default class SoundcloudPlayer
   implements DataSourceType {
   iFrameRef?: React.RefObject<HTMLIFrameElement>;
   soundcloudPlayer?: SoundCloudHTML5Widget;
-  supportsSearch = false;
```
MonkeyDo, line 61: This is probably leftover from trying something wy back when, was never used AFAIK and was never cleaned up.

PR #1564, opened 2021-07-25
LB-921 (I): A treasure trove of rabbitmq nuances

amCap1712:
### Fix consumer ack timeouts in spark request consumer

We originally tried to fix this in c13b9f0114ad93ec21ec2a2ce42891c453c4b102. The fix worked fine and we did not recompute requests again and again in spark. However, we continued to get a consumer ack timeout daily during the cron job runs.

It turns out that rabbitmq tries to send as many messages possible at a time by default. All of these messages are delivered and marked as unacked. The callback is called on first message. We ack it right away and then execute the spark request. Once the request has finished the method returns and the callback is executed on the next message. Now if the first request took a considerable amount of time to execute, we have already hit a timeout on the remaining messages before we reach them. This is probably why we have been seeing the consumer ack timeouts till now.

To fix this, we request rabbitmq to send only one message at a time and ack the message after it has been processed. (If we ack'ed right away, rabbitmq would send another message and we are back at the above described problem.) The next message is not sent until the current one has been ack'ed so the issue of timeout is mitigated. 

*Before applying this fix*
|name|messages|messages_ready|messages_unacknowledged|
|------|-----------|------------------|------------------------------|
|spark_request|10|0|10|

*After applying this fix*
|name|messages|messages_ready|messages_unacknowledged|
|------|-----------|------------------|------------------------------|
|spark_request|10|9 |1|

### DIFF
--- a/listenbrainz_spark/request_consumer/request_consumer.py
+++ b/listenbrainz_spark/request_consumer/request_consumer.py
@@ -57,20 +57,6 @@ class RequestConsumer:
             logger.error("Error while mapping query to function:", exc_info=True)
             return None
 
-        try:
-            # initialize connection to HDFS, the request consumer is a long running process
-            # so we try to create a connection everytime before executing a query to avoid
-            # affecting subsequent queries in case there's an intermittent connection issue
-            hdfs_connection.init_hdfs(config.HDFS_HTTP_URI)
-            return query_handler(**params)
-        except TypeError as e:
-            logger.error(
-                "TypeError in the query handler for query '%s', maybe bad params. Error: %s", query, str(e), exc_info=True)
-            return None
-        except Exception as e:
-            logger.error("Error in the query handler for query '%s': %s", query, str(e), exc_info=True)
-            return None
-
         return query_handler, params
 
     def callback(self, channel, method, properties, body):
--- a/listenbrainz_spark/request_consumer/request_consumer.py
+++ b/listenbrainz_spark/request_consumer/request_consumer.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+import threading
 
 import pika
 import json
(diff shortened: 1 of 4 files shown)


amCap1712:
I tested this with the all_time recordings task and it didn't work as well as expected. We start hitting into the ChannelWrongStateError. So in addition to this fix we need to offload the call to spark to a different thread.
https://stackoverflow.com/a/52951933

listenbrainz_spark/request_consumer/request_consumer.py
```diff
@@ -103,6 +103,7 @@ def push_to_result_queue(self, messages):
     def callback(self, channel, method, properties, body):
         request = json.loads(body.decode('utf-8'))
         logger.info('Received a request!')
+
         while True:
             try:
                 self.request_channel.basic_ack(delivery_tag=method.delivery_tag)
```
amCap1712, line 109: I went down the rabbit hole of the rabbitmq timeouts etc. and looks like this is problematic. RabbitMQ mandates that the `basic_ack` be sent from the same channel that received the message. So if we get an error during acking, it's already too late. There's no point of retrying.

amCap1712: https://github.com/metabrainz/listenbrainz-server/blob/b3fec9f9ea9294a69b0c588a60e3ecf6773559f8/listenbrainz_spark/request_consumer/request_consumer.py#L77-L91

This is an even bigger issue here. If we encounter an error during publishing the results, we close the connection and create new connections and channels for both consuming and publishing. This is erroneous because the original channel which received the request got closed so we cannot ack the request later. We should just be recreating the result channel here.

amCap1712:
https://github.com/metabrainz/listenbrainz-server/blob/b3fec9f9ea9294a69b0c588a60e3ecf6773559f8/listenbrainz_spark/request_consumer/request_consumer.py#L31

In the current version of rabbitmq, the max allowed heartbeat is 60s. The alternative is to disable heartbeats completely but that is only possible if heartbeats are disabled on server-side as well which means heartbeats are disabled across all MeB projects, which does not sound good.

mayhem (approved):
OK, lets give this a try and see if it improves things.

PR #1569, opened 2021-07-28
LB-867: Add unicode null checking to api endpoints

amCap1712:
Postgres does not allow `\u0000` character. This has brought down the TS Writer down twice before. Also, added a couple of tests so that we don't regress this in future.

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -121,9 +121,29 @@ def _send_listens_to_queue(listen_type, listens):
         )
 
 
+def _raise_error_if_has_unicode_null(value, listen):
+    if isinstance(value, str) and '\x00' in value:
+        raise APIBadRequest("{} contains a unicode null".format(value), listen)
+
+
+def check_for_unicode_null_recursively(listen: Dict):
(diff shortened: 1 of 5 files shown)


listenbrainz/webserver/views/api_tools.py
```diff
@@ -121,9 +121,29 @@ def _send_listens_to_queue(listen_type, listens):
         )
 
 
+def _raise_error_if_has_unicode_null(value, listen):
+    if isinstance(value, str) and '\x00' in value:
+        raise APIBadRequest("{} contains a unicode null".format(value), listen)
+
+
+def check_for_unicode_null_recursively(listen: Dict):
```
amCap1712, line 129: Alternative way is to first call `flatten_dict` and then check for nulls. That was the way it is done in `listen.py`. I think either way is fine but didn't use that currently because it doesn't handle lists.

amCap1712:
Failing test is trying to confirm the old way to check nulls works. I'll move that test to the api level instead.

mayhem (approved):
So much time spent checking incoming listens.. :( Clearly necessary buy very slow in python. We'll have to address the speed of this before long if our traffic picks up.

PR #1570, opened 2021-07-30
Refactor load recording from UUID functions

jdaok:
This PR refactors the helper functions added in #1547 (load_recording_from_msid(), load_recording_from_mbid(), to support doing lookups in one query. 

Both functions were renamed and changed so that they take in a list of UUID's instead of just one, and return a list of dicts in the same order instead of just one dict.


mayhem (approved):
Looks good, thanks!

PR #1572, opened 2021-08-02
Play icon button on listens

MonkeyDo:
At some point in the recent past, the play button that showed on listens when hovered was removed.
It was a good clear indication that the listens are playable rather than relying on the user knowing a double click would play the track.

On touch devices, the play icon will be shown at all times, since relying on hover does not make sense on touch devices.

image


akshaaatt (commented):
Amazing work! I really like the functionality and the distinction for both web and mobile experience is good. I have tested it and under ideal conditions, it works well.

I have two thoughts though:
- We should include a `pause` button once the user has clicked the play icon. Currently it just replays the recording. We could add this in another PR if you say so but I suggest we complete it with this.
- At one point, I did not have my spotify premium account and upon clicking the play button, the following error message was displayed 
`Objects are not valid as a React child (found: object with keys {code, message, errors, status}). If you meant to render a collection of children, use an array instead. in div (created by Alert) in div ....`
So we can do better error handling here.

Great work btw!

MonkeyDo:
Thanks for the review @akshaaatt .

I liked the idea of the pause button, but after trying to implement it I realize it's not at all easily done. The reason is that while we pass a `playListen` function from RecentListens to ListenCard, all of the pausing mechanism is done in BrainzPlayer, which is created by RecentListen.
We don't have access to any of the pausing mechanism in RecentListens to be able to pass it down to ListenCard and I think it would be quite the refactor to make it work.

Instead, we do have access to whether the listen is currently playing, and with that we can change the icon and action of the play button.
On top of the previous mechanism of hovering, now on the listen that is currently playing we are always showing a slightly different icon (play icon without the circle around it) and clicking the button/ double-clicking the card does nothing (instead of restarting the song):
image

I'll try to reproduce the issue you mentioned, thanks for reporting !

akshaaatt:
Works great! Thanks :100:

PR #1573, opened 2021-08-03
Improve artist credit recording pairs query

mayhem:
Improve artist credit recording pairs query to be more stable, which should improve stability of tracks being matched to releases that are identical for all of our criteria. 

Update tests for the new sort order created above.

Improve docs/indexes to reflect current reality.


amCap1712 (approved):
LGTM

PR #1574, opened 2021-08-04
LB-866 (I): Add user_id column to listen table

alastair:
This is the first part of LB-866

# Problem
The listens table includes a user_name field. This makes it difficult to rename user accounts, as it requires rewriting rows to make this change.

# Solution
Switch to using user_id instead.

# Action
This process will require a few PRs:
1. setup (this pr)
    1. Add a new user_id column without locking the table
    2. start writing user_id for new listens that come in
    3. in the background, update user_id for all old listens
2. switchover to new column
    1. Update all read methods to use the user_id field instead of user_name
    2. Update dumps to select user_name from the main database user table
3. clean up
    1. Remove files used for the migration
    2. delete user_name column

This PR has the following steps:
1. Add new user_id column.
We do a few magic steps here:
    * Postgres 11 can create a new non-null column with a default value without locking the table: https://dataegret.com/2018/03/waiting-for-postgresql-11-pain-free-add-column-with-non-null-defaults/
    this means that we can create the column and set the constraint without having to take down the site, and it means we don't need to lock the table at a later stage to add the constraint. I used a sentinel user_id of 0 to represent "null" in this case, but by the end of this process there should be no more listens with this id remaining.
    I found during development that timescale causes a few problems for us here. It seems that even in pg 11 it takes an exclusive lock and then disables autovacuum on each underlying chunk, before creating the column then deleting the lock. The [solution here seems to be to disable autovacuum](https://github.com/metabrainz/listenbrainz-server/blob/bbb8b41b55d479f26fa8e482fce9338b510ae9e2/admin/timescale/updates/2021-08-03-listens-add-user-id.sql#L60) on this table (which takes a while but doesn't require a lock), and then [add the new column](https://github.com/metabrainz/listenbrainz-server/blob/bbb8b41b55d479f26fa8e482fce9338b510ae9e2/admin/timescale/updates/2021-08-03-listens-add-user-id.sql#L11), which is fast.
    * Next, we need an index on this field to filter on user_id. Creating an index needs an exclusive lock, and timescale doesn't support `CREATE INDEX CONCURRENTLY`, which would solve this problem. Instead, we use `CREATE INDEX ... WITH (timescaledb.transaction_per_chunk)`, which only locks one chunk at a time. In my testing this seemed to be quite quick, and I was able to read from and insert into the listens table while this was happening
2. start writing user_id for new listens that come in.
    We [add both user_name and user_id](https://github.com/metabrainz/listenbrainz-server/compare/listen-userid?expand=1#diff-294c8f9593ce2f2f2e883afba6c54df5e236dc48dc90741311afe6e42ab0f84aR236). The site keeps reading from user_name for now
3. in the background, update user_id for all old listens
    This is done by `manage.py listen-add-userid`. It works per chunk (5 days), finds all unique usernames in this chunk, looks up the user_Id for each name, and performs a bulk update of the chunk.
    I initially tried to do a single update for each username, wrapping all updates in a transaction, but this took about 10x longer than the bulk update. I tested both approaches and they give the same result.
    This script can be quit and re-run. It will check each chunk, but if it has no missing user_id values then it will skip it and move on to the next.

While performing the user_id update, I found a few rows containing listens for users who don't exist in the users table. The update script will report these, and we can remove these listens.

### DIFF
--- a/admin/timescale/updates/2021-08-03-after-migration-listens-add-user-id.sql
+++ b/admin/timescale/updates/2021-08-03-after-migration-listens-add-user-id.sql
@@ -1,9 +1,0 @@
--- index so that we can quickly see which listens have user_id of 0. We'll need it when selecting
--- on this field after we remove user_name, too.
--- It doesn't appear to affect bulk update speed too much with this index in place
--- CREATE INDEX will lock the whole table, which means that it'll be unavailable while the index
--- is building. Timescale doesn't support CREATE INDEX CONCURRENTLY. The best we can do is this
--- extension which opens a transaction and creates the index one chunk at a time.
--- This lets inserts and selects work on the table while it's running
--- This operation doesn't support being run inside a transaction.
-create index user_id_listen on listen(user_id) WITH (timescaledb.transaction_per_chunk);
--- a/admin/timescale/updates/2021-08-03-after-migration-listens-add-user-id.sql
+++ b/admin/timescale/updates/2021-08-03-after-migration-listens-add-user-id.sql
@@ -1,0 +1,9 @@
+-- index so that we can quickly see which listens have user_id of 0. We'll need it when selecting
+-- on this field after we remove user_name, too.
+-- It doesn't appear to affect bulk update speed too much with this index in place
+-- CREATE INDEX will lock the whole table, which means that it'll be unavailable while the index
+-- is building. Timescale doesn't support CREATE INDEX CONCURRENTLY. The best we can do is this
+-- extension which opens a transaction and creates the index one chunk at a time.
+-- This lets inserts and selects work on the table while it's running
+-- This operation doesn't support being run inside a transaction.
+create index user_id_listen on listen(user_id) WITH (timescaledb.transaction_per_chunk);
--- a/admin/timescale/updates/2021-08-03-before-migration-listens-add-user-id.sql
+++ b/admin/timescale/updates/2021-08-03-before-migration-listens-add-user-id.sql
@@ -10,17 +10,6 @@ alter table listen set (autovacuum_enabled = false);
 -- After the migrate script runs we shouldn't have any listens with a user_id of 0
 alter table listen add column user_id integer not null default 0;
 
-COMMIT;
-
--- index so that we can quickly see which listens have user_id of 0. We'll need it when selecting
--- on this field after we remove user_name, too.
--- It doesn't appear to affect bulk update speed too much with this index in place
--- CREATE INDEX will lock the whole table, which means that it'll be unavailable while the index
--- is building. Timescale doesn't support CREATE INDEX CONCURRENTLY. The best we can do is this
--- extension which opens a transaction and creates the index one chunk at a time.
--- This lets inserts and selects work on the table while it's running
--- This operation doesn't support being run inside a transaction.
-create index user_id_listen on listen(user_id) WITH (timescaledb.transaction_per_chunk);
-
 -- turn autovacuum back on
 alter table listen set (autovacuum_enabled = true);
+COMMIT;
--- a/admin/timescale/updates/2021-08-03-listens-add-user-id.sql
+++ b/admin/timescale/updates/2021-08-03-listens-add-user-id.sql
@@ -1,0 +1,26 @@
+BEGIN;
+
+-- When adding a column, timescale goes through and disables the autovacuum for each chunk
+-- but it does this _after_ it takes an exclusive lock on the table.
+-- By disabling it manually first we make the column add operation fast
+alter table listen set (autovacuum_enabled = false);
+-- Postgres 11+ can add a column with non-null + default without locking the table
+-- https://dataegret.com/2018/03/waiting-for-postgresql-11-pain-free-add-column-with-non-null-defaults/
+-- we use user_id=0 as a sentinel (no user exists with this id)
+-- After the migrate script runs we shouldn't have any listens with a user_id of 0
+alter table listen add column user_id integer not null default 0;
+
+COMMIT;
+
+-- index so that we can quickly see which listens have user_id of 0. We'll need it when selecting
+-- on this field after we remove user_name, too.
+-- It doesn't appear to affect bulk update speed too much with this index in place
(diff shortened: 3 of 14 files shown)


alastair:
I ran this process on a test database of 20 million listens, while refreshing my user page and running a spotify importer to test reads and writes. The local site appeared to stay responsive while the column add/index creation was happening.
I'll download a copy of the full database (or a larger subset) and test it again to double-check, it's possible that due to the small size of my local db some things happened quickly which may take longer on a bigger db.

On my machine, it took about 10 seconds per chunk to process.

admin/timescale/updates/2021-08-03-listens-add-user-id.sql
```diff
@@ -0,0 +1,26 @@
+BEGIN;
+
+-- When adding a column, timescale goes through and disables the autovacuum for each chunk
```
mayhem, line 3: Thank you for these great comments.

admin/timescale/updates/2021-08-03-listens-add-user-id.sql
```diff
@@ -0,0 +1,26 @@
...
+
+-- index so that we can quickly see which listens have user_id of 0. We'll need it when selecting
+-- on this field after we remove user_name, too.
+-- It doesn't appear to affect bulk update speed too much with this index in place
+-- CREATE INDEX will lock the whole table, which means that it'll be unavailable while the index
+-- is building. Timescale doesn't support CREATE INDEX CONCURRENTLY. The best we can do is this
+-- extension which opens a transaction and creates the index one chunk at a time.
+-- This lets inserts and selects work on the table while it's running
+-- This operation doesn't support being run inside a transaction.
+create index user_id_listen on listen(user_id) WITH (timescaledb.transaction_per_chunk);
```
mayhem, line 23: Hmm, this creates an index on a column that has all the same value, which is... Questionable? I wonder if this step should be done later, AFTER data has been added, but BEFORE we start using the the column. thinking out loud...

admin/timescale/updates/2021-08-03-before-migration-listens-add-user-id.sql
```diff
@@ -0,0 +1,15 @@
...
+
+-- When adding a column, timescale goes through and disables the autovacuum for each chunk
+-- but it does this _after_ it takes an exclusive lock on the table.
+-- By disabling it manually first we make the column add operation fast
+alter table listen set (autovacuum_enabled = false);
+-- Postgres 11+ can add a column with non-null + default without locking the table
+-- https://dataegret.com/2018/03/waiting-for-postgresql-11-pain-free-add-column-with-non-null-defaults/
+-- we use user_id=0 as a sentinel (no user exists with this id)
+-- After the migrate script runs we shouldn't have any listens with a user_id of 0
+alter table listen add column user_id integer not null default 0;
```
amCap1712, line 11: Don't think it should matter much, but this will add the column to the end of table whereas the create table commands adds it after user_name column.

listenbrainz/listenstore/timescale_fill_userid.py
```diff
@@ -0,0 +1,144 @@
...
+        chunk_start += CHUNK_SECONDS
+
+    print("Finished.")
+    print(f"Usernames who have listens but no known user row: {global_missing_users}")
+
+
+def print_status_update(chunk_count, number_chunks, start_time):
+    """Print a basic status update based on how many chunks have been computed"""
+    chunk_time = time.monotonic()
+    chunk_percentage = chunk_count / number_chunks
```
alastair, line 136: one small bug that I noticed here, if number_chunks is 0 (for example if all chunks are done and there are none remaining) then it's a /0 error. I'd just move this whole calculation into the if block below, and then don't print any time data in the else

admin/timescale/updates/2021-08-03-after-migration-listens-add-user-id.sql
```diff
@@ -0,0 +1,9 @@
+-- index so that we can quickly see which listens have user_id of 0. We'll need it when selecting
+-- on this field after we remove user_name, too.
+-- It doesn't appear to affect bulk update speed too much with this index in place
+-- CREATE INDEX will lock the whole table, which means that it'll be unavailable while the index
+-- is building. Timescale doesn't support CREATE INDEX CONCURRENTLY. The best we can do is this
+-- extension which opens a transaction and creates the index one chunk at a time.
+-- This lets inserts and selects work on the table while it's running
+-- This operation doesn't support being run inside a transaction.
+create index user_id_listen on listen(user_id) WITH (timescaledb.transaction_per_chunk);
```
alastair, line 9: I created this specific index only to be able to check which listen rows hadn't been updated when restarting the update script if it crashed (see the comment in line 1).
We'll need an index on user_id in order to select listens, but that'll probably be an index over a few different columns - see the current indexes on the listen table that use user_name, they're all combined with other columns:
```
Indexes:
    "listened_at_track_name_user_name_ndx_listen" UNIQUE, btree (listened_at DESC, track_name, user_name)
    "listen_listened_at_idx" btree (listened_at DESC)
    "listened_at_user_name_ndx_listen" btree (listened_at DESC, user_name)
```

Therefore I think we can get rid of this file, and add the index in part 2 of the issue

alastair (commented):
@amCap1712 I re-ran the migration process without the index. If the process quits and we have to restart it, the initial query that is used to get the min/max listened_at does take a long time to complete, but 1) we only need to do this if the process quits, and 2) it's only a one-time process. So as we discussed this is fine.

PR looks good to go now. Let's do it.

admin/timescale/updates/2021-08-03-before-migration-listens-add-user-id.sql
alastair: I have no problem with this

admin/timescale/updates/2021-08-03-listens-add-user-id.sql
alastair: the idea for this index was only to make the migration script fast at startup if it quit and we had to restart it. We tested without the index and while it's slower to perform the initial query, we don't expect to run into this case when we perform the migration. Index has been removed.

amCap1712 (approved):

PR #1575, opened 2021-08-10
Fix invalid parameters on charts page, fix charts navigation links

alastair:
# Problems
When being passed a `?page=0` query parameter, the charts page subtracts 1 from this in order to compute an offset, resulting in a negative offset and an error from the server (https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/126368/)
-> Update the page parsing to not allow it to go below 1.

The next/previous links in the charts page had an empty a href. This means that the hover value of the link was invalid, and the pages were unable to be copied or opened in a new tab (LB-864)
-> Add values to the href property

### DIFF
--- a/listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
+++ b/listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
@@ -78,7 +78,7 @@ export default class UserEntityChart extends React.Component<
     window.history.replaceState(
       null,
       "",
-      `?page=${page}&range=${range}&entity=${entity}`
+      this.buildURLParams(page, range, entity)
     );
     this.syncStateWithURL();
     this.handleResize();
@@ -89,27 +89,13 @@ export default class UserEntityChart extends React.Component<
     window.removeEventListener("resize", this.handleResize);
   }
 
-  changePage = (
-    newPage: number,
-    event?: React.MouseEvent<HTMLElement>
-  ): void => {
-    if (event) {
-      event.preventDefault();
-    }
-
+  changePage = (newPage: number): void => {
     const { entity, range } = this.state;
     this.setURLParams(newPage, range, entity);
     this.syncStateWithURL();
   };
 
-  changeRange = (
-    newRange: UserStatsAPIRange,
-    event?: React.MouseEvent<HTMLElement>
-  ): void => {
-    if (event) {
-      event.preventDefault();
-    }
-
+  changeRange = (newRange: UserStatsAPIRange): void => {
     const { entity } = this.state;
     this.setURLParams(1, newRange, entity);
     this.syncStateWithURL();
@@ -378,23 +364,38 @@ export default class UserEntityChart extends React.Component<
     );
   };
 
+  /*
+  Build a url querystring (including the ?) for a page, range, and entity
+   */
   buildURLParams = (
     page: number,
     range: UserStatsAPIRange,
     entity: Entity
   ): string => {
-    const params = new URLSearchParams({
-      page: page.toString(),
-      range: range as string,
-      entity: entity as string,
-    }).toString();
-    return `?${params}`;
+    return `?page=${page}&range=${range}&entity=${entity}`;
   };
 
   handleResize = () => {
     this.setState({
       graphContainerWidth: this.graphContainer.current?.offsetWidth,
     });
+  };
+
+  /*
+  Handle a click on a link that will perform an action
+   - if control is held down, don't prevent the event from firing
+     this will allow ctrl/cmd-click to open links in a new tab
+   - otherwise, prevent the event from happening and call a callback
+     that performs some action (e.g. load a new page)
+   */
+  handleClickEvent = (
+    e: React.MouseEvent<HTMLAnchorElement>,
+    callback: () => any
+  ) => {
+    if (!e.ctrlKey) {
+      e.preventDefault();
+      callback();
+    }
   };
 
   render() {
@@ -466,7 +467,11 @@ export default class UserEntityChart extends React.Component<
                       <a
                         href={this.buildURLParams(1, "week", entity)}
                         role="button"
-                        onClick={(event) => this.changeRange("week", event)}
+                        onClick={(e) => {
+                          this.handleClickEvent(e, () => {
+                            this.changeRange("week");
+                          });
+                        }}
                       >
                         Week
                       </a>
@@ -475,7 +480,11 @@ export default class UserEntityChart extends React.Component<
                       <a
                         href={this.buildURLParams(1, "month", entity)}
                         role="button"
-                        onClick={(event) => this.changeRange("month", event)}
+                        onClick={(e) => {
+                          this.handleClickEvent(e, () => {
+                            this.changeRange("month");
+                          });
+                        }}
                       >
                         Month
                       </a>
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
```diff
@@ -373,10 +360,22 @@ export default class UserEntityChart extends React.Component<
...
+      this.buildURLParams(page, range, entity)
     );
   };
 
+  buildURLParams = (
+    page: number,
+    range: UserStatsAPIRange,
+    entity: Entity
+  ): string => {
+    return new URLSearchParams({
```
alastair, line 372: URLSearchParams isn't supported in IE, and honestly this method is huge... Happy to revert it back to string interpolation like the setURLParams method used to be if we want to make this shorter...

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
```diff
@@ -450,36 +449,48 @@ export default class UserEntityChart extends React.Component<
                   <ul className="dropdown-menu" role="menu">
                     <li>
                       <a
-                        href=""
+                        href={`?${this.buildURLParams(1, "week", entity)}`}
                         role="button"
-                        onClick={(event) => this.changeRange("week", event)}
+                        onClick={(e) => {
+                          e.preventDefault();
```
alastair, line 455: I moved this from changeRange (and from changePage) because to me it seems odd to pass a parameter into the method when it's not really used. While this change does result in more lines of code, I think it's clearer to show that the value of the event isn't important for the changeRange method.
Happy to discuss this

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
```diff
@@ -450,36 +449,48 @@ export default class UserEntityChart extends React.Component<
                   <ul className="dropdown-menu" role="menu">
                     <li>
                       <a
-                        href=""
+                        href={`?${this.buildURLParams(1, "week", entity)}`}
```
alastair, line 452: Some of these parameters are now duplicated, and in different places - the page number 1 is duplicated here and in the changeRange method. the range parameter ("week") is listed twice. Maybe this could be rewritten a bit so that we could have a single object (or search string?) that can be passed in to multiple places.
It does mean that changeRange, changeEntity and changePage are actually all the same method, just with some different parameters. Perhaps this could be simplified into a single method.

alastair:
This bug may also be present [in the recommendations page](https://github.com/metabrainz/listenbrainz-server/blob/6ad85cc449dd5b33a5fe90c89ddafec32f1b13ce/listenbrainz/webserver/static/js/src/recommendations/Recommendations.tsx#L166)

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
MonkeyDo: Good question. I can't really think of many arguments one way or the other.
I guess if URLSearchParams doesn't bring any extra functionality and has worse browser support we should probably revert to string interpolation, or if we think there are potential benefits to URLSearchParams we can check for support and use our previous version as a fallback (`if(window.URLSearchParams){…} else {//template string instead}`)

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
```diff
@@ -373,10 +360,23 @@ export default class UserEntityChart extends React.Component<
     window.history.pushState(
       null,
       "",
-      `?page=${page}&range=${range}&entity=${entity}`
+      this.buildURLParams(page, range, entity)
```
MonkeyDo, line 363: There is one more instance or the template string at line 81 in `componentDidMount` that should be changed

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
MonkeyDo: I think it could be an issue when for example we try to fix the bug we currently have with using ctrl+click to open in a new tab.

If the solution is to check a condition of the event (is ctrl key pressed?) before deciding what to do, it would be better to only have to do it in one spot.

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
```diff
@@ -553,9 +565,12 @@ export default class UserEntityChart extends React.Component<
                       }`}
                     >
                       <a
```
MonkeyDo, line 567: These anchor tags aren't applied the disabled style correctly anymore and remain clickable, even though the `disabled` class is applied to the parent <li> element.
Not sure what's going on there…

MonkeyDo:
Forgot to say, the fix for the original issue (the ?page=0 thing) definitely works :+1: !

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
alastair: I'll switch it back to a string interpolation, which takes this method back down to 1 line of code.

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
alastair: that's a good point - I found the control-click fix after I made this change. Keep in mind that we'd want this check in many different methods, not just `changeRange`, so I'm not sure if we should make another `handleClick` method (with callback?) or add it to each of the click handlers that we have around the app.

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
alastair: thanks!

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
alastair: strange, I didn't notice this when I made the changes. I'll have a look again. Is it a bit weird that the css class is related to `disabled` being on the `<li>`? Maybe it should be on the anchor anyway, or perhaps we should update the logic to use a `<span>` instead of anchor when it's disabled

alastair: This is because now that the anchor tag has an href, a rule for `a:visited` is overriding the rule that we have for this colour: https://github.com/metabrainz/listenbrainz-server/blob/f0bc6969ad8f5ddef1fb0d788998f2fcc45f3dad/listenbrainz/webserver/static/css/theme/links.less#L36

I was able to restore the correct colour by disabling these rules in the inspector:
Screenshot from 2021-09-13 14-25-43

I don't know how to update the styles to make this work

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
alastair: I updated this PR to include support for ctrl-click

alastair:
I've updated this, but not touched the css, as I'm not sure what the best change would be to fix this. @MonkeyDo, any ideas?

MonkeyDo:
The one failing front-end test is unrelated to this PR, and has to do with the changes brought in #1603 and will be addressed in a separate PR.
I'll merge this now then as it's the only failing test.

PR #1578, opened 2021-08-11
Check is row is not None before accessing similarity

amCap1712:
Bunch of errors in sentry like this one https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/128319/. Fix is to check for None before trying to access fields in the row.

### DIFF
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List
+from typing import List, Optional
 
 import sqlalchemy
 import uuid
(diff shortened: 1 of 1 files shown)


listenbrainz/db/user.py
```diff
@@ -492,10 +492,12 @@ def get_similar_users(user_id: int) -> SimilarUsers:
             'user_id': user_id,
         })
         row = result.fetchone()
-        users = {}
-        for user in row[1]:
-            users[user] = row[1][user][0]
-        return SimilarUsers(user_id=row[0], similar_users=users) if row else None
+        if row:
+            users = {}
+            for user in row[1]:
```
alastair, line 497: not directly related to this change, but can we use dict access here to make it clearer what we're selecting? `row['similar_users']`

alastair (approved):
bug fixes!

amCap1712:
Commenting from preview.

PR #1579, opened 2021-08-12
Minor PinnedRecording backend changes

jdaok:
(this PR doesn't depend on any others to be merged)

This PR mainly makes 2 changes to pins:
1) renaming `pinned_id `attribute to `row_id` everywhere: Pinned_id's were being returned by the API while row_id was used in front-end, so I'm renaming it to avoid confusion.
2)  `track_metadata` is now returned by the API when fetching pins. This way, we don't have to make additional API calls from the frontend to obtain the track name and artist name.


mayhem (approved):
Looks good. We'll merge after the passing tests.

PR #1581, opened 2021-08-13
Make pins listenable on BrainzPlayer

jdaok:
(this PR doesn't depend on any others to be merged)

This PR aims to make pins listenable on BrainzPlayer (and also changes to take an object instead of an #id.)
Props `className`, `playListen()`, and `removePinFromPinsList()` are added to PinnedRecordingCard component. 

chrome-capture (1)


listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -564,7 +570,14 @@ export default class RecentListens extends React.Component<
...
                   userName={user.name}
                   pinnedRecording={userPinnedRecording}
+                  className={
+                    this.isCurrentListen(getListenablePin(userPinnedRecording))
+                      ? " current-listen"
+                      : ""
+                  }
                   isCurrentUser={currentUser?.name === user?.name}
+                  playListen={this.playListen}
+                  removePinFromPinsList={() => {}}
```
amCap1712, line 580: To be implemented in future or forgot to commit?

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -564,7 +570,14 @@ export default class RecentListens extends React.Component<
                 <PinnedRecordingCard
                   userName={user.name}
                   pinnedRecording={userPinnedRecording}
+                  className={
+                    this.isCurrentListen(getListenablePin(userPinnedRecording))
```
amCap1712, line 574: I am not sure this will work as expected. Should this effect also take place if a listen of the same recording which is pinned is being played ? In that case, this won't work. I am fine either way but wanted to confirm the intent.

listenbrainz/webserver/static/js/src/RecentListens.tsx
jdaok: There's no need to pass in a function that does anything since there's no pin list in this component, only in userPins. removePinFromPinsList could also be made into an optional prop but I figured this looks cleaner.

listenbrainz/webserver/static/js/src/RecentListens.tsx
jdaok: Yes it works as intended, if BrainzPlayer is playing something only one card component should be highlighted at a time whether it's a listen or pin.

PR #1583, opened 2021-08-16
Pin history view

jdaok:
(based off #1581, diff link [here](https://github.com/jdaok/listenbrainz-server/compare/make-pins-listenable-on-brainzplayer..jdaok:pin-hist-view))

TO DO: 
- [x] add userPins.test.tsx

This PR aims to add a Pins tab to profiles to allow users to view any recordings they have pinned.
image


listenbrainz/webserver/views/user.py
```diff
@@ -349,6 +350,35 @@ def collaborations(user_name: str):
     )
 
 
+@user_bp.route("/<user_name>/pins/")
```
amCap1712, line 353: Need to add methods and CORS header like https://github.com/metabrainz/listenbrainz-server/pull/1588

listenbrainz/webserver/views/user.py
```diff
@@ -349,6 +350,35 @@ def collaborations(user_name: str):
...
+def pins(user_name: str):
```
amCap1712, line 354: This endpoint should probably be paged, like accepting a count and offset in the query parameters and passing it to `get_pin_history_for_user`.

amCap1712:
Sorry! You can ignore the above two comments. I mistook it to be an API endpoint.

amCap1712:
LGTM. Thanks!

PR #1585, opened 2021-08-22
CritiqueBrainz Review Modal

jdaok:
Please do not merge until
- [x] https://github.com/metabrainz/critiquebrainz/pull/306
- [x] https://github.com/metabrainz/critiquebrainz/pull/303
- [x] https://github.com/metabrainz/critiquebrainz/pull/371
- [x] `CBReviewModal.test.tsx` added

**To test the modal, build #1586 instead**

image
ezgif-3-60ef907d4b1d

I added two libraries: [iso-639-1](https://www.npmjs.com/package/iso-639-1) to get the languages needed for the dropdown selector, and [React Simple Star Rating](https://www.npmjs.com/package/react-simple-star-rating) to implement the 5 star rating, which requires that React is also updated to a newer version.


listenbrainz/webserver/templates/user/profile.html
```diff
@@ -12,4 +12,10 @@
 {% block scripts %}
   {{ super() }}
   <script src="{{ get_static_path('recentListens.js') }}" type="text/javascript"></script>
+  <script>
```
amCap1712, line 15: Can this be done in the react component instead?

listenbrainz/webserver/static/js/src/types.d.ts
```diff
@@ -40,6 +40,9 @@ interface AdditionalInfo {
   track_mbid?: string | null;
   tracknumber?: number | null;
   work_mbids?: Array<string> | null;
+  lastfm_artist_mbid?: string | null;
+  lastfm_release_mbid?: string | null;
+  lastfm_track_mbid?: string | null;
```
amCap1712, line 45: This looks unrelated.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,710 @@
...
+  };
+
+  getArtistEntity = () => {
+    this.setState({ artistEntity: null });
+
+    const { listen } = this.props;
+    const { additional_info } = listen.track_metadata;
+    let artist_mbid;
+
+    if (additional_info?.lastfm_artist_mbid)
```
amCap1712, line 241: Oh! I see its used here. My understanding from the chat logs was that lastfm supplied mbids are not trust worthy and hence should not be used. @mayhem thoughts?

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+            name="review-text"
+            onChange={this.handleTextInputChange}
+            rows={6}
+            style={{ resize: "vertical" }}
+            spellCheck="false"
+            required
+          />
+        </div>
+        <small style={{ display: "block", textAlign: "right" }}>
+          Words: {countWords(textContent)} / Chars: {textContent.length}
```
MonkeyDo, line 530: It would be nice to style this text to give a direct indication to the user if they are under the required.
Something like :
```
<small className={textContent.length < this.minTextLength ? 'text-danger' : ''} style={{ display: "block", textAlign: "right" }}>

File: listenbrainz/webserver/static/js/src/CBReviewModal.tsx [hunk 5]
```diff
@@ -0,0 +1,703 @@
...
+        </small>
+
+        <div className="rating-container">
+          <b>Rating: </b>
```
MonkeyDo, line 534: If the rating is optional, it should be marked as such (or the other fields marked as required)

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+            onClick={(rate: number) => this.setState({ rating: rate })}
+            ratingValue={rating}
+            transition
+            size={20}
+            stars={5}
+          />
+        </div>
+
+        <div className="dropdown">
+          <b>Language: </b>
```
MonkeyDo, line 546: Do you think this could be misinterpreted?
Is a clarification like this required?
```suggestion
          <b>Language of your review: </b>
```

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+          <label>
+            <input
+              id="acceptLicense"
+              type="checkbox"
+              checked={acceptLicense}
+              name="acceptLicense"
+              onChange={this.handleInputChange}
+              required
+            />
+            <small>
```
MonkeyDo, line 573: Damn, that's a long and intimidating checkbox…
I guess there's no good way around it, just like in CB itself.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+      <div
+        className="modal fade"
+        id="CBReviewModal"
+        tabIndex={-1}
+        role="dialog"
+        aria-labelledby="CBReviewModalLabel"
+        data-backdrop="static"
+      >
+        <div className="modal-dialog" role="document">
+          <form className="modal-content" onSubmit={(e) => e.preventDefault()}>
```
MonkeyDo, line 654: Here you override the onSubmit of the form with a "submit" button with an onclick event pointing to this.submitReviewToCB(…).
With my suggestions above regarding that, I think you should be able to use
```suggestion
          <form className="modal-content" onSubmit={this.submitReviewToCB}>
```
and on your submit button remove the onClick handler.
The `type="submit"` you have on the button should trigger the method passed as `onSubmit` to the <form> element

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+                type="button"
+                className="close"
+                data-dismiss="modal"
+                aria-label="Close"
+              >
+                <span aria-hidden="true">&times;</span>
+              </button>
+              <h4
+                className="modal-title"
+                id="PinRecordingModalLabel"
```
MonkeyDo, line 666: This looks incorrect :)

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -363,6 +364,12 @@ const getListenablePin = (pinnedRecording: PinnedRecording): Listen => {
   return pinnedRecListen;
 };
 
+const countWords = (str: string): number => {
```
MonkeyDo, line 367: That's elegant!
If it comes from the internet, do put a link to the source as a comment

listenbrainz/webserver/static/css/theme/boostrap/modals.less
```diff
@@ -97,6 +97,55 @@
 .modal-body {
   position: relative;
   padding: @modal-inner-padding;
+
```
MonkeyDo, line 100: All this new CSS code should live in a new separate file in the root of the css directory, rather than modifying this file in theme/bootstrap.

listenbrainz/webserver/static/css/theme/boostrap/modals.less
```diff
@@ -97,6 +97,55 @@
...
+      margin-left: 6px;
+    }
+  }
+
+  .dropdown-menu {
+    min-width: 140px;
+    border-radius: 4px;
+    padding: 10px 0px;
+    width: 250px;
+    color: @hate;
```
MonkeyDo, line 130: That doesn't look right, that `hate` variable is used for ratings but we shouldn't rely on it always being there.
But I do see the definition of @hate is just `@hate: @blue` . @blue is defined in static/css/colors.less and those aren't going anywhere, so you can use @blue directly.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
MonkeyDo:
```suggestion
          Words: {countWords(textContent)} / Characters: {textContent.length}
```

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+
+    /* Success message */
+    if (success && entityToReview) {
+      return (
+        <div>
+          Thanks for submitting your review for <b>{entityToReview.name}</b>!
+          <br />
+          <br />
+          You can access your CritiqueBrainz review by clicking{" "}
+          <a href={`${this.CBBaseUrl}/${reviewMBID}`}> here.</a>
```
MonkeyDo, line 447: That's an important missing word :p
```suggestion
          <a href={`${this.CBBaseUrl}/review/${reviewMBID}`}> here.</a>
```

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+      );
+    }
+
+    /* None of the three entities were found for the Listen */
+    if (!entityToReview) {
+      return (
+        <div>
+          We could not find any reviewable entities associated with{" "}
+          <b>{listen.track_metadata?.track_name}</b> by{" "}
+          <b>{listen.track_metadata?.artist_name}.</b> :(
```
MonkeyDo, line 458: I see this as a good opportunity to add missing data to MusicBrainz, if we couldn't find anything.

```suggestion
          We could not link <b>{listen.track_metadata?.track_name}</b> by{" "}
          <b>{listen.track_metadata?.artist_name}.</b> to any recording or artist on
          MusicBrainz. If you can't find them when searching <a href="https://musicbrainz.org/search">on MusicBrainz</a> either,
          please consider <a href="https://musicbrainz.org/doc/Introduction_to_Editing">adding them to our database</a>.
```

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+      loading: false,
+      reviewValidateAlert: null,
+
+      releaseGroupEntity: null,
+      artistEntity: null,
+      recordingEntity: null,
+
+      textContent: "",
+      rating: 0,
+      language: "en",
```
MonkeyDo, line 94: An open question to everybody: do we want to default to English?
There might be ways to [detect the user's preferred language](https://caniuse.com/mdn-api_navigatorlanguage_languages) from the navigator, and we could always fall back to english for non-supported browsers.

MonkeyDo (changes requested):
Overall this looks really good, well done !

I have a lot of small nitpicks and code styling suggestions, and a few more consequential improvements or questions.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
jdaok: Oops.. Fixed now

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
jdaok: Done, form changed to read "Rating (optional):"

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+          <a href={`${window.location.origin}/profile/music-services/details/`}>
+            {" "}
+            music services page.
+          </a>
+        </div>
+      );
+    }
+
+    /* Success message */
+    if (success && entityToReview) {
```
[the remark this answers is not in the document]
jdaok: Those cases don't exist, but `entityToReview.name` is included in the success message so the && is needed to make typescript happy (entityToReview can be null)

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
jdaok: The first suggestion is genius !

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+      rating,
+      language,
+      acceptLicense,
+    } = this.state;
+
+    /* Show warning if review text doesn't meet minnimum length */
+    if (textContent.length < this.minTextLength) {
+      this.setState({
+        reviewValidateAlert: `Text length needs to be between ${this.minTextLength} and ${this.maxTextLength} characters.`,
+      });
```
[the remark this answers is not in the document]
jdaok: Will go with the second option since the text area already limits the amount of characters that can be typed.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+    track_name: string
+  ): Promise<string> => {
+    const { APIService } = this.context;
+
+    try {
+      const response = await APIService.lookupMBReleaseFromTrack(mbid);
+      // MusicBrainz API returns multiple releases, medias, and tracks, so we need to
+      // search for the track with a name that matches the supplied track_name
+      const medias = response.releases[0].media;
+      const media = medias.find((res: any) => res.tracks);
```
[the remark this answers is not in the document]
jdaok: Thanks, I updated the comments.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
jdaok: Agreed.. I was wondering if we could move it to a popup window or it's own page instead, but this is dangerous since people might miss it entirely.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+    }
+    newAlert(
+      "danger",
+      title || "Error",
+      typeof error === "object" ? error.message : error
+    );
+  };
+
+  hasPermissions = (user?: CritiqueBrainzUser) => {
+    if (!user || !user.access_token) {
```
akshaaatt, line 132: This can be simplified.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+            reviewValidateAlert: null,
+            success: true,
+            reviewMBID: response.id,
+          });
+        }
+      } catch (error) {
+        if (maxRetries > 0 && error.message === "invalid_token") {
+          /* Need to refresh token and retry with new token */
+          const newToken = await this.refreshCritiquebrainzToken();
+          this.submitReviewToCB(newToken, maxRetries - 1);
```
akshaaatt, line 393: We are missing an `await` over here.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+            <input
+              id="acceptLicense"
+              type="checkbox"
+              checked={acceptLicense}
+              name="acceptLicense"
+              onChange={this.handleInputChange}
+              required
+            />
+            <small>
+              &nbsp;You acknowledge and agree that your contributed reviews to
```
akshaaatt, line 574: This does not provide satisfactory UX. This needs to be fixed. If not right now, it should be dealt with later.

akshaaatt (commented):
Great work! Although I see that the code isn't perfected yet, I have suggested a few fixes.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
MonkeyDo: Do you mean this?
```suggestion
    if (!user?.access_token) {
```

akshaaatt: Yes @MonkeyDo ! :)
Infact, we can make this a one liner as `return !user?.access_token;`

listenbrainz/webserver/templates/user/profile.html
```diff
@@ -12,4 +12,10 @@
...
+    // Enable popovers after page load
+    window.addEventListener("load", function(){
+      $('[data-toggle="popover"]').popover();
+    });
```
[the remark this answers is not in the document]
jdaok: Ok, I understand (apparently React shouldn't be used together with jQuery anyway). I tried out the library you pointed to but the styling it came with wasn't working with our pages, I ended up using the [react-tooltip library](https://www.npmjs.com/package/react-tooltip) instead.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
```diff
@@ -0,0 +1,703 @@
...
+      const { critiquebrainzAuth } = this.context;
+
+      return (
+        <button
+          type="submit"
+          id="submitReviewButton"
+          className={`btn btn-success ${reviewValidateAlert ? "disabled" : ""}`}
+          onClick={() =>
+            this.submitReviewToCB(critiquebrainzAuth?.access_token)
+          }
```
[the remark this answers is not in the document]
jdaok: Thanks, this makes sense to me, I'll remember to try to avoid arrow functions in render().

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
MonkeyDo: I guess we want to invert that boolean: `return !!user?.access_token;`

MonkeyDo:
I've had another quick in-passing thought: if a listen has no recording_mbid, artist_mbid, track_mbid or release_group_mbid (in short nothing to review), we should gray out the "Review this Recording" button in the dropdown menu.

Saves the user a click and the disappointment of the "We could not find any reviewable entities" message if there's no chance  it would work anyway.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
akshaaatt: Yasss @MonkeyDo you're on point. A correction regarding that! Thanks.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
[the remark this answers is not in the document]
jdaok: I meant to return a promise, and I updated the function so loading is only set to false on the last error, or on success.

jdaok:
While trying to commit my changes tonight I had discarded them (github desktop usually moves them to the recycle bin temporarily) but this time I cannot find more than half of them.. I'll get this PR ready again within a day or two.

jdaok:
> I've had another quick in-passing thought: if a listen has no recording_mbid, artist_mbid, track_mbid or release_group_mbid (in short nothing to review), we should gray out the "Review this Recording" button in the dropdown menu.

Done, these changes are added to #1586 now (it required modifying the listenControl and listenCard components).

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
jdaok: I think it's a good idea, it will save non english users an extra click every time they want to review. 
I will get this implemented in a separate PR once this one is merged so this one doesn't get too bloated.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
jdaok: I agree that the checkbox isn't ideal but I'm wondering what would be the best way to get around this. My understanding was that details of terms of service stuff should ideally be viewable upfront. Maybe it would be okay to move this into a popover, or give the text it's own page on CritiqueBrainz and just link it from the modal. @MonkeyDo what did you think?

MonkeyDo: It's not a question I'm comfortable answering, to be honest. Maybe @alastair can give us his point of view?
(alastair, we're talking about the checkbox with the super long license text in the modal, see the video in the description)
Does the license text need to be in full as-is, or could it be a link to a page with the full text, or a popover, or something else?

However I don't think this preoccupation should block this PR.

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
MonkeyDo: I wonder if there would be a way to use the user's browser language, and defaulting to english if it's not accessible.
Perhaps something with `const locale = navigator.language || navigator.userLanguage || "en";`, although I haven't checked how well the language codes match.
For example for my browser I get `fr-FR`; is that accepted, or should it be `fr` only?

Anyway, something for another PR !

listenbrainz/webserver/static/js/src/CBReviewModal.test.tsx
```diff
@@ -257,7 +251,7 @@ describe("submitReviewToCB", () => {
     const spy = jest.fn();
     instance.context.APIService.submitReviewToCB = spy;
 
-    await instance.submitReviewToCB(""); // access token not set
+    await instance.submitReviewToCB(); // access token not set
```
MonkeyDo, line 254: How does this test work?
I can't see anywhere where you're un-setting the access token, and I assume the globalProps have a CB token set by default, right?

MonkeyDo: I think this is currently the same test as the one below where `"does nothing if license was not accepted"`, looking at line 248

listenbrainz/webserver/static/js/src/CBReviewModal.test.tsx
```diff
@@ -297,7 +291,7 @@ describe("submitReviewToCB", () => {
     const spy = jest.fn();
     instance.context.APIService.submitReviewToCB = spy;
 
-    await instance.submitReviewToCB("valid token");
+    await instance.submitReviewToCB();
```
MonkeyDo, line 294: Similarly here on line 288 shouldn't `acceptLicense` be set to `true`?

listenbrainz/webserver/static/js/src/CBReviewModal.test.tsx
jdaok: Thanks for catching, I missed this while rewriting the submit function. Logic in tests should be fixed now :)

listenbrainz/webserver/templates/user/profile.html
```diff
@@ -12,4 +12,10 @@
 {% block scripts %}
   {{ super() }}
   <script src="{{ get_static_path('recentListens.js') }}" type="text/javascript"></script>
+  <script>
+    // Enable popovers after page load
+    window.addEventListener("load", function(){
```
MonkeyDo, line 17: Now that you've replace the popover functionality with a library, this should be removed

MonkeyDo (changes requested):
Last modification request, and after that it's a green light for me ! :rocket:

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
jdaok: @MonkeyDo Yeah you're right, it should be the two letter iso-639-1 code `fr` only. There's no direct way to get it from the browser it looks like but we can use something like `navigator.language.substring(0,2)`.

jdaok:
Thanks all for the help on this one, it's much appreciated 😅

MonkeyDo (approved):

listenbrainz/webserver/static/js/src/CBReviewModal.tsx
MonkeyDo: Or maybe split on characters `-` (and maybe `_` I'd need to review language codes), since I assume some languages will have more than 2 characters

PR #1592, opened 2021-08-24
Enable pinned recording feature for all

amCap1712:
The pinned recordings feature is finally complete. Cheers @jdaok and thanks for all the work 🎉. This PR will enable the feature for all. Will merge once @jdaok approves and confirms that there are no other feature flags/temporary things missed in the PR.


amCap1712:
I have also deployed this on beta.lb so that it can be tested.

jdaok:
Cheers! :D LGTM, we just need just need to update the snapshots for RecentListens and ListenCard (I don't have access to push commits here)

PR #1594, opened 2021-08-27
Add missing await's in APIService and minor refactoring

amCap1712:
Some of the functions in APIService.ts were missing `await`s, added those and did minor refactoring.


MonkeyDo (approved):
:+1:

PR #1599, opened 2021-08-27
Fix the title rendered in PinnedRecordingCard

jdaok:
This commit fixes the text rendered in the PinnedRecordingCard component. 

When the current user visits their own profile, **"My"** should be rendered in the title for pin cards. When they visit another user's profile, **the other user's name** should be rendered instead.


amCap1712 (approved):
Whoops! Thanks for fixing!

PR #1600, opened 2021-08-28
LB-772: Add playlists to admin interface

amCap1712:
LB-772

### DIFF
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -70,6 +70,11 @@ MBID_MAPPING_DATABASE_URI = "dbname=musicbrainz_json_dump user=musicbrainz host=
 MBID_MAPPING_DATABASE_URI = ""
 {{end}}
 
+# for use in playlists admin view
+SQLALCHEMY_BINDS = {
+   'timescale': SQLALCHEMY_TIMESCALE_URI
+}
+
 {{if service "typesense-listenbrainz"}}
 {{with index (service "typesense-listenbrainz") 0}}
 TYPESENSE_HOST = "{{.Address}}"
--- a/listenbrainz/model/playlist.py
+++ b/listenbrainz/model/playlist.py
@@ -1,0 +1,49 @@
+import uuid
+from datetime import datetime
+
+from sqlalchemy.dialects.postgresql import UUID, JSONB
+
+from listenbrainz.model import db
+from listenbrainz.webserver.admin import AdminModelView
+
+
+class Playlist(db.Model):
+    __bind_key__ = 'timescale'
+    __tablename__ = "playlist"
+    __table_args__ = {"schema": "playlist"}
+
+    id = db.Column(db.Integer, autoincrement=True, primary_key=True)
+    mbid = db.Column(UUID(as_uuid=True), nullable=False, default=uuid.uuid4)
+    creator_id = db.Column(db.Integer, nullable=False)
+    name = db.Column(db.String, nullable=False)
+    description = db.Column(db.String)
+    public = db.Column(db.Boolean, nullable=False)
+    created = db.Column(db.DateTime(timezone=True), default=datetime.utcnow, nullable=False)
+    last_updated = db.Column(db.DateTime(timezone=True), default=datetime.utcnow, nullable=False)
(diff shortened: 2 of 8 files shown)


consul_config.py.ctmpl
```diff
@@ -70,6 +70,11 @@ MBID_MAPPING_DATABASE_URI = "dbname=musicbrainz_json_dump user=musicbrainz host=
 MBID_MAPPING_DATABASE_URI = ""
 {{end}}
 
+# for use in playlists admin view
+SQLALCHEMY_BINDS = {
+   'timescale': SQLALCHEMY_TIMESCALE_URI
```
alastair, line 75: does this mean that the sqlalchemy timescale url is defined here, but the regular db url is defined in `SQLALCHEMY_DATABASE_URI`? Does it make sense to combine them? I guess that means that we'd have to set `__bind_key__` in all other db models as well?

listenbrainz/model/playlist.py
```diff
@@ -0,0 +1,49 @@
...
+
+from sqlalchemy.dialects.postgresql import UUID, JSONB
+
+from listenbrainz.model import db
+from listenbrainz.webserver.admin import AdminModelView
+
+
+class Playlist(db.Model):
+    __bind_key__ = 'timescale'
+    __tablename__ = "playlist"
```
alastair, line 12: quote consistency

listenbrainz/model/playlist.py
```diff
@@ -0,0 +1,49 @@
...
+    __table_args__ = {"schema": "playlist"}
+
+    id = db.Column(db.Integer, autoincrement=True, primary_key=True)
```
alastair, line 15: just out of interest - what is the purpose of `autoincrement=True, primary_key=True` in a model definition? I assume that this is purely so that if you use sqlalchemy to create the table, that it correctly creates necessary indexes/sequences/etc. Given that we make our tables manually, what do we think about having them defined here as well. Does it make sense?

listenbrainz/model/playlist.py
```diff
@@ -0,0 +1,49 @@
...
+    mbid = db.Column(UUID(as_uuid=True), nullable=False, default=uuid.uuid4)
```
alastair, line 16: similar to the comment above - in sqlalchemy does this `default` happen in python or in postgres? Given that we have a database-level default value, I'm not sure if we should _also_ have a python-level default here, which could cause confusion. What do you think?

alastair (commented):
:+1:

consul_config.py.ctmpl
amCap1712: https://flask-sqlalchemy.palletsprojects.com/en/2.x/binds/

The `SQLALCHEMY_DATABASE_URI` one is the default to be used if no other bind is specifically defined. The docs also define one default and then extra ones are in `SQLALCHEMY_BINDS`. I think its fine to keep it this way.

listenbrainz/model/playlist.py
amCap1712: Yeah, its unlikely we'll need those but I added them here anyways for consistency with the SQL table definition. Flask-admin also does some UI customization based on whether a column is primary_key or not but that's simple and one liner to implement if we wish to keep it.

listenbrainz/model/playlist.py
amCap1712: Right, this is a python default. This will only be useful if we want to add a playlist using admin UI. Should I remove it or make it a server side default?

alastair: my intuition would be to disable adding in the admin interface, and remove the python default values.

listenbrainz/model/playlist.py
alastair: I think that this is something that we should come up with a development guideline for. It's nice that flask-admin takes these hints for additional functionality, so I guess it makes sense to keep it here for now.

listenbrainz/model/playlist.py
amCap1712: Done.

alastair (approved):
This looks good to me

PR #1601, opened 2021-08-31
Gracefully handle errors due to invalid oauth codes

alastair:
# Problem

Identified in https://sentry.metabrainz.org/metabrainz/listenbrainz/issues/145812/?environment=production&referrer=alert_email

Although it doesn't happen in a normal successful oauth flow, there are some cases where a manually constructed URL with an invalid `code` parameter could cause an uncaught exception. 

# Solution

Handle the situation where:
 * the auth code isn't valid on the musicbrainz end, therefore returning an error message
 * the response from musicbrainz isn't valid json

# Action

I added another exception to handle the case where the user has no email address, but I didn't test this in detail. The behaviour should be the same. It might be worth checking it again with a test user to verify that this works as expected.

### DIFF
--- a/listenbrainz/webserver/login/provider.py
+++ b/listenbrainz/webserver/login/provider.py
@@ -13,6 +13,16 @@ _musicbrainz = None
 _session_key = None
 
 
+class MusicBrainzAuthSessionError(Exception):
+    """Raised when there is an error parsing the oauth response from MusicBrainz"""
+    pass
+
+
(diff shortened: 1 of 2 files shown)


amCap1712 (approved):
LGTM. Will test on beta.

amCap1712:
I had tested on beta. Login and email warnings/errors worked as expected.

PR #1603, opened 2021-09-02
Monkey refactor date year

MonkeyDo:
Context: https://chatlogs.metabrainz.org/libera/metabrainz/2021-09-02/?msg=4854802&page=1

Previously the year was shown on the listen card date only if the listen was from > 1 year ago.
It was confusing to not show the year for example now in September 2021 for a listen from November 2020; it would only show November Xth but no year.

Instead, we all agree it is more convenient (and what we all expect) to show the listen is not from _this year_

### DIFF
--- a/listenbrainz/webserver/static/js/src/utils.tsx
+++ b/listenbrainz/webserver/static/js/src/utils.tsx
@@ -243,7 +243,6 @@ const preciseTimestamp = (
     const now = new Date();
     const currentYear = now.getUTCFullYear();
     const listenYear = listenDate.getUTCFullYear();
-    const msDifference = now.getTime() - listenDate.getTime();
     // Date is today : format using timeago
     if (
       now.getUTCDate() === listenDate.getUTCDate() &&
--- a/listenbrainz/webserver/static/js/src/utils.tsx
+++ b/listenbrainz/webserver/static/js/src/utils.tsx
@@ -240,17 +240,26 @@ const preciseTimestamp = (
 
   // determine which display setting based on time difference to use if no argument was provided
   if (!display) {
-    const msDifference = new Date().getTime() - listenDate.getTime();
-    // over one year old : show with year
-    if (msDifference / (1000 * 3600 * 24 * 365) > 1) {
-      display = "includeYear";
+    const now = new Date();
+    const currentYear = now.getUTCFullYear();
+    const listenYear = listenDate.getUTCFullYear();
+    const msDifference = now.getTime() - listenDate.getTime();
+    // Date is today : format using timeago
+    if (
+      now.getUTCDate() === listenDate.getUTCDate() &&
+      now.getUTCMonth() === listenDate.getUTCMonth() &&
+      currentYear === listenYear
+    ) {
+      display = "timeAgo";
     }
-    // one year to yesterday : show without year
-    else if (msDifference / (1000 * 3600 * 24 * 1) > 1) {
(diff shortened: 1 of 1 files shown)


amCap1712 (approved):
Thanks! :rocket:

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -240,17 +240,25 @@ const preciseTimestamp = (
...
+    const now = new Date();
+    const currentYear = now.getUTCFullYear();
+    const listenYear = listenDate.getUTCFullYear();
+    // Date is today : format using timeago
+    if (
+      now.getUTCDate() === listenDate.getUTCDate() &&
+      now.getUTCMonth() === listenDate.getUTCMonth() &&
+      currentYear === listenYear
+    ) {
+      display = "timeAgo";
```
alastair, line 252: Because you're comparing with UTC dates here, the switchover from timeAgo to excludeYear may happen at midday for someone in NZ (right?). I think it might be better if we converted in the browser's timezone...

MonkeyDo:
@amCap1712 and @alastair Following the extra discussion in IRC, I also created a new helper to get a better full human-readable date, currently used on hovering on the Listen card date time.

It will produce a localized human-readable date instead of those ugly ISO dates.
For standard English locale it would be this format: `Friday, September 3, 2021 at 6:04:34 PM GMT+2`
For my locale it would be `vendredi 3 septembre 2021 à 18:03:16 UTC+2`
If you want to try it out, paste this in your browser console: `new Date().toLocaleString(undefined,{ dateStyle: 'full', timeStyle: 'long' })`

I guess that's a step in the right direction, although it would be useful to review date usage across the website to harmonize a bit (I guess in a following PR).

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -264,8 +265,13 @@ export default class ListenCard extends React.Component<
                       <span
                         className="listen-time text-muted"
                         title={
-                          listen.listened_at_iso?.toString() ||
-                          new Date(listen.listened_at * 1000).toISOString()
+                          listen.listened_at
```
alastair, line 268: there are 2 `<MediaQuery>` elements in this component, so it should be updated in both

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -264,8 +265,13 @@ export default class ListenCard extends React.Component<
...
+                            ? fullLocalizedDateFromTimestampOrISODate(
+                                listen.listened_at * 1000
+                              )
+                            : fullLocalizedDateFromTimestampOrISODate(
+                                listen.listened_at_iso
+                              )
                         }
                       >
                         {preciseTimestamp(
```
alastair, line 277: unrelated to this PR, but I noticed that the `...` dropdown on a listen card still has a title of "delete" (probably from when it was a delete button). Could this be updated? "more"? "tools" ?

alastair (approved):
After updating the other MediaQuery this looks fine to me. Happy for it to be merged after that's fixed

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
MonkeyDo: I went ahead and corrected that in this PR, might as well.

PR #1605, opened 2021-09-05
CritiqueBrainz review events to user timeline

jdaok:
(this PR doesn't depend on any others)
This PR contains the backend changes to add CritiqueBrainz review events to the user timeline.

~~A new endpoint is added to allow the CBReviewModal to submit review events. Review events include the review data (rating, text etc) and also the listen metadata. This allows for the recording linked to the review to be listenable on BrainzPlayer when it appears on the user feed.~~


admin/sql/updates/2021-08-28-add-user-timeline-enum-type-critiquebrainz_review.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
```
amCap1712, line 1: You can remove the BEGIN/COMMIT here, Postgres is weird about ALTER TYPE ADD VALUE running inside a transaction.

amCap1712:
@jdaok Having discussed this with @mayhem, we should not be storing the review data in two places. The right place for this data is in CB's database. LB can access it using CB api/directly accessing DB whichever makes more sense and then cache it if needed. This can however cause other reviews than "mini-reviews" to be pulled in. We think that is fine and can just format it on frontend side with `...` as necessary.

jdaok:
@amCap1712 The CB API has an endpoint for fetching reviews by user, but it requires the user's `cb uuid`. 
To use the API/db we may have to add an optional `musicbrainz username` to this endpoint (it looks like the sql query can search for the matching name in the `musicbrainz_id` column.

But do you think it would be appropriate to store _just_ the `review uuid` with the listen in LB's database?
We could use that id to fetch the review data with the API instead, and it would also keep only the 'mini-reviews'.
The advantage would be that this way, we could get the events listenable on brainzplayer, as the CB api doesn't return sufficient entity metadata needed for this.

Both would lead to many API calls on the user feed though, so maybe the database option? (I wouldn't know how to set this connection up though). Which option did you think would be the best to go with? :o

amCap1712:
Following the discussion in IRC today, we settled on storing only the review UUID in the timeline table.

amCap1712:
I have updated the PR to show reviews on the feed. Also, the reviews are playable now.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -432,6 +479,64 @@ def get_recording_recommendation_events(users_for_events: List[dict], min_ts: in
...
+        review_id = event.metadata.review_id
+        review_ids.append(review_id)
+        review_id_event_map[review_id] = event
+        entity_ids.append(event.metadata.entity_id)
+
+    reviews = CritiqueBrainzService().fetch_reviews(review_ids)
+    if reviews is None:
+        return []
+
+    mbid_metadatas = fetch_mapped_recording_data(entity_ids)
```
amCap1712, line 504: This works for recording reviews. But won't work for artist reviews or release group reviews. One way I think other reviews could be made playable is to always store recording mbid in the metadata even if it is a artist or release group review and play that. Thoughts?

amCap1712: Also, we don't store any other metadata like artist, track name etc for reviews in LB currently and CB doesn't store that either. It also only stores the mbid and retrieves the data from MB as needed.

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -344,12 +352,8 @@ export default class UserFeedPage extends React.Component<
             )}
             showUsername={false}
             showTimestamp={false}
-            listen={metadata as Listen}
-            additionalContent={
-              (metadata as PinEventMetadata).blurb_content
-                ? `"${(metadata as PinEventMetadata).blurb_content}"`
-                : ""
-            }
+            listen={listen}
```
amCap1712, line 355: @MonkeyDo The ListenCard uses the track_name and artist_name from the listen. However, if we have written a review for an artist, its likely we want to display the artist name first. (there's also some support for writing release group reviews so probably need to support that as well). Thoughts on how to do this? 

The entity_name and entity_type fields are the relevant ones here and should be used to display the name and link to the correct MB entity. For making the card playable, we can either just search the entity name and play it like in stats or use the attached recording_mbid mapped data to search for the original listen using which the review was written. Note that in the current implementation, I have attached recording mapped data in all cases with the listen in all cases.

To reiterate, the recording reviews are handled fine by the listen cards. The issue is how we support the display for other entity's reviews.

listenbrainz/db/model/review.py
```diff
@@ -0,0 +1,46 @@
...
+class CBReviewTimelineMetadata(BaseModel):
+    """Model to represent the data stored in user timeline event table
+    for a CB review. We only store review uuid, the entity's mbid and
+    the recording mbid (of the listen using which the review was written).
+    For the review of a recording, the entity_id and the recording_id is
+    the same. Other data is retrieved from the CB api as needed.
+    """
+    entity_name: str
+    review_id: str
+    entity_id: str
```
amCap1712, line 31: It is possible to get away with a subset of these fields here but I decided to keep all for simpler implementation. If we needed to support recording reviews only, we only need to store the recording mbid  and review id  here.  The entity name is obtained from the mapping tables when needed. But because we support artist and release group reviews, we need to store the name field because we don't have those in the mapping tables. 

Now for making these artists and release group reviews playable we could just search for the names. Alternative is we attach a recording mbid with each review and then retrieve the recording data from mapped tables and play that listen for the review. Regardless of the approach we go with I think its a good idea to store a recording mbid with each review. The reason is eveyrthing else in LB has a msid /mbid attached to it. Therefore, IMO It makes sense for reviews to be that way as well.

Now for recording reviews, the entity_id and entity_name become redundant because the entity_id is same as recording_id and entity_name can be retrieved from mapping tables. So we can set these fields to null if we want for recording reviews. I decided in favor of filling the redundant fields anyways because feels simpler to me.

listenbrainz/domain/critiquebrainz.py
```diff
@@ -1,21 +1,29 @@
...
 from listenbrainz.domain.external_service import ExternalService, ExternalServiceInvalidGrantError
 
+
 CRITIQUEBRAINZ_SCOPES = ["review"]
 
 OAUTH_AUTHORIZE_URL = "https://critiquebrainz.org/oauth/authorize"
 OAUTH_TOKEN_URL = "https://critiquebrainz.org/ws/1/oauth/token"
 
+CRITIQUEBRAINZ_REVIEW_SUBMIT_URL = "https://critiquebrainz.org/ws/1/review/"
+CRITIQUEBRAINZ_REVIEW_FETCH_URL = "https://beta.critiquebrainz.org/ws/1/reviews/"
```
amCap1712, line 24: This PR depends on https://github.com/metabrainz/critiquebrainz/pull/383 . A reminder to me to change this to cb prod once that PR is merged and deployed.

amCap1712:
Tests pending, will add once implementation is finalized.

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
MonkeyDo: I think I'm in favor of option # 1 and just using the entity name to search and play, like we do for the reports/charts pages ([example here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/stats/UserTopEntity.tsx#L203-L206)), but mainly for consistency.
I feel that if the user chose to review the artist rather than a specific recording we shouldn't tie a recording to their review on their behalf.

mayhem: Yes, given that I was arguing for consistency, I'm also good with #1

listenbrainz/db/model/review.py
amCap1712: I removed recording_id after discussion on IRC. We don't have a use case for this and will need to add more stuff to handle it.

amCap1712:
Thanks all!

PR #1606, opened 2021-09-06
added delete event functions and rudimentary API

hrik2001:
# Problem

This is a proposed solution to [LB-912](https://tickets.metabrainz.org/browse/LB-912). This is an endpoint that lets the creator of events to be deleted from the database. I was told that deletion of recommendation of tracks event is of utmost importance, hence I implemented it first. Once approved, and once I get the hang of the style of code you guys want, I will implement the deletion feature for other event types.

# Solution

My solution was to create a path `user/<user_name>/feed/events/delete` because it went well with the convention of other routes (I also got this route approved by lucifer in IRC). There is a function called `delete_user_timeline_event` that would actually delete any event you give it. But for now I have implemented an if condition for the deletion of recommendation event only, since events like unpin already have their dedicated path, later on it would be extended to if elif clause for other events. I thought that an if...elif clause would be safer, rather than making it completely up to the user to put the `event_type` and `metadata` and feed it directly to the database for deletion using that function. So it would check if the user's queried `event_type` is supported or not, if yes, then it will query further using the `metadata` and then proceed to delete it. If my description is not clear, then apologies, please ping me in IRC. Please do give feedback! :D 

# Action

I had mentioned this potential [flaw](https://github.com/metabrainz/listenbrainz-server/blob/b79dfb7194869939fe4a94c26cff8e8598b36bc8/listenbrainz/webserver/views/user_timeline_event_api.py#L256) in code once, I guess it got accidentally overlooked. Just a minor thing I found during my initial introspection of the codebase. `db_conn` gets passed whenever its called, and that too a fresh instance, yet the function disregards the passed argument and creates its own `db_conn`

### DIFF
--- a/listenbrainz/db/user_timeline_event.py
+++ b/listenbrainz/db/user_timeline_event.py
@@ -69,6 +69,33 @@ def create_user_track_recommendation_event(user_id: int, metadata: RecordingReco
         metadata=metadata,
     )
 
+def delete_user_timeline_event(user_id: int, metadata: RecordingRecommendationMetadata) -> UserTimelineEvent:
+    """ Deletes event from database and returns it.
+        Returns None if it couldn't find the queried event.
+    """
+    try:
+        meta = []
+        for i in metadata.dict().items():
+            if i[1] != None:
+                meta.append(" metadata ->> '" + i[0] + "' = '" + i[1] + "'")
+        meta_query = " AND ".join(meta)
+        with db.engine.connect() as connection:
+            result = connection.execute(sqlalchemy.text("""
+                DELETE FROM user_timeline_event WHERE
+            """ + meta_query + """
+                RETURNING *
+            """))
+            try:
+                r = dict(result.fetchone())
+            #when there is nothing to delete
+            except TypeError:
+                r = None
+        if not r:
+            return None
+        return UserTimelineEvent(**r)
+    except Exception as e:
+        raise DatabaseException(str(e))
+
 
 def create_user_notification_event(user_id: int, metadata: NotificationMetadata) -> UserTimelineEvent:
     """ Create a notification event in the database and returns it.
--- a/listenbrainz/webserver/views/user_timeline_event_api.py
+++ b/listenbrainz/webserver/views/user_timeline_event_api.py
@@ -238,6 +238,53 @@ def user_feed(user_name: str):
         'events': [event.dict() for event in all_events],
     }})
 
+@user_timeline_event_api_bp.route("/user/<user_name>/feed/events/delete", methods=['OPTIONS', 'POST'])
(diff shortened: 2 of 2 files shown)


listenbrainz/db/user_timeline_event.py
```diff
@@ -69,6 +69,33 @@ def create_user_track_recommendation_event(user_id: int, metadata: RecordingReco
...
 
+def delete_user_timeline_event(user_id: int, metadata: RecordingRecommendationMetadata) -> UserTimelineEvent:
+    """ Deletes event from database and returns it.
+        Returns None if it couldn't find the queried event.
+    """
+    try:
+        meta = []
+        for i in metadata.dict().items():
+            if i[1] != None:
+                meta.append(" metadata ->> '" + i[0] + "' = '" + i[1] + "'")
```
amCap1712, line 80: Don't construct the query by hand! This is a bad practice and makes the code vulnerable to SQL injection. A better way to do this is something like. https://github.com/metabrainz/listenbrainz-server/blob/b79dfb7194869939fe4a94c26cff8e8598b36bc8/listenbrainz/db/pinned_recording.py#L83-L89

amCap1712:
Thanks for working on this @hrik2001. Passing the metadata to delete an event is probably not a good idea. We usually use a unique id with/without a user_id for this purpose. I'll look into the code and see what's the right set of parameters needed here.

hrik2001:
Thanks for the review @amCap1712 , I was wondering too that its not safe (I had mentioned while opening this PR). I am actually not that comfortable with JSONB queries using `sqlalchemy.text`, so I opted for using `->>` operator. I saw that metadata was being passed to the frontend, so I thought that could be used.   

Do you plan to change the schema of `user_timeline_event` and implement some sort of indexing? Would it cause backwards compatibility issues?

listenbrainz/db/user_timeline_event.py
hrik2001: also I wanted to ask, does sqlalchemy provide sanitisation like this for JSONB queries?

amCap1712:
After discussing with @alastair and @mayhem, we decided that it makes sense to use `id` and `user_id` for deletion. This is similar to what we do for pinned recordings:
https://github.com/metabrainz/listenbrainz-server/blob/b79dfb7194869939fe4a94c26cff8e8598b36bc8/listenbrainz/db/pinned_recording.py#L71-L92
We'll also need to modify the https://github.com/metabrainz/listenbrainz-server/blob/b79dfb7194869939fe4a94c26cff8e8598b36bc8/listenbrainz/webserver/views/user_timeline_event_api.py#L163 as a next step so that it passes these ids to frontend.

hrik2001:
Done accordingly at #1610 . Closing this PR.

PR #1607, opened 2021-09-07
LB-442, LB-321: Update listen format documentation to include source and origin_url

alastair:
# Problem

LB-442, LB-321
https://community.metabrainz.org/t/what-do-you-want-in-listenbrainz/444888/38

This has been something that has been discussed for some time now. We want to know
* What tool someone is using to submit a file (name and version)
* What service someone is playing from - for local files this won't exist, but in the case that services have multiple clients we can distinguish the two
* Where the song that they're listening to can be found online (if it's online)

# Solution

Add some new recommended fields:

* `listening_from`: Now specifically refers to the **client** being used to submit the listen. This previously said client _or_ service
* `listening_from_version`: In the case that someone wants to submit a software version. I'm still not convinced how useful this is, but others seemed to want it. I think that we should add it as a separate field so that we can easily display listening_from without parsing it, and it's backwards compatible with what we currently have.
* `source`: The name of the place where the song is being listened from
* `origin_url`: the link to the song that is being listened to

This allows us to indicate for example that someone is listening to Jamendo from Rhythmbox, or Spotify from listenbrainz web player.

# Action

I haven't gone through the list of LB clients to ensure that they're compatible with what we've suggested here. Specifically, @MonkeyDo comments that some web-scrobblers use `origin_url` as the url of the webpage that a user is on while they have a player running. I don't think that this is a good idea, as to me having https://listenbrainz.org/user/alastairp/ as an url isn't useful. Soundcloud's player will continue playing from a playlist while the user is on the page of an unrelated song.
If the solution to this is to propose a new key that has this specific meaning then I'm open to doing that as well.

### DIFF
--- a/docs/dev/json.rst
+++ b/docs/dev/json.rst
@@ -81,7 +81,9 @@ A complete submit listen JSON document may look like:
           "listened_at": 1443521965,
           "track_metadata": {
             "additional_info": {
-              "listening_from": "spotify",
+              "media_player": "Rhythmbox",
+              "submission_client": "Rhythmbox ListenBrainz Plugin",
+              "submission_client_version": "1.0",
               "release_mbid": "bf9e91ea-8029-4a04-a26a-224e00a83266",
               "artist_mbids": [
                 "db92a151-1ac2-438b-bc43-b82e149ddd50"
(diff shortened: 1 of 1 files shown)


alastair:
Tagging many people for review, because I think that this is something that we should try and get right.

cc @phw too who had opinions on this

docs/dev/json.rst
```diff
@@ -177,7 +177,10 @@ element                 description
 ``isrc``                The ISRC code associated with the recording.
 ``spotify_id``          The Spotify track URL associated with this recording.  e.g.: http://open.spotify.com/track/1rrgWMXGCGHru5bIRxGFV0
 ``tags``                A list of user defined tags to be associated with this recording. These tags are similar to last.fm tags. For example, you have apply tags such as ``punk``, ``see-live``, ``smelly``. You may submit up to :data:`~webserver.views.api.MAX_TAGS_PER_LISTEN` tags and each tag may be up to :data:`~webserver.views.api.MAX_TAG_SIZE` characters large.
-``listening_from``       The source of the listen, i.e the name of the client or service which submits the listen.
+``listening_from``      The name of the client which submits the listen.
```
amCap1712, line 180: Does a website count as a client? Not something that would be mentioned in these docs but what about the importers we run?

docs/dev/json.rst
```diff
@@ -177,7 +177,10 @@ element                 description
...
+``listening_from_version`` (optional). If the client submitting the listen has a version, put it here.
```
amCap1712, line 181: Aren't all these attirbutes optional?

MonkeyDo:
> to me having https://listenbrainz.org/user/alastairp/ as a url isn't useful. Soundcloud's player will continue playing from a playlist while the user is on the page of an unrelated song.
If the solution to this is to propose a new key that has this specific meaning then I'm open to doing that as well.

I think the solution should instead be to clarify that `origin_url` should point to an existing URL where the resource is available (presumably from the `source`), *if available*.
In the case of Soundcloud for example, using WebScrobbler that field ends up being the current browser page which as you mention is not necessarily connected to the track being listened to.
If the full URL to the track can't be retrieved then the field should be left blank.

docs/dev/json.rst
alastair: to me, a client is a piece of software that knows how to send a POST to api.listenbrainz.org/1/submit-listen. If that's a webapp that has a musicplayer _and_ submitter built in, then sure. In the case of the spotify writer I think it's close enough that we can use "spotify". Do you have a specific example of a "website" that doesn't fit these criteria?

docs/dev/json.rst
alastair: yeah. Although earlier in the doc it says that they are _all_ optional, so perhaps I should just skip this additional parenthesis here

docs/dev/json.rst
```diff
@@ -177,7 +177,10 @@ element                 description
...
+``source``              If the song of this listen comes from an online source, the name of this source (e.g. Spotify, Soundcloud, YouTube, Jamendo). This should be a name, not a URL.
+``origin_url``          If the song of this listen comes from an online source, the URL to the place where it is available. This could be a spotify url (see ``spotify_id``), a YouTube video URL, a Soundcloud recording page URL, or the full URL to a public MP3 file.
```
MonkeyDo, line 183:
```suggestion
``origin_url``          If the song of this listen comes from an online source, the URL to the place where it is available. This could be a spotify url (see ``spotify_id``), a YouTube video URL, a Soundcloud recording page URL, or the full URL to a public MP3 file. This must be a valid public URL rather than a resource ID or URI.
```
?

alastair: so if we're going to be explicit, then perhaps we should add more json examples, showing the case of youtube, or soundcloud, or jamendo, or archive.org

docs/dev/json.rst
amCap1712: I see makes sense. I don't have an example but consider the case the case of a website that hosts its own music, should both source and listening_from be sent or source be present?

Also, What do you think about adding an examples section after parameters, showing how to decide what values to use in which fields?

mayhem:
I think `listening_from` isn't very clear. How about `listening_client` and `listening_client_version`?

amCap1712:
FWIW, `listening_from` is already in use currently. We can mention it in the docs that it is obsolete but some still clients may still keep sending it :/.

docs/dev/json.rst
MonkeyDo: Here's an abridged example of what I'm going to do for BrainzPlayer in case it's useful:
```json
"track_metadata": {
  "additional_info": {
    "listening_from": "listenbrainz",
    "origin_url": "https://www.youtube.com/watch?v=Bq7-B8_NZNU",
    "source": "youtube"
  }
}
```

phw: What about having the fields named `source` and `source_url` or `origin` and `origin_url`? As I see it they are related, and source and origin kind of mean the same here.

phw:
Yes, `listening_from` is already in use. But if this should be changed maybe now instead of later, but one must be aware that older clients stay around. Here is 

The LB clients for Rythmbox, Lollypop and Funkwhale all set `listening_from` to the application name, i.e. "Rythmbox" [1], "Lollypop" [2] and "Funkwhale" [3]. The foo_listenbrainz2 plugin for foobar2000 also [4]

Rhythmbox also sets the `source` field, if music is played from another source than local files. E.g. if playing via the magnatune plugin `source`  will be set to "magnatune". But existing Rhythmbox music source plugins have bitrotted quite a bit as the sevices behind them have lost their importance.

web-scrobbler sets `origin_url` to the current URL from which the playback is happening [5].

Simpe Scrobbler for Android does not set any of those fields.

[1] https://gitlab.gnome.org/GNOME/rhythmbox/-/blob/master/plugins/listenbrainz/listenbrainz.py#L184
[2] https://gitlab.gnome.org/World/lollypop/-/blob/master/lollypop/ws_listenbrainz.py#L191
[3] https://dev.funkwhale.audio/funkwhale/funkwhale/-/blob/develop/api/funkwhale_api/contrib/listenbrainz/funkwhale_ready.py#L24
[4] https://github.com/phw/foo_listenbrainz2/blob/d2e5b2fdcc5a59b97592f962e4499ab0f9456c10/src/play.cpp#L186
[5] https://github.com/web-scrobbler/web-scrobbler/blob/master/src/core/background/scrobbler/listenbrainz-scrobbler.js#L280

alastair:
> I think listening_from isn't very clear. How about listening_client and listening_client_version?

If we want to change these values, I think now is the time to do it. I think it'd be fine to propose new, clear values (including @phw's comments about source/source_url and origin/origin_url if wanted) and then deprecate the old values. As we have a list of clients and what they do, we can easily open tickets/propose patches to update it to our preferred structure. Also, by providing new fields with clear meanings we have a clear cutoff point where we know that the old fields have one meaning and the new ones a new meaning.

docs/dev/json.rst
alastair: Can you explain what each of these 4 fields would mean (perhaps with some examples?)

phw: I think your description in this PR already fits this very well, we only need to get the naming of the fields right. And I agree that it's better to get this right now then to wait. Taking into account Rob's suggestion of using `listening_client` and using a consistent naming for origin and origin URL I would suggest something like this:

```
"track_metadata": {
  "additional_info": {
    "listening_client": "Web Scrobbler",
    "origin": "bandcamp",
    "origin_url": "https://paradiselostofficial.bandcamp.com/track/the-longest-winter"
  }
}
```

I am also not sure about the need of `listening_client_version`, but if someone wants it I see no harm of having it.

alastair:
We had an in-person discussion about this.

IMG_20210916_173226974

Our recommendations are:

* media_player: the program that is being used to listen to music (spotify, rhythmbox, otter, vlc)
* submission_client, submission_client_version: the tool being used to submit to LB (lb spotify writer, musicbrainz-android, web-scrobbler, rhythmbox lb plugin)
* music_service: the _base url_ of the place where the music comes from: jamendo.org, archive.org, spotify.com, soundcloud.com
* origin_url: the url of the thing being played (youtube page url, spotify play url, bandcamp recording url)

some comments:
- we thought about player_name, but it was a little bit vague and made us think of video games
- We thought about "resource_url", but we weren't sure if this could be confused for the url of an audio stream, so we discarded it. Even though web-scrobler uses origin_url "incorrectly", we're not too worried about this.
- We decided that in the end it was a good idea to separate the player and the submission client - this allows us to show the player on the lb site, but also use the client information for stats (what are people using for submitting data), and error correction (in the unlikely event that a client submits lots of invalid data)
- music_service should be a fully qualified URL in order to prevent questions about how to spell/present/format the name ("internet archive" vs "archive" vs "Archive")
- we don't think that the other "origin url" makes sense (the page that the user was on when they were listening to a song), because this can be really different to what is being listened to. We have some future items about "context" (playlist, recommendations) that can fill in some of this functionality

todo: fill out these lists of examples of player/submitter/source combinations:
* brainzplayer on lb.org listening to youtube
* spotify+spotify+LB player
* Rhythmbox + jamendo
* vlc + android submitter

MonkeyDo:
Here are the first three examples you mention.
I'm not sure what the fourth one means @alastair: is that VLC on mobile? What's the "Android submitter" ?

ListenBrainz player listening to Youtube:
```
{
    "track_metadata": {
        "additional_info": {
            "media_player": "BrainzPlayer",
            "music_service": "https://www.youtube.com",
            "origin_url": "https://www.youtube.com/watch?v=JKFBiaoFHoY",
            "submission_client": "BrainzPlayer"
        },
        "artist_name": "Mdou Moctar",
        "release_name": "Ilana (The Creator)",
        "track_name": "Inizgam"
    }
}
```

ListenBrainz player listening to Spotify:
```
{
    "track_metadata": {
        "additional_info": {
            "media_player": "BrainzPlayer",
            "music_service": "https://open.spotify.com",
            "origin_url": "https://open.spotify.com/track/5fEjp2F0Sqr9fMuLSaDqz0",
            "submission_client": "BrainzPlayer"
        },
        "artist_name": "Les Filles de Illighadad",
        "release_name": "Eghass Malan",
        "track_name": "Inssegh Inssegh"
    }
}
```

Rhythmbox player listening to Jamendo playlist:
```
{
    "track_metadata": {
        "additional_info": {
            "media_player": "Rhythmbox",
            "music_service": "https://www.jamendo.com",
            "origin_url": "https://www.jamendo.com/playlist/500390216/electro-funk-frenzy",
            "submission_client": "Rhythmbox ListenBrainz plugin",
            "submission_client_version": "1.2.5"
        },
        "artist_name": "Duo Teslar",
        "track_name": "Universal Funk"
    }
}
```

phw:
In general I like the recommended fields. Just one note about 

> music_service: the base url of the place where the music comes from: jamendo.org, archive.org, spotify.com, soundcloud.com

This is a nice idea, but in case of e.g. the Rhythmbox plugin this either would not be a URL or we would need to hardcode the possible URLs into the plugin. The way this currently works that it includes the name of the plugin doing the playback, so e.g. `music_service` would be set to "soundcloud" or "jamendo". Of course we could hardocode a mapping between those names and the domain names, but then again this would not work of someone adds a new plugin for a service unknown to the plugin. So if I'd submit a patch to Rhythmbox to support the newly defined fields I would do it that the submission looks like this:

```
{
    "track_metadata": {
        "additional_info": {
            "media_player": "Rhythmbox",
            "music_service": "jamendo",
            "origin_url": "https://www.jamendo.com/playlist/500390216/electro-funk-frenzy",
            "submission_client": "Rhythmbox ListenBrainz plugin",
            "submission_client_version": "1.2.5"
        },
        "artist_name": "Duo Teslar",
        "track_name": "Universal Funk"
    }
}
```

In general I'm also not sure how useful `submission_client_version` is, but probably people would still try to add this and having this a separate field makes it clear that it should not be part of the `submission_client`.

alastair:
@phw our reasoning behind using a url instead of a name is because it's more "canonical" - we can insist on a single domain name, which is easier to get correct - e.g. which of Jamendo, jamendo, or Jamendo Music are correct?

@MonkeyDo I'm not sure about the `origin_url` in the jamendo example being a playlist. In my view this should be the specific song that is being played. The playlist should be part of a `context` field (not discussed in this PR).
The 4th example I mean VLC on android plus a separate android app that submits listens from all media players (could be VLC, or any other music player). The reason I wanted to add this was to show that the media player and the submission client could be completely separate (i.e. not even just a plugin to rhythmbox). Maybe we could come up with a better idea.

phw:
> our reasoning behind using a url instead of a name is because it's more "canonical" - we can insist on a single domain name, which is easier to get correct - e.g. which of Jamendo, jamendo, or Jamendo Music are correct?

@alastair I understand the reasoning. The question is what would I do if I don't have a URL available? Not submit this data? Set "music_service" to just "jamendo" or "soundcloud"?

alastair:
I updated this with the help of @MonkeyDo, adding in much more explicit documentation and more examples.

If you want to see the whole page, I uploaded it here as text - rename to .html and open in a browser (no formatting, but the content is all there).

[json.txt](https://github.com/metabrainz/listenbrainz-server/files/7476809/json.txt)

alastair:
@Freso your other PR looks mostly fine - note that we added some more docs today including `music_service_name` - I don't know if you might have this information in the client to be able to add. We also added media_player_version to the docs.

The domain parser looks a little fragile, but I'm not sure what a generic solution might be. I guess kodi could play anything from any URL. Do you know if you're going to get a "webpage url" or a "mp3 stream url"? It might not be as flexible, but perhaps you would have cleaner data if you had a list of valid domains instead of trying to guess it.

amCap1712 (approved):

PR #1610, opened 2021-09-09
LB-912 (I): Add endpoints and UI to delete feed events

hrik2001:
# Problem
[LB-912](https://tickets.metabrainz.org/browse/LB-912) wanted an API endpoint implemented that would delete different events from the user's feed. The problem in initial attempt was that the frontend client never had any unique identification to represent each recommendation or notification event so that it could simply post that to get those events deleted.

# Solution
Hence I created new data model to represent Notification and Recommendation events. Its a superset of `APITimelineEvent` with just one `id: int` field extra added.

# Action
I implemented a new data model for Notifications and Recommendation events. But one may wish to remove that and add `id: int` field to `APITimelineEvent` itself. Though I didn't do that because it might not be safe serving primary keys of different events, and I was told in IRC by lucifer to only enable `id` for only these 2 respectively.

### DIFF
--- a/data/model/user_timeline_event.py
+++ b/data/model/user_timeline_event.py
@@ -77,15 +77,7 @@ APIEventMetadata = Union[APIListen, APIFollowEvent, APINotificationEvent, APIPin
 
 
 class APITimelineEvent(pydantic.BaseModel):
-    event_type: UserTimelineEventType
-    user_name: str
-    created: int
-    metadata: APIEventMetadata
-
-class APIRecommendationNotificationTimelineEvent(pydantic.BaseModel):
-    # since `id` is not sent in APITimelineEvent
-    # this field is needed for deletion of Recommendation and Notification Events
-    id: int
+    id: Optional[int]
     event_type: UserTimelineEventType
     user_name: str
     created: int
--- a/data/model/user_timeline_event.py
+++ b/data/model/user_timeline_event.py
@@ -81,3 +81,12 @@ class APITimelineEvent(pydantic.BaseModel):
     user_name: str
     created: int
     metadata: APIEventMetadata
+
+class APIRecommendationNotificationTimelineEvent(pydantic.BaseModel):
+    # since `id` is not sent in APITimelineEvent
+    # this field is needed for deletion of Recommendation and Notification Events
+    id: int
+    event_type: UserTimelineEventType
+    user_name: str
+    created: int
+    metadata: APIEventMetadata
--- a/listenbrainz/db/user_timeline_event.py
+++ b/listenbrainz/db/user_timeline_event.py
@@ -79,6 +79,31 @@ def create_user_notification_event(user_id: int, metadata: NotificationMetadata)
         metadata=metadata
     )
 
+def delete_user_recommendation_notification_event(
+        id: int,
+        user_id:int
+) -> UserTimelineEvent:
+    ''' Deletes recommendation and notification event using id'''
+    try:
+        with db.engine.connect() as connection:
+            result = connection.execute(sqlalchemy.text('''
+                    DELETE FROM user_timeline_event
+                    WHERE user_id = :user_id
+                    AND id = :id
+                    RETURNING *
+                '''), {
+                    'user_id': user_id,
+                    'id': id
+                })
+            try:
+                r = dict(result.fetchone())
+            except TypeError:
+                r = None
+        if not r:
+            return None
+        return UserTimelineEvent(**r)
+    except Exception as e:
+        raise DatabaseException(str(e))
 
 def get_user_timeline_events(user_id: int, event_type: UserTimelineEventType, count: int = 50) -> List[UserTimelineEvent]:
     """ Gets user timeline events of the specified type associated with the specified user.
--- a/listenbrainz/db/user_timeline_event.py
+++ b/listenbrainz/db/user_timeline_event.py
@@ -79,7 +79,7 @@ def create_user_notification_event(user_id: int, metadata: NotificationMetadata)
         metadata=metadata
     )
 
-def delete_user_recommendation_notification_event(
+def delete_user_timeline_event(
         id: int,
         user_id:int
 ) -> bool:
--- a/listenbrainz/db/user_timeline_event.py
+++ b/listenbrainz/db/user_timeline_event.py
@@ -82,7 +82,7 @@ def create_user_notification_event(user_id: int, metadata: NotificationMetadata)
 def delete_user_recommendation_notification_event(
         id: int,
         user_id:int
-) -> UserTimelineEvent:
+) -> bool:
     ''' Deletes recommendation and notification event using id'''
     try:
         with db.engine.connect() as connection:
@@ -90,18 +90,11 @@ def delete_user_recommendation_notification_event(
                     DELETE FROM user_timeline_event
                     WHERE user_id = :user_id
                     AND id = :id
-                    RETURNING *
                 '''), {
                     'user_id': user_id,
                     'id': id
                 })
-            try:
-                r = dict(result.fetchone())
-            except TypeError:
-                r = None
-        if not r:
-            return None
-        return UserTimelineEvent(**r)
+            return result.rowcount == 1
     except Exception as e:
         raise DatabaseException(str(e))
 
--- a/listenbrainz/tests/integration/test_user_timeline_event_api.py
+++ b/listenbrainz/tests/integration/test_user_timeline_event_api.py
@@ -193,7 +193,6 @@ class UserTimelineAPITestCase(ListenAPIIntegrationTestCase):
             data=json.dumps({"metadata": metadata_not}),
             headers={'Authorization': 'Token {}'.format(approved_user['auth_token'])}
         )
-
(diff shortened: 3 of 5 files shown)


data/model/user_timeline_event.py
```diff
@@ -81,3 +81,12 @@ class APITimelineEvent(pydantic.BaseModel):
     user_name: str
     created: int
     metadata: APIEventMetadata
+
+class APIRecommendationNotificationTimelineEvent(pydantic.BaseModel):
+    # since `id` is not sent in APITimelineEvent
```
amCap1712, line 86: You can instead add the `id` field to APITimelineEvent. If all entities are not going to have it, you can make it as `id: Optional[int]`.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -238,6 +239,68 @@ def user_feed(user_name: str):
...
+        }
+    }
+    {
+        "event_type": "notification",
+        "metadata": {
+            id: int
+        }
+    }
+    {
+        "event_type": "recording_pin",
```
amCap1712, line 264: I think it makes more sense to just accept an `id` here. Even the event type is not needed becuse we do not use it in the where clause while deleting.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -238,6 +239,68 @@ def user_feed(user_name: str):
...
+        raise APIUnauthorized("You don't have permissions to delete from this user's timeline.")
+    try:
+        event = ujson.loads(request.get_data())
+        if event["event_type"] in [UserTimelineEventType.RECORDING_RECOMMENDATION.value,
+                UserTimelineEventType.NOTIFICATION.value]:
+            deleted_event = db_user_timeline_event.delete_user_recommendation_notification_event(event["metadata"]["id"],
+                    user["id"])
+            if not deleted_event:
+                raise APIBadRequest("No such event to delete")
+            event_data = deleted_event.dict()
```
amCap1712, line 289: We do not usually return the deleted data. Don't think its needed here either.

amCap1712 (commented):
I have left some comments + also needs tests.

amCap1712 (changes requested):
This looks much better. Thanks. Let's also add tests for these endpoints and database methods.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -239,6 +240,69 @@ def user_feed(user_name: str):
...
+            try:
+                event_deleted = db_user_timeline_event.delete_user_recommendation_notification_event(event["id"], user["id"])
+            except Exception as e:
+                raise APIInternalServerError("Something went wrong. Please try again")
+            if not event_deleted:
+                raise APINotFound("Cannot find '%s' event with id '%s' for user '%s'" % (event["event_type"], event["id"],
+                    user["id"]))
+            return jsonify({"status": "ok"})
+
+        elif event["event_type"] == UserTimelineEventType.RECORDING_PIN.value:
```
amCap1712, line 291: This should probably be removed as Pinned Recordings have their own endpoint to support deletion.

listenbrainz/tests/integration/test_user_timeline_event_api.py
```diff
@@ -182,3 +183,143 @@ def test_get_notification_event(self):
...
+            'track_name': 'Aruarian Dance',
+            'artist_msid':  str(uuid.uuid4()),
+            'recording_msid': str(uuid.uuid4()),
+        }
+        self.client.post(
+            url_for('user_timeline_event_api_bp.create_user_recording_recommendation_event',
+            user_name=new_user['musicbrainz_id']),
+            data=json.dumps({'metadata': metadata_rec}),
+            headers={'Authorization': 'Token {}'.format(new_user['auth_token'])},
+        )
```
amCap1712, line 275: store the response and `assert200` on it

listenbrainz/tests/integration/test_user_timeline_event_api.py
```diff
@@ -182,3 +183,143 @@ def test_get_notification_event(self):
...
```
amCap1712, line 210: store the response and assert200 on it

amCap1712 (commented):
One change, otherwise lgtm. Great job! Thanks.

amCap1712 (approved):

listenbrainz/db/user_timeline_event.py
```diff
@@ -79,6 +79,24 @@ def create_user_notification_event(user_id: int, metadata: NotificationMetadata)
...
+def delete_user_recommendation_notification_event(
+        id: int,
+        user_id:int
+) -> bool:
+    ''' Deletes recommendation and notification event using id'''
+    try:
+        with db.engine.connect() as connection:
+            result = connection.execute(sqlalchemy.text('''
+                    DELETE FROM user_timeline_event
+                    WHERE user_id = :user_id
```
mayhem, line 91: Is user_id even required for this query?

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -239,6 +239,57 @@ def user_feed(user_name: str):
...
+    :param user_name: The MusicBrainz ID of the user from whose timeline events are being deleted
+    :type user_name: ``str``
+    :statuscode 200: Successful deletion
+    :statuscode 400: Bad request, check ``response['error']`` for more details.
+    :statuscode 401: Unauthorized
+    :statuscode 404: User not found
+    :statuscode 500: API Internal Server Error
+    :resheader Content-Type: *application/json*
+    '''
+    # TODO: maybe implement logging once approved
```
mayhem, line 268: Logging for what? that an even was deleted? I don't think that is necessary.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -239,6 +239,57 @@ def user_feed(user_name: str):
...
+    user = validate_auth_header()
+    if user_name != user['musicbrainz_id']:
+        raise APIUnauthorized("You don't have permissions to delete from this user's timeline.")
+
+    try:
+        event = ujson.loads(request.get_data())
+
+        if event["event_type"] in [UserTimelineEventType.RECORDING_RECOMMENDATION.value,
```
mayhem, line 276: The docstring says that pin events are supported for deletion, but this if only allows REC and NOTIFICATION events. Are pins notifications as well, or is there a case missing here?

listenbrainz/db/user_timeline_event.py
```diff
@@ -79,6 +79,24 @@ def create_user_notification_event(user_id: int, metadata: NotificationMetadata)
         metadata=metadata
     )
 
+def delete_user_recommendation_notification_event(
```
mayhem, line 82: This function name seems poorly named, since it can delete more than just a recommendation event.

mayhem (commented):
Some questions for now, before I finish this review.

listenbrainz/db/user_timeline_event.py
amCap1712: To make sure we don't let a user delete another user's events. we use the same the technique for pinned recs.
https://github.com/metabrainz/listenbrainz-server/blob/29baea454b05677aed9f5909085cec3ce0c56db7/listenbrainz/db/pinned_recording.py#L82-L91

listenbrainz/webserver/views/user_timeline_event_api.py
amCap1712: Pinned recordings have their own delete endpoint so I asked it to be removed from here. Ideally, we should have just one delete endpoint for all events but the pinned recs has been in prod so not sure if somebody else is using it. And having pinned recs deletion in two places may cause confusion.

mayhem: So we should fix the docstring or what?

amCap1712: Oh right! I missed the docstring. Yes it should be fixed.

listenbrainz/webserver/views/user_timeline_event_api.py
hrik2001: Alrighty, I will remove this comment :)

listenbrainz/db/user_timeline_event.py
hrik2001: It was supposed to represent the fact that it deletes both recommendation and notification. But I can see how `recommendation_notification` maybe read as one entity. What do you suggest me to name it to? I realised that all methods in `listenbrainz-server/listenbrainz/db/user_timeline_event.py` that deal with only recommendation or notifications, have very generic names (like `create_user_timeline_event`). So should I just name it `delete_user_timeline_event`?

listenbrainz/webserver/views/user_timeline_event_api.py
hrik2001: yes, my bad, I will fix it in the next commit. Sorry :P

mayhem (approved):
Looks good now! Thanks!!

PR #1611, opened 2021-09-13
Show similar user's monthly reports page.

mayhem:
When a user clicks on a similar user from their feed, send them to the similar users monthly reports page -- that is a much better overview of a user than their recent listens.


amCap1712 (approved):

amCap1712:
LGTM. once tests are fixed.

PR #1618, opened 2021-09-17
Add all releases (not use non VA) into the mapping

mayhem:
The previous artist-credit recording pair mapping (1st half of the MBID mapping) didn't include VA releases, when it clearly should. This PR fixes this and is necessary in order to start computing the mbid_mapping that uses artist_mbids instead of artist-credit-id.


amCap1712 (approved):

mayhem:
21 passed, 0 failed.

Test pass, merging.

PR #1619, opened 2021-09-19
Move sitewide stats to new statistics table

amCap1712:
Migrate sitewide stats to new table and store as a special user's stats to simplify handling.

The PR also changes the structure of sitewide stats's API response but these stats haven't been generated since a year so it is unlikely any user is using this API. Hence, this change should be fine.

### DIFF
--- a/listenbrainz/db/stats.py
+++ b/listenbrainz/db/stats.py
@@ -37,7 +37,8 @@ from pydantic import ValidationError
 
 
 # sitewide statistics are stored in the user statistics table
-# as statistics for a special user with the following user_id
+# as statistics for a special user with the following user_id.
+# Note: this is the id from LB's "user" table and *not musicbrainz_row_id*.
 SITEWIDE_STATS_USER_ID = 15753
 
 
--- a/listenbrainz/db/stats.py
+++ b/listenbrainz/db/stats.py
@@ -27,8 +27,6 @@ from typing import Optional
 import sqlalchemy
 
 from data.model.common_stat import StatRange, StatApi
-from data.model.sitewide_artist_stat import (SitewideArtistStat,
-                                             SitewideArtistStatJson)
 from data.model.user_artist_map import UserArtistMapRecord
 from data.model.user_daily_activity import UserDailyActivityRecord
 from data.model.user_entity import UserEntityRecord
@@ -92,19 +90,6 @@ def insert_sitewide_jsonb_data(stats_type: str, stats: StatRange):
             stats: the data to be inserted
     """
     insert_user_jsonb_data(SITEWIDE_STATS_USER_ID, stats_type, stats)
-
-
-def insert_sitewide_artists(stats_range: str, artists: SitewideArtistStatJson):
-    """Inserts sitewide artist stats calculated from Spark into the database.
-
-       If stats are already present for a time range, they are updated to the new
-       values passed.
-
-       Args: stats_range: the range for which the stats have been calculated,
-             artists: the top artists for a particular stats_range
-    """
-    _insert_sitewide_jsonb_data(
-        stats_range, column='artist', data=artists.dict(exclude_none=True))
 
 
 def get_user_stats(user_id: int, stats_range: str, stats_type: str) -> Optional[StatApi[UserEntityRecord]]:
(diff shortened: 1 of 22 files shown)


listenbrainz/db/stats.py
```diff
@@ -38,6 +36,11 @@
 from pydantic import ValidationError
 
 
+# sitewide statistics are stored in the user statistics table
+# as statistics for a special user with the following user_id
+SITEWIDE_STATS_USER_ID = 15753
```
mayhem, line 41: Should this perhaps be user 1 -- the Anonymous user in musicbrainz for various purposes?

listenbrainz_spark/stats/sitewide/artist.py
```diff
@@ -1,49 +1,36 @@
 from listenbrainz_spark.stats import run_query
-from pyspark.sql.functions import collect_list, sort_array, struct
 
 
-def get_artists(table: str, date_format: str):
+def get_artists(table: str, limit: int = 1000):
```
mayhem, line 4: Should 1000 be a constant or documented somewhere?

listenbrainz_spark/stats/sitewide/entity.py
```diff
@@ -25,105 +23,70 @@
...
-
-    # Genarate a dataframe containing days of last and current week along with start and end time
-    time_range = []
-    while day < to_date:
-        time_range.append([day.strftime('%A %d %B %Y'), int(day.timestamp()), int(get_day_end(day).timestamp())])
-        day = offset_days(day, 1, shift_backwards=False)
-
-    time_range_df = listenbrainz_spark.session.createDataFrame(time_range, schema=time_range_schema)
-    time_range_df.createOrReplaceTempView('time_range')
+    from_date = offset_days(to_date, 7)
```
mayhem, line 32: Previously the range was 14 days, now is 7 days. Why has this changed?

listenbrainz_spark/stats/sitewide/entity.py
```diff
@@ -133,31 +96,24 @@ def get_entity_all_time(entity: str) -> Optional[List[SitewideEntityStatMessage]
     to_date = get_latest_listen_ts()
     from_date = datetime(LAST_FM_FOUNDING_YEAR, 1, 1)
 
-    # Generate a dataframe containing years from "from_date" to "to_date"
```
mayhem, line 136: This whole block was never used?

mayhem (approved):
I have a couple of questions, but overall this looks good! I love PRs that mostly delete complicated code, like date calculations!

listenbrainz_spark/stats/sitewide/artist.py
amCap1712: This is the default the code was using till now. I was thinking to make this configurable through spark request stats commands. Do you think that's needed or just making this a constant is fine?

listenbrainz/db/stats.py
amCap1712: This is the ListenBrainz User ID, not the musicbrainz row id. Currently, LB user id `1` is in use.

listenbrainz_spark/stats/sitewide/entity.py
amCap1712: This used to calculate the top artitst stats for each day of the last two weeks earlier. Not sure why because for user stats, we just calculate the overall top artist of the current week. So, I changed 14 to 7 to make it consistent.

mayhem: :+1:

listenbrainz_spark/stats/sitewide/entity.py
amCap1712: This is the part we discussed briefly in IRC last week. The way this had been implemented till now was that if spark was requested all time sitewide stats, it would generate the top artists for each year since founding i.e. 2005 top artists, 2006 top artists, 2007 top artists .... 2021 top artists. While all time user stats simply mean the overall top artists.

Similarly, year sitewide stats for eg: of 2021 meant top artists of Jan 2021, top artists of Feb 2021 and so on. While year user stats mean the overall top artists of the year.

listenbrainz_spark/stats/sitewide/artist.py
mayhem: I would guess a constant is fine for now -- until we realize that we do really need it to be changeable.

listenbrainz/db/stats.py
mayhem: Ah, I see.

PR #1620, opened 2021-09-20
Refactor ListenCard component

MonkeyDo:
The time has come to refactor the ListenCard!

The ListenCard component is one of the main components we have, the one used to display your listens or on the recent listens page.
1. It was in dire need of a refactor to get rid of MediaQuery react component (it's not good to have two different layouts which duplicate content, which is quite error prone and can be solved easily with CSS
2. The CSS also needed some refactoring
3. The component required some changes in order to be used it in all pages that display a list of tracks that should be playable, like/hatable, pinnable, etc. Those user features are currently not available on other pages such as the Feed page (!), recommendations page, etc. (which each use their own card component)

To limit the size and scope of this PR a bit, I'm only replacing the refactored ListenCard in the UserFeed component.
I will still have to do the same and replace the PlaylistItemCard, RecommendationCard and maybe PinnedRecordingCard components in other pages.

Desktop size:
image

Tablet size:
image

Small phone size:
image

And used in the UserFeed component (notice we now have like/hate and more actions menu):
image

### DIFF
--- a/listenbrainz/webserver/static/css/listens-page.less
+++ b/listenbrainz/webserver/static/css/listens-page.less
@@ -199,7 +199,6 @@
 	}
 	
 	> div {
-		padding: 0 1em;
 		display: flex;
 		flex:1;
 	}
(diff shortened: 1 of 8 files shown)


akshaaatt (approved):
Works great! Thanks for the much needed change. :100:

PR #1622, opened 2021-09-21
Hide Report User button

akshaaatt:
- If the user is logged in and is looking at their own profile, we hide the report button.

### DIFF
--- a/listenbrainz/webserver/static/js/src/UserPageHeading.tsx
+++ b/listenbrainz/webserver/static/js/src/UserPageHeading.tsx
@@ -51,7 +51,7 @@ const UserPageHeading = ({
           />
         )}
       </h2>
-      {loggedInUser && user.name !== loggedInUser.name && (
+      {loggedInUser && user?.name !== loggedInUser.name && (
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/static/js/src/UserPageHeading.tsx
```diff
@@ -51,7 +51,7 @@ const UserPageHeading = ({
           />
         )}
       </h2>
-      {loggedInUser && (
+      {loggedInUser && user.name !== loggedInUser.name && (
```
MonkeyDo, line 54:
```suggestion
      {loggedInUser && user?.name !== loggedInUser.name && (
```
 I'm not sure if user can be undefined, but it doesn't hurt to make sure we don't try to access a subproperty of undefined

MonkeyDo (commented):
Thanks ! :rocket:

listenbrainz/webserver/static/js/src/UserPageHeading.tsx
akshaaatt: I like the way you code :)

PR #1624, opened 2021-09-22
LB-934, LB-935 (I): Spark statistics time fix and code deduplication

amCap1712:
1) Refactor the spark code and extract common code into separate method.

2) Set from_date to midnight of that day. Till now, we used to calculate the latest listen's timestamp as the ending limit and then subtract 7 days to get from_ts. However, some stats manually set the from_ts to midnight of that date. For consistency, I suggest we do this always.

3) Refactor and fix all time listening activity code. All ranges in listening_activity add years for which the user has no listens as years with listen_count = 0. week, month and year code share the code to do this and worked as expected. Due to some reason, all_time range had difference code to do this but looking at current output of API that didn't work. I changed the all time code to use the same query and other ranges and it works now.

Will open a subsequent PR for some more code simplifiation to avoid making this one too large.

### DIFF
--- a/listenbrainz_spark/stats/sitewide/entity.py
+++ b/listenbrainz_spark/stats/sitewide/entity.py
@@ -16,126 +16,95 @@ logger = logging.getLogger(__name__)
 
 
 entity_handler_map = {
-    'artists': get_artists,
+    "artists": get_artists,
 }
 
 entity_model_map = {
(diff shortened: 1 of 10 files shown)


mayhem (approved):
Deleting code. \o/ Cleaner code \o\ /o/ Thanks!

PR #1626, opened 2021-09-25
Deduplicates spark queries

amCap1712:
The last batch of stats code deduplication 😄 

Having one query for eah type and stats_range combination add a lot of redundant code to maitain and write when adding new ranges and type. Hence, deduplicating this.


mayhem (approved):
Yay, cleaner code! Thanks!

PR #1627, opened 2021-09-26
Search Bar Addition and UI Cleanup

akshaaatt:
Screenshot from 2021-09-26 21-58-31
- A search box has been added to the navbar which facilitates us to search for fellow listeners.
- Tracks you might like has been moved from the dropdown and renamed to `For me` to house directly in the navbar since it is one of the best features of listenbrainz.

### DIFF
--- a/listenbrainz/webserver/static/css/main.less
+++ b/listenbrainz/webserver/static/css/main.less
@@ -62,7 +62,7 @@ body {
   .navbar-form {
     padding-right: 0px;
     .btn {
-      background-color: @turquoise;
+      background-color: #eb743b;
       margin-right: 0px;
     }
   }
--- a/listenbrainz/webserver/templates/navbar.html
+++ b/listenbrainz/webserver/templates/navbar.html
@@ -21,7 +21,8 @@
           {% if current_user.is_authenticated %}
             <li><a href="{{ url_for('user.profile', user_name=current_user.musicbrainz_id) }}">My Listens</a></li>
             <li><a href="{{ url_for('index.feed') }}">My Feed</a></li>
-          {% endif %}
+        <li><a href="{{ url_for('recommendations_cf_recording.info', user_name=current_user.musicbrainz_id) }}">For me</a></li>
+        {% endif %}
         {%- endif -%}
         <li><a href="{{ url_for('index.recent_listens') }}">Recent</a></li>
         <li class="dropdown">
@@ -72,10 +73,7 @@
               {{ current_user.musicbrainz_id }} <span class="caret"></span>
             </a>
             <ul class="dropdown-menu" role="menu">
-              <li><a href="{{ url_for('user.profile', user_name=current_user.musicbrainz_id) }}">My Listens</a></li>
-              <li><a href="{{ url_for('index.feed') }}">My Feed</a></li>
               <li><a href="{{ url_for('user.playlists', user_name=current_user.musicbrainz_id) }}">My Playlists</a></li>
-              <li><a href="{{ url_for('recommendations_cf_recording.info', user_name=current_user.musicbrainz_id) }}">Tracks you might like</a></li>
               <li><a href="{{ url_for('profile.music_services_details') }}">Connect Services</a></li>
               <li><a href="{{ url_for('profile.import_data') }}">Import listens</a></li>
               <li><a href="{{ url_for('profile.export_data') }}">Export</a></li>
@@ -86,7 +84,22 @@
         {%- endif -%}
       </ul>
     {% endif %}
+      <form id="searchForm" class="navbar-form">
+          <input id="searchQuery" type="text" class="form-control input-sm" placeholder="Search Listener">
+        <button type="submit" class="btn input-sm"><span class="glyphicon glyphicon-search"></span></button>
+      </form>
     </div>
-
   </div>
 </nav>
+<script type="text/javascript">
+  const query = document.getElementById('searchQuery');
+  const form = document.getElementById('searchForm');
+  form.onsubmit = function() {
+    console.log(query.value);
+    if(query.value.trim().length<1){
+      return false;
+    }
+    window.location = "https://listenbrainz.org/user/"+query.value;
+    return false;
+  };
+</script>
--- a/listenbrainz/webserver/templates/user/base.html
+++ b/listenbrainz/webserver/templates/user/base.html
@@ -5,7 +5,6 @@
 {% block content %}
   <div id="user-profile">
     <div id="user-page-heading-container"></div>
-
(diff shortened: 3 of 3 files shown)


listenbrainz/webserver/templates/navbar.html
```diff
@@ -86,7 +84,22 @@
         {%- endif -%}
       </ul>
     {% endif %}
+      <form id="searchForm" class="navbar-form">
+          <input id="searchQuery" type="text" class="form-control input-sm" placeholder="Search Listener">
```
amCap1712, line 88: Search sounds misleading because there's actually no search. If the user enters a wrong username, they'll be sent to a 404. Something like "Go to Profile" is probably clearer.

akshaaatt: Sounds reasonable. I'll make the change. Thanks!

mayhem:
I have some misgivings about this user experience of this PR -- namely handling cases where the user in question was not found. A 404 page is a pretty crap result for a "go to profile" feature that is so prominent. One way to fix this is to improve the 404 page and check to see if the page is shown as a result of a go to profile request. If so, we can show a better error page and also provide a link to MB's editor search. Thoughts?

amCap1712:
Customizing the user 404 and linking to MB search sounds good. But how about implementing the seach actually in LB? A simple LIKE search of musicbrainz_id in LB user table should suffice?

akshaaatt:
@mayhem Totally agree that a 404 page is not desirable. From my end, I am willing to take up improving the 404 page and provide a link to MB's editor search.
@amCap1712 I would love the search option and if you could work on that, that would be great. But if that feature takes a while to implement, we can go ahead with the first solution mentioned and accordingly with time add the search option :)

MonkeyDo:
I think the search functionality will be good, and its placement in the nav bar makes sense.

I'm not fully convinced by the "For me" menu option however. I think we should get together and discuss this topic some more as there are multiple music discovery features that we are currently working on.
I agree with you that we should make it more prominent, but it also makes sense to have them all in one place and that will take a bit more effort.
I think it would maybe be better to wait until we have a better idea of how we want to overhaul that music discovery page rather than just moving this one bit for now.

listenbrainz/webserver/templates/navbar.html
```diff
@@ -72,21 +73,32 @@
               {{ current_user.musicbrainz_id }} <span class="caret"></span>
             </a>
             <ul class="dropdown-menu" role="menu">
-              <li><a href="{{ url_for('user.profile', user_name=current_user.musicbrainz_id) }}">My Listens</a></li>
-              <li><a href="{{ url_for('index.feed') }}">My Feed</a></li>
-              <li><a href="{{ url_for('user.playlists', user_name=current_user.musicbrainz_id) }}">My Playlists</a></li>
```
MonkeyDo, line 77: "My Listens" and "My Feed" are duplicated in the nav bar, but there is no way to access "My playlists" directly so I think it should remain there.

As I commented on the PR already I think the recommendation below should also remain here until we design a better music discovery page

listenbrainz/webserver/templates/user/base.html
```diff
@@ -16,9 +15,6 @@
         <li {{ 'class=active' if active_section=='recommendations' }}><a href="{{ url_for('user.recommendation_playlists', user_name=user.musicbrainz_id) }}">Recommendations</a></li>
         <li {{ 'class=active' if active_section=='collaborations' }}><a href="{{ url_for('user.collaborations', user_name=user.musicbrainz_id) }}">Collaborations</a></li>
       </ul>
-
-    {% block profile_content %} {% endblock %}
-
```
MonkeyDo, line 21: Why was this removed?
I'm not so well versed in this templating system, but since this is a base for other templates I believe this profile_content block can be extended and some content added to it like it is done here: https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/templates/recommendations_cf_recording/top_artist.html#L3-L5

listenbrainz/webserver/static/css/main.less
```diff
@@ -62,7 +62,7 @@ body {
   .navbar-form {
     padding-right: 0px;
     .btn {
-      background-color: @turquoise;
+      background-color: #eb743b;
```
MonkeyDo, line 65:
```suggestion
      background-color: @orange;
```
See this file: https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/css/theme/variables.less#L11-L22

akshaaatt: Thanks for the correct implementation there. I missed correcting that during development and was testing variations on the spot. Hah!

listenbrainz/webserver/templates/user/base.html
akshaaatt: I actually decided to move this from under the listens to the `My profile` section.

listenbrainz/webserver/templates/navbar.html
akshaaatt: Makes sense. I will revert the `For me` changes and let the tracks discovery stay for now as we set some time to think of how the implementation can be done in a better way.

listenbrainz/webserver/templates/user/base.html
MonkeyDo: I don't think I understand your comment.
I just see `{% block profile_content %} {% endblock %}`  removed and not added anywhere else.

listenbrainz/webserver/templates/navbar.html
```diff
@@ -72,21 +73,33 @@
...
+      <form id="searchForm" class="navbar-form">
+          <input id="searchQuery" type="text" class="form-control input-sm" placeholder="Go to Profile">
+        <button type="submit" class="btn input-sm"><span class="glyphicon glyphicon-search"></span></button>
+      </form>
     </div>
-
   </div>
 </nav>
+<script type="text/javascript">
+  const query = document.getElementById('searchQuery');
```
amCap1712, line 95: No, need to do it this way. Change the form action to /search and method to get. For the empty check, make the query input required?

akshaaatt: @amCap1712 I was unable to make a url modification through that because that makes a query and appends that to the url instead of modifying it.

amCap1712: Oh sorry. I should have added more context. I have implemented actual user search in a different PR. Let me finish that up and see how we can fix this up.

amCap1712: Search box has now been added to #1628

PR #1628, opened 2021-09-28
Add user name search

amCap1712:
Add support for searching users by user name. We use `pg_trgm`'s `word_similarity` function to power the search. GIN indexes are usually better for read heavy data but the docs say that for these specific queries, its better to use GiST index. See [pg_trgm Index Support](https://www.postgresql.org/docs/12/pgtrgm.html#:~:text=This%20can%20be%20implemented%20quite%20efficiently%20by%20GiST%20indexes%2C%20but%20not%20by%20GIN%20indexes.%20It%20will%20usually%20beat%20the%20first%20formulation%20when%20only%20a%20small%20number%20of%20the%20closest%20matches%20is%20wanted.)


admin/sql/create_extensions.sql
```diff
@@ -1 +1,3 @@
 CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+CREATE EXTENSION IF NOT EXISTS "pg_trgm";
+ALTER DATABASE listenbrainz SET pg_trgm.word_similarity_threshold = 0.1;
```
amCap1712, line 3: The default threshold is 0.3 which eliminates a lot of possible matches.

mayhem: I believe that -- our "documents" are quite short in comparison with what others might put into a document.

listenbrainz/db/tests/test_user.py
```diff
@@ -215,3 +215,12 @@ def test_fetch_email(self):
 
         self.assertNotIn("email", db_user.get_by_mb_id(musicbrainz_id))
         self.assertEqual(email, db_user.get_by_mb_id(musicbrainz_id, fetch_email=True)["email"])
+
+    def test_search(self):
+        db_user.create(0, "Cécile")
+        db_user.create(1, "Cecile")
+        db_user.create(2, "lucifer")
+        db_user.create(3, "rob")
```
mayhem, line 223: Wooo, mission accomplished, I've made it into software tests!

listenbrainz/db/user.py
```diff
@@ -537,3 +537,18 @@ def update_user_email(musicbrainz_id, email):
             logger.error(err)
             raise DatabaseException(
                 "Couldn't update user's email: %s" % str(err))
+
+
+def search(search_term: str, limit: int) -> List[Tuple[str, float]]:
+    with db.engine.connect() as connection:
```
mayhem, line 543: docstring?

listenbrainz/webserver/templates/index/search-users.html
```diff
@@ -0,0 +1,17 @@
...
+{%- block content -%}
+    <h2 class="page-title">Search Results for {{ search_term }}</h2>
+    <table class="table table-striped">
+    {% for row in users %}
+        <tr>
+            <td>{{ loop.index }}</td>
+            <td><a href="{{ url_for("user.profile", user_name=row[0]) }}">{{ row[0] }}</a></td>
+        </tr>
+    {% else %}
+       <tr>
```
mayhem, line 12: I think it would be good to add a textbox that allows the user to search again to this page.

mayhem (approved):
Thanks for this! I think adding the search again text box to the page would be great, then merge!

listenbrainz/webserver/templates/index/search-users.html
amCap1712: Done.

listenbrainz/db/user.py
amCap1712: Done

listenbrainz/db/user.py
```diff
@@ -537,3 +537,45 @@ def update_user_email(musicbrainz_id, email):
...
+            """), {
+            "search_term": search_term,
+            "limit": limit
+        })
+
+        rows = result.fetchall()
+        if not rows:
+            return []
+        similar_users = get_similar_users(searcher_id) if searcher_id else None
+        logger.error("Similar users: %s", similar_users)
```
amCap1712, line 573: Forgot to remove after debuggin. Fixing

mayhem (approved):
Still looks good to me! :)

PR #1629, opened 2021-09-28
LB-860 (I): Add quarterly stats backend support

amCap1712:
Add quarterly stats range backend. Also, moved some stuff a bit around to reduce duplication.

### DIFF
--- a/listenbrainz/spark/request_manage.py
+++ b/listenbrainz/spark/request_manage.py
@@ -273,26 +273,31 @@ def request_similar_users(max_num_users):
 def cron_request_all_stats(ctx):
     ctx.invoke(request_user_stats, type_="entity", range_="week", entity="artists")
     ctx.invoke(request_user_stats, type_="entity", range_="month", entity="artists")
+    ctx.invoke(request_user_stats, type_="entity", range_="quarter", entity="artists")
     ctx.invoke(request_user_stats, type_="entity", range_="year", entity="artists")
     ctx.invoke(request_user_stats, type_="entity", range_="all_time", entity="artists")
 
     ctx.invoke(request_user_stats, type_="entity", range_="week", entity="releases")
     ctx.invoke(request_user_stats, type_="entity", range_="month", entity="releases")
+    ctx.invoke(request_user_stats, type_="entity", range_="quarter", entity="releases")
     ctx.invoke(request_user_stats, type_="entity", range_="year", entity="releases")
     ctx.invoke(request_user_stats, type_="entity", range_="all_time", entity="releases")
 
     ctx.invoke(request_user_stats, type_="entity", range_="week", entity="recordings")
     ctx.invoke(request_user_stats, type_="entity", range_="month", entity="recordings")
+    ctx.invoke(request_user_stats, type_="entity", range_="quarter", entity="recordings")
     ctx.invoke(request_user_stats, type_="entity", range_="year", entity="recordings")
     ctx.invoke(request_user_stats, type_="entity", range_="all_time", entity="recordings")
 
     ctx.invoke(request_user_stats, type_="listening_activity", range_="week")
     ctx.invoke(request_user_stats, type_="listening_activity", range_="month")
+    ctx.invoke(request_user_stats, type_="listening_activity", range_="quarter")
     ctx.invoke(request_user_stats, type_="listening_activity", range_="year")
     ctx.invoke(request_user_stats, type_="listening_activity", range_="all_time")
 
     ctx.invoke(request_user_stats, type_="daily_activity", range_="week")
     ctx.invoke(request_user_stats, type_="daily_activity", range_="month")
+    ctx.invoke(request_user_stats, type_="daily_activity", range_="quarter")
     ctx.invoke(request_user_stats, type_="daily_activity", range_="year")
     ctx.invoke(request_user_stats, type_="daily_activity", range_="all_time")
 
--- a/listenbrainz/spark/request_manage.py
+++ b/listenbrainz/spark/request_manage.py
@@ -6,6 +6,8 @@ import pika
 import ujson
 
 from flask import current_app
+
+from data.model.common_stat import ALLOWED_STATISTICS_RANGE
 from listenbrainz.webserver import create_app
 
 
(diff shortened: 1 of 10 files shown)


listenbrainz/spark/request_manage.py
```diff
@@ -269,30 +271,12 @@ def request_similar_users(max_num_users):
...
-    ctx.invoke(request_user_stats, type_="listening_activity", range_="week")
-    ctx.invoke(request_user_stats, type_="listening_activity", range_="month")
-    ctx.invoke(request_user_stats, type_="listening_activity", range_="year")
-    ctx.invoke(request_user_stats, type_="listening_activity", range_="all_time")
-
-    ctx.invoke(request_user_stats, type_="daily_activity", range_="week")
-    ctx.invoke(request_user_stats, type_="daily_activity", range_="month")
-    ctx.invoke(request_user_stats, type_="daily_activity", range_="year")
-    ctx.invoke(request_user_stats, type_="daily_activity", range_="all_time")
+    for stats_range in ALLOWED_STATISTICS_RANGE:
```
alastair, line 274: this does all types for a range and then moves on to the next range, as opposed to what was before when it was all ranges for a type. Does it matter in which order we do it?

listenbrainz_spark/stats/__init__.py
```diff
@@ -145,6 +165,9 @@ def get_dates_for_stats_range(stats_range: str) -> Tuple[datetime, datetime]:
     elif stats_range == "month":
         from_offset = relativedelta(months=-1, day=1)  # first day of previous month
         to_offset = relativedelta(months=+1)
+    elif stats_range == "quarter":
+        from_offset = get_last_quarter_offset(latest_listen_date)
+        to_offset = relativedelta(months=+3)
```
alastair, line 170: We had a 1 line discussion in IRC suggesting that we could also have the current quarter interpreted as "up-to-today". Do we also want this?

https://chatlogs.metabrainz.org/libera/metabrainz/2021-09-28/?msg=4865940&page=1

listenbrainz_spark/stats/user/listening_activity.py
```diff
@@ -25,6 +25,26 @@
 logger = logging.getLogger(__name__)
 
 
+def _get_quarter_offset(_date: date) -> relativedelta:
```
alastair, line 28: it's not immediately clear to me the reason for having both this method and the "1 quarter back" method in `listenbrainz_spark/stats/__init__.py`. At the very least, the name for this method is confusing without reading the docstring.

listenbrainz_spark/stats/user/listening_activity.py
```diff
@@ -60,6 +80,12 @@ def get_time_range(stats_range: str) -> Tuple[datetime, datetime, relativedelta,
         # compute listening activity for each day but no weekday
         step = relativedelta(days=+1)
         date_format = "%d %B %Y"
+    elif stats_range == "quarter":
+        from_offset = _get_quarter_offset(latest_listen_date)
+        to_offset = relativedelta(months=+6)
```
alastair, line 85: I see that all ranges in this method seem to compute "2 ranges back", but by reading the code and docs I don't understand why this is needed. Is there documetnation or a comment explaining why this is done and what behaviour it has?

listenbrainz/spark/request_manage.py
amCap1712: Yes, I mentioned the same in the commit message. I don't think we care.

listenbrainz_spark/stats/__init__.py
amCap1712: That should be a different category of stats if we decide to add it. For now, its last calendar period for all time ranges.

listenbrainz_spark/stats/user/listening_activity.py
amCap1712: I'll improve the docs and method name. The reason is that listening activity stats needs data for last two time periods. So that the frontend can compare those. Like listening activity generates the data for 2 weeks and frontend presents it as a comparision of each weekday of the two weeks.

listenbrainz_spark/stats/user/listening_activity.py
amCap1712: Explained in the comment above. I'll update documentation of code as well.

listenbrainz_spark/stats/user/listening_activity.py
amCap1712: Updated doc for this method. Let me know if I should add more details.

listenbrainz_spark/stats/user/listening_activity.py
amCap1712: Added docs to get_listening_activity.  Let me know if I should add more details. https://github.com/metabrainz/listenbrainz-server/pull/1629/files#diff-749661f792d900d163f3d9a7cabd382211ec8b2f886a525e018f5741f83cceb6R161-R165

listenbrainz_spark/stats/user/listening_activity.py
```diff
@@ -124,7 +155,15 @@ def calculate_listening_activity():
...
+    """ Calculate the number of listens of users for the given stats_range.
+
+    .. note::
+        Compared to other statistics, this stat needs listens for twice the time period.
+        For eg: for week range, this statistic needs the listen data of the last 2 weeks.
+        It will calculate individual listens for each weekday and send to LB webserver.
+        The LB frontend presents a chart comparing the listening activity on corresponding
+        weekdays of each week.
+    """
     logger.debug(f"Calculating listening_activity_{stats_range}")
```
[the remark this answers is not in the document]
amCap1712: Yes, that should be possible. The same should be done for other types of stats. However, this will need rewriting many stats tests so I'd prefer to take that up separately. I'd suggest to create a ticket for now and do this later. Thoughts?

listenbrainz_spark/stats/user/listening_activity.py
```diff
@@ -124,7 +155,15 @@ def calculate_listening_activity():
 
 
 def get_listening_activity(stats_range: str):
-    """ Calculate the number of listens of users for the given stats_range """
+    """ Calculate the number of listens of users for the given stats_range.
```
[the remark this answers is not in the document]
amCap1712: Thanks. fixed.

listenbrainz_spark/stats/user/listening_activity.py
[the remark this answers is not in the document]
alastair: sure, happy to create a ticket for this

alastair (approved):
great! thanks for stepping me through how it works.

Happy to create a ticket for this last item about making the join table otherwise it's good to go

PR #1631, opened 2021-09-28
Add sitewide stats to cronjob

amCap1712:
Sitewide stats should be generated regularly. Hence, add to the stats cron job.


listenbrainz/spark/request_manage.py
```diff
@@ -278,6 +278,9 @@ def cron_request_all_stats(ctx):
         for stat in ["listening_activity", "daily_activity"]:
             ctx.invoke(request_user_stats, type_=stat, range_=stats_range)
 
+        for entity in ["artists"]:
+            ctx.invoke(request_sitewide_stats, type_="entity", range_=stats_range, entity=entity)
```
alastair, line 282: it looks like this method doesn't take a `type_` parameter

amCap1712: 🤦 fixed.

PR #1632, opened 2021-09-29
LB-860 (II): Add half yearly stats backend

amCap1712:
As the title says.


alastair (approved):
I had a quick look at this, it looks like it just adds a new range to the existing functionality, right? If so, looks good!

PR #1637, opened 2021-09-30
Switch from curl to wget for key downloading

treeshateorcs:
# Problem
because curl fails with the following error

```
...
 => CACHED [listenbrainz-base  2/13] RUN wget https://github.com/jwilder/  0.0s
 => CACHED [listenbrainz-base  3/13] RUN wget -O /usr/local/bin/sentry-cl  0.0s
 => CACHED [listenbrainz-base  4/13] RUN apt-get update     && apt-get in  0.0s
 => CACHED [listenbrainz-base  5/13] RUN update-ca-certificates            0.0s
 => ERROR [listenbrainz-base  6/13] RUN curl https://www.postgresql.org/m  1.1s
------                                                                          
 > [listenbrainz-base  6/13] RUN curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -:                                                       
#9 0.229   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current                                                                        
#9 0.237                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
#9 0.812 curl: (60) server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
#9 0.812 More details here: http://curl.haxx.se/docs/sslcerts.html
#9 0.812 
#9 0.812 curl performs SSL certificate verification by default, using a "bundle"
#9 0.812  of Certificate Authority (CA) public keys (CA certs). If the default
#9 0.812  bundle file isn't adequate, you can specify an alternate file
#9 0.812  using the --cacert option.
#9 0.812 If this HTTPS server uses a certificate signed by a CA represented in
#9 0.812  the bundle, the certificate verification probably failed due to a
#9 0.812  problem with the certificate (it might be expired, or the name might
#9 0.812  not match the domain name in the URL).
#9 0.812 If you'd like to turn off curl's verification of the certificate, use
#9 0.812  the -k (or --insecure) option.
#9 1.091 gpg: no valid OpenPGP data found.
------
executor failed running [/bin/sh -c curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -]: exit code: 2
ERROR: Service 'web' failed to build : Build failed
```

# Solution
Switch from curl to wget

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # PostgreSQL client
-RUN curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
+RUN wget -O- https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
 ENV PG_MAJOR 12
 RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ xenial-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
 RUN apt-get update \


amCap1712:
Hi, The problem is not with curl but rather openssl or gnutls. A let's encrypt cert has expired and hence some part of ssl chain works incorrectly due to a bug in openssl. I am not sure why it seemingly worked in the GH action. For example, I tried it locally and still got the cert error with wget. 
```
docker run --rm -it metabrainz/python:3.7-20210115 bash
root@36a7054c1d0b:/# wget -O- https://www.postgresql.org/media/keys/ACCC4CF8.asc
--2021-10-01 12:08:03--  https://www.postgresql.org/media/keys/ACCC4CF8.asc
Resolving www.postgresql.org (www.postgresql.org)... 72.32.157.230, 87.238.57.232, 217.196.149.50, ...
Connecting to www.postgresql.org (www.postgresql.org)|72.32.157.230|:443... connected.
ERROR: cannot verify www.postgresql.org's certificate, issued by ‘CN=R3,O=Let's Encrypt,C=US’:
  Issued certificate has expired.
To connect to www.postgresql.org insecurely, use `--no-check-certificate'.
```
I do not know much about the intricate details of openssl etc. but acc. to my understanding this fix looks wrong. Can you tell why do you think changing curl to wget should fix the issue?

treeshateorcs:
i don't entirely understand why it happens with curl either, maybe wget doesn't use `ca-certificates` or something? 

would be great if someone explained what this is all about. so far my understanding is that curl downloads a certificate which doesn't pass another certificate or something, i can't wrap my head around it, but trying downloading with wget was my initial thought and it worked

amCap1712:
I don't understand this stuff very well but this might be useful https://aws.amazon.com/premiumsupport/knowledge-center/ec2-expired-certificate/

PR #1640, opened 2021-10-03
Temporary workaround for ssl cert expiry

amCap1712:
Just a temporary workaround, copy the changes to your branches as required. Preferably not merge this patch in master.


amCap1712:
Looks like something changed again due to which it not working :/

amCap1712:
See #1641

PR #1641, opened 2021-10-03
Fix ssl error by removing expired Let's Encrypt cert and update cacerts

amCap1712:
Preferred workaround for fixing let's encrypt expired cert. Install new certs (including new let's encrypt new cert), remove the expired ones, updates the local cert store.


amCap1712:
https://aws.amazon.com/premiumsupport/knowledge-center/ec2-expired-certificate/ suggests another workaround to update openssl but that didn't work. I tried updating all packages in the image, that works but will is not desirable. We'll need to find the subset of packages needed to upgrade if we want to go that route.

amCap1712:
Merged as part of #1635

PR #1644, opened 2021-10-06
LB-957: Separate APIService from BrainzPlayer

tandy-1000:
https://tickets.metabrainz.org/browse/LB-957

Initial commit adapts the code from APIService into the `submitListenToListenBrainz` method in BP.

### DIFF
--- a/listenbrainz/webserver/static/js/src/BrainzPlayer.test.tsx
+++ b/listenbrainz/webserver/static/js/src/BrainzPlayer.test.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import { mount } from "enzyme";
+import fetchMock from "jest-fetch-mock";
 import BrainzPlayer, { DataSourceType } from "./BrainzPlayer";
 import SoundcloudPlayer from "./SoundcloudPlayer";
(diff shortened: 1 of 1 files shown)


tandy-1000:
> There are two other instances in this component where we're using the APIService, and then we'll be free to remove any mention of it from this file: `APIService.refreshSpotifyToken` and `APIService.refreshYoutubeToken`, both of which should be passed as [new props to BrainzPlayer](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/RecentListens.tsx#L33-L42). Considering my comment below about passing `api_url` as a prop in the same way, it would make sense to do these three changes together (in this PR or in a new one, as you prefer).

No idea if I've done this last part right, wasn't really sure about it.. but should be done now!

tandy-1000:
ping @MonkeyDo :o

mayhem:
Hey @tandy-1000 : Today is a holiday and @MonkeyDo is taking a long weekend. Give him some more time, please!

tandy-1000:
Sorry!! didn't know it was a holiday! take your time monkey :o

PR #1648, opened 2021-10-07
Delete now unused statistics.user table

amCap1712:
statistics.user table is now replaced by statistics.user_new table. The latter has been working fine in prod for some time hence remove the old table. Also, update the dumps accordingly.


amCap1712:
I intended to do that in the future. Thinking again might as well do it now. will update the PR.

alastair (approved):

PR #1649, opened 2021-10-07
Upgrade to python 3.9 using a focal (20.04) base image

alastair:
# Problem
We need to upgrade our base images to a supported ubuntu!

# Solution
Remove pxz as xz provided by xz-utils is already installed and has
multithread support
Remove explicit pip installation as we no longer need to pin an old
version for dependency resolution

# Action

leaving this PR as a draft while we test it on test.lb


listenbrainz/db/dump.py
```diff
@@ -153,7 +153,7 @@
 }
 
 
-def dump_postgres_db(location, dump_time=datetime.today(), threads=None):
+def dump_postgres_db(location, dump_time=datetime.today(), threads=DUMP_DEFAULT_THREAD_COUNT):
```
amCap1712, line 156: Tests were failing because threads=None, caused xz to complain. Not sure why this worked earlier. pxz probably ignored invalid values of -T. Looking at docs, this is max number of threads to use for 
 compression/decompression. Some other dump methods use DUMP_DEFAULT_THREAD_COUNT for this so using it here as well.

alastair:
Closed in favour of #1902

PR #1650, opened 2021-10-07
Add instructions to push test images to Readme

amCap1712:


alastair:
To me these are really big images in our readme, which is a small feature used by only a handful of developers. I think that this isn't the place for this documentation. Perhaps in the docs would be better? Or otherwise syswiki? (I don't like to suggest this one, because we'll forget about it, but...)

README.md
```diff
@@ -64,6 +64,16 @@ also build the documentation locally:
     cd listenbrainz-server/docs
     pip install -r requirements.txt
     make clean html
+    
+## Pushing test images using Github Actions
+
+1. Go to Actions -> Push deployment image or click [here](https://github.com/metabrainz/listenbrainz-server/actions/workflows/push-dev-image.yml).
+2. Select the branch and enter the docker image tag (version).
+![Screenshot (54)](https://user-images.githubusercontent.com/27751938/136417112-20184ff0-3140-425f-89fe-bae625b820a3.png)
```
alastair, line 72: where is this image stored?

amCap1712: This image is curently on github servers, I'll move it to the repo so that we can use it on RTD. We have an images folder under docs for this.

amCap1712:
Moved instructions to RTD docs.

docs/maintainers/docker-image.rst
```diff
@@ -0,0 +1,50 @@
...
+^^^^^^^^^^^^^^^^^
+When a Github release is made, production images are automatically built and pushed by the
+`Publish image <https://github.com/metabrainz/listenbrainz-server/actions/workflows/deploy-image.yml>`_ action.
+The git tag associated with the Github release is used as docker image tag.
+
+
+Test Images
+^^^^^^^^^^^
+
+Time to time we want to build images to test PRs on beta.listenbrainz.org or test.listenbrainz.org. To build
```
alastair, line 14:
```suggestion
From time to time we want to build images to test PRs on beta.listenbrainz.org or test.listenbrainz.org. To build
```

docs/maintainers/docker-image.rst
```diff
@@ -0,0 +1,50 @@
+Building Docker Images
+======================
```
alastair, line 2: > note: this documentation is for ListenBrainz maintainers for when they deploy the website

or something like this?

docs/maintainers/docker-image.rst
```diff
@@ -0,0 +1,50 @@
...
+
+.. image:: ../images/release-result.png
+  :alt: Action Workflow Running Screenshot
+
+
+Using docker/push.sh script
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If Github Actions is unavailable or you want to take advantage of local docker build cache, you can use the
+`docker/push.sh` script. From the repository root, invoke the script with desired docker image tag. For eg:
```
alastair, line 46: e.g. or for example, not both :)

Add "You will need to be correctly authenticated to docker hub to push this image"

alastair (commented):
great to see this finally moved, thanks!
Just a few very small things

docs/maintainers/docker-image.rst
amCap1712: This is listed under the maintainers section so don't think its needed. But then we don't have maintainers documentation index page so probably fine to add.

alastair: right, I was going to suggest to put this on the main "maintainers page", but it seems that we don't have one of them? which is why I suggested here

amCap1712: Yeah, makes sense. Another PR adds a few more pages to maintainer docs but no index page there as welll. I'll add one there and then we can the note from here.

alastair: perfect

alastair (approved):

PR #1652, opened 2021-10-08
LB-908: Listens sent to front-end via websockets have a lot of unused keys

amCap1712:
1. Make the listen data sent by Websockets consistent with the API. 

2. The API response for now playing listens now also contains the now playing field and user name inside the listen body. This will make writing frontend code easier.

Its a field addition hence should be backward compatible. The fields at top level are not removed to maintain compatibility.

Old:
```
{
  payload: {
  count: 1,
  listens: [
    {
      track_metadata: {
        additional_info: { },
        artist_name: "All Time Low, Demi Lovato, blackbear",
        release_name: "Monsters (feat. Demi Lovato and blackbear)",
        track_name: "Monsters (feat. Demi Lovato and blackbear)"
      }
    }
  ],
  playing_now: true,
  user_id: "amcap1712"
  }
}
```

New:
```
{
  payload: {
  count: 1,
  listens: [
    {
      playing_now: true,
      track_metadata: {
        additional_info: { },
        artist_name: "All Time Low, Demi Lovato, blackbear",
        release_name: "Monsters (feat. Demi Lovato and blackbear)",
        track_name: "Monsters (feat. Demi Lovato and blackbear)"
      },
      user_name: "amCap1712"
    }
  ],
  playing_now: true,
  user_id: "amcap1712"
  }
}
```

### DIFF
--- a/listenbrainz/listen.py
+++ b/listenbrainz/listen.py
@@ -232,6 +232,13 @@ class NowPlayingListen:
             data['additional_info'] = flatten_dict(additional_info)
             self.data = data
 
+    def to_api(self):
+        return {
+            "user_name": self.user_name,
+            "track_metadata": self.data,
+            "playing_now": True
+        }
+
     def __repr__(self):
         from pprint import pformat
         return pformat(vars(self))
--- a/listenbrainz/listen.py
+++ b/listenbrainz/listen.py
@@ -234,7 +234,6 @@ class NowPlayingListen:
 
     def to_api(self):
         return {
-            "user_name": self.user_name,
             "track_metadata": self.data,
             "playing_now": True
         }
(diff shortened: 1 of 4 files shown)


amCap1712:
That inconsistency stems from the structure of the  json current listen endpoint. For example, see https://api.listenbrainz.org/1/user/amCap1712/listens?count=1 . The same issue appears here and this has been in production since almost the beginning of LB (I guess?). And my aim was to harmonise the structure of the now playing listens and normal submitted listens so I chose to inherit this evil. But I do think to add more weird stuff because there's already some weird stuff elsewhere is a double edged sword. One hand the API response gets weirder on the other frontend and possibly other consumers have to implement different logic for playing_now/submitted listens.

amCap1712:
I updated the PR to only use consistent structure in api and websockets for now playing listens and normal submitted listens. I have dropped all changes related to adding/removing field, will pick those up later.

PR #1672, opened 2021-10-22
Use ListenCard component for pinned recordings

MonkeyDo:
And another episode in the saga of using the ListenCard component everywhere !

In today's installment, the pinned recordings !

I also made it so that the page doesn't refresh when you pin a recording (that's how it was implemented apparently, instead of having a callback to show the new pinned recording)

Before:
image

After:
image


akshaaatt (changes requested):
Thanks for the updates! Works good for me, only a slight mismatch when the pin recording renders without refreshing.

Screenshot from 2021-11-02 22-09-50

akshaaatt:
Hi @MonkeyDo ! Check this out.

https://user-images.githubusercontent.com/51470769/139911734-99501ac6-cc57-4bcc-876a-489bdb8772c2.mp4

amCap1712 (approved):
LGTM after latest changes. The PR is also at beta currently as I have deployed another based on this one there.

akshaaatt (approved):
Works great for me! Thanks for the updates 💯

PR #1688, opened 2021-10-28
Improve mbid mapper error reporting

amCap1712:
Passing the exception as exc_info will cause logger.error to use the exception's stacktrace. This way we should get better error grouping in sentry and more info in logs.


mayhem (approved):
That simple, I see. :)

PR #1692, opened 2021-10-29
LB docs fixes

amCap1712:
1) Fix warnings due to invalid json blocks and indentation
2) Pin docutils version because docs were not building otherwise
3) Remove references to FAQs (FAQ page was removed some time ago)


mayhem:
Thanks for documenting the color stuff!

PR #1693, opened 2021-10-29
LB-984: Reject listens with a date before October 2002

amCap1712:
In #1679, we set this to 2005 but then later discovered that in an earlier version of the codebase it was October 2002. So changing to that and documenting. The comment about exception handling is dealt in a separate PR.


mayhem (approved):

docs/dev/json.rst
```diff
@@ -143,10 +143,11 @@ A minimal payload must include
 ``artist_name`` and ``track_name`` elements must be simple strings.
 
 The payload will also include the ``listened_at`` element which must be an integer
-representing the Unix time when the track was listened to. The only exception to this
-rule is when the listen is being played right now and has been retrieved from the
-endpoint to get listens being played right now. The ``listened_at`` element will be
-absent for such listens.
+representing the Unix time when the track was listened to. The minimum acceptable
```
alastair, line 146:
```suggestion
representing the Unix time when the track was listened to. The minimum accepted
```

docs/dev/json.rst
```diff
@@ -143,10 +143,11 @@ A minimal payload must include
 ``artist_name`` and ``track_name`` elements must be simple strings.
 
 The payload will also include the ``listened_at`` element which must be an integer
```
alastair, line 145:
```suggestion
The payload should also include the ``listened_at`` element, which must be an integer
```

docs/dev/json.rst
```diff
@@ -143,10 +143,11 @@ A minimal payload must include
...
-representing the Unix time when the track was listened to. The only exception to this
-rule is when the listen is being played right now and has been retrieved from the
-endpoint to get listens being played right now. The ``listened_at`` element will be
-absent for such listens.
+representing the Unix time when the track was listened to. The minimum acceptable
+value for this field is :data:`~webserver.views.api.LISTEN_MINIMUM_TS` The only
+exception to this rule is when the listen is being played right now and has been
+retrieved from the endpoint to get listens being played right now. The
+``listened_at`` element will be absent for such listens.
```
alastair, line 150: Not sure this is important. It seems to say "now playing items won't have this field because they don't have this field".

```suggestion
value for this field is :data:`~webserver.views.api.LISTEN_MINIMUM_TS`.
playing_now requests should not have a ``listened_at`` field
```

(we already talk about playing_now at the beginning of the doc, and say that listened_at isn't required. this just reinforces it)

alastair (commented):
Just a few small comments about grammar, otherwise good.

I left another comment here about lowest timestamps in the dumps, perhaps we can fix that here too: https://github.com/metabrainz/listenbrainz-server/pull/1678#discussion_r736710307

amCap1712:
@alastair Thanks for the documentation improvement. Committed.

alastair (approved):
We decided that we'd do these other changes in another PR. Open a ticket for them, but otherwise this one is good to go!

PR #1694, opened 2021-10-29
Fix errors returned by API compat for invalid listens 

amCap1712:
API compat uses the same validate_listen method as LB API to validate listens. validate_listens used to raise APIBadRequest when validation failed. This is problematic because the API compat error format is different from that of LB. To solve this issue, introduce a ValidationError exception.

### DIFF
--- a/listenbrainz/webserver/errors.py
+++ b/listenbrainz/webserver/errors.py
@@ -1,5 +1,4 @@
 from flask import render_template, make_response, jsonify, request, has_request_context, _request_ctx_stack, current_app
-from werkzeug.exceptions import InternalServerError
 from yattag import Doc
 import yattag
 import ujson
@@ -232,3 +231,15 @@ class InvalidAPIUsage(Exception):
             with tag('error', code=self.api_error.code):
                 text(self.api_error.message)
         return '<?xml version="1.0" encoding="utf-8"?>\n' + yattag.indent(doc.getvalue())
+
+
+class ValidationError(Exception):
+    """ Error class for raising when the submitted payload does not pass validation.
+    Only use for code paths common to LB API, API compat & API compat deprecated.
+    Throw this error from an util method, capture it in each of the corresponding
+    views and re-raise the API dependent error.
+    """
+
+    def __init__(self, message, payload=None):
+        self.message = message
+        self.payload = payload
--- a/listenbrainz/webserver/errors.py
+++ b/listenbrainz/webserver/errors.py
@@ -233,7 +233,7 @@ class InvalidAPIUsage(Exception):
         return '<?xml version="1.0" encoding="utf-8"?>\n' + yattag.indent(doc.getvalue())
 
 
-class ValidationError(Exception):
+class ListenValidationError(Exception):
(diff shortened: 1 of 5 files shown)


mayhem (approved):

listenbrainz/webserver/errors.py
```diff
@@ -232,3 +231,15 @@ def to_xml(self):
...
                 text(self.api_error.message)
         return '<?xml version="1.0" encoding="utf-8"?>\n' + yattag.indent(doc.getvalue())
+
+
+class ValidationError(Exception):
+    """ Error class for raising when the submitted payload does not pass validation.
+    Only use for code paths common to LB API, API compat & API compat deprecated.
+    Throw this error from an util method, capture it in each of the corresponding
+    views and re-raise the API dependent error.
+    """
```
alastair, line 241: good documentation!
Maybe we could call this ListenValidationError (or IncomingListenValidationError? SubmittedListen?) instead, to make it a bit clearer to what we're validating.

listenbrainz/webserver/views/api_tools.py
```diff
@@ -50,14 +52,8 @@ def insert_payload(payload, user, listen_type=LISTEN_TYPE_IMPORT):
...
-    try:
-        augmented_listens = _get_augmented_listens(payload, user)
-        _send_listens_to_queue(listen_type, augmented_listens)
-    except (APIInternalServerError, APIServiceUnavailable):
-        raise
-    except Exception as e:
-        current_app.logger.error("Error while inserting payload: %s", str(e), exc_info=True)
-        raise APIInternalServerError("Something went wrong. Please try again.")
+    augmented_listens = _get_augmented_listens(payload, user)
+    _send_listens_to_queue(listen_type, augmented_listens)
```
alastair, line 56: The original version of this will give a generic error to the user, whereas the changed version will send the exact error from the exception. I know we talked about this and decided to remove the try/except, but I think that I'm in favour of hiding implementation details from the user where possible.
Maybe we should update our error handler to give a generic message (I see that [we already do this for API requests](https://github.com/metabrainz/listenbrainz-server/blob/ab61aae4481bd9f52a16301ea05617d4e963f0bf/listenbrainz/webserver/errors.py#L178-L182)). If we do this, will the original error still get caught properly by sentry?

alastair (changes requested):

listenbrainz/webserver/errors.py
amCap1712: ListenValidationError  sounds good. will rename to that.

listenbrainz/webserver/views/api_tools.py
amCap1712: I'll test the sentry issue and see if we can give the user a generic message while retaining the full info in sentry.

amCap1712: Tested and verified that the user is shown a generic error message while sentry does record the full error.

listenbrainz/webserver/errors.py
amCap1712: Done.

PR #1696, opened 2021-10-29
Improve validation of listen submission

amCap1712:
The whack-a-mole of listen validation continues,
1.  Consolidate the validation of listened_at field in one method.
2.  Add some more validation checks:
    - int(listen["listened_at"]) raises TypeError if the value is null. So TypeError needs to be caught.
    - if the track_metadata field is missing from submission json, when we later try to access it we will get a KeyError.
    - track_metadata field is present but has value null.
Also add tests for the same.


mayhem (approved):
and yet there will be more shit to validate.

alastair (approved):
looks good!

PR #1699, opened 2021-10-31
Fix user listen timestamp update :eyes:

amCap1712:
The code was using listen.user_name which means it was referring to the batch's last listen user name. However, we are in a loop and comparing ts of each listen so we probably also want to use the user_name of each listen instead.


mayhem (approved):
🤦  wow.

PR #1701, opened 2021-10-31
Add support for giving feedback on now playing listens

amCap1712:
(Another half finished PR found in stash so pushing this out as draft for now.)

It is not possible to like now playing listens currently because those do not have a msid associated with them. So add an endpoint that lets users to like/hate by artist_name, track_name and release_name.

# Todos
- [ ] Discuss whether this should be a different endpoint or add this support to the existing one
- [ ] Discuss if we should accept an additional info because this endpoint may create a new msid like submit listens endpoint
- [ ] Add tests
- [ ] Add frontend support

### DIFF
--- a/docs/dev/feedback-json.rst
+++ b/docs/dev/feedback-json.rst
@@ -20,6 +20,19 @@ A sample feedback may look like:
         "score": 1
     }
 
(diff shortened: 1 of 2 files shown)


amCap1712:
We decided to go via an alternative route.

PR #1702, opened 2021-11-01
Hide delete listen if viewing another user's profile

amCap1712:
We had this a while ago as well but probably regressed during a refactoring. Added a test to avoid future regressions.


listenbrainz/webserver/static/js/src/RecentListens.test.tsx
```diff
@@ -96,8 +96,9 @@ describe("Recentlistens", () => {
 
     timeago.ago = jest.fn().mockImplementation(() => "1 day ago");
     const wrapper = mount<RecentListens>(
-      <RecentListens {...props} />,
-      mountOptions
+      <GlobalAppContext.Provider value={mountOptions.context}>
```
amCap1712, line 99: After hiding delete listen button when isCurrentUser is false in previous commit, I noticed the updated snapshot did not have any Delete Listen buttons. Looking at the test, I figured due to not wrapping in a context, it was causing the current user to be null. 

Fix: wrap in context because having current user in the snapshot seems more useful. If it was intentional to leave out the context here, can revert the change. 

(Snapshot still had to be updated because adding currentUser causes other changes too.)

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -660,6 +660,18 @@ export default class RecentListens extends React.Component<
                       return 0;
                     })
                     .map((listen) => {
+                      const isCurrentUser =
```
amCap1712, line 663: Similar checks are present in deleteListen method to not call the API method when clicking delete listen on another user's profile. Not sure if necessary to have the same check in two places. I am leaning in favour of just having the checks here and assume in deleteListen that if the method has been called its valid to delete the listen. Thoughts?

MonkeyDo: I don't mind the belt-and-braces approach, considering they are two separate concerns (one to restrict display elements, one to restrict business logic).

My only real argument is that it could make unit testing more difficult, i.e. we couldn't test calling the deleteListen function in a test without a logged in user to ensure it doesn't call the API.
I suppose with react devtools I could manually call that method, bypassing the need for a button to be displayed, but after all all they'll get is probably an error of some sort.

MonkeyDo (approved):
Thanks for fixing this, it must have been my fuckup during a merge !

listenbrainz/webserver/static/js/src/RecentListens.tsx
amCap1712: Yup, with devtools are bets are off anyways. I see testing is indeed easier this way. Cool, let's leave it as is then.

PR #1703, opened 2021-11-01
Try out kombu for rabbitmq connection pools

amCap1712:
Using pika with our custom connection pooling logic hasn't worked out well for us till now. Kombu supports connection pools out of the box so looks like a good fit to replace both pika and our custom logic.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -50,4 +50,4 @@ pandas==1.3.0
 pyarrow==4.0.1
 more-itertools==8.8.0
 kombu==5.1.0
-librabbitmq
+amqp==5.0.6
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -49,3 +49,5 @@ google-auth==1.30.0
 pandas==1.3.0
 pyarrow==4.0.1
 more-itertools==8.8.0
+kombu==5.1.0
+amqp==5.0.6
```
amCap1712, line 53: librabbitmq is also supported instead of amqp and the docs mention that it has superior performance. However, we need gcc and other build stuff in the image to install it. Hence, not doing it just yet.

PR #1704, opened 2021-11-01
Move brainzplayer_metadata object in submitted listens

MonkeyDo:
Previously, the `brainzplayer_metadata` object was nested in `track_metadata.additional_info.brainzplayer_metadata`.
We only allow two levels of deep nesting in the listens metadata, so this `brainzplayer_metadata` object was getting flattened.
We need to move it one level up to `track_metadata.brainzplayer_metadata` in order to preserve its structure.

Before:
image

After:
image


amCap1712 (approved):
LGTM. Should we also run a script to fix up the existing listens in the DB?

MonkeyDo:
> LGTM. Should we also run a script to fix up the existing listens in the DB?

Yes please !

PR #1706, opened 2021-11-02
Bypass CAA

amCap1712:
We can construct the url for the image thumbnail at archive.org using the release mbid and caa id. Since we have both available, we do not need to query CAA url just to be redirected to the archive. This helps reduce load from CAA.


mayhem (approved):
Thanks for changing the -writer too!

PR #1708, opened 2021-11-02
LB-837: Filter out playlist owner from collaborators

MonkeyDo:
We already do this on the back-end, but confusingly allow it on the front-end, leading to LB-837, which is a minor issue but still…
Added a test for it too.
Also fixed a React warning: "Each child in a list should have a unique "key" prop."


amCap1712 (commented):
Change LGTM. But how about instead show an error that you cannot add yourself as a collaborator and not adding the name to list when Add collaborator is clicked. Thoughts?

MonkeyDo:
OK, I opted for disabling the button when a logged in user writes the name of the playlist owner (or their own name for a playlist being created)

MonkeyDo:
Also updated the Playlist*s* page (I only did the Playlist page previously) with the previous filtering of playlist owner.

amCap1712 (approved):
LGTM. Thanks

PR #1709, opened 2021-11-02
LB-863: Empty playlist creation modal on success

MonkeyDo:
Small ticket !
This is basically a "reset" of the modal used to create new playlists, otherwise the details of the newly created playlist stay in the modal when you try to create another one.


akshaaatt (approved):
Thanks for fixing this, works great for me! :100:

PR #1713, opened 2021-11-04
User feedback improvements

MonkeyDo:
On the RecentListens page, user feedback was only loaded at component startup and not after each navigation ! 😱 
That means that after navigating from page 1 to page 2 of my listens, my user feedback wouldn't be displayed.
If I reloaded the page (with URL params, still page 2), then feedback is shown.

I also fixed a couple of issues with the UserFeedback page, where again feedback failed to load in some circumstances.
We already have feedback score in the page props so I made sure we used that and save API calls when we already have the info.


amCap1712 (approved):

listenbrainz/webserver/static/js/src/RecentListens.tsx
```diff
@@ -391,7 +391,7 @@ export default class RecentListens extends React.Component<
         }
       });
       try {
-        const data = await this.APIService.getFeedbackForUserForRecordings(
```
akshaaatt, line 394: What was the issue with using this implementation? We are instantiatng the APIService again just to avoid this?

MonkeyDo: We're not instantiating here,  it's already instantiated and available in the global context.
I'm slowly cleaning up all components to remove `this.APIService = APIService` and use the API from context directly, it's a bit cleaner and clearer.

MonkeyDo: The reason we still have `this.APIService = APIService` in some components is just leftover from how we previously did it; I wanted to minimize the number of changes when I did the global context refactor.

akshaaatt: Right, makes sense!

PR #1715, opened 2021-11-05
MessyBrainz Cleanup Part 3 - Move messybrainz inside listenbrainz

amCap1712:
Builds on #1714. Move messybrainz db components to listenbrainz.messybrainz so that we can reuse the configuration from listenbrainz.config. There is still some redundant code and opportunity to better arrange things. I intend to leave those for a later day.


mayhem (approved):
I still have conflicted feelings about a clean messybrainz.

PR #1716, opened 2021-11-05
MessyBrainz Cleanup 4 - Delete remaining unused parts of MsB DB

amCap1712:
This PR deletes more unused parts of MsB DB. I separated this one because it also contains script to delete these tables from prod.


mayhem (approved):
Thanks!

PR #1718, opened 2021-11-06
Use mapped mbid data in Pinned Recordings

amCap1712:
The PR modifies pinned recordings backend to use data from listen_mbid_mapping table when a mbid is available. This change has some implications. The data in listen_mbid_mapping table is populated from MB and not user submitted track/artist/release names whereas the existing practice is to use the data from recording_json table which has user submitted data. IMO, this is a step in the right direction but let's discuss it before finalizing.

For some context, the track_metadata from user submitted listens is used to populate the recording_json. So using data from listen table or recording_json table should be mostly the same. The data in listen_mbid_mapping is on the other hand filled from MB by mbid mapping writer.

We have quite a few combinations regarding use of this data in LB, I am describing all the ones that come to mind here. 

- Listens Page (Prefer user submitted over mapped) : Listens have user submitted data stored in track_metadata. When sending to the frontend, we add a mbid_mapping dict to this which contains the mapped mbids. Note, here we only send mapped mbids to the frontend. The mapped artist/track/release name are not used here. The frontend uses user submitted mbids if available otherwise tries to fallback on mapped mbids. The mapped track/artist/release names never displayed to the user.

- Stats (Prefer mapped over user submitted): Stats are generated from spark dumps. While doing spark dumps, we prefer mapped data over user submitted data. We use mapped names if available otherwise user submitted names. Only mapped mbids are used. User submitted mbids are never used.

- Feedback: Use user submitted names always. Add mapped mbids if available.

- Pinned Recordings Existing: Use user submitted data mbid data from recording_json table. If mbid not available, use msids to get data from recording_json.

- Pinned Recordings New: Use mapped mbid data if available including mapped names. If no mapping available, use user submitted data from recording_json using msids.

We should evaluate each use case and decide on what to use accordingly. I think feedback and pinned recording are similar enough that we should use the same way for these two at least.

### DIFF
--- a/listenbrainz/db/model/pinned_recording.py
+++ b/listenbrainz/db/model/pinned_recording.py
@@ -108,19 +108,19 @@ def fetch_track_metadata_for_pins(pins: List[PinnedRecording]) -> List[PinnedRec
         with timescale.engine.connect() as connection:
             mbid_metadatas = connection.execute(sqlalchemy.text(query), mbids=mbids)
 
-        # we can zip the pins and metadata because the query returns
-        # the metadata in same order of the mbid list passed to it
-        for pin, metadata in zip(mbid_pins, mbid_metadatas):
-            pin.track_metadata = {
-                "track_name": metadata["title"],
-                "artist_name": metadata["artist"],
-                "release_name": metadata["release"],
-                "additional_info": {
-                    "recording_mbid": metadata["recording_mbid"],
-                    "release_mbid": metadata["release_mbid"],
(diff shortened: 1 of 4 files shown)


amCap1712:
According to the discussion in IRC, we decided to use only mapped data in statistics, user submitted data and falling back to mapped data for listens, and prefer mapped data falling back to user submitted  data for everything else.

mayhem (approved):
I love how things are improving now that we have a mapping!

PR #1719, opened 2021-11-06
Fix pinned recording link color

MonkeyDo:
With #1672 merged and deployed in beta, I realized there's an issue with the link color on a currently pinned card. The links are the same blue as the card background.

I changed that to inherit the color of text from the parent.

I also modified the listen-card (in all its forms) to add underline when hovering links.
This is the default browser behavior which is helpful in making it clearer to users that those are links.

Before (links barely visible on hover, invisible without hovering):
image

After:
image


amCap1712 (approved):
Thanks!

PR #1720, opened 2021-11-08
Make new URL(… instantiation safe

MonkeyDo:
Getting some errors on the front-end: "Failed to construct 'URL': Invalid URL"
image

After a quick debug, it turns out some metadata is dirty (no way!) and a youtube video id ended up in `origin_url`.
Calling `new URL(x)` on that throws an error, which I didn't guard for.


amCap1712 (approved):
LGTM. Seeing this block of code getting repeated multiple times, a future enhancement can probably be refactoring it in a separate function,

MonkeyDo:
> LGTM. Seeing this block of code getting repeated multiple times, a future enhancement can probably be refactoring it in a separate function,

Yep, I actually did this just today for another PR :)

PR #1721, opened 2021-11-08
LB-999: Safer API response header access

MonkeyDo:
Saw this issue in sentry for multiple users:
`TypeError: Cannot read properties of undefined (reading 'get')`
The offending lines are
https://github.com/metabrainz/listenbrainz-server/blob/959f58d15b9e23105c26fe693918ecd97074883f/listenbrainz/webserver/static/js/src/LastFMImporter.tsx#L522-L523

We don't know for sure if the request will have headers, so we can't assume they are there.


alastair:
it's strange that these headers might not be here, since we add them to every outgoing API request. Were you seeing bugs reported due to them not being there?

MonkeyDo:
Created LB-1000 to track the other side of the issue (missing response headers)

PR #1722, opened 2021-11-09
Fix invalid order of ' and %

amCap1712:
Found stack trace in sentry.

### DIFF
--- a/listenbrainz/spark/spark_reader.py
+++ b/listenbrainz/spark/spark_reader.py
@@ -87,7 +87,7 @@ class SparkReader:
         try:
             response_handler(response)
         except Exception as e:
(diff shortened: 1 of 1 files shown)


listenbrainz/spark/spark_reader.py
```diff
@@ -87,7 +87,7 @@ def process_response(self, response):
         try:
             response_handler(response)
         except Exception as e:
-            current_app.logger.error('Error in the response handler: %s, data: %s %'
+            current_app.logger.error('Error in the response handler: %s, data: %s' %
```
alastair, line 90: remember that logger methods will do string interpolation automatically anyway, so you should be able to do `logger.error('message %s', var, exc_info=True)`

alastair (approved):

PR #1723, opened 2021-11-09
LB-992: Too many LB full dumps on FTP site

amCap1712:
Remove * from end of $SOURCE_DIR otherwise rsync --delete won't work as expected.


mayhem (approved):
Oh rsync and your cryptic use. I could totally give you up...

PR #1724, opened 2021-11-09
Create public (mbid mapping) and private (playlists) timescale tables dump

alastair:
# Problem

We want to make dump files of non-listens data in the timescale database:

- playlists
- mbid mapping

# Solution

The dump script is currently only designed to dump from the main postgres database, so in the interests of speed, I just copy/pasted a few things to make the db engine a configurable option.

Dump 2 extra files (private-timescale and public-timescale) that include the relevant public/private split of this database.

# Action

Still work in progress, things to fix:
- [x] Dump fails with `Expected 12 dump files, found 15. Aborting.`, update check
- [x] Also make public dump (mbid mapping)
- [x] Double-check that when this runs that we won't upload the private playlists dump to FTP
- [x] Tests need to be fixed, they'll be failing
- [x] Add import part, double-check that tables are dumped in the right order to fulfill fk requirements when importing to a db with fks set up
- [x] Sort out various schema ids. In discussion in IRC we thought that we should probably have 1 for pg db, 1 for ts db, 1 for listen dumps
- [x] Update SCHEMA_VERSION_CORE, related to https://github.com/metabrainz/listenbrainz-server/pull/1422 - wrong value was updated, so we should increase the db version as well.
- [x] Identify problem with dumping timescale public dump tables out of order
- [x] ~~Ideally deduplicate code by using https://github.com/metabrainz/brainzutils-python/pull/50/files~~ later

### DIFF
--- a/admin/create-dumps.sh
+++ b/admin/create-dumps.sh
@@ -190,6 +190,9 @@ add_rsync_include_rule \
     "listenbrainz-public-dump-$DUMP_TIMESTAMP.tar.xz"
 add_rsync_include_rule \
     "$FTP_CURRENT_DUMP_DIR" \
+    "listenbrainz-public-timescale-dump-$DUMP_TIMESTAMP.tar.xz"
+add_rsync_include_rule \
+    "$FTP_CURRENT_DUMP_DIR" \
     "listenbrainz-listens-dump-$DUMP_ID-$DUMP_TIMESTAMP-$DUMP_TYPE.tar.xz"
 add_rsync_include_rule \
     "$FTP_CURRENT_DUMP_DIR" \
--- a/listenbrainz/db/__init__.py
+++ b/listenbrainz/db/__init__.py
@@ -5,8 +5,12 @@ from sqlalchemy.pool import NullPool
 import time
 import psycopg2
 
-# This value must be incremented after schema changes on replicated tables!
-SCHEMA_VERSION = 5
+# The schema version of the core database. This includes data in the "user" database
+# (tables created from ./admin/sql/create-tables.sql) and includes user data,
+# statistics, feedback, and results of user interaction on the site.
+# This value must be incremented after schema changes on tables that are included in the
+# public dump
+SCHEMA_VERSION_CORE = 5
 
 engine = None
 
--- a/listenbrainz/db/__init__.py
+++ b/listenbrainz/db/__init__.py
@@ -10,7 +10,7 @@ import psycopg2
 # statistics, feedback, and results of user interaction on the site.
 # This value must be incremented after schema changes on tables that are included in the
 # public dump
-SCHEMA_VERSION_CORE = 5
+SCHEMA_VERSION_CORE = 6
 
 engine = None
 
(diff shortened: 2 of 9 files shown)


alastair:
I've updated this with imports and fixing schema numbers.
For now I've left the _CORE and _TIMESCALE schemas different, because I want to do a test where the schema id differs in the two types of dumps to check that the comparison works correctly.

I'm still running into a problem with the public ts dumps (msid mapping tables). In the [dump file, I list the mapping table before the join table](https://github.com/metabrainz/listenbrainz-server/pull/1724/files#diff-e4052e0a798cedeb7deb0436bdf9c0e0fb0237895de21f76a7cd15751c049f5eR114-R128) which means that it should appear first in the archive (dictionary iteration over keys should be stable in the order that they're defined), but the archive is created incorrectly:
```
$ tar tf listenbrainz-public-timescale-dump-20211109-192156.tar.xz
listenbrainz-public-timescale-dump-20211109-192156/SCHEMA_SEQUENCE
listenbrainz-public-timescale-dump-20211109-192156/TIMESTAMP
listenbrainz-public-timescale-dump-20211109-192156/COPYING
listenbrainz-public-timescale-dump-20211109-192156/lbdump/
listenbrainz-public-timescale-dump-20211109-192156/lbdump/listen_join_listen_mbid_mapping
listenbrainz-public-timescale-dump-20211109-192156/lbdump/listen_mbid_mapping
```
with the join table first in the file. This means that when we try to import it, it fails because it's missing the fk to items in the mapping table which hasn't been imported yet. I'm not sure why this is happening.
I updated the todo list in the first comment.

amCap1712:
@alastair even for the current public dumps, I see the following output:
```
tar tf listenbrainz-public-dump-20211101-040003.tar.xz
listenbrainz-public-dump-20211101-040003/SCHEMA_SEQUENCE
listenbrainz-public-dump-20211101-040003/TIMESTAMP
listenbrainz-public-dump-20211101-040003/COPYING
listenbrainz-public-dump-20211101-040003/lbdump/
listenbrainz-public-dump-20211101-040003/lbdump/"user"
listenbrainz-public-dump-20211101-040003/lbdump/recording_feedback
listenbrainz-public-dump-20211101-040003/lbdump/statistics.artist
listenbrainz-public-dump-20211101-040003/lbdump/statistics.recording
listenbrainz-public-dump-20211101-040003/lbdump/statistics.release
listenbrainz-public-dump-20211101-040003/lbdump/statistics.user
```

This is different from the order we dump in. Looking up on internet, I did not find definitive insights into what order tar files are decompressed.

alastair:
that's interesting - tar files should uncompressed in the order that the files are added to it. I just tested dumps here locally and we are in fact iterating through the dictionary keys in the order that they're defined, so I'm unsure as to why we have them in a different order here. I'll look into it again

mayhem (approved):
schema number definitions notwithstanding, looks good to me.

amCap1712 (approved):
LGTM

alastair:
I opened LB-1009 to improve our schema numbering

PR #1725, opened 2021-11-09
LB-904: Best effort to show cover art in ListenCard

MonkeyDo:
As LB-904 suggests, we should show some cover art in the ListenCard components.
This initial PR is a suggested way to do just that with limited resources.

First of all, if the listen has a Spotify ID in its metadata and the user has a Spotify token (required to call the API), we can get the cover art from Spotify directly.

If the listen has youtube metadata, here again we can derive the cover art URL from the youtube ID. No need to call an API.

And finally, if neither of these options are true and the listen has a release MBID (in metadata, originally or via the MBID mapper), we can query the Cover Art Archive API.

**We will need to reimplement this last scenario**. This particular release might not have cover art available, while the associated Release Group could have one, or even one of the other Releases in the same RG would be better that nothing.
We've discussed this and figure we should do that processing on the LB back-end since it has access to the MB database. It will be easier  than call the CAA and MB APIs a hundred times.

Best case scenario:
image

When images are missing we display a placeholder:
image

Also works with "compact" mode:
image


listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -251,7 +273,16 @@ export default class ListenCard extends React.Component<
           isCurrentlyPlaying ? "current-listen" : ""
         }${compact ? " compact" : " "} ${className || ""}`}
       >
-        {thumbnail && <div className="listen-thumbnail">{thumbnail}</div>}
+        {thumbnail || (
+          <div className={`listen-thumbnail ${!thumbnailSrc ? "empty" : ""}`}>
```
MonkeyDo, line 277: If no thumbnail is passed in props, we create the listen-thumbnail initially with an .empty css class. Once the image src is fetched the class is removed, and we get a bit of a transition.
Otherwise there's a jarring jump to the right of the track name when the  listen-thumbnail element is added.

MonkeyDo:
In case you notice this issue, I'm still doing some refactoring of the ListenCard design to solve the issue below (thumbnail height is fixed, ListenCard height isn't), but I'm doing it in a separate branch and will open a separate PR:
image

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -93,6 +92,31 @@ type BrainzPlayerState = {
...
+  if (!listen || !datasource) {
+    return undefined;
+  }
+  if (datasource instanceof SpotifyPlayer) {
+    return SpotifyPlayer.isListenFromThisService(listen);
+  }
+  if (datasource instanceof YoutubePlayer) {
+    return YoutubePlayer.isListenFromThisService(listen);
+  }
+  if (datasource instanceof SoundcloudPlayer) {
```
akshaaatt, line 114: We can directly return without explicitly checking here since we have already confirmed with the first `if` statement about the `datasource` being one of the three or not.

listenbrainz/webserver/static/js/src/SpotifyPlayer.tsx
```diff
@@ -106,16 +117,23 @@ export default class SpotifyPlayer
...
-    const spotifyTrack = spotifyId.split("https://open.spotify.com/")[1];
+    const spotifyTrack = spotifyId.split(
+      "https://open.spotify.com/track/"
+    )?.[1];
+    return spotifyTrack;
+  }
+
+  static getSpotifyUriFromListen(listen: Listen | JSPFTrack): string {
+    const spotifyTrack = SpotifyPlayer.getSpotifyTrackIDFromListen(listen);
     if (typeof spotifyTrack !== "string") {
```
akshaaatt, line 133: I think we can remove this now.

listenbrainz/webserver/static/js/src/YoutubePlayer.tsx
```diff
@@ -96,6 +105,20 @@ export default class YoutubePlayer
...
   }
 
+  public name = "youtube";
+  public domainName = "youtube.com";
+  youtubePlayer?: ExtendedYoutubePlayer;
+  checkVideoLoadedTimerId?: NodeJS.Timeout;
+
+  componentDidUpdate(prevProps: DataSourceProps) {
+    const { show } = this.props;
+    if (prevProps.show === true && show === false && this.youtubePlayer) {
```
akshaaatt, line 115: This check can be simplified to `prevProps.show && !show && this.youtubePlayer`

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
akshaaatt: Sounds good!

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -415,6 +421,53 @@ const pinnedRecordingToListen = (pinnedRecording: PinnedRecording): Listen => {
...
+      if (CAAResponse.ok) {
+        const body: CoverArtArchiveResponse = await CAAResponse.json();
+        if (!body.images?.[0]?.thumbnails) {
+          return undefined;
+        }
+        const { thumbnails } = body.images[0];
+        return (
+          thumbnails[250] ??
+          thumbnails.small ??
+          // If neither of the above exists, return the fisrt one we find
```
akshaaatt, line 459: A small typo here of `first`

listenbrainz/webserver/static/js/src/utils.tsx
```diff
@@ -415,6 +421,53 @@ const pinnedRecordingToListen = (pinnedRecording: PinnedRecording): Listen => {
...
+          // @ts-ignore
+          thumbnails[Object.keys(thumbnails)?.[0]]
+        );
+      }
+    } catch (error) {
+      // Do nothing with the error, I guess.
```
akshaaatt, line 465: I think we can log these errors here?  Or keep a record of these somewhere maybe?

akshaaatt (approved):
Thank you for the work here and really exciting about this long-due feature! :100: 
I have added a few comments about some parts of the code which are a very low priority for the fix but did need a mention.

listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
MonkeyDo: That's how it was initially, but unfortunately there is an issue with typescript that prevents us from doing it that way.

I had to make `isListenFromThisService` a static method in order to be able to use it in `getAlbumArtFromListenMetadata` in utils.tsx, where I don't have access to instances of these datasource classes.

Typescript freaks out when I try `datasource.isListenFromThisService(listen)` since it is not aware of static methods on these classes without referring to a class directly (`YoutubePlayer.isListenFromThisService(…)` works but `datasource.isListenFromThisService(…)` doesn't, and same goes for `YoutubePlayer.prototype.isListenFromThisService(…)` and `YoutubePlayer.constructor.isListenFromThisService(…)`)

It is a know TS issue which I left in a comment to above this method for posterity: https://github.com/Microsoft/TypeScript/issues/3841#issuecomment-337560146

PR #1726, opened 2021-11-10
Missing MusicBrainz Data Page

hrik2001:
# Problem
Displaying all the missing MusicBrainz data that spark has found till now, and adding an option to make users submit those data in MusicBrainz

# Solution
Made a public URL for each user where you can see each user's missing data and submit them. Also I have altered `ListenCard` quite a bit and need @MonkeyDo's opinion on it, since he is the one working on the whole `ListenCard` component revamp.

# Action
Need opinions on these things
- Spark outputs timestamp as `"2021-11-09 21:53:13"`. This is not any particular standard from what I know of. I have added a Z at the end to show that it is GMT, hence `"2021-11-09 21:53:13Z"`. But since this is not a standard thing, it may break things in older browsers. (Its generally advised that Date object should not parse the timestamp string unless it's a certain popular ISO Standard)
- Double click won't add the data, the user has to click the plus, I took this decision as an UX point of view, since in listen card, double click will mean play, but might not necessarily mean the same here in this case, since adding data is more effortful than listening.
- Don't have any tests for the link that serves the template , as the test will be merely a `self.assert200(response)` test. But I won't mind writing it.


listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -307,18 +347,30 @@ export default class ListenCard extends React.Component<
...
+          {isMissingData ? (
+            <button
+              title="Add to MusicBrainz"
+              className="btn-transparent"
+              onClick={this.submitMissingData}
+              type="button"
+            >
+              <FontAwesomeIcon size="2x" icon={faPlus as IconProp} />
+            </button>
+          ) : null}
```
MonkeyDo, line 373: Here's a proposition for another way to do this:

As you mentioned, I recently did a lot of refactoring of the ListenCard to use it everywhere, and found myself needing to be able to extend the functionality.
Instead of implementing each functionality directly in ListenCard (which would become very messy) I opted for passing some menu options as props (`additionalMenuItems`) to the ListenCard component, for any non-generic functionality.

You'll be able to move the `submitMissingData` method to the MissingMBData component and pass a menu item as prop to the ListenCard components.
You can see an example of this here: https://github.com/metabrainz/listenbrainz-server/blob/43d4c9df9cbeef480670065204859292961f70f7/listenbrainz/webserver/static/js/src/playlists/PlaylistItemCard.tsx#L64-L87

You'll probably need to bing the method like this `action={this.submitMissingData.bind(this, data)}` to ensure calling submitMissingData with the correct piece of data.
Similarly as what you did with the button, you can pass `icon={faPlus}` and the title as well.

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -307,18 +347,30 @@ export default class ListenCard extends React.Component<
...
-            onClick={this.playListen}
-            type="button"
-          >
-            {isCurrentlyPlaying ? (
-              <FontAwesomeIcon size="1x" icon={faPlay as IconProp} />
-            ) : (
-              <FontAwesomeIcon size="2x" icon={faPlayCircle as IconProp} />
-            )}
-          </button>
+          {disablePlay ? null : (
```
MonkeyDo, line 350: This is a potentially useful feature, but I have 2 comments:
1. You also need to check disablePlay on line 276 (`onDoubleClick={this.playListen}`)
2. Why don't we want this page to be playable?

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -91,6 +96,40 @@ export default class ListenCard extends React.Component<
     );
   };
 
+  submitMissingData = () => {
+    const { listen } = this.props;
+    const form = document.createElement("form");
```
MonkeyDo, line 101: This seems to me like a very odd way to go about this.

All this should be achievable with the `fetch` API ([docs here](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch)), which will take a javascript object as input (`{name: listen.track_metadata.release_name || "", …}`) 

There's a decent example code with `method: 'POST'` you can use as a base in that page (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#supplying_request_options)

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -50,6 +51,10 @@ export type ListenCardProps = {
   feedbackComponent?: JSX.Element;
   // These go in the dropdown menu
   additionalMenuItems?: JSX.Element;
+  // Mode for displaying missing MusicBrainz data
+  isMissingData?: boolean;
```
MonkeyDo, line 55: As explained in more detail below, we won't need this prop.

MonkeyDo (changes requested):
Good start :)

I've tried to explain clearly how the ListenCard can be extended in a composable manner, let me know if anything is unclear or if you have questions !

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
hrik2001: hey, so I wanted to discuss this too, since it indeed is a very odd way to do this, but I assure you, I did try fetch before opting for this obscure method

I had made myself an HTML form that would populate the musicbrainz form, my purpose was to submit it, record the submission in my browser, and copy it as a fetch request so I could use it in my code. But that didn't seem to work out (never redirected). I even tried to manually write the fetch but it **never redirected**. I played around with the redirect field of the object passed to the fetch, and it still didn't work.

I will look into it more, and update you on this. Thank you so much!

MonkeyDo: I recant myself!
As we discussed on IRC, this is actually the right method to achieve the result we are looking for.

The HTML content returned with the request response cannot be opened by the browser when using the fetch API. Booo !

Creating a hidden form and submitting it programmatically is an elegant solution in comparison to pretty much all the other options I thought of, and also less risky (this default browser function works with CORS, cookies, redirects, etc. ; not sure whatever hacky workaround we come up with will…).


I would just suggest to add a `redirect_uri` (as per [the docs](https://musicbrainz.org/doc/Development/Release_Editor_Seeding), bottom of the page) to redirect to LB on successful MB form submission.
I guess `redirect_uri = window.location.href` ?

And also to add an `edit_note` textarea element, something along the lines of  "Seeded by ListenBrainz missing data page: ${window.location.href}"

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
MonkeyDo: Ah, maybe instead of using `redirect_uri` you can set the from's `target` property to "_blank" , that should open it in a new tab. Haven't tried that myself so take with a pinch of salt :)

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
MonkeyDo: Ah, maybe instead of using redirect_uri you can set the form element's target property to "_blank" , that should open it in a new tab. Haven't tried that myself so take with a pinch of salt :)

MonkeyDo: I also think we should add a comment at the top to explain why we're doing it this way.
Maybe something along the lines of `The POST request returns an html page, which we can't do anything with in fetch. Instead we need to use a native html form, which once submitted will redirect to the MusicBrainz page appropriately.`?

hrik2001: Hi, sorry for replying late, I hadn't opened GitHub in the weekend. But I hear you, and I will implement it accordingly. 

>it programmatically is an elegant solution in comparison to pretty much all the other options I thought of, and also less risky

ikr! And I tried to think of potential alternatives, and one idea struck me, that was to create modal and then putting an iframe in there. But I realised that the serious music submitters will go back and forth into checking the MB database so they would need the back and front button of the browser, so the redirection using form seems like the best method till now. Which is why I didn't raise this idea in IRC.


You are also right about adding a comment to describe the reasoning behind the weird solution.

I will also try the `_blank` solution and let you know if it works or not.

> Add an `edit_note` textarea element     

That sounds quite neat, because suppose the user knows that some data is wrong, the person can just fill the form from scratch. (Seems like that was your intention). So I will add that. Thank you so much @MonkeyDo . Once again, really sorry for the late reply :D

MonkeyDo: To be clear, the `edit_note` I was thinking of would just be added to the form to identify that the data was auto-filled by ListenBrainz, the same way other auto-importers do it* and wouldn't be shown to the user on LB.
They can always add to it in the MB form if they wish.

* See for example this edit note: https://musicbrainz.org/edit/84600395#note-84600395-1

hrik2001: Ah that's helpful, this makes it seem more legitimate. I am gonna add this. Thanks!

listenbrainz/webserver/static/js/src/MissingMBData.tsx
```diff
@@ -0,0 +1,333 @@
...
+                    top: "50%",
+                    zIndex: 1,
+                  }}
+                >
+                  <Loader isLoading={loading} />
+                </div>
+                <div>
+                  {missingData.map((data) => {
+                    return (
+                      <div className="event-content" style={{ width: "100%" }}>
```
MonkeyDo, line 187: Since you are using a css class here, it's preferable to define styles for that class in  CSS (.less files) instead of inlined, which becomes hard to maintain pretty quickly.

Is this parent element needed by the way? The ListenCard should take up 100% of the parent's width.

MonkeyDo (changes requested):
Thanks for the modifications, you're heading in the exact right direction and almost at destination :)

hrik2001:
Last time I forgot to ask you one doubt.
So react complains (throws a warning) about how I have camel cased `additionalActions`, and how I shouldn't do that for a prop. But it never complains about `additionalMenuItems` or similar camelcased prop names. (which is why customicons was not camel cased) What voodoo did you pull off to tame react haha😅? I am sure I am missing something that is quite easy.

Apart from that, I have done done commits as you asked. Thanks again for such a detailed review @MonkeyDo.

MonkeyDo (changes requested):
Thanks again for the changes, and  sorry for being a stickler with ell of this.
I think we're really close now, the page  looking fine !

listenbrainz/webserver/templates/user/missing_data.html
```diff
@@ -0,0 +1,25 @@
...
+
+{% block title %}Missing MusicBrainz Data of "{{ user.musicbrainz_id }}" - ListenBrainz{% endblock %}
+
+{% block content %}
+  <h2 class="page-title">Missing MusicBrainz Data of {{ user.musicbrainz_id }}</h2>
+  {% if current_user.is_authenticated and current_user == user %}
+
+    <h3>Directly export missing MusicBrainz data</h3>
+      <p>Share a hand to the community effort of gathering and marking all known music in existence! Such small submissions help
+         us grow more :D.
```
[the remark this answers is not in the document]
hrik2001: That is a good idea

listenbrainz/webserver/static/js/src/MissingMBData.tsx
```diff
@@ -0,0 +1,338 @@
...
+                    user_name: user.name,
+                    track_metadata: {
+                      artist_name: data.artist_name,
+                      track_name: data.recording_name,
+                      release_name: data?.release_name,
+                    },
+                  };
+                  const additionalActions = (
+                    <>
+                      <ListenControl
```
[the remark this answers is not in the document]
hrik2001: Yes! Indeed looks better!

listenbrainz/webserver/static/js/src/MissingMBData.tsx
```diff
@@ -0,0 +1,338 @@
...
+                    top: "50%",
+                    zIndex: 1,
+                  }}
+                >
+                  <Loader isLoading={loading} />
+                </div>
+                {missingData.map((data) => {
+                  const listen = {
+                    listened_at:
+                      new Date(`${data.listened_at}Z`).getTime() / 1000,
```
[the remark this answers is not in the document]
hrik2001: Spark gives off timestamp as GMT. So I added a Z at the end to make the browser understand that. Or else it would automatically assume the timestamp to be of local time zone. I infact read up a bit and realised that Date constructor parsing non standard timestamp string should be kind of avoided for consistency reasons (across different browsers). So I think we should discuss this in IRC.

listenbrainz/webserver/static/js/src/MissingMBData.tsx
```diff
@@ -0,0 +1,338 @@
...
+                    user_name: user.name,
+                    track_metadata: {
+                      artist_name: data.artist_name,
+                      track_name: data.recording_name,
+                      release_name: data?.release_name,
+                    },
+                  };
```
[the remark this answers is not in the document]
hrik2001: Ahh fair point, a very embarrassingly silly mistake from my side. Added the brainzplayer in later commits and somehow missed the duplication part :P.

listenbrainz/webserver/static/js/src/MissingMBData.tsx
```diff
@@ -0,0 +1,338 @@
...
+            style={{
+              position: "-webkit-sticky",
+              // @ts-ignore
+              position: "sticky",
+              top: 20,
+            }}
+            /* eslint-enable no-dupe-keys */
+          >
+            <BrainzPlayer
+              direction={direction}
```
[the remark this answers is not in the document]
hrik2001: Will it be okay if this line is removed during the tackling of merge conflict? Since changing this rn here at my branch will cause trouble to test the page locally.

listenbrainz/webserver/templates/user/missing_data.html
```diff
@@ -0,0 +1,25 @@
...
+      </p>
+      <div class="row">
+      <div class="well col-sm-8 col-sm-offset-2 col-md-6 col-md-offset-3">
+        <div id="react-container">
+        </div>
+      </div>
+      </div>
```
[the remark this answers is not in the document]
hrik2001: Getting rid of the first div (that has row class), makes the footer glitch. The second div (that has well) makes the page non responsive, but I have removed the well class. Hope that works. Hence I _have not removed any div_.

(If there are any other methods to get the same effect, while making the codebase better, please do let me know, since I am not as adept in frontend as my backend, so feel free to drop advice and to be pedantic, as it will help me get better.)

listenbrainz/webserver/static/js/src/MissingMBData.tsx
```diff
@@ -0,0 +1,338 @@
...
+                      artist_name: data.artist_name,
+                      track_name: data.recording_name,
+                      release_name: data?.release_name,
+                    },
+                  };
+                  const additionalActions = (
+                    <>
+                      <ListenControl
+                        icon={faPlus}
+                        title=""
```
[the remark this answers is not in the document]
hrik2001: Adds a string in white, by the plus, are you sure it would look good? I can put it in if you want.

listenbrainz/webserver/static/js/src/MissingMBData.tsx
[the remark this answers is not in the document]
hrik2001: `btn btn-sm` makes the + non visible. Hence I put `btn-primary btn-success`, but there is a weird border around it

hrik2001:
@MonkeyDo while checking, you might need to add the data again, since now the timestamps come with Z. Though there won't be any significant difference.

@amCap1712 , I would be grateful if you check the last commit (67e5779) (changes in `listenbrainz/spark/handlers.py` and `listenbrainz/spark/test_handlers.py` basically),   so that the addition of Z doesn't break anything (since **I can't check this in spark locally**, hence kinda scared :P)

MonkeyDo:
I would suggest two main things:

1. Reorganizing the page structure:

I tried with this configuration which is working well for me:
- Remove all divs around the one with `react-container` id
- Change the css classes of the containers for the list and the brainzplayer
image

The result: 

Small screens: 
image

Larger screens:
image


2. The CSS classes for the buttons should be `btn btn-sm btn-success`:
image

hrik2001:
hello, I had reported earlier that `btn btn-sm btn-success` leads to the button being very weird
but I realised that it *works when playing around in component editor of browser*
but when one writes this class in react, the buttons looks like this
image
further looking into the issue shows this
`Buttons must have discernable text`.
So I thought of adding "h" to test. Well that approach did leave the warning at bay but didn't help

hrik2001:
image

MonkeyDo:
So there's definitely an issue with the `btn-…` css classes being applied to the <button> parent as well as the FAicon inside it.
I think I know what the fix should be, care to implement it?

We need two separate props to replace the `className` prop in the ListenControl component: `buttonClassName` and `iconClassName`, both optional, each one getting applied to the correct element.

This change should only affect one other component: ListenFeedbackComponent which is the only one using a ListenControl with `className` prop: https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/listens/ListenFeedbackComponent.tsx#L74

So in ListenFeedbackComponent the `className` should be changed to use the new `iconClassName` prop (since we also use the `iconOnly` prop to only render the icon without a <button>  parent).

As far as I can see in the codebase, this is the only component using ListenControl with the `className` prop (which is also why this issue didn't get caught before :) ) so no other changes would be needed.

You'll also need to regenerate test snapshots, since the prop names will have changed.

listenbrainz/webserver/static/js/src/MissingMBData.tsx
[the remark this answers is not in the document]
MonkeyDo: No, my mistake !

MonkeyDo:
I took the liberty to remove the commented-out direction prop, and to fix a failing test (it was nothing big but the way i wrote that test is hard to read IMO, needed some trial and error before I got to the simple change needed).

Thanks a lot for all your work !
I'll wait until tests pass and we're ready to merge ! :rocket:

MonkeyDo (approved):

hrik2001:
Oh shoot, looks like I forgot to uncomment the direction prop @MonkeyDo, should have done it.
Other than that, one of the test failed for missing mb data api. I have forgotten to update the tests for the new timestamps with a Z, so I would do that asap. Thank you so much for being patient.

hrik2001:
finally all checks passed! wew, I got scared for a moment a while ago :D

MonkeyDo:
Off we go ! :rocket:

PR #1729, opened 2021-11-13
Use kombu for rabbitmq connections in websockets

amCap1712:
We intend to replace pika with kombu in LB because kombu is easier to work with in general. To iron out issues, we begin with moving websockets off to kombu because those are a non-essential part of the infra.

### DIFF
--- a/listenbrainz/websockets/listens_dispatcher.py
+++ b/listenbrainz/websockets/listens_dispatcher.py
@@ -6,6 +6,9 @@ from kombu.mixins import ConsumerMixin
 from listenbrainz.utils import get_fallback_connection_name
 
 from kombu import Connection, Exchange, Queue, Consumer
+from kombu.utils.debug import setup_logging
+
+setup_logging()
 
 
 class ListensDispatcher(ConsumerMixin):
--- a/listenbrainz/websockets/listens_dispatcher.py
+++ b/listenbrainz/websockets/listens_dispatcher.py
@@ -17,8 +17,8 @@ class ListensDispatcher(ConsumerMixin):
         self.connection = None
         self.channel2 = None
 
-        self.unique_exchange = Exchange(app.config["UNIQUE_EXCHANGE"], "fanout")
-        self.playing_now_exchange = Exchange(app.config["PLAYING_NOW_EXCHANGE"], "fanout")
+        self.unique_exchange = Exchange(app.config["UNIQUE_EXCHANGE"], "fanout", durable=False)
+        self.playing_now_exchange = Exchange(app.config["PLAYING_NOW_EXCHANGE"], "fanout", durable=False)
         self.websockets_queue = Queue(app.config["WEBSOCKETS_QUEUE"], exchange=self.unique_exchange, durable=True)
         self.playing_now_queue = Queue(app.config["PLAYING_NOW_QUEUE"], exchange=self.playing_now_exchange, durable=True)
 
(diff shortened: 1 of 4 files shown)


listenbrainz/websockets/listens_dispatcher.py
```diff
@@ -1,81 +1,70 @@
...
-        else:
-            event_name = 'listen'
-        for listen in listens:
-            self.socketio.emit(event_name, json.dumps(listen), to=listen['user_name'])
-
-    def callback_listen(self, channel, method, properties, body):
-        listens = json.loads(body)
-        self.send_listens(listens, LISTEN_TYPE_IMPORT)
-        channel.basic_ack(delivery_tag=method.delivery_tag)
+        self.unique_exchange = Exchange(app.config["UNIQUE_EXCHANGE"], "fanout", durable=False)
```
amCap1712, line 22: We should probably be using durable=True exchanges. However, the existing ones are not durable so we cannot switch deleting the existing ones. Something to do later.

listenbrainz/websockets/listens_dispatcher.py
```diff
@@ -1,81 +1,70 @@
...
 
-class ListensDispatcher(threading.Thread):
+class ListensDispatcher(ConsumerMixin):
 
     def __init__(self, app, socketio):
-        threading.Thread.__init__(self)
         self.app = app
         self.socketio = socketio
+        self.connection = None
+        self.channel2 = None
```
mayhem, line 20: channel2 is not a great name -- what is it used for?

listenbrainz/websockets/listens_dispatcher.py
```diff
@@ -1,81 +1,70 @@
...
-    def create_and_bind_exchange_and_queue(self, channel, exchange, queue):
-        channel.exchange_declare(exchange=exchange, exchange_type='fanout')
-        channel.queue_declare(callback=lambda x: None, queue=queue, durable=True)
-        channel.queue_bind(callback=lambda x: None, exchange=exchange, queue=queue)
-
-    def on_open_callback(self, channel):
-        self.create_and_bind_exchange_and_queue(channel, current_app.config['UNIQUE_EXCHANGE'], current_app.config['WEBSOCKETS_QUEUE'])
-        channel.basic_consume(queue=current_app.config['WEBSOCKETS_QUEUE'], on_message_callback=self.callback_listen)
+    def send_listens(self, event_name, message):
+        self.app.logger.info("Callback called")
```
mayhem, line 29: remove?

mayhem (approved):
Thanks for this PR -- it will be really good to see how it behaves!

requirements.txt
```diff
@@ -49,3 +49,5 @@ google-auth==1.30.0
 pandas==1.3.0
 pyarrow==4.0.1
 more-itertools==8.8.0
+librabbitmq
```
alastair, line 52: this package hasn't been updated since 2018. Did you see https://pypi.org/project/librabbitmq-fork? This is updated recently, and also has wheels for more recent python versions, so we wouldn't have to build it from source.
The github page of librabbitmq-fork is forked from celery/librabbitmq, and all changes have actually been merged back to the celery repo too, so I think that for now using the -fork package from pypi is the better option

amCap1712: Done.

listenbrainz/websockets/listens_dispatcher.py
amCap1712: Done.

listenbrainz/websockets/listens_dispatcher.py
amCap1712: Renamed to playing_now_channel.

PR #1730, opened 2021-11-15
ListenCard responsive improvements

MonkeyDo:
Currently, the ListenCards have OK responsiveness, but the addition of cover art thumbnails doesn't look good in some cases, with the height of the listencard being variable.
Rob also suggested adding a placeholder thumbnail when no cover art is available, with a call to action

image

I reworked the ListenCard flexbox with some more magic:
image
image

iPhone 6/7/8 (375px wide):
image

iPad (768px wide):
image

Galaxy S5 (360px wide):
image

An extreme example: Galaxy Fold (280px wide):
I think here we could think about another design for extremely small screens
image

Desktop:
image

Pins:
image

Compact mode:
image


akshaaatt (approved):
Amazing work! Thanks :100:

PR #1733, opened 2021-11-16
Validate that payload in submission json is a list

amCap1712:
The payload in submission json has to be a list. If it is a dict, each key is treated as a listen by the code. If payload is a string,
each character is treated as a listen. In both of these cases, we get a 500 error because the later performed listen checks do not expect this. For instance, https://sentry.metabrainz.org/organizations/metabrainz/issues/42133/


mayhem (approved):
:bullettrain_side:

PR #1734, opened 2021-11-16
Huesound Listen Card showcase the menu options

akshaaatt:
Fixes Issue:
- https://tickets.metabrainz.org/browse/LB-1010


listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -238,7 +238,10 @@ export default class ListenCard extends React.Component<
 
     // Hide the actions menu if in compact mode or no buttons to be shown
     const hideActionsMenu =
-      compact || (!additionalMenuItems && !enableRecommendButton);
+      compact ||
+      (!additionalMenuItems &&
+        !enableRecommendButton &&
+        !(recordingMBID || spotifyURL || youtubeURL || soundcloudURL));
```
MonkeyDo, line 244: So at this point I think we should probably refactor this code.
It was really easy for me to forget to check if I needed to modify this, leading to the issue you reported.

I'm thinking something like this would be more readable and explicit. What do you think?
```
    let hideActionsMenu = false;
    if (
      !additionalMenuItems &&
      !enableRecommendButton &&
      !recordingMBID &&
      !spotifyURL &&
      !youtubeURL &&
      !soundcloudURL
    ) {
      hideActionsMenu = true;
    }
    if (compact) {
      hideActionsMenu = true;
    }
```

MonkeyDo: Or maybe something like this?
```
const hasActionOptions =
      additionalMenuItems ||
      enableRecommendButton ||
      recordingMBID ||
      spotifyURL ||
      youtubeURL ||
      soundcloudURL;
    const hideActionsMenu = !hasActionOptions;
```

akshaaatt: I think the latest one you sugested makes more sense

MonkeyDo:
Thanks for fixing this ! :rocket:

PR #1735, opened 2021-11-18
AISOTT: Refactor MBID mapping tables to allow easy rechecking on unmatched data

mayhem:
This PR renames the two mapping tables to be more clear and changes the FK relationship between the two tables. This change simplifies the ability to re-check items that were previously not matched. There is the beginnings of automatic rechecking, but I decided to finish that work later once we've manually rechecked listens for 2021 so that we can make better year end reports.

In order to recheck an item in the mapping, set the last_updated timestamp to 1970-01-01 and it will be checked during the next batch of legacy/recheck.

NOTE: Do not merge this yet -- it needs to be carefully released.


listenbrainz/db/dump.py
```diff
@@ -111,21 +111,21 @@
 }
 
 PUBLIC_TABLES_TIMESCALE_DUMP = {
-    'listen_mbid_mapping': (
-        'id',
+    'mbid_mapping_metadata': (
```
amCap1712, line 114: Update schema version also

listenbrainz/db/tests/test_feedback.py
```diff
@@ -58,32 +58,36 @@ def insert_test_data_with_metadata(self, user_id, neg_score=False):
         """ Insert test data with metadata into the database """
 
         with msb_db.engine.connect() as connection:
-            msid = submit_recording(connection, self.sample_recording)
+            msid = get_id_from_meta_hash(connection, self.sample_recording)
```
amCap1712, line 61: Why is this needed especially why in test only?

listenbrainz/mbid_mapping_writer/job_queue.py
```diff
@@ -86,11 +108,60 @@ def load_legacy_listens(self):
...
+                count += 1
+
+        return count
+
     def add_legacy_listens_to_queue(self):
         """Fetch more legacy listens from the listens table by doing an left join
            on the matched listens, finding the next chunk of legacy listens to look up.
            Listens are added to the queue with a low priority."""
 
+        legacy_query = """SELECT data->'track_metadata'->'additional_info'->>'recording_msid'::TEXT AS recording_msid
```
amCap1712, line 152: This query seemed weird at first look.  Maybe a comment to tell its to select not yet mapped msids? Also, would it make to use a WHERE EXISTS or WHERE IN?

amCap1712 (approved):
Other than the comments, LGTM.

listenbrainz/db/tests/test_feedback.py
mayhem: Running tests repeatedly (using test -u) causes this test to fail with a duplicate row violation. Checking to see if the data is already in the DB prevent this error.

listenbrainz/db/dump.py
mayhem: Updated both of them now.

listenbrainz/mbid_mapping_writer/job_queue.py
mayhem: :+1:

listenbrainz/db/tests/test_feedback.py
amCap1712: Oh, weird. because submit_recording is supposed to take care of that. Anyways fine for a test. We can investigate later.

PR #1738, opened 2021-11-19
LB-1001 : Show playlist title in newly created playlist alert message

MonkeyDo:
I've added a quick mention of whether the playlist is public/private in the message, too.
+ Same mechanism applied to duplicating a playlist.

playlist_name

### DIFF
--- a/listenbrainz/webserver/static/js/src/playlists/Playlists.tsx
+++ b/listenbrainz/webserver/static/js/src/playlists/Playlists.tsx
@@ -162,20 +162,20 @@ export default class UserPlaylists extends React.Component<
         currentUser.auth_token,
         playlistId
       );
+      // Fetch the newly created playlist and add it to the state if it's the current user's page
+      const JSPFObject: JSPFObject = await this.APIService.getPlaylist(
+        newPlaylistId,
+        currentUser.auth_token
+      );
       newAlert(
         "success",
         "Duplicated playlist",
         <>
           Duplicated to playlist&ensp;
-          <a href={`/playlist/${newPlaylistId}`}>Copy of {playlistTitle}</a>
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/static/js/src/playlists/Playlists.tsx
```diff
@@ -164,7 +167,7 @@ export default class UserPlaylists extends React.Component<
         "Duplicated playlist",
         <>
           Duplicated to playlist&ensp;
-          <a href={`/playlist/${newPlaylistId}`}>{newPlaylistId}</a>
+          <a href={`/playlist/${newPlaylistId}`}>Copy of {playlistTitle}</a>
```
alastair, line 170: is this name generated by the frontend when it's duplicated, or by the backend? If it's by the backend it should probably use a name returned by the backend, in the case that we decide to change this name.

MonkeyDo: It requires fetching the newly created playlist (since the API endpoint only returns the new playlist MBID), but it is probably wiser.

amCap1712 (approved):
LGTM. Thanks

listenbrainz/webserver/static/js/src/playlists/Playlists.tsx
```diff
@@ -159,20 +162,20 @@ export default class UserPlaylists extends React.Component<
         currentUser.auth_token,
         playlistId
       );
+      // Fetch the newly created playlist and add it to the state if it's the current user's page
```
amCap1712, line 165: Comment looks outdated

MonkeyDo: Mmm, the comment still describes what happens after it:
1. fetch the new playlist
2. on line 178 onwards : i on current user's page, add to state

PR #1739, opened 2021-11-19
LB-747: Update browser window title with currently playing track's name

MonkeyDo:
Update the browser's page title with currently playing song:
window_title


amCap1712 (approved):

amCap1712 (commented):
LGTM. Thoughts on reverting the window title if the track is paused or BP finishes playing the available tracks?

MonkeyDo:
> LGTM. Thoughts on reverting the window title if the track is paused or BP finishes playing the available tracks?

I've had a few thoughts on it, none of which sounded good :p
I think generally it would be great to revert, as you say, after BP is paused/stops playing, which would require saving the original document title on start.
I suppose that's not too tricky so maybe I should try it out.

PR #1741, opened 2021-11-19
LB-959: Redesign BrainzPlayer

MonkeyDo:
The current BrainzPlayer design was made at first for the recent listens page( the only React page at the time), and is not adapted anymore to other uses we want to make of it.

In particular, some ListenBrainz pages don't have the space for a big player, and once we make BrainzPlayer a separate library that can be embedded in third-party websites we will want it to be more flexible.

This new design follows industry standard: a thin player fixed at the bottom of the page which works just as well on mobile screens

Before playing: 
image

Playing a track:
image

Hovering over the cover art in the bottom left corner will pop the cover art or video (if playing with youtube) in a bigger size:
image

.


listenbrainz/webserver/static/js/src/BrainzPlayer.tsx
```diff
@@ -273,11 +273,13 @@ export default class BrainzPlayer extends React.Component<
...
-        invert === true ? currentListenIndex + 1 : currentListenIndex - 1 || 0;
+      nextListenIndex = invert
+        ? currentListenIndex + 1
+        : currentListenIndex - 1 || 0;
     } else if (direction === "down") {
-      nextListenIndex =
-        invert === true ? currentListenIndex - 1 || 0 : currentListenIndex + 1;
+      nextListenIndex = invert
+        ? currentListenIndex - 1 || 0
+        : currentListenIndex + 1;
```
MonkeyDo, line 282: This is incorrect and will break playback. See #1759

Generally, this section of BP has been refactored in #1728 and #1759 and shouldn't be modified before we merge the changes to avoid conflicts.

akshaaatt: I see. Thanks! Fixed.

MonkeyDo:
I know I'm flip-flopping on it being a draft or not, but I keep seeing things to refactor/improve :p

listenbrainz/webserver/static/css/brainzplayer.less
```diff
@@ -1,143 +1,269 @@
-@primary: #08949b;
-@secondary: #0d696e;
-@controlsHeight: 60px;
-@player-min-height: 320px;
+@listenbrainz-blue: #353070;
```
akshaaatt, line 1: We should probably move colors to colors.less and import from there. Would be convenient to define themes at one place then.

listenbrainz/webserver/static/css/brainzplayer.less
```diff
@@ -1,143 +1,269 @@
...
-			min-width: 100%;
-			background-image: url(../img/logo_big.svg);
-			background-repeat: no-repeat;
-			background-position: center top;
-			background-color: aliceblue;
-			opacity: 0.3;
-			text-align: center;
+		display: none;
+		position: absolute;
+		background-color: #f9f9f9;
```
akshaaatt, line 229: Maybe using pre-defined colors here would be better?

listenbrainz/webserver/static/css/main.less
```diff
@@ -20,7 +20,8 @@
 @pager-border-radius: @border-radius-base;
 
 body {
-  padding-bottom: 25px;
+  // Leave some space for BrainzPlayer
+  padding-bottom: 60px;
```
akshaaatt, line 24: I believe some math should be explained here in case we want to update the layouts in the future. Something like why 60px? Maybe a variable could help with that.

akshaaatt (approved):
Looks amazing! Thank you for all the hard work to get this done! :100:

listenbrainz/webserver/static/css/brainzplayer.less
MonkeyDo: I would normally 100% agree, but in this case if we're preparing to make BrainzPlayer an independent package we should reduce dependencies to other parts of Listenbrainz as much as possible.

And when we separate BrainzPlayer, we can replace some of these hardcoded values with a dependency on the design-system package

listenbrainz/webserver/static/css/main.less
MonkeyDo: I'll define a variable to make the name clearer.
We can't import the variable from the brainzplayer.less file unfortunately as they will eventually live in different repos

PR #1742, opened 2021-11-19
AISOTT: Minor mapping fixes and new index on listen table

mayhem:
This PR fixes a couple of minor issues that were not right with the latest mapping refactoring:

1. Fix the incoming listens match rate
2. Add an index on recording_msid (JSONB) on listens -- this becomes more important as we do more re-checks.

### DIFF
--- a/admin/timescale/create_indexes.sql
+++ b/admin/timescale/create_indexes.sql
@@ -3,6 +3,7 @@ BEGIN;
 CREATE INDEX listened_at_user_name_ndx_listen ON listen (listened_at DESC, user_name);
 CREATE INDEX created_ndx_listen ON listen (created);
 CREATE UNIQUE INDEX listened_at_track_name_user_name_ndx_listen ON listen (listened_at DESC, track_name, user_name);
+CREATE INDEX recording_msid_ndx_listen on listen ((data->'track_metadata'->'additional_info'->>'recording_msid'));
 
 -- View indexes are created in listenbrainz/db/timescale.py
 
--- a/listenbrainz/mbid_mapping_writer/job_queue.py
+++ b/listenbrainz/mbid_mapping_writer/job_queue.py
@@ -260,8 +260,7 @@ class MappingJobQueue(threading.Thread):
                         low_quality_rate=stats["low_quality"] - stats["last_low_quality"],
                         no_match_rate=stats["no_match"] - stats["last_no_match"],
                         listens_per_sec=listens_per_sec,
-                        listen_count=stats["listen_count"],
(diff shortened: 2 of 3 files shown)


admin/timescale/create_indexes.sql
```diff
@@ -3,6 +3,7 @@ BEGIN;
 CREATE INDEX listened_at_user_name_ndx_listen ON listen (listened_at DESC, user_name);
 CREATE INDEX created_ndx_listen ON listen (created);
 CREATE UNIQUE INDEX listened_at_track_name_user_name_ndx_listen ON listen (listened_at DESC, track_name, user_name);
+CREATE INDEX recording_msid_ndx_listen on listen ((data->'track_metadata'->'additional_info'->>'recording_msid'));
```
amCap1712, line 6: Assuming this is for https://github.com/metabrainz/listenbrainz-server/blob/3f24a5309480c52c9c872c63a56d18419ce9c56c/listenbrainz/mbid_mapping_writer/job_queue.py#L153 

Adding this index is probably fine but if postgres decides to do a hash join, the index probably won't improve performance. So may want to check execution plans before and after creating the index.

listenbrainz/mbid_mapping_writer/job_queue.py
```diff
@@ -260,15 +260,17 @@ def update_metrics(self, stats):
                         low_quality_rate=stats["low_quality"] - stats["last_low_quality"],
                         no_match_rate=stats["no_match"] - stats["last_no_match"],
                         listens_per_sec=listens_per_sec,
-                        listen_count=stats["listen_count"],
-                        listens_matched=stats["listens_matched"],
+                        listens_matched_p=stats["listens_matched"] / stats["listen_count"] * 100.0,
```
amCap1712, line 263: Calculation looks good but don't know what format etc grafana expects this to be in.

admin/timescale/create_indexes.sql
mayhem: This index is less for the current PR and more for the other things I often do in working with the mapping. For instance, in order to select MSIDs to recheck, I need to query on recording_msid frequently and its slow without an index. Much faster with one -- and in some cases the queries become near instant.

listenbrainz/mbid_mapping_writer/job_queue.py
mayhem: As per our current discussions in IRC, I find it easier to do the calculations in python than wrestling with the influx query syntax, so calculating what I want to show (%) is just easier.

PR #1744, opened 2021-11-19
Add sitewide release stats

amCap1712:
Add sitewide release statistics. Most of the code is copy paste from existing sitewide artist stats, only the Spark SQL query is the new code added. Tests are pending will add in a separate PR later.

### DIFF
--- a/listenbrainz/webserver/views/stats_api.py
+++ b/listenbrainz/webserver/views/stats_api.py
@@ -594,7 +594,65 @@ def get_sitewide_artist():
 @crossdomain()
 @ratelimit()
 def get_sitewide_release():
-    """ TODO: Add documentation """
+    """
+    A sample response from the endpoint may look like:
+
+    .. code-block:: json
+
+        {
+            "payload": {
+                "releases": [
+                    {
+                        "artist_mbids": [],
+                        "artist_name": "Coldplay",
+                        "listen_count": 26,
+                        "release_mbid": "",
+                        "release_name": "Live in Buenos Aires"
+                    },
+                    {
+                        "artist_mbids": [],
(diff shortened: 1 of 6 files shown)


listenbrainz/webserver/views/stats_api.py
```diff
@@ -587,21 +587,91 @@ def get_sitewide_artist():
...
+    stats = db_stats.get_sitewide_stats(stats_range, entity)
     if stats is None:
-        raise APINoContent('')
+        raise APINoContent("")
 
     entity_list, total_entity_count = _process_user_entity(stats, offset, count)
     return jsonify({
         "payload": {
-            "artists": entity_list,
+            entity: entity_list,
```
mayhem, line 674: Missing quotes?

mayhem (commented):
Too tired to finish this now. I'll try and get to this over the weekend.

listenbrainz/webserver/views/stats_api.py
amCap1712: Ah no, its intended. This method (see line 659) take entity variable as input which contains the value artists, releases or recordings etc. So the payload will have keys according to the stat endpoint called. For artists, it will 
`{ "payload": { "artists" : [] }}`. For releases like  `{ "payload": { "releases" : [] }}`. So using the value of entity variable here as the key.

(FWIW, I don't like changing the payload structure like this. I'd prefer another key which contains the name of the stat returned but this is what we do for user stats and having two slightly different formats for stats is probably more confusing so this is what we are stuck with for the time being.)

mayhem (approved):
:bullettrain_front:

PR #1746, opened 2021-11-19
Sitewide daily

amCap1712:
# Problem

# Solution

# Action


amCap1712:
Due to lack of timezones in listen timestamps, it is not possible to compare the listened at timestamps of users of different timezones in a meaningful and hence closing for now.

PR #1749, opened 2021-11-19
Update spotify reader imported listen counts more frequently

amCap1712:
Move the import listen count update code inside the loop so that it is checked after every user. If outside the loop, the condition won't be checked till all users in the batch have been processed (which currently takes ~7 mins). Therefore, updates to the metric are infrequent. Updating the metric at the same frequency (every 60s) as elsewhere (e.g: ts writer) helps simplify writing alerts in the grafana dashboard.

### DIFF
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -387,9 +387,9 @@ def process_all_spotify_users():
                                         u['musicbrainz_id'], exc_info=True)
             failure += 1
 
-    if time.monotonic() > _metric_submission_time:
-        _metric_submission_time += METRIC_UPDATE_INTERVAL
-        metrics.set("spotify_reader", imported_listens=_listens_imported_since_start)
+        if time.monotonic() > _metric_submission_time:
+            _metric_submission_time += METRIC_UPDATE_INTERVAL
+            metrics.set("spotify_reader", imported_listens=_listens_imported_since_start)
 
     current_app.logger.info('Processed %d users successfully!', success)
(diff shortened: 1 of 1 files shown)


mayhem (approved):
lol. What I think we should do here is to change the metric to actually output the change over the previous metric -- which makes all of this easier to display grafana...

zas:
> lol. What I think we should do here is to change the metric to actually output the change over the previous metric -- which makes all of this easier to display grafana...

Well, that's not the correct approach for a number of requests or the like. You just need to provide the count, as an ever increasing integer value. Provide the difference would imply to provide the start & end times for this difference.
Read for example: https://prometheus.io/docs/instrumenting/writing_exporters/#drop-less-useful-statistics (that's the same for  influxdb or any metrics system around, btw).
Provide a simple count, but don't do any calculation over raw data, tools that collect or represent those metrics are made to do that (and much more).

Also, don't forget I want to dump influxdb: we still use it because we have many existing dashboards using Influx query language. Prometheus format is much simpler, and very widely used.
I would recommend to stick to basic metrics types: counter & gauge mainly, see https://prometheus.io/docs/concepts/metric_types/#metric-types

The problem here isn't a real one: data wasn't updated often enough, so it creates artifacts in the representation of data that don't represent the reality.

image

Those peaks are due to the series we have, same number is stored for 5 minutes, then a sudden increase, then same value for another bunch of minutes. And that's exactly what is represented (now, after I fixed queries).

mayhem (approved):
:steam_locomotive: great, let me know when this is released and I'll update the graph!

amCap1712:
According to discussion in IRC, we decided to use listens imported since the last metric update as the value submitted.

PR #1750, opened 2021-11-20
Dump listens for spark in order based on dump type

amCap1712:
The spark listens dump is sorted using this column. full dumps are sorted on listened_at so that during loading listens in spark for stats calculation we can load a subset of files. however, sorting on listened_at creates the issue that the data imported today with listened_at in the past won't show up in the stats until the next full dump. to solve this, we sort and dump incremental listens using the created column. all incremental listens are always loaded by spark , so we can get upto date stats sooner.


mayhem (approved):
:train2:

PR #1751, opened 2021-11-20
Simplify spark request manage

amCap1712:
We never really send a request to spark without calling prepare_query_message on it so makes sense to deduplicate this. Also, use kwargs to avoid dicts everytime.

(Making these changes to because facing lot of repitition while adding new commands for year in music stuff)


mayhem (approved):
:train:

PR #1752, opened 2021-11-22
Timescale Writer: Use counts since last metric update

amCap1712:
Instead of sending counts as incoming/unique listens since the writer was started, use counts of incoming/unique listens since the last time the metric was updated.


mayhem (approved):
:train:

PR #1754, opened 2021-11-23
AISOTT: MBID Mapping explain endpoint and mapping fixes

mayhem:
This PR adds debugging info into the core of the matching engine to let us see exactly what the mapping is doing with any given track. To accomplish this a few things needed to be refactored. Also, artist_credit detuning was never working for punctuation based detuning because punctuation was removed before the detuning. D'oh.

The following fixes/improvements have been made:
- Detune artist_credits on , to help with spotify matching.
- Do not remove punctuation before detuning
- In the matcher make artist credit or recording detuning global (there we two separate local version before, leading to missed matches)
- Fix the incoming listens match rate
- Add an index on recording_msid (JSONB) on listens -- this becomes more important as we do more re-checks.


listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -0,0 +1,283 @@
...
+        return log
+
+    def detune_query_string(self, query, is_artist_credit):
+        """
+            Detune (remove known extra cruft) a metadata field. If a known
+            trailing string exists in the input string, that string and everything
+            after it is removed.
+        """
+
+        strings = ["(", "[", " ft ", " ft. ", " feat ", " feat. ", " featuring "]
```
MonkeyDo, line 91: I would also suggest " - " with spaces around. Would that be too agressive?
Most of the missing matches I see now have some form of " - {year}", " - alternate take", " - Radio Edit" and the like.

The matching has improved in big leap, enough that only those are the only missing matches I see now, making the pattern quite visible. The explainer tool is also super helpful to confirm that intuition.


Compare for example this query with and without the " - 2015 Remaster" at the end:

https://labs.api.listenbrainz.org/explain-mbid-mapping?artist_credit_name=Love&recording_name=Maybe+the+People+Would+Be+the+Times+or+Between+Clark+and+Hilldale+-+2015+Remaster

https://labs.api.listenbrainz.org/explain-mbid-mapping?artist_credit_name=Love&recording_name=Maybe+the+People+Would+Be+the+Times+or+Between+Clark+and+Hilldale

mayhem: Let's give it a try!

listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -0,0 +1,284 @@
...
+        if is_artist_credit:
+            strings.insert(0, ",")
+            strings.append(" with ")
+
+        for s in strings:
+            index = query.find(s)
+            if index >= 0:
+                return query[:index].strip()
+
+        # Yes, this is actually correct.
```
amCap1712, line 101: The documentation of this method says that it removes extra cruft so I would expect that if extra cruft does not exist the same string is returned while actually "" is returned. I think it makes sense to:
1) To mention this in the docstring of the method
2) Explain the comment here - why is this correct?

listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -0,0 +1,284 @@
...
+
+        ac_detuned = prepare_query(self.detune_query_string(artist_credit_name, True))
+        r_detuned = prepare_query(self.detune_query_string(recording_name, False))
+        self._log(f"ac_detuned: '{ac_detuned}' r_detuned: '{r_detuned}'")
+
+        self.is_ac_detuned = False
+        self.is_r_detuned = False
+
+        while True:
+            hit = self.lookup(artist_credit_name_p, recording_name_p)
```
amCap1712, line 227: This part can be extracted to method like `evaluate_and_lookup_hit` and call that method for the three cases. 
1) original names
2) ac detuned r original
3) ac detuned r detuned

The intent is to eliminate the while True loop and many local flags. IMO, that makes the code more readable. But since you mostly work with this code, feel free to choose the version you prefer.

listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -0,0 +1,284 @@
...
+                                <td>{hit['document']["release_name"]}</td>
+                                <td>{hit['document']["artist_credit_name"]}</td>
+                                <td>{hit['document']["recording_mbid"]}</td>
+                                <td>{hit['document']["release_mbid"]}</td>
+                                <td>{hit['document']["artist_credit_id"]}</td>
+                            </tr>
+                        </table>
+                    """))
+
+            if not hit:
```
amCap1712, line 253: My understanding is that this covers three cases:
1) original names
2) ac detuned r original
3) ac detuned r detuned

Do you think it makes sense to add the 4th, ac original r detuned?

listenbrainz/mbid_mapping_writer/mbid_mapper.py
amCap1712: With some effort, I think I was able to figure out why this is correct. My understanding is that L255 later in this file does not execute if there's nothing to detune.

listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -0,0 +1,284 @@
...
+                    r_dist <= self.MATCH_TYPE_MED_QUALITY_MAX_EDIT_DISTANCE:
+                self._log(Markup(f"""<b>med quality</b>: ac_detuned {self.is_ac_detuned}, r_detuned {self.is_r_detuned},
+                                     ac_dist {ac_dist:.3f} r_dist {r_dist:.3f}"""))
+                return (hit, MATCH_TYPE_MED_QUALITY)
+
+            # Poor results so far, lets try detuning fields and try again
+            if ac_dist > self.MATCH_TYPE_MED_QUALITY_MAX_EDIT_DISTANCE and ac_detuned:
+                ac_hit = ac_detuned
+                ac_detuned = ""
+                self.is_ac_detuned = True
```
amCap1712, line 161: self.is_ac_detuned and self.is_r_detuned is modified in the evaluate method as well the lookup method. That makes the overall flow difficult to read. While the log explainer helps to understand the checks it would be nide it the code could be refactored to improve readability.

listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -0,0 +1,284 @@
...
+                self._log(Markup(f"""<b>no match</b>, ac_dist too high {ac_dist:.3f} continuing: detune ac"""))
```
amCap1712, line 162: Since its possible to detune both the search artist credit name and the hit's artist credit name, we should mention in the logged line that its `ac_hit` that has been detuned. Similarly, `r_hit` in next case.

listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -0,0 +1,284 @@
...
+        search_parameters = {
+            'q': query,
+            'query_by': "combined",
+            'prefix': 'no',
+            'num_typos': self.MATCH_TYPE_MED_QUALITY_MAX_EDIT_DISTANCE
+        }
+
+        while True:
+            try:
+                hits = self.client.collections[COLLECTION_NAME].documents.search(
```
amCap1712, line 195: Despite all the fixes, one recording I mentioned earlier never got mapped. Looking at the [explain](https://labs.api.listenbrainz.org/explain-mbid-mapping?artist_credit_name=Selena+Gomez%2C+Rauw+Alejandro&recording_name=Baila+Conmigo+%28with+Rauw+Alejandro%29), I saw `HIT: artist: selena gomez w rauw alejandro recording: baila conmigo`.  This is interesting because of the `w` in there. Orginally, I thought it might be a with being clipped down to w somehow turns out its not.

I manually queried typesense for this query to find the recording mbid, it was hitting. It is [f90720d0-3c4e-42a9-94f1-df26dd4ad472](https://musicbrainz.org/recording/f90720d0-3c4e-42a9-94f1-df26dd4ad472).

The artist credit name on this recording is `Selena Gomez w/ Rauw Alejandro`. We remove the punctuation so `w/` becomes `w` and we get what we see above. Now looking at that recording's page, I do see the with version there but its the release artist name or track artist name. So I can see how this recording did not get matched. I don't know enough about this stuff to tell whether track_artist and release_artist fields are worthwhile in considering matches but it looks like a possible way to improve matches.

Looking closely at the explains again, we see it tried original first, the ac_dist was within limits but not r_dist. It then detunes artist credit which actually increases the ac_dist and then tries a detuned recording name but by this time ac_dist has become bigger than the threshold so no match. But if we had detuned recording first instead of artist credit, we would have got a match. So now, I think it makes sense to add the case for `original ac r detuned ` as suggested in the other comment. 

(Alternatively, try to detune on the field which has a greater dist first. But I think covering all cases if probably better.)

amCap1712:
We decided to take up the feedback when we do the follow up mapping work.

PR #1759, opened 2021-11-25
Correction of Brainzplayer playNextTrack mechanism

MonkeyDo:
In the process of refactoring Brainzplayer in #1728 , we broke the mechanism a bit and clicking on "next" in BrainzPlayer does the same thing as clicking "previous".
That is because we are checking if a condition is truthy instead of checking specifically if the condition === true


mayhem (approved):
Thanks!!

PR #1761, opened 2021-11-26
Remove unused props from user page

amCap1712:
While going through sentry found some errors around user page view. These removals are unrelated to the errors but these fields look unused so might as well remove those.


mayhem (approved):
deleting code/data FTW.

MonkeyDo (approved):
:rocket:

PR #1762, opened 2021-11-27
Fix broken feed due to incorrect dict access

amCap1712:
artist_msid is a part of additional_info inside track_metadata and not a top level key. Hence, fixing access of artist_msid key. The feed page is also not loading for any user due to an ISE occuring as a result of this error.


MonkeyDo (approved):
Good fix, thank you !

Tested on beta.LB.org, fixes currently broken feed page

PR #1763, opened 2021-11-29
Cap listen counts for sitewide stats

amCap1712:
We witnessed that a few users' listen counts were dominating the entire sitewide stats. So, we decided to impose a per user per recording listen count limit. This improved the sitewide stats by a lot.

For now only recording stats use the capped limit, I'll modify the artist and release stats once the recording query is reviewed and finalized.

### DIFF
--- a/listenbrainz_spark/stats/sitewide/entity.py
+++ b/listenbrainz_spark/stats/sitewide/entity.py
@@ -31,6 +31,18 @@ entity_model_map = {
 }
 
 
+def get_listen_count_limit(stats_range: str) -> int:
+    """ Return the per user per entity listen count above which it should
+    be capped. The rationale is to avoid a single user's listens from
+    over-influencing the sitewide stats.
+
+    For instance: if the limit for yearly recordings count is 500 and a user
+    listens to a particular recording for 10000 times, it will be counted as
+    500 for calculating the stat.
+    """
+    return 500
+
+
 def get_entity_stats(entity: str, stats_range: str) -> Optional[List[SitewideEntityStatMessage]]:
     """ Returns top entity stats for given time period """
     logger.debug(f"Calculating sitewide_{entity}_{stats_range}...")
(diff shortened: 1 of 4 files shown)


listenbrainz_spark/stats/sitewide/entity.py
```diff
@@ -31,6 +31,18 @@
 }
 
 
+def get_listen_count_limit(stats_range: str) -> int:
```
amCap1712, line 34: For now, the limit is set to 500 for all stats. However, we probably will want different listen count limits for different time ranges. Should we also make the limit change depending on whether its a artist stat or recording stat or release stat?

listenbrainz_spark/stats/sitewide/recording.py
```diff
@@ -17,33 +18,50 @@ def get_recordings(table: str, limit: int = SITEWIDE_STATS_ENTITY_LIMIT):
...
+                 , first(recording_name) AS recording_name
                  , recording_mbid
-                 , first(artist_name) AS any_artist_name
+                 , first(artist_name) AS artist_name
                  , artist_credit_mbids
-                 , nullif(first(release_name), '') as any_release_name
+                 , nullif(first(release_name), '') as release_name
                  , release_mbid
-                 , count(*) as listen_count
+                 , LEAST(count(*), {user_listen_count_limit}) as listen_count
```
amCap1712, line 29: This causes the final listen count to contain the capped count not the actual listen count. An alternative is to do both counts, the actual listen count and the capped count. Sort on the capped count and assign rank like 1, 2, 3 based on it and have the actual listen count field available separetely for anyone who's interested in it.

listenbrainz_spark/stats/sitewide/entity.py
mayhem: I think for now that is ok -- if things feel like they are wrong, we'll go look again.

listenbrainz_spark/stats/sitewide/recording.py
mayhem: Why should we care about the actual listen count here? I think as it is it sounds fine to me.

mayhem (approved):

listenbrainz_spark/stats/sitewide/recording.py
amCap1712: I was thinking some one might be interested in the actual listen counts. I am fine either way, will merge as is then for now.

PR #1765, opened 2021-12-03
Update test.sh commands to run as the local user

alastair:
# Problem

When running some test commands on linux + docker, the test process runs as root sometimes resulting in the creation of files owned by root (pyc files, jest snapshots), making it hard to clean up again when using your local user.

Also, there is a lot of duplication of `docker-compose -f .... -p .....`

# Solution

Use `docker-compose run --user` flag to run the process as your local user. This should have no effect on macos. We use the same process in develop.sh

# Action

I tested unit, int, fe tests (along with supporting commands) and they all succeeded.
Spark tests failed, but I don't think it was related to my change. @amCap1712, can you take a look at these and see if anything looks weird?

### DIFF
--- a/docker/docker-compose.integration.yml
+++ b/docker/docker-compose.integration.yml
@@ -24,6 +24,8 @@ services:
     image: listenbrainz
     volumes:
       - ..:/code/listenbrainz:z
+    environment:
+      PYTHONDONTWRITEBYTECODE: 1
     depends_on:
       - redis
       - lb_db
--- a/docker/docker-compose.spark.yml
+++ b/docker/docker-compose.spark.yml
@@ -30,3 +30,5 @@ services:
     command: dockerize -wait tcp://namenode:9000 -timeout 60s bash -c "cp listenbrainz_spark/config.py.sample listenbrainz_spark/config.py; PYTHONDONTWRITEBYTECODE=1 python -m pytest -c pytest.spark.ini"
     volumes:
       - ..:/rec
+    environment:
+      PYTHONDONTWRITEBYTECODE: 1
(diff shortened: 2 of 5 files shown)


amCap1712:
@alastair I took a look, https://github.com/metabrainz/listenbrainz-server/runs/4406978626?check_suite_focus=true#step:7:160 . It looks like not running as root causes some internal lookup in spark to fail. My guess is that the datanode/namenode (i.e. https://github.com/metabrainz/listenbrainz-server/pull/1765/files#diff-3722d9ba8feb2d3feac8ce71a209a638d4b404e1c53f937188761181594023e2R159) too need to have this same user mapping for tests to run as a non root user.

alastair:
thanks @amCap1712.
If the spark tests don't write any files to a mounted volume then we can continue running it as root.

amCap1712:
Its been some time since I ran these tests locally but looking at https://github.com/metabrainz/listenbrainz-server/blob/7d53e4cdcbdeda45f9845cb6158ca406c694ce3c/docker/docker-compose.spark.yml#L30,
I don't think any such files are written.

alastair:
Unfortunately, python still writes bytecode files when you run it, and we're mounting the source directory into the container.
I think the better solution for now is to set `PYTHONDONTWRITEBYTECODE=1` env variable in order to stop it from generating pyc files.

I know a few tricks to fix the user id error - the best way is to create a user while building the container, and then as part of your entrypoint run `usermod -u` to change that user's uid to the same id of the local user. This way there is a user in /etc/passwd with this uid, and the app should start. However, for now I think this is additional work for a problem that we can fix by preventing pycs to be written in the first place.

alastair:
I notice that develop.sh still runs as root anyway, I'm not sure what we should do here. Updating to run as a local user would work in most cases (unless we wanted to do a pip install in a shell). For now I'll focus on test.sh here, and will look at develop later.

alastair:
@amCap1712 updated this to output frontend test output files to /tmp to avoid permission problem

PR #1766, opened 2021-12-03
Standardise all LB ports to start with 7

alastair:
# Problem

In order to prevent conflicts when running multiple MeB projects, we
decided to isolate ports for each project. Continues from bcc5836d

# Solution

* Don't just update the published port, also update the default port for the service inside each container
* Also update api compat and websockets to use a 7 port
* publish postgres (on 127.0.0.1)
* Update config, documentation, and startup scripts

# Action


amCap1712 (approved):
When we merge this, we need to inform all devs to update their config.py (in case they have configured API_URL) and their redirect URIs in MB (possibly other services like spotify etc. too if they have configured those).

PR #1767, opened 2021-12-03
Moar submitter functionality

alastair:
# Problem

The musicbrainz submitter is great, except:
* I can never remember where it is
* I have to keep looking up my token in order to submit it
* submitted items don't include recording mbid

# Solution

* Add the command to manage.py
* Add the ability to submit based on username. This selects the user from the local dev database and extracts the token
* add recording mbid

# Action

Configuration is currently pointing to port 7000 (requires #1766 to be merged first)


manage.py
```diff
@@ -287,6 +287,30 @@ def refresh_continuous_aggregates():
     """
     ts_refresh_listen_count_aggregate()
 
+@cli.command()
+@click.option("-u", "--user", type=str)
+@click.option("-t", "--token", type=str)
+@click.argument("releaseid", type=str)
```
mayhem, line 293: minor nitpick: Given how many IDs we have floating about, I've started using the convention of being explicit about what type of id it is. Now I always try to use: 

- <entity>_id : for row ids
- <entity>_mbid : for mbids
- <entity>_msid : for msids
- <entity>_name : for the text value of one of the above.

mayhem (approved):

manage.py
alastair: thanks. I updated this to be releasembid

PR #1768, opened 2021-12-03
Return the data of submitted pins, including the pin row id

alastair:
# Problem

The row id is needed to show the pin on the frontend and allow it to be deleted

This prevents nan-nana-ann-nan on the frontend.

# Solution

# Action


alastair:
Another 2 bugs that I found related to pins when I was testing on my local environment:

* If I have a pin and load it (so there definitely is a pin id), then when I select delete, the highlight on the pin is removed, but it stays there. Not sure if this is intentional:
Screenshot from 2021-12-03 12-57-16

Sometimes when I pin something (it seems to appear with both mbid and msid pins), _sometimes_ the pin card doesn't show up at the top of the listens. I'm trying to reproduce it, but I can't do it consistently. Has this been seen before?

MonkeyDo:
> Another 2 bugs that I found related to pins when I was testing on my local environment:
> 
> * If I have a pin and load it (so there definitely is a pin id), then when I select delete, the highlight on the pin is removed, but it stays there. Not sure if this is intentional:
> Sometimes when I pin something (it seems to appear with both mbid and msid pins), _sometimes_ the pin card doesn't show up at the top of the listens. I'm trying to reproduce it, but I can't do it consistently. Has this been seen before?

Haven't seen either of these issues before, so definitely not intentional.

When you say "doesn't show up at the top of the listens", does that mean it appears somewhere else, or not at all?

alastair:
> When you say "doesn't show up at the top of the listens", does that mean it appears somewhere else, or not at all?

I mean that it doesn't appear at all. Initially I thought it was because I only had an MSID, then I was able to pin something with an MSID. then it happened to me when I had an MBID.
It's possible that it happened because the API call failed (e.g. dev server was restarting), but I never had the inspector open in the times that I saw it fail. I'll keep trying and see if I can reproduce it.

MonkeyDo (approved):
I've added the modifications required on the front-end to use the returned payload
I've only tested locally (and with some difficulty), but it works as expected.
RowId exists on the newly created pin, and I' able to delete it without having to refresh the page, so mission accomplished

I would appreciate if you could please review the JS side of things though.

listenbrainz/webserver/views/pinned_recording_api.py
```diff
@@ -63,12 +63,12 @@ def pin_recording_for_user():
...
 
     try:
-        db_pinned_rec.pin(recording_to_pin)
+        recording_to_pin_with_id = db_pinned_rec.pin(recording_to_pin)
     except Exception as e:
         current_app.logger.error("Error while inserting pinned track record: {}".format(e))
         raise APIInternalServerError("Something went wrong. Please try again.")
 
-    return jsonify({"status": "ok"})
+    return jsonify({"status": "ok", "data": _pinned_recording_to_api(recording_to_pin_with_id)})
```
amCap1712, line 71: Returning status: ok with data looks a bit weird!  On the top of my head, I don't think we do this elsewhere. So let's just send data and remove the status field?

alastair: good question. We can remove it.

alastair: @amCap1712 in this case are you thinking of `{"data": the data}`, or `return jsonify(_pinned_recording.....)`? One thing I don't like about the second option is that it makes it difficult to modify this response in the future. Keeping in mind the GET method for this returns a dictionary with a key `"pinned_recordings"`... maybe returning `"pinned_recording": whatever` (singular) is a good way of being consistent?

amCap1712: "pinned_recording": whatever sounds good.

PR #1770, opened 2021-12-06
Remove load_recordings_from_mbids

amCap1712:
We don't use this method anywhere currently. We should use mbid data from mapping tables wherever needed so removing this.


mayhem (approved):
Thanks for fixing this PR -- make a lot more sense now than it did earlier.

amCap1712:
Yeah, it had got borked during a rebase. Fixed now :D

PR #1771, opened 2021-12-07
Remove artist_msid from recommending track event

amCap1712:
We do not use artist msid anywhere in recommending track event. The frontend also does not use this info. Hence, removing this. This makes having optional recording_msids easier because then we no longer need to lookup messybrainz db for each item. Currently, we lookup messybrainz for each pin because artist msid is required to recommend anything. Now that we do not need artist msid for this event, we can go ahead and make msids optional elsewhere.


mayhem (approved):
Less code FTW!

PR #1772, opened 2021-12-07
Add API support for getting and fetching the algorithm_metadata field.

mayhem:
This PR adds support for submitting algorithm_metadata with a playlist if the submitter is an approved playlist submitter. This functionality is needed for troi-bot to submit metadata about playlists so that periodic playlists can be updated in sane and predictable manner. Playlists will be given a slug to identify them as belonging to a given patch that was run for a given time period, so that they may be periodically updated. 

NOTE: This will need to be deployed before we can delivery the YIM playlists.

### DIFF
--- a/listenbrainz/db/playlist.py
+++ b/listenbrainz/db/playlist.py
@@ -3,6 +3,7 @@ import datetime
 from typing import List, Optional
 
 import sqlalchemy
+import ujson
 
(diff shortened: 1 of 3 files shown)


mayhem:
Docs updated to mention the algorithm_metadata field: https://musicbrainz.org/doc/jspf

amCap1712 (approved):

PR #1773, opened 2021-12-07
LB-296: Disable root logger configuration in listenstore causing duplicate messages

alastair:
# Problem

All log messages were appearing twice:

> spotify_reader_1    | 2021-12-07 15:05:09,784 listenbrainz.webserver INFO     Processed 1 users successfully!
> spotify_reader_1    | 2021-12-07 15:05:09,784 INFO Processed 1 users successfully!

We specifically [set up a `listenbrainz` logger](https://github.com/metabrainz/listenbrainz-server/blob/dfb0fcff6bef52195b3c5fd9a2bea0804420a0b4/listenbrainz/__init__.py#L14) for all LB packages, but somehow messages were being sent to a root logger too which was mysteriously being created.

# Solution

Turns out it wasn't much of a mystery. The root logger was implicitly created in listenstore, way back in one of the earliest LB commits

# Action

Now that we don't have a root logger, there may be some messages that will get lost, however the only instance of logging that I can find outside of the `listenbrainz` package is in the spark package, and we already set up a root logger for that, so everything should be fine.

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import pprint
 import sys
@@ -77,6 +78,7 @@ def gen_app(debug=None):
 
     In the Flask app returned, blueprints are not registered.
     """
+
     app = CustomFlask(
         import_name=__name__,
         use_flask_uuid=True,
@@ -85,6 +87,11 @@ def gen_app(debug=None):
     load_config(app)
     if debug is not None:
         app.debug = debug
+    # As early as possible, if debug is True, set the log level of our 'listenbrainz' logger to DEBUG
+    # to prevent flask from creating a new log handler
+    if app.debug:
+        logger = logging.getLogger('listenbrainz')
+        logger.setLevel(logging.DEBUG)
 
     # initialize Flask-DebugToolbar if the debug option is True
     if app.debug and app.config['SECRET_KEY']:
--- a/listenbrainz/__init__.py
+++ b/listenbrainz/__init__.py
@@ -7,11 +7,10 @@ DUMP_LICENSE_FILE_PATH = os.path.join(os.path.dirname(os.path.realpath(__file__)
                                       'db', 'licenses', 'COPYING-PublicDomain')
 
 _handler = logging.StreamHandler()
-_handler.setLevel(logging.INFO)
 _formatter = logging.Formatter("%(asctime)s %(name)-20s %(levelname)-8s %(message)s")
 _handler.setFormatter(_formatter)
 
 _logger = logging.getLogger("listenbrainz")
+# This level is set to DEBUG in listenbrainz.webserver.gen_app if Flask DEBUG=True
 _logger.setLevel(logging.INFO)
 _logger.addHandler(_handler)
-
--- a/listenbrainz/listenstore/__init__.py
+++ b/listenbrainz/listenstore/__init__.py
@@ -1,5 +1,3 @@
-import logging
-
 ORDER_DESC = 0
 ORDER_ASC = 1
 ORDER_TEXT = [ "DESC", "ASC" ]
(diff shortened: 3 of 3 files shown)


mayhem (approved):
THANK YOU!

alastair:
I've pushed some further changes to this, things are a bit more complex than they first appeared.

1. When you start flask, with `app = Flask(__name__)`, flask will use the value of the argument as the name of the logger. In our case, this would be `listenbrainz.webserver`
2. The first time you access `app.logger`, flask will [look for a logger in the tree](https://github.com/pallets/flask/blob/9486b6cf57bd6a8a261f67091aca8ca78eeec1e3/src/flask/logging.py#L26) until it finds one, in this case it'll look for listenbrainz.webserver, then listenbrainz, then root. It should find the 'listenbrainz' one.
3. *However*, in the case that you have debugging set, and the default logger _doesn't have a level set_, [flask will set the desired logging level to DEBUG](https://github.com/pallets/flask/blob/9486b6cf57bd6a8a261f67091aca8ca78eeec1e3/src/flask/logging.py#L68-L69)
    * we were in this situation, where our 'listenbrainz' logger was set to INFO, but the newly requested 'listenbrainz.webserver' one wasn't configured
4. Then, when flask looks for a handler that can support DEBUG, it skipped our one (because it's only INFO), and then went and created a new one. This means that listenbrainz.webserver had 2 handlers which reponded to INFO, flask's debug one, and our default INFO one.

I tried to fix this by setting the argument to `app = Flask()` to "listenbrainz", so that it correctly finds the logger, and then manually change this logger's level to debug if we're in debug mode.
Unfortunately this breaks other parts of flask, such as templates because it uses this full path to look for template files.

Still looking into whether the better solution here is to hard-code the flask templates folder, or manually configure a logger for 'listenbrainz.webserver' to stop flask doing magic.

alastair:
Another option would be to let flask go and create the handler magically, but then explicitly delete it afterwards: https://flask.palletsprojects.com/en/2.0.x/logging/#removing-the-default-handler
Though this also feels a bit magic (and you have to know in what cases the handler is created, in our case deleting a random line of code in a different place in the file could also prevent this from working)

alastair:
Looking further at the flask documentation: https://flask.palletsprojects.com/en/2.0.x/api/#application-object

> About the First Parameter 
>  For example the Flask-SQLAlchemy extension will look for the code in your application that triggered an SQL query in debug mode. If the import name is not properly set up, that debugging information is lost. (For example it would only pick up SQL queries in yourapplication.app and not yourapplication.views.frontend)

It may be worth looking into this in more depth - as I understand it, if we set the argument to `listenbrainz.webserver` then other extensions may not be able to find data in `listenbrainz.db`, for example

alastair:
After discussion with @amCap1712 we decided to fix the logger in this PR, and deal with other stuff like fixing the first arg to `Flask()` in a separate task. I reverted the argument back to `__name__`. In order to prevent flask from creating a new handler, I explicitly set our handler to DEBUG if flask debug is True

listenbrainz/webserver/__init__.py
```diff
@@ -77,6 +78,12 @@ def gen_app(debug=None):
 
     In the Flask app returned, blueprints are not registered.
     """
+    # As early as possible, if debug is True, set the log level of our 'listenbrainz' logger to DEBUG
+    # to prevent flask from creating a new log handler
+    if debug is True:
+        logger = logging.getLogger('listenbrainz')
```
alastair, line 84: one thought here is that we could put a comment in `listenbrainz/__init__.py` where the logger is initially created, informing that the level is set to DEBUG here in gen_app only if FLASK_DEBUG=1?

listenbrainz/webserver/__init__.py
```diff
@@ -77,11 +78,18 @@ def gen_app(debug=None):
...
         use_flask_uuid=True,
     )
 
+    # As early as possible, if debug is True, set the log level of our 'listenbrainz' logger to DEBUG
+    # to prevent flask from creating a new log handler
+    if app.debug:
+        logger = logging.getLogger('listenbrainz')
+        logger.setLevel(logging.DEBUG)
+
     load_config(app)
```
amCap1712, line 93: Uh! Approved too soon. Shouldn't load config happen before `app.debug` check?

alastair: good catch, thanks. This was from some testing that I was doing and forgot to revert. It's fixed now

PR #1775, opened 2021-12-08
Use LEFT JOINs for medium format in mapping because it can be NULL

amCap1712:
We found that some releases (eg: https://musicbrainz.org/release/a4be0448-accc-4f12-bf23-08122434270e) were not being mapped and digging deeper discovered it was because medium_format is nullable. To include these releases in the mapping use a LEFT JOIN but order by NULLS LAST so that releases with a medium format available are preferred.


mayhem (approved):
Sweet!

PR #1778, opened 2021-12-13
Codebase Cleanup

akshaaatt:
- Move all frontend tests to a separate folder
- Rearrange images folder
- Rearrange all files in a systematic order


listenbrainz/webserver/static/js/tests/utils/utils.test.tsx
```diff
@@ -54,6 +57,7 @@ describe("preciseTimestamp", () => {
       new Date(testDate).toLocaleString(undefined, {
         day: "2-digit",
         month: "short",
+        year: "numeric",
```
MonkeyDo, line 60: Why was there a need to change the test assertion?
We need to be careful of modifying tests in such a huge PR with so many changes.
Is it definitely the same result? If so a PR comment would be welcome.

MonkeyDo (changes requested):
For future reference, it would be nicer to chop such a big job into multiple PRs, it's a lot of work to review 213 files !

There are files that have been deleted, including a bunch of test snapshots.
Is that intentional?
I see the index.jsx file was empty, but for example the 5-seconds-of-silence.mp3  file shouldn't be removed
Regarding the snapshots, it's dangerous to be removing or changing pieces of the test harness during a big change like this.

MonkeyDo:
Looks like the front-end tests also point to missing methods, have a look :)

PR #1781, opened 2021-12-13
Update LB public ports. again.

alastair:
# Problem
In bcc5836d and 42e72fd2 we standardised on port 7000 for the webserver and other closeby ports for other services.
However, macos 12 introduced an AirPlay receiver which uses ports 5000 and 7000, causing chaos.

# Solution

Try again, using 81xx ports.

Additionally, we missed a few places in the configuration and
documentation that were pointing at the old port 80 (especially in oauth
configuration to external services like spotify). Update these

# Action

Consider applying ranges of 100 ports to other MeB apps. Proposal: MB: 80xx, LB: 81xx, CB: 82xx, AB: 83xx


MonkeyDo (approved):
Fine by me.
In the interest of not having to change it again-again, is this range safe-ish or should we aim for a higher range that is less likely to be used anytime soon, say 20100-20900 for example ?

PR #1783, opened 2021-12-15
Look up coverart for top_releases

alastair:
# Problem

Find coverart for top releases when they're imported from spark.

Find coverart for a release, and if it doesn't exist, see if there is any coverart for that release's releasegroup

# Solution

# Action

### DIFF
--- a/listenbrainz/db/year_in_music.py
+++ b/listenbrainz/db/year_in_music.py
@@ -234,18 +234,20 @@ def caa_id_to_archive_url(release_mbid, caa_id):
 
 def get_coverart_for_top_releases(top_releases):
     """Use the CAA database connection to find coverart for each release
-    1. For all releases, find coverart
-    2. For releases where there is no coverart, get their releasegroup
-    3. Find release mbids for the above release groups
-    4. For these release mbids, find coverart
+    1. Get release id and releasegroup id for each release (also considering release mbid redirects)
+    2. Get coverart for these releases
+    3. For releases with no cover art, see if any other release in their releasegroup has coverarat
     """
+
+    if musicbrainz_db.engine is None:
+        current_app.logger.warning("get_coverart_for_top_releases: No connection to MusicBrainz database")
+        return {}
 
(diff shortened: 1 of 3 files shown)


listenbrainz/db/year_in_music.py
```diff
@@ -227,6 +228,123 @@ def insert_playlists(troi_patch_slug, import_file):
...
+                        , release 
+                     FROM cover_art_archive.index_listing
+                    WHERE release in :release_ids
+                      AND is_front = 't';
+        """
+        res = connection.execute(sqlalchemy.text(query), {"release_ids": tuple(release_id_to_mbid.keys())})
+        for row in res.fetchall():
+            release_id = row["release"]
+            caa_id = row["id"]
+            release_mbid = release_id_to_mbid[release_id]
```
alastair, line 287: there are a handful of places where I'm getting data out of a dict that I'm pretty sure is always going to be there, but I don't know if it might fail in some unexpected situations

listenbrainz/webserver/static/js/src/year-in-music/YearInMusic.tsx
```diff
@@ -356,7 +357,11 @@ export default class YearInMusic extends React.Component<
               >
                 {yearInMusicData.top_releases.slice(0, 10).map((release) => (
                   <img
-                    src="/static/img/cover-art-placeholder.jpg"
+                    src={
+                      yearInMusicData.top_releases_coverart[
+                        release.release_mbid
```
alastair, line 362: no checking here to see if this mbid is set, or if it exists in the coverart array

alastair: @MonkeyDo

MonkeyDo: You can add ` ?? "/static/img/cover-art-placeholder.jpg"` to default to the placeholder image when no item is found in the coverart dict

PR #1785, opened 2021-12-16
Fix YIM email sending bug

amCap1712:
Do not try to send email to users if they don't have one but do create a timeline event for all.

### DIFF
--- a/listenbrainz/db/year_in_music.py
+++ b/listenbrainz/db/year_in_music.py
@@ -423,6 +423,7 @@ def notify_yim_users():
               FROM statistics.year_in_music yim
               JOIN "user"
                 ON "user".id = yim.user_id
+             WHERE email IS NOT NULL
         """)
         rows = result.fetchall()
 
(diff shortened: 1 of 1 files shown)


listenbrainz/db/year_in_music.py
```diff
@@ -445,15 +450,19 @@ def notify_yim_users():
...
+                subject="Year In Music 2021",
+                text=render_template("emails/year_in_music.txt", **params),
+                to_email=row["email"],
+                to_name=user_name,
+                html=render_template("emails/year_in_music.html", **params),
+                lb_logo_cid=lb_logo_cid,
+                lb_logo=lb_logo
+            )
+        except Exception:
+            current_app.logger.error("Could not send YIM email", exc_info=True)
```
alastair, line 464: will this include enough context information (in sentry?) for us to know who it was sent to?

amCap1712: Sentry will capture local variable so user_name will be available in sentry to identify it.

alastair (approved):

PR #1786, opened 2021-12-17
Deduplicate new releases of top artists

amCap1712:
If a release has multiple artist credits, the release may duplicated when joined with the top artists. To avoid duplication, use collect_set. However, this does not prevent multiple releases of a release group released in the same year from appearing in the list again.


mayhem (approved):
sooo simple!

PR #1787, opened 2021-12-17
YIM: Warning and placeholders when data not calculated

MonkeyDo:
Friendly warning text and some placeholder text if any of the top level keys of the YIM data is missing.
Also added contact details for questions and feedback in the intro text.

image


amCap1712 (approved):
💯 💯 💯

PR #1788, opened 2021-12-17
YIM: Fix coverflow navigation

MonkeyDo:
Puts the navigation arrows back and visible on the Coverflow

Note: This disables clicking on the covers to right/left of the middle one to navigate, unfortunately

image


amCap1712 (approved):
LGTM. Thanks!

PR #1789, opened 2021-12-17
YIM: Add a follow button next to user name

MonkeyDo:
If currently logged-in user is not looking at their own page

Also cleaned up use of `listens` array which shouldn't be in the state or be modified in the render function like that.
image


amCap1712 (approved):
🚢

PR #1790, opened 2021-12-17
Add UI for this year, this month and this week stats

amCap1712:
These time ranges were added some time ago and track to date stats. This PR adds the frontend UI to browse these stats.


akshaaatt (approved):
Works well and the stats look good.
Thanks for the amazing addition! :100:

PR #1791, opened 2021-12-18
Fix OutOfMemory error in site wide listening activity stats

amCap1712:
Empirical testing has shown that joining time_range table with listens table before grouping is more expensive than converting each listen's listened_at to the desired date format, grouping on that and then joining with time_range table to fill any gaps. The first approach is so much more expensive that it causes a OOM in the Spark Driver.

### DIFF
--- a/listenbrainz_spark/stats/common/listening_activity.py
+++ b/listenbrainz_spark/stats/common/listening_activity.py
@@ -55,9 +55,15 @@ def get_two_quarters_ago_offset(_date: date) -> relativedelta:
         return relativedelta(month=4, day=1)
 
 
-def get_time_range(stats_range: str) -> Tuple[datetime, datetime, relativedelta, str]:
-    """ Returns the start time, end time, segment step size and date format to use for
-    calculating the listening activity stats
+def get_time_range(stats_range: str) -> Tuple[datetime, datetime, relativedelta, str, str]:
+    """ Returns the start time, end time, segment step size, python date format and spark
+     date format to use for calculating the listening activity stats
+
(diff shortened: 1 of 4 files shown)


listenbrainz_spark/stats/common/listening_activity.py
```diff
@@ -134,29 +145,34 @@ def get_time_range(stats_range: str) -> Tuple[datetime, datetime, relativedelta,
...
         step = relativedelta(days=+1)
         date_format = "%d %B %Y"
+        spark_date_format = "d MMMM y"
     elif stats_range == "quarter":
         from_offset = get_two_quarters_ago_offset(latest_listen_date)
         to_offset = relativedelta(months=+6)
         # compute listening activity for each week with date format as day
         step = relativedelta(weeks=+1)
         date_format = "%d %B %Y"
+        spark_date_format = "d MMMM y"
```
amCap1712, line 162: Thinking again, this won't work for quarters. For quarters we want to round off each timestamp to week which is not simple. Need to rethink quarterly stats or special case the query for this.

mayhem: how does this comment impact this PR? Merge for now and fix later?

mayhem (approved):
Aside from the one pending comment, this looks fine.

listenbrainz_spark/stats/common/listening_activity.py
amCap1712: As per IRC discussion, modified to use daily steps here instead of weekly.

PR #1794, opened 2021-12-20
Add recording mbid to recording feedback

amCap1712:
Adds support to give feedback using recording mbid only. The feedback endpoints till now only accepted recording msid, however we now want to support recording msids and recording mbids both. This creates issues if we want to maintain backward compatibility, we either need to add uglish workarounds to the existing endpoints and default to the current behaviour to maintain compat. Otherwise, we need to add separate endpoints. Also, in future we probably want to remove the support for msids so many of these measures will look redundant at that time. 

For example:

1) `GET /user/<user_name>/get-feedback-for-recordings?recordings=[foobar]` accepts a recordings param which should be a list of recording msids. We need to add another param to support recording mbids. I think it would be nice if the returned json contained two top level keys - mbids and msids to denote the feedback for each but that json structure will probably be incompatible with existing. 

2) `GET /recording/<recording_msid>/get-feedback` accepts a recording msid in the path. It should still work if we a mbid is sent but need a way to differentiate whether its a msid or mbid. Maybe add a query param `is_mbid` but uhhh?

### DIFF
--- a/listenbrainz/db/feedback.py
+++ b/listenbrainz/db/feedback.py
@@ -46,11 +46,12 @@ def delete(feedback: Feedback):
             DELETE FROM recording_feedback
              WHERE user_id = :user_id
                AND recording_msid = :recording_msid
+               AND recording_mbid = :recording_mbid
             """), {
             'user_id': feedback.user_id,
             'recording_msid': feedback.recording_msid,
-        }
-        )
+            'recording_mbid': feedback.recording_mbid
+        })
 
 
 def get_feedback_for_user(user_id: int, limit: int, offset: int, score: int = None, metadata: bool = False) -> List[Feedback]:
@@ -69,14 +70,17 @@ def get_feedback_for_user(user_id: int, limit: int, offset: int, score: int = No
     """
 
     args = {"user_id": user_id, "limit": limit, "offset": offset}
-    query = """ SELECT user_id,
-                       "user".musicbrainz_id AS user_name,
-                       recording_msid::text, score,
-                       recording_feedback.created
+    query = """ SELECT user_id
+                     , "user".musicbrainz_id AS user_name
+                     , recording_msid::text
+                     , recording_mbid::text
+                     , score
+                     , recording_feedback.created
                   FROM recording_feedback
                   JOIN "user"
                     ON "user".id = recording_feedback.user_id
-                 WHERE user_id = :user_id """
+                 WHERE user_id = :user_id
+    """
 
     if score:
         query += " AND score = :score"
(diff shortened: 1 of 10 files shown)


listenbrainz/db/feedback.py
```diff
@@ -22,15 +23,15 @@ def insert(feedback: Feedback):
         connection.execute(sqlalchemy.text("""
             INSERT INTO recording_feedback (user_id, recording_msid, score)
                  VALUES (:user_id, :recording_msid, :score)
-            ON CONFLICT (user_id, recording_msid)
+            ON CONFLICT (user_id, recording_msid, recording_mbid)
```
amCap1712, line 26: The issue with this is that recording_mbid are currently all NULL so we'll get duplicate feedbacks. However, we end up in the same situation if we insert using just recording_mbid because we already have msid feedback. One way could be to run a script to insert mbids for each msid.

listenbrainz/db/tests/test_feedback.py
```diff
@@ -197,9 +191,8 @@ def test_get_feedback_for_user_with_metadata(self):
         self.assertEqual(result[0].score, self.sample_feedback_with_metadata[0]["score"])
         self.assertEqual(result[0].track_metadata["artist_name"], "Portishead")
         self.assertEqual(result[0].track_metadata["track_name"], "Strangers")
-        self.assertEqual(result[0].track_metadata["additional_info"]["recording_mbid"], "076255b4-1575-11ec-ac84-135bf6a670e3")
-        self.assertEqual(result[0].track_metadata["additional_info"]["release_mbid"], "1fd178b4-1575-11ec-b98a-d72392cd8c97")
-        self.assertEqual(result[0].track_metadata["additional_info"]["artist_msid"], self.saved_artist_msid)
+        self.assertEqual(result[0].track_metadata["mbid_mapping"]["recording_mbid"], "076255b4-1575-11ec-ac84-135bf6a670e3")
```
amCap1712, line 194: This made me realise that currently feedback endpint serves mapped mbids inside additional_info key whereas pinned recordings and listens do that in mbid_mapping. The frontend checks both so no issues there. Since this is a MSID endpoint, I doubt there are any 3rd party consumers of this API. So I am in favour of making the switch to using the mbid_mapping key here right away.

amCap1712:
PR updated according to IRC discussion. Pending: Adding new tests to cover cases when when feedback is given using mbids current tests only cover msids,

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -40,29 +36,31 @@ def recording_feedback():
...
-    try:
-        if feedback.score == 0:
-            db_feedback.delete(feedback)
-        else:
-            db_feedback.insert(feedback)
-    except Exception as e:
-        current_app.logger.error("Error while inserting recording feedback: {}".format(e))
-        raise APIInternalServerError("Something went wrong. Please try again.")
+
+    if feedback.score == 0:
```
mayhem, line 60: Dont we have a constant defined for this?

mayhem (approved):
Loooks good, one possible tiny nitpick to deal with.

amCap1712:
This completes the backend support for submitting/fetching feedback using mbids. Frontend support to utilise these apis is pending and will be done in a separate PR.

PR #1795, opened 2021-12-22
Chunk listening activity and daily activity stats

amCap1712:
To speed up insertion of stats in the database, use bulk inserts. An earlier PR (#1764) implemented this for entity (top artists/releases/recording stats). This PR extends it to daily and listening activity stats.


mayhem (approved):
Oy, this was long. Thankfully only refactoring!

PR #1796, opened 2021-12-24
Combine unit tests and integration tests

amCap1712:
Combine docker file setup and test scripts of unit and integration tests. The actions setup still runs the tests separately because its slow to run all tests together.


amCap1712:
Another improvement we can make here or in a follow up PR: https://github.com/metabrainz/listenbrainz-server/pull/1819#r784643934

PR #1797, opened 2021-12-24
LB-676: Listenbrainz reports git commit error on startup

amCap1712:
LB-676

Only perform .git-version check in prod where the file will always be present. For local development, this probably doesn't make much sense anyway because we can run git normally.


mayhem (approved):
Do we even need it for test?

amCap1712:
Don't think it matters much but easier to do for all.

PR #1798, opened 2021-12-25
LB-993: Increase buffer-size to avoid 502 errors

amCap1712:
LB-993

When the user feedback endpoint is queried with too many recordings, it errors with 502. This is because the recording mbids are sent as a part of the query params in the URL and cause the size of request to exceed uwsgi buffer size. Double the buffer size to allow us to do ~100 mbids in one go.

Tested the fix on test.lb to confirm it works.


alastair (approved):
perfect!

PR #1801, opened 2021-12-28
LB-564: Add lastfm feedback import

amCap1712:
https://tickets.metabrainz.org/browse/LB-564
Import loved tracks from last.fm. This PR depends on recording feedback supporting mbids first.

### DIFF
--- a/listenbrainz/domain/lastfm.py
+++ b/listenbrainz/domain/lastfm.py
@@ -9,6 +9,10 @@ from listenbrainz import db
 
 
 def bulk_insert_loved_tracks(user_id: int, feedback: List[Tuple[int, str]]) -> int:
+    # DO NOTHING because we can only import loved tracks.
+    # if a user has imported something before from last.fm then the score doesn't change.
+    # if the user has unloved a track on last.fm, we won't receive it this time so can't
+    # remove it from feedback in any case.
     query = """
         INSERT INTO recording_feedback (user_id, created, recording_mbid, score)
         VALUES %s ON CONFLICT (user_id, recording_mbid) DO NOTHING
--- a/listenbrainz/domain/lastfm.py
+++ b/listenbrainz/domain/lastfm.py
@@ -1,0 +1,82 @@
+import requests
+from flask import current_app
+from psycopg2.extras import execute_values
+from requests import RequestException
+from requests.adapters import HTTPAdapter, Retry
+from typing import List, Dict, Tuple
+
+from listenbrainz import db
+
+
+def bulk_insert_loved_tracks(user_id: int, feedback: List[Tuple[int, str]]) -> int:
+    query = """
+        INSERT INTO recording_feedback (user_id, created, recording_mbid, score)
+        VALUES %s ON CONFLICT (user_id, recording_mbid) DO NOTHING
+    """
+    with db.engine.raw_connection() as connection:
+        cursor = connection.cursor()
+        execute_values(cursor, query, feedback, template=f"({user_id}, %s, %s, 1)")
+        return cursor.rowcount
+
+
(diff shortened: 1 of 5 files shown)


MonkeyDo:
> This PR depends on recording feedback supporting mbids first.

Looks like we're going to be able to look at this PR again soon :)

listenbrainz/domain/lastfm.py
```diff
@@ -0,0 +1,119 @@
...
+from listenbrainz import db
+from brainzutils import musicbrainz_db
+
+from listenbrainz.webserver.errors import APINotFound
+
+
+def bulk_insert_loved_tracks(user_id: int, feedback: list[tuple[int, str]]):
+    # DO NOTHING because we can only import loved tracks.
+    # if a user has imported something before from last.fm then the score doesn't change.
+    # if the user has unloved a track on last.fm, we won't receive it this time so can't
```
mayhem, line 18: I have a feeling that this case doesn't happen all that often.

listenbrainz/domain/lastfm.py
```diff
@@ -0,0 +1,119 @@
...
+        ON CONFLICT (user_id, recording_mbid)
+         DO NOTHING
+    """
+    connection = db.engine.raw_connection()
+    with connection.cursor() as cursor:
+        execute_values(cursor, query, feedback, template=f"({user_id}, to_timestamp(%s), %s, 1)")
+    connection.commit()
+
+
+def load_recordings_from_tracks(track_mbids: list) -> dict[str, str]:
```
mayhem, line 32: While this function is rather simple and well named, its existence isn't quite clear. Maybe a docstring to clarify?

listenbrainz/domain/lastfm.py
```diff
@@ -0,0 +1,119 @@
...
+          JOIN recording
+            ON track.recording = recording.id
+         WHERE track.gid IN :tracks
+    """
+    with musicbrainz_db.engine.connect() as connection:
+        result = connection.execute(text(query), {"tracks": tuple(track_mbids)})
+        return {row["track_mbid"]: row["recording_mbid"] for row in result.mappings()}
+
+
+def fetch_lfm_feedback(lfm_user):
```
mayhem, line 46: docstring mssing.

Also, there is potential for a bad user experience here. While it is unlikely that a lot of users have pages upon pages of feedback, this function could take seconds, possibly minutes (in case of retries) to complete, in which the user's browser will time out the page load, leaving the user in the dark about the import of their feedback.

Sadly, there is no simple solution for fixing this. :(

listenbrainz/domain/lastfm.py
```diff
@@ -0,0 +1,119 @@
...
+    counts = {
+        "total": total_count,
+        "missing_mbid": missing_mbid_count,
+        "invalid_mbid": invalid_mbid_count
+    }
+
+    return track_mbids, track_data, counts
+
+
+def import_feedback(user_id, lfm_user) -> dict:
```
mayhem, line 100: docstring

mayhem (approved):
Please add some docstrings. As far as dealing with the connection timeout problem, maybe we should deploy the feature as is and if this turns out to be a problem for users, then we can revisit it when the time comes.

listenbrainz/domain/lastfm.py
amCap1712: Makes sense, discussed on IRC to explore running this import in background in future.

listenbrainz/domain/lastfm.py
amCap1712: Yes probable. If someone complains, we can check if anything can be done.

PR #1802, opened 2021-12-28
Spotify mapping

amCap1712:
First rough cut of spotify mapping. We originally intended to use continous aggregates for this but when I tried to implement the mapping using it I hit a few issues.

-  Continous aggregates need to include the time column of the hypertable but we don't have the use of listened_at column here. 
- Already materialized chunks are not refreshed automatically we need to refresh them manually like we do for listen count 30 day  view currently.
- When a chunk is refreshed from which listens have been deleted, some data may get removed from the mapping table.
- Normal postgres materialized views can be a solution but the issue with those is that those cannot be updated incrementally.

This PR currently creates the mapping by processing each chunk and putting data in a mapping table. After initial run, process all listens since the last run. To be added: a cron job to run this periodically (daily?).

### DIFF
--- a/listenbrainz/mbid_mapping/mapping/spotify_mapping.py
+++ b/listenbrainz/mbid_mapping/mapping/spotify_mapping.py
@@ -1,0 +1,45 @@
+from datetime import datetime
+
+from sqlalchemy import text
+
+from listenbrainz.db import timescale
+
+base_query_begin = text("""
+    INSERT INTO spotify_mapping
+         SELECT data->'track_metadata'->'additional_info'->>'recording_msid'::UUID
+              , data->'track_metadata'->'additional_info'->>'spotify_id'
+              , data->'track_metadata'->'additional_info'->>'spotify_album_id'
+              , data->'track_metadata'->'additional_info'->>'spotify_artist_ids'
+              , data->'track_metadata'->'additional_info'->>'spotify_album_artist_ids'
+              , data->'track_metadata'->'additional_info'->>'isrc'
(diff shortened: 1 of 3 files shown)


listenbrainz/mbid_mapping/mapping/spotify_mapping.py
```diff
@@ -0,0 +1,45 @@
...
+            AND """)
+initial_query = text("listened_at >= :start AND listened_at < :end")
+incremental_query = text("created >= :watermark")
+# spotify_id field is considered while calculating a msid, so one msid can only have
+# one spotify track id associated with it. Assuming, given a spotify track id other
+# spotify ids do not change all ids for a given msid will be fixed. Thus, ignoring rows
+# if msid is already present in mapping is enough to avoid duplicates and not lose data.
+base_query_end = text(" ON CONFLICT (recording_msid) DO NOTHING")
+
+LISTEN_MINIMUM_TS = int(datetime(2002, 10, 1).timestamp())
```
amCap1712, line 27: These constants are already exisiting in other open PRs, will update to use those if those get merged before this.

listenbrainz/mbid_mapping/mapping/spotify_mapping.py
```diff
@@ -0,0 +1,45 @@
...
+CHUNK_SECONDS = 432000
+
+
+# TODO: decide what to do with watermark - run for last time period or store last run timestamp somewhere
```
amCap1712, line 31: .

PR #1803, opened 2021-12-28
LB-912 (II): Add endpoints and UI to delete feed events

hrik2001:
# Problem
There were no UI to delete feed events. The backend (endpoint) of this feature was implemented in #1610 .

# Solution

This PR implements a menu item in the `recording_recommendation` and `recording_pin` feed events to delete them. Menu item only appears if the event was created by the same user who is browsing the feed.


hrik2001:
@MonkeyDo, there has been an issue where timestamps generated in previous year are showing up, since 2021 was hardwired in the tests. Hence the Missing MB Data snapshot got updated too and hence I pushed it here.

hrik2001:
@MonkeyDo, there has been an issue where timestamps generated in previous year are showing up, since 2021 was hardwired in the tests. Hence frontend tests are failing. I attempted to solve this issue at #1808.

Also, some backend tests are failing that are false positives and seems like some occasional fault with db mocking.
[Chat about false positive](https://chatlogs.metabrainz.org/libera/metabrainz/2022-01-05/?msg=4932195&page=1)

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -327,9 +329,94 @@ export default class UserFeedPage extends React.Component<
...
+            "Successfully deleted!",
+            <>Successfully deleted!</>
+          );
+          const new_events = _remove(events, (element) => {
+            return !(
+              element?.id === event.id &&
+              element.event_type === EventType.RECORDING_PIN
+            );
+          });
+          this.setState({ events: new_events });
```
MonkeyDo, line 397: At a glance this isn't the utility you're looking for.
According to the docs (https://lodash.com/docs/4.17.15#remove), Lodash's remove mutates the array (which we don't want) and returns an array of the removed values (which isn't exactly what we need either.
I think that's why you had to invert the value (`return !(…`), which produces the results we want but is difficult to read.

Consider using _.filter instead (https://lodash.com/docs/4.17.15#filter) which returns "an array of all elements predicate returns truthy for" AND doesn't mutate the original array.

listenbrainz/webserver/static/js/src/APIService.ts
```diff
@@ -969,4 +969,23 @@ export default class APIService {
     await this.checkStatus(response);
     return response.json();
   };
+
+  deleteFeedEvent = async (
+    eventType: string,
+    username: string,
+    userToken: string,
+    id?: number
```
MonkeyDo, line 977: I don't know how it is implemented on the server but if an ID is required then this method should probably throw an error if there is no `id`
It would be great to add a test case for that as well.

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -327,9 +329,94 @@ export default class UserFeedPage extends React.Component<
...
+      try {
+        const status = await APIService.deletePin(
+          currentUser.auth_token as string,
+          event.id as number
+        );
+        if (status === 200) {
+          newAlert(
+            "success",
+            "Successfully deleted!",
+            <>Successfully deleted!</>
```
MonkeyDo, line 389: I realize the message should be optional for cases like this where we don't have much else to say.
Let's try not to duplicate the title, and pass an empty string instead.

There's another instance of the same thing in this file.

MonkeyDo (changes requested):
Just a couple of changes needed, as well as merging in the master branch (to make sure the front-end tests can pass).

This is looking great, thank you ! :)

listenbrainz/webserver/static/js/src/APIService.ts
hrik2001: hi @MonkeyDo, really sorry for late reply because of exams,
so, id is indeed implemented in the backend
https://github.com/metabrainz/listenbrainz-server/pull/1803/files#diff-68b3baf8382d61b3b29a2894bb9edca6db7a948d6f0794bf28275c8581f859b4R462

When I started working on LB-912, we realised that ID needs to be passed in multiple APIs, hence I implemented id to be an optional parameter in #1610 .
https://github.com/metabrainz/listenbrainz-server/pull/1610/files#diff-c99f9398509233652fe3be847b91ec0e6b89d968b4bdc68cf81063767aed7683R80

so since `id` is indeed being passed by the backend for all those events that can be deleted from the feed page, hence, it shouldn't be a problem.

MonkeyDo: No problem, take your time! Your exams are a lot more important, no question about that :)

Regarding the optional ID, as far as I can tell the argument for this API method should be `id: number`, i.e. not optional, as it doesn't make sense to call `/feed/events/delete` without an id.

I'd still like you to cover the case of a missing id parameter and throw an error with a helpful message.
Otherwise, if something changes in the future and we call this method without the id, something is likely to break in less clear ways.

hrik2001: Hi, this totally makes sense, the embarrassing thing is that I mistakenly thought that this is a code review for `types.d.ts` and initially thought you wanted me to remove `?` from `id?: number` from the same file (since I had no recollection of putting `id?: number` in `deleteFeedEvent`). But now I do remember that I had put it because I was facing some typescript error and then forgot to switch it back and it completely vanished from my mind. ;_; , which is why I deleted my earlier comment :P

also, the edge case of `id` not being there for future changes is a very good idea, hadn't crossed my mind before. I would implement that for sure. Thank you so much @MonkeyDo!!

MonkeyDo: Let me know if you need a hand with the typescript issues :)

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
hrik2001: seems like the boolean nature of both `remove` and `filter` are same
image
what we can do for better code readability is this, and not mess around with `new_events` at all (this mutates the original array)
```js
         _remove(events, (element) => {
            return !(
              element?.id === event.id &&
              element.event_type === EventType.RECORDING_PIN
            );
          });
          this.setState({ events: events });
```


I also wanted to ask, why is mutating the array a bad idea?

MonkeyDo (approved):

MonkeyDo:
Thanks a lot @hrik2001  !

PR #1804, opened 2021-12-29
Make Year In Music runnable for past years

amCap1712:
Make year in music runnable for past year by changing all `this_year`s to use the specified year.


mayhem (approved):
Looks good!

PR #1806, opened 2022-01-04
LB-673, LB-794: Include artist list for each country in "Artist Origin" endpoint

amCap1712:
Show artists from each country in artist origin chart.

Screenshot (228)

There are a few issues which we should resolve before deploying:

- Artist names from the same country are repeated because we `explode` the list of artist mbids when calculating countries. One possible solution is to only show the name only once and link it to the first artist mbid only. This is what we do elsewhere from artist links anyways.
- The list can become quite large in many cases. We can limit it to say show top 5-10 artists from that country?
- The links are there but can't be clicked because the tooltip continously moves along with the cursor. Maybe we can reduce the width of the artist map and add a side bar ish thing which uses the top artists component to list the artists from the country which was clicked last.

Screenshot (229)


listenbrainz/webserver/static/js/src/APIService.ts
```diff
@@ -420,7 +420,7 @@ export default class APIService {
   getUserArtistMap = async (
     userName: string,
     range: UserStatsAPIRange = "all_time",
-    forceRecalculate: boolean = false
+    forceRecalculate: boolean = true
```
amCap1712, line 423: Revert this before merging but invalidate all existing artist map stats manually from db.

listenbrainz/webserver/static/js/src/stats/Choropleth.tsx
```diff
@@ -133,14 +133,46 @@ export default function CustomChoropleth(props: ChoroplethProps) {
...
     return (
-      <BasicTooltip
-        id={feature.label}
-        color={feature.color}
-        value={`${feature.formattedValue} ${suffix}`}
-        enableChip
-      />
+      <div
+        style={{
+          background: "white",
```
amCap1712, line 141: I copied these styles from the source of BasicTooltip.

mayhem (approved):
I'm very much looking forward to seeing this live!

MonkeyDo (approved):
Looks really nice, thanks !
Indeed that tooltip isn't ideal and we'll have to figure out another way to display the artists, but for the time being this is a great improvement !

I took the liberty of tweaking the custom tooltip a little:
image

As a reminder, there's a [comment](https://github.com/metabrainz/listenbrainz-server/pull/1806/files#r778328908) about reverting a change in APIService.ts before merging.

PR #1809, opened 2022-01-06
LB-1000: Missing headers in API error responses

hrik2001:
# Problem
In LB-1000, there was mention of rate limit headers not passing through. Personally I had experienced CORS headers not passing through because of ISE. So this is an attempt to fix both of these issues

# Solution
Hence implemented error handling in only the `crossdomain()` decorator, since in all APIs (at least in all the ones I checked), `crossdomain()` was being called first.

# Action
Should some sort of logging need to be implemented in this PR?

### DIFF
--- a/listenbrainz/webserver/errors.py
+++ b/listenbrainz/webserver/errors.py
@@ -140,7 +140,6 @@ def init_error_handlers(app):
         if current_app.config.get('IS_API_COMPAT_APP') or request.path.startswith(API_PREFIX):
             response = make_response(jsonify({'code': code, 'error': error.description}), code)
             response.headers["Access-Control-Allow-Origin"] = "*"
-            response.headers["Access-Control-Expose-Headers"] = "X-RateLimit-Remaining,X-RateLimit-Limit,X-RateLimit-Reset,X-RateLimit-Reset-In"
             return response
         return error_wrapper('errors/{code}.html'.format(code=code), error, code)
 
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/errors.py
```diff
@@ -138,7 +138,7 @@ def handle_error(error, code):
                 otherwise
         """
         if current_app.config.get('IS_API_COMPAT_APP') or request.path.startswith(API_PREFIX):
-            return jsonify({'code': code, 'error': error.description}), code
+            return crossdomain(jsonify({'code': code, 'error': error.description}), code)
```
amCap1712, line 141: Only pass the jsonify to crossdomain which will add headers to the response. The `handle_error` method has to return something like `(Response, code)` so code should be outside call to crossdomain.
```suggestion
            return crossdomain(jsonify({'code': code, 'error': error.description})), code
```

amCap1712 (commented):
With the suggested change, LGTM. Please mark this as ready, i'll do a final test on beta and merge.

listenbrainz/webserver/errors.py
```diff
@@ -138,7 +138,10 @@ def handle_error(error, code):
                 otherwise
         """
         if current_app.config.get('IS_API_COMPAT_APP') or request.path.startswith(API_PREFIX):
-            return jsonify({'code': code, 'error': error.description}), code
+            response = make_response(jsonify({'code': code, 'error': error.description}), code)
```
amCap1712, line 141: shouldn't need make_response here because jsonify returns a Response object itself.

amCap1712 (approved):
LGTM. Thanks.

PR #1811, opened 2022-01-10
Simplify redis connection initializations

amCap1712:
During of creation of flask app, we initialize cache so there's no need to do it again in listenstore's which are created after cache has been initialized. There are a couple of places where we still init_cache manually (in mbid mapping writer and recalculate user data cron job because we don't create a flask app there


mayhem (approved):
Thanks for cleaning this up!

alastair (approved):
go go go!

PR #1812, opened 2022-01-10
Reduce initializations of timescale listenstore

amCap1712:
Just like RedisListenstore, TimescaleListenstore is inited at time of app creation. So there is no need to create a new one in each request, just use the same again and again. At many places, the imports are changed from `_ts` to `timescale_connection._ts`, for normal usage either is fine because the app init happens before imports but in tests that's not always the case so the latter one is preferred as it reduces unrelated errors in tests.


listenbrainz/listenstore/tests/test_test.py
```diff
@@ -55,12 +45,7 @@ def setUp(self):
         self.reset_timescale_db()
 
         self.ns = config.REDIS_NAMESPACE
-        self.logstore = init_timescale_connection(self.log, {
-            'REDIS_HOST': config.REDIS_HOST,
-            'REDIS_PORT': config.REDIS_PORT,
-            'REDIS_NAMESPACE': config.REDIS_NAMESPACE,
-            'SQLALCHEMY_TIMESCALE_URI': config.SQLALCHEMY_TIMESCALE_URI,
-        })
+        self.logstore = timescale_connection._ts
```
mayhem, line 48: This is sooo much better!

mayhem (approved):

PR #1813, opened 2022-01-10
Some simplifications to TimescaleWriter

amCap1712:
ListenWriter does not have any use of its own and only subclass is TimescaleWriter. It only has a few utility methods which can be nicely moved into TimescaleWriter.


mayhem (approved):
I'm also very happy that this useless base class is gone.

alastair (approved):
🎉

PR #1814, opened 2022-01-11
Simplify ListenStores

amCap1712:
We have TimescaleListenStore and RedisListenStore both of which inherit from ListenStore. However, both of the classes share nothing in common with each other. Further, ListenStore class on itself has no use anyways. So remove the ListenStore class. Further, the TimescaleListenStore is a bit large. It has many methods to create dumps which can be logically carved into a separate file. For the lack of a better name, calling it DumpListenStore.


listenbrainz/listenstore/listenstore.py
```diff
@@ -1,70 +0,0 @@
...
-
-from listenbrainz.listenstore import ORDER_ASC, ORDER_DESC, DEFAULT_LISTENS_PER_FETCH
-
-
-class ListenStore(object):
-    MAX_FETCH = 5000          # max batch size to fetch from the db
-    MAX_FUTURE_SECONDS = 600  # 10 mins in future - max fwd clock skew
-
-
-    def __init__(self, logger):
```
amCap1712, line 15: The `self.log` parameter is used at a few places by both RedisListenStore and TimescaleListenStore. For now, I have added an __init__ method accepting logger to both but IIRC the intent is to move away from using `current_app` logger everywhere in which the `self.log` will also go away.

alastair (approved):
splitting up dumps code into a separate class is a great idea, thanks

PR #1815, opened 2022-01-11
LB-795: Update label formatting in artist origin chart

amCap1712:
LB-795

Adding ~ to format specifier trims unnecessary decimals. Also do not pluralize if formatted value is 1.

Screenshot (230)


akshaaatt (approved):
Works fine, thanks!

PR #1816, opened 2022-01-11
Use gen_app over create_app where possible

amCap1712:
Almost all LB services run inside a flask app context, we use create_app to create flask app instances for use there. However, create_app also registers endpoints and related jinja stuff in the flask app. This is not needed at most of those places, only configurations and connections are. We also have a gen_app method which does exactly that so we should be using gen_app at all places except the LB web api app.

However, there are many places to replace create_app with gen_app and we are habitual of using create_app everywhere so we'll forget this and end up using the create_app instead gen_app again. For this reason, I am renaming create_app to create_web_app because it registers web blueprints. gen_app is renamed to create_app and so using create_app is fine now. If
someone uses create_app and gets an error about missing endpoints, then use create_web_app at that place instead.

### DIFF
--- a/listenbrainz/server.py
+++ b/listenbrainz/server.py
@@ -1,3 +1,3 @@
-from listenbrainz.webserver import create_app
+from listenbrainz.webserver import create_web_app
 
-application = create_app()
+application = create_web_app()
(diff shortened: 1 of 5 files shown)


alastair (approved):
This looks like a good change, thanks! 👍

PR #1817, opened 2022-01-11
Additional cookie security measures

amCap1712:
We only use cookies for user login and remembering login sessions. We had discussed taking some additional measures to secure these cookies but then forgot to deploy these.


mayhem (approved):

alastair (approved):
This looks fine, but I'd like to test it in beta first in the following situations:

Log in, close tab, re-visit site - check you're logged in
Log in, close browser, open & revisit- check you're logged in
Log in, click link from other https site (e.g. a link in gmail) and ensure that you're logged in - check you're logged in (this is the one that I'm unsure about - I believe that we discussed that lax is fine for this situation and is what other MeB sites use? I remember being bitten by a similar issue, but I think it was with strict that caused the bug)

alastair:
I just checked the other project where i had this problem and we switched from strict to lax to fix it, so I'm happy with this change 👍

PR #1820, opened 2022-01-14
Top discoveries 2021

mayhem:
This PR adds two more mapping modules, AKA reports:

1. Top discoveries for year -- this creates a table in the mapping schema of the provided mb-docker database that lists all the tracks that a user listened to the first time in the given year. 
2. Tracks for the year -- this also creates a table in the mapping schema, where this one creates a historgram of which tracks and how many times a user played for a given year.

From these intermediate tables Troi can generate two YIM playlists, via the datasets hosted on Bono.


listenbrainz/mbid_mapping/reports/top_discoveries.py
```diff
@@ -0,0 +1,135 @@
+from datetime import datetime, timezone
+
+import psycopg2
+import psycopg2.extras
+from psycopg2.errors import OperationalError
+
+from mapping.utils import insert_rows, log
```
amCap1712, line 7: Not sure if this import will work, may need to modify it to `..mapping.utils`.

listenbrainz/mbid_mapping/reports/top_discoveries.py
```diff
@@ -0,0 +1,135 @@
...
+              FULL OUTER JOIN mbid_mapping m
+                           ON (data->'track_metadata'->'additional_info'->>'recording_msid')::uuid = m.recording_msid
+              FULL OUTER JOIN mbid_mapping_metadata mm
+                              ON mm.recording_mbid = m.recording_mbid
+                        WHERE user_name in %s
+                          AND mm.recording_mbid IS NOT NULL
+                     GROUP BY user_name, artist_name, mm.artist_mbids, track_name, m.recording_mbid
+                       HAVING (array_agg(extract(year from to_timestamp(listened_at))::INT ORDER BY
+                                         extract(year from to_timestamp(listened_at))::INT))[1] = %s
+                     ORDER BY user_name, array_length(array_agg(extract(year from to_timestamp(listened_at))::INT), 1) DESC"""
```
amCap1712, line 101: A few comments explaining how the query works would be nice. I have seen this query before but took a bit of time to remember how it works.

listenbrainz/mbid_mapping/reports/top_discoveries.py
```diff
@@ -0,0 +1,135 @@
...
+                            , mm.artist_mbids
+                         FROM listen
+              FULL OUTER JOIN mbid_mapping m
+                           ON (data->'track_metadata'->'additional_info'->>'recording_msid')::uuid = m.recording_msid
+              FULL OUTER JOIN mbid_mapping_metadata mm
+                              ON mm.recording_mbid = m.recording_mbid
+                        WHERE user_name in %s
+                          AND mm.recording_mbid IS NOT NULL
+                     GROUP BY user_name, artist_name, mm.artist_mbids, track_name, m.recording_mbid
+                       HAVING (array_agg(extract(year from to_timestamp(listened_at))::INT ORDER BY
```
amCap1712, line 99: I think this could become `min(extract(year from to_timestamp(listened_at))::INT) = %s` and then the array_agg in select cause could become just a `count(*)`.

listenbrainz/mbid_mapping/reports/tracks_of_the_year.py
```diff
@@ -0,0 +1,122 @@
...
+    return [r[0] for r in rows]
+
+
+def chunks(lst, n):
+    """ Helper function to break a list into chunks """
+    for i in range(0, len(lst), n):
+        yield lst[i:i + n]
+
+
+def fetch_tracks_listened_to(lb_conn, mb_conn, ts):
```
amCap1712, line 68: missing 1 arg and different name from one used inside function. should probably be `lb_conn, mb_conn, start_ts, end_ts`.

listenbrainz/mbid_mapping/manage.py
```diff
@@ -91,6 +93,24 @@ def cron_log():
...
+    calculate_top_discoveries(year)
+
+
+@cli.command()
+@click.argument('year')
+def top_tracks(year):
+    """
+        Calculate the top tracks data for the given year.
+    """
+    calculate_top_tracks_of_the_year(year)
```
amCap1712, line 111: should be `calculate_tracks_of_the_year`

listenbrainz/mbid_mapping/reports/tracks_of_the_year.py
```diff
@@ -0,0 +1,122 @@
...
+                    to_insert = []
+                    mb_conn.commit()
+
+            insert_rows(mb_curs, "mapping.tracks_of_the_year", to_insert)
+            mb_conn.commit()
+
+
+def calculate_tracks_of_the_year(year):
+    """
+        Main entry point for creating top discoveries table.
```
amCap1712, line 113: Let's also add the text from the PR to the docstrings of the respective reports.

amCap1712 (approved):
LGTM. Thanks!

listenbrainz/mbid_mapping/reports/top_discoveries.py
mayhem: It has worked so far...

listenbrainz/mbid_mapping/reports/tracks_of_the_year.py
mayhem: Shit, I guess I didn't actually check both scripts. Sigh.

listenbrainz/mbid_mapping/reports/top_discoveries.py
mayhem: I find that PG is not entirely predictable for when this works, but it doesn't always work. I'm going to call this good enough and leave it be.

listenbrainz/mbid_mapping/reports/tracks_of_the_year.py
mayhem: Good idea.

PR #1821, opened 2022-01-15
Fix bug in artist country code api endpoint

amCap1712:
In #1818, artist_name was added to the artist country code api. In that , the intent was probably to append r instead of row. Due to not calling dict on the result row, it is appended as an array to json. Due to this artist map endpoint which calls this labs api endpoint errors and returns a 500. Weird that the labs api test didn't catch this bug.


mayhem (approved):
Indeed odd -- thanks for fixing.

PR #1822, opened 2022-01-18
Fix LB server local development

amCap1712:
Due to the changes in #1816, local development broke. Flask by default creates application using create_app but we want to use create_web_app so need to specify it manually. The other option is to create a command in manage.py like run_api_compat_server, I didn't go for that version because werkzeug.run_simple's defaults are different from flask run's
so need more work to make sure it works as desired. We can go that way later if we think it has advantages but currently need to get local development working again.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -40,7 +40,7 @@ services:
       args:
         GIT_COMMIT_SHA: HEAD
     environment:
-      FLASK_APP: listenbrainz.webserver
+      FLASK_APP: listenbrainz.webserver:create_web_app()
       FLASK_ENV: development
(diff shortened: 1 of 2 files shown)


alastair:
Interesting, I wonder if `run_api_compat_server` is like that because it was copied from the original start-server command that we had in manage.py (we used werkzeug.run_simple there because I wanted to reload the server if the manifest file changed. we solved this problem by not adding hashes to js files during development and went back to using flask run)

This specific change (updating the FLASK_APP) is fine by me. I'd also be open to replacing run_api_compat_server with flask run, too. This is only a requirement for development, as we use uwsgi in prod.

amCap1712:
Moving api_compat_server to similar way sounds good, will do.

PR #1825, opened 2022-01-18
Reorganize docs - 2

amCap1712:
- Split API endpoint docs across pages for better readability.
- Rename sections


docs/users/api/playlist.rst
```diff
@@ -0,0 +1,9 @@
+Playlists
+=========
+
+The playlists API allows for the creation and editing of lists of recordings
+
+.. autoflask:: listenbrainz.webserver:create_app_rtfd()
+   :blueprints: playlist_api_v1
```
alastair, line 7: we put the "/1/user/(playlist_user_name)/playlists" endpoint in the core blueprint, because we wanted it related to users, rather than under the /playlist namespace. However, from the perspective of a developer wanting to use the API, I think it makes sense to put all playlist endpoints in a single location.
Is there a way to individually add the other endpoints to here? (same with the user similarity endpoints)

docs/developers/spark-devel-env.rst
```diff
@@ -1,5 +1,5 @@
-Set up ListenBrainz Spark development environment
-=================================================
+ListenBrainz Spark development
+==============================
```
alastair, line 2: 1 line sentence saying what the spark environment is for?

The ListenBrainz Spark environment is used for computing statistics and computing recommendations

alastair: Maybe we don't need the 3 lines below this either? Just a 1 sentence introduction, and then the comment that says that you don't need this if you're just developing the website

alastair (commented):
This breakdown looks great, thanks!
I added a few comments about cleaning up the distribution of a few methods to make it easier to find them.
I think that eventually we can clean up some of the text, but I didn't focus on that in this review - let's get the organisation of the docs cleaned up first and then we can focus on that

docs/users/api/playlist.rst
amCap1712: Yes, we can manually exclude and include endpoints. I'll do the needful changes.

docs/users/api/core.rst
```diff
@@ -0,0 +1,66 @@
+Core
+====
+
+The ListenBrainz server supports the following end-points for submitting and fetching listens.
+
+.. autoflask:: listenbrainz.webserver:create_app_rtfd()
+   :blueprints: api_v1
```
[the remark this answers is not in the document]
amCap1712: To order the endpoints, we have two choices let it be default which seems like is the order in which they appear. Alternatively, we can ask for it to be ordered by the url of the endpoint. 

Yes to break a subsection, we either need to do it manually by specifying the endpoints individually or create a blueprint.

docs/users/api/playlist.rst
amCap1712: Where should we move the simlarity endpoints to?

alastair: I was thinking of duplicating these endpoints - keep them in core (grouped by user-specific endpoints), but also additionally add them to playlist, so that if people go to the playlist section they'll also see them.

In that case, we don't need to move the similarity endpoints anywhere - just leave them in core if we don't have another place for them.

PR #1829, opened 2022-01-23
Do not run CI on draft PRs

amCap1712:
We probably do not want CI to be run for draft PRs. I think its fine to disable it for all workflows but we can configure this on a per workflow basis if we want some jobs to run for draft PRs.

### DIFF
--- a/.github/workflows/build-prod-image.yml
+++ b/.github/workflows/build-prod-image.yml
@@ -5,6 +5,7 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ '*' ]
+    types: [ opened, synchronize, reopened, ready_for_review ]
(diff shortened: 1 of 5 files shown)


alastair:
can we have a link somewhere (commit message? comment?) which shows the possible values of `types`? By having an explicit list, there's a possibility that if new ones get added, we'll not handle them

amCap1712:
The solution is taken from https://github.community/t/dont-run-actions-on-draft-pull-requests/16817/13 and all types plus the default ones are listed at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request . I'll add the docs link to the commit and explain why/how the change works.

PR #1830, opened 2022-01-27
Make created NOT NULL and update values for historic rows

amCap1712:
created field was not always a part of listen table. at the time it was added, it was set to 'epoch' for pre-existing rows and left as nullable. however, it makes more sense to set created according to listened_at field. we can also add the not null constraint.


alastair (approved):
I see that `timescale/create_tables.sql` already has this set to NOT NULL, so looks good to me. In my local dev environment this column is already not null, I guess that's because I created the table with this file and then imported a dump.

I didn't know about the special timestamp value 'epoch', interesting!

I ran this on my local large database, and it ran fine. Looks good.

PR #1831, opened 2022-01-27
Fix handling of multiple items (pinned recordings) having same mbid/msid

amCap1712:
It is possible to that multiple items have same msid/mbid. for example, pinning the same recording twice. Since the dict is keyed by mbid/msid the corresponding value should be a iterable of all items having that mbid so that such items are handled. Added test for the same.

### DIFF
--- a/listenbrainz/db/msid_mbid_mapping.py
+++ b/listenbrainz/db/msid_mbid_mapping.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from typing import Iterable, Dict, List, TypeVar
 from typing import Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


listenbrainz/db/msid_mbid_mapping.py
```diff
@@ -79,47 +80,49 @@ def fetch_track_metadata_for_items(items: List[ModelT]) -> List[ModelT]:
...
 
     mapping_mbid_metadata, mapping_msid_metadata = load_recordings_from_mapping(mbid_item_map.keys(), msid_item_map.keys())
     _update_items_from_map(mbid_item_map, mapping_mbid_metadata)
     _update_items_from_map(msid_item_map, mapping_msid_metadata)
     return items
 
 
-def _update_items_from_map(models: Dict[str, ModelT], metadatas: Dict) -> Dict[str, ModelT]:
-    for _id, item in models.items():
+def _update_items_from_map(models: Dict[str, Iterable[ModelT]], metadatas: Dict):
```
alastair, line 111: comment saying that it updates `models` in place would be helpful

listenbrainz/db/msid_mbid_mapping.py
```diff
@@ -79,47 +80,49 @@ def fetch_track_metadata_for_items(items: List[ModelT]) -> List[ModelT]:
...
+            metadata = metadatas[_id]
+            item.track_metadata = {
+                "track_name": metadata["title"],
+                "artist_name": metadata["artist"],
+                "release_name": metadata["release"],
+                "additional_info": {
+                    "recording_msid": metadata["recording_msid"],
+                    "recording_mbid": metadata["recording_mbid"],
+                    "release_mbid": metadata["release_mbid"],
+                    "artist_mbids": metadata["artist_mbids"]
```
alastair, line 126: not directly related to this PR, but are we always going to be sure that the `metadata` dict has all of these fields?

listenbrainz/db/msid_mbid_mapping.py
```diff
@@ -79,47 +80,49 @@ def fetch_track_metadata_for_items(items: List[ModelT]) -> List[ModelT]:
...
     msid_metadatas = load_recordings_from_msids(msid_item_map.keys())
     for metadata in msid_metadatas:
         msid = metadata["ids"]["recording_msid"]
-        item = msid_item_map[msid]
-        item.track_metadata = {
-            "track_name": metadata["payload"]["title"],
-            "artist_name": metadata["payload"]["artist"],
-            "additional_info": {
-                "recording_msid": msid
+        for item in msid_item_map[msid]:
```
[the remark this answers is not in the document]
amCap1712: Nothing is deleted from MsB ever so don't think this should be an issue. Also, if this is missing we need to fix code elsewhere as well because other places in codebase assume presence of title, artist name.

listenbrainz/db/msid_mbid_mapping.py
amCap1712: Yes, these are the exact fields selected in `load_recordings_from_mapping` (above in this file) and this function is only called with data from that query.

alastair (approved):

PR #1832, opened 2022-01-28
Update listen additional_info metadata added to spotify listens

alastair:
We updated the documentation for what metadata to put in a listen. This PR updates listens added by our spotify reader to follow this format.

We added the following new fields to the documentation, but I didn't add all of them:

* origin_url - added, same as spotify_id
* music_service - added, 'spotify.com' as per our guides
* submission_client - added, set to 'listenbrainz'

Not added:
* media_player, media_player_version - we don't know what client the user is using to listen, so don't fill this in. It's not always going to be the spotify app
* submission_client_version - does it make sense to have a "version" of the spotify reader?
* music_service_name - we proposed music_service (as a domain), and music_service_name only if you are unable to identify a domain, so I skipped this. We could add it in anyway if we wanted, but I don't think it's needed.

Deprecated fields:
* `'listening_from': 'spotify'` because it's replaced by `'music_service': 'spotify.com'`

### DIFF
--- a/listenbrainz/spotify_updater/spotify_read_listens.py
+++ b/listenbrainz/spotify_updater/spotify_read_listens.py
@@ -99,7 +99,6 @@ def _convert_spotify_play_to_listen(play, listen_type):
         'tracknumber': track.get('track_number'),
         'spotify_artist_ids': spotify_artist_ids,
         'artist_names': artist_names,
-        'listening_from': 'spotify',
         'discnumber': track.get('disc_number'),
         'duration_ms': track.get('duration_ms'),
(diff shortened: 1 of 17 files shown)


alastair:
Additional question - should we also add submission_client to the last.fm importer? I don't think we have data from this API to fill in any other fields.

amCap1712 (approved):
LGTM, Thanks!

amCap1712:
Adding the fields to Last.FM importer sounds good.

MonkeyDo (approved):
:rocket:

PR #1833, opened 2022-01-28
Do not read user_name from timescale

amCap1712:
The user_name column of listen table will soon go away. So we need to stop reading it from listen table in both dumps and fetch_listens. At both places, we should instead read user_names from primary LB db.

### DIFF
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -379,7 +379,17 @@ class TimescaleListenStore(ListenStore):
                         break
 
                     user_name = user_id_map[result["user_id"]]
-                    listens.append(Listen.from_timescale(**result, user_name=user_name))
+                    listens.append(Listen.from_timescale(
+                        listened_at=result["listened_at"],
+                        track_name=result["track_name"],
+                        user_id=result["user_id"],
+                        created=result["created"],
+                        data=result["data"],
+                        recording_mbid=result["recording_mbid"],
+                        release_mbid=result["release_mbid"],
+                        artist_mbids=result["artist_mbids"],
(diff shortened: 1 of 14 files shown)


listenbrainz/webserver/login/__init__.py
```diff
@@ -32,6 +32,16 @@ def from_dbrow(cls, user):
             login_id=user['login_id'],
         )
 
+    def to_dict(self):
```
amCap1712, line 35: Ideally, I'd have converted dicts to object but this way was less changes so this way for now.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -428,8 +431,8 @@ def fetch_recent_listens_for_users(self, user_list, limit=2, max_age=3600):
                 result = curs.fetchone()
                 if not result:
                     break
-
-                listens.append(Listen.from_timescale(*result[0:8]))
+                user_name = user_id_map[result["user_id"]]
+                listens.append(Listen.from_timescale(*result[0:8], user_name=user_name))
```
alastair, line 435: it'd be great to update calls to `from_timescale` to use the dict format from `result`, and explicitly list them all. This way we won't have unexpected errors if we happen to change the order of the columns in the select.

amCap1712: How about `**result`?

alastair: this would require the column names and parameter names in from_timescale to be the same, which might not be the case. I think that explicitly listing them might be safest.

amCap1712: Makes sense, fixed.

PR #1834, opened 2022-01-30
Migrate spark and spark dumps to user id

amCap1712:
We need to use user_id in spark as well so that stats remain correct even when users are renamed.


listenbrainz/db/stats.py
```diff
@@ -100,9 +100,10 @@ def insert_multiple_user_jsonb_data(data):
                   , {from_ts}
                   , {to_ts}
                   , NOW()
-               FROM (VALUES %s) AS t(user_name, count, stats)
-               JOIN "user"
-                 ON "user".musicbrainz_id = user_name
+               FROM (VALUES %s) AS t(user_id, count, stats)
+               -- this JOIN serves no other purpose than to filter out users for whom stats were calculated but
+               -- no longer exist in LB. if we don't filter, we'll get a FK conflict when such a case occurs
+               JOIN "user" ON "user".id = user_id
```
amCap1712, line 106: I guess this could be rewritten to use a CTE and WHERE but not sure if that's more readable?

mayhem: fine as is.

mayhem (approved):
Phew long, but fortunately not too complex!

alastair (approved):
looks fine to me too

listenbrainz_spark/recommendations/recording/candidate_sets.py
```diff
@@ -1,6 +1,6 @@
 """
 This script is responsible for creating candidate sets for all users. Candidate sets are dataframes containing
-(user_id, recording_id). The generated candidate sets will be given as input to the recommender that will assign ratings
+(spark_user_id, recording_id). The generated candidate sets will be given as input to the recommender that will assign ratings
```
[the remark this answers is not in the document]
amCap1712: spark_user_id is a consecutive integer. user_id may be non consecutive.

PR #1835, opened 2022-01-31
Update username whenever user logs in

amCap1712:
And with this change, we fully support username changes in LB . Earlier PRs migrated various components to use LB user id instead of username. With this, whenever a user logs in their username will be updated in LB so that website uses the new username as well.

### DIFF
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -516,7 +516,7 @@ def report_user(reporter_id: int, reported_id: int, reason: str = None):
 
 
 def update_user_details(lb_id: int, musicbrainz_id: str, email: str):
-    """ Update the email field for user with specified MusicBrainz ID
+    """ Update the email field and MusicBrainz ID of the user specified by the lb_id
 
     Args:
         lb_id: listenbrainz row id of the user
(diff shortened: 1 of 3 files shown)


listenbrainz/db/user.py
```diff
@@ -515,21 +515,24 @@ def report_user(reporter_id: int, reported_id: int, reason: str = None):
         })
 
 
-def update_user_email(musicbrainz_id, email):
+def update_user_details(lb_id: int, musicbrainz_id: str, email: str):
     """ Update the email field for user with specified MusicBrainz ID
```
alastair, line 519: update docstring

listenbrainz/db/user.py
```diff
@@ -515,21 +515,24 @@ def report_user(reporter_id: int, reported_id: int, reason: str = None):
         })
 
 
-def update_user_email(musicbrainz_id, email):
+def update_user_details(lb_id: int, musicbrainz_id: str, email: str):
```
alastair, line 518: does this method have a test? Looks like an easy win that we can add

amCap1712: Added.

listenbrainz/db/user.py
amCap1712: Done.

alastair (approved):

PR #1836, opened 2022-01-31
cron: fix feedback dump command

amCap1712:
Oh well! Broken for 8 months. In our defense, we haven't put these to use anyways ... :sweat_smile:

### DIFF
--- a/admin/create-dumps.sh
+++ b/admin/create-dumps.sh
@@ -95,7 +95,7 @@ elif [ "$DUMP_TYPE" == "incremental" ]; then
 elif [ "$DUMP_TYPE" == "feedback" ]; then
     SUB_DIR="spark"
 else
-    echo "Dump type must be one of 'full', 'incremental' or 'feedback'"
+    echo "ERROR: Dump Type $DUMP_TYPE is invalid. Dump type must be one of 'full', 'incremental' or 'feedback'"
     exit
 fi
 
--- a/listenbrainz/db/dump.py
+++ b/listenbrainz/db/dump.py
@@ -46,6 +46,7 @@ from listenbrainz.webserver import create_app
 MAIN_FTP_SERVER_URL = "ftp.eu.metabrainz.org"
 FULLEXPORT_MAX_AGE = 17  # days
 INCREMENTAL_MAX_AGE = 26  # hours
+FEEDBACK_MAX_AGE = 9  # days
 
 # this dict contains the tables dumped in public dump as keys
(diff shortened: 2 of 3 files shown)


admin/create-dumps.sh
```diff
@@ -95,7 +95,7 @@ elif [ "$DUMP_TYPE" == "incremental" ]; then
 elif [ "$DUMP_TYPE" == "feedback" ]; then
     SUB_DIR="spark"
 else
-    echo "Dump type must be one of 'full', 'incremental' or 'feedback'"
+    echo "ERROR: Dump Type $DUMP_TYPE is invalid. Dump type must be one of 'full', 'incremental' or 'feedback'"
```
amCap1712, line 98: Whenever I see this message in the logs, its not entirely obvious to me something failed. I have to come back to the script to check it. So print ERROR to make it obvious.

alastair:
We have the checker script for spark and regular dumps - does it make sense to add all types of dumps to this checker?

amCap1712:
Sure, sounds good. Added the check and updated PR.

listenbrainz/db/dump.py
```diff
@@ -46,6 +46,7 @@
 MAIN_FTP_SERVER_URL = "ftp.eu.metabrainz.org"
 FULLEXPORT_MAX_AGE = 17  # days
 INCREMENTAL_MAX_AGE = 26  # hours
+FEEDBACK_MAX_AGE = 9  # days
```
alastair, line 49: looking at the size of a feedback dump, it's pretty small and so should dump really quickly. maybe 8 days?

alastair (approved):
This specific change looks fine.

I understand that the feedback recommendations are only used to import into spark to tweak recommendations, is that right?
It seems a bit strange to have a folder `spark/` in the download archive that doesn't actually contain the (daily incremental?) spark updates, in fact the files aren't even parquet. What do you think about options for naming this dump differently, or including it in a different dump type? (we can open a ticket for this to do it later)

listenbrainz/db/dump.py
amCap1712: I had chosen 9 origianally because the check_dumps scripts run only once daily so was unsure whether the time difference of 8 days would be sufficient. I  now checked manually, it is. So changed to 8.

amCap1712:
We currently do not use these dumps anywhere. Its likely we'll revamp this before we start using feedback dumps in spark. We fix these inconsistencies at that time. Will open a ticket so that we don't forget.

PR #1838, opened 2022-02-02
Mention cause of potential 502s in Feedback API documentation

amCap1712:
We were facing these on the frontend (LB-993), the header size has now been increased but the error will still occur on large lists so mentioning in docs may be useful for future.


mayhem (approved):
Easy one!

PR #1839, opened 2022-02-02
LB-1058: use 'artist_identifiers' as a key for playlist track artists

alastair:
# Problem

Playlist json serialisation used 'artist_identifier' as the key for a list of artist urls, but there could be many of them and so the plural makes more sense.

# Action

This results in a change to the format of our API, which may affect people who are reading this data. Is this a concern for us?

See associated PR in Troi repo:
https://github.com/metabrainz/troi-recommendation-playground/pull/44


alastair:
converted to draft, waiting on related frontend changes by @MonkeyDo

listenbrainz/webserver/static/js/src/year-in-music/YearInMusic.tsx
```diff
@@ -151,6 +151,16 @@ export default class YearInMusic extends React.Component<
...
+          const trackExtension =
+            newTrack?.extension?.[MUSICBRAINZ_JSPF_TRACK_EXTENSION];
+          // See https://github.com/metabrainz/listenbrainz-server/pull/1839 for context
+          if (trackExtension && has(trackExtension, "artist_mbids")) {
+            //  Using some forbiddden (but oh so sweet) ts-ignore as we're fixing objects that don't fit the expected type
+            // @ts-ignore
+            trackExtension.artist_identifiers = trackExtension.artist_mbids;
+            // @ts-ignore
+            delete trackExtension.artist_mbids;
+          }
```
MonkeyDo, line 163: Protect the legacy !

A summary of the generated playlists is stored in the year-in-music table in the database, and with the change of JSPF key name we need to ensure we don't break the playlist cards on the YIM page.

Consider re-writing the playlists in the YIM table to use `artist_identifiers` instead of `artist_mbids` and nuking this code block in the future.

PR #1841, opened 2022-02-03
Add a simple playlist submit endpoint

mayhem:
We often need to have a random list of recording_mbid turned into a playable playlist for testing purposes. This adds the simple /playlist/create-simple endpoint that allows a single line POST that includes the recording_mbids separated by comma.

Returns playlist id, like the regular endpoint.

### DIFF
--- a/listenbrainz/webserver/views/playlist_api.py
+++ b/listenbrainz/webserver/views/playlist_api.py
@@ -364,12 +364,7 @@ def create_playlist_simple():
     for mbid in recording_mbids:
         playlist.recordings.append(WritablePlaylistRecording(mbid=UUID(mbid), added_by_id=user["id"]))
 
-    try:
-        playlist = db_playlist.create(playlist)
-    except Exception as e:
-        current_app.logger.error("Error while creating new playlist: {}".format(e))
-        raise APIInternalServerError("Failed to create the playlist. Please try again.")
-
+    playlist = db_playlist.create(playlist)
     return jsonify({'status': 'ok', 'playlist_mbid': playlist.mbid})
 
 
--- a/listenbrainz/webserver/views/playlist_api.py
+++ b/listenbrainz/webserver/views/playlist_api.py
@@ -328,6 +328,51 @@ def create_playlist():
     return jsonify({'status': 'ok', 'playlist_mbid': playlist.mbid})
 
 
+@playlist_api_bp.route("/create-simple", methods=["POST", "OPTIONS"])
+@crossdomain(headers="Authorization, Content-Type")
+@ratelimit()
+@api_listenstore_needed
+def create_playlist_simple():
+    """
+    Create a quick and dirty public playlist. The POST data must be nothing but a list of recording_mbids
+    separated by , all on one line.
+
+    :reqheader Authorization: Token <user token>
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/playlist_api.py
```diff
@@ -328,6 +328,51 @@ def create_playlist():
...
+    :statuscode 200: playlist accepted.
+    :statuscode 400: invalid recording_mbid list sent
+    :statuscode 401: invalid authorization. See error message for details.
+    :statuscode 403: forbidden. The submitting user is not allowed to create playlists for other users.
+    :resheader Content-Type: *application/json*
+    """
+
+    user = validate_auth_header()
+
+    data = str(request.get_data(), "utf-8")
```
amCap1712, line 350: Since this is POST request, you can instead accept json here. We should do that before we put it in prod. Fine either way for now. 

Maybe also accept the name/description of the playlist optionally here?

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -328,6 +328,51 @@ def create_playlist():
...
+    for mbid in recording_mbids:
+        playlist.recordings.append(WritablePlaylistRecording(mbid=UUID(mbid), added_by_id=user["id"]))
+
+    try:
+        playlist = db_playlist.create(playlist)
+    except Exception as e:
+        current_app.logger.error("Error while creating new playlist: {}".format(e))
+        raise APIInternalServerError("Failed to create the playlist. Please try again.")
+
+    return jsonify({'status': 'ok', 'playlist_mbid': playlist.mbid})
```
amCap1712, line 373: Remove status: ok for consistency with other endpoints.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -328,6 +328,51 @@ def create_playlist():
...
+                                creator_id=user["id"],
+                                description="Instant playlist -- nothing else is known. Can probably be deleted! :)",
+                                public=True)
+
+    for mbid in recording_mbids:
+        playlist.recordings.append(WritablePlaylistRecording(mbid=UUID(mbid), added_by_id=user["id"]))
+
+    try:
+        playlist = db_playlist.create(playlist)
+    except Exception as e:
```
amCap1712, line 369: We have an error handler configured to capture all unhandled (50x) errors, give the users a specific message and report it to sentry. That error handler is going to run next anyways because APIInternalServerError is raised which is a 50x error so the try/except clause is redundant here.

We have a bunch of places in the codebase where such redundant clauses exists, I usually remove the ones when I am modifying related code and also haven't added those to new endpoints.

listenbrainz/webserver/views/playlist_api.py
mayhem: Huh, this is copied from the playlist create endpoint. I'll remove it there too.

mayhem: All the functions on this module do this. lets clean this up later.

listenbrainz/webserver/views/playlist_api.py
mayhem: Ah, ok.

PR #1844, opened 2022-02-07
Check user_id has a corresponding user_name before accessing

amCap1712:
I had triggered full dumps after deploying user_id stuff so that spark stuff could be migrated. The normal listens dump at that time failed because a user_id didn't not have a corresponding user_name in the dict. I investigated and discovered that this was the cause:
- All usernames stored in a dict
- New user registers and submit listens
- Query fetches those listens, the user_id does not have a user_name causes a KeyError.

This is possible during incremental dumps too but there dump completes within 2-3 minutes so the issue is rare to occur there but full dumps take over 12 hours to complete so the chances are high. The current fix applies to both queries anyways.

This is a case I haven't previously thought out in depth, what happens when a user registers in the middle of a full dump and submits listen. For normal listens dumps, any behaviour is probably fine but for spark dumps this can cause inconsistent stats for the user for a while. Note that spark dumps are infact not affected by this because there is no use user_name there but still it is worthwhile to know what the current behaviour for this situation is in spark dumps.

For now, let's roll the current fix and keep dumps working.

### DIFF
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -465,7 +465,6 @@ class TimescaleListenStore(ListenStore):
                      FROM listen
                     WHERE listened_at >= :start_time
                       AND listened_at <= :end_time
-                      AND user_id != 0
                  ORDER BY listened_at ASC"""
         args = {
             'start_time': start_time,
@@ -484,7 +483,6 @@ class TimescaleListenStore(ListenStore):
                      FROM listen
                     WHERE created > :start_ts
                       AND created <= :end_ts
-                      AND user_id != 0
                  ORDER BY created ASC"""
 
         args = {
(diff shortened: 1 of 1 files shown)


mayhem (approved):

alastair (approved):
make sure we open a ticket to fix the larger underlying issue, but otherwise good 👍

alastair:
> This is a case I haven't previously thought out in depth, what happens when a user registers in the middle of a full dump and submits listen. 

I mentioned in IRC, adding a comment here for documentation: 
incremental updates already filter on `listen.created <= somevalue`, while full dumps use `listened_at`. We could update the query to be `where listened_at >= start and listened_at <= end and created <= end`, which means that we'll be sure that we don't include any listens submitted after we start the dump (and which perhaps may fix a small bug in which we dump listens in an incremental dump that are also included in the full dump)

PR #1845, opened 2022-02-07
Fix migration of user_name -> user_id in Spark related stuff

amCap1712:
Looking through sentry errors for the regular spark jobs, I found I had missed multiple places to change user_name -> user_id in spark and spark handlers. These changes are necessary to keep spark jobs working in prod and are currently deployed in prod as well.


mayhem (approved):
LGTM

PR #1847, opened 2022-02-09
Refactor fetch_listen methods

amCap1712:
We have 3 main entry points to fetch listens: user's get-listens, get-recent-listens endpoint and feed listen events. The first one only fetches the listens of 1 user and the other fetch listens for multiple users.

While working on migrating timestamp stuff, I saw that the feed listen events had a note to use an alternative query in the future. So it happens that query already exists in another name fetch_recent_listens_for_users. Modify the feed endpoint to use that method. Incidentally, this allows to simplify a chain of other fetch_listen methods.


mayhem (approved):
Thanks for these -- this cleans up long niggling crufty code. Happy!

PR #1849, opened 2022-02-09
Remove get_recent_listens_for_user_list api endpoint

amCap1712:
This endpoint is not used by LB frontend, and we found only a very low number of requests from 2 external users for this endpoint in logs. We reached out to one of those in the IRC channel and found that the requests were being triggered by CI for a client library. The request pattern of the other user also resembles that of a CI, so we believe there are no users for this endpoint and hence removing.


mayhem (approved):
We used to have a use for it -- it was an internal endpoint mostly, but I guess can can nuke it -- if we need it back, we know where to find it.

PR #1852, opened 2022-02-10
Refactor code Part One

akshaaatt:
A brainzplayer folder has been made and all the files related to it have been moved here.


listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
```diff
@@ -29,8 +29,8 @@ import {
 } from "../AlertNotificationsHOC";
 import APIServiceClass from "../APIService";
 import GlobalAppContext, { GlobalAppContextT } from "../GlobalAppContext";
-import SpotifyAPIService from "../SpotifyAPIService";
-import BrainzPlayer from "../BrainzPlayer";
+import SpotifyAPIService from "../brainzplayer/SpotifyAPIService";
```
MonkeyDo, line 32: As mentioned, this is going to need reverting

MonkeyDo (changes requested):
Thanks, that PR is a whole lot more manageable :)

There is one file that shouldn't be moved: SpotifyAPIService.ts
It's not used for BrainzPlayer, but rather for other components where we interact with Spotify (I think for now only to export playlists)

Apart from that, :rocket: !

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
akshaaatt: Done!

PR #1854, opened 2022-02-11
LB-1060: Follow button on followed users' pages doesn't show as "Following"

hrik2001:
# Problem
When viewing the page of a user you're following, the "Follow" button at the top only displays properly on the "Listens" tab. any other tab (Reports, Playlists, etc.) it appears as though you aren't following the user

# Solution
Added `logged_in_user_follows_user` to the props, so that the follow button can render correctly.

### DIFF
--- a/listenbrainz/webserver/views/user.py
+++ b/listenbrainz/webserver/views/user.py
@@ -446,6 +446,13 @@ def delete_listens_history(musicbrainz_id):
 
 
 def logged_in_user_follows_user(user):
+    """ Check if user is being followed by the current user.
+    Args:
+        user : User object
+    Raises:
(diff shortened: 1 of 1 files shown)


MonkeyDo (commented):
Thanks for finding this bug !

There is a lot of repeated code, i think it would make sense to make it a utility function and reduce the duplication.

hrik2001:
@MonkeyDo, btw pretty sure the failure is a false positive, since
- The tests pass locally
- I didn't edit that part of the backend
Very weird. If you want, I can look into it.

MonkeyDo (approved):
Thanks !
:rocket:

PR #1861, opened 2022-02-14
Refactor Codebase Part Four

akshaaatt:
- CB Reviews moved
- Missing MB data moved
- Report User moved
- LastFM moved


MonkeyDo (approved):
Yep, all good !
Thanks for part 4!
The codebase already looks a lot more organized

PR #1862, opened 2022-02-15
Bootstrap 5 upgrade

akshaaatt:
Get rid of local bootstrap v3 and replace it with the design system v5.

Please Hide Whitespace in order to review the PR with ease.


akshaaatt:
This PR got complicated because we tried to remove the bloated and extra CSS we have written all over. But seems like it will be rather hard to get rid of it and should be done gradually.

PR #1865, opened 2022-02-17
Redirect logged in users to Listens

akshaaatt:
Heading over to the Homepage seems redundant for already logged-in users. Hence it is best we redirect them to their Listens.


mayhem (approved):
I agree this is way overdue, but I think previously we agreed that we should redirect users to the feed page. But, lets just use this for now and when we have more stuff on the feed page, we can switch to it.

Aerozol:
Feed and Listens are both good options for landing pages imo.

07jasjeet:
> Feed and Listens are both good options for landing pages imo.

Reviving this pr and voting fir listens to be the homepage

PR #1866, opened 2022-02-17
Add note about max number of recording_mbids

mayhem:
We had some errors in our logs that we wanted to track down and found that too long URLs were being sent to the instant playlist endpoint. I've added a note about not sending more than 100 recording_mbids to the instant playlist page.

### DIFF
--- a/listenbrainz/webserver/views/player.py
+++ b/listenbrainz/webserver/views/player.py
@@ -74,6 +74,9 @@ def load_instant():
     the recording_mbid's metadata and creates a JSPF file from this data and sends it to the front end
     so a playlist can be instantly played.
 
+    NOTE: We recommend that you do not send more than 100 recording_mbids in one request -- our
+    server infrastructure will likely give you a gateway error (502) if you do.
+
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/views/player.py
```diff
@@ -74,6 +74,9 @@ def load_instant():
     the recording_mbid's metadata and creates a JSPF file from this data and sends it to the front end
     so a playlist can be instantly played.
 
+    NOTE: We recommend that you do not send more than 50 recording_mbids in one request -- our
```
amCap1712, line 77: If you write using `.. note::`, 
https://github.com/metabrainz/listenbrainz-server/blob/9e315466678033071c9ad5cf98d7abda8a6ccbb5/listenbrainz/webserver/views/feedback_api.py#L206-L208

reathedocs will render it nicely in a [blue box](https://listenbrainz.readthedocs.io/en/latest/users/api/recordings/#get--1-feedback-user-(user_name)-get-feedback-for-recordings) that stands out.

amCap1712 (approved):

PR #1868, opened 2022-02-17
Fix feed page timeout

amCap1712:
The feed page is timing out because the query to fetch listen events is timing out. This is happening because no min_ts and thus no lower bound exists for the listen query. To solve this we set min_ts to 14 days earlier if it is unspecified. In this particular case, we can get away with this workaround because the feed isn't intended to show old listen events anyway.


mayhem (approved):
Thanks for the swift fix!

PR #1869, opened 2022-02-17
Moving listen timestamps to database

amCap1712:
Currently, the listen timestamps are stored in redis. However, this is fragile and over the time many people have reported issues which could be traced to invalid timestamps. Listen counts already live in database so it makes sense to move timestamps to database as well.

TODO: Update frontend to show message the listen is not deleted immediately.


listenbrainz/listenstore/tests/test_timescale_utils.py
```diff
@@ -0,0 +1,107 @@
...
+        with timescale.engine.connect() as connection:
+            result = connection.execute(
+                text("""
+                    SELECT count, min_listened_at, max_listened_at
+                      FROM listen_user_metadata
+                     WHERE user_id = :user_id
+                """), user_id=user["id"])
+            return dict(**result.fetchone())
+
+    def _get_pending_deletes(self):
```
mayhem, line 47: A get function that returns nothing and prints stuff. Odd?

mayhem: I see that you have some debugging stuff leftover here. Are you confident enough to clean that up now?

listenbrainz/listenstore/timescale_utils.py
```diff
@@ -85,42 +277,16 @@ def add_missing_to_listen_users_metadata():
 
 
 def recalculate_all_user_data():
```
mayhem, line 279: Missing docs for this function.

mayhem: This function is intended to run only once, right?

listenbrainz/listenstore/timescale_utils.py
```diff
@@ -1,41 +1,228 @@
...
+        connection.execute(text(update_listen_min_ts))
+
+        logger.info("Update maximum listen timestamp affected by deleted listens")
+        connection.execute(text(update_listen_max_ts))
+
+        logger.info("Clean up listen delete metadata table")
+        connection.execute(text(delete_user_metadata), max_id=max_id)
+
+
+def update_user_listen_data():
```
mayhem, line 205: docs!

mayhem: This function never gets called except for tests and the cron job, yes?

mayhem (changes requested):
Looks quite good and I can follow most of it quite well. There are some bits of cleanup to do however...

listenbrainz/listenstore/timescale_utils.py
```diff
@@ -1,41 +1,228 @@
...
+    # count deleted listens, checked created and update listen counts
+    delete_listens_and_update_listen_counts = """
+        WITH deleted_listens AS (
+            DELETE FROM listen l
+             USING listen_delete_metadata ldm
+             WHERE ldm.id <= :max_id
+               AND l.user_id = ldm.user_id
+               AND l.listened_at = ldm.listened_at
+               AND l.data -> 'track_metadata' -> 'additional_info' ->> 'recording_msid' = ldm.recording_msid::text
+         RETURNING l.user_id, l.created
```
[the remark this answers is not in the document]
amCap1712: I'll need to check but why do we need to identify the listen? We only want to count the listens.

listenbrainz/listenstore/timescale_utils.py
amCap1712: Yes.

listenbrainz/listenstore/timescale_utils.py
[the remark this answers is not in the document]
mayhem: Ah yes, I see that now. Ignore me.

listenbrainz/listenstore/timescale_utils.py
amCap1712: Yes. Only need to run this when we we want to recalculate everything from scratch which should be only when we deploy this unless something goes wrong.

listenbrainz/listenstore/timescale_utils.py
```diff
@@ -1,41 +1,245 @@
...
+    with timescale.engine.begin() as connection:
+        result = connection.execute(select_max_id)
+        row = result.fetchone()
+
+        if not row:
+            logger.info("No pending deletes")
+            return
+
+        max_id = row["max_id"]
+        logger.info("Found max id in listen_delete_metadata table: %d", max_id)
```
mayhem, line 204: Is there any value to leaving these info statements?

mayhem (approved):
Thank you so much for getting this rather tricky task done! Its been a long journey, but I think this time we'll get much better results!

listenbrainz/listenstore/timescale_utils.py
amCap1712: Yes, I think it may come useful if something goes wrong during the cron job run. At least in tracking how much time each query is taking to execute.

amCap1712:
While moving timestamps, we also made deletes to not happen immediately but when the hourly cron job runs. So calling the delete-listen endpoint, adds the listens to `listen_delete_metadata` instead of deleting. The cron job picks it up every hour and does the acutal listen. 

We do so because we need to store the maximum and minimum listen timestamp of each user for efficiently querying their listens and to facilitate pagination accurately. Say a user deletes their latest listen then we need to update their maximum timestamp in the database. Doing this for check for every delete/insert is inefficient so instead we batch deletes and do it once in the cron job.

PR #1870, opened 2022-02-17
LB-912 (III): Add endpoints and UI to delete feed events

hrik2001:
# Problem
`ListenControl`'s `iconOnly` was bugging out. Also, a uniform UX was needed to give user's provision to delete notifications, recommendations and pins.

# Solution

Removed `iconsOnly` prop and also added trash icon in front of every events. (Greyed trash denotes non deletable events)


listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -607,6 +607,26 @@ export default class UserFeedPage extends React.Component<
...
+                            event.event_type === EventType.NOTIFICATION ? (
+                              <ListenControl
+                                title=""
+                                icon={faTrash}
+                                buttonClassName="btn btn-link"
+                                // eslint-disable-next-line react/jsx-no-bind
+                                action={this.deleteFeedEvent.bind(this, event)}
+                              />
+                            ) : (
+                              <ListenControl
```
MonkeyDo, line 623: So after seeing the page with all the garbage cans, I'm starting to have misgivings about the solution.

I propose one of two things:

1. Add the class `.btn-xs` to make all the buttons much smaller, therefore taking less space and being less conspicuous visually:
image


2. Put the trash icon somewhere else and only on events the user can delete.
The first thing that came to mind to me was simply after the event description:
image

We could also make it appear on the left-hand side, maybe when we hover over the event for example?
image



Sorry about the back and forth 😅 
What are your thoughts on this?

MonkeyDo (commented):
Thanks for working on this second part !

The ListenControl changes look great. This was the solution we needed all along, getting rid of the weird edge cases I put in there.

I have some doubts about my design suggestions though:

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -607,6 +607,26 @@ export default class UserFeedPage extends React.Component<
 
                           <span className="event-time">
                             {preciseTimestamp(created * 1000)}
+                            {((event.event_type ===
+                              EventType.RECORDING_RECOMMENDATION ||
+                              event.event_type === EventType.RECORDING_PIN) &&
+                              event.user_name === currentUser.name) ||
+                            event.event_type === EventType.NOTIFICATION ? (
+                              <ListenControl
+                                title=""
```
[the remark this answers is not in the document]
hrik2001: The title would make text appear besides the garbage can. What I think should be done is, use aria-label here. What do you say?

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
MonkeyDo: Follow up of this conversation in IRC, for posterity:
https://chatlogs.metabrainz.org/libera/metabrainz/2022-02-22/?msg=4957942&page=1

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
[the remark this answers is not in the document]
MonkeyDo: Ah, yes, absolutely :)
I forgot the `title` was added as text; that's annoying. Thanks, past monkey.

The aria label will be a good start, but it does not give us the feature that the title property does of showing text after hovering for a bit.
Let's leave it at that for now.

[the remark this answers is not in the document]
hrik2001: so should I add `title={ariaLabel}` at the button component in `ListenControl`? Or do you want to proceed with some other method to pass the button title

listenbrainz/webserver/static/js/src/listens/ListenControl.tsx
```diff
@@ -28,13 +29,14 @@ const ListenControl = (props: ListenControlProps) => {
...
     icon,
-    title,
+    text,
     dataToggle,
     dataTarget,
     disabled,
     link,
     anchorTagAttributes,
     ariaLabel,
+    title,
```
MonkeyDo, line 39: Considering how the component is used in ListenCard, where we duplicate title and text, how do you feel about defaulting the `title` to the value of `text`?
Something like this maybe?

`const titleAttribute = title ?? text;`

That way if you pass a text and are OK with the title being the same (which will probably happen most of the time) there's no need to pass a separate prop with the same content.

It would be worth adding a comment on the prop definition in ListenControlProps just to say it defaults to `text` if omitted.

MonkeyDo (commented):
Thanks for the title/text props, I think it makes sense to give us that flexibility.

There is another use of `{title}` that should be replace with `{text}` on line 48 inside the `if (link) {`  bloc

listenbrainz/webserver/static/js/src/listens/ListenControl.tsx
hrik2001: @MonkeyDo exactly! I had the same thought regarding this, since we took a very similar approach for `aria-label` in the same component. I had planned to ask you, but guess it slipped out of my mind!

Also, the `title` at line 48 was a big oopsie, my bad. Thanks a lot again.

listenbrainz/webserver/static/js/src/listens/ListenControl.tsx
```diff
@@ -43,9 +44,14 @@ const ListenControl = (props: ListenControlProps) => {
     // When using the link property,
     // render an anchor tag with an href instead of onClick
     return (
-      <a href={link} title={title} {...anchorTagAttributes}>
+      <a
+        href={link}
+        aria-label={ariaLabel ?? text}
```
MonkeyDo, line 49: We can even do this to make sure there is something, anything, in the aria label.
Same for below in the button.

And then we don't even need the `aria-label="Open in MusicBrainz"` anymore

```suggestion
        aria-label={ariaLabel ?? title ?? text}
```

MonkeyDo (commented):
Last suggestion, and ready to merge :)
Thanks for your patience !

MonkeyDo (approved):
Thank you so much !

PR #1871, opened 2022-02-18
Apply overall limit to listen events in feed

amCap1712:
Currently, feed includes max of 2 listen events per user you are following. If one is following a large number of users, then the
feed will get overwhelmed with just listen events. To avoid this situation, apply an overall limit of max 10 listen events across
users for the feed. Also, sorting it by newer first so that the newer listens are the ones shown if necessary to trim the list.


mayhem (approved):
Nice tweak, I approve!

PR #1872, opened 2022-02-18
Set custom handler for listenbrainz logger

amCap1712:
It turns out we forgot to set the handler on the listenbrainz logger. I was tracing why some log messages were missing during testing of the listen delete PR and noticed that adding the handler is missing. After adding it, the earlier missing log messages appear.


mayhem (approved):

alastair (approved):
Looks like this was accidentally removed in https://github.com/metabrainz/listenbrainz-server/commit/6a01bac94495dfdc7c93469e0aaa4458abc590de and https://github.com/metabrainz/listenbrainz-server/commit/f4c0b3b4cb599157c8382856c2be2be71fa0cc7d

good catch!

amCap1712:
:sweat:

PR #1874, opened 2022-02-19
Fix Unknown type: Decimal('value') error

amCap1712:
I was looking into LB-463 and checking why the count was still inaccurate since we moved to the database. Turns out we had an old key set without expiry for total listen count. I deleted it from redis so that a new would be calculated. However, it errored out as follows:

The SUM() function in postgres returns a DECIMAL value even if summing int columns. So psycopg2 returns the `value` as a DECIMAL type which is not recognized by msgpack/redis. This causes Unknown type: Decimal() error when trying to put the value in cache. To fix, cast the value to an int in python. Tested this fix on beta.lb.

So, I set the count manually in redis without an expiry key. Once the fix is merged, we need to remove the without expiry key so that total listen counts are recalculated on a regular basis.


mayhem (approved):
Thanks for tracking this down and how it happened. Well done and thanks!

PR #1877, opened 2022-02-21
Upgrade BU to v2.3.1

amCap1712:
1) Pin dependency versions to exclude versions incompatible with Flask
2) Update sentry_sdk initialization

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -4,9 +4,8 @@ import pprint
 import sys
 from time import sleep
 
-from brainzutils import cache, metrics
+from brainzutils import cache, metrics, sentry
 from brainzutils.flask import CustomFlask
-from brainzutils.sentry import init_sentry
 from flask import request, url_for, redirect
 from flask_login import current_user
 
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/__init__.py
```diff
@@ -82,11 +83,7 @@ def create_app(debug=None):
     if app.debug and app.config['SECRET_KEY']:
         app.init_debug_toolbar()
 
-    # Logging
-    app.init_loggers(
-        file_config=app.config.get('LOG_FILE'),
-        sentry_config=app.config.get('LOG_SENTRY')
-    )
+    init_sentry(**app.config.get('LOG_SENTRY'))
```
alastair, line 86:
```suggestion
    sentry.init_sentry(**app.config.get('LOG_SENTRY'))
```
with the associated change to import `sentry` as a module instead of the plain method (alongside cache/metrics)

alastair (approved):

listenbrainz/webserver/__init__.py
amCap1712: Fixed alongside checking if config is not None.

PR #1879, opened 2022-02-24
Add message to inform user of listen deletion policy

amCap1712:
LB website should inform the user that the listen is not deleted immediately but scheduled for deletion when the cron job runs next time. Also, put the error message in delete listen endpoint docs.


mayhem (approved):

MonkeyDo (approved):
LGTM.

As I mentioned on IRC, we *might* want to implement a system to cache in the user's browser which listens the were deleted so that if they refresh the page we can decide to hide them.
However I'm not convince it is worth the effort.

We could alternatively include the list of listens pending deletion in the RecentListens page props which is probably easier to implement.

amCap1712:
The pending listen thing sounds doable. I'll work on it in a separate PR.

PR #1881, opened 2022-02-25
Use self-hosted fonts from staticbrainz.org

amCap1712:
Keeping in mind ORG-51, we decided to self host google fonts on staticbrainz.org. The fonts themselves are now available on SB but we also need to modiffy the css `@import` urls and `link` to fonts.google.com. This is an initial version of that. I'll check and confirm there are no other usages in LB which need to be updated.

# TODO

- staticbrainz.org currently does not  add CORS headers for localhost so the fonts cannot be loaded in dev setup. This is not good so we should wait to fix that before merging.


alastair:
As we mentioned in IRC, musicbrainz hosts static assets from the dev server during local development. We could modify our `get_static_path` to handle both of these cases so that we don't need CORS headers for localhost.

PR #1882, opened 2022-02-26
Cleanup mbids lookup in MsB code

amCap1712:
First, mbid fields are a part of the dict of which hash is calculated when creating msids. Therefore if a listen has the same msid as another, the mbids fields are going to be same or missing in both. So there is no point in assigning mbids from the data  loaded from MsB.

Second, the code to do this lookup is broken. In timescale_writer, we update the listen's additional info with mbids fields. These fields are filled in load_recordings_from_msids where `"artist_mbids": []`. Therefore, the check len(artist_mbids) > 0 in timescale_writer always fails and the mbids are never assigned anyway.

Considering all this, it is better to remove these mbids from "ids". If someone ever needs it, they'll present in the payload key anyway.

### DIFF
--- a/listenbrainz/timescale_writer/timescale_writer.py
+++ b/listenbrainz/timescale_writer/timescale_writer.py
@@ -111,7 +111,7 @@ class TimescaleWriterSubscriber:
             if 'additional_info' in listen['track_metadata']:
                 ai = listen['track_metadata']['additional_info']
                 if 'artist_mbids' in ai and isinstance(ai['artist_mbids'], list):
-                    messy_dict['artist_mbids'] = ai['artist_mbids']
+                    messy_dict['artist_mbids'] = sorted(ai['artist_mbids'])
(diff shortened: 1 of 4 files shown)


listenbrainz/timescale_writer/timescale_writer.py
```diff
@@ -111,7 +111,7 @@ def messybrainz_lookup(self, listens):
             if 'additional_info' in listen['track_metadata']:
                 ai = listen['track_metadata']['additional_info']
                 if 'artist_mbids' in ai and isinstance(ai['artist_mbids'], list):
-                    messy_dict['artist_mbids'] = ai['artist_mbids']
+                    messy_dict['artist_mbids'] = sorted(ai['artist_mbids'])
```
amCap1712, line 114: LB-368

mayhem (approved):

PR #1700, opened 2021-10-31
Store listen counts in PG

amCap1712:
An inital draft- will need more iterations and testing to finalise. Design based on https://docs.google.com/document/d/1FRBilrLO3xy9BPoa0xRzEPpuk2y9mjNgPKlmwhvKrLE/edit . 

I tried a few queries and examined plans to estimate the costs invloved but the my local dev has limited data so probably best to create the tables in prod and run the query plans there instead.

PR #1887, opened 2022-02-28
Remove all traces of listen count cont. agg

amCap1712:
We used cont. agg for listen counts earlier. That has been replaced in #1700. The replacement has been working well so far, time to remove the cont. agg.


mayhem (approved):
delete delete delete!!

PR #1888, opened 2022-03-01
Add sitewide stats frontend

amCap1712:
Add dedicated pages to view sitewide stats. For the frontend side of changes, I have made the user prop optional in stat components. If present, that particular user's stats are loaded otherwise the sitewide stats are loaded.

WIP. Test Pending.

### DIFF
--- a/listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
+++ b/listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
@@ -12,7 +12,7 @@ import { isInvalidStatRange } from "./utils";
 
 export type UserArtistMapProps = {
   range: UserStatsAPIRange;
-  user: ListenBrainzUser;
+  user?: ListenBrainzUser;
   apiUrl: string;
 };
 
@@ -110,8 +110,13 @@ export default class UserArtistMap extends React.Component<
   getData = async (): Promise<UserArtistMapResponse> => {
     const { range, user } = this.props;
     try {
-      const data = await this.APIService.getUserArtistMap(user.name, range);
-      return data;
+      let data;
+      if (user) {
+        data = this.APIService.getUserArtistMap(user.name, range);
+      } else {
+        data = this.APIService.getSiteWideArtistMap(range);
+      }
+      return await data;
     } catch (error) {
       if (error.response && error.response.status === 204) {
         this.setState({
--- a/listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
+++ b/listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
@@ -110,13 +110,10 @@ export default class UserArtistMap extends React.Component<
   getData = async (): Promise<UserArtistMapResponse> => {
     const { range, user } = this.props;
     try {
-      let data;
-      if (user) {
-        data = this.APIService.getUserArtistMap(user.name, range);
-      } else {
-        data = this.APIService.getSiteWideArtistMap(range);
-      }
-      return await data;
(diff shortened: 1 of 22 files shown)


MonkeyDo (commented):
I think it would make sense to call a single API method from APIService.ts and let that method decide which endpoint to call depending on the presence of a username.

So for example, instead of `getUserListeningActivity` and `getSiteWideListeningActivity` duplicating a bunch of code, we could have a single `getListeningActivity` method with an optional `userName`, and do our if/else choice there:

```
let url;
if(userName){
  url = `${this.APIBaseURI}/stats/user/${userName}/listening-activity`;
}
else {
  url = `${this.APIBaseURI}/stats/sitewide/listening-activity`;
}
const response = await fetch(`${url}?range=${range}`);
```
I even moved common query parameters in the `fetch` call, but maybe that isn't necessary.

That would reduce the number of if/else logic scattered across components.
What do you think?


Another thought for later on: it would be great to add a way to go from user X's stats to sitewide stats.
I'm thinking of something near the heading of the page:

<h2>CoolUsername's stats
<br><a href="#" style="font-size: small">See sitewide stats</a></h2>

listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
```diff
@@ -110,8 +110,10 @@ export default class UserArtistMap extends React.Component<
   getData = async (): Promise<UserArtistMapResponse> => {
     const { range, user } = this.props;
     try {
-      const data = await this.APIService.getUserArtistMap(user.name, range);
-      return data;
+      return await this.APIService.getUserArtistMap(
+        user ? user.name : undefined,
```
MonkeyDo, line 114: I'm a fan of this shorter version which safely accesses a nested property, and returns undefined otherwise
This applies to other instances too:
```suggestion
        user?.name,
```

listenbrainz/webserver/static/js/src/utils/APIService.ts
```diff
@@ -363,13 +363,19 @@ export default class APIService {
   };
 
   getUserEntity = async (
-    userName: string,
+    userName: string | undefined,
```
MonkeyDo, line 366: Our convention in other files is to denote optional props like this (whihc ends up as string | undefined):
```suggestion
    userName?: string,
```

MonkeyDo (changes requested):
Looking great! I'm excited to see this feature out.

And on top of that, a PR I'm working on can make use of the new sitewide stats endpoints to show some stats on the Recent page, awesome timing :)

listenbrainz/webserver/static/js/src/utils/APIService.ts
amCap1712: Right, I had seen that at a few places and used it earlier but I got an error because its followed by entity which is not optional. So I had gone with this format everywhere. Changing all other instance in the PR except this one should be fine will do.

listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
amCap1712: Makes sense. fixed.

listenbrainz/webserver/static/js/src/utils/APIService.ts
amCap1712: Fixed.

mayhem (approved):
Python bits are fine.

MonkeyDo (approved):
GTG :rocket:
Looking forward to seeing this deployed :)

PR #1889, opened 2022-03-02
Fix CBReview modal MBIDs

MonkeyDo:
Currently, CBReviewmodal is broken.
Clicking on "write a review" for a listen that definitely has recording and artist MBIDs somewhere in its metadata (i.e from mbid_mapping), this is what appears:
image

The is some duplicated code to get MBIDs from listens that doesn't target the correct properties.
image

Instead, we should use existing utilities that were created for that purpose


akshaaatt (approved):
LGTM!

PR #1890, opened 2022-03-03
Fill right hand side column

MonkeyDo:
### Currently deployed on test.lb: https://test.listenbrainz.org/user/mr_monkey/

After redesigning BrainzPlayer in #1741, we are left with an empty column on the right.

We all agree making all pages full width looks bad (listen cards in full width = 💩 ), so we decided to fill the space with useful info.

Home page:
image

On smaller screen sizes when the columns take the full page width, the two columns are reordered to have the stats pins etc. first, and social features are hidden:
image

Charts page is made full width:
image

The content of other remaining pages where we couldn't think of relevant info to put in the right column have been centered (love/hate, pins, playlist):
image

The Recent page I want to do in a separate PR.
We want to show some selected sitewide stats in the right column, but it requires some refactoring.
Currently user's listens and recent listens are based on the same component with a "mode" sued to choose what to display.
This is a poopy legacy of when we didn't have much on those pages and now a blocker that needs to be refactored, and doing it in a separate PR will be better for everyone…


akshaaatt (approved):
Amazing work! Thanks for the fixes :)

PR #1891, opened 2022-03-04
LB-535: Refactor RecentListens component

MonkeyDo:
The RecentListens component was the first React component we created, and at the time we saved ourselves some sweat by reusing the component for different purposes ("Recent" page and user's listens page) with a `mode` property to conditionally render different layouts.

We've arrived at the point where this is no longer sustainable.
We added some user stats to the user's listens page in #1890 and now we want to add some sitewide statistics in the same spot.

Time to split the components !
The bulk of the RecentListens component and functionality is renamed to the file and component src/user/Listens.tsx

A new src/recent/RecentListens.tsx component is created which is a pared down version of Listens.tsx with a lot of functionality removed.

In a follow-up PR I will add sitewide statistics. better to wait until #1888 is merged, as it implements using stats components for both sitewide and user.


listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+                    const isListenReviewable =
+                      Boolean(recordingMBID) ||
+                      artistMBIDs?.length ||
+                      Boolean(trackMBID) ||
+                      Boolean(releaseGroupMBID);
+                    /* eslint-disable react/jsx-no-bind */
+                    const additionalMenuItems = (
+                      <>
+                        <ListenControl
```
amCap1712, line 174: Listens cannot be pinned unless they have either msid or mbid. So, a now playing listen without mbids submitted by the user cannot be pinned. So we should check that here.

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
```diff
@@ -0,0 +1,306 @@
...
+                          listen.track_metadata?.additional_info?.recording_msid
+                        )}
+                        updateFeedbackCallback={this.updateFeedback}
+                        newAlert={newAlert}
+                        additionalMenuItems={additionalMenuItems}
+                      />
+                    );
+                  })}
+                </div>
+                {!listens.length && (
```
amCap1712, line 212: Given Line 150 in this file, this seems to be wrong. We only need this message once, no?

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
```diff
@@ -0,0 +1,306 @@
...
+                onSuccessfulPin={(pinnedListen) =>
+                  newAlert(
+                    "success",
+                    "",
+                    `Successfully pinned ${pinnedListen.track_metadata.track_name}`
+                  )
+                }
+              />
+              <CBReviewModal
+                listen={recordingToReview || listens[0]}
```
amCap1712, line 233: `listens[0]` here feels a bit wrong. Maybe the component should handle null/undefined better?

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
```diff
@@ -0,0 +1,306 @@
...
+        const data = await APIService.getFeedbackForUserForRecordings(
+          currentUser.name,
+          recordings
+        );
+        return data.feedback;
+      } catch (error) {
+        if (newAlert) {
+          newAlert(
+            "danger",
+            "Playback error",
```
amCap1712, line 91: Error message looks wrong given context.

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
amCap1712: I realized that since this is RecentListens page, now playing listens won't appear here. But I'll probably get confused in future again about this. A comment mentioning why we don't need the check would be nice.

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -774,111 +763,109 @@ export default class RecentListens extends React.Component<
...
-                    </li>
-                  </ul>
-                )}
+                      &#x21E5;
+                    </a>
+                  </li>
+                </ul>
                 {currentUser && (
                   <PinRecordingModal
                     recordingToPin={recordingToPin || listens[0]}
```
amCap1712, line 871: I can't comment on the exact line due to Github limitations so writing here. The Review Modal just a  few lines below should be inside the currentUser check because not logged in users cannot write a review. Its that way in the new RecentListens.tsx already.

listenbrainz/webserver/views/index.py
```diff
@@ -139,12 +139,10 @@ def recent_listens():
 
     props = {
         "listens": recent,
-        "mode": "recent",
     }
 
     return render_template("index/recent.html",
         props=ujson.dumps(props),
-        mode='recent',
         active_section='listens')
```
amCap1712, line 146: active_section shouldn't be needed for recent listen page now.

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -774,111 +763,109 @@ export default class RecentListens extends React.Component<
                 {listens.length < this.expectedListensPerPage && (
```
amCap1712, line 763: Same as recent listens. looks redundant.

amCap1712 (approved):

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
amCap1712: Also, this comment is probably still relevant to Listens.tsx see Line 716 in that file.

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
MonkeyDo: I was thinking the same

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
MonkeyDo: Correct. And also this would never have been displayed as it was in a block displayed only if listens.length > 0

listenbrainz/webserver/static/js/src/user/Listens.tsx
MonkeyDo: In this case we can have some listens but not enough to fill the page (i.e new user only has 5 listens to show)
This is probably why RecentListens still had something similar still in there, I must have modified the condition to remove expectedListensPerPage, but it didn't make sense in Recent page.

akshaaatt (approved):
Looks good to me! Thanks for the updates :)

PR #1892, opened 2022-03-06
CORS: Always allow `Authorization` header

tandy-1000:
This would simplify things for those wanting to use the ListenBrainz API.

### DIFF
--- a/listenbrainz/webserver/decorators.py
+++ b/listenbrainz/webserver/decorators.py
@@ -13,9 +13,11 @@ def crossdomain(origin='*', methods=None, headers=None,
     if methods is not None:
         methods = ', '.join(sorted(x.upper() for x in methods))
     default_headers = {"AUTHORIZATION", "CONTENT-TYPE"}
-    headers.update(default_headers)
     if headers is not None and not isinstance(headers, string_types):
+        headers.add(x for x in default_headers)
         headers = ', '.join(x.upper() for x in headers)
+    else:
+        headers = default_headers
     if not isinstance(origin, string_types):
         origin = ', '.join(origin)
     if isinstance(max_age, timedelta):
--- a/listenbrainz/webserver/decorators.py
+++ b/listenbrainz/webserver/decorators.py
@@ -17,7 +17,7 @@ def crossdomain(origin='*', methods=None, headers=None,
         headers.add(x for x in default_headers)
         headers = ', '.join(x.upper() for x in headers)
     else:
-        headers = default_headers
+        headers = ', '.join(default_headers)
     if not isinstance(origin, string_types):
         origin = ', '.join(origin)
     if isinstance(max_age, timedelta):
--- a/listenbrainz/webserver/decorators.py
+++ b/listenbrainz/webserver/decorators.py
@@ -1,12 +1,14 @@
 from functools import update_wrapper, wraps
 from datetime import timedelta
+from typing import List
+
 from flask import request, current_app, make_response, redirect, url_for
 from six import string_types
 
 from listenbrainz.webserver import timescale_connection
 
 
-def crossdomain(origin='*', methods=None, headers=None,
+def crossdomain(origin='*', methods=None, headers: List[str] = None,
                 max_age=21600, attach_to_all=True,
                 automatic_options=True):
     # Based on snippet by Armin Ronacher located at http://flask.pocoo.org/snippets/56/.
@@ -15,7 +17,7 @@ def crossdomain(origin='*', methods=None, headers=None,
     all_headers = {"AUTHORIZATION", "CONTENT-TYPE"}
     if headers is not None:
         all_headers.update(headers)
-    all_headers = ', '.join(x.upper() for x in headers)
+    all_headers = ', '.join(x.upper() for x in all_headers)
     if not isinstance(origin, string_types):
         origin = ', '.join(origin)
     if isinstance(max_age, timedelta):
@@ -42,8 +44,7 @@ def crossdomain(origin='*', methods=None, headers=None,
             h['Access-Control-Allow-Origin'] = origin
             h['Access-Control-Allow-Methods'] = get_methods()
             h['Access-Control-Max-Age'] = str(max_age)
-            if all_headers is not None:
-                h['Access-Control-Allow-Headers'] = all_headers
+            h['Access-Control-Allow-Headers'] = all_headers
             return resp
 
         f.provide_automatic_options = False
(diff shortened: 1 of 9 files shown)


amCap1712:
We think a better way to handle this would be modify the crossdomain decorator so that we don't miss adding this method at any place and also avoid repetition. 

Set these Authorization and Content-Type as default allowed headers as default; allow more to be added (append to the list, don't overwrite it); make sure we dedup it in case we add one which is already a default.

I can make the change if its fine with you or help you with making it yourself.

tandy-1000:
> We think a better way to handle this would be modify the crossdomain decorator so that we don't miss adding this method at any place and also avoid repetition.
> 
> Set these Authorization and Content-Type as default allowed headers as default; allow more to be added (append to the list, don't overwrite it); make sure we dedup it in case we add one which is already a default.
> 
> I can make the change if its fine with you or help you with making it yourself.

Do you mean I should add these as the default here?
https://github.com/metabrainz/listenbrainz-server/blob/e0c7c5d129cafbb95a53432da75072779dc2c685/listenbrainz/webserver/decorators.py#L9

amCap1712:
Not there but inside the method because if set as default in the method parameter it will get overriden when the user specifies which header to allow.

https://github.com/metabrainz/listenbrainz-server/blob/e0c7c5d129cafbb95a53432da75072779dc2c685/listenbrainz/webserver/decorators.py#L15

You'll need to modify the code around here to always add Authorization, Content-Type headers. If the user already specified one of those headers, then make sure it isn't added twice.

listenbrainz/webserver/decorators.py
```diff
@@ -12,8 +12,12 @@ def crossdomain(origin='*', methods=None, headers=None,
     # Based on snippet by Armin Ronacher located at http://flask.pocoo.org/snippets/56/.
     if methods is not None:
         methods = ', '.join(sorted(x.upper() for x in methods))
+    default_headers = ["AUTHORIZATION", "CONTENT-TYPE"]
     if headers is not None and not isinstance(headers, string_types):
-        headers = ', '.join(x.upper() for x in headers)
+        for x in headers:
+            if x.upper() not in default_headers:
+                headers += ', '.join(x.upper())
```
amCap1712, line 19: `join` takes an iterable not str as input. If str is given as input then it'll instead split characters and join those. So `", ".join("ABC")` will become `"A, B, C"` which is incorrect.

tandy-1000: doh!

listenbrainz/webserver/decorators.py
```diff
@@ -12,8 +12,12 @@ def crossdomain(origin='*', methods=None, headers=None,
...
+    headers += ', '.join(x for x in default_headers)
```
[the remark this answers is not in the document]
tandy-1000: are you talking about L20?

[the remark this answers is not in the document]
tandy-1000: nvm i see it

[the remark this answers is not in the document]
tandy-1000: do you mean restrict headers type to an iterable?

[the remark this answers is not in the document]
amCap1712: Yes, as discussed in IRC. Let's restrict headers to an iterable and use a set instead for dedup.

listenbrainz/webserver/views/api.py
```diff
@@ -105,7 +105,7 @@ def submit_listen():
 
 
 @api_bp.route("/user/<user_name>/listens", methods=['GET', 'OPTIONS'])
-@crossdomain()
+@crossdomain(headers="Authorization")
```
amCap1712, line 108: Once the crossdomain changes work, these headers can be removed from everywhere.

tandy-1000:
Yay all green :) I think its ready now

listenbrainz/webserver/decorators.py
```diff
@@ -12,6 +12,8 @@ def crossdomain(origin='*', methods=None, headers=None,
     # Based on snippet by Armin Ronacher located at http://flask.pocoo.org/snippets/56/.
     if methods is not None:
         methods = ', '.join(sorted(x.upper() for x in methods))
+    default_headers = {"AUTHORIZATION", "CONTENT-TYPE"}
+    headers.update(default_headers)
     if headers is not None and not isinstance(headers, string_types):
         headers = ', '.join(x.upper() for x in headers)
```
amCap1712, line 18: This is a bit incorrect because headers can be None so that'll cause an error. The other way around i.e. `default_headers.update(headers)` can work because default_headers is never None. But then passing None to update can cause error.

I am almost sure this change here
```suggestion
    all_headers = {"AUTHORIZATION", "CONTENT-TYPE"}
    if headers is not None:
        all_headers.update(headers)
    all_headers = ', '.join(x.upper() for x in headers)
```

, removing the redundant check at L43 and changing L44 in the file to use `all_headers` variable should work.

listenbrainz/webserver/views/api.py
```diff
@@ -181,7 +181,7 @@ def get_listen_count(user_name):
 
 
 @api_bp.route("/user/<user_name>/playing-now", methods=['GET', 'OPTIONS'])
-@crossdomain()
+@crossdomain(headers="Authorization")
```
amCap1712, line 184: Other removals look good but this and 1 or 2 other instances of passing Authorization header in the file are present and should also be removed.

tandy-1000:
Hmm that doesn't seem to work, should I revert to this?

```
    default_headers = {"AUTHORIZATION", "CONTENT-TYPE"}
    if headers is not None and not isinstance(headers, string_types):
        headers.add(x for x in default_headers)
        headers = ', '.join(x.upper() for x in headers)
    else:
        headers = default_headers
```

listenbrainz/webserver/decorators.py
```diff
@@ -12,8 +12,10 @@ def crossdomain(origin='*', methods=None, headers=None,
     # Based on snippet by Armin Ronacher located at http://flask.pocoo.org/snippets/56/.
     if methods is not None:
         methods = ', '.join(sorted(x.upper() for x in methods))
-    if headers is not None and not isinstance(headers, string_types):
-        headers = ', '.join(x.upper() for x in headers)
+    all_headers = {"AUTHORIZATION", "CONTENT-TYPE"}
+    if headers is not None:
+        all_headers.update(headers)
+    all_headers = ', '.join(x.upper() for x in headers)
```
amCap1712, line 18:
```suggestion
    all_headers = ', '.join(x.upper() for x in all_headers)
```

PR #1897, opened 2022-03-10
LB-759, LB-962, LB-1047, LB-1074, LB-1075: Relinking Spotify throws `duplicate key` error

amCap1712:
For a long time we were unable to reproduce this error, @alastair  reproduced this reliably by double clicking the button causing spotify to send multiple payload to the callback and trying multiple inserts. There may be other possible reasons as well which we may not know of. ON CONFLICT DO UPDATE seems like a good thing to do in these cases.


alastair:
A test for this would be good, but otherwise looks fine to me

alastair (approved):

PR #1898, opened 2022-03-10
Some fixes for Missing MB Data page

amCap1712:
Originally discovered due to missing login_required decorator raising errors in sentry. But we decided that page should be public so fixes accordingly.

The page take user_name in url but renders the current user's missing MB data anyway. This is wrong. Some other minor code fixes/simplifications. There is no link to open the page for another user directly for now (the link in dropdown is for current user) but that can added in future as part of the redesign.


amCap1712:
Tested on test.lb.org, works fine.

MonkeyDo (approved):
Good to go, thanks !

amCap1712:
Thanks!

PR #1900, opened 2022-03-11
Use ListenCountCard component

amCap1712:
1) Localize listen count number
2) Improve listen count message for no listens
3) Use ListenCountCard component
4) In ListenCountCard, change h4 to p and change margin of hr for
 better formatting.
5) Add/update basic snapshots tests for ListenCountCard


listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
```diff
@@ -9,15 +9,17 @@ const ListenCountCard = (props: ListenCountCardProps) => {
...
-    <Card id="listen-count-card">
-      <h4>
-        You have listened to
-        <hr />
-      </h4>
-      <div>
-        <h4>{listenCount}</h4>
-        <small className="text-muted">songs so far</small>
-      </div>
+    <Card id="listen-count-card" style={{ fontSize: 24 }}>
```
MonkeyDo, line 12: It would be better to define font size using css instead of inline.
The `listen-count-card` id makes that easy.

listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
```diff
@@ -9,15 +9,17 @@ const ListenCountCard = (props: ListenCountCardProps) => {
...
+      {listenCount && (
+        <p>
+          You have listened to
+          <hr style={{ marginTop: 12, marginBottom: 12 }} />
```
MonkeyDo, line 16: Similarly, we can target `#listen-count-card tr {` for styling in css

MonkeyDo (changes requested):
Looks good, thanks for finding this in the codebase :)

Some suggestions to keep from mixing css and inline styling:

MonkeyDo (approved):
Awesome, thanks :) :rocket:

PR #1902, opened 2022-03-15
Upgrade to Python 3.10

alastair:
This replaces https://github.com/metabrainz/listenbrainz-server/pull/1649. Because of merge conflicts, I just applied the changes again to a new branch.

I updated some dependencies - these are ones that explicitly caused failures during build or just by running tests. I'm sure there are also other ones that can be upgraded at some time.

Most tests run, but I still saw a few failures. I'm not sure if these are some of our ghost failures that sometimes happen then go away, or if they are a problem.


amCap1712 (approved):
LGTM. Thanks!

PR #1904, opened 2022-03-16
Simplify @crossdomain decorator

amCap1712:
The @crossdomain decorators takes many parameters as input which we never use. We had copied it from the popular CORS snippet at some point. Since then some flask internal behaviour with regards to automatic options has changed as well. In any case, we do not see a use to maintain this complexity so removing all args and always use fixed set of constants. For methods, determine it from flask.


alastair (approved):
Tested using curl, I can see the cors headers on GET and on OPTIONS. looks good.

PR #1905, opened 2022-03-16
Allow multiple alpha values for request_model job

amCap1712:
Allowing multiple values let us compare multiple models and know which params work better.


mayhem (approved):
Things are going to get slowwwww, but hey, testing.

PR #1906, opened 2022-03-16
BrainzPlayer: show draggable youtube player on the page when playing with youtube

MonkeyDo:
In a separate little video screen, and not in the BrainzPlayer cover art thumbnail.
This is required for compliance with Youtube's requirements, and also better user UX.

image


amCap1712 (approved):
On mobile screens, a better UX would be if the player does not overlay the listen cards etc and does not always stay on screen (moves as per scroll). But even if we want to do that, it can come later. Thanks!

MonkeyDo:
> On mobile screens, a better UX would be if the player does not overlay the listen cards etc and does not always stay on screen (moves as per scroll). But even if we want to do that, it can come later. Thanks!

So you would have the video under the listencard?

What do you mean by "moves as per scroll"?
I'm keen to makes this better for mobile devices, but the only idea I've had so far is to take advantage of picture-in-picture functionality, which I know don't know enough about :p

MonkeyDo:
OK, I added the react-draggable library as a dependency, and added an icon allowing the user to drag the video anywhere on screen:

image

amCap1712 (approved):
LGTM, Thanks!

Aerozol:
Wow, I like the draggable option, with the icon. Very clear.
I think this is a good example of an annoying and restrictive requirement turning into a big + thanks to your care and attention Monkey!

PR #1907, opened 2022-03-16
BrainzPlayer: Warn before closing page if BrainzPlayer is playing

MonkeyDo:
Closing the page or navigating by mistake will stop music playback, which is pretty bad user experience.
[A user has suggested](https://community.metabrainz.org/t/introducing-the-new-listenbrainz-player/576230/3?u=mr_monkey) prompting before closing/navigating, which will also make sense when we have a SPA.

### DIFF
--- a/listenbrainz/webserver/static/js/src/brainzplayer/BrainzPlayer.tsx
+++ b/listenbrainz/webserver/static/js/src/brainzplayer/BrainzPlayer.tsx
@@ -180,7 +180,7 @@ export default class BrainzPlayer extends React.Component<
   componentDidMount = () => {
     window.addEventListener("storage", this.onLocalStorageEvent);
     window.addEventListener("message", this.receiveBrainzPlayerMessage);
-
+    window.addEventListener("beforeunload", this.alertBeforeClosingPage);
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/static/js/src/brainzplayer/BrainzPlayer.tsx
```diff
@@ -194,9 +194,23 @@ export default class BrainzPlayer extends React.Component<
...
     window.removeEventListener("storage", this.onLocalStorageEvent);
     window.removeEventListener("message", this.receiveBrainzPlayerMessage);
+    window.removeEventListener("beforeunload", this.alertBeforeClosingPage);
     this.stopPlayerStateTimer();
   };
 
+  alertBeforeClosingPage = (event: BeforeUnloadEvent) => {
+    const { playerPaused } = this.state;
+    if (!playerPaused) {
+      // Some old browsers may allow to set a custom message, but this is deprecated.
```
amCap1712, line 204: Yeah, the current FF already doesn't show custom message when I tried the PR :/

amCap1712 (approved):

PR #1909, opened 2022-03-17
Add CodeQL analysis

amCap1712:


amCap1712:
Didn't turn out to useful currently, will try someother time.

PR #1910, opened 2022-03-17
Reports world map: Show tooltip on click instead of hover

MonkeyDo:
The standard tooltips used in nivo components, including Choropleth (the world map we have on user reports page, only work on hovering over a country.
This prevents users from clicking the links to artists as the tooltip disappears when the mouse moves.

I also threw in some very minor refactoring (no change in functionality) while I was in there.

Instead of using the tooltip property to pass a custom tooltip component, we'll render and place said custom component ourselves !
This allows us to bind this to an onClick event (as opposed to onMouseEnter etc.) and listen to clicks on the outside to hide it.

This will look the same as currently deployed, but trust me, links are clickable now !
image

### DIFF
--- a/listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
+++ b/listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
@@ -179,6 +179,7 @@ export default class UserArtistMap extends React.Component<
             className="col-md-2 col-xs-4 text-right"
             style={{ marginTop: 20 }}
           >
+            <span>Rank by</span>
             <span className="dropdown">
               <button
                 className="dropdown-toggle btn-transparent capitalize-bold"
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/static/js/src/stats/UserArtistMap.tsx
```diff
@@ -196,7 +197,7 @@ export default class UserArtistMap extends React.Component<
                       this.changeSelectedMetric("listen", event)
                     }
                   >
-                    Listens
+                    number of listens
```
akshaaatt, line 200: I'm not sure about this change because a normal user wouldn't care too much about the long wordings here. For me just Listens or Artists was more clear. Also, the sentence can be capitalized (the first alphabet) if we were to continue with this.

akshaaatt (approved):
Thank you for fixing this! LGTM :)

PR #1911, opened 2022-03-18
Fix "does nothing if datasource already submits listens" test

amCap1712:
While investigating LFM importer issues, I was looking into usage of APIService.submitListens. I saw BP is no longer using it but one of it tests incorrectly spies on it. Fix the test by checking fetchMock instead.


MonkeyDo (approved):
Thanks for finding this !

PR #1912, opened 2022-03-18
Fetch recording data from MsB if not available in mbid_mapping

amCap1712:
We allow pinning recordings and submitting feedback if either msid or mbid is available. While retrieving data, this was the workflow before the PR.

1) Fetch msid only pins data from MsB
2) Fetch data from mapping for all pins (mbid only, msid only, mbid msid both).
    If mbid is available, we prefer that. For msid only pins, this means  mapping data will override MsB data if available.

However, this misses a case: A user submitted a listen with mbid to LB. LB assigns a msid to it but is unable to find mapping to it so the listens mbid doesn't exist in mapping table. If this listen is pinned, while retrieving data for it we'll only check mbid_mapping under table using mbid. So data will be missed leading to errors.

To handle this case the workflow is modified as follows:

1) Fetch data from mapping for all pins (mbid only, msid only, mbid msid both).
    If mbid is available, we prefer that. Store the pins we couldn't find data   for in mapping separately.
2) For all the pins which we couldn't find data for, lookup data in MsB.

There is still 1 rare case which is not handled by either worflows. Its described in comments but difficult to handle so not inclined to handle it unless someone is affected.

### DIFF
--- a/listenbrainz/db/msid_mbid_mapping.py
+++ b/listenbrainz/db/msid_mbid_mapping.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 from typing import Iterable, Dict, List, TypeVar
 from typing import Optional, Tuple
 
+from flask import current_app
 from pydantic import BaseModel, validator
 from sqlalchemy import text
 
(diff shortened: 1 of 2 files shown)


listenbrainz/db/msid_mbid_mapping.py
```diff
@@ -102,19 +111,32 @@ def fetch_track_metadata_for_items(items: List[ModelT]) -> List[ModelT]:
...
-
         for item in items:
+            # this means we could not find data for this item in the mapping tables. so we now add it to
+            # remaining items map using msid as the key because this data will be looked up from messybrainz
+            # afterwards.
+            if _id not in metadatas:
+                if item.recording_msid:
+                    remaining[item.recording_msid].append(item)
+                else:
+                    # well we are in the weeds now. we have an item which has only mbid, its msid is None. but we
```
amCap1712, line 130: The unhandled case.

mayhem (approved):

PR #1914, opened 2022-03-18
Cancel in progress jobs if new commit is pushed

amCap1712:
In almost all cases, we don't need the existing to complete if a new commit is pushed so cancel those.

As documented here: https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow


MonkeyDo (approved):
Great !

PR #1915, opened 2022-03-18
Add external service table to private dumps

amCap1712:
We learnt it the hard way. Lost user tokens due to mishandling of a spotify downtime. We maybe able to recover that from a barman backup but those are a pain to restore because we need to restore entire cluster for using those. Private dumps are easier to manage compared to barman backups.


mayhem (approved):
Thanks

PR #1916, opened 2022-03-18
More resilient access to nested track_metadata properties across the front-end

MonkeyDo:
In some rare cases, a track may not have a track_metadata object, and we don't want everything to break trying to access a nested property!
This PR creates two utility functions `getTrackName` and `getArtistName` to ensure we safely access `listen.track_metadata?.nested_prop`, and uses it across the website wherever we accessed `track_metadata?.track_name` or `track_metadata?.artist_name`.

In other places where we access `track_metadata.some_other_nested_prop` I made sure nothing breaks if there is no metadata (`track_metadata?.` to safely access nested prop)


amCap1712 (approved):
LGTM, Thanks!

PR #1922, opened 2022-03-19
MetaBrainz Design System v1.0.6 in ListenBrainz

akshaaatt:
- This PR focuses on replacing the locally added bootstrap version with the Design System bootstrap.
- bootstrap.min.js has been updated to support bootstrap version 5.1.3


MonkeyDo (commented):
We discussed these changes on IRC, decided to merge this initial PR into a separate branch as a lot of things are currently broken.

~~One thing in particular remains: the file `listenbrainz/webserver/static/js/lib/bootstrap.min.js` was deleted, but it is still loaded into the base template `listenbrainz/webserver/templates/base.html`.
The  PR description states "bootstrap.min.js has been updated to support bootstrap version 5.1.3", so perhaps we should update the file instead of deleting it?~~
I've got 💩 for eyes, it's updated, not deleted.

PR #1925, opened 2022-03-22
STFU typesense empty query message

mayhem:
Sometimes detuning metadata can leave nothing left at all and for obvs reasons typesense complains about it. Just ignore that noise and move on.

### DIFF
--- a/listenbrainz/mbid_mapping_writer/mbid_mapper.py
+++ b/listenbrainz/mbid_mapping_writer/mbid_mapper.py
@@ -183,7 +183,7 @@ class MBIDMapper():
 
             query = " ".join(cleaned_query)
 
-        if not query:
+        if query is None or query == "":
             return None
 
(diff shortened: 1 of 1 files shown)


amCap1712 (approved):

listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -183,6 +183,9 @@ def lookup(self, artist_credit_name_p, recording_name_p):
 
             query = " ".join(cleaned_query)
 
+        if query is None or query == "":
```
amCap1712, line 186: `if not query` ?

mayhem: Thats what I had, but it didn't work. I guess this is pointless, eh?

PR #1926, opened 2022-03-23
MB metadata cache

mayhem:
This PR implements the creation of a mb_metadata_cache table in the timescale DB. The components included are:

- Code to fetch the data using a GIANT query.
- Swap new table into production in one query.
- DB model, DB implementation
- Metadata blueprint and metadata fetch endpoint.

Not included in this PR is the ability to update this data in-place, but the provisions are in place to make it possible.


docker/docker-compose.yml
```diff
@@ -115,5 +115,5 @@ services:
 # TODO: re-comment these before merging this code
 #networks:
 #  default:
-#    name: musicbrainzdocker_default
+#    name: musicbrainz-docker_default
```
mayhem, line 118: This now matches what musicbrainz-docker does.

amCap1712 (approved):
LGTM, Thanks!

PR #1929, opened 2022-03-23
Refactor MBID Mapper and add a detuning case

amCap1712:
The PR refactors the MBID Mapper to improve readability. It also adds a case for detuning recording only but not artist credit which leads to matches being found in some cases where earlier match wasn't found. This is particularly useful for some spotify listens when spotify repeats the artist name in the recording name field as well.

Once the review is done I'll duplicate the file for use in the mapping endpoint as well. There will 2 be identical queries for the time being but the backend mapping query will get enhancements in future which may take a long time to run hence not suitable for use in the endpoint.

### DIFF
--- a/listenbrainz/mbid_mapping_writer/mbid_mapper.py
+++ b/listenbrainz/mbid_mapping_writer/mbid_mapper.py
@@ -68,8 +68,6 @@ class MBIDMapper:
             'connection_timeout_seconds': timeout
         })
         self.remove_stop_words = remove_stop_words
-        self.is_ac_detuned = False
-        self.is_r_detuned = False
 
     def _log(self, str):
         if self.debug:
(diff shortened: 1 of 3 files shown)


listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -221,65 +312,34 @@ def search(self, artist_credit_name, recording_name):
...
+                return hit
 
-            break
+        # this case is the last one because it didn't exist in earlier versions and
+        # preserving order of cases with older versions is probably sensible.
+        if r_detuned:
+            self._log("Detune only recording")
+            hit = self.lookup_and_evaluate_hit(artist_credit_name_p, r_detuned, False, True)
+            if hit:
+                return hit
```
amCap1712, line 340: This is the new case. Everything else in the PR is aimed to be a refactoring preserving existing behaviour.

listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -109,9 +107,45 @@ def compare(self, artist_credit_name, recording_name, artist_credit_name_hit, re
         self._log(Markup(f"""QUERY: artist: <b>{artist_credit_name}</b> recording: <b>{recording_name}</b>"""))
         self._log(Markup(f"""HIT: artist: <b>{artist_credit_name_hit}</b> recording: <b>{recording_name_hit}</b>"""))
 
-        return (distance(artist_credit_name, artist_credit_name_hit), distance(recording_name, recording_name_hit))
-
-    def evaluate_hit(self, hit, artist_credit_name, recording_name):
+        return distance(artist_credit_name, artist_credit_name_hit), distance(recording_name, recording_name_hit)
+
+    def check_hit_in_threshold(self, artist_credit_name, recording_name, ac_hit, r_hit, is_ac_detuned, is_r_detuned):
```
mayhem, line 112: docstring, please.

mayhem (approved):
Please proceed, but do add docstrings in the next phase. Also, I like your approach of doing the multiple passes over the data, while not hitting the DB more. Could be really promising!

listenbrainz/mbid_mapping_writer/mbid_mapper.py
amCap1712: Added.

PR #1931, opened 2022-03-27
Navbar Migration

akshaaatt:
- Bootstrap imports have been fixed in this PR so that JS can be worked on as well.
- The navbar broke after switching from version 3 to 5 and hence the necessary fixes have been made.


MonkeyDo:
Finally coming back to this, thanks for your patience !

This is what I see first thing after compiling everything and opening the landing page locally:
Either I'm doing something wrong, or you are :)
Any ideas? Would you mind trying to run it with a fresh local copy?

image

MonkeyDo:
A thought occurred regarding the non-collapsing navbar: could it be an issue with the JS rather than CSS/HTML?

MonkeyDo:
(Sorry, wrong button !)

listenbrainz/webserver/static/css/theme/variables.less
```diff
@@ -259,101 +257,6 @@
...
-@nav-tabs-active-link-hover-bg: @body-bg;
-@nav-tabs-active-link-hover-color: @gray;
-@nav-tabs-active-link-hover-border-color: #ddd;
-
-@nav-tabs-justified-link-border-color: #ddd;
-@nav-tabs-justified-active-link-border-color: @body-bg;
-
-// Pills
-@nav-pills-active-link-hover-bg: @component-active-bg;
-@nav-pills-active-link-hover-color: @white;
```
MonkeyDo, line 355: These don't seem related to the navbar itself (there are other nav elements like tabs etc.).
I think this results in the tabs looking broken (although I'm guessing we'll also replace those in another PR I gather):
image

listenbrainz/webserver/templates/navbar.html
```diff
@@ -1,96 +1,195 @@
...
-              <li><a href="{{ url_for('profile.export_data') }}">Export</a></li>
-              <li><a href="{{ url_for('user.missing_mb_data', user_name=current_user.musicbrainz_id) }}">Add Missing MusicBrainz Data</a></li>
-              <li><a href="{{ url_for('login.logout') }}">Sign out</a></li>
-            </ul>
-          </li>
+        <li class="nav-item dropdown">
+          <a class="nav-link dropdown-toggle" data-bs-toggle="dropdown" href="#">
+            {{ current_user.musicbrainz_id }} <span class="caret"></span>
+          </a>
+          <ul class="dropdown-menu" role="menu">
```
MonkeyDo, line 147:
```suggestion
          <ul class="dropdown-menu dropdown-menu-end" role="menu">
```

To ensure the menu is aligned to the right, otherwise if it is too wide i will make the page wider (and scrollable)
Here's what it looks like (with `width: max-content;` in the `.dropdown-menu` class):
image

MonkeyDo (commented):
Nice to see it all work!
Thanks for working on this, it's not an easy cycle you've started, good work so far :)

I have a few items of feedback, happy to discuss with you at your convenience:

1. Navbar dropdowns
I find the navbar dropdowns too narrow, which results in a lot of unnecessarily wrapping text:
image

Modifying the `.dropdown-menu` class to have a wider min-width  (for example `13rem` instead of `10rem`) or even using `width: max-content;`  gives the container a bit more width to avoid wrapping text, which results in less space on the page:
image

Considering these changes would have to be done in the design system (IIUC), they don't necessarily need to be done for this PR; I'm sure we will have other css changes to do and it seems somehow wasteful to deploy a new version for a one number change :)

2.  Navbar items hover
Did we decide to use the orange color as hover color for the menu items?
I did like that each project's navbar had a bit of flavor each with its own color.
Similarly, I'm not in love with the gray color of the navbar items' text or the font change, and the hover color should (probably?) be lighter and not darker. Grey/dark gray on orange is not very readable IMO; for comparison here it is with white text:
<span>image</span>
<span>image</span>

3. One last non-navbar related thing: I definitely definitely think the absence of any padding on the page is very jarring and looks like a mistake.

MonkeyDo:
Ah, found another couple of issues:

1. the navbar doesn't seem to collapse in small sizes. the hamburger button doesn't do anything.

2. in screen sizes between collapsed navbar and full width, the navbar doesn't shrink and some items are hidden behind a horizontal scroll:
image

listenbrainz/webserver/templates/navbar.html
akshaaatt: Fixed

akshaaatt:
> Ah, found another couple of issues:
> 
> 1. the navbar doesn't seem to collapse in small sizes. the hamburger button doesn't do anything.
> 2. in screen sizes between collapsed navbar and full width, the navbar doesn't shrink and some items are hidden behind a horizontal scroll:
>    image

Thank you! All fixed :)

MonkeyDo (changes requested):
That looks a lot better, thanks!

There is still [one comment from the previous review](https://github.com/metabrainz/listenbrainz-server/pull/1931/files#r876885676) that is still relevant: 
I'm not sure we should be removing the `nav-tabs-…` and `nav-pills-…`  LESS rules.
Navbar one yes, but nav ones are more general…

listenbrainz/webserver/templates/navbar.html
```diff
@@ -1,96 +1,195 @@
...
+                Your Year In Music
+              </a>
+            </li>
           </ul>
         </li>
-        <li class="dropdown">
-          <a class="dropdown-toggle" data-toggle="dropdown" href="#">
-             About<span class="caret"></span>
+        <li class="nav-item dropdown">
+          <a class="nav-link dropdown-toggle" data-bs-toggle="dropdown">
```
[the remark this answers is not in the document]
akshaaatt: Fixed

akshaaatt:
> That looks a lot better, thanks!
> 
> There is still [one comment from the previous review](https://github.com/metabrainz/listenbrainz-server/pull/1931/files#r876885676) that is still relevant: I'm not sure we should be removing the `nav-tabs-…` and `nav-pills-…` LESS rules. Navbar one yes, but nav ones are more general…

Makes sense! Fixed

MonkeyDo (approved):

PR #1933, opened 2022-03-30
Update dirty items in the metadata cache

alastair:
# Problem

Metadata cache items get out of date

# Solution
Find dirty items and perform the large metadata query on them.
In batches in a transaction delete the dirty items and replace them
with the newly queried data.

### DIFF
--- a/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
+++ b/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
@@ -381,11 +381,13 @@ def create_cache(mb_conn, mb_curs, lb_conn, lb_curs):
     log("mb metadata cache: drop old tables, create new tables")
     create_tables(lb_conn)
 
-    # Uncomment these 4 lines to select a subset of the musicbrainz database for testing, and use execute_values instead of execute.
-    #recording_mbids = ['e97f805a-ab48-4c52-855e-07049142113d', 'e95e5009-99b3-42d2-abdd-477967233b08', '97e69767-5d34-4c97-b36a-f3b2b1ef9dae']
-    #values = [(uuid.UUID(u),) for u in recording_mbids]
-    #query = get_metadata_cache_query(with_values=True)
-    config_postgres_join_limit(mb_curs)
+    # Uncomment these 4 lines to select a subset of the musicbrainz database for testing,
+    # and use execute_values instead of execute.
+    # recording_mbids = ['e97f805a-ab48-4c52-855e-07049142113d', 'e95e5009-99b3-42d2-abdd-477967233b08',
+    #                    '97e69767-5d34-4c97-b36a-f3b2b1ef9dae']
+    # values = [(uuid.UUID(u),) for u in recording_mbids]
+    # query = get_metadata_cache_query(with_values=True)
+    # config_postgres_join_limit(mb_curs)
 
     rows = []
     count = 0
--- a/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
+++ b/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
@@ -13,7 +13,7 @@ import config
 BATCH_SIZE = 5000
 
 
-def mark_rows_as_dirty(lb_conn, recording_mbids: List[uuid.UUID], artist_mbids: List[uuid.UUID]):
+def mark_rows_as_dirty(lb_conn, recording_mbids: List[uuid.UUID], artist_mbids: List[uuid.UUID], release_mbids: List[uuid.UUID]):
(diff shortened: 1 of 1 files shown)


listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -414,6 +431,67 @@ def create_cache(mb_conn, mb_curs, lb_conn, lb_curs):
     log("mb metadata cache: done")
 
 
+def delete_rows(lb_curs, recording_mbids):
+    query = """
```
mayhem, line 435: missing docstrings here and following two functions, please.

mayhem (changes requested):
some quick docstrings and this should be good to go. thanks!

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -414,6 +431,67 @@ def create_cache(mb_conn, mb_curs, lb_conn, lb_curs):
...
+        DELETE FROM mb_metadata_cache
+              WHERE recording_mbid IN %s
+    """
+    lb_curs.execute(query, (tuple(recording_mbids),))
+
+
+def config_postgres_join_limit(mb_curs):
+    # Because of the size of query we need to hint to postgres that it should continue to use joins
```
amCap1712, line 443:
```suggestion
    # Because of the size of query we need to hint to postgres that it should continue to
    # reorder JOINs in an optimal order. Without these settings, PG will take minutes to
    # execute the metadata cache query for even for 3-4 mbids. With these settings,
    # the query planning time increases by few milliseconds but the query running time
    # becomes instantaneous.
```

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -13,16 +13,17 @@
 BATCH_SIZE = 5000
 
 
-def mark_rows_as_dirty(lb_conn, recording_mbids: List[uuid.UUID], artist_mbids: List[uuid.UUID]):
+def mark_rows_as_dirty(lb_conn, recording_mbids: List[uuid.UUID], artist_mbids: List[uuid.UUID], release_mbids: List[uuid.UUID]):
```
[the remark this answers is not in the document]
amCap1712: I think fine to make these arguments optional but in any case need to check PG doesn't error if the list is empty.

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -13,16 +13,17 @@
...
     """Mark rows as dirty if the row is for a given recording mbid or if it's by a given artist mbid"""
     try:
         with lb_conn.cursor() as curs:
             query = """UPDATE mb_metadata_cache
                                SET dirty = 't'
                              WHERE recording_mbid = ANY(%s)
                                 OR %s && artist_mbids
+                                OR release_mbid = ANY(%s)
```
amCap1712, line 24: Replace = ANY with IN ? Both are semantically same afaik but we use IN everywhere else.

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -414,6 +431,67 @@ def create_cache(mb_conn, mb_curs, lb_conn, lb_curs):
...
+    recording_mbids = lb_curs.fetchall()
+
+    config_postgres_join_limit(mb_curs)
+
+    log(f"mb metadata update: got {len(recording_mbids)} to update")
+
+    log("mb metadata update: Running looooong query on dirty items")
+    query = get_metadata_cache_query(with_values=True)
+    values = [(row[0],) for row in recording_mbids]
+    psycopg2.extras.execute_values(mb_curs, query, values, page_size=len(values))
```
amCap1712, line 468: How many rows do you expect to update in one run of this job? Setting page size to insert all rows in 1 insert may be problematic, I vaguely remember issues in inserting LB stats or YIM when stats payload were huge. Though here the JSON will be much smaller so probably not an issue. In any case, having an estimate as to what we are expecting here would be nice.

amCap1712 (approved):
LGTM, Thanks!

mayhem:
https://tickets.metabrainz.org/browse/LB-1079 has been filed to address the issues in this PR. I'm going to merge, so I can continue working on this project.

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
amCap1712: Uh, I see. This cursor is iterated for results afterwards so page_size has to be eqaul to len otherwise won't work.

PR #1939, opened 2022-03-31
Add year to mb cache

mayhem:
NOTE: This PR is based off @alastair's branch.

This PR joins the canonical_musicbrainz_data table so that the year when the recording was released can be included in the release data of the cache. Also, automatically select a data sub-set if the USE_MINIMAL_DATASET config is True.

I realized that the canonical data was already being used here, so it will be a requirement to first build the full canonical data and save it to the MB instance that is being used to process the data, before the mb cache can be generated. Given that we hope to not update the cache often, this should be ok.


amCap1712 (commented):
LGTM, Thanks!

amCap1712 (approved):

PR #1941, opened 2022-04-03
[LB-1078] update API endpoint to /user/ 

abhijeet007rocks8:
# Problem
[API endpoint for user pins should be under /user/](https://tickets.metabrainz.org/projects/LB/issues/LB-1078?filter=allopenissues)
Currently, the API endpoint to get a user's pins is {}{{/1/$user_name/pins.
{}}}For consistency with other endpoints relating to a user's data, the endpoint should be {{/1/user/$user_name/pins }} with an extra /user/ in there.

Same goes for /pins/following

# Solution
Alter the endpoints on function handling the pings from `/<user_name>/pin` to `/user/<user_name>/pin` in files. 

# Action

### DIFF
--- a/listenbrainz/webserver/views/pinned_recording_api.py
+++ b/listenbrainz/webserver/views/pinned_recording_api.py
@@ -130,7 +130,7 @@ def delete_pin_for_user(row_id):
     return jsonify({"status": "ok"})
 
 
-@pinned_recording_api_bp.route("/<user_name>/pins", methods=["GET", "OPTIONS"])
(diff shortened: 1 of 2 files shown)


abhijeet007rocks8:
@MonkeyDo Can you review the PR.

amCap1712:
Hi! Thanks for the PR and sorry for the delay. But we can't merge this one yet, because this is an API breaking change which needs to be announced a few weeks in advance.

PR #1942, opened 2022-04-05
Hide Events Backend

hrik2001:
# Problem
Some events in your feed cannot be deleted, so a provision to hide few of them is needed. To achieve that, that too across all devices of a user, a backend is needed to store the events that a user might want to hide. This PR tries to do exactly that :)
## Eligibility of "Hidden Events"
Only `recording_recommendation` and `recording_pin` events that are created by the people you follow will be hidden by this endpoint.

# Solution
Added a new table, `hide_user_timeline_event`. Along with that, created a new pydantic data structure to represent rows in this table.

I also implemented `get_pins_by_id` and `get_user_timeline_by_id` functions, since there were no methods to fetch these specifically by their IDs. I believe these functions can help us later too for future features.

I also added two endpoints to hide and unhide events
`user/<user_name>/feed/events/hide` and `user/<user_name>/feed/events/hide/delete`


admin/sql/create_tables.sql
```diff
@@ -101,6 +101,14 @@ CREATE TABLE user_timeline_event (
   created               TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL
 );
 
+CREATE TABLE hide_user_timeline_event (
+    id           SERIAL,  --PK
+    user_id      INTEGER, --FK to "user"
+    event_type   hide_user_timeline_event_type_enum,
+    event_id     INTEGER,
```
amCap1712, line 108: I think it makes sense to add a comment here that this event_id is a row id. It can be a row id of a pin in the pins table or a timeline event in that table etc.

listenbrainz/db/user_timeline_event.py
```diff
@@ -168,3 +190,66 @@ def get_user_notification_events(user_id: int, count: int = 50) -> List[UserTime
...
         count=count
     )
+
+def hide_user_timeline_event(user_id: int, event_type: UserTimelineEventType, event_id: int) -> bool:
+    """ Adds events that are to be hidden """
+    try:
+        with db.engine.connect() as connection:
+            result = connection.execute(sqlalchemy.text('''
+                INSERT INTO hide_user_timeline_event (user_id, event_type, event_id)
+                    SELECT :user_id, :event_type, :event_id WHERE NOT EXISTS
```
amCap1712, line 200: Why not a simple insert? If you are trying to avoid duplicates, create a UNIQUE index and use ON CONFLICT DO NOTHING. We have some existing examples in the codebase for this, you can find them by searching for this clause.

listenbrainz/db/user_timeline_event.py
```diff
@@ -168,3 +190,66 @@ def get_user_notification_events(user_id: int, count: int = 50) -> List[UserTime
...
+                'user_id': user,
+                'event_type': event_type.value,
+                'count': count
+                }
+            )
+            return [HideUserTimelineEvent(**row) for row in result.fetchall()]
+    except Exception as e:
+        raise DatabaseException(str(e))
+
+def delete_hidden_timeline_events(user: int, row_id: int) -> bool:
```
amCap1712, line 238: I think unhide is a better name than delete_hidden.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -454,12 +581,21 @@ def get_recording_pin_events(users_for_events: List[dict], min_ts: int, max_ts:
                     )
                 )
             )
+            
+            hidden = False
```
amCap1712, line 585: Same  as above.

amCap1712: Actually also no need to 2 queries for this, my suggestion is do the database query for all events types at once and then in python build 2 dicts of different events and do the hiding.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -293,6 +295,109 @@ def delete_feed_events(user_name):
...
+        else:
+            raise APIBadRequest("This event type is not supported for hiding");
+    except (ValueError, KeyError) as e:
+        raise APIBadRequest(f"Invalid JSON: {str(e)}")
+
+
+@user_timeline_event_api_bp.route("/user/<user_name>/feed/events/hide/delete", methods=['OPTIONS', 'POST'])
+@crossdomain
+@ratelimit()
+def delete_hidden_user_timeline_event(user_name):
```
amCap1712, line 367: IMO, unhide instead of delete_hidden.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -293,6 +295,109 @@ def delete_feed_events(user_name):
...
+    .. code-block:: json
+
+        {
+            "event_type": "recording_pin",
+            "event_id": "<integer id of the event>"
+        }
+
+    :param user_name: The MusicBrainz ID of the user from whose timeline events are being deleted
+    :type user_name: ``str``
+    :statuscode 200: Successful deletion
```
amCap1712, line 324: Successful hide.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -293,6 +295,109 @@ def delete_feed_events(user_name):
...
+    For example:
+
+    .. code-block:: json
+
+        {
+            "id": "<integer row id of the hidden event>"
+        }
+
+    :type user_name: ``str``
+    :statuscode 200: Successful deletion
```
amCap1712, line 379: Something like even unhidden succesfully.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -293,6 +295,109 @@ def delete_feed_events(user_name):
...
+    try:
+        data = ujson.loads(request.get_data())
+        if data["event_type"] == UserTimelineEventType.RECORDING_RECOMMENDATION.value:
+            row_id = data["event_id"]
+            result = db_user_timeline_event.get_user_timeline_events_by_id(row_id);
+            if db_user_relationship.is_following_user(user['id'], result.user_id):
+                if db_user_timeline_event.hide_user_timeline_event(user['id'], data["event_type"], data["event_id"]):
+                    return jsonify({"status": "ok"});
+                else:
+                    raise APIBadRequest("Event already hidden");
```
amCap1712, line 345: @alastair thoughts on this? I'd say if user tries to hide twice, backend should ignore without raising APIBadRequest.

data/model/user_timeline_event.py
```diff
@@ -88,3 +88,11 @@ class APITimelineEvent(BaseModel):
     user_name: constr(min_length=1)
     created: NonNegativeInt
     metadata: APIEventMetadata
+    hidden: Optional[bool]
+
+class HideUserTimelineEvent(BaseModel):
```
amCap1712, line 93: I think `HiddenUserTimelineEvent` is better.

data/model/user_timeline_event.py
```diff
@@ -88,3 +88,11 @@ class APITimelineEvent(BaseModel):
...
+    id: NonNegativeInt
+    user_id: NonNegativeInt
+    event_type: UserTimelineEventType
+    event_id: NonNegativeInt
+    created: Optional[datetime]
```
amCap1712, line 98: Why optional, this column is NOT NULL in table schema?

admin/sql/updates/2022-03-32-hide-user-timeline-event.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
+
+CREATE TYPE hide_user_timeline_event_type_enum AS ENUM('recording_recommendation', 'recording_pin');
+
+CREATE TABLE hide_user_timeline_event (
+    id           SERIAL,  --PK
+    user_id      INTEGER, --FK to "user"
+    event_type   hide_user_timeline_event_type_enum,
+    event_id     INTEGER,
```
amCap1712, line 9: All of these columns can be NOT NULL?

data/model/user_timeline_event.py
hrik2001: Sorry my bad, I had put it there during testing, and forgot to remove it 😅

listenbrainz/webserver/views/user_timeline_event_api.py
hrik2001: hey @amCap1712 , wanted to ask, should the API endpoint remain the same, or should I change it to `events/unhide` instead of `events/hide/delete`.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -213,12 +214,32 @@ def user_feed(user_name: str):
...
+            # hidden_events_recommendation.append(hidden_event)
+            hidden_events_recommendation[hidden_event.event_id] = hidden_event
+        else:
+            hidden_events_pin[hidden_event.event_id] = hidden_event
+
+    for i in range(len(recording_recommendation_events)):
+        if recording_recommendation_events[i].id in hidden_events_recommendation:
+            recording_recommendation_events[i].hidden = True
+
+    for i in range(len(recording_pin_events)):
```
amCap1712, line 239: probably rewrite it as `for event in recording_pin_events:` and then `event.hidden = True` inside.

amCap1712: Same for above loop.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -213,12 +214,32 @@ def user_feed(user_name: str):
...
         count=count,
     )
 
+    hidden_events = db_user_timeline_event.get_hidden_timeline_events(user['id'], count)
+    hidden_events_pin = {}
+    hidden_events_recommendation = {}
+
+    for hidden_event in hidden_events:
+        if hidden_event.event_type.value == UserTimelineEventType.RECORDING_RECOMMENDATION.value:
+            # hidden_events_recommendation.append(hidden_event)
```
amCap1712, line 230: Comment can be removed.

listenbrainz/db/user_timeline_event.py
```diff
@@ -168,3 +190,62 @@ def get_user_notification_events(user_id: int, count: int = 50) -> List[UserTime
...
+                user_id = :user_id AND
+                event_type = :event_type AND
+                event_id = :event_id
+            '''), {
+                'user_id': user,
+                'event_type': event_type,
+                'event_id': event_id
+                }
+            )
+            return result.rowcount == 1
```
amCap1712, line 249: We don't verify in the API that a user hasn't already hidden an event so same event can have multiple entries in table. So better to rewrite as this > 0.

amCap1712 (changes requested):
Sorry for the very late review. Some nits. otherwise LGTM, I'll test on beta/test before merging.

listenbrainz/webserver/views/user_timeline_event_api.py
hrik2001: hi @amCap1712, if that is done, the changes wouldn't reflect in `recording_pin_events` list. Which is later being served in response. It would only change the event object in the current loop.

amCap1712: If you use `for event in recording_pin_events`, the variable event in the current loop will refer to the original object in the list so it should work fine.

For example, try running this snippet of code and see how the value changes even if we do it this way.

```python
class Container:
    
    def __init__(self, value):
        self.value = value
        
    def __repr__(self):
        return f"Container[{self.value}]"
    
    def __str__(self):
        return f"Container[{self.value}]"
        
items = [Container(i) for i in range(1, 4)]
print(items)

for item in items:
    item.value = item.value * 5
    
print(items)
```

hrik2001: Oops, I see, I didn't know objects behaved like that. I was merely thinking of them as regular datatypes in a list. Thank you so much @amCap1712 for taking your time to explain this. I have committed according to your review.

Once again thanks a lot!

amCap1712 (approved):
I tested on beta and it revealed a few issues with error handling. So I updated some code,  removed redundant error handling and added comment to tests. Hope that is fine.

Thanks again!

hrik2001:
@amCap1712, pleasure is all mine! Thanks a lot for being patient and taking your time out to help me out while working on this PR! It actually means a lot. Hope you have a good one 😃

PR #1943, opened 2022-04-06
Migrate MB metadata cache to bulk table

mayhem:
This PR moves the MB metadata cache to the bulk table system. In the process of doing so, I needed to add support for setting up the cursor prior to running the fetch for insert query and I needed to make the index specification process more general to allow for more complex types of indexes to be built.

NOTE: It looks as though the large query was changed, but that isn't actually the case. None of the internals have been reworked as part of this PR.


amCap1712 (approved):
LGTM, thanks!

PR #1944, opened 2022-04-07
Add Profile link to user navigation menu

da2x:
# Problem
There’s no obvious way to navigate to my own profile page after I’ve logged in to the service. The quickest method requires three clicks and three page loads from the front page: 1) click on the username menu: 2) click playlists: 3) click listens.

# Solution
Add a link to the profile page as the top item in the username menu. This is a design pattern used on lots of other websites with username-based drop down menus like this one.

# Action
This is completely untested. The code is my best guess at the correct implementation based on how other menu links are created and how profile links where created on the _similar users_ page.

### DIFF
--- a/listenbrainz/webserver/templates/navbar.html
+++ b/listenbrainz/webserver/templates/navbar.html
@@ -71,6 +71,7 @@
               {{ current_user.musicbrainz_id }} <span class="caret"></span>
             </a>
             <ul class="dropdown-menu" role="menu">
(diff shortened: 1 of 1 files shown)


MonkeyDo:
We do have a way! :)

The "Home" link on the left-hand side of the navbar (only visible when logged in) takes you to your listens page.

We're working on a redesign of the front-end pages (including navigation), so if this is not clear it's good information for us to know.

PR #1947, opened 2022-04-11
Use zip_longest in BulkInsertTable

amCap1712:
When using the BulkInsertTable class to build mapping tables, test values can be specified optionally to build a limited test table instead. If we don't want a test table, we return an empty list in get_insert_queries_test_values. This is zipped with get_insert_queries. This is problematic because zip stops at the end of the smallest list. So zipping with empty list will return an empty list and no queries will be executed.

Instead, use zip_longest which goes till the end of the longest list and substitutes None (configurable to something else) once no more elements are left in the shorter lists.


mayhem (approved):
thanks!

PR #1949, opened 2022-04-12
Automatically shut down the listenstore based on consul config value

alastair:
# Problem
Sometimes we need to shut down the timescale database, and we should stop the webserver and other tools from accessing it. Before now we have typically done this by rebuilding an image with the DB connection strings set to "" and then re-deploying. This takes extra work each time we need to do this process.

# Solution
Add a new consul config option to disable the datastore. If this is set to True, automatically set the connection string to "", even if the service is available and running

# Action

This just takes care of the webserver. At the moment the timescale writer is shut down manually, though we should be able to use this same process to make that shut down automatically too.

Mapping tools write to a schema in the timescale database, see if we can shut these down too.

### DIFF
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -53,21 +53,27 @@ MB_DATABASE_URI = 'postgresql://musicbrainz_ro@{{.Address}}:{{.Port}}/musicbrain
 MB_DATABASE_URI = "SERVICEDOESNOTEXIST_pgbouncer-slave_pgbouncer-master"
 {{end}}
 
-SQLALCHEMY_TIMESCALE_URI = ""
-TIMESCALE_ADMIN_URI=""
-TIMESCALE_ADMIN_LB_URI=""
 
+# Use a key 'listenstore_up' in consul (python boolean True or False) to decide if we connect to timescale.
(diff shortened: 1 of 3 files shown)


consul_config.py.ctmpl
```diff
@@ -53,22 +53,28 @@ MB_DATABASE_URI = 'postgresql://musicbrainz_ro@{{.Address}}:{{.Port}}/musicbrain
...
 {{end}}
 
-SQLALCHEMY_TIMESCALE_URI = ""
-TIMESCALE_ADMIN_URI=""
-TIMESCALE_ADMIN_LB_URI=""
 
+# Use a key 'listenstore_up' in consul (python boolean True or False) to decide if we connect to timescale.
+# If we set SQLALCHEMY_TIMESCALE_URI to an empty string, then the @api_listenstore_needed and @web_listenstore_needed
+# decorators will prevent the webserver from accessing the timescale database
+if {{template "KEY" "listenstore_up"}}:
```
alastair, line 60: This results in a python if statement in the rendered file:
```
if True:
    URI = "real_url"
else:
    URI = ""
```

I'm not sure if there's a difference here between doing it like this or using a consul if statement to render only one of them. Any preference?

amCap1712 (approved):
LGTM, Thanks!

PR #1950, opened 2022-04-13
LB-119: Add track duration as optional field to listens

YouzhiL:
# Problem

fix problem mentioned in https://tickets.metabrainz.org/browse/LB-119

# Solution

1.update doc to specify the type of duration
2.add validation of duration field submitted by clients
* `duration` and `duration_ms` will only accept integer type. 
* If both fields are given, `duration` will be ignored without raising an error.

# Action

failures not fixed:
https://gist.github.com/YouzhiL/cb0810c811fec50c5f6dbb38640c0e92

### DIFF
--- a/listenbrainz/tests/integration/test_api.py
+++ b/listenbrainz/tests/integration/test_api.py
@@ -476,7 +476,7 @@ class APITestCase(ListenAPIIntegrationTestCase):
         self.assertEqual(response.json['code'], 400)
     
     def test_multi_duration(self):
-        """ Test for duration and duration_both given """
+        """ Test for duration and duration_ms both given """
         with open(self.path_to_data_file('multi_duration.json'), 'r') as f:
             payload = json.load(f)
         response = self.send_data(payload)
--- a/listenbrainz/tests/integration/test_api.py
+++ b/listenbrainz/tests/integration/test_api.py
@@ -236,6 +236,13 @@ class APITestCase(ListenAPIIntegrationTestCase):
         self.assert401(response)
         self.assertEqual(response.json['code'], 401)
 
+    def test_valid_single(self):
+        """ Test for valid submissioon of listen_type listen """
+        with open(self.path_to_data_file('valid_single.json'), 'r') as f:
+            payload = json.load(f)
+        response = self.send_data(payload)
+        self.assert200(response)
+        self.assertEqual(response.json['status'], 'ok')
 
     def test_single_more_than_one_listen(self):
         """ Test for an invalid submission which has listen_type 'single' but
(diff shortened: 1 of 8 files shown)


listenbrainz/webserver/views/api_tools.py
```diff
@@ -263,6 +274,16 @@ def log_raise_400(msg, data=""):
     raise APIBadRequest(msg)
 
 
+def validate_duration_field(listen, key):
+    if key in listen['track_metadata']['additional_info']:
```
amCap1712, line 278: Hi! I, @alastair and @mayhem discussed it in IRC. We think its better to throw a Validation error if both duration and duration_ms are present. Also, if 1 is present and its value is not a positive integer we should throw a ValidationError,

listenbrainz/tests/integration/test_api.py
```diff
@@ -236,14 +236,6 @@ def test_unauthorized_submission(self):
         self.assert401(response)
         self.assertEqual(response.json['code'], 401)
 
-    def test_valid_single(self):
```
amCap1712, line 239: Deleted accidentally?

amCap1712 (changes requested):
Thanks for the PR. I looked at the failing tests. There is some issue with LFM comapt API which is coercing integers to strings. I'll try to debug it and share more details.

listenbrainz/webserver/views/api_tools.py
YouzhiL: Hi, Lucifer. Thanks for the advice. I modified the validation part and add new commits.

listenbrainz/tests/integration/test_api.py
YouzhiL: Yes, sorry for that. I add it back in new commits.

listenbrainz/webserver/views/api_tools.py
```diff
@@ -215,6 +215,17 @@ def validate_listen(listen: Dict, listen_type) -> Dict:
...
                     raise ListenValidationError("JSON document may not contain track_metadata.additional_info.tags "
                                                 "longer than %d characters." % MAX_TAG_SIZE, listen)
+        
+        
+        # duration
+        duration_key = ["duration", "duration_ms"]
+        for key in duration_key:
+            validate_duration_field(listen, key)
+        # if both duration and duration_ms are given and valid, an error will be raised.
+        if 'duration' in listen['track_metadata']['additional_info'] and 'duration_ms' in listen['track_metadata']['additional_info']:
```
amCap1712, line 225: This check should happen before trying to validate the duration fields.

listenbrainz/tests/integration/test_api.py
```diff
@@ -482,6 +481,26 @@ def test_invalid_recording_mbid(self):
...
         self.assert400(response)
         self.assertEqual(response.json['code'], 400)
+    
+    def test_multi_duration(self):
+        """ Test for duration and duration_ms both given """
+        with open(self.path_to_data_file('multi_duration.json'), 'r') as f:
+            payload = json.load(f)
+        response = self.send_data(payload)
+        self.assert400(response)
+        self.assertEqual(response.json['code'], 400)
```
amCap1712, line 491: Let's also verify the error message.

amCap1712 (changes requested):
Hi! Sorry for the late review. Some more changes needed but looks mostly good. I haven't looked into the LFM failures yet but will do that soon.

YouzhiL:
Hi! Thanks for the review. Sorry for the late commit. I was struggling with final exams. I modify the sequence of validation and verify the error message.

amCap1712 (approved):
Hi! No worries. The latest changes look good.

I just made some small changes before merging. I reworded the error message slightly so as to look like other messages. Also added another test. I hope that is fine.

I tested the PR on beta as well, works great. Thanks!

PR #1951, opened 2022-04-14
Improve model training output

mayhem:
This trivial PR improves two things:

1. Change the format of the html files so that they sort by date in the directory.
2. Add milligram CSS to the html files for simple prettififcation.


amCap1712 (approved):
LGTM, thanks!

PR #1953, opened 2022-04-14
Some enhancements to model training

amCap1712:
Log model params for each iteration and the best chosen model at the end. Also rename alpha to alphas.

### DIFF
--- a/listenbrainz_spark/recommendations/recording/train_models.py
+++ b/listenbrainz_spark/recommendations/recording/train_models.py
@@ -313,12 +313,10 @@ def save_training_html(time_, num_training, num_validation, num_test, model_meta
     save_html(model_html, context, 'model.html')
 
 
-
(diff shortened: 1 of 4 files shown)


listenbrainz_spark/recommendations/recording/train_models.py
```diff
@@ -328,7 +327,7 @@ def main(ranks=None, lambdas=None, iterations=None, alpha=None):
         logger.critical('model param "iterations" missing')
         raise
 
-    if alpha is None:
+    if alphas is None:
         logger.critical('model param "alpha" missing')
```
mayhem, line 331: nitpick, alphas in the error message.

mayhem (approved):

listenbrainz_spark/recommendations/recording/train_models.py
amCap1712: Fixed, thanks

PR #1954, opened 2022-04-14
[LB-1080] - Remove the bottom whitespace in pages with no BrainzPlayer

Fanny-Yuan:
Add style="padding: 0px;" to body

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -148,3 +148,5 @@ listenbrainz_spark/recommendations/html_files*
 
 # IDE configs
 .idea
+/.vs
+/listenbrainz/delete_listens.py
--- a/listenbrainz/webserver/static/css/main.less
+++ b/listenbrainz/webserver/static/css/main.less
@@ -22,8 +22,8 @@
 @brainzplayer-height: 60px;
 
 body {
-  // Leave some space for BrainzPlayer
-  padding-bottom: @brainzplayer-height;
+    // Leave some space for BrainzPlayer
+    //padding-bottom: @brainzplayer-height;
 }
 
 .table {
--- a/listenbrainz/webserver/templates/base.html
+++ b/listenbrainz/webserver/templates/base.html
@@ -1,4 +1,4 @@
-{% from 'macros.html' import print_message %}
+﻿{% from 'macros.html' import print_message %}
 
 <!DOCTYPE html>
 <html>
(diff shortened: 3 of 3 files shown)


MonkeyDo (changes requested):
Hi there! Thanks for opening a PR :)

There are a few issues with this PR:
1. This solution is applied to all pages, not just the ones where a player is present at the bottom of the page which is what the ticket is about. You would need to come up with a way to only apply these when a BrainzPlayer element is not in the page. That's not currnetly possible just with css, so you would have to find another way.
2. This would be better targeted with CSS classes rather than element `style` property; see for example [where the padding is currently defined](https://github.com/metabrainz/listenbrainz-server/blob/63826071c8/listenbrainz/webserver/static/css/main.less#L15-L17)
3. The changes you committed refactored the entire file unnecessarily, making the PR much harder to review. I'd suggest disabling your code editor's auto-formatting option before trying to get a clean PR with only the required changes.

Good luck, and let me know if you have questions.

Fanny-Yuan:
> Hi there! Thanks for opening a PR :)
> 
> There are a few issues with this PR:
> 
> 1. This solution is applied to all pages, not just the ones where a player is present at the bottom of the page which is what the ticket is about. You would need to come up with a way to only apply these when a BrainzPlayer element is not in the page. That's not currnetly possible just with css, so you would have to find another way.
> 2. This would be better targeted with CSS classes rather than element `style` property; see for example [where the padding is currently defined](https://github.com/metabrainz/listenbrainz-server/blob/63826071c8/listenbrainz/webserver/static/css/main.less#L15-L17)
> 3. The changes you committed refactored the entire file unnecessarily, making the PR much harder to review. I'd suggest disabling your code editor's auto-formatting option before trying to get a clean PR with only the required changes.
> 
> Good luck, and let me know if you have questions.

Hi Monkey, thank you for your suggestions! Before making a solution like this, I tried editing the main.less to change the padding from 60px to 0px. However, the page shows no change even if I refreshed it several times. It is so strange and I was still wondering why... So I tried another solution and it works (but the method is "rude" as it changes the all pages as you said.)

Fanny-Yuan:
@MonkeyDo Hi, Monkey! I edited the main.less to remove the padding instead of changing the body. Could you take a review of that? And sorry for the unclean PR last time.

MonkeyDo (commented):
Hello again @Fanny-Yuan,
thanks for giving this another shot.

We still have the issue I mentioned earlier: this change applies to all pages, which is not what we want.
Some pages have a player at the bottom (for example https://listenbrainz.org/feed/) while others do not (for example https://listenbrainz.org/similar-users/).
On the first page, the padding is necessary so that the player does not cover the page footer.
On the second page it is too much white space at the bottom since we don't have a player.

Honestly  I can't think of a clean and easy solution for this.
We have a plan for next month to refactor a lot of the front-end code (moving to a single-page application model) which should help with the matter, but until that is done I'm not sure how you could do this cleanly.
I would suggest putting this issue aside for now and we can revisit a better solution once we have refactored the front-end.

amCap1712:
We currently don't intend to solve it this way so closing for now. We should probably look at this again after the refactoring.

PR #1955, opened 2022-04-14
Add time to html filenames and add padding to the html output

mayhem:
Add time to html filenames and add padding to the html output of the html reports that are generated by the CF model training.


amCap1712:
LGTM.

PR #1956, opened 2022-04-15
BrainzPlayer: always submit playing-now listen when playing a new track

MonkeyDo:
We shouldn't rely on the Spotify listen ingestion cycle —which takes minutes and only getting longer as we grow— to get the currently playing track info out there.
Some of our pages as well as upcoming tools need to be more reactive.

Now that we're working on a tool to visualize metadata about the currently playing track, this is essential to the user experience.
I have a browser extension which submits a now-playing straight away and the reactiveness when playing on bandcamp for example feels absolutely great.

This PR implements the same for BrainzPlayer: submit a playing-now every time a new track starts regardless of if the service we are playing the track from is supposed to do this too through other means.


amCap1712 (approved):
LGTM, Thanks!

PR #1957, opened 2022-04-17
[LB-1086] ListenCountCard: Fix card message when visiting another user's page

chinmaykunkikar:
# Problem

Listen count says "You have…" instead of "'username' has" when visiting another user's page.
Ticket link - https://tickets.metabrainz.org/browse/LB-1086

# Solution

A `user` prop is passed to `ListenCountCard` component from `Listens` component.
ListenCountCard uses `GlobalAppContext` to get the current user and compares
the current user name to the user name from props to determine the
`isCurrentUser` boolean.

isCurrentUser is then used to conditionally render the message on the
ListenCountCard component.

## Screenshots

<details><summary>Screenshots</summary>

**1. When users see their own page**
When user sees their own page

**2. When no user is logged in**
When no user is logged in

**3. When a user visits another user's page**
When a user visits another user's page
</details>

### DIFF
--- a/listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
+++ b/listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
@@ -3,7 +3,6 @@ import Card from "../components/Card";
 import GlobalAppContext from "../utils/GlobalAppContext";
 
 export type ListenCountCardProps = {
-  isCurrentUser: Boolean;
   listenCount?: number;
   user: ListenBrainzUser;
 };
--- a/listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
+++ b/listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
@@ -1,25 +1,38 @@
 import * as React from "react";
 import Card from "../components/Card";
+import GlobalAppContext from "../utils/GlobalAppContext";
 
 export type ListenCountCardProps = {
+  isCurrentUser: Boolean;
   listenCount?: number;
+  user: ListenBrainzUser;
 };
 
 const ListenCountCard = (props: ListenCountCardProps) => {
-  const { listenCount } = props;
+  const { currentUser } = React.useContext(GlobalAppContext);
+  const { listenCount, user } = props;
+  const isCurrentUser = currentUser?.name === user?.name;
 
   return (
     <Card id="listen-count-card">
       {listenCount && (
         <div>
-          You have listened to
+          {isCurrentUser
+            ? "You have listened to"
+            : `${user.name} has listened to`}
           <hr />
           {listenCount.toLocaleString()}
           <br />
           <small className="text-muted">songs so far</small>
         </div>
       )}
-      {!listenCount && <p>You have not listened to any songs so far</p>}
+      {!listenCount && (
+        <p>
+          {isCurrentUser
+            ? "You have not listened to any songs so far"
+            : `${user.name} has not listened to any songs so far`}
+        </p>
+      )}
     </Card>
   );
 };
--- a/listenbrainz/webserver/static/js/tests/listens/ListenCountCard.test.tsx
+++ b/listenbrainz/webserver/static/js/tests/listens/ListenCountCard.test.tsx
@@ -1,15 +1,50 @@
 import * as React from "react";
 import { mount } from "enzyme";
 
+import GlobalAppContext, {
+  GlobalAppContextT,
+} from "../../src/utils/GlobalAppContext";
 import ListenCountCard from "../../src/listens/ListenCountCard";
+import APIService from "../../src/utils/APIService";
+
+const user = {
+  id: 1,
+  name: "track_listener",
+};
+
+const loggedInUser = {
+  id: 2,
+  name: "iliekcomputers",
+};
+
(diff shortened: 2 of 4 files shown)


listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
```diff
@@ -1,25 +1,38 @@
 import * as React from "react";
 import Card from "../components/Card";
+import GlobalAppContext from "../utils/GlobalAppContext";
 
 export type ListenCountCardProps = {
+  isCurrentUser: Boolean;
```
MonkeyDo, line 6: This prop isn't required, considering you user the `user` prop and the currentUser from the global context

MonkeyDo (changes requested):
Apart from a small change, the PR looks good.
Thank you for fixing this, it's been bugging me :)

MonkeyDo:
Could you also please add a couple of tests to `ListenCountCard.test.tsx` to ensure we are displaying everything correctly?

chinmaykunkikar:
@MonkeyDo **Update:**
1. Removed the redundant `isCurrentUser` prop.
2. Added tests and updated snapshots to make sure everything works as expected. Tests are passing locally 😄
3. _"Why the force push?"_. I wanted to format the description of the commit. Hope that won't be a problem?

listenbrainz/webserver/static/js/tests/listens/ListenCountCard.test.tsx
```diff
@@ -1,15 +1,50 @@
...
+    const wrapper = mount(<ListenCountCard user={user} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+  it("renders user's name instead of 'You' when visiting another user's page", () => {
+    const wrapper = mount(
+      <GlobalAppContext.Provider value={globalContext}>
+        <ListenCountCard user={user} listenCount={100} />
+      </GlobalAppContext.Provider>
+    );
+    expect(wrapper).toMatchSnapshot();
```
MonkeyDo, line 40: Snapshots are especially useful to make sure that the overall component renders as expected.
As you saw it's easy to update the snapshots, but harder to make sure nothing else in the snapshot file has changed that shouldn't have.

In this case, we want to test the rendered content directly in the test.
Instead of expect.toMatchSnapshot in the two new tests, we can ensure the name is rendered properly with the [enzyme library](https://enzymejs.github.io/enzyme/docs/api/) that we use to mount the components in the tests:

1. find the element by id
2. get the first child element
3. get the text content and compare it

Something like this: 
```
const outerCard = wrapper.find("#listen-count-card").first();
const innerDiv = outerCard.children().first();
expect(innerDiv.text()).toEqual("You have listened to 100 songs so far")
```

MonkeyDo (changes requested):
Thanks, looks great !

One last artificial hurdle for the test and we're good to go :)

listenbrainz/webserver/static/js/tests/listens/ListenCountCard.test.tsx
MonkeyDo: P.S: the removal of `expect(wrapper).toMatchSnapshot();` in the last two tests means you'll also have to update the snapshots again.

chinmaykunkikar: Hey! Thank you for the feedback. I understand what is to be done here.

Since the Listen count is surrounded by an `<hr />` and `<br />` respectively, they aren't considered as white space. So the expected text `"You have listened to 100 songs so far"` fails as it receives `"You have listened to100songs so far"` while running the test.

Code:
```html
<div class="card " id="listen-count-card">
    <div>You have listened to
        <hr />
        100
        <br />
        <small class="text-muted">songs so far</small>
    </div>
</div>
```

Is there any enzyme API that can be used for this? I have gone through their documentation, but no combination of their API is able to pass this test easily. All I can think apart from that is regex.

I do have a couple hacks to get around this -
1. Either test with `"You have listened to100songs so far"` to make it work. Probably not a good idea.
2. Or add white space before and after the Listen count in `ListenCountCard` component like so - 
```js
{` ${listenCount.toLocaleString()} `}
```

But again, these are hacks. We should write tests for the code and not the other way round.

I'd like to hear your thoughts on this.

chinmaykunkikar: I've also tried using `.html()` over `.text()` to match the HTML directly, but it didn't look like an elegant solution.

MonkeyDo: I think you are correct with the `.html()` solution, that was also my first thought.
I think that's elegant enough; in this case we can't easily test the text content, but we trust the HTML

chinmaykunkikar: Alright, using `.html()` then :)

chinmaykunkikar: @MonkeyDo Check out the new commit :)

MonkeyDo (approved):
Perfect, thank you! :)

PR #1867, opened 2022-02-17
Final Refactoring of the Frontend Codebase

akshaaatt:
- Last few files moved
- Mock Data moved
- Snapshots Updated

PR #1958, opened 2022-04-19
LB-1088: Separate North Korea and South Korea in artist origins map

MonkeyDo:
Yes, we're fixing Korea. No, we're not talking about international politics.

MusicBrainz has separate areas for North Korea and South Korea but the LB artist origins map does not distinguish between the two. We're fixing that.
I added the geoJSON entries for the separate countries, which I got from https://geojson-maps.ash.ms/

Also moved the GeoJSON (`world_countries.json`) file to stats UI folder, it was incorrectly moved to the tests folder during a repo structure refactor (#1867)

The result:

image
image


amCap1712 (approved):
LGTM, Thanks!

PR #1959, opened 2022-04-21
Do not harcode alpha in html report

amCap1712:
We now support multiple alphas for a run so display the alpha used for each model in the report.


mayhem (approved):
Thanks!

PR #1960, opened 2022-04-21
Use entire training data to generate candidate sets

amCap1712:
By default, candidate sets used to use a subset of listens to calculate top artists and similar artists. In experiments, this has led to worse results. Therefore, now changing the default to use entire training data instead of a subset.


mayhem (approved):
LGTM

PR #1961, opened 2022-04-22
Include CF model id and url with the recommendations

amCap1712:
In order to compare recommendations from various runs, we intend to create playlists of the recommendations linking back to the original model details that generated it. For this, metabrainz/troi-recommendation-playground#52 added a patch. Adding model id and model url to the api for the same.


mayhem (approved):
Great, thanks!

PR #1962, opened 2022-04-22
Trust input from approved bots

amCap1712:
troi-bot needs to add urls to the playlist descriptions which aren't in the allowlist. Its easier to disable all input verification for troi-bot instead of disabling checks incrementally when needed.


mayhem (commented):
OK, if test pass.

PR #1963, opened 2022-04-22
Add support for running troi after recommendations are run. 

mayhem:
This PR add the ability to run scripts as troi-bot once recommendations are complete. The only patch running right now is the one that sends a playlist to LB with a link to the model metadata.

The inner core of this code (which invokes troi) has been tested, but the rest... no idea how to do that.

### DIFF
--- a/listenbrainz/spark/troi_bot.py
+++ b/listenbrainz/spark/troi_bot.py
@@ -1,0 +1,22 @@
+""" This function contains code to run the various troi-bot functions after
+    recommendations have been generated.
+"""
+from troi.core import generate_playlist
+import config
+
+USER_TO_PROCESS = ["mr_monkey", "rob", "akshaaatt", "Damselfish", "lucifer", "alastairp", "CatCat", "atj"]
+
+def run_post_recommendation_troi_bot(user):
+    """
+        Top level function called after spark CF recommendations have been completed.
+    """
+
+    if user in USERS_TO_PROCESS:
+        make_playlist_from_recommendations(user)
+        # Add others here
+
+def make_playlist_from_recommendations(user):
+
+    token = config.WHITELISTED_AUTH_TOKENS[0]
+    for type in ["top", "similar"]:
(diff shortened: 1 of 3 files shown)


listenbrainz/spark/troi_bot.py
```diff
@@ -0,0 +1,22 @@
...
+        Top level function called after spark CF recommendations have been completed.
+    """
+
+    if user in USERS_TO_PROCESS:
+        make_playlist_from_recommendations(user)
+        # Add others here
+
+def make_playlist_from_recommendations(user):
+
+    token = config.WHITELISTED_AUTH_TOKENS[0]
```
amCap1712, line 20: We should use `flask.current_app` to get config here.

mayhem: Is this code running in a flask context??

amCap1712: Yup, entire spark reader thus all handlers run in flask context (https://github.com/metabrainz/listenbrainz-server/blob/28b27f9a9bd3af933aa42d53848391ecebe4152f/listenbrainz/spark/spark_reader.py#L127)

mayhem: Updated.

mayhem:
Thanks for catching the typos. Last week was really rough for me. :(

requirements.txt
```diff
@@ -23,6 +23,7 @@ pika == 1.2.0
 git+https://github.com/metabrainz/brainzutils-python.git@v2.3.1
 spotipy == 2.17.1
 git+https://github.com/metabrainz/data-set-hoster.git@v-2021-02-10.0#egg=datasethoster
+git+https://github.com/metabrainz/troi-recommendation-playground.git#egg=troi
```
amCap1712, line 26: Let's still pin it to some commit. lest we run the risk of breaking LB when working on troi.

mayhem: Done.

amCap1712: Thanks!

PR #1966, opened 2022-04-25
Navbar: Fix logo width

MonkeyDo:
Recent changes in logos 'broke' the navbar (making it wrap):

image

This PR makes sure the logo (with padding) is set to a specific width of 180px (not sure why we were at 181px before, decided to round it up). Adjusted the CSS to remove that 1 pixel.
image


amCap1712 (approved):

amCap1712:
LGTM, Thanks!

PR #1967, opened 2022-04-26
Metadata viewer: Part I

MonkeyDo:
In this episode, we introduce the new metadata viewer that allows user to see what they are currently playing.

This new page consists of a wrapper that fetches the currently playing listen and reacts to incoming playing-now listens, and the MetadataViewer itself which renders all that information.

We're not 100% of the way there, but we can deploy this advanced work in progress and keep the `/playing-now` route 'secret' until we are ready to unveil the feature.

Ideally, the same MetadataViewer component will be used by BrainzPlayer to show extra information about the track.

image

image


alastair (commented):
Python code looks good to me!

amCap1712 (approved):
Thanks!

PR #1968, opened 2022-04-27
Separate Missing MB data part into separate job

amCap1712:
Missing MB data is generated each time we create dataframes which is not ideal because we do a lot of testing about recommendations these days and do not want to spend extra time on these each time needlessly.

Also, dataframes are during similar users calculation as well which happens regularly and overwrites the data generated during recommendations run. By separating the run, we also avoid redundant work here.

### DIFF
--- a/listenbrainz/spark/request_manage.py
+++ b/listenbrainz/spark/request_manage.py
@@ -236,6 +236,14 @@ def request_dataframes(days, job_type, listens_threshold):
     )
 
 
+@cli.command(name="request_missing_mb_data")
+@click.option("--days", type=int, default=180, help="Request model to be trained on data of given number of days")
+def request_dataframes(days):
+    """ Send the cluster a request to generate missing MB data.
+    """
+    send_request_to_spark_cluster('cf.missing_mb_data', days=days)
+
+
 def parse_list(ctx, args):
     return list(args)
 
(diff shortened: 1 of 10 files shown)


listenbrainz_spark/missing_mb_data/missing_mb_data.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Send messages for data that has been submitted to ListenBrainz but is missing from MusicBrainz.
+    To avoid overwhelming amount of data, only top `TOP_LISTENS_LIMIT` number of listens for each user
+    are sent.
+
+    Args:
+        days: Number of days to calculate missing MB data from.
+    """
+    logger.info('Fetching listens to find missing MB data...')
+    to_date, from_date = get_dates_to_train_data(days)
```
mayhem, line 66: I suppose this function name ought to be updated during the spark nomenclature improvements.

mayhem (approved):
One minor docstring to fix, but looking good!

listenbrainz_spark/missing_mb_data/missing_mb_data.py
amCap1712: Yes, makes sense to do that. Currently haven't done that because this function is still used at other places.

listenbrainz/spark/request_manage.py
```diff
@@ -236,6 +236,14 @@ def request_dataframes(days, job_type, listens_threshold):
     )
 
 
+@cli.command(name="request_missing_mb_data")
+@click.option("--days", type=int, default=180, help="Request model to be trained on data of given number of days")
```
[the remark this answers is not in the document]
amCap1712: Fixed.

PR #1969, opened 2022-04-27
Replace usage of MLLib and RDDs with ML and DFs

amCap1712:
Spark ML is the newer machine library of spark and is based on dataframes instead of RDDs. This PR replaces usage of Spark MLLib the older version with the newer one. This also helps with replacing some hand rolled code with Spark built-in apis.


listenbrainz_spark/recommendations/recording/train_models.py
```diff
@@ -92,7 +58,7 @@ def preprocess_data(transformed_listencounts_df):
             test_data (rdd): Used for testing.
     """
     logger.info('Splitting dataframe...')
-    training_data, validation_data, test_data = transformed_listencounts_df.rdd.map(parse_dataset).randomSplit([4, 1, 1], 45)
+    training_data, validation_data, test_data = transformed_listencounts_df.randomSplit([4.0, 1.0, 1.0], 45)
```
mayhem, line 61: at least documenting these constants would be good.

mayhem (commented):
So far so good! I really appreciate how this simplifies things AND brings us to use the modern APIs, which match up when you search the docs. Big thumbs up!

listenbrainz_spark/recommendations/recording/train_models.py
amCap1712: Done.

mayhem (approved):
LGTM

PR #1975, opened 2022-04-29
Use TrainValidationSplit for tuning ALS Model

amCap1712:
Instead of hand rolling the code to do the training validation split, use Spark's inbuilt TrainValidationSplit API. Also, use ParamGridBuilder for running the training over different parameter combinations.


mayhem (approved):
LGTM!

PR #1976, opened 2022-04-29
Replace TrainValidationSplit with CrossValidation

amCap1712:
CrossValidation splits the training data multiple times as opposed to TrainValidationSplit which only does it once. Therefore, CrossValidation takes more time to train but can also help in less randomness in models.

### DIFF
--- a/listenbrainz_spark/recommendations/recording/tests/test_models.py
+++ b/listenbrainz_spark/recommendations/recording/tests/test_models.py
@@ -1,6 +1,5 @@
-import re
 from unittest import mock
-from unittest.mock import patch, Mock, MagicMock
+from unittest.mock import patch, MagicMock
 
 from listenbrainz_spark.recommendations.recording.tests import RecommendationsTestCase
 from listenbrainz_spark.recommendations.recording.train_models import Model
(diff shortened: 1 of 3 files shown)


listenbrainz_spark/recommendations/recording/tests/test_models.py
```diff
@@ -73,8 +72,9 @@ def test_get_best_model(self, mock_als_cls, mock_params, mock_tvs):
             estimator=mock_als_cls.return_value,
             estimatorParamMaps=mock.ANY,
             evaluator=mock_evaluator,
-            trainRatio=0.80,
-            collectSubModels=True
+            numFolds=5,
```
mayhem, line 75: Can we please document these values? And maybe make constants for them, if we anticipate changing them?

listenbrainz_spark/recommendations/recording/train_models.py
```diff
@@ -148,20 +155,21 @@ def train_models(training_data, evaluator, ranks, lambdas, iterations, alphas, c
...
         .build()
 
-    tvs = TrainValidationSplit(
+    cv = CrossValidator(
         estimator=als,
         estimatorParamMaps=params,
         evaluator=evaluator,
-        trainRatio=0.80,
-        collectSubModels=True
+        numFolds=5,
```
mayhem, line 162: Yes, lets make constants if they are being used in two places.

mayhem (approved):
One minor change, then good to go.

listenbrainz_spark/recommendations/recording/tests/test_models.py
amCap1712: Done

mayhem:
:+1:

PR #1977, opened 2022-05-02
LB-1085: Show track duration in ListenCard

chinmaykunkikar:
**Ticket - https://tickets.metabrainz.org/browse/LB-1085**

- Add a helper util method `getTrackDuration()` to get track duration from listen.

- Add css class `title-duration` that creates a flex row to render the track name and duration inline.
  Discussion about the layout - https://tickets.metabrainz.org/browse/LB-1085

- Update `ListenCard` to conditionally render track duration element.
  Convert ms to mm:ss using the millisecondsToStr helper method.

- Change `artistName` element from `<span />` to `<div />` to avoid unexpected behavior since span is an inline element tag.

- Add `listen.additional_info.duration_ms` property to the dummy `listen` object.

- Update snapshots by running `./test.sh fe -u`.

image


listenbrainz/webserver/static/css/listens-page.less
```diff
@@ -310,7 +310,12 @@
 				padding-right:initial;
 			}
 		}
-		
+
+		.title-duration {
+			display: flex;
+			flex-direction: row;
```
MonkeyDo, line 316: I would also add `align-items: center;` so that if the track title wraps around, the duration is vertically centered:

image

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -326,12 +329,24 @@ export default class ListenCard extends React.Component<
...
+                  <div className="small text-muted" title="Duration">
+                    {isNumber(trackDuration) &&
+                      millisecondsToStr(trackDuration)}
+                  </div>
+                </div>
+              ) : (
+                <div title={trackName} className="ellipsis-2-lines">
+                  {getTrackLink(listen)}
+                </div>
+              )}
```
MonkeyDo, line 346: I would suggest always rendering the title-duration parent element, and only conditionally render the duration element:

```suggestion
                <div className="title-duration">
                  <div title={trackName} className="ellipsis-2-lines">
                    {getTrackLink(listen)}
                  </div>
                  {trackDuration ? (
                    <div className="small text-muted" title="Duration">
                      {isNumber(trackDuration) &&
                        millisecondsToStr(trackDuration)}
                    </div>
                    )}
                </div>
```

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -326,12 +329,24 @@ export default class ListenCard extends React.Component<
...
-                {getTrackLink(listen)}
-              </div>
-              <span className="small text-muted ellipsis" title={artistName}>
+              {trackDuration ? (
+                <div className="title-duration">
+                  <div title={trackName} className="ellipsis-2-lines">
+                    {getTrackLink(listen)}
+                  </div>
+                  <div className="small text-muted" title="Duration">
+                    {isNumber(trackDuration) &&
```
MonkeyDo, line 338: How would you feel about adding a dash before the duration as a separator? Superfluous?

image

If so you would need a non-breaking space between the dash and the number (`&nbsp;` in HTML encoding): 
`-&nbsp;{isNumber(trackDuration) && …`

MonkeyDo (commented):
Great work ! Thank you !
Very clean PR, and a dearly missed feature that I'm personally excited about !

I've got some suggested improvements below, let me know what you think.

listenbrainz/webserver/static/css/listens-page.less
```diff
@@ -310,7 +310,12 @@
...
+			column-gap: 0.5rem;
```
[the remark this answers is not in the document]
MonkeyDo: Just in case, a translation of the less/css:
`>` :  Target direct children of title-duration
`*`: Target any element type (that way if we change the element type one day we won't be breaking everything)
`:not(…)`: pretty straightforward, inverts whatever condition is in the parens
`:first-child`: no explanation needed :)

All in all this translates to "target any and every child of title-duration except for the first one"

listenbrainz/webserver/static/js/src/utils/utils.tsx
```diff
@@ -137,6 +137,9 @@ const getReleaseGroupMBID = (listen: Listen): string | undefined =>
 const getTrackName = (listen?: Listen | JSPFTrack | PinnedRecording): string =>
   _.get(listen, "track_metadata.track_name", "") || _.get(listen, "title", "");
 
+const getTrackDuration = (listen?: Listen): number =>
+  _.get(listen, "track_metadata.additional_info.duration_ms", "");
```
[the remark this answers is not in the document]
chinmaykunkikar: Haha, I had originally put JSPFTrack in there but I wasn't sure about it so I got rid of it before the PR 😛

listenbrainz/webserver/static/css/listens-page.less
chinmaykunkikar: Noted.

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
chinmaykunkikar: I'll test this and let you know :)

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
chinmaykunkikar: While it sounds like a good idea, adding the dash can make it look out of place like in the first `ListenCard` of your screenshot.
I also think that the separator doesn't fit in the contemporary UI practices 😅

chinmaykunkikar:
@MonkeyDo I have made the suggested changes in the latest commit. Please take a look 😄

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
chinmaykunkikar: I couldn't commit this suggestion because the usage of the ternary operator is incorrect (missing `:`). But because of that I was able to test it locally. I have included it in the latest commit.

listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
MonkeyDo: I'm OK with that :)

I agree it looks weird with the wrapping text, that's why I put it in the screenshot.

chinmaykunkikar: I see :)

Then there's nothing else to do here :smile:

MonkeyDo (approved):
Works great, looks great ! Thank you ! :rocket:

PR #1979, opened 2022-05-03
LB-1079: Fix mb_metadata_cache update methods to work with bulk table code

alastair:
# Problem

* Some outstanding feedback on https://github.com/metabrainz/listenbrainz-server/pull/1933 which needed to be applied
* After migrating to the BulkInsertTable class, some update methods no longer worked

# Solution

Address feedback and fix update methods


mayhem (approved):
Thanks!

PR #1981, opened 2022-05-05
LB-1099: Do not show feedback cards without track name 

amCap1712:
LB-1099, discovered when a user's feedback page errored with ISE. When feedback is added, we only check that the mbid and msid are valid uuids but not whether the those actually exist in the database and refer to some track. Therefore, loosen the check on retrieval and ignore such MSIDs. Also, do not show feedback cards without track name.


MonkeyDo (commented):
Looks good apart from the change below :+1:

PR #1982, opened 2022-05-05
LB-1095: Ignore all dumps in .gitignore

chinmaykunkikar:
Current `gitignore` only ignores `listens` dumps, leaving out other dumps.
This change will ignore all kinds of dumps.

# Problem

https://tickets.metabrainz.org/browse/LB-1095

# Solution
Modify the pattern to ignore all dumps
```gitignore
listenbrainz-*dump*.tar.*
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,8 @@ env
 # tmp folder
 tmp/
 
-# Listenbrainz listens dump
-listenbrainz-listens*
+# Listenbrainz dumps
+listenbrainz-*dump*.tar.*
 
 # ListenBrainz application data
 docker/data
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ env
 tmp/
 
 # Listenbrainz dumps
-listenbrainz-*dump*.tar.*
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -10,8 +10,8 @@ env
 # tmp folder
 tmp/
 
-# Listenbrainz listens dump
-listenbrainz-listens*
+# Listenbrainz dumps
+listenbrainz-*dump*.tar.*
```
amCap1712, line 14: This one doesn't match the spark dumps which are in tar format only without gzip/bzzip2 compression, eg: `listenbrainz-spark-dump-853-20220415-040003-full.tar`. probably remove the `.` after tar and it should work fine in all cases then.

amCap1712 (commented):
Thanks for the PR. One minor change needed otherwise looks good.

chinmaykunkikar:
@amCap1712 Thanks for the review. I have changed the pattern according to your suggestion in my last commit :)

amCap1712 (approved):
LGTM, Thanks!

PR #1984, opened 2022-05-09
Add option to switch between playcounts and transformed listencounts

amCap1712:
Adding the option so that we can evaluate the usefulness of using a custom transformation function to transform listen counts before calculating confidence values from them.

### DIFF
--- a/listenbrainz_spark/recommendations/recording/create_dataframes.py
+++ b/listenbrainz_spark/recommendations/recording/create_dataframes.py
@@ -103,9 +103,12 @@ def save_dataframe_metadata_to_hdfs(metadata: dict, df_metadata_path: str):
         raise
 
 
-def describe_listencount_transformer():
-    """ Returns a human-readable description of the algorithm used to transform listen counts"""
-    return "No transformation applied to listen counts"
+def describe_listencount_transformer(use_transformed_listencounts):
+    """ Returns a human-readable description of the algorithm used to transform listen counts """
+    if use_transformed_listencounts:
+        return "No transformation applied to listen counts"
+    else:
+        return "No transformation applied to listen counts"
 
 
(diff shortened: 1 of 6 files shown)


listenbrainz_spark/recommendations/recording/create_dataframes.py
```diff
@@ -103,9 +103,12 @@ def save_dataframe_metadata_to_hdfs(metadata: dict, df_metadata_path: str):
         raise
 
 
-def describe_listencount_transformer():
-    """ Returns a human-readable description of the algorithm used to transform listen counts"""
-    return "No transformation applied to listen counts"
+def describe_listencount_transformer(use_transformed_listencounts):
+    """ Returns a human-readable description of the algorithm used to transform listen counts """
+    if use_transformed_listencounts:
+        return "No transformation applied to listen counts"
```
mayhem, line 109: I guess this string should be updated.

mayhem (approved):
Please update that one string and then after tests pass, feel free to merge. Thanks!

listenbrainz_spark/recommendations/recording/create_dataframes.py
amCap1712: Uh no, this PR only adds the switch for using playcounts or transformed listencounts. For this PR for both transformed listencounts and playcounts are same so the string is same in both cases. The next PR adds the transform function.

This is because we only merged the code to allow use of transformed listencount last time and not the actual transfer function as there was no switch to turn it off at that time and we wanted to play with both.

mayhem: Yes, I saw that in the latter PR. Thanks for clarifying. Lets proceed!

PR #1985, opened 2022-05-09
Apply a transformation function on listen counts before calculating confidence in CF

amCap1712:
Use a custom function to map listencounts before calculating confidence values. Playcounts range from 0 to infinity. It might be useful to scale them to fix a range instead. So experimenting it.


mayhem (approved):
Great!

PR #1986, opened 2022-05-10
Hide Events Frontend

hrik2001:
# Problem
To hide events, API endpoints were implemented at #1942. But no frontend exists for the implemented PR, hence this PR tries to implement one. (The backend PR hasn't been merged yet, hence, this PR is a descendant of the backend PR. If further changes are made to that PR, changes would be merged here)

# Solution
Frontend with an UI like this has been coded in this PR
image


listenbrainz/db/user_timeline_event.py
```diff
@@ -158,6 +159,27 @@ def get_recording_recommendation_events_for_feed(user_ids: List[int], min_ts: in
         return [UserTimelineEvent(**row) for row in result.fetchall()]
 
 
+def get_user_timeline_events_by_id(id: int) -> UserTimelineEvent:
```
MonkeyDo, line 162: If this method returns only one item the name probably shouldn't be plural:
```suggestion
def get_user_timeline_event_by_id(id: int) -> UserTimelineEvent:
```

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -405,8 +407,122 @@ export default class UserFeedPage extends React.Component<
...
+    try {
+      const status = await APIService.hideFeedEvent(
+        event.event_type,
+        currentUser.name,
+        currentUser.auth_token as string,
+        event.id!
+      );
+
+      if (status === 200) {
+        newAlert("success", "", <>Successfully hidden</>);
```
MonkeyDo, line 424: Not sure we want to show an alert toast, because the visual feedback  of the event being hidden should be enough

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -405,8 +407,122 @@ export default class UserFeedPage extends React.Component<
...
+          ) {
+            // eslint-disable-next-line no-param-reassign
+            traversedEvent.hidden = true;
+          }
+          return traversedEvent;
+        });
+        this.setState({ events: new_events });
+      }
+    } catch (error) {
+      newAlert("danger", "", <>Could not hide event</>);
```
MonkeyDo, line 438: This could potentially be useful:
```suggestion
      newAlert("danger", error.toString(), <>Could not hide event</>);
```

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -405,8 +407,122 @@ export default class UserFeedPage extends React.Component<
...
+    try {
+      const status = await APIService.unhideFeedEvent(
+        event.event_type,
+        currentUser.name,
+        currentUser.auth_token as string,
+        event.id!
+      );
+
+      if (status === 200) {
+        newAlert("success", "", <>Successfully unhidden</>);
```
MonkeyDo, line 456: Same as above, probably no need for a toast notification

listenbrainz/webserver/static/js/tests/utils/APIService.test.ts
```diff
@@ -1184,3 +1184,141 @@ describe("deleteFeedEvent", () => {
...
+
+    apiService.checkStatus = jest.fn();
+  });
+
+  it("calls fetch with correct parameters", async () => {
+    await apiService.hideFeedEvent(
+      "recording_recommendation",
+      "riksucks",
+      "testToken",
+      1337
```
MonkeyDo, line 1206: 🥇

MonkeyDo (commented):
The React bits look good to me.
I don't trust myself enough to comment much on the python code though, I'll tag @amCap1712

listenbrainz/db/user_timeline_event.py
```diff
@@ -212,17 +208,18 @@ def hide_user_timeline_event(user_id: int, event_type: UserTimelineEventType, ev
         raise DatabaseException(str(e))
 
 
-def get_hidden_timeline_events(user: int, count: int) -> List[HiddenUserTimelineEvent]:
+def get_hidden_timeline_event(user_id: int, count: int) -> List[HiddenUserTimelineEvent]:
```
MonkeyDo, line 211: Inversely to my original comment, the docline says "Retrieves all events…" so here the  plural name `get_hidden_timeline_events` was appropriate

MonkeyDo (commented):
👌 for me for the front-end part.

listenbrainz/db/user_timeline_event.py
amCap1712: Pluralizing makes sense, thanks for catching it! :D

MonkeyDo (approved):
Thanks a lot!

We deployed it on test.listenbrainz.org and everything is working fine.
Nice work !

PR #1987, opened 2022-05-11
Removing extra Z from missing MB data API and view

hrik2001:
# Problem
Earlier we had manually add `Z` at the end of timestamps to denote UTC timezone. Now spark adds the Z and hence this causes double Z
image
image

# Solution
Removed the Z from API, view and tests


amCap1712 (approved):
LGTM, Thanks!

PR #1988, opened 2022-05-11
Metadata viewer: Show better message if no listen to show

MonkeyDo:
Currently, the metadata viewer (https://listenbrainz.org/playing-now) only show a bit of text if there is no playing-now to show.
It's ugly and not user friendly at all.

Instead, let's show a more helpful message with a link to how to send data to LB as well as some basic expectation management text (and a nice frosted glass transparency effect).
Feedback on the wording is very welcome ! 
image

### DIFF
--- a/listenbrainz/webserver/static/css/metadata-viewer.less
+++ b/listenbrainz/webserver/static/css/metadata-viewer.less
@@ -26,6 +26,7 @@
 		z-index: 2;
 	}
 	.no-listen {
+		max-width: 50em;
 		min-width: 340px;
 		height: 300px;
 		// frosted glass effect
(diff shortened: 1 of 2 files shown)


mayhem:
Maybe this?

"Sorry, we do not know what music you are currently listening to, since we have not received any recent Playing Now events for your account. In order to receive these events, you'll need to [send listens] to ListenBrainz and the service that sends these events also needs to send the Playing Now event to us. (add link to API docs, perhaps?)

We work hard to make this data available to you as soon as we receive it, but until your music service sends us this event, we cannot display anything here."

mayhem (approved):
In general looks good, but I think we probably want to massage the message a bit.

amCap1712 (approved):
LGTM, Thanks!

mayhem:
Yes, looks good. I'm not a huge fan of the text centering, but IANAGD. :D

PR #1989, opened 2022-05-13
Custom timestamp on Recommendations' ListenCard

MonkeyDo:
Based on #1983 (and the **PR is set to merge into the recording-discovery-spark branch**)

This PR refactors the ListenCard to allow for some custom text prefix before the timestamp:
image

The reason we want this is we are using the location and style of usual timestamp which usually mean "you listened to this song on:".
The introduction of the concept of "you **first listened to** this song on:" in recommendations pages requires some clarification.


amCap1712 (approved):
LGTM, Thanks!

PR #1990, opened 2022-05-16
Add musicbrainz_up consul flag

amCap1712:
When MB is unavailable, we want to LB to remain up but disable the features that require MB to be available. Currently, only login and registration are affected.

### DIFF
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -46,7 +46,7 @@ POSTGRES_ADMIN_URI = "SERVICEDOESNOTEXIST_pgbouncer-master"
 # webserver from accessing the musicbrainz database.
 {{- /* This needs to use key/printf again here instead of our pre-defined template at the top of the file as you cannot
      call a template in an if statement or variable declaration */}}
-{{- $musicbrainz_up := (key (printf "docker-server-configs/LB/config.%s.json/$musicbrainz_up" (env "DEPLOY_ENV"))) }}
+{{- $musicbrainz_up := (key (printf "docker-server-configs/LB/config.%s.json/musicbrainz_up" (env "DEPLOY_ENV"))) }}
 {{- if $musicbrainz_up | parseBool}}
     {{if service "pgbouncer-slave"}}
     {{with index (service "pgbouncer-slave") 0}}
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -59,6 +59,7 @@ MB_DATABASE_URI = 'postgresql://musicbrainz_ro@{{.Address}}:{{.Port}}/musicbrain
     {{else}}
 MB_DATABASE_URI = "SERVICEDOESNOTEXIST_pgbouncer-slave_pgbouncer-master"
     {{end}}
+{{else}}
 MB_DATABASE_URI = ""
 {{end}}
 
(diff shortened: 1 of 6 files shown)


listenbrainz/webserver/templates/index/musicbrainz-offline.html
```diff
@@ -0,0 +1,21 @@
+{%- extends 'base.html' -%}
+{%- block title -%}MusicBrainz currently unavailable - ListenBrainz{%- endblock -%}
+{%- block content -%}
+  <h2 class="page-title">MusicBrainz currently unavailable</h2>
+
+  <p>
+     The MusicBrainz database is currently offline for maintenance. As a result,
```
alastair, line 7: I'd consider making this explanation a bit more generic, it might be confusing to people to read something about MusicBrainz while they're on LB. How about something like "ListenBrainz login is currently unavailable due to database maintenance"

listenbrainz/webserver/decorators.py
```diff
@@ -56,3 +57,18 @@ def decorator(*args, **kwargs):
...
+
+def web_musicbrainz_needed(func):
+    """
+        This web decorator checks to see if musicbrainz db is online (by having
+        a DB URI) and if not, it redirects to an error page telling the user
+        that the musicbrainz db is offline.
+    """
+    @wraps(func)
+    def decorator(*args, **kwargs):
+        if musicbrainz_db.engine is None:
```
alastair, line 70: This is going to cause problems in local development if people don't have a musicbrainz database - perhaps we need a separate config item in this case?

consul_config.py.ctmpl
```diff
@@ -41,16 +41,26 @@ POSTGRES_ADMIN_URI = "SERVICEDOESNOTEXIST_pgbouncer-master"
...
 
 
-{{if service "pgbouncer-slave"}}
-{{with index (service "pgbouncer-slave") 0}}
+# Use a key '$musicbrainz_up' in consul (json boolean true or false) to decide if we connect to musicbrainz.
+# If we set MB_DATABASE_URI to an empty string, then the @mb_database_needed decorator will prevent the
+# webserver from accessing the musicbrainz database.
+{{- /* This needs to use key/printf again here instead of our pre-defined template at the top of the file as you cannot
+     call a template in an if statement or variable declaration */}}
+{{- $musicbrainz_up := (key (printf "docker-server-configs/LB/config.%s.json/musicbrainz_up" (env "DEPLOY_ENV"))) }}
```
alastair, line 49: not important for this PR, but perhaps we can have another look into how to do this key interpolation, so that we don't have to copy/paste printf statements all over this file...

alastair (commented):
I had a look through all of the other places that use `brainzutils.musicbrainz_db`, and it seems that all of them already have a check in place to skip if the engine is None (e.g. player metadata, huesound metadata, email address check on lastfm api)

listenbrainz/webserver/templates/index/musicbrainz-offline.html
amCap1712: Fixed.

listenbrainz/webserver/decorators.py
amCap1712: Fixed by adding a separate config variable.

PR #1991, opened 2022-05-16
Return empty results if MusicBrainz database connection isn't available

alastair:
# Problem

related: https://github.com/metabrainz/listenbrainz-server/pull/1990
When the MusicBrainz database connection isn't available we don't want an exception to be raised when accessing labs-api endpoints that require the MB database

# Solution
Identify if the db is unavailable, and if so return an empty result instead of trying to connect to it.

# Action

In the future we could update dataset-hoster to return http 503 service unavailable in this case instead of a valid response that could be interpreted as "no data available for this query".

### DIFF
--- a/listenbrainz/labs_api/labs/api/artist_credit_from_artist_mbid.py
+++ b/listenbrainz/labs_api/labs/api/artist_credit_from_artist_mbid.py
@@ -26,7 +26,7 @@ class ArtistCreditIdFromArtistMBIDQuery(Query):
         if not current_app.config["MB_DATABASE_URI"]:
             return []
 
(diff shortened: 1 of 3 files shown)


listenbrainz/labs_api/labs/api/artist_credit_from_artist_mbid.py
```diff
@@ -23,6 +23,9 @@ def outputs(self):
     def fetch(self, params, count=-1, offset=-1):
 
         with psycopg2.connect(config.MB_DATABASE_URI) as conn:
+            if not config.MB_DATABASE_URI:
```
amCap1712, line 26: I had earlier looked from mobile so missed this. This needs to be outside the with,

PR #1995, opened 2022-05-18
Add information about cron stop script to maintainers docs

alastair:
# Problem

I keep forgetting about the tools that we have available to help us deploy cron and other containers to prod

# Solution

Add some docs about the cron check script. This document could be further updated with instructions about how to deploy other containers (and especially an explicit indication if they can be restarted at any time).


amCap1712 (approved):
LGTM, Thanks!

PR #1996, opened 2022-05-18
Add initial mapping documentation for developers and maintainers

alastair:
# Problem

This is an initial attempt at documenting the MBID mapper for the following purposes:
* Information for potential developers, showing them how to build the metadata tables and configure the mapper
* Information for people who want to use the metadata dumps and 
* Information for maintainers who want to debug why the mapping isn't working for some submissions.

# Solution

Write some docs! I split it into two parts, one general description for developers, and another page with debugging information for maintainers.

# Action

I expect that much of this document may be incorrect, the idea was to get some text in place so that someone can tell me that I'm wrong and point me in the correct direction to fix it.
I added a few TODOs that I'd like to fill in before merging this. Any other recommendations for todos would be good too.
I added a few names of management commands or python commands to run the tools, but I'm not sure if we should directly link to lines of code too - these links would go out of date quickly if they link to a specific line number in master branch, so maybe we could just list the important functions instead?

There is still no docker-compose setup for local development to run all of these services (however there are some run commands in the mbid_mapping README). We can decide if we want to make a new docker-compose file for this process in this PR, or if we add it at a later stage.

### DIFF
--- a/docs/maintainers/mapping.rst
+++ b/docs/maintainers/mapping.rst
@@ -1,0 +1,51 @@
+MBID Mapping
+============
+
+For a background on how the mapping works, see :ref:`developers-mapping`
+
+
+Containers
+^^^^^^^^^^
+
+The mapping tools run in two containers:
+
+ * ``mbid-mapping-writer-prod``: Populates the ``mbid_mapping`` table for new listens. Built from the 
+    main ListenBrainz dockerfile.
+
+ * ``mbid-mapping``: Periodically generates the MBID Mapping supplemental tables, typesense index,
+    and huesound index. Built from ``listenbrainz/mbid_mapping/Dockerfile``
+
+
+Data sources
+^^^^^^^^^^^^
+
+In the production environment, the ``mbid-mapping`` container reads from the MB replica on aretha.
+
+
+Debugging lookups
+^^^^^^^^^^^^^^^^^
+
+If a listen isn't showing up as mapped on ListenBrainz, one of the following might be true:
+
+* The item wasn't in musicbrainz at the time that the lookup was made
+* There is a bug in the mapping algorithm
+
+If the recording doesn't exist in MusicBrainz during mapping, a row will be added to the ``mbid_mapping`` table
+with the MSID and a ``match_type`` of ``no_match``. Currently no_match values aren't looked up again automatically.
+
+You can test the results of a lookup by using `https://labs.api.listenbrainz.org/explain-mbid-mapping <https://labs.api.listenbrainz.org/explain-mbid-mapping>`
+This uses the same lookup process that the mapper uses. If this returns a result, but there is no mapping present
+it could be due to data being recently added to MusicBrainz or improvements to the mapping algorithm.
+
+If no data is returned or an incorrect match is being returned, this should be reported to us, by adding a comment
+to `LB-1036 <https://tickets.metabrainz.org/browse/LB-1036>`.
+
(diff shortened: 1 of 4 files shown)


docs/maintainers/mapping.rst
```diff
@@ -0,0 +1,27 @@
+MBID Mapping
+============
+
+For a background on how the mapping works, see :ref:`developers-mapping`
```
amCap1712, line 4: Probably add that the source of the data in the tables is the json dump database on aretha and that a daily cronjob in listenbrainz-mbid-mapping container (not the listenbrainz-mbid-mapping-writer-prod) recreates it.

docs/maintainers/mapping.rst
```diff
@@ -0,0 +1,27 @@
...
+You can test the results of a lookup by using https://labs.api.listenbrainz.org/explain-mbid-mapping
+This uses the same lookup process that the mapper uses. If this returns a result, but there is no mapping present
+it could be due to data being recently added to MusicBrainz or improvements to the mapping algorithm.
+
+In this case you can retrigger a lookup by seting the ``mbid_mapping.last_updated`` field to '1970-01-01'. 
+The mapper will pick up these items and put them on the queue again.
+
+.. code:: sql
+
+    UPDATE mbid_mapping SET last_updated = 'yyyy-mm-dd' WHERE recording_msid = '00000737-3a59-4499-b30a-31fe2464555d';
```
amCap1712, line 26: can use last_updated = 'epoch' also.

docs/maintainers/mapping.rst
```diff
@@ -0,0 +1,27 @@
...
+
+If a listen isn't showing up as mapped on ListenBrainz, one of the following might be true:
+
+* The item wasn't in musicbrainz at the time that the lookup was made
+* There is a bug in the mapping algorithm
+
+If the recording doesn't exist in MusicBrainz during mapping, a row will be added to the ``mbid_mapping`` table
+with the MSID and a ``match_type`` of ``no_match``. Currently no_match values aren't looked up again automatically.
+
+You can test the results of a lookup by using https://labs.api.listenbrainz.org/explain-mbid-mapping
```
amCap1712, line 17: Probably turn into a hyperlink.

docs/maintainers/mapping.rst
```diff
@@ -0,0 +1,27 @@
...
+This uses the same lookup process that the mapper uses. If this returns a result, but there is no mapping present
+it could be due to data being recently added to MusicBrainz or improvements to the mapping algorithm.
```
amCap1712, line 19: If no data is returned or an incorrect match is being returned, probably report the case like add to LB-1036.

docs/maintainers/mapping.rst
```diff
@@ -0,0 +1,27 @@
...
+Debugging lookups
+^^^^^^^^^^^^^^^^^
+
+If a listen isn't showing up as mapped on ListenBrainz, one of the following might be true:
+
+* The item wasn't in musicbrainz at the time that the lookup was made
+* There is a bug in the mapping algorithm
+
+If the recording doesn't exist in MusicBrainz during mapping, a row will be added to the ``mbid_mapping`` table
+with the MSID and a ``match_type`` of ``no_match``. Currently no_match values aren't looked up again automatically.
```
mayhem, line 15: This will change very soon, if Lucifer has something to say about it. :)

docs/maintainers/mapping.rst
mayhem: It should be noted that these are only processed once a day currently, so its not a very speedy process. But this could be changed to happen hourly, if we wish.

alastair:
comments applied and merged into #2157, closing this one

PR #1997, opened 2022-05-18
Add `listenbrainz-mpd` submission client

elomatreb:
This pull request adds a link to the [`listenbrainz-mpd`](https://codeberg.org/elomatreb/listenbrainz-mpd) submission client. It also updates the name and URL of the existing client for MPD (renamed from "libra" to "wylt").

I took the liberty of listing my client first because it is capable of submitting MusicBrainz Identifiers (among other things), which "wylt" does not support.


amCap1712 (approved):
LGTM, Thanks!

PR #1998, opened 2022-05-19
Recheck unmatched incoming listens periodically

amCap1712:
If the msid of an incoming listen is unmatched and was last checked at least 2 weeks ago, recheck it.


mayhem (approved):
LGTM, with future improvements as planned in IRC

PR #2001, opened 2022-05-20
Some misc improvements to recommendation processing and display

amCap1712:
1) Remove redundant usages of getattr. The attr name is fixed so we can always directly access the property.
2) Refactor recommendation post processing to more in SQL query.
3) Use database directly for fetching playable data for recommendations.


mayhem (approved):
I'm amazed at how you can continually make this code simpler. Looking better every step!

PR #2002, opened 2022-05-20
Add release_tags

mayhem:
When I first created the mb metadata cache I skipped the release_tags. Finally adding them now.

This PR uses the MBID mapping supplemental tables to find the canonical release and then pluck the release_tags from it to be included in the index. Nothing really new here.


amCap1712 (approved):
LGTM, Thanks!

PR #2003, opened 2022-05-22
Fix check for discovery timestamp

amCap1712:
The check intends to look for listened_at_iso field if nothing was found in the latest_listened_at field but currently it doesn't do that. Discovered when experimenting on the recommendations page on LB.


MonkeyDo (approved):
That went past both of use 🤦

amCap1712:
Indeed!

PR #2004, opened 2022-05-23
Add raw recommendations

amCap1712:
Add a new category of recommendations - raw. These recommendations are directly calculated from the model based on the training data without passing any candidate data set to the model.

### DIFF
--- a/listenbrainz_spark/recommendations/recording/recommend.py
+++ b/listenbrainz_spark/recommendations/recording/recommend.py
@@ -232,6 +232,7 @@ def create_messages(model_id, model_html_file, top_artist_recs_df, similar_artis
             'recommendations': {
                 'top_artist': data['top_artist'],
                 'similar_artist': data['similar_artist'],
+                'raw': data['raw'],
                 'model_id': model_id,
                 'model_url': f"http://michael.metabrainz.org/{model_html_file}"
             }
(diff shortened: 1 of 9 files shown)


listenbrainz_spark/recommendations/recording/recommend.py
```diff
@@ -87,9 +87,21 @@ def process_recommendations(recommendation_df, limit):
               FROM recommendation
         )
         SELECT user_id
-             , recording_mbid
-             , score
-             , date_format(latest_listened_at, "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'") AS latest_listened_at
+             , array_sort(
```
amCap1712, line 90: This is change is not to do with raw_recs but it felt to me that collecting in a list in python was slower and doing it in spark seemed somewhat faster.

listenbrainz_spark/recommendations/recording/recommend.py
```diff
@@ -170,47 +182,48 @@ def get_user_name_and_user_id(top_artist_candidate_set_df, users):
...
-        user_rec[row.user_id]['similar_artist'].append(
-            {
-                "recording_mbid": row.recording_mbid,
-                "score": row.score,
-                "latest_listened_at": row.latest_listened_at
-            }
-        )
+        row_dict = row.asDict(recursive=True)
+        user_rec[row_dict["user_id"]]["similar_artist"] = row_dict["recs"]
+        similar_artist_rec_user_count = similar_artist_rec_user_count + 1
```
mayhem, line 219: Any reason not to use += here?

mayhem (approved):
I'm looking forward to seeing these in action!

listenbrainz_spark/recommendations/recording/recommend.py
amCap1712: No reason, will fix.

PR #2005, opened 2022-05-23
Fix homepage

akshaaatt:
- Homepage cleanup and footer separation


listenbrainz/webserver/templates/footer.html
```diff
@@ -0,0 +1,323 @@
...
+<!--                    <img src="{{ url_for('static', filename='img/listenbrainz-logo.svg') }}" width="180" alt="ListenBrainz"/>-->
+<!--                </h3>-->
+<!--                <br/>-->
+<!--                <p class="color-gray">-->
+<!--                    ListenBrainz keeps tracks of what music you listen to and provides you with insights into your listening habits.-->
+<!--                    <br>-->
+<!--                    You can use ListenBrainz to track your listening habits, discover new music with personalized recommendations, and share your musical taste with others using our visualizations.-->
+<!--                </p>-->
+<!--                <ul class="list-unstyled">-->
+<!--                    <li class="color-a">-->
```
MonkeyDo, line 242: I suppose all this commented code should be removed?

listenbrainz/webserver/templates/navbar.html
```diff
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-expand-lg navbar-light bg-white" role="navigation">
+<nav class="navbar fixed-top navbar-expand-lg navbar-light bg-white border-bottom border-" role="navigation">
```
MonkeyDo, line 1: The last `border-` class doesn't look right

listenbrainz/webserver/templates/index/index.html
```diff
@@ -43,61 +43,57 @@ <h2>Discover new music</h2>
...
+        </div>
+        <div class="panel-footer" style="text-align: center">
+          <small>Updated every 5 minutes</small><br />
+          <small><a href="{{ url_for('index.current_status') }}">more on the current status page</a></small>
         </div>
       </div>
-
     </div>
-
     <div class="col-md-5"
```
MonkeyDo, line 97: I suppose this will be coming in another PR, but I just wanted to mention since we're looking at the homepage that the stats and news panels look like they're missing css styles:
image

listenbrainz/webserver/templates/footer.html
```diff
@@ -0,0 +1,323 @@
...
+                        </a>
+                    </li>
+                </ul>
+            </div>
+            <div class="col-sm-12 col-md-3">
+                <h3 class="ms-3">
+                    Fellow Projects
+                </h3>
+                <ul class="list-style-type">
+                    <li class="pt-2 pb-2">
```
MonkeyDo, line 45: I think there's an error here, there's no MusicBrainz but ListenBrainz is here which we don't need.
I also think that means that maybe (?) you can remove the `meb-icons/ListenBrainz.svg` file you added in this PR (if it's only used in the footer)

listenbrainz/webserver/templates/footer.html
```diff
@@ -0,0 +1,323 @@
...
+                            community.metabrainz.org
+                        </a>
+                    </li>
+                </ul>
+            </div>
+            <div class="col-sm-12 col-md-3">
+                <h3 class="ms-3">
+                    Fellow Projects
+                </h3>
+                <ul class="list-style-type">
```
MonkeyDo, line 44: Not sure what styles would need to be transferred from before, but I quite liked the chevrons you used in the footer:
image

listenbrainz/webserver/templates/footer.html
akshaaatt: Right!

listenbrainz/webserver/templates/footer.html
akshaaatt: Bitmap left a reply suggesting to me that we shouldn't add extra styles and hence I should remove the chevrons and keep things standard. I do think that makes sense, hence made the move. Let me know if that if fine.

listenbrainz/webserver/templates/footer.html
akshaaatt: Thanks! Fixed.

listenbrainz/webserver/templates/index/index.html
akshaaatt: Fixed :)

listenbrainz/webserver/templates/footer.html
MonkeyDo: Or perhaps removing the list style entirely?

image

MonkeyDo:
For this PR as well, I'd be happy to see some footer-specific PR.

Since we have redesigned LB front-end completely a lot of things in this PR aren't relevant anymore.

PR #2006, opened 2022-05-24
Fetch release_group tags, not release tags

mayhem:
The last mb metadata cache PR added support for release_tags, which are useless. This PR changes this support to be release_group tags, which are useful.


amCap1712 (approved):
LGTM, Thanks!

PR #2009, opened 2022-05-24
Allow a maximim number of listens per import

alastair:
# Problem

Processing too many items in a single import (rabbitmq message) can cause delays and issues with the timescale writer.

# Solution

Limit the max number of messages to 1k per import.

# Action

The documentation is incorrect - it says

> The complete JSON document may not exceed :data:`~webserver.views.api.MAX_LISTEN_SIZE` bytes in size

but we actually check that num listens * max_listen_size is not higher than the submitted payload. We should decide which is true, and update the docs or the check.

Also need to check what is the max number of items that the lastfm import uses at once and check it's under this limit.

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -21,9 +21,10 @@ from listenbrainz.webserver.errors import APIBadRequest, APIInternalServerError,
     APIUnauthorized, ListenValidationError
 from listenbrainz.webserver.models import SubmitListenUserMetadata
 from listenbrainz.webserver.utils import REJECT_LISTENS_WITHOUT_EMAIL_ERROR
-from listenbrainz.webserver.views.api_tools import insert_payload, log_raise_400, validate_listen, parse_param_list, \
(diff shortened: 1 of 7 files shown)


amCap1712 (approved):
LGTM, Thanks!

alastair:
Regarding my question about the listen size check - 
the [sample listen](https://listenbrainz.readthedocs.io/en/latest/users/json/) is about 700 bytes (round up to 1k), so if we said that the whole message can't exceed 10k then that's only about 10 messages, too small. I'll update the docs.

We also decided to add another check before parsing json, to check that the body is less than 10mb in size (10k per listen * 1000 max listens), so that we don't even bother loading json if the body is too large

listenbrainz/webserver/views/api.py
```diff
@@ -58,6 +58,12 @@ def submit_listen():
         raise APIUnauthorized(REJECT_LISTENS_WITHOUT_EMAIL_ERROR)
 
     raw_data = request.get_data()
+
+    if len(raw_data) > MAX_LISTENS_PER_REQUEST * MAX_LISTEN_SIZE:
```
amCap1712, line 62: Can use MAX_PAYLOAD_SIZE here.

PR #2014, opened 2022-05-26
Fix get_timestamps_for_user to account for more cases

amCap1712:
We need to do coalesce in last_update CTE  and not at the end because  max_listened_at and min_listened_at is set to NULL:

1) for new users
2) users who have deleted listen history and re-imported
 (and cron job hasn't run yet)
3) development (cron job never runs)

The where clause in listens_after_update immediately evaluates to false if listened_at is compared to NULL and thus preventing the query from finding new listens submitted since the cron job ran last.


mayhem (approved):
:ship:

PR #2017, opened 2022-05-26
Fix extraneous 0 in listen count card

amCap1712:
Don't use {listen_count && components} because of the way React and JS treat falsy values like 0. JS will evaluate 0 && as false so the part after && is not rendered. However, React will still render the listen count as 0. To use { && form, you'll need to do !!listen_count, but it is easier to just use an explicit if.

Before:
Screenshot from 2022-05-26 21-51-10

After:
Screenshot from 2022-05-26 22-59-49


MonkeyDo (approved):
Yus. JS is weird.

PR #2018, opened 2022-05-28
Update and Fix Home section

akshaaatt:
- Fix the tabs present in the Home section
- Improve the common components


listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -624,43 +624,7 @@ export default class Listens extends React.Component<
...
-              <div
-                className="card hidden-xs hidden-sm"
-                style={{ paddingTop: "1.5em" }}
-              >
-                <UserSocialNetwork user={user} newAlert={newAlert} />
-              </div>
-            )}
-          </div>
-          <div className="col-md-8 col-md-pull-4">
+          <div className="col-md-8">
```
MonkeyDo, line 627: The reason why I arranged the two parts lie I did with `col-md-push-8` and `col-md-pull-4` is so that the playing now, pinned listen and listen count cards end up *above* the list of listens on mobile screens, rather than below.

MonkeyDo:
I'm going to close this, as there have been a lot of changes in the front-end and I can't guarantee this can be merged cleanly.

Happy to see another PR for the footer work for example.

PR #2021, opened 2022-05-30
Add check_again column for MBID mapping writer

amCap1712:
We need a way to recheck msids later for which a match was not found. Initially if a match is not found, check again a day later. After that retry, every 2 * (NOW() - last_updated) INTERVAL later till no match is found.


mayhem (commented):
The changes to the existing tables look fine to me.

mayhem (approved):
LGTM! Thanks!

PR #2023, opened 2022-06-01
Refactor `boostrap` -> `bootstrap`

chinmaykunkikar:
**Boo! 👻**

Other "boostrap" jokes from IRC -
>  - "Boozestrap would have been better"
>  - "I mean, why fix it now? We apparently liked it!"
>  - "I suspect it would be this monkey's fault."..."Yeah, actually not this monkey!"
>  - "Wouldn't want to throw you under the bus(trap)"


akshaaatt (approved):
Thank you! :)

PR #2024, opened 2022-06-01
Fix ISEs due to duplicate feedback

amCap1712:
Delete the existing feedback and then insert new feedback. we cannot use ON CONFLICT DO UPDATE because it is possible for a user to submit the feedback using recording_msid only and then using both recording_msid and recording_mbid at once. Then the ON CONFLICT clause won't work well. We can use partial unique indexes to make it work but there will still be duplicates. therefore, we delete first then insert new feedback in the same transaction.

Originally, I tried using partial unique indexes but that still leaves room for duplicates.


listenbrainz/db/feedback.py
```diff
@@ -67,20 +80,22 @@ def delete(feedback: Feedback):
...
+    params = {"user_id": feedback.user_id}
 
-    if feedback.recording_msid:
-        conditions.append("recording_msid = :recording_msid")
-        args["recording_msid"] = feedback.recording_msid
-
-    if feedback.recording_mbid:
-        conditions.append("recording_mbid = :recording_mbid")
-        args["recording_mbid"] = feedback.recording_mbid
+    if feedback.recording_msid is not None and feedback.recording_mbid is not None:
```
alastair, line 85: I see we're explicitly checking against None here, is there a case when they might be `""`? (same question for the insert method)

In fact, I see that the column type for these fields is UUID, is there a case where we might insert invalid data and get an exception? I see that the `Feedback` data model has a validator for these fields which we use in the API endpoint, so I think we're fine here right?

alastair (commented):
looks good, two small questions otherwise this looks good to go

listenbrainz/db/feedback.py
amCap1712: As discussed will fix in separate PR and also add validator for at least one of mbid and msid to be present.

listenbrainz/db/feedback.py
```diff
@@ -49,16 +53,25 @@ def insert(feedback: Feedback):
...
+        insert_query = INSERT_QUERIES["msid"]
 
-    with db.engine.connect() as connection:
-        connection.execute(text(query), params)
+    with db.engine.connect() as connection, connection.begin():
+        # delete the existing feedback and then insert new feedback. we cannot use ON CONFLICT DO UPDATE
+        # because it is possible for a user to submit the feedback using recording_msid only and then
+        # using both recording_msid and recording_mbid at once. Then the ON CONFLICT clause won't work
+        # well. We can use partial unique indexes to make it work but there will still be duplicates.
+        # therefore, we delete first then insert new feedback in the same transaction.
```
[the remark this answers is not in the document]
amCap1712: Done

PR #2027, opened 2022-06-02
Fix env variable issues with cron_lock script

amCap1712:
cronjob doesn't observe environment variables set by docker. Therefore, the deploy_env check always fails in production and the cron lock is never created. This forfeits the purpose of the script. Fixing the issue by reading the admin/config.sh file instead. If that file doesn't exist, exit with error.

### DIFF
--- a/admin/cron_lock.py
+++ b/admin/cron_lock.py
@@ -7,23 +7,25 @@ import click
 
 
 CRON_LOGS_DIR = "/logs"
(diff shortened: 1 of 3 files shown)


amCap1712:
I tested with the test cron job script added in the PR, will remove it before merging.

mayhem (approved):
This is an interesting way of doing this!

PR #2029, opened 2022-06-03
Fix daily jams

amCap1712:
Troi had a serialization bug, so update troi to fix that. Also, fix passing usernames to daily jams, currently we pass a dict instead it
should be the username's value.


mayhem (approved):
thanks again for bailing me out on this mess. sorry!

PR #2031, opened 2022-06-03
Improve validation in MsidMbidModel

amCap1712:
1. Error if the mbid or msid specified is empty string.
2. Error if both msid and mbid are None. (at least 1 is required)


alastairporter (approved):
looks good, though a test would be nice!

PR #2032, opened 2022-06-04
Setup stylelint for styles

chinmaykunkikar:
`package.json`
 - Introduce npm script `format:less`
 - **stylelint** - Main stylelint package.
 - **postcss-less** - Custom syntax to parse less styles (grammar for .less).
 - **stylelint-config-recommended-less** - Recommended linting rules for .less.
 - **stylelint-config-prettier** - Configure prettier to pretty format `.less` files.

`.stylelintrc`
 - Extend stylelint with "stylelint-config-recommended-less" for `.less` files.
 - Use customSyntax, the grammar for `.less` files.
 - Use prettier plugin and configure it as a rule for stylelint.
 - Ignore all bootstrap files.

`"no-descending-specificity": null`
  Lot of styles don't honor descending specificity. [[1]](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) [[2]](https://stylelint.io/user-guide/rules/list/no-descending-specificity)
  We should try and eliminate all such linting errors to eliminate unexpected behavior of some styles.
  Following specificity rules will also help eliminate few override styles that are using `!important`. [[3]](https://css-tricks.com/the-surprising-behavior-of-important-in-css-custom-property-values/)
  I have temporarily ignored the linting rule to keep the lint output short.


MonkeyDo (commented):
Looks pretty good to me !

I think we should also run this automatically with Webpack, I guess using this plugin: https://webpack.js.org/plugins/stylelint-webpack-plugin/

Then we can set it up to auto-fix the files when not in production mode like we [currently do for eslint](https://github.com/metabrainz/listenbrainz-server/blob/master/webpack.config.js#L43).


There are also some issues with the automated build CI test: https://github.com/metabrainz/listenbrainz-server/runs/6739670781?check_suite_focus=true#step:5:264

First thing to check would be if the Docker image builds properly for you locally with these changes (perhaps without the suggested changes above for starters).
If not, you can try to resolve the issue locally; if the image builds then we would need to figure out what is different between local and CI environment/commands being run.

Let me know if you get stuck !

chinmaykunkikar:
@MonkeyDo an update:
1. Got rid of the CI error from earlier.
2. Added Stylelint webpack plugin.
3. Make stylelint work with docker.
4. Run stylelint check during `./test.sh fe -f`
5. Eliminate more lint errors.

Almost all the tests are passing. I have no idea why unit tests aren't passing :(

MonkeyDo:
Thanks @chinmaykunkikar I'll have another review.
The unit test failure is a phantom concurrency bug we sometimes get in the CI tests :poop:.
I re-ran the suite and it should be sorted.

chinmaykunkikar:
@MonkeyDo tests are passed, I'll wait for the review :)

chinmaykunkikar:
Should we explore [lint-staged](https://www.npmjs.com/package/lint-staged#examples) as well?

MonkeyDo:
Yes I think running locally on commits is also useful, but opinions may
vary.
Perhaps it could be done in a separate PR based on this one, so that other
devs can chime in?

Le jeu. 9 juin 2022, 19:25, Chinmay Kunkikar ***@***.***> a
écrit :

chinmaykunkikar:
> Perhaps it could be done in a separate PR based on this one, so that other devs can chime in?

Sounds good to me :)

MonkeyDo (approved):
Great job, thanks !

PR #2034, opened 2022-06-06
 Fix recording discovery timestamp

amCap1712:
Instead of using last, need to use max to get latest listened at timestamp. Also, change discovered on to last listened at because discovery may mean first listened at date whereas we are displaying the latest one here.


MonkeyDo (approved):
:+1:

mayhem (approved):
I also like the terminology change, thanks!

PR #2035, opened 2022-06-06
create new table, modify UI and endpoint for timezone

YouzhiL:
# Problem
Listens on ListenBrainz are collected in the UTC instead of their local timestamp currently, without taking DST in to consideration. Adding timezone preference for users will contribute to the analysis and thus provide better user experience.

# Solution
## database: 
1. Create a new table `user_setting`. The user_timezone information and other user preferences can be stored here.
2. Create a function to get all timezone recognized by PostgreSQL.
3. Create functions to create, get, and modify user settings.
4. Create a function to convert timezones retrieved by PostgreSQL to standardized format for display.
## UI: 
1. Add a `Reset timezone`  UI module in the general setting page.
2. Create a new webpage for user to select timezone.
## endpoint
Create an endpoint for getting and resetting user timezone preference by creating a `TimezoneForm` extended from `FlaskForm`.

### DIFF
--- a/admin/sql/create_indexes.sql
+++ b/admin/sql/create_indexes.sql
@@ -57,6 +57,6 @@ CREATE INDEX user_id_ndx_pinned_recording ON pinned_recording (user_id);
 CREATE INDEX release_mbid_ndx_release_color ON release_color (release_mbid);
 CREATE UNIQUE INDEX caa_id_ndx_release_color ON release_color (caa_id);
 
-CREATE INDEX user_id_ndx_user_setting ON user_setting (user_id);
+CREATE UNIQUE INDEX user_id_ndx_user_setting ON user_setting (user_id);
 
 COMMIT;
--- a/admin/sql/create_indexes.sql
+++ b/admin/sql/create_indexes.sql
@@ -57,4 +57,6 @@ CREATE INDEX user_id_ndx_pinned_recording ON pinned_recording (user_id);
 CREATE INDEX release_mbid_ndx_release_color ON release_color (release_mbid);
 CREATE UNIQUE INDEX caa_id_ndx_release_color ON release_color (caa_id);
 
+CREATE INDEX user_id_ndx_user_setting ON user_setting (user_id);
+
 COMMIT;
--- a/listenbrainz/db/user_setting.py
+++ b/listenbrainz/db/user_setting.py
@@ -1,0 +1,118 @@
+import sqlalchemy
+from datetime import datetime, timedelta
+from listenbrainz import db
+from listenbrainz.db.exceptions import DatabaseException
+
+import logging
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+def get_pg_timezone():
+    """ Get list of time zones PostgreSQL supports.
+
+    Returns:
+        list of tuple('zone_name', 'utc_offset')
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT * FROM pg_timezone_names
+            ORDER BY name
+        """))
+        timezones = [(row["name"], row["utc_offset"]) for row in result.fetchall()]
+        timezones = standardize_timezone(timezones)
+        return timezones
+
+
+def get(user_id: int):
+    """Get user setting with the row ID of the user in the DB, or create if there's setting history.
+    Args:
+        user_id (int): the row ID of the user in the DB
+    Returns:
+        zone_name (str): the user selected timezone name. If not selected by user, return 'UTC'
+    """
+    with db.engine.connect() as connection:
+        try:
+            result = connection.execute(sqlalchemy.text("""
+                SELECT timezone_name
+                FROM user_setting
+                WHERE user_id = :user_id
+            """), {
+                "user_id": user_id,
+            })
+            if result.rowcount:
+                return dict(result.fetchone())
+            return None
+            # print("get from db...user_timezone: ", row)
+        except sqlalchemy.exc.ProgrammingError as err:
+            logger.error(err)
+            raise DatabaseException(
+                "Couldn't get user's timezone: %s" % str(err))
+
+def create(user_id: int, timezone_name: str = "UTC"):
+    """Create setting for user
+
+    Args:
+        user_id (int): the row ID of the user in the DB
+        timezone_name(str): with default as "UTC
+
+    Returns:
+        ID of newly created user setting.
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            INSERT INTO user_setting (user_id, timezone_name)
+                 VALUES (:user_id, :timezone_name)
+              RETURNING id
+        """), {
+            "user_id": user_id,
+            "timezone_name": timezone_name,
+        })
+        return result.fetchone()["id"]
+
+   
+def get_or_create(user_id: int):
+    """Get user setting with the row ID of the user in the DB, or create if there's setting history.
+    Args:
+        user_id (int): the row ID of the user in the DB
+    Returns:
+        dict: user settings
+    """
+    user_settings = get(user_id)
+    if not user_settings:
+        create(user_id)
+        user_settings = get(user_id)
+    return user_settings
+
+def update_timezone(user_id: int, timezone_name:str):
+    """update user's timezone
+    Args:
+        user_id (int): the row ID of the user in the DB
+        zone_name (str): the user selected timezone name
+
+    """
+    with db.engine.connect() as connection:
+        try:
+            connection.execute(sqlalchemy.text("""
+                UPDATE "user_setting"
+                SET timezone_name = :timezone_name
+                WHERE user_id = :user_id
+                """), {
+                "user_id": user_id,
+                "timezone_name": timezone_name,
+            })
+            # print('db: ',get_user_timezone(user_id) )
+        except sqlalchemy.exc.ProgrammingError as err:
+            logger.error(err)
+            raise DatabaseException(
+                "Couldn't update user's timezone: %s" % str(err))
+
+def standardize_timezone(timezones):
+    result = []
+    for (name, offset) in timezones:
+        if offset.days > -1:
+            result.append((name, "+" + str(offset) + " GMT"))
+        else:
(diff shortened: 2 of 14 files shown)


YouzhiL:
Hi mayhem,

Thanks for the reply. That progression is also nice to me. I don't have much experience in javascript based UI. This will be a good chance for me to learn new things.

mayhem:
I've finally gotten the chance to run this code and it works! Well done!

Some minor things to tweak to improve wording and layout:

image

On this view, I think it would be good to show the current timezone in bold and then to start a new paragraph for clarity. Then lets have the text say:

```
By choosing your local time zone, you will have a local timestamps part of your submitted listens. This also informs as when to generate daily playlists and other recommendations for you.
```

Next, this page title and the section that leads to this page should say: "Select timezone" rather than Reset Timezone.

Finally, the text on the profile page should read:

```
Your current timezone is <bold>[timezone]</bold>

Please select an appropriate timezone so that we may associated a timezone with your listens and know when to generate recommendations for you.
```

mayhem:
Next, up see the comments left by pep8speaks, which indicates that you ought to fix up some formatting issues in your code. I normally use the command line tool autopep8 to accomplish this:

```
autopep8 -i --max-line-length 130 <filename>
```

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,125 @@
...
+            SELECT * FROM pg_timezone_names
+            ORDER BY name
+        """))
+        timezones = [(row["name"], row["utc_offset"]) for row in result.fetchall()]
+        timezones = standardize_timezone(timezones)
+        return timezones
+
+
+def get(user_id: int):
+    """Get user setting with the row ID of the user in the DB, or create if there's setting history.
```
mayhem, line 28: ", or create if there's setting history."

Not sure this still applies since this function doesn't create anything. Maybe just remove?

mayhem: This function is called get, but it acts more like get_timezone. I wonder if we should keep the word get, but return a dict with all of the settings that this function fetches. For starters that would only be: { "timezone": "utc" }, but this allows us to add future user settings to this dict as needed.

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,125 @@
...
+                SELECT timezone_name
+                FROM user_setting
+                WHERE user_id = :user_id
+            """), {
+                "user_id": user_id,
+            })
+            if result.rowcount:
+                return dict(result.fetchone())
+            return None
+            # print("get from db...user_timezone: ", row)
```
mayhem, line 46: Remove?

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,125 @@
...
+        timezones = standardize_timezone(timezones)
+        return timezones
+
+
+def get(user_id: int):
+    """Get user setting with the row ID of the user in the DB, or create if there's setting history.
+    Args:
+        user_id (int): the row ID of the user in the DB
+    Returns:
+        zone_name (str): the user selected timezone name. If not selected by user, return 'UTC'
```
mayhem, line 32: This does not reflect what the code does -- if there is no user setting right now it will return None. I think we ought to define a constant in this module called 'DEFAULT_TIMEZONE "UTC"' and then return that constant when no row exists.

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,125 @@
...
+                WHERE user_id = :user_id
+            """), {
+                "user_id": user_id,
+            })
+            if result.rowcount:
+                return dict(result.fetchone())
+            return None
+            # print("get from db...user_timezone: ", row)
+        except sqlalchemy.exc.ProgrammingError as err:
+            logger.error(err)
```
mayhem, line 48: Also remove?

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,125 @@
...
+            raise DatabaseException(
+                "Couldn't get user's timezone: %s" % str(err))
+
+def create(user_id: int, timezone_name: str = "UTC"):
+    """Create setting for user
+
+    Args:
+        user_id (int): the row ID of the user in the DB
+        timezone_name(str): with default as "UTC
```
mayhem, line 57: nitpick: missing "

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,125 @@
...
+                 VALUES (:user_id, :timezone_name)
+              RETURNING id
+        """), {
+            "user_id": user_id,
+            "timezone_name": timezone_name,
+        })
+        return result.fetchone()["id"]
+
+   
+def get_or_create(user_id: int):
```
mayhem, line 74: I feel that this function is not needed. If we go with the default return value from get being a dict with a default timezone of UTC, then the user will always get this until they actually set a different timezone.

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,125 @@
...
+    Returns:
+        dict: user settings
+    """
+    user_settings = get(user_id)
+    if not user_settings:
+        create(user_id)
+        user_settings = get(user_id)
+    return user_settings
+
+def update_timezone(user_id: int, timezone_name:str):
```
mayhem, line 87: And if we change this function to set_timezone() and change the SQL query to use an upsert type query, we can make all these functions flow a little smoother. Change the query to use an INSERT ... ON CONFLICT DO UPDATE ... construct and it will either insert the row if it doesn't exist, or update the value if it does exist.

See https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-upsert/

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,125 @@
...
+                "user_id": user_id,
+                "timezone_name": timezone_name,
+            })
+            # print('db: ',get_user_timezone(user_id) )
+        except sqlalchemy.exc.ProgrammingError as err:
+            logger.error(err)
+            raise DatabaseException(
+                "Couldn't update user's timezone: %s" % str(err))
+
+def standardize_timezone(timezones):
```
mayhem, line 110: I'm not sure what this function really is supposed to do. Can you please improve the docstring?

mayhem (changes requested):
First off, please don't get discouraged by all the comments I've left you. This PR is great work for the first attempt at this feature and it implements the feature fully, which is great this early on!

Most of the comments I've left are more nuanced things that make your code flow more like our existing code -- things that you could not really know yet. Aside from having to implement this in React, rather than Flask there is nothing substantially wrong with this PR -- just a pile of minor improvements that you'll knock out quickly, I am sure.

Thanks!

mayhem:
Hi! I've run the code and it works fine, thanks!

I did notice a typo in my text suggestion for you: "Please select an appropriate timezone so that we may associated a timezone" The "associated" should not have a "d". Sorry my bad.

listenbrainz/webserver/views/profile.py
```diff
@@ -51,6 +54,35 @@ def reset_token():
...
+        try:
+            update_timezone = str(form.timezone.data)
+            db_usersetting.set_timezone(current_user.id, update_timezone)
+            flash.info("timezone reset")
+        except DatabaseException:
+            flash.error("Something went wrong! Unable to update timezone right now.")
+        return redirect(url_for("profile.info"))
+
+    if form.csrf_token.errors:
+        flash.error('Cannot update timezone due to error during authentication, please try again later.')
```
mayhem, line 75: For this error, lets give fewer details, since that might mean someone is trying to hack us. If that is the case, lets not give then any help. Lets have this say: "Unable to update timezone."

mayhem (changes requested):
Some minor wording changes, then lets add the test and SQL update script.

admin/sql/create_indexes.sql
```diff
@@ -57,4 +57,6 @@ CREATE INDEX user_id_ndx_pinned_recording ON pinned_recording (user_id);
 CREATE INDEX release_mbid_ndx_release_color ON release_color (release_mbid);
 CREATE UNIQUE INDEX caa_id_ndx_release_color ON release_color (caa_id);
 
+CREATE INDEX user_id_ndx_user_setting ON user_setting (user_id);
```
amCap1712, line 60: Can use `CREATE UNIQUE INDEX` instead of declaring attribute as UNIQUE in create_tables.sql. This also matches how we do it elsewhere.

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,101 @@
+import sqlalchemy
+from datetime import datetime, timedelta
+from listenbrainz import db
+from listenbrainz.db.exceptions import DatabaseException
+
+import logging
+
+logger = logging.getLogger(__name__)
```
amCap1712, line 8: Do not create loggers here. Instead use current_app.logger everywhere. You can look into other db files for reference.

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,101 @@
...
+                VALUES (:user_id, :timezone_name)
+                ON CONFLICT (user_id)
+                DO 
+                    UPDATE SET timezone_name = :timezone_name
+                """), {
+                "user_id": user_id,
+                "timezone_name": timezone_name,
+            })
+        except sqlalchemy.exc.ProgrammingError as err:
+            logger.error(err)
```
amCap1712, line 80: Same here. No log needed here.

listenbrainz/db/user_setting.py
```diff
@@ -0,0 +1,101 @@
...
+            raise DatabaseException(
+                "Couldn't update user's timezone: %s" % str(err))
+
+
+def standardize_timezone(timezones):
```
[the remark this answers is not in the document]
YouzhiL: Thanks for the suggestion. But the offset format is a little bit strange when I retrieve it. It has something like "(-1day, 15:00:00)". This function converts it to "(-9:00:00 GMT)".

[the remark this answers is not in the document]
amCap1712: I see, that's because psycopg2 adapts interval types to timedelta in python and the string representation for those is weird. There doesn't appear to be a built in method to nicely format timedeltas so indeed this function is fine.

mayhem (changes requested):
The create index call is missing from the SQL update script, other than that it looks good to me!

PR #2036, opened 2022-06-07
Update Flask, BU and other dependencies

amCap1712:
Update all possible dependencies.

### DIFF
--- a/listenbrainz/webserver/testing.py
+++ b/listenbrainz/webserver/testing.py
@@ -25,7 +25,8 @@ class ServerTestCase(flask_testing.TestCase):
         not_redirect = "HTTP Status %s expected but got %d" % (valid_status_code_str, response.status_code)
 
         self.assertIn(response.status_code, valid_status_codes, message or not_redirect)
-        self.assertTrue(response.location.endswith(location), message)
+        location_mismatch = "Expected redirect location %s but got %s" % (response.location, location)
+        self.assertTrue(response.location.endswith(location), message or location_mismatch)
 
     assert_redirects = assertRedirects
 


listenbrainz/db/dump.py
```diff
@@ -53,16 +54,16 @@
...
         'id',
         'created',
         'musicbrainz_id',
         'musicbrainz_row_id',
         # the following are dummy values for columns that we do not want to
         # dump in the public dump
-        'null',  # auth token
-        'to_timestamp(0)',  # last_login
-        'to_timestamp(0)',  # latest_import
+        SQL('null'),  # auth token
```
alastair, line 64: should we add some docs here explaining the difference between strings and these specific objects?

listenbrainz/db/dump.py
```diff
@@ -631,12 +632,20 @@ def copy_table(cursor, location, columns, table_name):
                      that should be dumped
             table_name: the name of the table to be copied
     """
+    fields = []
+    for column in columns:
```
alastair, line 636: perhaps a small doc here too explaining the difference between a composable and an Identifier (perhaps just one doc for these two items?)

listenbrainz/db/dump.py
```diff
@@ -783,8 +792,19 @@ def _import_dump(archive_path, db_engine: sqlalchemy.engine.Engine,
                         current_app.logger.info(
                             'Importing data into %s table...', file_name)
                         try:
-                            cursor.copy_from(tar.extractfile(member), '%s' % file_name,
-                                             columns=tables[file_name])
+                            fields = []
+                            for column in tables[file_name]:
```
alastair, line 796: this is duplicated on line 636, is it worth abstracting out? I guess the only difference between these two queries is that one is TO STDOUT and the other is TO STDIN?

listenbrainz/webserver/views/test/test_index.py
```diff
@@ -26,7 +26,8 @@ def test_index(self):
 
     def test_downloads(self):
         resp = self.client.get(url_for('index.downloads'))
-        self.assert_redirects(resp, url_for('index.data'))
+        self.assertEqual(resp.status_code, 302)
+        self.assertEqual(resp.location, url_for('index.data'))
```
alastair, line 30: we didn't do it in CB, but I realised that we have a base test class (ServerTestCase) where we could implement our own assertRedirects that checks both the status code and the location without the full protocol/hostname. Do you think it's worth it? Note that we have a "variant" of this test also in https://github.com/metabrainz/listenbrainz-server/blob/981e2fe6413e344e78f255a2d605fd5045d5fb11/listenbrainz/webserver/views/test/test_user.py#L48-L57

amCap1712: Makes sense, added

listenbrainz/db/dump.py
amCap1712: Refactored into a separate function.

listenbrainz/db/dump.py
amCap1712: Added

listenbrainz/db/dump.py
amCap1712: Added.

amCap1712:
I also tested doing a dump and importing it locally and it went fine.

listenbrainz/webserver/testing.py
```diff
@@ -15,6 +15,20 @@ def temporary_login(self, user_login_id):
...
+        if permanent:
+            valid_status_codes = (301, 308)
+        else:
+            valid_status_codes = (301, 302, 303, 305, 307, 308)
+
+        valid_status_code_str = ', '.join(str(code) for code in valid_status_codes)
+        not_redirect = "HTTP Status %s expected but got %d" % (valid_status_code_str, response.status_code)
+
+        self.assertIn(response.status_code, valid_status_codes, message or not_redirect)
+        self.assertTrue(response.location.endswith(location), message)
```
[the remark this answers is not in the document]
amCap1712: Done.

PR #2037, opened 2022-06-10
Add recent releases backend

amCap1712:
Recent Releases page shows recently released and soon to be released releases which the user may find relevant. To show only relevant releases, we show a release for a user only if the user has listened to an artist of that release in the past X days. Further, each release is assigned a confidence score for each user which is equal to the number of times the user has listened to the artists of that release in the past.


mayhem (approved):
Looks good -- even the spark queries are starting to make more sense to me now!

PR #2038, opened 2022-06-11
Dockerfile: Always install `postgresql-client` from current apt repository

chinmaykunkikar:
PostgreSQL doesn't currently support xenial.

1. This approach (`$(lsb_release -cs)`) will eliminate the need to change repository source names every time the base image is updated to a newer version.
2. This approach is mentioned on [PostgreSQL docs](https://www.postgresql.org/download/linux/ubuntu/).

Output of `cat /etc/apt/sources.list.d/pgdg.list`:
<pre>
deb http://apt.postgresql.org/pub/repos/apt/ <b>focal</b>-pgdg main 12
</pre>
...which is correct for current base image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,7 +36,7 @@ RUN apt-get update \
 # PostgreSQL client
 RUN curl https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
 ENV PG_MAJOR 12
-RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ xenial-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
+RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
 RUN apt-get update \
     && apt-get install -y --no-install-recommends postgresql-client-$PG_MAJOR \
     && rm -rf /var/lib/apt/lists/*


alastair (approved):
Thanks! I guess we didn't update this when we updated our base image (we've already had to make this change manually once before, making it automatic is a great idea)

chinmaykunkikar:
@alastair thanks :)

I have a question about base image -
Where in the dockerfile configs have me mentioned `ubuntu` or `ubuntu 20.04` as our base image? I couldn't find it anywhere.
I see gh workflow config files configured to use `ubuntu-latest`, which resolves to "ubuntu 20.04", as the base image for GH workflows. But those are only useful for gh workflows.

Or does docker defaults to ubuntu for base image so that there's no need to explicitly configure the base image in the dockerfiles?

amCap1712:
@chinmaykunkikar There is a hierarchy of base images in MeB to promote reuse among projects. The Ubuntu version is defined at the lowermost level.

- [ListenBrainz Image](https://github.com/metabrainz/listenbrainz-server/blob/d70112e6be391a5c53f8bf0b70bbc2c110576527/Dockerfile#L2) (used in all ListenBrainz service containers)
- [MetaBrainz Python 3.10 Image](https://github.com/metabrainz/docker-python/blob/680b0c84ce0053800a292d7f2c5af41bae3385dc/3.10/Dockerfile#L1) (Base image for MetaBrainz Python projects)
- [MetaBrainz Base Consul Image](https://github.com/metabrainz/docker-consul-template-base/blob/8d26a77b8e16f07ecd13fe1713d72fe8dbbd5ae9/Dockerfile#L1) (Base Image for all projects using consul, all python projects use consul so the python images are also based off this)
- [MetaBrainz Base Image](https://github.com/metabrainz/base-image/blob/0b78def90174ef172617e9ade500519b6668804d/Dockerfile#L1) (The base image for all metabrainz projects, docker consul image is based off this)
- The MetaBrainz Base image is based off [Phusion Base Image](https://github.com/phusion/baseimage-docker/blob/1443b1f1995abe8123bac7325ad3a3f1d7085851/image/Dockerfile#L1). Here, you can see Ubuntu being used.

PR #2039, opened 2022-06-13
Add useful messages and call to actions for empty states on listens page

chinmaykunkikar:
A newly signed up user usually needs to connect to their Spotify to track their listens and import any existing listens from last.fm.
The user will also most probably navigate to "Home" after signing up. They may or may not have read or the _"Get started..."_ paragraph on the main page of LB.

In such cases, have something for the user to do when they are on the Listens page and have no idea where to go from there.

The current listens page has no call to action links anywhere except in the profile button drop-down, which is hidden away from the interface.

<h2>Screenshots</h2>
<details><summary><b>Current Listens Page</b></summary>

</details>

<details><summary><b>Changes after this PR</b></summary>

</details>

---
### `Listens` section changes -
1. Suggest user to import their listening history and connect to music services.
2. Replace "No listens yet" with a simple Vinyl illustration for empty state.

---
### `ListenCountCard` changes -
Explicitly tell new users what the card shows instead of just showing them -
> "You have not listened to any songs so far"

It didin't tell new users what'd they see had they listened to any song.

The message -

> "Your listens count" followed by "You haven't listened to any songs."

...will subtly tell them what to expect in that place.

---
### `UserSocialNetwork` changes - 
1. Move UserSocialNetwork modals to individual cards.
2. Conditionally show helpful messages on all modals if their state is empty.
4. Adjust and add appropriate styles.


listenbrainz/webserver/static/js/src/follow/FollowerFollowingModal.tsx
```diff
@@ -50,13 +56,15 @@ export default class FollowerFollowingModal extends React.Component<
         <div className="text-center follower-following-pills">
           <div className="btn-group btn-group-justified" role="group">
             <Pill
+              style={{ margin: "0px 6px" }}
```
MonkeyDo, line 59: This would be better done in the follow.less file where `.follower-following-pills` is defined

listenbrainz/webserver/static/js/src/follow/FollowerFollowingModal.tsx
```diff
@@ -65,24 +73,46 @@ export default class FollowerFollowingModal extends React.Component<
...
-                )}
-                updateFollowingList={updateFollowingList}
-              />
-            );
-          })}
-        </div>
+
+        {activeModeList.length === 0 ? (
+          <>
+            <hr style={{ margin: "0px 2em", borderTop: "1px solid #eee" }} />
```
MonkeyDo, line 79: Maybe better defined in a less file as well. Would it work as a border-top and margins applied to the follower-following-empty class?

listenbrainz/webserver/static/js/src/follow/FollowerFollowingModal.tsx
```diff
@@ -65,24 +73,46 @@ export default class FollowerFollowingModal extends React.Component<
...
+            ) : (
+              <div className="follower-following-empty text-center text-muted">
+                {user.name === currentUser?.name
+                  ? "You aren't"
+                  : `${user.name} isn't`}{" "}
+                following anyone.
+              </div>
+            )}
+          </>
+        ) : (
```
MonkeyDo, line 96: The nested ternary operations get pretty hard to read, I think it would be better to have something like this for legibility:
```jsx
if(activeModeList.length === 0){
  if(activeMode === "follower"){
  // return "no followers" markup
  }

  // return "not following" markup
}
// return regular markup
return <div className="follower-following-list">…
```

listenbrainz/webserver/static/js/src/follow/SimilarUsersModal.tsx
```diff
@@ -25,22 +25,33 @@ const SimilarUsersModal = (props: SimilarUsersModalProps) => {
...
-              mode="similar-users"
-              key={listEntry.name}
-              user={listEntry}
-              loggedInUserFollowsUser={loggedInUserFollowsUser(listEntry)}
-              updateFollowingList={updateFollowingList}
-            />
-          );
-        })}
-      </div>
+      {similarUsersList.length === 0 ? (
```
MonkeyDo, line 31: Same comment as above to break up the ternary expression

listenbrainz/webserver/static/js/src/follow/SimilarUsersModal.tsx
```diff
@@ -25,22 +25,33 @@ const SimilarUsersModal = (props: SimilarUsersModalProps) => {
...
+        <>
+          <hr style={{ margin: "0px 2em", borderTop: "1px solid #eee" }} />
```
MonkeyDo, line 33: Should probably be a style for the `similar-users-empty` class defined in a less file as well

listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
```diff
@@ -28,11 +26,16 @@ const ListenCountCard = (props: ListenCountCardProps) => {
...
-          : `${user.name} has not listened to any songs so far`}
-      </p>
+      <>
+        <p className="text-muted">
+          {isCurrentUser ? "Your" : `${user.name}'s`} listens count
+        </p>
+        <hr style={{ margin: "10px 0px" }} />
+        <div style={{ fontSize: "14px" }} className="text-muted">
+          {isCurrentUser ? "You haven't" : `${user.name} hasn't`} listened to
+          any songs.
```
MonkeyDo, line 36: What do you think about this optimistic note? :)

```suggestion
          any songs yet.
```

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -687,19 +688,30 @@ export default class Listens extends React.Component<
...
+                  listened to any songs yet.
+                </div>
+                {currentUser?.name === user?.name && (
+                  <div className="empty-action">
+                    Import <a href="/profile/import/">your listening history</a>{" "}
+                    and track your listens by{" "}
+                    <a href="/profile/music-services/details/">
+                      connecting to a music service
+                    </a>
+                    , or use <a href="/add-data/">one of the players</a> to
```
MonkeyDo, line 711:
```suggestion
                    , or use <a href="/add-data/">one of these music players</a> to
```

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -687,19 +688,30 @@ export default class Listens extends React.Component<
...
+                <div className="lead empty-text">Empty record</div>
+                <div className="empty-message text-muted">
+                  {currentUser?.name === user?.name
+                    ? "You haven't"
+                    : `${user.name} hasn't`}{" "}
+                  listened to any songs yet.
+                </div>
+                {currentUser?.name === user?.name && (
+                  <div className="empty-action">
+                    Import <a href="/profile/import/">your listening history</a>{" "}
```
MonkeyDo, line 706: Maybe spelling this out? What do you think?
```suggestion
                    Import <a href="/profile/import/">your listening history</a> from Last.fm/Libre.fm{" "}
```

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -687,19 +688,30 @@ export default class Listens extends React.Component<
...
+                    and track your listens by{" "}
+                    <a href="/profile/music-services/details/">
+                      connecting to a music service
```
MonkeyDo, line 709:
```suggestion
                      connecting to a music streaming service
```

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -687,19 +688,30 @@ export default class Listens extends React.Component<
...
-            )}
+            {user && <UserSocialNetwork user={user} newAlert={newAlert} />}
           </div>
           <div className="col-md-8 col-md-pull-4">
             {!listens.length && (
-              <div className="lead text-center">
-                <p>No listens yet</p>
+              <div className="empty-listens">
+                <FontAwesomeIcon icon={faRecordVinyl as IconProp} size="10x" />
+                <div className="lead empty-text">Empty record</div>
```
MonkeyDo, line 697: Somehow this doesn't quite sound right.I like the play on words though, so I'm a little torn.
I keep playing it in my head (like a record) and I'm not convinced, but I don't have any good suggestion.
Maybe "Get started" ?

MonkeyDo (commented):
Loving your work !
Thank you for this good improvement!

listenbrainz/webserver/static/js/src/user/Listens.tsx
MonkeyDo: In line with that, I didn't fully recognize a vinyl icon.
I have a few suggestions for icons I think would work well in context:
- https://fontawesome.com/icons/music-slash?s=solid
- https://fontawesome.com/icons/user-music?s=solid 
- https://fontawesome.com/icons/album-collection-circle-plus?s=solid 
- maybe weird? https://fontawesome.com/icons/head-side-headphones?s=solid

listenbrainz/webserver/static/js/src/follow/FollowerFollowingModal.tsx
chinmaykunkikar: Ah, I need to improve my relationship with CSS 😅

listenbrainz/webserver/static/js/src/follow/FollowerFollowingModal.tsx
chinmaykunkikar: No because the <hr/> is not a  child of `follower-following-empty` class. I have applied the properties on <hr/> element in follow.less file. It's working.

listenbrainz/webserver/static/js/src/follow/FollowerFollowingModal.tsx
chinmaykunkikar: I agree, nested ternary operations are confusing to read. I was trying to use if-else statements to get it done because it was hurting the readability. But it turns out you can't use normal if-else statements in JSX. It is explained in the following links why is that the case:
1. https://react-cn.github.io/react/tips/if-else-in-JSX.html
2. https://reactjs.org/docs/conditional-rendering.html
3. https://www.digitalocean.com/community/tutorials/7-ways-to-implement-conditional-rendering-in-react-applications

I can try separating it out as a new component though. I'll give you an update on this.

listenbrainz/webserver/static/js/src/listens/ListenCountCard.tsx
chinmaykunkikar: I think I like it :)

listenbrainz/webserver/static/js/src/user/Listens.tsx
chinmaykunkikar: The reason I didn't consider writing "_streaming_" originally was because the [music services page](https://listenbrainz.org/profile/music-services/details/) has an option to also connect with CritiqueBrainz, which is a music service but isn't a music streaming service.

Need I be this particular about the wording?

listenbrainz/webserver/static/js/src/follow/SimilarUsersModal.tsx
chinmaykunkikar: Applying the properties to hr element in follow.less earlier took care of this hr as well :)

listenbrainz/webserver/static/js/src/follow/FollowerFollowingModal.tsx
MonkeyDo: Ah, there is a subtlety here: you cannot use if-else statements *inside a JSX expression* meaning in a JSX block (the part that starts with an opening html bracket `<` ) , but you can definitely use them outside of it in the same file.
Any react component after all is just a standard javascript function…

From your first link (emphasis mine):
> If a ternary expression isn't robust enough, you can use if statements *outside of your JSX* to determine which components should be used

listenbrainz/webserver/static/js/src/follow/FollowerFollowingModal.tsx
MonkeyDo: It's a good skill to have :)

To be honest, for simple things its very accessible.

listenbrainz/webserver/static/js/src/follow/SimilarUsersModal.tsx
MonkeyDo: Even better !

listenbrainz/webserver/static/js/src/user/Listens.tsx
MonkeyDo: I think in this context (getting new user onboarded) I would prefer spelling it out, even though CB is also in there.

listenbrainz/webserver/static/js/src/user/Listens.tsx
MonkeyDo: Dang,  I didn't realized those were all "pro" icons…
Welp, perhaps just the CD then? https://fontawesome.com/icons/compact-disc?s=solid

I'm with you on "get started" being a bit meh. 
Following your suggestion of "get playing", after some though I think "Start listening" or "Get listening" would work well (considering we call the pieces of data "listens")

listenbrainz/webserver/static/css/follow.less
```diff
@@ -75,3 +82,12 @@
...
 }
+
+.follower-following-empty, .similar-users-empty {
+  padding: 10px 20px;
+}
+
+hr {
+  margin: 0px 2em;
+  border-top: 1px solid #eee;
+}
```
MonkeyDo, line 93: This now applies to all `hr` elements across the website. 
The fact that we're in a separate .less file makes no difference to that, in the end everything is compiled into a single css file.

If you want it to apply only to hr inside certain classes, you can move this whole block inside of it like this:
```less
.follower-following-empty, .similar-users-empty {
  padding: 10px 20px;
  hr {
    margin: 0px 2em;
    border-top: 1px solid #eee;
  }
}
```

This will be compiled as 
```less
.follower-following-empty hr,  .similar-users-empty hr {…}
```
This nesting capability comes from LESS, it's not a native CSS feature.

listenbrainz/webserver/static/css/follow.less
```diff
@@ -47,8 +47,16 @@
...
+.card-user-sn {
+  margin-bottom: 20px;
+}
+
 .follower-following-pills {
-  padding-bottom: 10px;
+  padding: 20px 0px;
+
+  button {
+    margin: 0px 6px !important;
```
MonkeyDo, line 58: As much as possible we try to stay away from `!important`, but after starting to write a whole paragraph, I realized looking into the page code that the pills styles are applied directly to the element, not with a class.
Unfortunate, but so it goes.
These are the only good reasons to use `!important` :)

listenbrainz/webserver/static/js/src/user/Listens.tsx
chinmaykunkikar: I like the CD as well! :)

And "Start listening" goes really well with Listens. I'll do the changes according to this then :)

listenbrainz/webserver/static/css/follow.less
chinmaykunkikar: Oh right! I forgot it compiles to a single css file. I'll work on it.

listenbrainz/webserver/static/js/src/follow/SimilarUsersModal.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/prop-types */
```
MonkeyDo, line 1: This isn't the way to go to fix an issue ;)

Here, removing the disable rule gives us an error that we should pay attention to: something not right with the props. The error itself is not very understandable, but it points to an issue with how we're using the props.

In this case, React doesn't like the pattern you're using: each time any of the props change, the `renderSimilarUsersList` function will be recreated. You can mamke a memoized version of it that will detect changes to certain props dependencies and only render if those dependencies have changed.

I will comment directly on the lines concerned

chinmaykunkikar:
I'll re run tests, update snapshots and resolve the merge conflict in some time.

MonkeyDo:
Just a heads up that with the last commit there is duplicated code:
The section that starts with `return (      <div className="similar-users-list">…` is there twice, once inside the renderSimilarUsersList callback and once after it (which is probably left over from before)

PR #2040, opened 2022-06-15
Include user submitted mbids in spark dumps

amCap1712:
This allows us to fall back to user submitted mbids when those are available and the listen was unmapped.

### DIFF
--- a/listenbrainz/listenstore/dump_listenstore.py
+++ b/listenbrainz/listenstore/dump_listenstore.py
@@ -353,9 +353,9 @@ class DumpListenStore:
                           , mm.recording_mbid::TEXT AS m_recording_mbid
                           , data->'track_metadata'->'additional_info'->>'recording_mbid' AS l_recording_mbid
                        FROM listen l
-            FULL OUTER JOIN mbid_mapping mm
+                  LEFT JOIN mbid_mapping mm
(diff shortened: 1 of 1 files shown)


listenbrainz/listenstore/dump_listenstore.py
```diff
@@ -333,26 +333,33 @@ def write_parquet_files(self,
...
+                          , release_name AS m_release_name
+                          , data->'track_metadata'->>'release_name' AS l_release_name
+                          , release_mbid::TEXT AS m_release_mbid
+                          , data->'track_metadata'->'additional_info'->>'release_mbid' AS l_release_mbid
+                          , recording_name AS m_recording_name
+                          , track_name AS l_recording_name
+                          , mm.recording_mbid::TEXT AS m_recording_mbid
+                          , data->'track_metadata'->'additional_info'->>'recording_mbid' AS l_recording_mbid
+                       FROM listen l
+            FULL OUTER JOIN mbid_mapping mm
```
mayhem, line 356: I never got around to making sure we needed full outer join here. Didn't we think that a left join would suffice? Should we do this now that we're mucking with this?

mayhem (approved):
Thanks!

listenbrainz/listenstore/dump_listenstore.py
amCap1712: Yes, a LEFT JOIN should be enough. will update.

PR #2041, opened 2022-06-16
Lookup MBIDs for now playing listens

amCap1712:
Now Playing listens are received on frontend using Websockets and through the listens in props on initial page load. Now playing listens don't have msids or mbid mapping data when first received. To add mbid mapping data to now playing listens, we lookup the metadata using the track and artist name and according update the listen.

### DIFF
--- a/listenbrainz/webserver/static/js/src/user/Listens.tsx
+++ b/listenbrainz/webserver/static/js/src/user/Listens.tsx
@@ -119,7 +119,8 @@ export default class Listens extends React.Component<
   componentDidMount(): void {
     const { newAlert } = this.props;
     // Get API instance from React context provided for in top-level component
-    const { APIService, currentUser } = this.context;
+    const { APIService } = this.context;
+    const { playingNowListen } = this.state;
     this.APIService = APIService;
 
     this.connectWebsockets();
@@ -141,6 +142,9 @@ export default class Listens extends React.Component<
             error?.toString()
           );
         });
+    }
+    if (playingNowListen) {
+      this.receiveNewPlayingNow(playingNowListen);
     }
     this.loadFeedback();
   }
@@ -211,7 +215,8 @@ export default class Listens extends React.Component<
       this.receiveNewListen(data);
     });
     this.socket.on("playing_now", (data: string) => {
-      this.receiveNewPlayingNow(data);
+      const playingNow = JSON.parse(data) as Listen;
+      this.receiveNewPlayingNow(playingNow);
     });
   };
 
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -612,6 +658,77 @@ export default class Listens extends React.Component<
...
+    const canDelete =
+      isCurrentUser && Boolean(listenedAt) && Boolean(recordingMSID);
+
+    const isListenReviewable =
+      Boolean(recordingMBID) ||
+      artistMBIDs?.length ||
+      Boolean(trackMBID) ||
+      Boolean(releaseGroupMBID);
+    // All listens in this array should have either an MSID or MBID or both,
+    // so we can assume we can pin them. Playing_now listens are displayed separately.
```
MonkeyDo, line 682: This comment should probably be reviewed since playing_now listens are not separate anymore

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -243,9 +248,17 @@ export default class Listens extends React.Component<
...
+    const playingNow = newPlayingNow;
+    const { APIService } = this.context;
+    const metadata = await APIService.lookupRecordingMetadata(
+      playingNow.track_metadata.track_name,
+      playingNow.track_metadata.artist_name,
+      false
+    );
+    playingNow.track_metadata.mbid_mapping = metadata as MbidMapping;
+
+    this.loadFeedbackForNowPlaying(playingNow);
```
MonkeyDo, line 261: This should be wrapped in a try-catch block in case the request fails.

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -243,9 +248,17 @@ export default class Listens extends React.Component<
     }
   };
 
-  receiveNewPlayingNow = (newPlayingNow: string): void => {
-    const playingNow = JSON.parse(newPlayingNow) as Listen;
-    playingNow.playing_now = true;
```
MonkeyDo, line 248: We used to manually add the playing_now = true, IIRC because not all playing now listened we received had it, and we rely on that field in other components (ListenCard).
Are you 100% sure this is not needed anymore?

test.sh
```diff
@@ -65,6 +65,10 @@ function docker_compose_run {
     invoke_docker_compose run --rm --user `id -u`:`id -g` "$@"
 }
 
+function docker_compose_run_fe {
+  invoke_docker_compose run --rm frontend_tester npm run "$@"
+}
+
```
MonkeyDo, line 71: This and the rest of the changes for this file can probably be removed after updating from master?

MonkeyDo (commented):
Looks sane !
I haven't tried it locally, but happy to help test once deployed on test or beta

test.sh
amCap1712: Done.

listenbrainz/webserver/static/js/src/user/Listens.tsx
amCap1712: Done.

listenbrainz/webserver/static/js/src/user/Listens.tsx
amCap1712: Done.

MonkeyDo (approved):
💥

PR #2042, opened 2022-06-17
Fix listen dedup in database

amCap1712:
The Unique Index used for deduplication in the listen table is case-sensitive which is undesirable. msids already ignores casing while calculating the hash. Update the index to be case-insensitive. Found while investigating LB-1109.

Hypertables do not support concurrent or per chunk indexing so we cannot use either of those optimizations for listen table. My intent for the upgrade is to stop TS writer, add the new index and start TS writer with new changes. The old index will be kept around for a while so that it can be used in case we need to downgrade a service later.


mayhem (approved):
Done with the easy part. :)

amCap1712:
On hold, because intend to fix with LB-1118

PR #2045, opened 2022-06-29
Debug Github Actions Failure

amCap1712:


alastair:
I'm a little unsure about going back to running these as root - especially because
* On local development, anything that runs in docker and creates files (e.g. snapshot creation) will create these new files as root
* we've already discussed running things as non-root users in prod and everywhere else, and so I'd like to standardise on this

However, if this fix currently works for the issue, let's go for it. We can address it again when we get to our security update changes.
I did have a quick look around and found https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user which is interesting, it says

> Docker actions must be run by the default Docker user (root). Do not use the USER instruction in your Dockerfile, because you won't be able to access the GITHUB_WORKSPACE. For more information, see "[Using environment variables](https://docs.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables)" and [USER reference](https://docs.docker.com/engine/reference/builder/#user) in the Docker documentation

Another possible fix to this might be to change the uid of our in-container user to the user that we're running as. We do this in freesound: https://github.com/MTG/freesound/blob/master/docker/entrypoint.sh#L10

```
if [[ -v FS_USER_ID && $FS_USER_ID != 0 && -n $FS_USER_ID ]]; then
    echo Set fsweb UID to: "$FS_USER_ID"
    usermod -u "$FS_USER_ID" fsweb

    exec gosu fsweb "$@"
else
    exec "$@"
fi
```

this could be a solution for when we introduce our own user inside docker.

alastair (approved):

PR #2046, opened 2022-06-30
Daily jams fine tuning

mayhem:
This PR improves the daily jams feature by:

1. Not generating models each day (only sundays as the last thing, ready for monday's recs)
2. Use the latest troi release
3. Put debugging playlists under troi-bot-debug user, freeing up troi-bot for non dev purposes.

@amCap1712 : Please sanity check the modified cron calls -- I forget if we need to create dataframes (I think so) even if we're not generating a model each time.


amCap1712:
As discussed in IRC< we currently cannot train models and use those separately.

mayhem:
Updated to back to to calculating a model daily and wasting less time with gaps in the crontab.

amCap1712 (approved):
LGTM, Thanks!

PR #2048, opened 2022-07-04
Display error message on Last.fm import page if user has no email 

amCap1712:
Multiple users have reported issues using the Last.fm importer. A few times the reason behind the issues was that the user hadn't added an email to their account. To improve the UX somewhat, do not show the importer at all if the user doesn't have an email associated.

This is what the error looks like for users without email opening the importer page.

Screenshot from 2022-07-05 00-30-21

### DIFF
--- a/listenbrainz/webserver/templates/user/import.html
+++ b/listenbrainz/webserver/templates/user/import.html
@@ -4,53 +4,54 @@
 
 {% block content %}
     <h2 class="page-title">Import to user {{ user.musicbrainz_id }}</h2>
-    {% if user_has_email %}
-        {% if current_user.is_authenticated and current_user == user %}
-
-            <p>
-                Most users will want to import from Last.fm directly.
-                <br>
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/templates/user/import.html
```diff
@@ -3,45 +3,55 @@
...
+            Clicking the "Import now!" button will import your profile now without the need to open lastfm.<br/>
+            You need to keep this page open for the tool to work, it might take a while to complete. Though, you can
+            continue doing your work. :)
+        </p>
+        <p>
+            <em>NOTE</em>: Please note that your listen counts may not be accurate for up to 24 hours after
+            you import your data!
+            Also, running the Last.fm importer twice will not cause duplicates in your listen history.
+        </p>
+        <div class="row">
```
MonkeyDo, line 43: I would wrap this whole row with the same `{% if user_has_email %}` to avoid the empty container

amCap1712: Done.

PR #2049, opened 2022-07-05
Simplify the code to get Youtube URL from a listen

staticssleever668:
# Problem

If `origin_url` in submitted listen contains a shortened YouTube link (`youtu.be`), modified function `getYoutubeURLFromListen` doesn't return anything.

So, when you hover on the cover art of the listen in the profile, the link leads to user's profile page (not sure this is intended) instead of YouTube. 

Example: 
See the first listen [here](https://listenbrainz.org/user/waggery_assonant/?min_ts=1657020000).

# Solution

There is already a `getVideoIDFromListen` function, which works with `youtu.be` links as well, so instead of having two separate places with link parsing, just use the dedicated function. 

# Action

I couldn't get local environment to work (I can't sign into MusicBrainz account), so please, try this in practice.
I'm not even sure the problem is actually related to the modified file. :)


listenbrainz/webserver/static/js/src/brainzplayer/YoutubePlayer.tsx
```diff
@@ -92,25 +95,12 @@ export default class YoutubePlayer
   static getYoutubeURLFromListen(
     listen: Listen | JSPFTrack
   ): string | undefined {
-    // Checks if there is a youtube ID in the listen
-    const youtubeId = _get(listen, "track_metadata.additional_info.youtube_id");
+    const youtubeId = this.getVideoIDFromListen(listen);
```
MonkeyDo, line 98: My concern with this is that we don't know for sure what the format of youtube_id will be (is it always just the id, or will users send the full URL? for the similar spotify_id we [recommend sending the full url](https://listenbrainz.readthedocs.io/en/latest/users/json.html#payload-json-details) for example).
However I do agree with this decuplication of code.

One way to make sure we strip the potential URL would be to modify `getVideoIDFromListen`: in the new `if (youtubeId) {` block we can use the existing `youtubeURLRegexp`  to test for the URL, and return
 `return match?.[1] ?? youtubeId` -> if match is an array and the second items exists return that, otherwise default to returning the youtube_id directly.

MonkeyDo (changes requested):
Thanks for cleaning up that code ! :)

The cover art link pointing to the current page is a separate bug, I'll open a ticket for it, thanks !

listenbrainz/webserver/static/js/src/brainzplayer/YoutubePlayer.tsx
staticssleever668: Thanks for reviewing! I changed the patch as described and added some comments to make it clear  `youtube_id` can be either of URL and ID.

I tested it like so (Node JS code):

```node
function _isString(input) {
  return typeof input === "string" || input instanceof String;
}

function getVideoIDFromListen(listen_fields) {
  const youtubeId = listen_fields["id"];

  const videoIdOrUrl = youtubeId ?? listen_fields["url"];
  if (_isString(videoIdOrUrl) && videoIdOrUrl.length) {
    const youtubeURLRegexp =
      /(?:https?:\/\/)?(?:www\.|m\.)?youtu(?:\.be\/|be.com\/\S*(?:watch|embed)(?:(?:(?=\/[^&\s?]+(?!\S))\/)|(?:\S*v=|v\/)))([^&\s?]+)/g;
    const match = youtubeURLRegexp.exec(videoIdOrUrl);
    return match?.[1] ?? youtubeId;
  }

  return undefined;
}

const assert = require("assert");

assert(
  getVideoIDFromListen({
    id: "123",
    url: "https://youtu.be/123",
  }) === "123"
);

assert(
  getVideoIDFromListen({
    url: "https://youtu.be/123",
  }) === "123"
);

assert(
  getVideoIDFromListen({
    id: "123",
  }) === "123"
);

assert(
  getVideoIDFromListen({
    id: "https://youtu.be/123",
  }) === "123"
);

assert(
  getVideoIDFromListen({
    id: "https://youtu.be/123",
    url: "https://youtu.be/123",
  }) === "123"
);

// It is impossible to do better in such case.
assert(
  getVideoIDFromListen({
    id: "https://bogus_link_in_id",
    url: "https://bogus_link",
  }) === "https://bogus_link_in_id"
);

assert(
  getVideoIDFromListen({
    url: "https://bogus_link",
  }) === undefined
);

assert(getVideoIDFromListen({}) === undefined);
```

MonkeyDo (approved):
Thanks a lot for fixing this !

PR #2053, opened 2022-07-06
Fix data-tip in Critiquebrainz Modal 

anshg1214:
* Fix HTML not being rendered for data-tip
* Removed extra space before entity type


MonkeyDo (approved):
Thanks !

listenbrainz/webserver/static/js/src/cb-review/CBReviewModal.tsx
```diff
@@ -84,7 +84,7 @@ export default class CBReviewModal extends React.Component<
           style={{ color: "black" }}
         />
       </span>
-      <ReactTooltip place="bottom" globalEventOff="click" clickable />
+      <ReactTooltip place="bottom" globalEventOff="click" clickable html={true} />
```
amCap1712, line 87: To comply with lint:
```suggestion
      <ReactTooltip place="bottom" globalEventOff="click" clickable html />
```

listenbrainz/webserver/static/js/src/cb-review/CBReviewModal.tsx
```diff
@@ -546,7 +546,7 @@ export default class CBReviewModal extends React.Component<
               data-toggle="dropdown"
               type="button"
             >
-              {`${entityToReview.name} (
+              {`${entityToReview.name}(
               ${entityToReview.type.replace("_", " ")})`}
```
MonkeyDo, line 550: I think this should work to remove the space before the entity type while keeping the space between the entity name and the parens:
```suggestion
              {`${entityToReview.name} (${entityToReview.type.replace(
                "_",
                " "
              )})`}
```

PR #2054, opened 2022-07-06
Fix spotify player

MonkeyDo:
We have two new-ish issues linked to an update of the Spotify Web SDK (Spotify's JS player library):

1. Tracks played with the SDK on LB don't appear in the spotify listening history anymore (and consequently aren't seen by LB's spotify importer), so BrainzPlayer needs to submit the listens from now on.
    Let's hope they don't reverse course on that implementation and create duplicates for us…
   No response on https://community.spotify.com/t5/Spotify-for-Developers/Tracks-played-with-Web-Playback-SDK-don-t-appear-in-recent/td-p/5386327
2. Detecting the end of a track was always a hack, and some recent implementation change broke it.
    The player now receives state updates when you play from another device, and brainzplayer wrongly detects that as the end of the currently playing track, and so it plays the next song on BP, taking back control of spotify from the device you were trying to use it on.
    No responses on https://community.spotify.com/t5/Spotify-for-Developers/Web-Playback-SDK-Detect-Track-End/m-p/5391515 so apart from that the general implementation of track detection stays the same (until next time…)


listenbrainz/webserver/static/js/src/utils/types.d.ts
```diff
@@ -137,7 +137,8 @@ declare type SpotifyPlayerSDKState = {
   position: number;
   duration: number;
   track_window: {
-    current_track: SpotifyTrack;
+    current_track: SpotifyTrack | null;
+    previous_tracks?: SpotifyTrack[]; // always an array even if empty
```
amCap1712, line 141: Where and how is previous_tracks? I see a comment about it in SpotifyPlayer.tsx but there is no code change accompanying so a bit confused.

MonkeyDo: Right, that's left over from me trying to fix the track end detection, but that was not compatible with another fix in this PR reported by mayhem (https://github.com/metabrainz/listenbrainz-server/pull/2054/files#diff-939dbd0e74ca9fa5dea63f2a3c6a7600a5de30a317cef08fd44f9dbb495f2712R471-R475)

I'll remove it

amCap1712 (approved):
LGTM, Thanks!

PR #2055, opened 2022-07-06
Personal recommendation backend

hrik2001:
This PR implements the backend that is necessary to support sending personal recording recommendation to your followers.

The backend is implemented in the following manner:
- When personal recommendation is given by user A to user B
- First it is checked that if B is a follower of A or not
- If yes, then A's id is added to `user_id` column of `user_timeline_event` table, while the details of B and the song and text of the recommendation is stored in the metadata

_This PR will only implement the backend needed for personal recommendation. Implementation of hiding of personal recording recommendation will be implemented as separate PR_


listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -732,3 +812,50 @@ def get_recording_pin_events(
...
+                recording_mbid=event.metadata.recording_mbid,
+                recording_msid=event.metadata.recording_msid,
+                recommendee_id=event.metadata.recommendee_id,
+                blurb_content=event.metadata.blurb_content
+            )
+
+            events.append(APITimelineEvent(
+                id=event.id,
+                event_type=UserTimelineEventType.PERSONAL_RECORDING_RECOMMENDATION,
+                user_name=id_following_username_map[event.user_id],
```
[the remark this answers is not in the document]
amCap1712: Note that this case doesn't happen in general recommendations because there we show recommendations from whomever we are currently following.

amCap1712 (approved):
Sorry for the delay. The PR looks great. 1 case needs to be handled and some minor refactorings.

amCap1712 (changes requested):
Uh! approved by error, still needs some changes.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -493,6 +503,76 @@ def unhide_user_timeline_event(user_name):
...
+    metadata = data['metadata']
+
+    try:
+        metadata_db = {
+            "artist_name": metadata["artist_name"],
+            "track_name": metadata["track_name"],
+            "recording_msid": metadata["recording_msid"],
+            "release_name": metadata["release_name"],
+            "recording_mbid": metadata["recording_mbid"],
+            "recommendee_id": 0,
```
amCap1712, line 554: I am unsure about this one. If we assign recommendee_id later on then it won't be validated. But if we do it for each user then the remaining fields are validated multiple times. That is a bit wasteful but unless we know its a noticeable slowdown, I suggest that we validate recommendee_id as well.

To be clear, validate each metadata first and throw a 400 if any validation error happens. After all are validated, then start inserting in database.

hrik2001:
@amCap1712 I have implemented the backend of personal recommendation as we talked about in the IRC (ie storing an array of followers in the metadata). Feel free to review when you are free. Thanks!

listenbrainz/tests/integration/test_user_timeline_event_api.py
```diff
@@ -815,3 +815,123 @@ def test_get_cb_review_events(self, mock_requests):
         self.assertEqual(metadata_3["rating"], review_event_3["metadata"]["rating"])
         self.assertEqual(metadata_3["text"], review_event_3["metadata"]["text"])
         self.assertEqual(review_mbid_3, review_event_3["metadata"]["review_mbid"])
+
```
amCap1712, line 818: These tests look good we should probably also add 1 more where the user tries to recommend to a user they don't follow.

listenbrainz/tests/integration/test_user_timeline_event_api.py
```diff
@@ -815,3 +815,123 @@ def test_get_cb_review_events(self, mock_requests):
...
+    def test_personal_recommendation_writes_to_db(self):
+        # Let's create 2 users, who follow the request sender
+        user_one = db_user.get_or_create(2, "riksucks")
```
amCap1712, line 821: In this test, it would be nice to test that the personal recommendation is still after the user has unfollowed the user who recommended. Whether we want to show those or not, may be a choice but the aim of the test is to ensure that case doesn't crash the page.

listenbrainz/db/user_timeline_event.py
```diff
@@ -170,6 +183,40 @@ def get_recording_recommendation_events_for_feed(user_ids: List[int], min_ts: in
...
 
 
+def get_personal_recommendation_events_for_feed(user_id: int, min_ts: int, max_ts: int, count: int) -> List[UserTimelineEvent]:
+    """ Gets a list of personal_recording_recommendation events for specified users.
+
+    user_ids is a tuple of user row IDs.
+    """
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT user_timeline_event.id
```
amCap1712, line 193: I reformatted the query myself to make a bit easier to read. Hope you don't mind.

listenbrainz/db/user_timeline_event.py
```diff
@@ -170,6 +183,40 @@ def get_recording_recommendation_events_for_feed(user_ids: List[int], min_ts: in
...
+                 , user_timeline_event.user_id
+                 , user_timeline_event.event_type
+                 , user_timeline_event.metadata
+                 , user_timeline_event.created
+                 , "user".musicbrainz_id as user_name
+              FROM user_timeline_event
+              JOIN "user"
+                ON user_timeline_event.user_id = "user".id
+             WHERE (user_timeline_event.metadata ->> 'followers')::jsonb @> (:user_id)::text::jsonb
```
amCap1712, line 202: AFAIK, `(user_timeline_event.metadata ->> 'followers')::jsonb` should be equivalent to `user_timeline_event.metadata -> 'followers'`. Notice one less arrow. `->` keeps the output as jsonb whereas `->>` converts it to text. Since we want jsonb here anyway use the former and avoid the cast. Also, no need to cast user_id to jsonb, just casting to text is sufficient.

See https://stackoverflow.com/a/46206119 for reference.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -493,6 +500,59 @@ def unhide_user_timeline_event(user_name):
...
+    if user_name != user['musicbrainz_id']:
+        raise APIUnauthorized("You don't have permissions to post to this user's timeline.")
+
+    try:
+        data = ujson.loads(request.get_data())
+    except ValueError as e:
+        raise APIBadRequest(f"Invalid JSON: {str(e)}")
+
+    metadata = data['metadata']
+
```
amCap1712, line 545: There should be a is_following check here for all users, no? I'd suggest to create a new function db_follow which receives an array and checks that the users follows all of the other users. Also, the metadata will need to convert usernames  of the person to whom the track is recommended into user ids. 

We want this so that external api users can easily use this api (they don't have access to listenbrainz user ids , at least not by our intention). Secondly, so that everything keeps working smoothly when a user renames themself.

amCap1712 (changes requested):
Hi! Sorry for the late review. PR looks great, only some changes needed before its ready to merge.

listenbrainz/db/user_relationship.py
```diff
@@ -58,6 +58,29 @@ def is_following_user(follower: int, followed: int) -> bool:
         })
         return result.fetchone()['cnt'] > 0
 
+def multiple_users_by_username_following_user(followed: int, followers: List[str]):
+    # returns a dictionary, keys being usernames
```
amCap1712, line 62: Docstring missing.

listenbrainz/db/user_relationship.py
```diff
@@ -58,6 +58,29 @@ def is_following_user(follower: int, followed: int) -> bool:
...
+    # values being boolean
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT "user".musicbrainz_id,
+                   coalesce((
+                    SELECT "user".id = user_0
```
amCap1712, line 68: can change to `SELECT 't'`

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -493,6 +500,66 @@ def unhide_user_timeline_event(user_name):
...
+    try:
+        metadata = PersonalRecordingRecommendationMetadata(**metadata)
+        follower_results = db_user_relationship.multiple_users_by_username_following_user(user['id'], metadata.followers_username)
+        non_followers = []
+        for follower in metadata.followers_username:
```
amCap1712, line 550: We should probably call the key name in the api `users` instead of followers because we may want to change the follower restriction in future etc.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -493,6 +500,66 @@ def unhide_user_timeline_event(user_name):
...
+            if not follower_results[follower]:
+                non_followers.append(follower)
+        if non_followers:
+            raise APIBadRequest(f"These people don't follow you {str(non_followers)}")
```
amCap1712, line 554: Probably also add that You cannot tracks to people who don't follow you.

listenbrainz/db/user_timeline_event.py
```diff
@@ -170,6 +217,53 @@ def get_recording_recommendation_events_for_feed(user_ids: List[int], min_ts: in
...
+                        'followers_username', jsonb_agg("user".musicbrainz_id),
+                        'blurb_content', user_timeline_event.metadata -> 'blurb_content'
+                    ) AS metadata
+                    FROM jsonb_array_elements_text(user_timeline_event.metadata -> 'followers') AS arr
+                   INNER JOIN "user" ON arr.value::int = "user".id
+                 )
+                 , user_timeline_event.created
+                 , "user".musicbrainz_id as user_name
+              FROM user_timeline_event
+              JOIN "user"
```
amCap1712, line 247: We will always the user name available in the view so we can just use that and avoid this join.

amCap1712 (commented):
Thanks for the changes. The PR looks great. Awesome job on the SQL queries especially. Some more nits and after that we should be good to go.

amCap1712 (approved):
LGTM Thanks

PR #2056, opened 2022-07-09
Fix dumps query

amCap1712:
Dumps broke because the user submitted mbids l_artist_credit_mbids and mapped mbids m_artist_credit_mbids had different data types. The first was returned as text whereas the second was returned as a text[]. Align the data type to fix the error.


mayhem (approved):
I had to stare at this one for a while before I got it, lol.

PR #2058, opened 2022-07-10
Limit top stats in Spark SQL query itself

amCap1712:
Instead of collecting all stats and then removing extra stats in post-processing, modify the SQL query to only collect the desired number of stats at max.


mayhem (approved):
Great!

PR #2059, opened 2022-07-11
[package.json] Major dependencies update - LB towards React 18

chinmaykunkikar:
### Intro
My strategy in this PR is to upgrade all dependencies to their latest major versions while upgrading React to `react 17.0.2`.
This is because of two reasons -
1. Changes that need to be done to the source to make it work with React 18.
2. Some packages may be yet to support React 18.

While many libraries were upgraded to their latest versions smoothly with some minor API changes, it's a pain to get a few libraries working properly. There are also couple libraries that are either completely incompatible with newer dependencies, or have deprecated a beloved feature/API that have subpar alternatives, which can stop us from upgrading few other important libraries if we decide to keep using the existing versions.

---
#### 1. [`react-coverflow`](https://github.com/andyyou/react-coverflow/blob/master/README.md#looking-for-someone-to-take-over-the-project-as-ive-recently-changes-jobs-and-had-a-child-and-am-no-longer-capable-of-supporting-this-dependency-if-anyone-is-interested-please-feel-free-to-take-ownership-of-it) has put up a deprecation notice since 2019.
- `react-coverflow` is an unmaintained library and ends it's support at [React 16.x](https://github.com/andyyou/react-coverflow/blob/master/package.json#L48).
- Alternatives like `SwiperJS` are available but are tricky to work with in React, especially since it has poor support for pure ES modules. SwiperJS is also a heavy library and increases the overall bundle size.
- Switching back to `react-multi-carousel` is looking like a good idea because it is maintained actively.
- But having a coverflow in the first place insted of carousel was the idea behind ditching react-multi-carousel.

_Maybe we can think of having a home-grown solution for the coverflow?_

#### 2. `@nivo/*` packages
The visualization library that we use for our charts and stats has gone through major changes in their codebase from `0.62.0` to `0.70.x`.
- Our current version (`0.62.0`) doesn't support react 16+. Upgrading to `^0.70.0` is mandatory for react 17.
- They have not included enough examples (if any) to show implementation of some of the APIs that we use. There is no list of  _"Alternatives to the deprecated APIs"_ or similar.
- No examples with TypeScript.

_Having a hard time fixing issues related to these libraries._

#### 3. The `fork-ts-checker-webpack-plugin`.
- I'm torn. They [retired the `eslint` option](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/601) from the plugin starting [form `7.0.0`](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/pull/607). I think it's one of the most useful and loved feature of the LB build system.
- Looking at the announcement thread, the only alternative to this feature is the [`eslint-webpack-plugin`](https://www.npmjs.com/package/eslint-webpack-plugin), which is [reported significantly slower](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/601#issuecomment-1024849682) to fork-ts-checker-webpack-plugin.
- You may think _"Simple. Let's keep using this version without upgrading."_ But doing so will create another problem - the last supported `eslint` version for this plugin is 6.8. We'll have to keep using `eslint ^6.8` and `prettier` along with most of their plugins/configs and on the same version forever.
- This creates a dilemma: If we go through with the upgrade, we'll have to stick with `eslint-webpack-plugin` and give up some performance until they hopefully address those performance issues. Or we stick to this version and give up upgrading `eslint` and `prettier` along with their plugins and configs and hope that it wont cause troubles with new TypeScript versions.

_speaking of..._

#### 4. Upgrade TypeScript `3.1` --> `4.7`
- Upgrading TypeScript is a major change which has it's own challenges. Like optional operands for `delete` operator, and `catch` clause changes.
- Fix the optional operands for `delete` operator here - https://github.com/metabrainz/listenbrainz-server/commit/fc0c8b8f4d4b79b2443d731b0c2edf024142b909
- Preserve the old behavior for `catch` by adding `"useUnknownInCatchVariables": false` to tsconfig.json. More information on this - [**TS 4.0 announcement blog post**](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#unknown-on-catch) and [**TS Breaking Changes wiki**](https://github.com/microsoft/TypeScript/wiki/Breaking-Changes#using-unknown-in-catch-variables).

---

### TODO
1. Solution for coverflow/carousal.
2. Solution for `ForkTsCheckerPlugin`.
3. Fix `nivo` library errors while keeping same functionality.
4. Upgrade to React 18.2.

_**P.S. Tests won't pass at this point.**_


MonkeyDo:
Thanks for opening this! Definitely a bigger job than I expected!

First of all, I think we will probably want to break the job into multiple PRs, each merged into a special feature branch so that we can break things and fix them in the next PR without disturbing the master branch.
This will make it easier to work on, to review, and to make sure we don't break something else along the way since we're touching so many different parts of the codebase.

1. coverflow:
Time to pay our technical debt! The react-coverflow library outdated, time to replace it.
I think SwiperJS will be our best bet, even with the caveats you cited.
I can help solve issues with implementation, and I'm not sure the size of the library matters all that much: if we don't do tree shaking right to only compile the imported components, then that's its own issue.
Found this which could be of interest?: https://github.com/coderboy-raiyan/Swiper-js-guidline

2. Nivo
Yeah, I regularly find the documentation lacking as well.
But we should aim to keep the library up to date, and I can help you figure out the api changes, I've got some experience reverse-engineering their components.

3. fork-ts-checker-webpack-plugin
Ah, that's unfortunate! Well, let's update it and implement the suggested eslint-webpack-plugin, and wait for them to fix/improve the build times (they are aware of the issues with multithreading).

4. Upgrade TypeScript

MonkeyDo:
To be clearer, I propose we divide the work between PRs like so:

1. Update typescript (least amount of dependencies?)  — mostly done?
2. Update dependencies by group - a PR for each group
    a. All Nivo/D3 dependencies in one PR
    b. Other dependencies that just need a package update but no other changes
    c. Replace the coverflow
3. Changes to the build/compilation/test setup — everything webpack, babel, eslint, jest, less etc.

I know that seems like a lot of PRs, but as I touched on it will help us make sure nothing slides by, and the conversations will be a bit easier to have on a single subject at a time.

Does that sound like a good plan to you? If so I'll create a feature branch that will be our merge target.

chinmaykunkikar:
@MonkeyDo That sounds great! I was struggling to keep up with the changes myself. It'd have been a task to review them.
Yes please create a feature branch for this :)

1. TypeScript PR
Yes, most of the work is done and is least dependent on other packages.

3. Coverflow PR
I'm not sure if the repo will be much of a use. It has example for Swiper v6.x. 6.x works fine, things start breaking from 7.x 😛
But that's alright. We'll get it working :)

4. Nivo PR
Awesome. I'll take your help with that then :)
Note - To upgrade nivo packages to `0.79`, it'll need to be on react 17 at least. So I'm thinking this PR will update all deps that won't work without react 17.

5. Webpack/ESLint/Prettier PR
This will have the fork-ts-checker-webpack-plugin, eslint, etc.

6. Remaining deps PR
Update rest of the deps here.

chinmaykunkikar:
Thanks a lot for the work!

MonkeyDo:
> Thanks a lot for the work!

I'm still looking into alternative for the coverflow library but I think after that I should be able to update to React 18 (which was my original goal, I just want to use the new React Router versions 😭 )

MonkeyDo:
And thank you for doing most of the research @chinmaykunkikar, that made my job much easier

chinmaykunkikar:
Hahaha I know the pain. There's virtually NO coverflow libraries. Was quite surprised.
We'll have to write our own coverflow component or bring back the carousel 🥶

I also noticed that the [time-ago](https://www.npmjs.com/package/time-ago) library isn't maintained anymore. I was thinking if we should use something like [react-time-ago](https://www.npmjs.com/package/react-time-ago) or [javascript-time-ago](https://www.npmjs.com/package/javascript-time-ago) as they are more maintained.

PS Excited to see react router in action on LB. It's going to make things so much easier 😄

PPS Thank you so much!

MonkeyDo:
I think I'll give Swiper.js a go. IMO we can have some nice coverflow-ish effect.
If we don't need it to look exactly like the current coverflow (no real reason it needs to look exactly that way) this looks to me like the best solution, as swiper is well maintained.

In particular these two plugin effects would look nice I think:
https://swiperjs.com/demos#effect-coverflow and  https://swiperjs.com/demos#effect-cards

chinmaykunkikar:
That's right, Swiper is really good. We can definitely come up with something as per our needs.

PR #2061, opened 2022-07-12
Store stats in couchdb

amCap1712:
We refresh the entire stats data daily, storing it all in PG for 24 hours seems wasteful. To improve the situation, adding a nosql datastore - couchdb. Store each day's stat in its own database with the date as the suffix. Once the day's database for a stat is ready, delete older databases for that stat.


listenbrainz/db/couchdb.py
```diff
@@ -0,0 +1,104 @@
...
+    databases_url = f"{get_base_url()}/{database}"
+    response = requests.put(databases_url)
+    response.raise_for_status()
+
+
+def list_databases(prefix: str) -> list[str]:
+    """ List all couchdb database whose name starts with the given prefix
+    sorted in the descending order of creation.
+
+    This method is useful because we name
```
mayhem, line 38: Something odd happened here, I think....

listenbrainz/db/couchdb.py
```diff
@@ -0,0 +1,104 @@
...
+
+def insert_data(database: str, data):
+    """ Insert the given data into the specified database. """
+    with start_span(op="serializing", description="serialize data to json"):
+        docs = ujson.dumps({"docs": data})
+
+    with start_span(op="http", description="insert docs in couchdb using api"):
+        couchdb_url = f"{get_base_url()}/{database}/_bulk_docs"
+        response = requests.post(couchdb_url, data=docs, headers={"Content-Type": "application/json"})
+        response.raise_for_status()
```
mayhem, line 104: I wonder if we should be more intelligent about inserting the data here. If couchdb is unavailable, this will result in one sentry error message per user, which may not be all that good.

listenbrainz_spark/stats/user/entity.py
```diff
@@ -106,10 +106,19 @@ def create_messages(data, entity: str, stats_range: str, from_date: datetime, to
...
             handle this message. can be "user_entity" or "year_in_music_top_stats"
+        database: the name of the database in which the webserver should store the data
 
     Returns:
         messages: A list of messages to be sent via RabbitMQ
     """
+    if database is None:
+        database = f"{entity}_{stats_range}"
+
+    yield {
```
mayhem, line 117: If the couch DB creation fails, then we'll have a ton of failed insertion messages and presumably loads of sentry messages. I think we need to be more careful about the possible failures to create DB.

mayhem (changes requested):
Looking quite good, but I think better error handling is needed.

listenbrainz/db/couchdb.py
amCap1712: The issue is that we don't persist state between messages so its either 1 error per user or none at all. However, sentry is often good at clustering messages. For instance:

image

So, it does have 22k error reports but those are grouped together so we can view it together as one.

listenbrainz_spark/stats/user/entity.py
amCap1712: Agree that we should be more careful for failures to create db. Regarding number of sentry messages, see other comment.

amCap1712:
Need to update dumps as well before merging this PR.

listenbrainz/db/stats.py
```diff
@@ -40,13 +42,53 @@
...
+from listenbrainz.db import couchdb
+
 
 # sitewide statistics are stored in the user statistics table
 # as statistics for a special user with the following user_id.
 # Note: this is the id from LB's "user" table and *not musicbrainz_row_id*.
 SITEWIDE_STATS_USER_ID = 15753
 
 
+def insert_stats_in_couchdb(database, from_ts, to_ts, values):
```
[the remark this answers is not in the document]
amCap1712: Fixed.

amCap1712:
Dumps are now working but tests for dumps and probably some more testing for couchdb is pending.

amCap1712:
Added dump tests as well.

listenbrainz/db/couchdb.py
```diff
@@ -0,0 +1,249 @@
...
+from sentry_sdk import start_span
+
+from urllib3 import Retry
+
+# stat type followed by a underscore followed by a date in YYYYMMDD format
+DATABASE_NAME_PATTERN = re.compile(r"([a-zA-Z]+)_(\w+)_(\d{8})")
+
+DATABASE_LOCK_FILE = "LOCK"
+
+_user = None
```
mayhem, line 17: When I was working with this code for the spotify cache, I was wondering why you chose not to make this module a class...

listenbrainz/db/couchdb.py
```diff
@@ -0,0 +1,249 @@
...
+
+def lock_database(database: str):
+    """ 'Lock' the database so that it does not get deleted.
+
+        Note that, this is not a couchdb feature but a way we made up to co-ordinate process in LB.
+        The onus is the on other users to check for existence of the LOCK file before deleting a
+        database.
+    """
+    document_url = f"{get_base_url()}/{database}/{DATABASE_LOCK_FILE}"
+    # TODO: figure out why PUT works but POST fails with a weird referer header error
```
mayhem, line 175: Still to do or should this be removed?

listenbrainz/db/couchdb.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    databases = list_databases(prefix)
+    # remove the latest database from the list then delete the databases remaining in the list.
+    databases.pop(0)
+
+    deleted, retained = [], []
+
+    for database in databases:
+        databases_url = f"{get_base_url()}/{database}"
+        response = requests.get(f"{databases_url}/{DATABASE_LOCK_FILE}")
```
mayhem, line 99: I think for clarity we should have a check_database_lock() function, rather than having the lock check embedded here.

listenbrainz/db/couchdb.py
```diff
@@ -0,0 +1,249 @@
...
+def unlock_database(database: str):
+    """ 'Unlock' the database so that it can be cleaned up when needed. """
+    delete_data(database, DATABASE_LOCK_FILE)
+
+
+def _assert_status_hook(r, *args, **kwargs):
+    r.raise_for_status()
+
+
+def _get_requests_session():
```
mayhem, line 189: docstring?

listenbrainz/db/tests/test_dump_manager.py
```diff
@@ -195,15 +132,15 @@ def test_create_full_db(self, mock_notify):
                            '--location', self.tempdir])
         self.assertEqual(len(os.listdir(self.tempdir)), 1)
         dump_name = os.listdir(self.tempdir)[0]
-        mock_notify.assert_called_with(dump_name, 'fullexport')
 
         # make sure that the dump contains a full listens dump, a public and private dump (postgres),
         # a public and private dump (timescale) and a spark dump.
         archive_count = 0
+        print(os.listdir(os.path.join(self.tempdir, dump_name)))
```
mayhem, line 139: intended or leftover?

mayhem (approved):
I've skimmed the whole PR and reviewed changed files since the last review -- I think its good to go now, but some minor cleanup could be good as commented.
Thanks!

listenbrainz/db/couchdb.py
amCap1712: I do not remember distinctly, it may be that I forgot about trying that. How about I try moving this to a class in a follow up PR?

listenbrainz/db/couchdb.py
amCap1712: Still to do

listenbrainz/db/couchdb.py
mayhem: Sounds good.

listenbrainz/db/couchdb.py
amCap1712: added.

listenbrainz/db/couchdb.py
amCap1712: done.

listenbrainz/db/tests/test_dump_manager.py
amCap1712: leftover. removed.

PR #2062, opened 2022-07-12
Fix mbid mapping writer after ujson upgrade :sweat:

amCap1712:
When upgrading ujson, the format of listens sent to mbid mapper changed. This is because at some places we were not explicitly dumping python objects and their fields became the json keys. To fix, update mapper to use new format.


mayhem (approved):
I'm confused by the old code... lol.

PR #2069, opened 2022-07-14
LB-1112: profile link in drop down

Aerozol:
Adding a 'Profile' dropdown under the user menu to match all other MetaBrainz sites (MB, BB)/expected functionality
(yes, this is doubling up with the 'Home' button. But gives us future flexibility there)


MonkeyDo (approved):
Why not ? :)

It does harmonize with other websites, although the other websites aren't quite as much centered on the user's profile.
I guess all the more reason to make it easier to get to.

For completion, MB, BB and AB use "Profile" while CB and MeB websites use "Your Profile".

I think "Profile" is clear enough.

PR #2070, opened 2022-07-15
Timezone support in react

YouzhiL:
# Problem

The current version of timezone-support function is implemented in Flask/html. 

# Solution

This pull request re-implement  timezone-support function in React/js.

# Action
1.  Modify the view function `select_timezone` to support react based UI. No more flask forms needed.
2. Create `user-setting` folder in static/js/src.
3. Implement a drop-down for user to select the local timezones.
4. Add APIService function `resetUserTimezone` to fetch response from backend API.
5. Add backend API `reset_timezone` to reset timezone.


listenbrainz/webserver/templates/profile/selecttimezone.html
```diff
@@ -1,35 +1,14 @@
...
 
-{% block content %}
-  <h3 class="page-title">Select Timezone</h3>
-  <p>
-    Your current timezone setting is <b>{{ user_timezone }}</b>. 
-    <br>By choosing your local time zone, 
-    you will have a local timestamps part of your submitted listens. 
-    This also informs as when to generate daily playlists and other recommendations for you.
-  </p>
+{% block profile_content %}
```
amCap1712, line 5: No need for the profile_content block because base-react defines one already. This one is only needed in user/* template files because template structure is different there.

mayhem:
Hi!

I've run this, but when I select a new timezone, it doesn't save the change. Should it be doing that now or am I jumping the gun?

listenbrainz/webserver/views/user_setting_api.py
```diff
@@ -0,0 +1,36 @@
...
+from listenbrainz.webserver.errors import APIInternalServerError,APIBadRequest
+
+
+user_setting_api_bp = Blueprint('user_setting_api_v1', __name__)
+
+
+@user_setting_api_bp.route('/<user_name>/timezone', methods=["POST", "OPTIONS"])
+@crossdomain
+@ratelimit()
+def reset_timezone(user_name):
```
amCap1712, line 19: For normal views, `@login_required` performs the authentication but for API it has to be done manually by reading the Authorization header of the request. We have a helper method for this:

https://github.com/metabrainz/listenbrainz-server/blob/f20cedc430b29600674ce2d83badb42dc3db9f53/listenbrainz/webserver/views/api_tools.py#L459

Example usage is simple, just call it:
https://github.com/metabrainz/listenbrainz-server/blob/f20cedc430b29600674ce2d83badb42dc3db9f53/listenbrainz/webserver/views/pinned_recording_api.py#L48

The helper will also return the user who possesses the token in Authorization header.

But now if you have username in the url as well, then you have to verify that the token matches that user. So, I would suggest to get rid of the username from the url. `/1/user-setting/timezone` is probably enough.

YouzhiL: Thanks, Lucifer! I made the change in my new commit. Also, the user's timezone will be saved and updated on the page after clicking the `save timezone` button. Please take a look at it.

listenbrainz/webserver/views/user_setting_api.py
```diff
@@ -0,0 +1,48 @@
...
+  :resheader Content-Type: *application/json*
+  """
+  user = validate_auth_header()
+  try:
+      data = ujson.loads(request.get_data())
+  except (ValueError, KeyError) as e:
+      raise APIBadRequest(f"Invalid JSON: {str(e)}")
+
+  if 'zonename' not in data:
+      raise APIBadRequest("JSON document must contain zonename", data)
```
mayhem, line 41: Let's update this to say "JSON document must contain timezone name."

mayhem (approved):
The python bits look fine to me and the code runs as expected. Thanks!

amCap1712 (commented):
Hi! The react bits look good to me. Some other comments:
1) The .DS_Store file should be removed from the commit.
2) Tests should be added for frontend component and settings api,
3) In the url, change user-setting/ to settings/

listenbrainz/webserver/views/user_setting_api.py
amCap1712: Thanks

listenbrainz/webserver/static/js/tests/user-settings/SelectTimezone.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <GlobalAppContext.Provider value={globalProps}>
+        <SelectTimezone {...extraProps} />
+      </GlobalAppContext.Provider>
+    );
+    expect(wrapper.html()).toMatchSnapshot();
+  });
+});
+
+describe("resetTimezone", () => {
+  it("calls API, and sets state + creates a new alert on success", async () => {
```
amCap1712, line 42: The description mentions it but this test doesn't have a assert that a alert is created. You can add it like this:

https://github.com/metabrainz/listenbrainz-server/blob/f2096ec1b80302a02fae65b551934c0e69cabf0a/listenbrainz/webserver/static/js/tests/pins/PinnedRecordingCard.test.tsx#L85-L99

amCap1712 (commented):
Thanks for adding the tests. I think we should add 2 more tests but other than that the PR is ready.

1. A test for APIService.resetUserTimezone
2. A test for SelectTimezone component that error is displayed when call to resetTimezone fails.

amCap1712:
Thanks for adding the new tests. Those look good to me. But the unresolved comment above to add an assert for success alert still needs to fixed.

amCap1712 (approved):
LGTM, Thanks!

PR #2071, opened 2022-07-15
Improve Listening Now page

MonkeyDo:
A few improvements rolled up into this PR:

1. As [previously discussed on IRC](https://chatlogs.metabrainz.org/libera/metabrainz/2022-05-19/?msg=5008384&page=2) we adopted the name "Listening Now" instead of "Playing Now". The route has also changed accordingly to `/listening-now`
2. Add release group tags to the album section
3. Fetch metadata for the listen on startup (previously metadata was not fetched and we had to wait to receive another listen)
4. Tweak and move the "no listen"  message:
image
5. Disable "add tag" button that is not implemented. Future PR coming with tag improvements and allowing users to tag directly from in here
6. In the ListenCard component, for playing-now listens make the "listening now" text link to the /listening-now page to increase discovery. We should think about how this page will be accessible; in which menu does it fit?

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -264,7 +264,7 @@ def _register_blueprints(app):
     _register_blueprint_with_context(app, player_bp, url_prefix='/player')
 
     from listenbrainz.webserver.views.metadata_viewer import metadata_viewer_bp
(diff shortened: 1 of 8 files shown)


amCap1712 (approved):
LGTM, Thanks! Let's test on test.lb or beta.lb once before merging. Regarding placement, this can probably go in Explore section for now.

PR #2072, opened 2022-07-19
Listening now: allow users to vote on and add tags

MonkeyDo:
Currently on the Listening Now page we show tags, but users can't interact with them directly.

In this PR we add the UI and mechanism for users to vote on tags as well as submit new tags.
We will have to wait for LB-1120 to be implemented in order to have an MB auth token to send along with the XML POST request.

image

Close up of the "add tag" component once clicked:
image


MonkeyDo:
Hello @alastair !

I wanted to check in with you regarding the work you did on MB about auth tokens, in order to pass on the rest of the implementation work (LB-1120) to lucifer.
Where was your stopping point on this project, and how can I help with passing the project along to the relevant people?

MonkeyDo:
Closing in favor of #2475

PR #2074, opened 2022-07-20
Improve mapping rg type sort

mayhem:
The release group primary and secondary types were not really being sorted (or I just previously got kinda lucky). The primary type accidentally works well to sort on id, but the secondary type, that was not the case.

This PR add a new custom sort order for secondary release types and sort the canonical releases using this new sort and then adds some new tests that where highlighted by Pratamesh's work.

### DIFF
--- a/listenbrainz/mbid_mapping/mapping/custom_sorts.py
+++ b/listenbrainz/mbid_mapping/mapping/custom_sorts.py
@@ -1,6 +1,18 @@
 import psycopg2
 from psycopg2.errors import OperationalError, UndefinedTable
 
+RELEASE_GROUP_SECONDARY_TYPES = [
+    (1, "Compilation"),
+    (2, "Soundtrack"),
+    (6, "Live"),
+    (7, "Remix"),
+    (8, "DJ-mix"),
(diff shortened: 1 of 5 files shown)


listenbrainz/mbid_mapping/mapping/custom_sorts.py
```diff
@@ -123,3 +135,23 @@ def create_formats_table(conn):
...
         raise
+
+    try:
+        with conn.cursor() as curs:
+            curs.execute("DROP TABLE IF EXISTS mapping.release_group_secondary_type_sort")
+            curs.execute("CREATE TABLE mapping.release_group_secondary_type_sort ( secondary_type integer, sort integer )")
+
+            id = 1
+            for type_id, _ in RELEASE_GROUP_SECONDARY_TYPES:
+                curs.execute("INSERT INTO mapping.release_group_secondary_type_sort values (%s, %s);",  tuple((id, type_id)))
```
amCap1712, line 146: Shouldn't this be (type_id, id) ?

amCap1712 (approved):

PR #2076, opened 2022-07-21
Update link to an active, existing rust library

phinjensen:
The repository for listenbrainz-rust is gone from Github, and the crates.io entry hasn't been updated in 3 years: https://crates.io/crates/listenbrainz-rust

Instead, this points to https://crates.io/crates/listenbrainz which was last updated 2 months ago and still has a live repository on Github.

### DIFF
--- a/docs/users/clients.rst
+++ b/docs/users/clients.rst
@@ -13,7 +13,7 @@ Go
 
 Rust
 ^^^^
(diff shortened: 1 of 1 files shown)


amCap1712:
LGTM. Thanks for  the contribution.

PR #2080, opened 2022-07-26
Set metrics when the various spark jobs complete

mayhem:
Currently we have no way of knowing when a spark job fails until one of us notices the data is out of date. This PR adds a metric submission to each of the spark jobs so that we can set grafana alerts for each of them. Since I have no meaningful way to test this PR, this PR IS UNTESTED. 

@amCap1712 I am not sure if I got the stats all correct -- some things are not clear and straightforward there. Please have a look and suggest improvements.

Once this PR is merged, we will need to setup the alerts in grafana to complete this task.

### DIFF
--- a/listenbrainz/spark/handlers.py
+++ b/listenbrainz/spark/handlers.py
@@ -24,6 +24,7 @@ from listenbrainz.spark.troi_bot import run_post_recommendation_troi_bot
 TIME_TO_CONSIDER_STATS_AS_OLD = 20  # minutes
 TIME_TO_CONSIDER_RECOMMENDATIONS_AS_OLD = 7  # days
 
+
 def is_new_user_stats_batch():
     """ Returns True if this batch of user stats is new, False otherwise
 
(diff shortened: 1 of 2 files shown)


listenbrainz/spark/handlers.py
```diff
@@ -48,6 +50,7 @@ def notify_user_stats_update(stat_type):
             from_name='ListenBrainz',
             from_addr='noreply@'+current_app.config['MAIL_FROM_DOMAIN']
         )
+        metrics.set("listenbrainz-spark-user-stats", 1)
```
amCap1712, line 53: This function is unused. I have cleaned it up in the couchdb PR, so should remove this metrics.set.

listenbrainz/spark/spark_reader.py
```diff
@@ -58,6 +59,8 @@
 
 RABBITMQ_HEARTBEAT_TIME = 60 * 60  # 1 hour, in seconds
 
+metrics.init("listenbrainz")
```
amCap1712, line 62: metrics are init in create_app which happens in __init___ so can remove this.

PR #2081, opened 2022-07-26
Move artist map stats to redis cache

amCap1712:
We generate artist map stats on demand and cache them in PG. Soon, we'll be using CouchDB for stats which is not a natural fit for caching. So instead use Redis for caching.


mayhem (approved):
Looks good, but lets not forget about invalidating the existing cache items when new data has been generated/inserted.

amCap1712:
Storing in couchdb only because we don't want stale artist maps and keeping in sync with artist stats is easier there.

PR #2082, opened 2022-07-26
Update clients.rst

rain0r:
# Problem

There's no linkt to a Java client.

# Action

Added `listenbrainz-java` to the list of clients.


amCap1712 (approved):
LGTM, Thanks!

PR #2083, opened 2022-07-27
fixed path to docker-compose.labs.api.yml in start-labs-api.sh

Prathamesh-Ghatole:
# Problem
start-labs-api.sh was calling docker-compose-labs.api.yml instead of docker-compose.labs.api.yml

# Solution
replaced the "docker-compose-labs.api.yml" with "docker-compose.labs.api.yml" in start-labs-api.sh

# Action
N/A


alastair (approved):
thank you!

PR #2085, opened 2022-07-29
Add global react props to huesound template

alastair:
# Problem

Reported at https://community.metabrainz.org/t/huesound-error/594804, huesound API queries are hitting https://listenbrainz.org/1/whatever instead of https://api.listenbrainz.org
This appears to be due only to a missing global prop containing the api_key

# Solution

Add global props to the flask template

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -285,7 +285,7 @@ def _register_blueprints(app):
     app.register_blueprint(api_bp_compat)
 
     from listenbrainz.webserver.views.explore import explore_bp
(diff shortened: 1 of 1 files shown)


amCap1712 (approved):
LGTM, Thanks!

alastair:
This is wrong, the huesound html already inherits from react-base, which already includes the global props. In that case, I'm not sure why the global props are not showing on the page in prod.

alastair:
@amCap1712 got it this time - we were missing the context when registering the blueprint

amCap1712 (approved):

PR #2087, opened 2022-07-31
Fix typo in ReleaseGroupSecondaryType

amCap1712:
Fresh Releases API is currently erroring because the Audio Drama is named incorrectly due to which validation fails.


mayhem (approved):
Gah what drama over such a small thing. lol

PR #2090, opened 2022-08-01
Fix huesound endpoint url

amCap1712:
#2085 added global props to the page but huesound still doesn't work because the endpoint url was invalid. Fix the url. Tested on beta works now.


alastair (approved):
was just checking this and saw it wasn't working yet. thanks!

PR #2092, opened 2022-08-03
Spotify metadata cache (WIP)

mayhem:
This PR is very much a work in progress, but I could use some head-checks to see if anyone sees anything badly wrong with my approach for building a spotify metadata cache.

The idea is that to warm up this giant cache, we're going to seed all the known spotify artist ids as a set of "pending ids" to be fetched. As these artists are fetched and stored in couchdb, we discover new artist ids and then add them to the list of pending ids. Once we've collected 1000 pending ids, these ids are written to a special document in couchdb. 

This program, which will run indefinitely, has a queue of what artists to look up next. If that runs dry, the next pending ids document is fetched from couchdb, inserted into the internal queue and then deleted from the db. The main loop of the program will continually fetch artists (by fetching all fo their albums and their tracks) and then write the results to couchdb. 

In order to keep the pending_ids from exploding, there is a recent_ids dict that tracks ids that have either been recently added or recently added to the pending list, so that we can forget about it for the time being. RIght now the recently added dict has no pruning in place, so this is something to keep an eye on.

Actually using this data is beyond the scope of this PR. We have several months before we've collected enough data.

TODO: Clean up couchdb additions, docstrings, add more gutt to README.

### DIFF
--- a/listenbrainz/mbid_mapping/mapping/spotify_cache.py
+++ b/listenbrainz/mbid_mapping/mapping/spotify_cache.py
@@ -91,7 +91,12 @@ class SpotifyMetadataCache():
         return artist
 
     def fetch_artist_ids_from_track_id(self, track_id):
-        return [a["id"] for a in self.sp.track(track_id)["artists"]]
+        try:
+            track = self.sp.track(track_id)
+        except spotipy.exceptions.SpotifyException:
+            return None
+
+        return [a["id"] for a in track["artists"]]
 
     def insert_artist(self, artist):
         artist["_id"] = artist["id"]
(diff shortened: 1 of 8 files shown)


listenbrainz/mbid_mapping/mapping/spotify_cache.py
```diff
@@ -0,0 +1,244 @@
...
+        if spotify_id not in self.recent_ids:
+            self.recent_ids[spotify_id] = datetime.now()
+            return True
+
+        return False
+
+    def add_pending_spotify_ids(self, spotify_ids):
+        try:
+            iter(spotify_ids)
+        except TypeException:
```
amCap1712, line 122: except TypeError

listenbrainz/mbid_mapping/mapping/spotify_cache.py
```diff
@@ -0,0 +1,244 @@
...
+        self.id_queue = UniqueQueue()
+        self.pending_ids = []
+        self.recent_ids = {}
+        self.sp = spotipy.Spotify(auth_manager=SpotifyClientCredentials(client_id=config.SPOTIFY_APP_CLIENT_ID,
+                                                                        client_secret=config.SPOTIFY_APP_CLIENT_SECRET))
+        self.couch = couchdb.init(config.COUCHDB_USER, config.COUCHDB_PASSWORD, config.COUCHDB_HOST, config.COUCHDB_PORT)
+        dbs = couchdb.list_databases("")
+        if self.COUCHDB_NAME not in dbs:
+            couchdb.create_database(self.COUCHDB_NAME)
+        couchdb.create_index(self.COUCHDB_NAME, {
```
amCap1712, line 57: This should probably be inside the if as well.

PR #2093, opened 2022-08-11
Cleanup unused files

amCap1712:
See individual commit messages for more details and rationale.


consul_config.py.ctmpl
```diff
@@ -223,12 +223,8 @@ CRITIQUEBRAINZ_CLIENT_ID = '''{{template "KEY" "critiquebrainz/client_id"}}'''
 CRITIQUEBRAINZ_CLIENT_SECRET = '''{{template "KEY" "critiquebrainz/client_secret"}}'''
 CRITIQUEBRAINZ_REDIRECT_URI = '''{{template "KEY" "critiquebrainz/redirect_uri"}}'''
 
-
 # YOUTUBE
 YOUTUBE_API_KEY = '''{{template "KEY" "youtube/api_key"}}'''
-YOUTUBE_REDIRECT_URI = '''{{template "KEY" "youtube/redirect_uri"}}'''
```
alastair, line 229: will this be removed from serer-configs too?

alastair (approved):
looks fine to me, thanks for cleaning up

consul_config.py.ctmpl
amCap1712: Yes, after a few releases though. So that in case we have to roll back to a release whose config file still requires this keys, it doesn't block startup.

PR #2094, opened 2022-08-11
[Do not merge] Speed up LB tests

amCap1712:
This is a catch all changes PR to benchmark speeding up LB tests. It touches lots of parts of the codebase, some with partial or no test coverage. As such its not possible to review it well, the intent is to keep this PR as a record of all the changes that need to be made and open smaller PRs and improve speed incrementally.

### DIFF
--- a/listenbrainz/db/__init__.py
+++ b/listenbrainz/db/__init__.py
@@ -2,7 +2,7 @@ from typing import Optional
 
 import sqlalchemy
 from sqlalchemy import create_engine
-from sqlalchemy.pool import NullPool, AssertionPool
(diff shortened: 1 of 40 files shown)


MonkeyDo:
As an aside, I recently removed the `action-docker-layer-caching` steps in the BB repo github actions after seeing reports of improved speed, and indeed my builds are now a lot faster without the caching steps.
Maybe something to remove from LB too?

amCap1712:
Oh, I see. The caching does take a lot of time so makes sense to remove that if the actual build time is less. I'll do some quick tests and will remove if its slower. Thanks!

PR #2095, opened 2022-08-13
Only allow pinning if playing now listen has msid or mbid

amCap1712:
Add a guard to only show the pin recording control if the listen has a msid or mbid. Normal listens always have msid but playing now listens never do. And metadata lookup may not always be able to find a match for the playing now listen so need the check.

### DIFF
--- a/listenbrainz/webserver/static/js/src/user/Listens.tsx
+++ b/listenbrainz/webserver/static/js/src/user/Listens.tsx
@@ -691,17 +691,20 @@ export default class Listens extends React.Component<
       artistMBIDs?.length ||
       Boolean(trackMBID) ||
       Boolean(releaseGroupMBID);
+    const canPin = Boolean(recordingMBID) || Boolean(recordingMSID);
 
     /* eslint-disable react/jsx-no-bind */
     const additionalMenuItems = (
       <>
-        <ListenControl
-          text="Pin this recording"
-          icon={faThumbtack}
-          action={this.updateRecordingToPin.bind(this, listen)}
-          dataToggle="modal"
-          dataTarget="#PinRecordingModal"
-        />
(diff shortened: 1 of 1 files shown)


MonkeyDo (commented):
Yes, thanks for fixing this. I'm sure we had that check at some point, but it was removed as we thought we would always have one id or the other.

You might want to wait until #2096 is merged and apply the same method to conditionally add the pin option.

A more general open question: I'd love to move this pin button to the ListenCard component so that it is always available (with the caveat of `canPin` the way you did) ?
Currently we add it separately on some pages (and so it's missing in other places).

The issue is that we only have access to the `PinRecordingModal` from the root component.
Not sure what the best architecture for dealing with this issue is, but I'm sensing something combining a higher-order component and a React Context (the first would give us a wrapper for the base page and the second a method we can call from the listencard).

amCap1712:
#2096 is based on this PR so I already modified the method for pinning there :). At a quick look and IIUC, I see that `onSuccessfulPin` is the only piece that is needed in the root component. So maybe we could that as an optional prop to ListenCard and then use PinnedRecordingModal from inside ListenCard directly.

MonkeyDo:
> #2096 is based on this PR so I already modified the method for pinning there :). At a quick look and IIUC, I see that `onSuccessfulPin` is the only piece that is needed in the root component. So maybe we could that as an optional prop to ListenCard and then use PinnedRecordingModal from inside ListenCard directly.

The problem with that is that we'll be creating a PinnedRecordingModal component for each ListenCard, instead of the current single component. That's why I mentioned a HOC as a good solution to provide the PinnedRecordingModal component to the entire page.

amCap1712:
Ah! Makes sense. HOC sounds good indeed then.

PR #2096, opened 2022-08-13
Hide actions menu if empty

amCap1712:
Empty Fragment <></> also evaluates to true, so we end up showing an empty menu if an empty fragment is passed as additionalMenuItems. Instead, accept an array of action item elements and only render if array is not empty.

### DIFF
--- a/listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
+++ b/listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
@@ -245,7 +245,7 @@ export default class ListenCard extends React.Component<
 
     // Hide the actions menu if in compact mode or no buttons to be shown
     const hasActionOptions =
-      (additionalMenuItems && additionalMenuItems.length) ||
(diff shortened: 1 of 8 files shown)


listenbrainz/webserver/static/js/src/listens/ListenCard.tsx
```diff
@@ -245,7 +245,7 @@ export default class ListenCard extends React.Component<
 
     // Hide the actions menu if in compact mode or no buttons to be shown
     const hasActionOptions =
-      additionalMenuItems ||
+      (additionalMenuItems && additionalMenuItems.length) ||
```
MonkeyDo, line 248: A little more succinct, this should work nicely.
The `?.` syntax allows us to check if additionalMenuItems exists before accessing the length, and will evaluate to `undefined` (coerced to `false` by the OR operator) if it doesn't exists.

```suggestion
      additionalMenuItems?.length ||
```

MonkeyDo (commented):
Changes look good

PR #2097, opened 2022-08-14
Upgrade Spark Cluster

amCap1712:
All of these upgrades are already in place on production, updating here for testing and documentation purposes.

1. Upgrade Spark to 3.3.0, Hadoop to 3.3.4, Java to 17 (only for spark, hadoop still runs on Java 11)
2. Replace YARN with Spark Standalone cluster
3. Add systemd service file for request consumer


docker/spark-cluster-config/michael/request_consumer.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=The ListenBrainz Request Consumer for Spark
+
+[Service]
+WorkingDirectory=/home/request_consumer/listenbrainz-server/
+ExecStart=/home/request_consumer/listenbrainz-server/docker/start-spark-request-consumer.sh
+User=root
```
amCap1712, line 7: I tried using the request_consumer user but ran into troubles with file permissions errors in /data/tmp directories and HDFS. So leaving it as root for now. Also, request_consumer user doesn't exist on workers nodes so will need to figure out how/if it'll work,

mayhem: Agreed, but this is a good start. Maybe leave a TODO comment here?

mayhem (approved):
easy enough!

docker/spark-cluster-config/michael/request_consumer.service
amCap1712: Done.

PR #2101, opened 2022-08-17
Add new fields to mapping.mb_metadata_cache

amCap1712:
1. We already had length but forgot to use it in process_row, need it for session based mbid similarity.
2. Need recording_name, artist_credit_name, release_name to replace usages of mbid_mapping_metadata in future.


mayhem (approved):
This slightly changes the format of the artist data, no? This means that this PR needs to be deployed once the data consumers (e.g. listenging now page) is updated as well.

amCap1712:
Makes sense, updated.

mayhem (approved):

PR #2105, opened 2022-08-17
this needs to be 80, otherwise development server will run on wrong port

treeshateorcs:
sorry if i've done something wrong. this is my first (well, second, first one was unsuccessful) pull request in this repo. i'm just starting out with the codebase. if you have any tips, please tell me

# Problem
server starts in docker on the wrong port, so when you point your browser to http://localhost it shows that it can't connect to server

# Solution
change port from 8100 to 80

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -49,7 +49,7 @@ services:
       - ..:/code/listenbrainz:z
       - ../listenbrainz/webserver/static:/static
     ports:
-      - "8100:80"
+      - "80:80"
     depends_on:
       - redis
       - lb_db


amCap1712:
Hi @treeshateorcs, It is intentional to run on port 8100 to avoid conflicts with other services running on the user's system. To access the website locally, you should try opening `http://localhost:8100`.

treeshateorcs:
hello @amCap1712 ! thank you for your review. in that case, documentation needs to be updated, here

https://listenbrainz.readthedocs.io/en/production/dev/devel-env/#run-the-magic-script

> By default, the web service listens on port 80.

PR #2106, opened 2022-08-19
Update labs api consul/uwsgi to match other flask apps

amCap1712:
Every some days, labs api container is plagued with this error:

probably another instance of uWSGI is running on the same address (0.0.0.0:3031)
    bind(): Address already in use

When looking at the consul config, labs api's config is different from that of web and api compat. My hunch is when consul template restarts the process say due to the restart of a service dependency, the old uwsgi process is not shut down properly. Updating the labs api config to match other flask apps to hopefully fix the issue.

### DIFF
--- a/docker/services/labs_api/consul-template-labs-api.conf
+++ b/docker/services/labs_api/consul-template-labs-api.conf
@@ -4,9 +4,9 @@ template {
 }
 
 exec {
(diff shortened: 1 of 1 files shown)


alastair (approved):
great catch, thanks!

PR #2110, opened 2022-08-20
Use kombu in mbid mapping writer

amCap1712:
We have been facing issue with using pika, so we replaced it with kombu sometime ago in websockets container. It has been working fine so now using kombu everywhere in the codebase.


mayhem (approved):
Gawd, I hope this makes our rabbitmq issues go away!

PR #2119, opened 2022-08-24
Remove reliance on implicit autocommit 

amCap1712:
This PR is the first in a series of PRs to prepare the codebase for upgrading to SQLAlchemy 2.0. The 2.0 version isn't out yet but we can use the future flag in version 1.4 to use the newer API. Some of these API changes are useful when migrating to transactional tests.

This PR enables the SQLAlchemy Deprecation warnings in tests and fixes one type of those warnings:

The current statement is being autocommitted using implicit autocommit, which will be removed in SQLAlchemy 2.0. Use the .begin() method of Engine or Connection in order to use an explicit transaction for DML and DDL statements.

For rationale, see section: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#library-level-but-not-driver-level-autocommit-removed-from-both-core-and-orm

### DIFF
--- a/docker/docker-compose.integration.yml
+++ b/docker/docker-compose.integration.yml
@@ -34,6 +34,7 @@ services:
       - ..:/code/listenbrainz:z
     environment:
       PYTHONDONTWRITEBYTECODE: 1
+      SQLALCHEMY_WARN_20: 1
     depends_on:
       - redis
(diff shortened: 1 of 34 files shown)


alastair (approved):
looks sane, thanks

PR #2120, opened 2022-08-24
Access result rows as namedtuples

amCap1712:
Builds on #2119

This PR fixes the warning:

Using non-integer/slice indices on Row is deprecated and will be removed in version 2.0; please use row._mapping[<key>], or the mappings() accessor on the Result object.

For rationale, see section: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#result-rows-act-like-named-tuples


listenbrainz/db/dump.py
```diff
@@ -671,7 +671,7 @@ def get_dump_entries():
               ORDER BY created DESC
             """))
 
-        return [dict(row) for row in result]
```
alastair, line 674: try `result.mappings().all()` or `[dict(row) for row in result.mappings()]`

listenbrainz/db/dump.py
```diff
@@ -684,7 +684,7 @@ def get_dump_entry(dump_id):
             'dump_id': dump_id,
         })
         if result.rowcount > 0:
-            return dict(result.fetchone())
+            return result.fetchone()._asdict()
```
alastair, line 687: try `return result.mappings().first()`

alastair (commented):
We had a discussion in IRC about this, and mentioned that it'd be nicer to use the `result.mappings()` method to return us dictionary-looking items when we need dictionaries, so that we don't have to use `_hidden` style variables. I think that in the cases where we just get a few values and return them we can use this new style (e.g. `return Session(row.id, row.user_id, row.sid, row.api_key, row.ts)` is OK by me)

amCap1712:
Just to clarify, `_asdict()` is named like a private member but isn't one. Its intended for public use but has been named with an _ for "reasons".

listenbrainz/db/dump.py
amCap1712: Done

listenbrainz/db/dump.py
amCap1712: Done

PR #2121, opened 2022-08-24
Only pass text-wrapped string to connection.execute

amCap1712:
Builds on #2120

This PR fixes the warning:

Passing a string to Connection.execute() is deprecated and will be removed in version 2.0.  Use the text() construct, or the
Connection.exec_driver_sql() method to invoke a driver-level SQL string.

For rationale, see section:
https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#execute-method-more-strict-execution-options-are-more-prominent

### DIFF
--- a/listenbrainz/db/__init__.py
+++ b/listenbrainz/db/__init__.py
@@ -1,7 +1,7 @@
 from typing import Optional
 
 import sqlalchemy
-from sqlalchemy import create_engine
+from sqlalchemy import create_engine, text
 from sqlalchemy.pool import NullPool
 import time
 import psycopg2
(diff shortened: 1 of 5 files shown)


alastair (approved):
great, not too many of these to fix.

Does it make sense to try and do something with the 3 duplicated `run_sql_script_without_transaction` methods for ts/msb/lb? I guess this will be going down to 2 in the future with msb going away, but should we abstract this out to take an engine parameter and use the same method in all db init methods?

amCap1712:
Yes, abstracting run_sql_script and run_sql_script_without_transaction makes sense to me.

PR #2122, opened 2022-08-24
Pass parameters to connection.execute() as a dictionary

amCap1712:
Based on #2121

The connection.execute() method in SQLAlchemy 2.0 will accept parameters as a single dictionary or a single sequence of dictionaries only. Parameters passed as keyword arguments, tuples or positionally oriented dictionaries and/or tuples will no longer be accepted.

For rationale, see section:
https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#execute-method-more-strict-execution-options-are-more-prominent


alastair (approved):
wow, I didn't even know that you could pass kwargs! I was sure that we were always using dictionaries, but I guess not... thanks for the updates

PR #2123, opened 2022-08-24
Return event metadata from create_user_notification_event endpoint

amCap1712:
Currently, the notification event id is hardcoded in tests and it works fine because the relevant sequence is reset before each test. However, as a part of upcoming test speed improvements that will no longer be the case. So, we need to retrieve the event id from the api. We already return this metadata for other event types, so it should be fine.

### DIFF
--- a/listenbrainz/webserver/views/user_timeline_event_api.py
+++ b/listenbrainz/webserver/views/user_timeline_event_api.py
@@ -154,11 +154,14 @@ def create_user_notification_event(user_name):
     metadata = NotificationMetadata(creator=creator['musicbrainz_id'], message=message)
 
     try:
-        db_user_timeline_event.create_user_notification_event(user['id'], metadata)
+        event = db_user_timeline_event.create_user_notification_event(user['id'], metadata)
     except DatabaseException:
         raise APIInternalServerError("Something went wrong, please try again.")
 
-    return jsonify({'status': 'ok'})
+    event_data = event.dict()
+    event_data['created'] = event_data['created'].timestamp()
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -154,11 +154,14 @@ def create_user_notification_event(user_name):
...
 
     try:
-        db_user_timeline_event.create_user_notification_event(user['id'], metadata)
+        event = db_user_timeline_event.create_user_notification_event(user['id'], metadata)
     except DatabaseException:
         raise APIInternalServerError("Something went wrong, please try again.")
 
-    return jsonify({'status': 'ok'})
+    event_data = event.dict()
+    event_data['created'] = event_data['created'].timestamp()
```
alastair, line 162: do we use a timestamp elsewhere in this API? (rather than a rfc 3339 date)

amCap1712: The retrieve endpoint accepts timestamps:

https://github.com/metabrainz/listenbrainz-server/blob/331c9a93dd60cfa5174c382699ac69e729d5caa9/listenbrainz/webserver/views/user_timeline_event_api.py#L238-L239

The other create event endpoints also return a timestamp, for one example:

https://github.com/metabrainz/listenbrainz-server/blob/331c9a93dd60cfa5174c382699ac69e729d5caa9/listenbrainz/webserver/views/user_timeline_event_api.py#L221-L226

mayhem (approved):

PR #2125, opened 2022-08-24
Reduce time.sleep 's in tests

amCap1712:
We had a few unnecessary time.sleep 's in tests, removed those to speed up tests. The remaining ones are in listen integration tests and cannot be removed entirely. Those sleeps are there to wait for timescale writer to pick up and insert the listens. Reducing the sleep duration there a bit to poll more frequently.

### DIFF
--- a/listenbrainz/db/tests/test_user_relationship.py
+++ b/listenbrainz/db/tests/test_user_relationship.py
@@ -120,10 +120,10 @@ class UserRelationshipTestCase(DatabaseTestCase):
         db_user_relationship.insert(self.main_user['id'], self.followed_user_1['id'], 'follow')
         db_user_relationship.insert(self.main_user['id'], self.followed_user_2['id'], 'follow')
 
-        time.sleep(3)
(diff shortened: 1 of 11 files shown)


listenbrainz/db/tests/test_user_relationship.py
```diff
@@ -120,11 +120,11 @@ def test_get_follow_events_honors_timestamp_parameters(self):
         db_user_relationship.insert(self.main_user['id'], self.followed_user_1['id'], 'follow')
         db_user_relationship.insert(self.main_user['id'], self.followed_user_2['id'], 'follow')
 
-        time.sleep(3)
+        ts2 = time.time()
```
mayhem, line 123: This is a clever improvement. :)

mayhem (approved):

listenbrainz/db/tests/test_user_relationship.py
amCap1712: :)

PR #2126, opened 2022-08-25
Create flask app once per TestCase

amCap1712:
To improve test speed, initialize the flask app only once per TestCase instead of doing it for each test case. This speeds up unit tests by 30-45s and integration tests by ~60s on wolf.

### DIFF
--- a/listenbrainz/webserver/testing.py
+++ b/listenbrainz/webserver/testing.py
@@ -1,9 +1,22 @@
-import flask_testing
+import unittest
+
+from flask import template_rendered, message_flashed
 
 from listenbrainz.webserver import create_api_compat_app, create_web_app
 
 
-class ServerTestCase(flask_testing.TestCase):
+class ServerTestCase(unittest.TestCase):
+    """ TestCase for Flask App tests, most of the code here has been borrowed from flask_testing.TestCase
+    (https://github.com/jarus/flask-testing/blob/5107691011fa891835c01547e73e991c484fa07f/flask_testing/utils.py#L118).
+    A key difference is that flask_testing's TestCase creates the flask app for each test method whereas
+    our ServerTestCase creates it once per class. flask_testing's test case allows also for custom Response mixins
+    and other stuff to be compatible across flask versions. Since, we don't need that the implementation can be
+    simplified.
+
+    Notably, the implementation of the setUpClass, setUp, tearDown, tearDownClass, assertMessageFlashed,
+    assertTemplateUsed has been adapted to use class level variables. The implementation of assertRedirects has been
+    modified to fix concerning absolute and relative urls in Location header.
+    """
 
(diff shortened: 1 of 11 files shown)


listenbrainz/webserver/testing.py
```diff
@@ -1,11 +1,14 @@
-import flask_testing
+import unittest
+
+from flask import template_rendered, message_flashed
 
 from listenbrainz.webserver import create_api_compat_app, create_web_app
 
 
-class ServerTestCase(flask_testing.TestCase):
+class ServerTestCase(unittest.TestCase):
```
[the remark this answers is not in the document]
alastair: this would also be great because it allows us to add a bit about how we fixed assertRedirects too

listenbrainz/webserver/views/test/test_index.py
```diff
@@ -173,9 +167,9 @@ def test_menu_logged_in_error_dont_show_user_loaded(self, mock_user_get):
 
         mock_user_get.return_value = user
 
-        @self.app.route('/page_that_returns_500')
+        @self.app.route('/page_that_returns_500_2')
```
alastair, line 170: was this for testing, or is it required?

listenbrainz/webserver/views/test/test_player.py
```diff
@@ -1,14 +1,9 @@
...
-from listenbrainz.webserver.testing import ServerTestCase
+from listenbrainz.tests.integration import IntegrationTestCase
 
 
-class PlayerViewsTestCase(ServerTestCase, DatabaseTestCase):
-
-    def setUp(self):
-        ServerTestCase.setUp(self)
-        DatabaseTestCase.setUp(self)
+class PlayerViewsTestCase(IntegrationTestCase):
```
alastair, line 6: I see that there are 3 other subclasses of `IntegrationTestCase` in the `listenbrainz.webserver` package. I would have thought that they'd all be in `listenbrainz.tests.integration` Do you think that we should move these 5-6 test files to the integration package, or leave them for now with the understanding that we're going to be merging them into a single test runner in the short term future anyway?

listenbrainz/webserver/views/test/test_index.py
amCap1712: It is required because the flask app is same for all tests in this class and another method registers a view of the same name.

listenbrainz/webserver/views/test/test_player.py
amCap1712: Leaving as is makes sense. Further, the change to combine integration and unit tests has been merged now.

listenbrainz/webserver/testing.py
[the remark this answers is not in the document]
amCap1712: Done.

PR #2128, opened 2022-08-25
LB-768: Combine integration and unit tests

amCap1712:
Combine docker file setup and test scripts of unit and integration tests


test.sh
```diff
@@ -55,14 +48,8 @@ function invoke_docker_compose_spark {
...
 
-function invoke_docker_compose_int {
-    docker-compose -f $INT_COMPOSE_FILE_LOC \
-                   -p $INT_COMPOSE_PROJECT_NAME \
-                   "$@"
-}
-
 function docker_compose_run {
-    invoke_docker_compose run --rm --user `id -u`:`id -g` "$@"
+    invoke_docker_compose run --rm --user "$(id -u)":"$(id -g)" "$@"
```
alastair, line 52: 🎉

alastair (approved):
This looks fine, thanks.

One thing I learned recently was [compose file profiles](https://docs.docker.com/compose/compose-file/#profiles), which would allow us to do things like "only start up the timescale writer in certain cases". However, this would have to be done in test.sh when calling docker-compose, which would likely introduce extra complexity.
I don't think it's a problem to run the ts writer during unit tests.

PR #2131, opened 2022-08-29
Personal Recommendation Frontend

hrik2001:
# Problem
The backend for personal recommendation has already been merged, but to recommend, we need a frontend, this PR (draft PR for now) attempts to solve that.

# Solution
Creating a modal and choosing followers to whom you want to recommend. UI/UX of this PR has been followed according to the [GSoC proposal](https://community.metabrainz.org/t/gsoc-2022-personal-recommendation-of-a-track/578905) for the same

### DIFF
--- a/listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
+++ b/listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
@@ -29,6 +29,7 @@ import {
 } from "../utils/utils";
 import CBReviewModal from "../cb-review/CBReviewModal";
 import ListenControl from "../listens/ListenControl";
+import PersonalRecommendationModal from "../personal-recommendations/PersonalRecommendations";
 
 export type RecentListensProps = {
   listens: Array<Listen>;
@@ -39,6 +40,7 @@ export interface RecentListensState {
   listenCount?: number;
   recordingToPin?: Listen;
   recordingToReview?: Listen;
+  recordingToPersonallyRecommend: Listen;
   recordingMsidFeedbackMap: RecordingFeedbackMap;
   recordingMbidFeedbackMap: RecordingFeedbackMap;
 }
@@ -56,6 +58,7 @@ export default class RecentListens extends React.Component<
       listens: props.listens || [],
       recordingToPin: props.listens?.[0],
       recordingToReview: props.listens?.[0],
+      recordingToPersonallyRecommend: props.listens?.[0],
       recordingMsidFeedbackMap: {},
       recordingMbidFeedbackMap: {},
     };
@@ -67,6 +70,12 @@ export default class RecentListens extends React.Component<
 
   updateRecordingToPin = (recordingToPin: Listen) => {
     this.setState({ recordingToPin });
+  };
+
+  updateRecordingToPersonallyRecommend = (
+    recordingToPersonallyRecommend: Listen
+  ) => {
+    this.setState({ recordingToPersonallyRecommend });
   };
 
   updateRecordingToReview = (recordingToReview: Listen) => {
@@ -174,7 +183,12 @@ export default class RecentListens extends React.Component<
   };
 
   render() {
-    const { listens, recordingToPin, recordingToReview } = this.state;
+    const {
+      listens,
+      recordingToPin,
+      recordingToReview,
+      recordingToPersonallyRecommend,
+    } = this.state;
     const { newAlert } = this.props;
     const { APIService, currentUser } = this.context;
 
@@ -213,6 +227,16 @@ export default class RecentListens extends React.Component<
                         action={this.updateRecordingToPin.bind(this, listen)}
                         dataToggle="modal"
                         dataTarget="#PinRecordingModal"
+                      />
+                      <ListenControl
+                        text="Personally recommend this recording"
+                        icon={faThumbtack}
+                        action={this.updateRecordingToPersonallyRecommend.bind(
+                          this,
+                          listen
+                        )}
+                        dataToggle="modal"
+                        dataTarget="#PersonalRecommendationModal"
                       />
                       {isListenReviewable && (
                         <ListenControl
@@ -264,6 +288,10 @@ export default class RecentListens extends React.Component<
               <CBReviewModal
                 listen={recordingToReview}
                 isCurrentUser
+                newAlert={newAlert}
+              />
+              <PersonalRecommendationModal
+                recordingToPersonallyRecommend={recordingToPersonallyRecommend}
                 newAlert={newAlert}
               />
             </>
--- a/listenbrainz/webserver/static/js/src/utils/types.d.ts
+++ b/listenbrainz/webserver/static/js/src/utils/types.d.ts
@@ -516,6 +516,12 @@ declare type UserTrackRecommendationMetadata = {
   recording_msid: string;
 };
 
+/** For recommending a track personally from the front-end */
+declare type UserTrackPersonalRecommendationMetadata = UserTrackRecommendationMetadata & {
+  blurb_content: string | null;
+  users: Array<string> | null;
+};
+
 declare type PinEventMetadata = Listen & {
   blurb_content?: string;
 };
--- a/listenbrainz/webserver/static/css/main.less
+++ b/listenbrainz/webserver/static/css/main.less
@@ -15,6 +15,7 @@
 @import "messybrainz.less";
 @import "metadata-viewer.less";
 @import "musicbrainz-entity-icons.less";
+@import "personalrecommendationmodal.less";
 
 @icon-font-path: "/static/fonts/";
 
--- a/listenbrainz/webserver/static/css/personalrecommendationmodal.less
+++ b/listenbrainz/webserver/static/css/personalrecommendationmodal.less
@@ -1,0 +1,41 @@
+@white-background: #fff;
+
+#PersonalRecommendationModal {
+  .searchdropdown {
+    &.open {
+      display: block;
+    }
+    display: none;
+    position: absolute;
+    width: calc(100% - 40px);
+    z-index: 6;
+    box-shadow: 0px 6px 12px rgba(0, 0, 0, 0.175);
+    list-style: none;
+    border: 1px solid #ccc; // IE8 fallback
+    border: 1px solid rgba(0, 0, 0, 0.15);
+    background-color: @white-background;
+    max-height: 102px;
+    /*overflow: hidden;*/
+    overflow-y: scroll;
+
+    // Links within the dropdown menu
+    > a,
+    > button {
+      display: block; // necessary for <a> tags
+      width: 100%;
+      background: none;
+      color: inherit;
+      border: none;
+      padding: 5px 20px;
+      font: inherit;
+      cursor: pointer;
+      outline: inherit;
+      text-align: left;
+
+      &:hover {
+        color: @white-background;
+        background-color: #eb743b;
+      }
+    }
+  }
+}
--- a/listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendations.tsx
+++ b/listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendations.tsx
@@ -1,0 +1,186 @@
+import * as React from "react";
+import { faUser } from "@fortawesome/free-solid-svg-icons";
+import { uniq, includes } from "lodash";
+import GlobalAppContext from "../utils/GlobalAppContext";
+import ListenControl from "../listens/ListenControl";
+
+export type PersonalRecommendationModalProps = {
+  recordingToPersonallyRecommend?: Listen;
+  newAlert: (
+    alertType: AlertType,
+    title: string,
+    message: string | JSX.Element
+  ) => void;
+  onSuccessfulPersonalRecommendation?: UserTrackPersonalRecommendationMetadata;
+};
+
+export interface PersonalRecommendationModalState {
+  blurbContent: string;
+  users: Array<string> | null;
+  followers: Array<string> | null;
+  searchBoxFocus: Boolean;
+}
+
+export default class PersonalRecommendationModal extends React.Component<
+  PersonalRecommendationModalProps,
+  PersonalRecommendationModalState
+> {
+  static contextType = GlobalAppContext;
+  declare context: React.ContextType<typeof GlobalAppContext>;
+  readonly maxBlurbContentLength = 280;
+
+  constructor(props: PersonalRecommendationModalProps) {
+    super(props);
+    this.state = {
+      blurbContent: "",
+      users: [],
+      followers: [],
+      searchBoxFocus: false,
+    };
+  }
+
+  componentDidMount(): void {
+    const { APIService, currentUser } = this.context;
+    APIService.getFollowersOfUser(currentUser.name).then((respose) => {
+      this.setState({ followers: respose.followers });
+    });
+  }
+
+  handleBlurbInputChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
+    event.preventDefault();
+    const input = event.target.value.replace(/\s\s+/g, " "); // remove line breaks and excessive spaces
+    if (input.length <= this.maxBlurbContentLength) {
+      this.setState({ blurbContent: input });
+    }
+  };
+
+  handleError = (error: string | Error, title?: string): void => {
+    const { newAlert } = this.props;
+    if (!error) {
+      return;
+    }
+    newAlert(
+      "danger",
+      title || "Error",
+      typeof error === "object" ? error.message : error
+    );
+  };
+
+  addUser = (user: string) => {
+    const { users } = this.state;
+    users!.push(user);
+    this.setState({ users: uniq(users) });
+  };
+
+  render() {
+    const { recordingToPersonallyRecommend } = this.props;
+    if (!recordingToPersonallyRecommend) {
+      return null;
+    }
+    const { blurbContent, users, followers, searchBoxFocus } = this.state;
+    const { track_name } = recordingToPersonallyRecommend.track_metadata;
+    const { artist_name } = recordingToPersonallyRecommend.track_metadata;
+    return (
+      <div
+        className="modal fade"
+        id="PersonalRecommendationModal"
+        tabIndex={-1}
+        role="dialog"
+        aria-labelledby="PersonalRecommendationModalLabel"
+        data-backdrop="static"
+      >
+        <div className="modal-dialog" role="document">
+          <form className="modal-content">
+            <div className="modal-header">
+              <button
+                type="button"
+                className="close"
+                data-dismiss="modal"
+                aria-label="Close"
+                onClick={() => {
+                  this.setState({ users: [], blurbContent: "" });
+                }}
+              >
+                <span aria-hidden="true">&times;</span>
+              </button>
+              <h4 className="modal-title" id="PersonalRecommendationModalLabel">
+                Personally recommend this recording
+              </h4>
+            </div>
+            <div className="modal-body">
+              <div className="input-group">
+                <input
+                  type="text"
+                  className="form-control"
+                  onFocus={() => {
+                    this.setState({ searchBoxFocus: true });
+                  }}
+                  onBlur={() => {
+                    this.setState({ searchBoxFocus: false });
+                  }}
+                />
+                <span className="input-group-btn">
+                  <button className="btn btn-default" type="button">
+                    Add
+                  </button>
+                </span>
+              </div>
+              <div className={`searchdropdown ${searchBoxFocus ? "open" : ""}`}>
+                {followers!.map((name) => {
+                  return (
+                    <ListenControl
+                      text={name}
+                      // eslint-disable-next-line react/jsx-no-bind
+                      action={this.addUser.bind(this, name)}
+                      icon={faUser}
+                    />
+                  );
+                })}
+              </div>
+              <p>
+                Why will you recommend personally{" "}
+                <b>
+                  {" "}
+                  {track_name} by {artist_name}
+                </b>
+                ? (Optional)
+              </p>
+              <div className="form-group">
+                <textarea
+                  className="form-control"
+                  id="blurb-content"
+                  placeholder="Let your followers know why you recommended this song"
+                  value={blurbContent}
+                  name="blurb-content"
+                  rows={4}
+                  style={{ resize: "vertical" }}
+                  spellCheck="false"
+                  onChange={this.handleBlurbInputChange}
+                />
+              </div>
+              <small style={{ display: "block", textAlign: "right" }}>
+                {blurbContent.length} / {this.maxBlurbContentLength}
+              </small>
+            </div>
+            <div className="modal-footer">
+              <button
+                type="button"
+                className="btn btn-default"
+                data-dismiss="modal"
+              >
+                Cancel
+              </button>
+              <button
+                type="submit"
+                className="btn btn-success"
+                data-dismiss="modal"
+              >
+                Personally Recommend
+              </button>
+            </div>
+          </form>
+        </div>
+      </div>
+    );
+  }
+}


listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+      try {
+        const status = await APIService.submitPersonalRecommendation(
+          currentUser.auth_token,
+          currentUser.name,
+          metadata
+        );
+        if (status === 200) {
+          newAlert(
+            "success",
+            "You personally recommended a track!",
```
MonkeyDo, line 135: I wonder about the phrasing here, it sounds a bit off.
Perhaps it would make more sense to say something like `You recommended this track to ${users.length} user${users.length >1 ? 's' : ''}`

Similarly in the error I would just say "Error while recommending a track"

What do you think?

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+  };
+
+  render() {
+    const { recordingToPersonallyRecommend } = this.props;
+    if (!recordingToPersonallyRecommend) {
+      return null;
+    }
+    const { blurbContent, users, suggestions } = this.state;
+    const { track_name } = recordingToPersonallyRecommend.track_metadata;
+    const { artist_name } = recordingToPersonallyRecommend.track_metadata;
```
MonkeyDo, line 153: These can be grouped like so:

```suggestion
    const { track_name, artist_name } = recordingToPersonallyRecommend.track_metadata;
```

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+                  });
+                  // update the followers list
+                  APIService.getFollowersOfUser(currentUser.name)
+                    .then((response) => {
+                      this.setState({ followers: response.followers });
+                    })
+                    .catch((error) => {
+                      this.handleError(error, "Error while fetching followers");
+                    });
+                }}
```
MonkeyDo, line 186: It's best no to have anonymous functions defined in the render method, because they will get re-created on every re-render (in theory that's not optimal).

What you want is to have a method defined on the class such as `onCloseModal` that will call setState.
You will also want to refactor the code to get followers that you have in `componentDidMount` into its own method `getFollowers`, so that you can call it from componentDidMount and from onCloseModal without having duplicate code.

P.S: why do we need to update the followers list when we close the modal? These two things don't seem like they should be related

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+              >
+                <span aria-hidden="true">&times;</span>
+              </button>
+              <h4 className="modal-title" id="PersonalRecommendationModalLabel">
+                Personally recommend this recording
+              </h4>
+            </div>
+            <div className="modal-body">
+              {users!.map((user) => {
```
MonkeyDo, line 195: Why the need for the `!`?
That's usually a "bad smell" (as programmers call it) and point to the type not being well defined.

I see in `PersonalRecommendationModalState` the arrays are also described as being potentially `null`, but looking at the code they will never be so and should simply be an array.
For example in the constructor you explicitly set them as an empty array, and in all your setState calls you are careful and always them them to arrays too.
That way the need for the `!` will go away.

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+                // eslint-disable-next-line react/jsx-no-bind
+                action={this.addUser}
+              />
+              <p style={{ marginTop: "24px", marginBottom: "24px" }}>
+                Tell the users selected above why are you recommending them{" "}
+                <b>
+                  {" "}
+                  {track_name} by {artist_name}
+                </b>
+                ? (Optional)
```
MonkeyDo, line 222: I think this is left over from previous wording
```suggestion
                (Optional)
```

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+              <div className="form-group">
+                <textarea
+                  className="form-control"
+                  id="blurb-content"
+                  placeholder="Let your followers know why you recommended this song"
+                  value={blurbContent}
+                  name="blurb-content"
+                  rows={4}
+                  style={{ resize: "vertical" }}
+                  spellCheck="false"
```
MonkeyDo, line 233: Y u no specllheck?

:p 
Jokes aside, what is the rationale for disabling this?

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+                  onChange={this.handleBlurbInputChange}
+                />
+              </div>
+              <small style={{ display: "block", textAlign: "right" }}>
```
MonkeyDo, line 237: Since you already have a .less file defined for this component, I'd like to see all the `style={…` in this file moved to LESS definitions instead. It's much easier to keep things up to date when the style are defined in only one place.

listenbrainz/webserver/static/js/src/personal-recommendations/SearchDropDown.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from "react";
+import { faUser } from "@fortawesome/free-solid-svg-icons";
+import ListenControl from "../listens/ListenControl";
+
+export type SearchDropDownProps = {
+  action: (event: string) => void;
+  suggestions?: Array<string> | null;
```
MonkeyDo, line 7: Same comment as on the other file, I don't think the prop can be `null` so this should be:
```suggestion
  suggestions?: Array<string>;
```

The `?` already marks it as optional, which translates to `Array<string> | undefined`.


With that change, the `!` on line 14 should be removed.

listenbrainz/webserver/static/js/src/utils/types.d.ts
```diff
@@ -516,6 +516,12 @@ declare type UserTrackRecommendationMetadata = {
   recording_msid: string;
 };
 
+/** For recommending a track personally from the front-end */
+declare type UserTrackPersonalRecommendationMetadata = UserTrackRecommendationMetadata & {
+  blurb_content: string | null;
+  users: Array<string> | null;
```
MonkeyDo, line 522: These should probably be:
I assume blurb content could be an empty string rather than null, and same for an empty users array
```suggestion
  blurb_content: string;
  users: Array<string>;
```

listenbrainz/webserver/static/js/src/utils/utils.tsx
```diff
@@ -631,6 +649,25 @@ export function getReviewEventContent(
...
+export function getPersonalRecommendationEventContent(
+  eventMetadata: UserTrackPersonalRecommendationMetadata,
+  isCreator: Boolean
+): JSX.Element {
+  const additionalContent = getAdditionalContent(eventMetadata);
+  return (
+    <>
+      {isCreator && (
+        <div style={{ paddingBottom: "10px" }}>
+          {eventMetadata.users!.map((userName) => {
```
MonkeyDo, line 661: Another `!` I'd like to see removed ;)

listenbrainz/webserver/static/js/tests/personal-recommendations/NamePill.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+    const wrapper = mount(<NamePill {...props} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("works when you click", () => {
+    const wrapper = mount(<NamePill {...props} />);
+    wrapper.find("button").at(0).simulate("click");
+    expect(mockAction).toHaveBeenCalledTimes(1);
+  });
+});
```
MonkeyDo, line 32: I'd like to see another test to ensure the button is not rendered if `closeButton` is false  (or according to my suggestion above if `closeAction`  is not a Function)

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
```diff
@@ -202,6 +220,10 @@ export default class RecentListens extends React.Component<
                     artistMBIDs?.length ||
                     Boolean(trackMBID) ||
                     Boolean(releaseGroupMBID);
+
+                  const isListenPersonallyRecommendable = Boolean(
+                    getRecordingMSID(listen)
```
MonkeyDo, line 225: Flagging this, but I don't have a direct suggestion:
we're moving aay from using MSIDs as our primary ids for anything, so this feature should be based first on the `recordingMBID`  and only if not available should it be the `recordingMSID`.
Similarly in the API call on the back end we should use MBID first with an optional fallback to MSID

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,263 @@
...
+    const { APIService, currentUser } = this.context;
+
+    if (currentUser?.auth_token) {
+      const metadata: UserTrackPersonalRecommendationMetadata = {
+        artist_name: getArtistName(recordingToPersonallyRecommend),
+        track_name: getTrackName(recordingToPersonallyRecommend),
+        release_name: recordingToPersonallyRecommend!.track_metadata
+          .release_name,
+        recording_mbid: getRecordingMBID(recordingToPersonallyRecommend!),
+        recording_msid: getRecordingMSID(recordingToPersonallyRecommend!),
```
MonkeyDo, line 122: Please remove the `!`  once the type has been corrected.

MonkeyDo (changes requested):
Good job @hrik2001 !

I haven't run it locally yet, there are a few things to improve in the code. Most of it is related to some subtleties of React and of Typescript, and I've only seen minor issues.

I will have another round of review, mainly for the visual aspect which I haven't looked at yet (you were proactive with the requests for feedback on your mockups so I'm not worried).

But all in all it looks quite good, including some nice looking tests!

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
hrik2001: Hi, sorry for the late reply as I was busy conducting Software Freedom Day event :)

I disabled spellcheck because other modals have it disabled (Pin Recording Modal and CB Review Modal)
I did think if it's needed or not, but then I realized that I often use "Hinglish" (Hindi but with latin script) which might come off as wrong in the textbox.

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
hrik2001: I see, this is interesting, I will make changes to the backend and frontend. But I have a question
All the mbid fields in local machine are usually empty. If we do move, development will get hard. I believe we should work on some mechanism to generate dummy MB for development purposes. What do you say?

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
hrik2001: Makes sense, thanks

> P.S: why do we need to update the followers list when we close the modal? These two things don't seem like they should be related

I did it mostly for the UX
for example, when someone sends you an email, you would refresh until you get the email. Similarly, suppose you are recommending someone, and you wrote a huge recommendation, only to realise that someone doesn't follow you, and you make ask that person to follow you. But sadly you would have to reload the whole page to have an updated list.

MonkeyDo: I think that's not a bad idea, but maybe not the right design for that.
Without a clear indication I wouldn't suspect that the list would refresh automatically.
Maybe a well-placed refresh button might be helpful for that? I wonder how much use it will get, but at the same time if it just calls a function that is already implemented it doesn't cost anything to add it.

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
MonkeyDo: Hm, interesting . Didn't realize that was the case on the pin modal.
I tentatively think we shouldn't disable it (for either modal, but let's focus on this one for this PR).
If we're struggling to find a rationale for disabling it it's probably best to keep browser defaults that users would expect.

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
MonkeyDo: For posterity: there was some discussion on IRC about this, lucifer suggested using an existing script to populate the local DB with an album from MusicBrainz: https://listenbrainz.readthedocs.io/en/latest/developers/commands.html#develop-sh-manage-submit-release

MonkeyDo:
@hrik2001 Hi! Is this ready for another review?
If not, anything I can do to help?

hrik2001:
Hi @MonkeyDo, the PR is ready for review, very very sorry for leaving you hanging, I was down with covid apparently, but I have recovered! I couldn't even attend the MB summit stream :/

The new commits do everything you asked for except of the shift to mbid and the `isFunction` function that you suggested. Opted it out for now because jest's mock function doesn't get qualified as a function.

hrik2001:
Also @MonkeyDo, I do need your help in one thing. For some reason, after merging the master, I have errors in frontend tests (and those errors don't appear to be happening in the tests here in github). Due to this reason, I am not able to generate snapshots. (You can see the only tests that are failing are snapshot tests)

MonkeyDo:
> Hi @MonkeyDo, the PR is ready for review, very very sorry for leaving you hanging, I was down with covid apparently, but I have recovered! I couldn't even attend the MB summit stream :/

Oh damn, I hope it wasn't too hard on you! Glad to hear you're doing better though.
Sorry if I sounded pushy, I just wanted to make sure the PR doesn't end in limbo as sometimes happens; I'm just so much looking forward to using the feature ! :)

I'll do another review pass and look at the tests. Without having looking at anything, I'd wager it's probably related to some recent package updates I've been working on, so might require rebuilding some docker images (`./develop.sh build web static_builder`).

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
MonkeyDo: Is there a changed required here? I have no visibility on how the whole MBID vs MSID thing is implemented on the back-end so you'll have to tell me :)
Or do we want to change both front-end and back-end at the same time in a separate PR?

amCap1712: Since https://github.com/metabrainz/listenbrainz-server/pull/2153, recommendations support both MBID and MSID. Either one of the two is sufficient.

MonkeyDo: Thanks Captain @amCap1712 !
So no change needed then?

amCap1712: IIUC, a change is needed to allow for the case when there is only mbid present. Instead of `Boolean(getRecordingMSID(listen))`, something like 
`Boolean(getRecordingMSID(listen)) ||Boolean(getRecordingMBID(listen))` I guess?

MonkeyDo:
Well, I was surprised and a bit confused by the comment about isFunction and the jest mocks, because I was expecting jest mock functions to pass as regular functions.
So I tried it…and it works! Not sure what issue you hit when you tried it out, but I took the liberty of pushing those extra changes.

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+                // eslint-disable-next-line react/jsx-no-bind
+                action={this.addUser}
+              />
+              <p>
+                Tell the users selected above why are you recommending them{" "}
+                <b>
+                  {" "}
+                  {track_name} by {artist_name}
+                </b>
+                (Optional)
```
MonkeyDo, line 218: Making this smaller and gray text, as well as adding a missing space:
```suggestion
                <small> (Optional)</small>
```
This also requires changing some css used for the character count: currently the rule is applied to all `<small>` elements (`#PersonalRecommendationModal small`), so we probably want to add a className `character-count` (for example) to the corresponding element and target that instead (`#PersonalRecommendationModal .character-count`

Before 
image

After:
image

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+                type="button"
+                className="close"
+                data-dismiss="modal"
+                aria-label="Close"
+                onClick={this.closeModal}
+              >
+                <span aria-hidden="true">&times;</span>
+              </button>
+              <h4 className="modal-title" id="PersonalRecommendationModalLabel">
+                Personally recommend this recording
```
MonkeyDo, line 188: I think what would be great here would be to use the recording name instead.
Maybe something like `Recommend <b>${track_name}</b> to a friend`, what do you think ?

image

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+              </h4>
+            </div>
```
MonkeyDo, line 190: I wonder what this should look like if the user doesn't have any followers to recommend to.
I think maybe instead of the modal body below we can return a bit of text explaining that in order to recommend directly to another user they need to follow you, that you can follow people by clicking next to their username on their listens page, and finally that closing and reopening this popup window will refresh the list.

listenbrainz/webserver/static/js/src/personal-recommendations/PersonalRecommendationsModal.tsx
```diff
@@ -0,0 +1,258 @@
...
+              >
+                Cancel
+              </button>
+              <button
+                type="submit"
+                className="btn btn-success"
+                data-dismiss="modal"
+                onClick={this.submitPersonalRecommendation}
+              >
+                Personally Recommend
```
MonkeyDo, line 250: I'm on the fence with the wording (both here and on the listencard dropdown menu item).
What do you think of "Send to a friend" or "Send to a follower"?
Or if we want to stay with "recommend", "Recommend to a follower"?

You could ask around on the IRC channel, perhaps someone else has a good idea.

listenbrainz/webserver/static/js/src/user-feed/UserFeed.tsx
```diff
@@ -150,6 +157,8 @@ export default class UserFeedPage extends React.Component<
           review.entity_type
         )}`;
       }
+      case EventType.PERSONAL_RECORDING_RECOMMENDATION:
+        return "personally recommended a track";
```
MonkeyDo, line 161: ~~In the same discussion about phrasing, how about "UserX recommended you a track" or "UserX sent you a track" ?~~
Now that I saw the other end of it (i.e. on my own feed), I see that wouldn't work ("You recommended you a track").
Back to the drawing board for that one…

listenbrainz/webserver/static/js/src/utils/utils.tsx
```diff
@@ -648,6 +666,25 @@ export function getReviewEventContent(
...
+export function getPersonalRecommendationEventContent(
+  eventMetadata: UserTrackPersonalRecommendationMetadata,
+  isCreator: Boolean
+): JSX.Element {
+  const additionalContent = getAdditionalContent(eventMetadata);
+  return (
+    <>
+      {isCreator && (
+        <div style={{ paddingBottom: "10px" }}>
+          {eventMetadata.users.map((userName) => {
```
MonkeyDo, line 678:
```suggestion
          Sent to: {eventMetadata.users.map((userName) => {
```

listenbrainz/webserver/static/js/src/utils/utils.tsx
```diff
@@ -648,6 +666,25 @@ export function getReviewEventContent(
...
 
+export function getPersonalRecommendationEventContent(
+  eventMetadata: UserTrackPersonalRecommendationMetadata,
+  isCreator: Boolean
+): JSX.Element {
+  const additionalContent = getAdditionalContent(eventMetadata);
+  return (
+    <>
+      {isCreator && (
+        <div style={{ paddingBottom: "10px" }}>
```
MonkeyDo, line 677: Can we put this in the less file instead?

listenbrainz/webserver/static/js/tests/personal-recommendations/NamePill.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+describe("NamePill", () => {
+  it("renders correctly", () => {
+    const wrapper = mount(<NamePill {...props} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it("works when you click", () => {
+    const wrapper = mount(<NamePill {...props} />);
+    wrapper.find("button").at(0).simulate("click");
+    expect(mockAction).toHaveBeenCalledTimes(1);
```
MonkeyDo, line 29: Here you can add another expectation:
`expect(mockAction).toHaveBeenCalledWith(expected arguments)`
just to make sure it's called with the right arguments

listenbrainz/webserver/static/js/tests/personal-recommendations/PersonalRecommendationsModal.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    );
+    spy.mockImplementation((userToken, userName, metadata) => {
+      return Promise.resolve(200);
+    });
+
+    wrapper.setState({ blurbContent: "hii", users: ["riksucks", "hrik2001"] });
+    const setStateSpy = jest.spyOn(instance, "setState");
+    await instance.submitPersonalRecommendation();
+    expect(setStateSpy).toHaveBeenCalledTimes(1);
+    expect(wrapper.state("blurbContent")).toEqual("");
```
MonkeyDo, line 112: The "belt and braces" approach here is to first ensure that the state is indeed what you expect as a "normal" condition before you take an action.
In this case, it means adding `expect(wrapper.state("blurbContent")).toEqual("hii");` right after your setState call.

MonkeyDo (changes requested):
Nice work @hrik2001 !
I've run this locally and it's working smoothly !

Some comments below, but we're now down to details which is great.

Don't hesitate to bring the phrasing discussion on IRC, I suspect people like @Aerozol might have good suggestions.

hrik2001:
Hi @MonkeyDo, it wasn't pushy at all, I just felt very guilty for not updating you at all, I should have done that.

Also thank you so much for helping me out with the commits, the new reviews are crystal clear, will update you soon chief 🫡

Aerozol:
My only feedback:
Can we move ‘Optional:’ to the front of that second sentence?
Or even just have ‘Optional:’ and get rid of everything else, since the text in the box explains it clearly enough?

hrik2001:
Hi @MonkeyDo @Aerozol. This is what I came up with so far
image
1) There was dilemma over "Send Song" and "Personally Recommend", so I felt like "Send Recommendation" sounds better
2) Added a small write-up to explain the case of not having followers
3) What do you think about remove the paragraph which says "Tell your above chosen followers why are you recommending them 1985 by Freddie Gibbs, The Alchemist (Optional)" as @Aerozol suggested. I support it, but other modals have a similar sentence, hence I have kept it here for now. Should we delete from other modals too if it's decided to remove from here?


Do you think there is more room for improvement? Please do let me know


Also, sorry for the late PR, the work was mostly done way earlier but couldn't be pushed because of a very embarrassing reason. My tests were broken and to my surprise rebuilding every image and creating container wouldn't fix anything, merging master added more problem. After spending a lot of time, it came to my notice, that the build was cached. Even though the files were updated, and the images were deleted, docker still somehow used the old cache. All I had to do was add `--no-cache` flag and everything worked like a charm! XD

Aerozol:
@hrik2001: Looking good, my input is to simplify stuff as much as possible:

The title ‘Recommend [song] to certain follower(s)’ is clunky.
Maybe just ‘Recommend [song]’? Then in the next box have ‘Search and add followers*’, and add the asterisk (*) to the start of the last paragraph explaining why you might not be able to find the user in the drop down.

If we’re keeping the ‘tell your chosen…’ line let’s just have something like: ‘Add a message (optional)’.

For the last paragraph maybe something like:
*Can’t find a user? Make sure they are following you, and then refresh the page.

hrik2001:
@Aerozol fair point. You are letting the UI and the UX do the job for explaining to the user rather than being explicit. I will update the UI and show it here and then in the IRC too.

@MonkeyDo do you think explaining that the user should go to the listens page and then click the button next to it is necessary, or do you think it's fairly understood

MonkeyDo:
Nice job ! Looking better and better.

I agree with everything aerozol said, including the shortened help text at the bottom.
The UI is likely to change dramatically soon, and I can guarantee we'll forget to update this bit of text if the process changes :)
If users can't figure how to follow another user, that's another (big) problem that shouldn't be your concern at this point.

For the modal title, maybe "Send [song] to friends" or "Share [song]"  ?

> other modals have a similar sentence […] Should we delete from other modals too if it's decided to remove from here?
That's not necessary in this PR at least. I don't think this is a required style of how we do modals, and other modals might need the extra helping text — but then again perhaps they don't and we can revisit that. I think it's only the pin recording modal? The CB review one doesn't have one.

listenbrainz/webserver/static/js/src/recent/RecentListens.tsx
MonkeyDo: @hrik2001 We still need the change that lucifer suggested above.
You can use `Boolean(recordingMBID)` since we already use that for `isListenReviewable`

listenbrainz/webserver/static/js/tests/personal-recommendations/NamePill.test.tsx
MonkeyDo: Bump :)

MonkeyDo (commented):
Good going! 
Apart form the comments about UI, there are two —relatively minor— pieces of feedback from before that haven't been addressed. I added a comment to each so they should be easy to find.

hrik2001:
Oops, I had accidentally changed it to `recordingMSID`

Also, I wanted to ask you about the parameter one, when `NamePill` calls the function, it can't call with any parameter (apart from mouse event).
In the modal code, the bind method creates a function, that wraps an existing function with a parameter, which is how the parameter is passed to the `NamePill` component.

So should I write a test to see if the function is called with a mouse event or not?

MonkeyDo:
> 

Ah, you're right my comment was a generic one, but it won't be of any this in this particular situation, so let's leave it like it is. :)

hrik2001:
Hi, I believe everything has been done, please feel free to leave feedbacks. :)

MonkeyDo (changes requested):
Only small tidbits left !

I see a couple of our previous comments about the wording were not addressed; it's always open for discussion, but if you don't think they're good ideas, I'd like to heard some arguments :)
If you just forgot about them, ignore the above :p

Wording comments are here:
https://github.com/metabrainz/listenbrainz-server/pull/2131#issuecomment-1292930816
https://github.com/metabrainz/listenbrainz-server/pull/2131#issuecomment-1293408203

hrik2001:
image
image
@MonkeyDo @Aerozol 
Opinions?

MonkeyDo:
Starting to look great !
I think the "search and add followers*" text should go above the user selection component, that'll make more sense.
Apart form that I'm happy to merge, deploy and get a feel for it.

Aerozol:
Nice!
Yup, my suggestion was the ‘search and add followers*’ to go in the search box (actually could just be ‘search followers*’ now that I think about it)…
And then ‘Add a message (optional)’ in the space under it

MonkeyDo (approved):

MonkeyDo:
Awesome work @hrik2001 !
So happy to see this feature out !

PR #2132, opened 2022-08-30
Cache wordpress API locally in LB and serve to clients from the cache

alastair:
# Problem
When a user visits LB, we shouldn't share their IP address to wordpress when we render the list of recent blog posts

# Solution
Add an endpoint on the LB server which downloads and caches the wordpress recent posts API endpoint. Cache this data for an hour, return an error back to the client and don't cache anything if wordpress returns us bad data.

### DIFF
--- a/listenbrainz/webserver/static/js/src/home/Blog.tsx
+++ b/listenbrainz/webserver/static/js/src/home/Blog.tsx
@@ -1,23 +1,23 @@
 import React, { useEffect, useState } from "react";
 
-const Blog = () => {
-  const [blogDetails, setBlogDetails] = useState([{}]);
+const Blog = ({ apiUrl }: { apiUrl: string }) => {
+  const [blogDetails, setBlogDetails] = useState<object[]>();
 
   const fetchBlogDetails = async () => {
     try {
-      let response: any = await fetch(
-        `https://public-api.wordpress.com/rest/v1.1/sites/blog.metabrainz.org/posts/`
-      );
-      response = await response.json();
-      const objectArray: any = [];
-      for (let postIndex = 0; postIndex <= 6; postIndex += 1) {
-        const object: any = {};
-        object.id = response.posts[postIndex].ID;
-        object.title = response.posts[postIndex].title;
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/static/js/src/home/Blog.tsx
```diff
@@ -1,23 +1,25 @@
...
+        for (let postIndex = 0; postIndex <= 6; postIndex += 1) {
+          const object: any = {};
+          object.id = response.posts[postIndex].ID;
+          object.title = response.posts[postIndex].title;
+          object.link = response.posts[postIndex].URL;
+          objectArray.push(object);
+        }
+        setBlogDetails(objectArray);
+        // eslint-disable-next-line no-console
+        console.log("SETTINGGGG");
```
alastair, line 21: 😳

amCap1712 (commented):
Lets check once on test or beta once but otherwise LGTM, Thanks!

amCap1712 (approved):

listenbrainz/webserver/static/js/src/home/Blog.tsx
```diff
@@ -1,23 +1,23 @@
...
-      );
-      response = await response.json();
-      const objectArray: any = [];
-      for (let postIndex = 0; postIndex <= 6; postIndex += 1) {
-        const object: any = {};
-        object.id = response.posts[postIndex].ID;
-        object.title = response.posts[postIndex].title;
-        object.link = response.posts[postIndex].URL;
-        objectArray.push(object);
+      let response: any = await fetch(`${apiUrl}/blog-data/`);
```
amCap1712, line 8: The endpoint is on the website url not the api url so should probably use `${window.location.origin}` instead of api url.

alastair: fixed

amCap1712:
Failing tests are fixed in https://github.com/metabrainz/listenbrainz-server/pull/2143. can ignore for now.

amCap1712:
Thanks for the fix!

PR #2135, opened 2022-08-31
Simplify mb_metadata_cache queries

amCap1712:
1. Remove redundant joins in mb_metadata_cache query

   We join to artist and artist_credit_name in the last query in mb_metadata_cache but as far as I can see we don't use any data from those or removing the JOINs.

2. Remove JOINs to artist_credit where not needed

    At multiple places, we have JOINs of the form:

    ```sql
     FROM recording r
     JOIN artist_credit ac
       ON r.artist_credit = ac.id
     JOIN artist_credit_name acn
       ON acn.artist_credit = ac.id
     ```

     where we do not use the values of artist_credit table. In these cases we can simplify the JOINs to:

     ```sql
      FROM recording r
      JOIN artist_credit_name acn
   USING (artist_credit)
   ```


mayhem (approved):
Thanks for these improvements!

PR #2136, opened 2022-09-05
add timezone field in listens

YouzhiL:
# Problem

Submitted listens are not converted into the correct timezone when displayed to the user if someone is traveling.

# Solution

Allow clients to submit the local timestamp in ISO8601 format with a local timezone specified.

# Action

1. Add new attribute `listened_at_local` in `additional_info`.
2. validate `listened_at_local` format.

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -238,6 +238,16 @@ def validate_listen(listen: Dict, listen_type) -> Dict:
         for key in multiple_mbid_keys:
             validate_multiple_mbids_field(listen, key)
 
+        # check listened_at_local validity. listened_at_local should be an ISO8601 string which follows the format:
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/api_tools.py
```diff
@@ -238,6 +238,16 @@ def validate_listen(listen: Dict, listen_type) -> Dict:
...
             validate_multiple_mbids_field(listen, key)
 
+        # check listened_at_local validity. listened_at_local should be an ISO8601 string which follows the format:
+        # YYYY-MM-DDThh:mm:ss+hh:mm or YYYY-MM-DDThh:mm:ss-hh:m
+
+        if 'listened_at_local' in listen['track_metadata']['additional_info']:
+            listened_at_local = listen['track_metadata']['additional_info']['listened_at_local']
+            try:
+                iso_time = datetime.fromisoformat(listened_at_local)
+            except:
```
alastair, line 248: We should be as specific as possible when catching exceptions:

```suggestion
            except ValueError:
```

PR #2139, opened 2022-09-07
Update start-spark-request-consumer.sh

amCap1712:
These changes are part of the move from yarn to spark standalone cluster . I had forgotten to check in these changes earlier and only realised it when debugging the cluster's failure to launch after today's release.

### DIFF
--- a/docker/start-spark-request-consumer.sh
+++ b/docker/start-spark-request-consumer.sh
@@ -1,3 +1,4 @@
+  GNU nano 4.8                                                                      docker/start-spark-request-consumer.sh                                                                       Modified
 #!/bin/bash
 set -e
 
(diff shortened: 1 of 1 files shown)


docker/start-spark-request-consumer.sh
```diff
@@ -1,3 +1,4 @@
+  GNU nano 4.8                                                                      docker/start-spark-request-consumer.sh                                                                       Modified
```
mayhem, line 1: Eh?

mayhem (approved):
Not too familiar with the various options, but looks sane. Just remove that line 1. :)

amCap1712:
Uhh, nvm. I see now that I forgot to merge #2097.

docker/start-spark-request-consumer.sh
amCap1712: :facepalm:

PR #2140, opened 2022-09-09
Footer typo

Aerozol:
# Problem

Typo in the footer.

# Solution

Removed the extra ‘s’, removed an extraneous word while I was at it.
Text could be made a bit more snappy in general but just fixing the typo for now!

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 
 ## About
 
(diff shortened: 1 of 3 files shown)


MonkeyDo (approved):
Thanks !

PR #2141, opened 2022-09-13
Fix test_get_listens_ts_unavailable

amCap1712:
This test sets timescale_connection._ts to None in order to test the behaviour when listenstore behaviour. However, it never reverts the listenstore to its actual value. This is an issue since we do not create a flask app for every method now but only once per class. Any call to get listens endpoint in any test run after this test in the same class will fail with a 503.

Use mock.patch to auto restore the value after the test to fix the issue.


alastair (approved):
👌

PR #2144, opened 2022-09-13
Don't fail webpack build for style issues in development

MonkeyDo:
As requested by @alastair

In development mode, configure the Stylelint webpack plugin to not fail module compilation if there is a linting error.
This does not change the production mechanism, so CI tests will still fail on linting errors.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -46,6 +46,7 @@ module.exports = function (env, argv) {
     }),
     new StylelintPlugin({
       configFile: ".stylelintrc.js",
+      failOnError: isProd,
       files: "**/static/css/**/*.less",
       fix: !isProd,
       threads: true,


alastair:
We discussed this a bit - what this means is that the static builder in development won't fail in some cases (such as adding console.log without an ignore comment above it). This will make local development a little bit easier in some cases. Note that the build/lint process will still automatically re-format code if necessary.

When opening a PR, the "ListenBrainz Build Production Image" workflow will run, which runs webpack in prod mode, and means that the build process will fail if a lint check fails. This allows us to easily find js errors that need to be fixed before the PR is merged.

In the case that we want to run the linters locally, we can run eslint manually with `./test.sh fe -f`

alastair (approved):
thank you!

PR #2147, opened 2022-09-14
Add foobar2000 player and its plugin to compatible players list

regorxxx:
Adds [foobar2000](https://www.foobar2000.org/components) player and its [plugin](https://github.com/phw/foo_listenbrainz2) to compatible players list.

Entire list has been rearranged to put cross platform players at top, then windows-only players, then unix-only ones, Android ones and finally web browsers plugins. Previously they were scattered without any specific order.

VLC entry has been reworked to link both the software and its plugin, like the other entries. Previously only the plugin was linked to


MonkeyDo (approved):
Great, thank you very much ! :rocket:

PR #2149, opened 2022-09-17
Clicking user now takes you to their listens page

Aerozol:
As discussed with Monkey on IRC, charts is not often where you want to go when clicking a users name. Changed URL to the default/Listens page.

(chat log: https://chatlogs.metabrainz.org/libera/metabrainz/2022-09-15/?msg=5055517&page=1)


MonkeyDo:
I took the liberty to update some test things for you.
In short, as part of our tests we have snapshots of what the HTML of our pages or elements should look like, so that if something changes we know because the tests fail.
This is what happened in this case, it was showing two snapshot-related failures, which is expected since we changed the anchor tag link.

To figure out what is going wrong you can click on "details" next to a failing automated test suite:
image
This should automatically scroll you down to the test results, in which in this case you can see there are two failing tests all in all, and both are snapshot-related failures (which means nothing 'wrong' per se, just needs an update): https://github.com/metabrainz/listenbrainz-server/actions/runs/3072103750/jobs/4963357026#step:7:8311

To update the snapshots, with the project set up on your local machine you can run this command in your console/terminal to run the test suite while also updating the snapshots at the same time: `./test.sh fe -u`
You can find that documented in this section of our dev docs: https://listenbrainz.readthedocs.io/en/latest/developers/devel-env.html?highlight=tests#test-your-changes-with-unit-tests

Then you will see there will be changes to the snapshot files which you'll need to commit and push, same process as you usually use to commit and push changes.

(Or in any case, you can always come ask me, I'm happy to regenerate the snapshots for you :) )

MonkeyDo (approved):
With tests passing this is good to go ! :rocket:

PR #2150, opened 2022-09-17
Changing default modal view to ‘following’

Aerozol:
As discussed on IRC, feedback welcome.

Reasons:
- ‘followers’ is not necessarily curated/picked by/a reflection of a user
- I follow my friends and then regularly use this panel to visit their pages. I rarely check on followers
- If it’s empty it might encourage users to follow someone


MonkeyDo (commented):
Thanks for making the change !

There are two things required to makes the tests pass, if you want to try it out yourself:
1. updating the test snapshots (I detailed the process in #2149 but don't hesitate if you have questions—or if you can't be bothered ;) )
2. Modifying the following test to replace 'following' with 'follower' and vice-versa: https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/tests/follow/FollowerFollowingModal.test.tsx#L63-L71

let me know if you have questions !

MonkeyDo:
Here are the instructions for setting up a local development environment of ListenBrainz, so that you can run tests and all that:
https://listenbrainz.readthedocs.io/en/latest/developers/devel-env.html
You can most likely ignore the bit about Last.fm/Spotify importers and CritiqueBrainz reviews and after the configuration step go straight to [#test-your-changes-with-unit-tests](https://listenbrainz.readthedocs.io/en/latest/developers/devel-env.html#test-your-changes-with-unit-tests)

It can seem pretty overwhelming at first (I still remember the first time I had to set up a MusicBrainz development environment way back when I made the logos, I didn't code a whole lot at the time…), but the instructions are fairly easy to follow and we're here to answer any questions !

Aerozol:
Thanks Monkey! (and lucifer) I followed the steps, and got a bunch of errors when running the test - I’m assuming because I only set up part of the server on docker?

I also got a lot of the following when I staged everything:
```
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in listenbrainz/webserver/static/js/tests/follow/__snapshots__/UserSocialNetwork.test.tsx.snap.
```

Anyway, I thought maybe it’s easiest if I just add everything to the pull request and then you can have a look and let me know if there’s anything I should worry about :)

listenbrainz/config.py.sample
```diff
@@ -76,8 +76,8 @@ COUCHDB_USER = "listenbrainz"
 COUCHDB_ADMIN_KEY = "listenbrainz"
 
 # MusicBrainz OAuth
-MUSICBRAINZ_CLIENT_ID = "CLIENT_ID"
-MUSICBRAINZ_CLIENT_SECRET = "CLIENT_SECRET"
+MUSICBRAINZ_CLIENT_ID = "SO_FPZ4wKjl3NRIrIYE0EEqU7W3_UBo2"
```
MonkeyDo, line 79: Ah, here we get to the point where "I'll just commit everything" becomes an issue :)

You always want to check exactly what you want to commit.

In this case, you modified the config *sample* file (you may also have modified the actual config.py file but changes to that one are ignored by git) and leaked your MB application credentials.

Not a biggie but we'll want to do two things:
1. revert this change (you can simply modify the file again to be "CLIENT_ID" and "CLIENT_SECRET", then commit that change)
2. you'll also want to delete your [MB application](https://musicbrainz.org/account/applications) and create a new one, that way the leaked credentials don't matter anymore

Aerozol:
Thanks Monkey! ‘Un-leaked’ (hah) my credentials, and lucifer helped me fix a conflict.

I made a related pull request updating the docs with some things that tripped me up (https://github.com/metabrainz/listenbrainz-server/pull/2185), here are my notes as I followed the instructions if you’re curious:
https://chatlogs.metabrainz.org/libera/metabrainz/2022-09-28/?msg=5061259&page=4

Edit: ahh now frontend tests have failed again >.>

PR #2151, opened 2022-09-19
Remove redundant LEFT JOINs from mb_metadata_cache query

amCap1712:
The logic behind removing these LEFT JOINs is that each individual CTE only needs to collect data for the recordings that have the respective data. The LEFT JOINs in the final select will ensure that all recordings all included in the final result even when some of the intermediate CTEs don't have any results for them. 

We only need LEFT JOINs in the intermediate CTEs when:
1. we are collecting more than one column of data
2. at least one of those columns can be NULL independent of the values in the other columns
3. we want that data should be collected even if not all but some columns are present.

See commit message for more details on which LEFT JOIN was removed why.


mayhem (approved):
thanks for this cleanup!

PR #2152, opened 2022-09-19
Fix order of artist credit mbids

amCap1712:
artist_credit_mbids should be order'ed by position column in the artist_credit_name table. The website otherwise cannot link to the first artist's mbid correctly.


mayhem (approved):
ohhh, that is how to do this. #ilearnedsomethingtoday

PR #2153, opened 2022-09-20
Allow recommendations with only mbids

amCap1712:
Currently, we require msid and optionally mbid for recommending tracks. But with mbid_mapping_metadata table, this restriction becomes arbitrary. As with feedback and pinned recordings, just one of msid or mbid is sufficient.


mayhem (approved):
Looks good, thanks!

PR #2154, opened 2022-09-20
Enable recommend track button if only recording mbid is present

amCap1712:
Currently, we require a msid for recommending tracks but this restriction is relaxed in #2153. Now that either mbid or msid is sufficient, update the conditions under which the button is rendered accordingly.


MonkeyDo (approved):
Good to go, thanks ! :rocket:

PR #2155, opened 2022-09-20
Incrementally update mb_metadata_cache

amCap1712:
The comment in the methods covers most of the details of the queries.

Need to figure out how to store/handle the value of timestamp - the mark above which changed values are retrieved, store last time incremental ran in redis or maybe add a last_updated column to mb_metadata_cache ?

Can also try to add a test for checking that incremental updates is not missing cases but crafting a comprehensive test dataset is hard.

### DIFF
--- a/listenbrainz/mbid_mapping/docker/crontab
+++ b/listenbrainz/mbid_mapping/docker/crontab
@@ -6,3 +6,10 @@
 
 # Rebuild the spotify metadata index every friday at 1 A.M.
 0 1 * * 5 listenbrainz /usr/local/bin/python /code/mapper/manage.py build-spotify-metadata-index >> /code/mapper/cron-spotify-metadata-index.log 2>&1
+
+
+# Build the mb metadata cache from scratch on first thursday of every month, MB dumps run on Wed and Sat so avoid those days
+0 15 * * 4 listenbrainz [ $(date +\%d) -le 7 ] && /usr/local/bin/python /code/mapper/manage.py cron-build-mb-metadata-cache >> /code/mapper/lb-cron.log 2>&1
+
+# Update the mb metadata cache incrementally every 4 hours
+0 */4 * * * listenbrainz /usr/local/bin/python /code/mapper/manage.py update-mb-metadata-cache >> /code/mapper/lb-cron.log 2>&1
--- a/listenbrainz/mbid_mapping/docker/crontab
+++ b/listenbrainz/mbid_mapping/docker/crontab
@@ -9,7 +9,7 @@
 
 
 # Build the mb metadata cache from scratch on first thursday of every month, MB dumps run on Wed and Sat so avoid those days
-0 15 * * 4 listenbrainz [ $(date +\%d) -le 7 ] && /usr/local/bin/python /code/mapper/manage.py cron-build-mb-metadata-cache >> /code/mapper/lb-cron.log 2>&1
+0 15 * * 4 listenbrainz [ $(date +\%d) -le 7 ] && /usr/local/bin/python /code/mapper/manage.py cron-build-mb-metadata-cache >> /code/mapper/cron-mb-metadata-cache.log 2>&1
 
 # Update the mb metadata cache incrementally every 4 hours
-0 */4 * * * listenbrainz /usr/local/bin/python /code/mapper/manage.py update-mb-metadata-cache >> /code/mapper/lb-cron.log 2>&1
+0 */4 * * * listenbrainz /usr/local/bin/python /code/mapper/manage.py update-mb-metadata-cache >> /code/mapper/cron-mb-metadata-cache.log 2>&1
--- a/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
+++ b/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
@@ -583,6 +583,7 @@ class MusicBrainzMetadataCache(BulkInsertTable):
 
         try:
             with self.mb_conn.cursor() as curs:
+                self.config_postgres_join_limit(curs)
                 recording_mbids = set()
 
                 log("mb metadata cache: querying recording mbids to update")
--- a/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
+++ b/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
@@ -599,7 +599,7 @@ class MusicBrainzMetadataCache(BulkInsertTable):
                   UPDATE {self.table_name}
                      SET dirty = 't'
                     FROM dirty_mbids
-                   WHERE {self.table_name}.artist_mbids @> dirty_mbids.artist_mbid::text
+                   WHERE {self.table_name}.artist_mbids::text[] @> dirty_mbids.artist_mbid::text
                 """
                 execute_values(curs, query, [(mbid,) for mbid in artist_mbids], page_size=len(artist_mbids))
 
--- a/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
+++ b/listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
@@ -715,6 +715,7 @@ def incremental_update_mb_metadata_cache(use_lb_conn: bool):
         log("mb metadata cache: starting incremental update")
 
         timestamp = select_metadata_cache_timestamp(lb_conn or mb_conn)
+        log(f"mb metadata cache: last update timestamp - {timestamp}")
         if not timestamp:
             return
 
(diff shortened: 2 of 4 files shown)


listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -488,3 +663,27 @@ def create_mb_metadata_cache(use_lb_conn: bool):
...
+            lb_conn = psycopg2.connect(config.SQLALCHEMY_TIMESCALE_URI)
+
+        cache = MusicBrainzMetadataCache(mb_conn, lb_conn)
+        if not cache.table_exists():
+            log("mb metadata cache: table does not exist, first create the table normally")
+            return
+
+        log("mb metadata cache: starting incremental update")
+
+        # TODO: Update logic to get last update timestamp
```
amCap1712, line 684: TODO: Decide the schema for a PG table to store the last update timestamp.

listenbrainz/mbid_mapping/docker/crontab
```diff
@@ -3,3 +3,10 @@
 
 # Run the huesound color sync hourly
 10 * * * * listenbrainz /usr/local/bin/python /code/mapper/manage.py update-coverart >> /code/mapper/lb-cron.log 2>&1
+
+
+# Build the mb metadata cache from scratch on first thursday of every month, MB dumps run on Wed and Sat so avoid those days
+0 15 * * 4 listenbrainz [ $(date +\%d) -le 7 ] && /usr/local/bin/python /code/mapper/manage.py cron-build-mb-metadata-cache >> /code/mapper/lb-cron.log 2>&1
+
+# Update the mb metadata cache incrementally every 4 hours
+0 */4 * * * listenbrainz /usr/local/bin/python /code/mapper/manage.py update-mb-metadata-cache >> /code/mapper/lb-cron.log 2>&1
```
alastair, line 12: should we have a way to stop this from running when the main cache is updating? I guess the swap to the new tables happens in a single transaction, but does it make sense to perform updates when the table is about to be deleted?

listenbrainz/mbid_mapping/docker/crontab
```diff
@@ -3,3 +3,10 @@
 
 # Run the huesound color sync hourly
 10 * * * * listenbrainz /usr/local/bin/python /code/mapper/manage.py update-coverart >> /code/mapper/lb-cron.log 2>&1
+
+
+# Build the mb metadata cache from scratch on first thursday of every month, MB dumps run on Wed and Sat so avoid those days
+0 15 * * 4 listenbrainz [ $(date +\%d) -le 7 ] && /usr/local/bin/python /code/mapper/manage.py cron-build-mb-metadata-cache >> /code/mapper/lb-cron.log 2>&1
```
alastair, line 9: Should we use the same log file for all cron jobs, or since we have an entire folder here, could we have a separate log file per process?

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -396,51 +399,223 @@ def config_postgres_join_limit(self, curs):
         curs.execute('SET from_collapse_limit = 15')
         curs.execute('SET join_collapse_limit = 15')
 
-    def mark_rows_as_dirty(self, recording_mbids: List[uuid.UUID], artist_mbids: List[uuid.UUID], release_mbids: List[uuid.UUID]):
-        """Mark rows as dirty if the row is for a given recording mbid or if it's by a given artist mbid, or is from a given release mbid"""
+    def query_last_updated_items(self, timestamp):
+        # there queries here try to mirror the structure and logic of the main cache building queries
```
alastair, line 403:
```suggestion
        # there queries mirror the structure and logic of the main cache building queries
```

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -396,51 +399,223 @@ def config_postgres_join_limit(self, curs):
...
+        # note that the tags queries in any of these omit the count > 0 clause because possible removal
+        # of a tag is also a change.
+        # the last_updated considered and last_updated ignored columns below together list all the last_updated
+        # columns a given CTE touches. any other tables touched by a given CTE do not have a last_updated column.
+
+        # further note that these queries only take updates and deletions into consideration, not deletes
```
alastair, line 409:
```suggestion
        # these queries only take updates and deletions into consideration, not deletes
```

alastair: something about _why_ not deletes? because the data is no longer in the db, so we're unable to identify this case?

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -396,51 +399,223 @@ def config_postgres_join_limit(self, curs):
...
+              FROM artist a
+              JOIN l_artist_url lau
+                ON lau.entity0 = a.id
+              JOIN url u
+                ON lau.entity1 = u.id
+              JOIN link l
+                ON lau.link = l.id
+              JOIN link_type lt
+                ON l.link_type = lt.id
+             WHERE lt.gid IN ('99429741-f3f6-484b-84f8-23af51991770'
```
alastair, line 434: is there any value to unify this list of linktypes in the update and create query?

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -396,51 +399,223 @@ def config_postgres_join_limit(self, curs):
...
+    def update_dirty_cache_items(self, recording_mbids: Set[uuid.UUID]):
         """Refresh any dirty items in the mb_metadata_cache table.
 
         This process first looks for all recording MIBDs which are dirty, gets updated metadata for them, and then
         in batches deletes the dirty rows and inserts the updated ones.
         """
-        dirty_query = f"""
-            SELECT recording_mbid
-              FROM {self.table_name}
-             WHERE dirty = 't'
```
alastair, line 428: does this mean that the dirty column is no longer used? Does it make sense to do this process in 2 steps - get mbids and mark as dirty, and then update?

listenbrainz/mbid_mapping/docker/crontab
amCap1712: I was thinking about that but figured it wouldn't be straightforward to do so. We'll have to add locks etc like we have in LB main cron. So inclined to going ahead with this for now. If some issue comes up, reevaluate.

listenbrainz/mbid_mapping/docker/crontab
amCap1712: Separate logs sound good.

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
amCap1712: Yes, dirty column is no longer used. We already have the list of recording mbids at hand so figured no need to mark dirty then query and update. I guess we could do that if we wanted to manually marks rows as dirty and want the update process to consider those as well?

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -396,51 +399,223 @@ def config_postgres_join_limit(self, curs):
...
-            with conn.cursor() as curs:
-                query = f"""UPDATE {self.table_name}
-                               SET dirty = 't'
-                             WHERE recording_mbid IN %s
-                                OR %s && artist_mbids
-                                OR release_mbid IN %s
-                        """
-                curs.execute(query, (tuple(recording_mbids), artist_mbids, tuple(release_mbids)))
-                conn.commit()
+            with self.mb_conn.cursor() as curs:
```
mayhem, line 581: I find it amazing to see the massive queries above, only to collect one column of data. :laughing:

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
mayhem: Why should it be different from the one we use everywhere else?

mayhem (approved):
Looks good to me, but I admit, I didn't check all the queries in great detail -- on the surface they all look sane.

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
amCap1712: Mostly because we want to create a single table and reuse it for all cases going forward instead of creating a new table for each use case.

listenbrainz/mbid_mapping/docker/crontab
amCap1712: Done.

listenbrainz/mbid_mapping/mapping/mb_metadata_cache.py
amCap1712: Done

alastair (approved):

PR #2156, opened 2022-09-21
LB-1126: show track duration properly on ListenCard

MonkeyDo:
Currently, a lot of listens are showing "0:00" as track duration on the ListenCard, because they are not converted to milliseconds.
When using the duration field (i.e. the duration_ms filed is not present), transform to milliseconds before formatting in human readable format.
Added two tests to check the formatted duration as well.


amCap1712 (approved):
LGTM, Thanks!

PR #2157, opened 2022-09-21
Add dumps for musicbrainz metadata tables

alastair:
# Problem

We generate a bunch of intermediate metadata files from musicbrainz that might be useful for other people, but they can't get the data

# Solution

Generate periodic data dumps of Canonical recording redirect and Canonical MB metadata.

I generated a new py file specifically for the metadata dumps because of some differences to the existing LB dumps:
- Compress with zstd instead of xz
- Because this is just a periodic one-off dump that doesn't need to be imported to a database, don't include a schema number file
- Different filename structure to show it's not an LB dump
- Add explicit filenames to the archive (e.g. `canonical_musicbrainz_data.csv`), not just the table name (such as `mapping.canonical_musicbrainz_data`)
- Dump in CSV format, not postgres default text format
- Remove the special-casing around different types of dumps (e.g. feedback, parquet, etc)

I tried to special-case all of these differences into the main dump file, but I didn't like the way that we added a whole bunch of additional `if` statements. It's possible that we could go back to @amCap1712's BU pull request with a set of helper classes for making dumps that we could subclass with additional settings, however I think this is a separate task.

Documentation for the contents of these dumps will be done as part of https://github.com/metabrainz/listenbrainz-server/pull/1996

This creates a dump with a structure like this:

```
$ tar tf metabrainz-metadata-dump-20220921-113421.tar.zst
metabrainz-metadata-dump-20220921-113421/TIMESTAMP
metabrainz-metadata-dump-20220921-113421/COPYING
metabrainz-metadata-dump-20220921-113421/metabrainz/canonical_musicbrainz_data.csv
metabrainz-metadata-dump-20220921-113421/metabrainz/canonical_recording_redirect.csv
```

At the moment, I'm writing both data files to the same archive. It's 1.4Gb compressed, and uncompressed:
```
4.8G	metabrainz-metadata-dump-20220921-113421/metabrainz/canonical_musicbrainz_data.csv
620M	metabrainz-metadata-dump-20220921-113421/metabrainz/canonical_recording_redirect.csv
```

If we decide that we want each data file to be in its own archive, we could do this so that users can download just the files that they want, though I don't see a huge difference in sizes. I'm open to feedback here

Pending tasks:
- [x] Write canonical release table
- [x] Dump parquet for users who want this in spark or other tools
- [ ] Table generation scripts can dump into MB or LB database connection, if we want to keep this, we should modify the dump script to also be able to dump from either.

# Action

- After releasing, we'll need to modify the upload scripts to include these dumps into FTP
- Update the [dataset page](https://metabrainz.org/datasets) on metabrainz.org to include more information about these

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ RUN apt-get update \
                        redis-tools \
                        rsync \
                        uuid \
+                       zstd \
     && rm -rf /var/lib/apt/lists/*
 
 # PostgreSQL client
--- a/docs/developers/mapping.rst
+++ b/docs/developers/mapping.rst
@@ -1,0 +1,76 @@
+:orphan:
+
+.. _developers-mapping:
+
+MBID Mapping
+============
+
+The MBID mapping scripts allow us to take metadata from the messybrainz database and look up recording MBIDs
+from the MusicBrainz database.
+
+.. note::
+    The MBID Mapping source code lives in ``listenbrainz/mbid_mapping`` but is run independently
+    from the main listenbrainz web docker image. You can use your own virtual environment or use
+    ``listenbrainz/mbid_mapping/build.sh`` to build a standalone docker image.
+
+Database tables
+^^^^^^^^^^^^^^^
+
+The MBID Mapping supplemental tables hold preprocessed data from the MusicBrainz database.
+
+* ``mapping.canonical_musicbrainz_data``: The MBID and Name of Recordings, Artists (and credits), and Releases for all recordings in MusicBrainz
+* ``mapping.canonical_recording_redirect``: A mapping to find the "canonical" recording given an artist credit + recording name
+* ``mapping.canonical_release_redirect``: A mapping to find the "canonical" release given an artist credit + release name
+
+These tables can be populated by running 
+
+.. code:: bash
+
+    python mapper/manage.py canonical-data
+
+The update process build the new data in a temporary table and then
+replaces them in a single transaction. This means that lookups can continue to run on the 
+existing tables while the new ones are being built.
+
+Fuzzy lookups
+^^^^^^^^^^^^^
+
+We use typesense as a way of performing quick, fuzzy lookups based on artist name and recording name
+
+Build the typesese index with
+
+.. code:: bash
+
+    python mapper/manage.py build-index
+
+As with the data tables, a new typesense collection is created and then swapped into place in a
+single operation.
+
+Build the mapping tables and then the typesense index directly afterwards with 
+
+.. code:: bash
+
+    python mapper/manage.py create-all
+
+MBID Mapper
+^^^^^^^^^^^
+
+The mapper looks for new MSIDs submitted to messybrainz and finds a matching MBID in MusicBrainz
+
+    ``python3 -u -m listenbrainz.mbid_mapping_writer.mbid_mapping_writer``
+
+A background thread pushes items to be processed onto a queue - recent submissions first, and then if nothing
+is to be done, old items.
+The processing thread pops items off the queue and then looks them up, adding them to the ``mbid_mapping`` table.
+
+There is also a background thread that fires off daily, which looks for listens that have been written to the 
+listens table, but for some reason do not have a matching mapping entry. (This could happen due to 
+restarts or problems with the mapper itself). These are called legacy listens.
+
+The background thread will walk the entire listens table once a day to find these legacy listens and attempt to
+map them. In the same thread we also look for mapping items with timestamp of the unix epoch (1970-01-01 00:00:00),
+which indicates that they ought to be re-checked. Currently we have no automated mechanism in place for setting any mapping 
+entries to the epoch.
+
+TODO: Detuning algorithm
+TODO: match quality types
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -35,6 +35,7 @@ Contents
    developers/spark-devel-env
    developers/architecture
    developers/spark-architecture
+   developers/mapping
    developers/commands
 
 .. toctree::
@@ -44,6 +45,7 @@ Contents
    maintainers/deploy
    maintainers/docker-image
    maintainers/dumps
+   maintainers/mapping
    maintainers/spotify-reader
    maintainers/updating-prod-db-schema
    maintainers/pull-requests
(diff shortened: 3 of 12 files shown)


amCap1712:
> Dump parquet for users who want this in spark or other tools
Almost all tools including spark support importing from CSV so I guess good enough for now or till someone specially asks?

alastair:
> Almost all tools including spark support importing from CSV so I guess good enough for now or till someone specially asks?

Yes, good point. This data isn't overly complex, so I think it does fit the structure of csv just fine. The only thing I'd consider doing for this is a quick evaluation of load time and file size to see if parquet gives us any wins there.

alastair:
I just double-checked. The canonical_musicbrainz_data table only uses core musicbrainz data, so the license of this dump (CC0) is correct.

If we also dump the recording metadata cache, this will include user data (tags), and so parts of it are attribution-nc-sa, we'll have to mark this clearly.

listenbrainz/mbid_mapping/mapping/dump.py
```diff
@@ -0,0 +1,219 @@
+""" This module contains data dump creation and import functions.
```
amCap1712, line 1: Adding this file here is fragile. 

The code inside `listenbrainz/mbid_mapping` runs in a separate docker image that only contains code inside `listenbrainz/mbid_mapping`. This is in contrast to all other containers which have all listenbrainz code in them. As a result, any import `listenbrainz` module inside `mbid_mapping` module will result in a `ModuleNotFoundError` at runtime.

While we know that this script is only run by dump_manager which is invoked through the normal listenbrainz containers, having this file here may give the wrong impression that it is runnable just inside the mbid_mapping container as well.  Further, if we import dump.py somewhere else inside the mbid_mapping module, it will probably to crash there.

Due to these reasons, I have been wondering for some time that `listenbrainz/mbid_mapping` should instead be a top level directory `mbid_mapping` to make this clearer.

amCap1712: There's also issues that `listenbrainz/mbid_mapping` has a different `requirements.txt` file from rest of LB. So many deps we use elsewhere in codebase are unable there.

listenbrainz/mbid_mapping/mapping/dump.py
```diff
@@ -0,0 +1,219 @@
...
+                raise
+            except Exception as e:
+                current_app.logger.error(
+                    'Exception while adding dump metadata: %s', str(e), exc_info=True)
+                raise
+
+            archive_tables_dir = os.path.join(temp_dir, 'metabrainz')
+            create_path(archive_tables_dir)
+
+            with db_engine.connect() as connection:
```
amCap1712, line 157: You can use `db.engine.raw_connection()` to get a psycopg2 connection directly. Also, can probably also remove the explicit transaction stuff.

listenbrainz/mbid_mapping/mapping/dump.py
```diff
@@ -0,0 +1,219 @@
...
+                            current_app.logger.error(
+                                'IOError while copying table %s', table, exc_info=True)
+                            raise
+                        except Exception as e:
+                            current_app.logger.error(
+                                'Error while copying table %s: %s', table, str(e), exc_info=True)
+                            raise
+                    transaction.rollback()
+
+            # Add the files to the archive in the order that they are defined in the dump definition.
```
amCap1712, line 178: Since no FKs or import concern apply here, probably best to simplify the comment to just `Add the files to the archive in the order that they are defined in the dump definition.`

listenbrainz/mbid_mapping/mapping/dump.py
```diff
@@ -0,0 +1,219 @@
...
+        Returns:
+            the path to the dumped data
+    """
+    current_app.logger.info('Beginning dump of PostgreSQL database...')
+    current_app.logger.info('dump path: %s', location)
+
+    current_app.logger.info('Creating dump of canonical metadata data...')
+    try:
+        dump_location = create_mapping_dump(location, dump_time)
+    except IOError as e:
```
amCap1712, line 88: Nit: I know we do this in normal dumps as well but I think multiple excepts are not useful because we do the exactly same thing and `exc_info` would record the exception info anyway.

listenbrainz/mbid_mapping/mapping/dump.py
```diff
@@ -0,0 +1,219 @@
...
+
+            try:
+                timestamp_path = os.path.join(temp_dir, "TIMESTAMP")
+                with open(timestamp_path, "w") as f:
+                    f.write(dump_time.isoformat(" "))
+                tar.add(timestamp_path,
+                        arcname=os.path.join(archive_name, "TIMESTAMP"))
+                tar.add(DUMP_LICENSE_FILE_PATH,
+                        arcname=os.path.join(archive_name, "COPYING"))
+            except IOError as e:
```
amCap1712, line 145: Nit: Same as above.

listenbrainz/mbid_mapping/mapping/dump.py
amCap1712: Ah ok, I checked why this was added to main LB dumps. It is to ensure that all tables as visible at a given point are dumped. I guess that makes for these dumps as well.

amCap1712 (approved):
Overall. LGTM Thanks! 

We should probably create 1 dump before merging to avoid unexpected surprises later.

listenbrainz/mbid_mapping/mapping/dump.py
alastair: moved to `listenbrainz/db/mapping_dump.py`

listenbrainz/mbid_mapping/mapping/dump.py
alastair: So you mean that it's best to keep the explicit transaction? In this case should we use raw_connection or stay with connect()?

alastair: we discussed in IRC, given that having a transaction is useful we wil just leave this as-is

amCap1712 (approved):
Overall changes look good, need to add a cronjob to produce the dumps and also updating the script to copy these over to FTP.

PR #2158, opened 2022-09-21
Rename follow-button class to prevent adblocking

alastair:
# Problem

ublock origin on chrome comes by default with a block list called "Fanboy's Annoyance List", which removes any element in a page with a class of .follow-button
Work around this by naming our class .lb-follow-button instead

# Solution

Rename the class


MonkeyDo (approved):
:rocket:

PR #2160, opened 2022-09-21
When the stars align…

MonkeyDo:
Not sure why this CSS was required, but some rules (specifically `top`) are overwritten by the ratings library (in the element's styles) and the stars looks like they're hovering over the placeholders:
image

After: 
image

Also modified license text which talks about a choice between two licenses, even though we don't offer a choice in this modal.


alastair:
you left stars in my eyes ⭐

PR #2161, opened 2022-09-22
Call mapping start/init method before starting consumer

alastair:
# Problem
Mapping writer wasn't starting up. @amCap1712 told me that he was running this change on a branch that hadn't been merged to master.

# Solution

# Action


amCap1712 (approved):
LGTM, Thanks!

PR #2162, opened 2022-09-22
LB-1113: Allow ratings on CB reviews to be optional

alastair:
# Problem

Even though the CB review writing popup says that rating is optional, submitting a text only review with no rating returned an error.

# Solution

By setting a default of 0 in LB, the CB api returned a validation error - the rating needs to be None, or 1-5
Make rating optional in LB validation

### DIFF
--- a/listenbrainz/db/model/review.py
+++ b/listenbrainz/db/model/review.py
@@ -1,8 +1,5 @@
-import sqlalchemy
+from typing import Optional
 from pydantic import BaseModel
-from typing import List
-
-from listenbrainz.db import timescale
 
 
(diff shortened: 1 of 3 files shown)


amCap1712 (approved):
LGTM, Thanks!

You can also omit the field entirely from the payload by changing
https://github.com/metabrainz/listenbrainz-server/blob/898c2bb548db20f767cdb9cd860c602c112b823e/listenbrainz/domain/critiquebrainz.py#L104 
to `review.dict(exclude_none=True)` but since CB works fine either way, upto you to update or not.

amCap1712 (approved):
LGTM, Thanks!

PR #2164, opened 2022-09-22
Turn sqlalchemy.exc.Base20DeprecationWarning into errors in tests

amCap1712:
The previous PRs in this series fixed a large number of Base20DeprecationWarning warnings. To avoid introducing more warnings between now and a future PR that will switch the future flag to turn these warnings into errors in prod, turn these warning into errors in tests.

BU and mbdata also need to be updated to avoid failing every test since the existing mbdata version init has a deprecated import.


amCap1712:
Depends on https://github.com/metabrainz/brainzutils-python/pull/95

mayhem (approved):

PR #2165, opened 2022-09-22
LB-1012: Missing cover art image links to release page if MBID available

MonkeyDo:
+ updated snapshots

As suggested in the ticket, if we have a release_mbid (either as part of the original listen or through the mbid_mapper) let's make the placeholder cover art image link to the release page's cover art tab so that users can add them easily.


amCap1712 (approved):
LGTM, Thanks

PR #2166, opened 2022-09-23
Use kombu in API

amCap1712:
We have already migrated all other places in LB to use kombu over pika. For APIs, errors will lead to 500 errors visible to the user. So we need to be more careful about this change.


amCap1712:
I have put this PR up on test/beta for testing but its likely issues will only appear in constant submission of listens. How about we put this up on spotify reader? If there is something wrong, it can automatically recover.

amCap1712:
I had put this on spotify_reader a couple of days ago and it has been working normally so far.

alastair (approved):
cool, let's do it!

PR #2167, opened 2022-09-23
Import collections as playlist

amCap1712:
TBA


amCap1712:
MusicBrainz collections now have a Play on ListenBrainz button which takes users to ListenBrainz and allows them to create a playlist there.

image

PR #2168, opened 2022-09-23
LB-1067: Bypass CAA redirect

MonkeyDo:
Firefox users are seeing security warnings about loading mixed content.
This is due to the CoverArtArchive image URLs being http rather than https or scheme-less.
Considering LB uses https we get a redirection when loading the images, resulting in the warning.

Instead of using the URLs from the CAA json response, we can constructing the URL to the underlying archive.org resource from the information in the json.
This is based on 9e40ad440d0b280b6c53d13e804f911657469c8b


amCap1712 (approved):
LGTM, Thanks!

PR #2169, opened 2022-09-23
LB-1139: Retry CAA calls after 429 (rate-limit exceeded)

MonkeyDo:
When a whole page of non-spotify* listens is loaded, we call the CoverArtArchive for each listen on that page.
There is a gateway-level rate-limit applied and consequently a lot of these requests will result in a status 429 (too many requests).

Retrying fetch calls being an issue in multiple places, I've installed a library ([fetch-retry](https://github.com/jonbern/fetch-retry)) that very easily implements customizable retry strategies and set it up here for the CAA calls.

*for the spotify listens we fetch the cover art from spotify's API directly


amCap1712 (approved):
LGTM, Thanks!

PR #2170, opened 2022-09-23
LB-1020: Remove 12-hour formatting from listencard timestamps

MonkeyDo:
We should trust toLocaleString to do the right thing depending on the user's locale


MonkeyDo:
@alastair Since you opened the ticket Ill let you test and see if it fixes the issue you saw. I do believe so from my tests, and this also makes a lot more sense for my french locale (AM/PM not really required, it's usually clearer to have 24h format)

PR #2175, opened 2022-09-24
LB-1123: Show total artist, recording, release count 

amCap1712:
We limit the actual stats to top 1000 to conserve resources but the total overall counts will now be accurate.

Also, make counts accurate in case entries are discarded during validation.


mayhem (approved):
This PR kept repeating itself over and over again. :) :)

PR #2177, opened 2022-09-25
Adds playlists management page

regorxxx:
Adds new page to navbar (data menu) for playlists management. Includes a link to [foobar2000 playlist manager ](https://github.com/regorxxx/Playlist-Manager-SMP)to sync playlists on ListenBrainz with local playlists (with content resolution) and Spotify [third party service connection](https://listenbrainz.org/profile/music-services/details/). Also a note about advanced usage related to[ content resolution following XSPF specs.](https://xspf.org/spec#34-content-resolver)

Rationale: if ListenBrainz is gonna make use of playlists (either user made, collaborative or bots), some kind of software support will be required... and it would be great to encourage it with its own page. This would be the first step.

Structure is mostly a clone of 'Adding data' tab.

### DIFF
--- a/listenbrainz/webserver/templates/index/playlists-management.html
+++ b/listenbrainz/webserver/templates/index/playlists-management.html
@@ -1,0 +1,34 @@
+{%- extends 'base.html' -%}
+{%- block title -%}Playlists management - ListenBrainz{%- endblock -%}
+{%- block content -%}
+  <h2 class="page-title">Playlists managementg </h2>
+  {# Playlists management information #}
+  <h3>Synchronizing Playlists</h3>
+  <p>
+    There are many ways to sync your own playlists to ListenBrainz:
+  </p>
+  <h4>Music players</h4>
+  <ul>
+    <li><em><a href="https://www.foobar2000.org/">Foobar2000</a></em>, full-fledged audio player for Windows: <a href="https://github.com/regorxxx/Playlist-Manager-SMP"><code>Playlist-Manager-SMP</code></a></li>
+  </ul>
+
+  <h3>Synchronizing via third-party music services</h3>
+  <p>
+    <a href={{ url_for('profile.music_services_details') }}>Connect your ListenBrainz account</a> and import/export your playlists to some third-party services.
+  </p>
+  <ul>
+    <li><em><a href="https://open.spotify.com/">Spotify</a></em></li>
+  </ul>
+  <p>
+    
+  </p>
+
+  <h3>For advanced users</h3>
+  <p>
+    Developers are able to import/export and edit their playlists on Listenbrainz using the Listenbrainz API. Information on how to do this
+    can be found in the <a href="https://listenbrainz.readthedocs.io">API docs</a>.
+  </p>
+  <p>
+	Note playlist importing requires track MBID identification for every item; exporting requires some kind of content resolution of tracks by track MBID, title and/or artist, following <a href="https://xspf.org/spec#34-content-resolver">JSPF or XSPF specs</a>.
+  </p>
+{%- endblock -%}
--- a/listenbrainz/webserver/templates/navbar.html
+++ b/listenbrainz/webserver/templates/navbar.html
@@ -32,6 +32,7 @@
             <li><a href="{{ url_for('index.add_data_info') }}">Adding data</a></li>
             <li><a href="{{ url_for('index.import_data_info') }}">Importing data</a></li>
             <li><a href="{{ url_for('index.data') }}">Using our data</a></li>
+            <li><a href="{{ url_for('index.playlists_management_info') }}">Playlists management</a></li>
           </ul>
         </li>
         <li class="dropdown">
--- a/listenbrainz/webserver/views/index.py
+++ b/listenbrainz/webserver/views/index.py
@@ -80,6 +80,11 @@ def import_data_info():
     return render_template("index/import-data.html")
 
 
+@index_bp.route("/playlists-management/")
+def playlists_management_info():
+    return render_template("index/playlists-management.html")
+
+
 @index_bp.route("/lastfm-proxy/")
 def proxy():
     return render_template("index/lastfm-proxy.html")
--- a/listenbrainz/webserver/views/test/test_index.py
+++ b/listenbrainz/webserver/views/test/test_index.py
@@ -42,6 +42,10 @@ class IndexViewsTestCase(IntegrationTestCase):
         resp = self.client.get(url_for('index.import_data_info'))
         self.assert200(resp)
 
+    def test_playlists_management_info(self):
+        resp = self.client.get(url_for('index.playlists_management_info'))
+        self.assert200(resp)
+
     def test_404(self):
         resp = self.client.get('/canyoufindthis')
         self.assert404(resp)


listenbrainz/webserver/templates/index/playlists-management.html
```diff
@@ -0,0 +1,34 @@
+{%- extends 'base.html' -%}
+{%- block title -%}Playlists management - ListenBrainz{%- endblock -%}
+{%- block content -%}
+  <h2 class="page-title">Playlists managementg </h2>
```
MonkeyDo, line 4:
```suggestion
  <h2 class="page-title">Playlist management </h2>
```

This might end up renamed as "Importing playlists", but thought I'd comment in any case.

listenbrainz/webserver/templates/index/playlists-management.html
```diff
@@ -0,0 +1,34 @@
...
+  <ul>
+    <li><em><a href="https://www.foobar2000.org/">Foobar2000</a></em>, full-fledged audio player for Windows: <a href="https://github.com/regorxxx/Playlist-Manager-SMP"><code>Playlist-Manager-SMP</code></a></li>
+  </ul>
+
+  <h3>Synchronizing via third-party music services</h3>
+  <p>
+    <a href={{ url_for('profile.music_services_details') }}>Connect your ListenBrainz account</a> and import/export your playlists to some third-party services.
+  </p>
+  <ul>
+    <li><em><a href="https://open.spotify.com/">Spotify</a></em></li>
```
MonkeyDo, line 20: I think here what would be more useful would be third-party services mentioned above that allow you to export and/or transform playlists.

Looking around quickly for OSS I found  https://github.com/watsonbox/exportify, as well as some non-open-source options like https://www.tunemymusic.com and https://freeyourmusic.com that can export to CSV

There are also tools that allow to transform other formats to XSPF such as https://github.com/kevlened/csvToXspf , and there must be good OSS alternatives out there to do the same for M3U playlists and other standard playlist formats

listenbrainz/webserver/templates/index/playlists-management.html
```diff
@@ -0,0 +1,34 @@
...
+    There are many ways to sync your own playlists to ListenBrainz:
+  </p>
+  <h4>Music players</h4>
+  <ul>
+    <li><em><a href="https://www.foobar2000.org/">Foobar2000</a></em>, full-fledged audio player for Windows: <a href="https://github.com/regorxxx/Playlist-Manager-SMP"><code>Playlist-Manager-SMP</code></a></li>
+  </ul>
+
+  <h3>Synchronizing via third-party music services</h3>
+  <p>
+    <a href={{ url_for('profile.music_services_details') }}>Connect your ListenBrainz account</a> and import/export your playlists to some third-party services.
```
MonkeyDo, line 17: We are likely to soon add an option  to import playlists from Spotify, but in the meantime this is a good start

listenbrainz/webserver/templates/index/playlists-management.html
```diff
@@ -0,0 +1,34 @@
...
+    
+  </p>
+
+  <h3>For advanced users</h3>
+  <p>
+    Developers are able to import/export and edit their playlists on Listenbrainz using the Listenbrainz API. Information on how to do this
+    can be found in the <a href="https://listenbrainz.readthedocs.io">API docs</a>.
+  </p>
+  <p>
+	Note playlist importing requires track MBID identification for every item; exporting requires some kind of content resolution of tracks by track MBID, title and/or artist, following <a href="https://xspf.org/spec#34-content-resolver">JSPF or XSPF specs</a>.
```
MonkeyDo, line 32: We can link to our tool we use for that purpose: https://labs.api.listenbrainz.org/mbid-mapping which can also be called as an API

MonkeyDo (commented):
Thanks for opening this PR !

As a main comment, I agree with the need for a section about importing playlists, but I think it might find a better home in the "Import data" page which is currently underused, and seems to me like a perfect fit.

It can be reworked a bit to add a new section "Importing playlists".
What do you think?

listenbrainz/webserver/templates/index/playlists-management.html
regorxxx: Although I agree it would be great to at least link to services like those (specially to give people tools to make use of their spotify data for open source services), there is a caveat: current problem of ListenBrainz is not being able to export the JSPF playlists in an user-friendly way.

regorxxx: I mean, even if I add more info about what to do with Spotify playlists, currently there is only one software example being  able to export ListenBrainz playlists to working files. And that's the weak part.

listenbrainz/webserver/templates/index/playlists-management.html
regorxxx: Thanks! Didn't knew about it... in my foobar plugin the use-case was sending a playlist to the server with tracks already tagged with 'MUSICBRAINZ_TRACKID' so it was not required. But I will make good use of that to cover sending playlists with non tagged files.

What I found most needed **was an api to link MBIDs to youtube urls.** i.e. if your export a JSPF playlist for local use, it would be great to have the **youtube link found on the ListenBrainz player as default location source.** Right now you need to use content resolution to find matches with local tracks, but for missing items, those tracks would be simply skipped (while there is software out there able to play youtube URLs). Foobar2000 is an example.

MonkeyDo: In the ListenBrainz player we do content resolution on the spot too with the youtube search API.
But we are indeed working on better content resolution, initially to get Spotify IDs for a given recording MBID with a goal to extend it to other sources.
We want it for the same reasons: content resolution on the fly is a bit uneven, and some search APIs leave to be desired, so we want to serve content URLs along with the rest of the metadata.
But we're not quite there yet.

listenbrainz/webserver/templates/index/playlists-management.html
MonkeyDo: I'll create a ticket for exporting playlists in an easier manner

MonkeyDo:
> > Thanks for opening this PR !
> > As a main comment, I agree with the need for a section about importing playlists, but I think it might find a better home in the "Import data" page which is currently underused, and seems to me like a perfect fit.
> > It can be reworked a bit to add a new section "Importing playlists". What do you think?
> 
> Makes sense to me too. Although note playlists should be considered data meant to be imported and also exported, so maybe a rename is required in that page. Something like "Managing data" for the page name.
> 
> Then I could rework the current importing data part and add the third party services part there as you note.
> 
> Finally add the playlists section too.

I see your point about exporting playlists; we do already have an "export" page but accessible through the user dropdown menu: https://listenbrainz.org/profile/export/
We should add the ability to export all playlists in that export page (created a ticket for that too: [LB-1144](https://tickets.metabrainz.org/browse/LB-1144))

I wonder if for simplicity's sake we should focus on importing only on this page, but also add a subsection at the bottom with a link to the export page as well.

listenbrainz/webserver/templates/index/playlists-management.html
regorxxx: **Btw, note I already coded a JSPF/XSPF library in javascript, which may be reused by the team with minimal changes.**

XSPF parser to JSPF
https://github.com/regorxxx/Playlist-Manager-SMP/blob/main/helpers-external/xspf-to-jspf-parser/xspf_parser.js
JSPF parser to XSPF
https://github.com/regorxxx/Playlist-Manager-SMP/blob/main/helpers/helpers_xxx_playlists_files_xspf.js

Only change required on your side would be merging both files, and scrapping the "include" part.

Also when exporting to XSPF or JSPF, it would be great to add as location the youtube link found on the ListenBrainz player, as commented below. Not sure if at least the current youtube content resolution could be used when exporting, since it's already coded. It could mean hundreds of queries at once though.

MonkeyDo: Love the *SPF libraries, that could indeed come in very handy :)

If I'm honest, I'm not at all impressed by the youtube search results when used in the context of the ListenBrainz player. Since the youtube API tries to return a result, any result rather than none, we often get results that are quite trash. I wouldn't rely on it for a good playlist…
However when we do have a better form of content resolution (we're keen on this for obvious reasons) we'll be sure to include the track locations in the playlists.

MonkeyDo:
Hi @regorxxx !
Just wanted to check in, I' not eactly sure where we left this off.Are there any questions

PR #2181, opened 2022-09-27
Add Fresh Releases frontend

chinmaykunkikar:
Fresh Releases page is at its final stages of completion ✨

Things that require attention are -
1. **Tests** - **Edit**: Done.

2. The **timeline slider logic** is still a bit buggy. My take on possible bug and it's fix -> Currently the `marks` are created based on the percent value of `releases per day/total releases`. This creates marks from 0%-100%. So the slider behaves almost exactly like the browser window scrollbar. The percent value at the end of the slider will scroll past the releases into the footer of the page. We want the slider marks to start from where the card grid starts and end where the grid ends. The start/end of the slider should have a percent value where the grid starts/ends.
To explain it with an example -
Suppose `offsetTop` of grid is 2% of the whole page, and the footer is 8% of the page, then the slider range should be between 2% (start) and 92% (end).
Also, another improvement can be that if the end of grid is at the bottom of the screen, it should stop scrolling there and never show footer.
Does my approach to the problem make sense here, @MonkeyDo?

3. **BrainzPlayer integration** - **Edit**: Will be a new PR

This completes Fresh Releases.


MonkeyDo:
So I did try my suggestion for scrolling the page (directly in browser, just for testing) and I think it'll work nicely

With the slider's min and max values being 0 and 100, we get a percentage value which is convenient for scrolling the page.
I use onAfterChange which only fires after the user stops interacting with the slider, rather than firing on every move.
```
onAfterChange = (percent)=> {
const pageScrollHeight = document.documentElement.scrollHeight;
// Invert percentage, as slider's top (100%) should equal scrolling to 0
const scrollYPx = pageScrollHeight- (percent / 100) * pageScrollHeight;
window.scrollTo(0, Math.round(scrollYPx));
}
```

Then to be make the slider look super nice we can calculate the number of consecutive days we have to display, and set the `step` prop to  a rounded 100/numberOfDays and activate the `dots` feature.
This will show a dot on the slider for each day we have.

Let me know if you need some help with all of this !

chinmaykunkikar:
@MonkeyDo I tried going this way. One concern with this approach is that we cannot be sure about the number of releases per day i.e we cannot divide the slider into equal parts based on the number of days. A particular day may see only a couple releases while a weekend may see hundreds.

listenbrainz/webserver/static/js/src/fresh-releases/ReleaseTimeline.tsx
```diff
@@ -31,14 +31,36 @@ export default function ReleaseTimeline(props: ReleaseTimelineProps) {
...
+    const datesArr = Object.keys(releasesPerDate).map((item) =>
+      formattedReleaseDate(item)
+    );
+    const percentArr = Object.values(releasesPerDate)
+      .map((item) => Math.floor((item / data.length) * 100))
+      .map((_, index, arr) =>
+        arr.slice(0, index + 1).reduce((prev, curr) => prev + curr)
+      );
+    percentArr.unshift(0);
+    percentArr.pop();
```
chinmaykunkikar, line 45: `unshift` and `pop` explaination - we want the percentage values to start at 0 to get the first date at the start of the slider where the respective releases are. And then we wan to remove the last value.

listenbrainz/webserver/static/css/fresh-releases.less
```diff
@@ -337,25 +336,49 @@
         }
       }
     }
+
+    @media (min-width: 1200px) {
+      grid-template-columns: repeat(5, 1fr);
```
chinmaykunkikar, line 341: Large screens can have more columns

listenbrainz/webserver/static/css/rc-slider.less
```diff
@@ -0,0 +1,298 @@
...
+  position: absolute;
+  height: 6px;
+  background-color: @background_color_2;
+  border-radius: 6px;
+}
+.rc-slider-handle {
+  position: absolute;
+  width: 3rem;
+  height: 3rem;
+  margin-top: -12px;
```
MonkeyDo, line 74: I would make this quite a bit smaller (at least on desktop sizes):
```suggestion
  width: 2rem;
  height: 2rem;
  margin-top: -7px;
```

listenbrainz/webserver/static/js/src/fresh-releases/ReleaseCard.tsx
```diff
@@ -0,0 +1,105 @@
...
+  async function getCoverArt() {
+    const coverartURL = await getAlbumArtFromReleaseMBID(releaseMBID);
+    if (coverartURL) {
+      setCoverartSrc(coverartURL);
+    }
+  }
+
+  React.useEffect(() => {
+    getCoverArt();
+  }, []);
```
MonkeyDo, line 69: I think this should be the same pattern you found for async calls with useEffect, and passing the required dependencies:

```suggestion
  React.useEffect(() => {
    async function getCoverArt() {
      const coverartURL = await getAlbumArtFromReleaseMBID(releaseMBID);
      if (coverartURL) {
        setCoverartSrc(coverartURL);
      }
    }
    getCoverArt();
  }, [releaseMBID,setCoverartSrc]);
```

listenbrainz/webserver/static/js/src/fresh-releases/ReleaseTimeline.tsx
MonkeyDo: Might be good to move this comment from Github to inside the code.

Although if I'm honest I'm still not clear on why we're removing the first and last elements. Is that why there is no date for the end of the slider?

listenbrainz/webserver/static/js/src/fresh-releases/ReleaseTimeline.tsx
```diff
@@ -0,0 +1,87 @@
...
+    // Scroll to the current date
+    setCurrentValue(middle);
+    window.scrollTo({ top: changeHandler(middle), behavior: "smooth" });
+
+    return zipObject(percentArr, datesArr);
+  }
+
+  const handleScroll = React.useCallback(
+    debounce(() => {
+      // TODO change to relative position of #release-cards-grid instead of window
```
MonkeyDo, line 55: Todo todo?
(todo todo-todo-todo-todo-todooooooo  todododo)
https://images.squarespace-cdn.com/content/v1/566a4af357eb8d3974390587/1454570174741-4Y47MU9KC0IEMF87J7JE/ke17ZwdGBToddI8pDm48kHnqq_UrsdjXT8IYn8eSSjYUqsxRUqqbr1mOJYKfIPR7LoDQ9mXPOjoJoqy81S2I8PaoYXhp6HxIwZIk7-Mi3Tsic-L2IOPH3Dwrhl-Ne3Z2qO_jwV24K38aH_dPXd7Kl7ikppHhGne0pYe9e5QogUDdZR9z9mxWb0yLUToVqwSd/PinkPantherPinkSax-LtdEdCel.jpg

listenbrainz/webserver/static/js/src/fresh-releases/utils.tsx
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from "react";
+
+export function formattedReleaseDate(releaseDate: string) {
```
MonkeyDo, line 3: I'm more used to action verbs for functions that perform an action:
```suggestion
export function formatReleaseDate(releaseDate: string) {
```

listenbrainz/webserver/static/js/src/utils/APIService.ts
```diff
@@ -1172,4 +1172,26 @@ export default class APIService {
...
     return response.json();
   };
+
+  fetchSitewideFreshReleases = async (
+    release_date?: string,
+    days?: number
+  ): Promise<any> => {
+    let url = `${this.APIBaseURI}/explore/fresh-releases/`;
+    if (!isUndefined(release_date)) url += `?release_date=${release_date}`;
+    if (!isUndefined(days)) url += `&days=${days}`;
```
MonkeyDo, line 1182: Not sure this is very safe; if release_date is undefined but not days, you end up with a malformed url `/explore/fresh-releases/&days=123` without  the required `?`.

Have a look at the implementation of query parameters in `getListensForUser` for example.

MonkeyDo (commented):
Cool, we're down to small details now, that's great !
Some comments on implementation while I figure out the tests part (question # 1) :)
This includes a solution for your question # 2.

As for question # 3 (BrainzPlayer integration), I think it should be implemented in another PR regardless of what solution we go for.
For what it's worth, IMO the solution would be to redirect users to a playable album page on LB. For example https://listenbrainz.org/player/release/f5d6d909-06dc-4811-8e13-811e6af31b82/
Currently this would navigate away from the page which might not be too problematic, and in the future when the front-end is a SPA with React Router, that will be even less of a problem.

listenbrainz/webserver/static/js/tests/fresh-releases/FreshReleases.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const props = {
+  ...freshReleasesProps,
+  newAlert: () => { },
+};
+
+// Create a new instance of GlobalAppContext
+const mountOptions: { context: GlobalAppContextT } = {
+  context: {
+    APIService: new APIService("foo"),
```
MonkeyDo, line 34: I'm not clear on the differences for testing funcional components vs classes.
Here's something you could try if you wanted to mock some API methods (not sure it'll work but…):
Along the lines of
```

const fakeAPIService = new APIService("foo");
const mockFetchSitewideFreshReleases = jest.fn()
fakeAPIService.fetchSitewideFreshReleases = mockFetchSitewideFreshReleases
const wrapper = mount(
  <GlobalAppContext.Provider value={{...mountOptions.context, APIService: fakeAPIService}}>
    <FreshReleases {...props} />
  </GlobalAppContext.Provider>
);
expect(mockFetchSitewideFreshReleases).toHaveBeenCalled()
```

listenbrainz/webserver/static/css/rc-slider.less
```diff
@@ -0,0 +1,298 @@
...
+  }
+}
+.rc-slider-mark {
+  position: absolute;
+  top: 18px;
+  left: 0;
+  width: 100%;
+  font-size: 12px;
+}
+.rc-slider-mark-text {
```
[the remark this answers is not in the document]
chinmaykunkikar: I like the css transform version better than the odd/even approach.

Now that we're on this topic I have this whole idea for marks and the handle logic. It may take some time to figure it out. The idea goes like this -
Have the marks placed equidistantly no matter what how many releases are per day. The handle will scroll to the positon where the marks are while actually the grid is getting scrolled to the correct position i.e. the respective date. This way the timeline looks more polished and easier to navigate to the wanted date. As the main purpose of having the timeline is to have something like a map legend to make it easily navigable between dates. The placement of dates doesn't matter as long as they are there (in a visually appealing way). We can leverage the `steps` property of rc-slider but we'll have to figure out the logic.

But for now I'll go with css transform to declutter them.

listenbrainz/webserver/static/css/rc-slider.less
chinmaykunkikar: Was wondering the same few days ago. Will be done :)

chinmaykunkikar:
> As for question # 3 (BrainzPlayer integration), I think it should be implemented in another PR regardless of what solution we go for. 
Sure!

> For what it's worth, IMO the solution would be to redirect users to a playable album page on LB. For example https://listenbrainz.org/player/release/f5d6d909-06dc-4811-8e13-811e6af31b82/ 

I didn't know such album pages existed! Yeah, it will be better experience with a router.

listenbrainz/webserver/static/js/src/fresh-releases/utils.tsx
chinmaykunkikar: Good catch :D I like having properly worded function and variable names in the code too!

listenbrainz/webserver/static/js/src/fresh-releases/ReleaseTimeline.tsx
chinmaykunkikar: Hahaha 🤣 

It's for the scroll logic problem that I mentioned in the description.

chinmaykunkikar: Logic behind the `unshift(0)` and `pop()` - 
It may look hacky.
And yes, that's the reason there is no date at the end of the slider but there is one at the start (because of `unshift(0)`).

If we don't do that, the timeline ends up looking like this - 
image

We want the timeline dates or marks to start where the grid starts. So the 0% should always have the first date. With the same logic, we don't want the last date to be at 100% because that will mean we're at the bottom of the grid. The last date should start before 100%. That explains the `pop()`.

listenbrainz/webserver/static/js/tests/fresh-releases/FreshReleases.test.tsx
chinmaykunkikar: Will try it out

listenbrainz/webserver/static/js/src/utils/APIService.ts
chinmaykunkikar: Actually the request will go as `/explore/fresh-releases/?release_date=&days=123` (as it currently does).

But I agree that it's not the ideal query params implementation. I'll check out the `getListensForUser`.

chinmaykunkikar: Update: url query params are safe now :)

listenbrainz/webserver/static/css/rc-slider.less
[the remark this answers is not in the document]
chinmaykunkikar: Update: Rotated those marks!

listenbrainz/webserver/static/js/tests/fresh-releases/FreshReleases.test.tsx
chinmaykunkikar: Update: It works. I'll try to take it forward from here :)

MonkeyDo: Phew! Great!

listenbrainz/webserver/static/js/src/fresh-releases/ReleaseTimeline.tsx
MonkeyDo: Yeah, I see it now, thanks for the explanation. A comment would be useful for posterity.

chinmaykunkikar: Done

listenbrainz/webserver/static/js/src/fresh-releases/ReleaseTimeline.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const screenMd = useMediaQuery("(max-width: 992px)"); // @screen-md
+
+  const changeHandler = React.useCallback((percent: number | number[]) => {
+    setCurrentValue(percent);
+    const element: HTMLElement | null = document.getElementById(
+      "release-cards-grid"
+    )!;
+    const scrollHeight = ((percent as number) / 100) * element.scrollHeight;
+    const scrollTo = scrollHeight + element.offsetTop;
+    window.scrollTo({ top: scrollTo, behavior: "smooth" });
```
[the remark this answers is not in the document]
chinmaykunkikar: @MonkeyDo: I'm finally onto this one. I didn't quite understand your idea of `handleScroll`  reverse adjustments and use of `ref`.

amCap1712:
IIUC, this PR is ready to be merged? If so, let's merge and deploy.

mayhem:
oh please oh please!

amCap1712:
Thanks @chinmaykunkikar for this work and @monkey for reviewing and all the help!

PR #2183, opened 2022-09-27
Run troi bot hourly

amCap1712:
Instead of running troi bot after recommendation are generated, run troi bot hourly. Every check which for which followers the new day has just started and generate daily jams for them. If we do not know the timezone of a user assume it to be GMT.

Note that daily jams patch currently does not accept a date or timezone parameter so the date in it can be incorrect for the time being.

### DIFF
--- a/listenbrainz/troi/troi_bot.py
+++ b/listenbrainz/troi/troi_bot.py
@@ -36,7 +36,7 @@ def get_users_for_daily_jams():
     query = """
         SELECT "user".musicbrainz_id AS musicbrainz_id
              , "user".id as id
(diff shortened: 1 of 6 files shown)


amCap1712:
Depends on https://github.com/metabrainz/troi-recommendation-playground/pull/67

listenbrainz/troi/troi_bot.py
```diff
@@ -28,6 +28,25 @@ def run_post_recommendation_troi_bot():
         # Add others here
 
 
+def get_users_for_playlist():
```
mayhem, line 31: this function needs a better name too.

mayhem (approved):
A couple of minor improvements and good to merge.

PR #2184, opened 2022-09-28
Make develop.sh and test.sh compatible with compose v2

amCap1712:
The command to execute docker compose in v2 is `docker compose` as opposed to `docker-compose` in v1. I recently upgraded versions on my local machine and faced errors while trying to execute the scripts. Since, no else has complained so far they are likely using v1. Therefore, updating the scripts to check the version of docker compose installed and invoke the appropriate command for compatibility with both versions.

Note that there are some differences between the two commands but AFAIU those are not relevant to our uses.

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -8,8 +8,16 @@ if [[ ! -d "docker" ]]; then
     exit -1
 fi
 
+echo "Checking docker compose version"
+if docker compose version &> /dev/null; then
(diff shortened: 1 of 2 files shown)


amCap1712:
I tested locally where I have v2 and on wolf where we have v1 and the scripts work in both cases.

PR #2185, opened 2022-09-28
Minor fool-proofing

Aerozol:
# Problem

Expanded a few things that tripped me up

# Solution

- Mentioning that you can clone into your own fork/branch, it doesn’t have to be the master url specified
- Explicitly stating not to touch config.py.sample


docs/developers/devel-env.rst
```diff
@@ -69,6 +69,7 @@ configuration.
 
     cp listenbrainz/config.py.sample listenbrainz/config.py
 
+Don't change anything in config.py.sample.
```
MonkeyDo, line 72: We can even lay it thicker I think:
```suggestion
This will create a config.py file. Don't change anything in config.py.sample: as its name indicates it is only a sample file.
```

And the following sentence which tripped you up initially:
```suggestion
Next, open the new config.py file with your favorite text editor
```

MonkeyDo (commented):
Thanks for making instructions better 🥇

docs/developers/devel-env.rst
Aerozol: I actually like the brief instructions without any fluff, I don’t think it needs that extra line. What tripped me up is that ‘sample’ is mentioned earlier in that part:
“If this is your first time configuring ListenBrainz, copy the sample to a live configuration.”
I thought changing both would be a safe bet in case the rest of the section was talking about ‘sample’ stuff.

How about ’Now, open the new config.py file (don’t change config.py.sample) with your favorite text editor’?

’New’ implies that the config.py file was created by the previous step without having to go into it too much.

MonkeyDo: Yes, I think that will work absolutely fine

Aerozol:
Thanks Monkey, updated!

MonkeyDo (approved):
🚀

docs/developers/devel-env.rst
```diff
@@ -19,7 +19,7 @@ Clone listenbrainz-server
 -------------------------
 
 ListenBrainz is hosted on GitHub at https://github.com/metabrainz/listenbrainz-server/.
-You can use ``git`` to clone it to your computer
+You can use ``git`` to clone it (or your own fork/branch) to your computer
```
alastair, line 22:
```suggestion
You can use ``git`` to clone it (or your own fork) to your computer
```
Just a small technical correction here - you are cloning a fork, but not a branch

alastair (changes requested):
one teeny tiny correction but otherwise looks good to me!

alastair (approved):

PR #2189, opened 2022-09-30
Tweak CAA retry strategy

MonkeyDo:
After some investigation of the CAA 429 responses, I decided to add a bit more delay between retries as well as an exponential backoff, in order not to hammer the CAA servers if there are already issues on the server (the 429s were caused by hitting a global limit of number of requests per second, so more retries = more 429s)


amCap1712 (approved):
LGTM, Thanks!

PR #2190, opened 2022-10-03
Update Typescript to v4.x

MonkeyDo:
Paying some technical debt here, with lots of packages that will soon require TS > 4.

We still have some clean up to do afterwards after a change in the way errors in catch block are typed, but that will touch a lot of files so I though it would be better to keep it separate
See the addition of [useUnknownInCatchVariables](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables) option to tsconfig file

Many thanks to @chinmaykunkikar who did most of the work here in #2059


chinmaykunkikar (commented):
Ah, just as I suspected where all the eslint checks were failing. All eslint and prettier related packages should only be upgraded to latest versions after getting rid of the `fork-ts-checker-webpack-plugin` (see the 3rd point in #2059's description). This is because there are major updates rule changes to these plugins. See if you can only upgrade to TS4.8 without updating any eslint packages.
If not, we'll have to do all the typescript, fork-ts-checker-webpack-plugin, eslint, and prettier related work in a single PR.

PR #2191, opened 2022-10-03
Update nivo graph library

MonkeyDo:
In our ongoing saga of paying technical debt to get to React 18, we need to update our graph library nivo (composed of multiple packages). See #2059 for details.
Even though it is a point release, the nivo library is < v1 so technically point releases can have breaking changes, as is the case here.
A few APIs have changed, typescript types are cleaner.

Thanks to @chinmaykunkikar for providing the basis for this PR.


amCap1712:
I put this on beta for testing. Works fine except this particular tooltip (first is prod and second is beta, this PR):
Screenshot from 2022-10-04 23-02-29
Screenshot from 2022-10-04 23-02-48

MonkeyDo:
One last issue to fix with the tooltip on the BarDualTone component:
image

MonkeyDo:
Caught another issue with tooltip on the charts page, now fixed:
image

PR #2192, opened 2022-10-04
Remove configuration options for websocket server

alastair:
# Problem

This was unused since we moved the production websockets server to a sub-url of the main listenbrainz.org domain instead of a subdomain.

# Solution
Delete it

# Action

Websockets still don't work on local development, we will fix this later


alastair:
force-pushed to remove `webSocketsServerUrl` from frontend code too

amCap1712 (approved):
LGTM, thanks.

PR #2195, opened 2022-10-08
Fix spark user fresh releases

amCap1712:
The earlier PR to add caa.id to fresh releases response didn't trigger spark tests due to a bug in the actions setup. On merging the tests were triggered and revealed a bug. Fix the issue and actions setup.


mayhem (approved):
LGTM!

PR #2196, opened 2022-10-08
Remove satackey cache step from GH actions

amCap1712:
We have observed that sometimes this cache step takes a long time to complete even longer than the test or build itself so we might as well do without it.


MonkeyDo (approved):
:+1:

PR #2197, opened 2022-10-09
Add couchdb container in local dev

amCap1712:
Should have been added earlier but forgot.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -10,6 +10,7 @@ volumes:
   rabbitmq:
   # The contents of /root, so that bash and ipython can store history
   web_home:
+  couchdb:
 
 services:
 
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
I did the same locally to fix my environment.
But is it required for running a local copy? Can we make the server gracefully ignore couchdb connection if the related config.py items are empty/missing?

amCap1712:
@MonkeyDo I think the issue is with running Flask as `DEBUG=True`. We could make this container optional and have the server throw a 500 error in case the container is missing but that would likely lead to same issue you saw just with a different error. Setting DEBUG=False in config.py should probably make this go away.

Another alternative could be to always return a 200 no stats message if couchdb is missing but that seems less ideal.

However, we do intend to make as many containers optional in dev as possible and can look into a general solution at that time.

PR #2198, opened 2022-10-10
Add spotify metadata cache

amCap1712:
Supersedes #2092. That PR describes the rationale somewhat.

Some major changes since that PR are:
1. Store data in PG instead of couchdb because 6 months is long enough time to justify storing there. For now, we only store things like stats in couchdb which have life of 24 hours.
2. Base the cache on albums instead of artists for easier invalidation. We still use artists to discover new albums.
3. In addition to storing spotify api response as raw data, normalize the data into artist, track and album tables to optimize the generation of spotify release index. Generating index from raw data is way too slow.
4. To avoid redundant lookups and optimize, store looked up album ids in redis with a TTL = cache expiry time.
5. The ids are queued for lookup by reading listens from "unique" queue.


admin/timescale/updates/2022-10-10-spotify-cache-normalized-1.sql
```diff
@@ -0,0 +1,56 @@
+BEGIN;
+
+CREATE SCHEMA spotify_cache;
```
amCap1712, line 3: Forgot to do it earlier but will copy these statements to create_tables.sql etc after review.

listenbrainz/db/couchdb.py
```diff
@@ -129,6 +129,37 @@ def fetch_data(prefix: str, user_id: int):
 
     return None
 
+def get(database: str, document_id: str):
```
mayhem, line 132: All this will be removed before merging, yes?

listenbrainz/spotify_metadata_cache/store.py
```diff
@@ -0,0 +1,137 @@
+import ujson
+from psycopg2.extras import execute_values
+from psycopg2.sql import SQL, Literal
+
+
+def insert_album(curs, data, last_refresh, expires_at):
```
mayhem, line 6: docstrings

mayhem (approved):
Overall this looks good. But I feel that there are a lot of docstrings missing and TODOs leftover. Perpaps time to remove TODOs, add tickets or tasks to the planning doc?

Still, amazing that this is so close -- what a boring project, but incredibly important!

admin/timescale/updates/2022-10-10-spotify-cache-normalized-1.sql
```diff
@@ -0,0 +1,48 @@
...
+CREATE UNLOGGED TABLE spotify_cache.artist (
+    id                      INTEGER GENERATED ALWAYS AS IDENTITY NOT NULL,
+    spotify_id              TEXT NOT NULL,
+    name                    TEXT NOT NULL,
+    data                    JSONB NOT NULL
+);
+
+CREATE UNIQUE INDEX spotify_cache_artist_uniq_idx ON spotify_cache.artist (spotify_id, name); -- use name in unique idx for now because we don't know how spotify handles track credits
+
+CREATE UNLOGGED TABLE spotify_cache.track (
```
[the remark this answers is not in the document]
amCap1712: I guess we can add it when we need it as a column when needed? the data is always present is the JSONB.

listenbrainz/db/couchdb.py
amCap1712: Yes

amCap1712: Done.

listenbrainz/spotify_metadata_cache/store.py
amCap1712: Added.

PR #2200, opened 2022-10-11
Update ESLint and associated Webpack workflow

MonkeyDo:
Thanks again to @chinmaykunkikar for laying the groundwork and doing the research for this.
See #2059 for more details.

In short, this PR does three main things:
1. Replace the plugin that runs ESLint as part of our Webpack workflow (fork-ts-checker-webpack-plugin) with a new one (eslint-webpack-plugin) as the ESLint option has been retired.
2. Update ESLint version, and with it update a few associated eslint plugin libraries
3. Review ESLint configuration and fix warnings and errors that followed from the updates mentioned above.
Apart from the package.json, webpack.config.js and .eslintrc.js, the rest of the files should only have changes related to formatting and TS typing, and which follow updated best practices and usage warnings


MonkeyDo:
I still have a few Typescript errors to fix, but I wanted to run the CI tests and see how they are reported.

jest.config.js
```diff
@@ -2,42 +2,45 @@
...
+  // An array of regexp pattern strings that are matched against all test paths, matched tests are skipped
+  testPathIgnorePatterns: ["\\\\node_modules\\\\"],
 
-    // An array of regexp pattern strings that are matched against all source file paths, matched files will skip transformation
-    transformIgnorePatterns: ['<rootDir>/node_modules/'],
+  // This option sets the URL for the jsdom environment. It is reflected in properties such as location.href
+  testURL: "http://localhost",
+  transform: {
+    "\\.[jt]sx?$": "ts-jest",
+  },
```
MonkeyDo, line 37: Making it easier for y'all, this is the only addition to this file, the rest are indentation changes

.eslintrc.js
```diff
@@ -37,10 +37,16 @@ module.exports = {
...
-    "react/static-property-placement": ["error", "static public field"]
+    "jsx-a11y/label-has-associated-control": [
+      "error",
+      {
+        assert: "either",
+      },
+    ],
+    "react/static-property-placement": "off",
+    "class-methods-use-this": "off",
+    "react/no-unused-class-component-methods": "off",
```
MonkeyDo, line 48: We have many instances of methods defined on a class but not used in the same file.

.eslintrc.js
```diff
@@ -37,10 +37,16 @@ module.exports = {
...
+    "default-param-last": "off",
```
MonkeyDo, line 49: We can revisit this if we wish at a later point, but I'm not too pressed.
It would require rewriting a fair amount of APIService methods and their use across the front-end codebase.
IMO with typescript and a decent integration in your IDE this isn't required.

listenbrainz/webserver/static/js/src/brainzplayer/SpotifyPlayer.tsx
```diff
@@ -258,7 +258,9 @@ export default class SpotifyPlayer
       if (response.status === 404) {
         // Device not found
         // Wait a second, reconnect and try again
-        await new Promise((resolve) => setTimeout(resolve, 1000));
+        await new Promise((resolve) => {
+          setTimeout(resolve, 1000);
```
MonkeyDo, line 262: Apparently returning a value directly from a promise is a no-no, even though in this specific case the code works exactly the same way.
https://eslint.org/docs/latest/rules/no-promise-executor-return

listenbrainz/webserver/static/js/src/components/Card.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from "react";
 
 type CardProps = {
   style?: React.CSSProperties;
+  children?: React.ReactNode;
```
MonkeyDo, line 5: https://solverfox.dev/writing/no-implicit-children/

listenbrainz/webserver/static/js/src/huesound/ColorPlay.tsx
```diff
@@ -124,7 +123,6 @@ export default class ColorPlay extends React.Component<
                   shadowColor: "grey",
                   shadowBlur: 5,
                 }}
-                colours={defaultColors}
```
MonkeyDo, line 127: Setting a default for those, no need to pass them directly here

listenbrainz/webserver/static/js/src/huesound/ColorWheel.tsx
```diff
@@ -37,6 +37,16 @@ type ColorWheelState = {
   centerCircleOpen: boolean;
 };
 
+const defaultProps = {
```
MonkeyDo, line 40: I don't even want to talk about this. The interaction between typescript typing and React's defaultProps is awful with optional props with defaults.

listenbrainz/webserver/static/js/src/stats/BarDualTone.tsx
```diff
@@ -148,14 +148,14 @@ export default function BarDualTone(props: BarDualToneProps) {
     let dateString: string;
     let listenCount: number;
     if (id === "lastRangeCount") {
-      const lastRangeDate = new Date(datum.lastRangeTs! * 1000);
+      const lastRangeDate = new Date((datum.lastRangeTs ?? 0) * 1000);
```
MonkeyDo, line 151: These should be defined but both props are marked optional because they depend on the `id` check above (if/else).
But TS is right, better play it safe and default to a number

listenbrainz/webserver/static/js/src/stats/UserEntityChart.tsx
```diff
@@ -561,7 +560,9 @@ export default class UserEntityChart extends React.Component<
                   <div
                     className="col-xs-6"
                     style={{
-                      height: listenContainerHeight ?? `${65 * data?.length}px`,
+                      height:
+                        listenContainerHeight ??
+                        `${65 * (data?.length ?? 1)}px`,
```
MonkeyDo, line 565: default to a number if undefined

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "allowJs": true,
```
MonkeyDo, line 3: Required to ensure js test files are processed by jest

listenbrainz/webserver/static/js/src/user/Listens.tsx
```diff
@@ -792,7 +791,7 @@ export default class Listens extends React.Component<
     const isOlderButtonDisabled =
       !nextListenTs || nextListenTs <= oldestListenTs;
     const isOldestButtonDisabled =
-      listens?.[listens?.length - 1]?.listened_at <= oldestListenTs;
+      listens?.[(listens?.length ?? 0) - 1]?.listened_at <= oldestListenTs;
```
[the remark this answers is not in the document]
MonkeyDo: Yes, I think I took a shortcut when I was correcting this.
My reasoning was that in that case `listens[-1]` will be undefined and with the optional chaining (`?.`) this will be the same as an undefined `listens[0]`.
But perhaps I should rewrite it so it's clearer, there's a lot of assumption in all that.

PR #2201, opened 2022-10-11
Improve BulkInsertTable.table_exists performance

amCap1712:
count(*) is super slow for large tables. Replace the query with something trivial and use curs.rowcount to optimize speed.


mayhem (approved):
Oh, that is very interesting.

amCap1712:
FWIW, a relevant link from PG wiki. https://wiki.postgresql.org/wiki/Slow_Counting Didn't use that method because we don't need an approximate count just want to whether a table exists.

PR #2203, opened 2022-10-13
Update release_group_secondary_type_sort order

amCap1712:
The new order for release group secondary type sort as discussed on IRC with @reosarevok and @mayhem.


reosarevok (approved):
Seems generally sensible to me (obviously, since I proposed it). The center of the list could be ordered in almost any way tbh, but I expect those not to clash too much anyway.

mayhem (approved):

PR #2205, opened 2022-10-14
Update spotify playlist export

amCap1712:
The existing spotify export from the frontend is replaced by an export playlist endpoint which uses troi to do the export.


listenbrainz/webserver/views/playlist_api.py
```diff
@@ -691,3 +691,47 @@ def copy_playlist(playlist_mbid):
...
+    if service != "spotify":
+        raise APIBadRequest(f"Service name {service} is invalid or currently not supported")
+
+    spotify_service = SpotifyService()
+    token = spotify_service.get_user(user["id"])
+    if not token:
+        raise APIBadRequest(f"Service {service} is not linked. Please link your {service} account first.")
+
+    if not SPOTIFY_LISTEN_PERMISSIONS.issubset(set(token["scopes"])):
+        raise APIBadRequest(f"Missing required scopes to export playlists."
```
mayhem, line 725: I think it would be good to state explicitly which scopes, using the exact same terminology at the spotify site uses, the user needs.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -691,3 +691,47 @@ def copy_playlist(playlist_mbid):
...
+    :statuscode 404: Playlist not found
+    :resheader Content-Type: *application/json*
+    """
+    user = validate_auth_header()
+
+    if not is_valid_uuid(playlist_mbid):
+        log_raise_400("Provided playlist ID is invalid.")
+
+    if service != "spotify":
+        raise APIBadRequest(f"Service name {service} is invalid or currently not supported")
```
mayhem, line 717: "Service {service} is not supported. We currently only support 'spotify'."

I love it when error messages answer any possible questions I have, like "what services are provided".

listenbrainz/troi/export.py
```diff
@@ -0,0 +1,23 @@
+import spotipy
+from troi.core import generate_playlist
+from troi.patches.playlist_from_listenbrainz import TransferPlaylistPatch
+
+
+def export_to_spotify(lb_token, spotify_token, playlist_mbid, is_public):
+    sp = spotipy.Spotify(auth=spotify_token)
+    # TODO: store spotify user ids in external_service_oauth table
```
mayhem, line 8: Should this be implemented in this PR or be added as a ticket?

mayhem (approved):
I've only got minor suggestions for how to improve error message a bit -- looks good!

listenbrainz/troi/export.py
amCap1712: LB-1158

listenbrainz/webserver/views/playlist_api.py
amCap1712: Makes sense. I updated the text. However, note that currently I perform the same check on frontend side and hide the button entirely if the scopes are absent. Do you think we should do something else on the frontend in that case?

listenbrainz/domain/spotify.py
```diff
@@ -30,6 +30,11 @@
     'playlist-modify-private',
```
MonkeyDo, line 30: I originally added the playlist permissions to the listen permissions we request, but there isn't much point to doing anymore that if you're implementing them separately. AFAIR we only need them to export playlists to spotify from the front-end, which should be irrelevant now?
Or do we want to keep the ease of use aspect if a user connects their account they can export playlists manually without needing to go update their preferences?

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
```diff
@@ -560,58 +560,26 @@ export default class PlaylistPage extends React.Component<
...
-      if (error.error?.status === 401) {
-        try {
-          const newUserToken = await this.APIService.refreshSpotifyToken();
-          this.SpotifyAPIService.setUserToken(newUserToken);
-        } catch (err) {
-          this.handleError(err.error ?? err);
-        }
-      } else if (
+      if (
         error.error?.status === 403 &&
```
MonkeyDo, line 583: You might need to change the error message below to point to the profile page rather than the music-services page.

listenbrainz/domain/spotify.py
amCap1712: I think it makes sense to let playlist permissions to be a part of listening only for ease of use. If we separate those, we'll probably have to support multiple combinations as well. The separate variable I added for playlist perms is only to keep backend code clear.

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
```diff
@@ -650,12 +618,16 @@ export default class PlaylistPage extends React.Component<
...
     } = this.state;
-    const { APIService } = this.context;
+    const { APIService, spotifyAuth } = this.context;
     const { newAlert } = this.props;
     const { track: tracks } = playlist;
     const hasRightToEdit = this.hasRightToEdit();
     const isOwner = this.isOwner();
 
+    const showSpotifyExportButton = spotifyAuth?.permission?.includes(
+      "playlist-modify-public"
```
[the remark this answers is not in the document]
amCap1712: The backend already supports exporting private playlists but I am not sure about how to do the frontend UI for it so  currently all playlists are public.

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
```diff
@@ -560,58 +560,26 @@ export default class PlaylistPage extends React.Component<
...
-          {spotifyURIs.length !== playlist.track.length && (
-            <b>
-              <br />
-              {playlist.track.length - spotifyURIs.length} tracks were not found
-              on Spotify, and consequently skipped.
-            </b>
-          )}
         </>
       );
     } catch (error) {
```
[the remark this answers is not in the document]
amCap1712: The invalid scopes error is handled internally by the endpoint as it can refresh tokens so that error should never reach the frontend. Even if it bubbles up, it'll appear as a 500 error.

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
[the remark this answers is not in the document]
MonkeyDo: FWIW the way I did it previously was to copy the LB playlists public boolean.
If the playlist is private on LB, make it private on Spot, and vice-versa.

But the additional message in the success dialog would probably be good UX regardless. "Heads up: the new playlist is public on Spotify"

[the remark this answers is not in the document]
amCap1712: Added the better message for now, let's do the copying visibility thing afterwards.

listenbrainz/webserver/static/js/src/playlists/Playlist.tsx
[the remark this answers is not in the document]
amCap1712: Fixed,

PR #2208, opened 2022-10-17
 Remove artist msid and release msid

amCap1712:
These fields were never widely used in LB and with the recent MsB migration we have stopped calculating artist and release msids altogether. Therefore, getting rid of these fields from LB codebase.


listenbrainz/tests/integration/test_stats_api.py
```diff
@@ -218,6 +218,7 @@ def assertUserStatEqual(self, request, response, entity, stats_range, total_coun
             this method also accepts a count and offset parameter denoting how many stats and which
             entries to check.
         """
+        self.maxDiff = None
```
mayhem, line 221: leftover?

mayhem (approved):
Yay for removing code!

listenbrainz/tests/integration/test_stats_api.py
amCap1712: Yes, will fix.

PR #2211, opened 2022-10-19
Split production image into multiple stages

amCap1712:
To improve caching in image, split the JS files compilation into a separate build stage. For the final production image, copy over js files from the intermediate stages. This helps in cases when a python dependency is changed but the frontend code doesn't change.


Dockerfile
```diff
@@ -62,6 +63,30 @@ RUN pip3 install --no-cache-dir -r ./docs/requirements.txt
 COPY . /code/listenbrainz
 
 
+#########################################################################
+# NOTE: The javascript files for production are compiled in this image. #
+#########################################################################
+FROM node:$NODE_VERSION as listenbrainz-frontend
```
MonkeyDo, line 69: This seems to be basically equivalent to the Dockerfile.webpack file.
Do we still need that separate dockerfile? or can we change the static_builder in docker-compose to use `target: listenbrainz-frontend` ?

MonkeyDo (commented):
Writing this question here, not sure if I can remember enough details to make sense:

At some point when we set up the static_builder container, we have to do some annoying path fuckery in order for it to play nice with other containers.

We mount the sourcelistenbrainz/webserver/static folder twice in the same container: once in `/code/static` and also in `/static` where we compile the output files:
https://github.com/metabrainz/listenbrainz-server/blob/master/docker/docker-compose.yml#L117-L122

The web container also mounts listenbrainz/webserver/static, and so it is basically used as a shared volume.

Is there any way that in this new process we could do without that path fuckery? The goal would be to allow the webpack config file to point to existing paths in the repo rather than virtual paths that  only  exist inside the docker container.
Currently I have a separate webpack.local config file with the correct local paths so that I can run webpack locally rather than in docker.

Dockerfile
amCap1712: Hmm, good idea let me see if I can make that work.

Dockerfile
```diff
@@ -62,6 +63,30 @@ RUN pip3 install --no-cache-dir -r ./docs/requirements.txt
...
+RUN mkdir /static
+WORKDIR /static
+
+COPY package.json package-lock.json /static/
+RUN npm install
+
+# Compile front-end (static) files
+COPY webpack.config.js babel.config.js .eslintrc.js .stylelintrc.js tsconfig.json ./listenbrainz/webserver/static /static/
+RUN npm run build:prod
+
```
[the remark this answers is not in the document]
amCap1712: This is a separate image altogether. Only the content of /static folder is copied to the final image.

https://github.com/metabrainz/listenbrainz-server/blob/508b8db46bf85dec14966f2b793ea72ffc25ddfa/Dockerfile#L179

I guess it makes sense to do this if npm will not install the dev dependencies at all. As that may save time, if it removes as an extra step later then it'll just slow the build for no reason.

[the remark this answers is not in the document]
MonkeyDo: We definitely need the dev dependencies in order to build (`npm run build:prod`)  so this suggestion would be indeed an extra step.
But I disagree with "for no reason": it would remove a lot of packages which we don't need once the JS is compiled, and which currently will get copied ot the final prod image (since it's all in listenbrainz-frontend's /static folder)

webpack.config.js
```diff
@@ -26,6 +21,15 @@ const babelExcludeLibrariesRegexp = new RegExp(
 );
 module.exports = function (env, argv) {
   const isProd = argv.mode === "production";
+  let baseDir;
+  if (isProd) {
+    baseDir = "/static";
+  } else {
+    baseDir = "/code/listenbrainz/webserver/static";
```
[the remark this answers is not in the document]
MonkeyDo: Turns out, no we can't, Webpack throws an error:

>[webpack-cli] Invalid configuration object. Webpack has been initialized using a configuration object that does not match the API schema.
>configuration.output.path: The provided value "listenbrainz/webserver/static/dist" is not an absolute path!
   -> The output directory as **absolute path** (required).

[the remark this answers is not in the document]
MonkeyDo: Although I think I have a solution to that, currently testing locally.

Dockerfile
```diff
@@ -62,6 +63,39 @@ RUN pip3 install --no-cache-dir -r ./docs/requirements.txt
...
+RUN mkdir /static
+
+#########################################################################
+# NOTE: The javascript files for production are compiled in this image. #
+#########################################################################
+FROM listenbrainz-frontend-dev as listenbrainz-frontend-prod
+
+# Compile front-end (static) files
+COPY ./listenbrainz/webserver/static /static
```
MonkeyDo, line 95: Out of curiosity, is there a specific reason for copying and building into the /static folder, rather than what we do in dev  (/code/…) ?
From my understanding since we are copying an arbitrary folder from the image the path shouldn't matter.

The advantage would be that we would only have one path to support instead of different one for dev and prod.

But maybe I'm missing a detail?

.github/workflows/deploy-image.yml
```diff
@@ -32,7 +32,8 @@ jobs:
       with:
         build-args: GIT_COMMIT_SHA=${{ steps.get_tag.outputs.TAG }}
         cache-from: metabrainz/listenbrainz:cache
-        cache-to: metabrainz/listenbrainz:cache
+        # use type=max to cache intermediate layers as well
+        cache-to: type=registry,mode-max,oci-mediatypes=true,compression=zstd,compression-level=7,ref=metabrainz/listenbrainz:cache
```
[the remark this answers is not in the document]
amCap1712: yes.

Dockerfile
MonkeyDo: All the useful files should end up in  /code/listenbrainz/webserver/static/dist once compiled, and we shouldn't need to copy anything else, right?
node modules are all the way at the root /code folder.

I'm thinking `COPY --from=listenbrainz-frontend-prod  /code/listenbrainz/webserver/static/dist /static/dist` might do the trick, but I haven't tried (and my brain stopped working so I shouldn't try that right now :p )

amCap1712: That only has the compiled assets, we still want other stuff like /static/img, /static/sound, /static/fonts, robots.txt. But we could explicitly list those I guess.

MonkeyDo: That would make sense to me, and remove some arbitrary confusing path fuckery :) 

Or perhaps we can explicilty ignore the couple of folders we don't need (js, css), if there's an easy syntax for that.

MonkeyDo: I gave it a try in c30da47 , let me know what you think (and if I missed anything !)

amCap1712: This copies the contents of `/code/listenbrainz/webserver/static/fonts` to `/static` instead of `/static/fonts` which is likely not desired. I'll split this into multiple COPY --from statements to fix.

amCap1712: I put the latest version up on beta for testing as well.

MonkeyDo: Thanks for catching that !
On beta currently all the dropdowns are broken. Fro example in the navbar and on the charts page.
I forgot we need the js/lib folder…

PR #2212, opened 2022-10-20
Feedback Page duplicate tracks fix

amCap1712:
@Aerozol reported a feedback page bug in IRC where a feedback entry showed up in both love and hate page. On multiple switches the track duplicates in the list.

https://usercontent.irccloud-cdn.com/file/K5Nyz55I/image.png

On debugging, I found that the track was a loved track. There was another loved track on the page with the same created timestamp. When the user clicks on hate button, the page is rendered again. As we use timestamp as the key in the list, react is unable to remove the items correctly and that one track is not removed from the list showing up on the hate page as well.

React recommends unique keys in a list. To make the key unique, include recording msid and recording mbid in the key name.

Also, the backend ensures that the msid and mbid are included in the additional_info. So can simplify the RecordingMetadataToListenFormat function. Also, update test props to match the api response.


listenbrainz/webserver/static/js/src/user/UserFeedback.tsx
```diff
@@ -522,11 +517,15 @@ export default class UserFeedback extends React.Component<
                         dataTarget="#PinRecordingModal"
                       />,
                     ];
+                    const recording_msid =
+                      listen.track_metadata.additional_info?.recording_msid;
+                    const recording_mbid =
+                      listen.track_metadata.additional_info?.recording_mbid;
```
MonkeyDo, line 523: We have utilities to do this which can support the various formats we use, and various places where the fields might be: 
I've been trying to use it everywhere: https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/utils/utils.tsx#L133-L138

MonkeyDo (commented):
:rocket: Thanks for fixing!
Started the day late, didn't realize you had fixed it already :D

listenbrainz/webserver/static/js/src/user/UserFeedback.tsx
amCap1712: fixed.

MonkeyDo (approved):
:+1:

PR #2214, opened 2022-10-20
Fix element key in user feedback page

MonkeyDo:
@Aerozol reported an issue on IRC: when he switches from loved to hated tracks on his feedback page (https://listenbrainz.org/user/aerozol/feedback) one specific track is added to the hated tracks each time he switches even though it's a loved track.
Repeat the process and get many duplicates:
image
This is due to two of the tracks having the exact same listened_at timestamp, which is used as a key for the React element.
If two elements have the same key, React won't be able to target them individually and they won't be updated resulting in some ghost duplicates.
image


MonkeyDo:
Duplicate of #2212

PR #2219, opened 2022-10-23
Add a save playlist button to instant playlists

amCap1712:
So that users can save the instant playlist for listening later or export it to another platform.

### DIFF
--- a/listenbrainz/webserver/static/js/src/player-pages/PlayerPage.tsx
+++ b/listenbrainz/webserver/static/js/src/player-pages/PlayerPage.tsx
@@ -119,7 +119,7 @@ export default class PlayerPage extends React.Component<
 
   savePlaylist = async () => {
     const { currentUser } = this.context;
-    if (!currentUser || !currentUser.auth_token) {
+    if (!currentUser?.auth_token) {
       return;
     }
     const { newAlert, playlist } = this.props;
@@ -137,7 +137,7 @@ export default class PlayerPage extends React.Component<
         </>
       );
     } catch (error) {
-      newAlert("danger", "Error", error.message);
+      newAlert("danger", "Could not save playlist", error.message);
     }
   };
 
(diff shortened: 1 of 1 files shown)


mayhem (approved):
Not that I understand this well, but it looks simple enough. :)

listenbrainz/webserver/static/js/src/player-pages/PlayerPage.tsx
```diff
@@ -115,6 +117,30 @@ export default class PlayerPage extends React.Component<
     );
   }
 
+  savePlaylist = async () => {
+    const { currentUser } = this.context;
+    if (!currentUser || !currentUser.auth_token) {
```
MonkeyDo, line 122: Not particularly important, but this could be shortened to `!f (!currentUser?.auth_token)`

listenbrainz/webserver/static/js/src/player-pages/PlayerPage.tsx
```diff
@@ -115,6 +117,30 @@ export default class PlayerPage extends React.Component<
...
+      newAlert(
+        "success",
+        "Created playlist",
+        <>
+          Created a new public{" "}
+          <a href={`/playlist/${newPlaylistId}`}>instant playlist</a>
+        </>
+      );
+    } catch (error) {
+      newAlert("danger", "Error", error.message);
```
MonkeyDo, line 140: Instead of just "Error", let's make this more user-friendly. How about "Could not save playlist" ?

listenbrainz/webserver/static/js/src/player-pages/PlayerPage.tsx
```diff
@@ -197,6 +225,18 @@ export default class PlayerPage extends React.Component<
                       />
                     </li>
                   )}
+                  {currentUser?.name && (
```
MonkeyDo, line 228: Is there a specific reason for needing the name?
If it's just to make sure we have a user object, let's change it to `auth_token` which would make a bit more sense since it's used in savePlaylist.

MonkeyDo (approved):
Yep, thanks for fixing this !

listenbrainz/webserver/static/js/src/player-pages/PlayerPage.tsx
amCap1712: Done.

listenbrainz/webserver/static/js/src/player-pages/PlayerPage.tsx
amCap1712: Done.

listenbrainz/webserver/static/js/src/player-pages/PlayerPage.tsx
amCap1712: Done

PR #2220, opened 2022-10-25
Fix reading export_to_spotify preference

amCap1712:
The name of the preference was wrong but was treating it as a string so the check evaluated to true was everyone. Fix the preference and the cast.

### DIFF
--- a/listenbrainz/troi/troi_bot.py
+++ b/listenbrainz/troi/troi_bot.py
@@ -45,7 +45,7 @@ def get_users_for_daily_jams():
         SELECT "user".musicbrainz_id AS musicbrainz_id
              , "user".id as id
              , to_char(NOW() AT TIME ZONE COALESCE(us.timezone_name, 'GMT'), 'YYYY-MM-DD Dy') AS jam_date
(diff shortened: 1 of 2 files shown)


mayhem (approved):

listenbrainz/troi/troi_bot.py
```diff
@@ -45,7 +45,7 @@ def get_users_for_daily_jams():
         SELECT "user".musicbrainz_id AS musicbrainz_id
              , "user".id as id
              , to_char(NOW() AT TIME ZONE COALESCE(us.timezone_name, 'GMT'), 'YYYY-MM-DD Dy') AS jam_date
-             , COALESCE(us.troi->>'auto_export_to_spotify', 'f') AS export_to_spotify
+             , COALESCE(us.troi->>'export_to_spotify', 'f')::bool AS export_to_spotify
```
alastair, line 48: this is interesting, because it immediately makes me think of using a real column to make sure that the value is correct!
Does it make sense to make this fieldname a constant and use that constant everywhere in the codebase to prevent this?

amCap1712: Indeed, a separate column would avoid the issue but here we are. Sounds good to turn this into a constant.

PR #2222, opened 2022-10-26
New "Created For You" page

MonkeyDo:
The recommendations we offer our users are scattered here and there  instead of being presented in a consolidated, user-friendly manner:
- The [recommendations](https://listenbrainz.org/user/mr_monkey/recommendations/) tab lists playlists with a pagination, can get very messy
- The [collaborations](https://listenbrainz.org/user/mr_monkey/collaborations/) tab shows collaborative playlists, but essentially all recommendation playlists are collaborative so we duplicate playlists there
- The [Tracks you might like](https://listenbrainz.org/recommended/tracks/mr_monkey/) page (access from the user menu in the navbar) has three tabs, isn't particularly user-friendly or exciting, and was meant as a temporary solution

We talked with @alastair about how to improve this situation, and made some doodles about a new page that would gather these various resources on a single page.
Not
Part two of this work will be to find where this page fits in the new redesign of the website. In the meantime, this improves the current situation.

So far this is what the page looks like, just throwing some stuff down on a page:
image

### DIFF
--- a/listenbrainz/webserver/static/css/recommendation-page.less
+++ b/listenbrainz/webserver/static/css/recommendation-page.less
@@ -38,10 +38,6 @@
     repeating-linear-gradient( fade(@color1, 80%), fade(@color1, 40%) );
   }
   &.top-100 {
-    color: white;
-    > h4 {
-      color: white;
-    }
     background-image: linear-gradient(154deg, #f2e859 5%, #fa9f55 5%, #fa9f55 50%, #f2e859 50%, #f2e859 55%, #fa9f55 55%, #fa9f55 100%);
 background-size: 20.00px 9.75px;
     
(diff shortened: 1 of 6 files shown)


MonkeyDo:
@Aerozol Another project I'd love to get some initial feedback on; I'm sure you'll love that there's more colors :)
For now it's really just a reproduction of a doodle I had on paper here at the office, so that it doesn't get lost, so see it more as brainstorming than anything set in stone (about all of it: wording, placement of the tab, do we want a tab?, colors, content, etc.) .

Aerozol:
@MonkeyDo looks good, this is the kind of consolidation I’d like the redesign to tackle anyway. Playlists aren’t in my list of LB interests so I probably can’t give more feedback, at a glance, than ‘looks like an improvement on the current Recommendations tab’. Looks like a logical grouping for everything in there 👍

I like the Daily Jams background, the rest are a bit distracting!

MonkeyDo:
This PR was mostly to get some ideas out of our heads and into code, partly to be able to make a screenshot.
I don't think the PR will be useful as-is and it will slowly rot. I'm closing it, but there might be *some* useful code in here (like that purple css background) for the future.

PR #2223, opened 2022-10-27
Fix error handing in daily jams

amCap1712:
This should be continue not return because we do not want errors from one user's jam to block another user. :facepalm:


mayhem (approved):
whooops!

PR #2224, opened 2022-10-27
Fix metadata viewer's use of release MBID

MonkeyDo:
Atj reported that when using the Listening Now  page he wasn't seeing cover art, whereas it was loading fine in the listening-now widget on his listens page.
It appears that in the MetadataViewer component I was getting the MBID from a single place which is not guaranteed to have the MBID, and the MBID could be at another place in the metadata.
That's a terrible sentence, but I hope it makes some sense. Here's an image, I'm too tired:
image


amCap1712 (approved):
LGTM, Thanks

MonkeyDo:
Thanks @atj for reporting the issue !

PR #2225, opened 2022-10-28
Add spotify metadata cache seeder

amCap1712:
Every day query, spotify's new releases endpoint to retrieve new releases and submit those album ids to spotify metadata cache for processing.

Added a separate queue and a direct exchange for this so that we can directly enqueue spotify album ids.


mayhem (approved):
Could use some docstrings, but otherwise good!

PR #2227, opened 2022-10-31
LB-1160: Fix Spotify player playback issues

MonkeyDo:
This PR aims to fix two main issues:

1. The Spotify player SDK does not have an "end of track“ event, so we have to rely on some workaround logic:
https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/brainzplayer/SpotifyPlayer.tsx#L482-L484
It previously had a flaw which we are solving here with the addition of another check using information available in the SDK player state update.

2. Token refresh issue: the Spotify API tokens are time-limited, and when they expire the SDK has a mechanism in place (a callback function) to take in an updated token.
I coded this with a bad closure design which  relied on errors that the SDK does not return in this case, and also meant we always sent an expired token (from when we first loaded the page):
https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/webserver/static/js/src/brainzplayer/SpotifyPlayer.tsx#L398-L408
To fix this, we call the LB API to refresh the token directly in the SDK's `getOAuthToken` function

In the process we also improve error handling a bit.

### DIFF
--- a/listenbrainz/webserver/static/js/src/brainzplayer/SpotifyPlayer.tsx
+++ b/listenbrainz/webserver/static/js/src/brainzplayer/SpotifyPlayer.tsx
@@ -201,7 +201,7 @@ export default class SpotifyPlayer
         this.handleAccountError();
         return;
       }
-      handleError(errorObject);
+      handleError(errorObject.message ?? errorObject);
     }
   };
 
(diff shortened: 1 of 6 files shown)


mayhem:
I've spent quite a bit of time listening to playlists in test and it seems better, but I've had it stop on my a couple of times. There are a lot of problems relating to test vs prod, so lets get this into prod to see what is real and what is not.

MonkeyDo:
Agreed!
Already a big improvement in any case, so I'm happy to ship it

PR #2228, opened 2022-10-31
ListenBrainz Art

mayhem:
This PR adds support for cover art generation. It adds three new endpoints:

- /art/grid -- post an JSON file with cover art composition instructions, returns SVG file
- /art/grid-stats -- given a user, stats range and design, create a grid of cover art based on user stats
- /art/<custom> -- given a user, stats range and design, create a grid of custom cover art based on user stats


docs/users/api/art.rst
```diff
@@ -0,0 +1,28 @@
+Art
+===
+
+ListenBrainz has a (cover) art infrastructure that creates new cover art from a user's statstics or
```
MonkeyDo, line 4:
```suggestion
ListenBrainz has a (cover) art infrastructure that creates new cover art from a user's statistics or
```

docs/requirements.txt
```diff
@@ -3,4 +3,56 @@ sphinxcontrib-httpdomain==1.8.0
 sphinx_rtd_theme==0.5.1
 docutils==0.17.1
 sphinx-click==4.3.0
--r ../requirements.txt
+
```
MonkeyDo: Don't know python much, but is this long addition replacing `-r ../requirements.txt` expected?

docs/users/api/art.rst
```diff
@@ -0,0 +1,28 @@
...
+a user's instructions on how to composite a cover art grid.
+
+All art endpoints have this root URL for our current production site.
+
```
MonkeyDo, line 8: It might be worth to add a note for the usage of the SVG the endpoints return:

As these endpoints return SVGs rather than images, you must embed them in an html `<object data="covert_art_url" type="image/svg+xml">` element* rather than an `` element.
Otherwise external resources such as cover art images  and fonts will not be loaded and the result will be useless.

* See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object for reference

listenbrainz/webserver/views/art_api.py
```diff
@@ -0,0 +1,220 @@
...
+
+    if custom_name in ("designer-top-5"):
+        try:
+            artists, metadata = cac.create_artist_stats_cover(user_name, time_range)
+            if artists is None:
+                raise APIInternalServerError("Failed to artist cover art SVG")
+        except ValueError as error:
+            raise APIBadRequest(error)
+
+        return render_template(f"/art/svg-templates/{custom_name}.svg",
```
amCap1712, line 195: should be `art/` instead of `/art` I think?

listenbrainz/webserver/views/art_api.py
```diff
@@ -0,0 +1,220 @@
...
+        cac = CoverArtGenerator(current_app.config["MB_DATABASE_URI"], r["dimension"], r["image_size"], r["background"],
+                                r["skip-missing"], r["show-caa"])
+        tiles = r["tiles"]
+    else:
+        if "layout" in r:
+            layout = r["layout"]
+        else:
+            layout = 0
+        cac = CoverArtGenerator(current_app.config["MB_DATABASE_URI"], r["dimension"], r["image_size"], r["background"],
+                                r["skip-missing"], r["show-caa"], r["layout"])
```
amCap1712, line 76: I get an extra warning at `r["layout"]`

amCap1712 (approved):
LGTM, Thanks

docs/requirements.txt
mayhem: Fixed already.

PR #2229, opened 2022-10-31
Introduce constants to reduce hardcoded values

chinmaykunkikar:
- Many color values are hardcoded around the TS codebase.

- Introduce a constants file like what we have for CSS.

- This will also make it easier to add the repeating hex values like the LB orange (`#eb743b`) without the need to go hunting for the value elsewhere or memorize it.

- We can export and use other repeating values as well.

 - Pill: increase color contrast of Pill component by replacing `#bbbbbb` with a more used `#8d8d8d` (or COLOR_LB_LIGHT_GRAY) color.

Current `#bbbbbb`:
Current Pill color

Using `#8d8d8d`:
Better contrast Pill


listenbrainz/webserver/static/js/src/components/Pill.tsx
```diff
@@ -12,9 +18,9 @@ type PillProps = {
 export default function Pill(props: React.PropsWithChildren<PillProps>) {
   const { active, children, type, style: propStyle, ...buttonProps } = props;
 
-  let background = "#EB743B";
+  let background = COLOR_LB_ORANGE;
```
MonkeyDo, line 21: Can't help but feel that in this case we could do with moving all these styles to CSS instead, with appropriate states (`.active`, `.secondary`, etc.)

MonkeyDo (approved):
Great bit of cleanup, thank you !

Left a comment about my pet peeve of having style definitions in the JSX components, but it's more of a general comment rather than anything blocking for this PR. Ready to merge as far as I'm concerned !

listenbrainz/webserver/static/js/src/components/Pill.tsx
chinmaykunkikar: I agree with this. Using `CSSProperties` is uncommon in LB codebase and styles should move to a CSS file.
I'm adding it to my list :)

PR #2230, opened 2022-11-02
Only mount static source files in builder; run as local user

alastair:
# Problem

If we map ..:/code with a volume mount then the node_modules folder created during image build will be overwritten.
Instead, just mount the webserver static directory into static_builder

# Solution

By default, static_builder runs as root, which means that built static files are created as root. To avoid this, set a `user` flag in docker-compose for the static_builder using a docker-compose environment variable, and create the .env file at startup if required with develop.sh

# Action

### DIFF
--- a/develop.sh
+++ b/develop.sh
@@ -16,6 +16,16 @@ else
 fi
 
 
+if [[ -f .env ]]; then
+    . .env
+    if [ -z $STATIC_BUILD_USER ]; then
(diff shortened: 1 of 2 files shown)


amCap1712 (approved):
LGTM, Thanks

MonkeyDo (approved):
:rocket:

PR #2231, opened 2022-11-03
Fix the invocation of the cover art updater.

mayhem:
What looks like a bad merge sometime in the past caused the cover art updater to no longer run, because it expects a year argument it never uses. Removing the line should fix the cover art updating.


amCap1712 (approved):
LGTM, Thanks

PR #2235, opened 2022-11-07
Add release group cover art, artist credit id and individual artist name to mb_metadata_cache

amCap1712:
1. artist_credit_id is needed so that mb_metadata_cache can replace mbid_mapping_metadata in future, see LB-1128.
2. individual artist names and join phrases are needed for LB-1138.
3. add release group cover art to reduce number of calls needed to CAA.


mayhem (approved):
So, when does this code/these queries become sentient? Feels like its gotta be getting close. :) Another great action packed PR, thanks!

PR #2236, opened 2022-11-07
Use mb db master/standby for mb metadata cache

amCap1712:
We have MBID_MAPPING_DATABASE_URI to connect to aretha replica of musicbrainz database. However, aretha has slow disks and the cache also competes with json dumps for resources. To improve the reliability of the cache building job, we want to MB standby db. However, we also need to create the canonical musicbrainz data tables to build the metadata cache. The standby is a read-only replica so we write the tables to floyd and let those be replicated to standby and then create the cache.

We cannot ensure the tables have been replicated when we trigger the cache build but there will always be some version of the table available. The canonical table will be rebuilt some time before the cache is to be built. If the replication finishes, then the latest version is used else an older version will be used.

### DIFF
--- a/listenbrainz/mbid_mapping/config.py.sample
+++ b/listenbrainz/mbid_mapping/config.py.sample
@@ -9,6 +9,8 @@ SQLALCHEMY_TIMESCALE_URI = "dbname=listenbrainz_ts user=listenbrainz_ts host=lis
 
 # If you are connecting to a musicbrainz-docker instance for MusicBrainz data, this string should work:
 MBID_MAPPING_DATABASE_URI = "dbname=musicbrainz_db user=musicbrainz host=localhost port=25432 password=musicbrainz"
+MB_DATABASE_MASTER_URI = "dbname=musicbrainz_db user=musicbrainz host=localhost port=25432 password=musicbrainz"
+MB_DATABASE_STANDBY_URI = "dbname=musicbrainz_db user=musicbrainz host=localhost port=25432 password=musicbrainz"
 
 # Typesense connection
 TYPESENSE_HOST = "typesense"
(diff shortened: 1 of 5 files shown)


listenbrainz/mbid_mapping/config.py.sample
```diff
@@ -9,6 +9,8 @@ SQLALCHEMY_TIMESCALE_URI = "dbname=listenbrainz_ts user=listenbrainz_ts host=lis
 
 # If you are connecting to a musicbrainz-docker instance for MusicBrainz data, this string should work:
 MBID_MAPPING_DATABASE_URI = "dbname=musicbrainz_db user=musicbrainz host=localhost port=25432 password=musicbrainz"
+MB_DATABASE_MASTER_URI = "dbname=musicbrainz_db user=musicbrainz host=localhost port=25432 password=musicbrainz"
```
mayhem, line 12: I would like to see more comments here, documenting what these are and why someone might need to use them. Or, rather, that most people won't need to bother with this at all.

listenbrainz/mbid_mapping/docker/consul_config.py.ctmpl
```diff
@@ -8,14 +8,37 @@ MBID_MAPPING_DATABASE_URI = "dbname=musicbrainz_json_dump user=musicbrainz host=
...
 {{else}}
 SQLALCHEMY_DATABASE_URI = "SERVICEDOESNOTEXIST_pgbouncer-master"
+MESSYBRAINZ_SQLALCHEMY_DATABASE_URI = "SERVICEDOESNOTEXIST_pgbouncer-master"
+POSTGRES_ADMIN_URI = "SERVICEDOESNOTEXIST_pgbouncer-master"
+MB_DATABASE_MASTER_URI = "SERVICEDOESNOTEXIST_pgbouncer-master"
+{{end}}
+
+MB_DATABASE_STANDBY_URI = ""
+
+{{if service "pgbouncer-slave"}}
```
mayhem, line 32: For some reason, I am awestruck that this sort of config is possible. I love our infrastructure. :)

mayhem (approved):
A couple more comments, then it looks good to merge.

listenbrainz/mbid_mapping/config.py.sample
amCap1712: Done

PR #2238, opened 2022-11-08
Use api as the subdomain if in production

mayhem:
Remove the well intentioned. but poorly received art subdomain.


amCap1712 (approved):
LGTM, Thanks

PR #2239, opened 2022-11-09
Add additional_metadata field to playlist and playlist tracks

amCap1712:
We need these fields to store arbitrary metadata about the playlist and track. For instance, the immediate use case is storing external links of the playlist and the track. In future, we may way to store more types of metadata.

We already have an algorithm_metadata field at the playlist level. It does not seem prudent to have two dicts for arbitrary metadata at the same level so going forward algorithm_metadata will be a key of the additional_metadata dict.


mayhem (approved):
:+1:

PR #2240, opened 2022-11-09
Add an option to show the listen source

MonkeyDo:
### ~~This PR is based on #2233 , please wait until it is merge before merging this one~~

For debugging purposes as well as for advanced users who want easy access to the listen JSON, we need a way to show the source of a listen on the front-end.

This PR adds a global modal element (i.e. created once on the page root, ref passed in GlobalContext) that can be called from inside the ListenCards.
It will show and allow to easily copy the content of the listen JSON, complete with formatting and code highlighting!

image

image


chinmaykunkikar:
I have an opinion about this - If in future we have other such options in other parts of LB, then maybe we can have a "Show debugging and development options" toggle in preferences that will enable such features in the UI for users who are interested.

MonkeyDo:
> I have an opinion about this - If in future we have other such options in other parts of LB, then maybe we can have a "Show debugging and development options" toggle in preferences that will enable such features in the UI for users who are interested.

We actually talked about that yesterday with mayhem and alastair, adding an "advanced user" mode that is opt-in.
I think that might possibly be useful, but I'm not entirely certain it is required (at least for now).

PR #2242, opened 2022-11-10
Update existing daily jams playlist on spotify if available

amCap1712:
Spotify doesn't allow to delete playlists. It allows us to unfollow a playlist to hide for the user but that is not what we desire. So instead of creating new playlists, try to update the existing playlist on spotify if possible.


mayhem (approved):
Excited to see this deployed -- last big step as far as I am concerned. <3

PR #2247, opened 2022-11-15
Listening Now: Use caa_release_mbid field

MonkeyDo:
The caa_release_mbid field was added in the metadata API response in #2235 , which uses release groups instead of release to reliably get cover art. That means our listen can be matched with a canonical release which might not have cover art, in which case we get the cover art from another release via the release group.
The Listening Now page is now using the release MBID as it did before, but sometimes with the wrong caa_id (Right CAA id, wrong release MBID). This results in broken images.

We need to use the caa_release_mbid instead which is the release associated with the caa_id field.

### DIFF
--- a/listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
+++ b/listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
@@ -166,7 +166,7 @@ export default function MetadataViewer(props: MetadataViewerProps) {
     // Backup if we don't have the CAA ID.
     // Try fetching using CAA release MBID and fall back on Release MBID
     coverArtSrc = `https://coverartarchive.org/release/${
-      CAAReleaseMBID ?? releaseMBID
+      releaseMBID ?? CAAReleaseMBID
     }/front`;
   }
 
--- a/listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
+++ b/listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
@@ -160,7 +160,8 @@ export default function MetadataViewer(props: MetadataViewerProps) {
   const CAAID = metadata?.release?.caa_id;
   let coverArtSrc = "/static/img/cover-art-placeholder.jpg";
   if (CAAReleaseMBID && CAAID) {
-    coverArtSrc = `https://coverartarchive.org/release/${CAAReleaseMBID}/${CAAID}-500.jpg`;
+    // Bypass the Cover Art Archive redirect since we have the info to directly fetch from archive.org
+    coverArtSrc = `https://archive.org/download/mbid-${CAAReleaseMBID}/mbid-${CAAReleaseMBID}-${CAAID}_thumb500.jpg`;
   } else {
     // Backup if we don't have the CAA ID.
     // Try fetching using CAA release MBID and fall back on Release MBID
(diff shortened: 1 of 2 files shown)


amCap1712:
LGTM, Thanks.

amCap1712 (approved):

listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
```diff
@@ -156,14 +156,17 @@ export default function MetadataViewer(props: MetadataViewerProps) {
...
-      // Backup if we don't have the CAA ID
-      coverArtSrc = `https://coverartarchive.org/release/${releaseMBID}/front`;
-    }
+  if (CAAReleaseMBID && CAAID) {
+    coverArtSrc = `https://coverartarchive.org/release/${CAAReleaseMBID}/${CAAID}-500.jpg`;
+  } else {
+    // Backup if we don't have the CAA ID.
+    // Try fetching using CAA release MBID and fall back on Release MBID
+    coverArtSrc = `https://coverartarchive.org/release/${
+      CAAReleaseMBID ?? releaseMBID
```
amCap1712, line 168: This should be the other way around I think to prefer user submitted mbids.

listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
```diff
@@ -156,14 +156,17 @@ export default function MetadataViewer(props: MetadataViewerProps) {
...
   let coverArtSrc = "/static/img/cover-art-placeholder.jpg";
-  if (releaseMBID) {
-    if (metadata?.release?.caa_id) {
-      coverArtSrc = `https://coverartarchive.org/release/${releaseMBID}/${metadata.release.caa_id}-500.jpg`;
-    } else {
-      // Backup if we don't have the CAA ID
-      coverArtSrc = `https://coverartarchive.org/release/${releaseMBID}/front`;
-    }
+  if (CAAReleaseMBID && CAAID) {
+    coverArtSrc = `https://coverartarchive.org/release/${CAAReleaseMBID}/${CAAID}-500.jpg`;
```
amCap1712, line 163: Can go one more step ahead `coverArtSrc = `https://archive.org/download/mbid-${releaseMBID}/mbid-${releaseMBID}-${caaId}_thumb500.jpg`;`

listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
MonkeyDo: Since this is in the metadata lookup, the releaseMBID in question is our matched canonical release, right?
There's not user submitted data in this AFAIU, but maybe I'm missing something?

listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
MonkeyDo: Ah yes, did that for the listencard, indeed makes sense

listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
amCap1712: Ah right, so that one is really of no use then. The backend will have already checked canonical releases. However, the user submitted mbid is available under `recordingData?.release_mbid` so maybe we can use that to get releaseMbid.

MonkeyDo: OK, makes sens, we do already use it as first choice (`const releaseMBID = recordingData?.release_mbid ?? metadata?.release?.mbid;`) so as you suggested I'm inverting the order to prefer user-submitted data in that fallback: `releaseMBID ?? CAAReleaseMBID`

PR #2248, opened 2022-11-15
Make test for production a bit more MeB specific and not use FLASK_ENV.

mayhem:
Improve the test for our production by checking for an actual value that we set in production. Hopefully this will fix the art page.


amCap1712 (approved):
LGTM, Thanks

PR #2250, opened 2022-11-15
Link to each artist separately in listen card if possible

amCap1712:
We have artist credit names and join phrases available in mb metadata cache now and can use it for linking to each artist separately in listen cards.

Screenshot from 2022-11-15 23-04-49

To consider:
1. Should we do it always? Until now we have preferred user submitted names on the main listens page, elsewhere user submitted names aren't available so mapped names are used always. This changes that.
2. The name of the artist fields in the api, how should the artists array look and what should be its keys? The metadata lookup endpoint has structure from what this PR adds to mbid_mapping section.

### DIFF
--- a/listenbrainz/webserver/static/js/src/utils/types.d.ts
+++ b/listenbrainz/webserver/static/js/src/utils/types.d.ts
@@ -43,7 +43,7 @@ interface AdditionalInfo {
   work_mbids?: Array<string> | null;
 }
 
-declare type MbidMappingAritst = {
+declare type MbidMappingArtist = {
   artist_mbid: string;
   artist_credit_name: string;
   join_phrase: string;
@@ -53,7 +53,7 @@ declare type MbidMapping = {
   recording_mbid: string;
   release_mbid: string;
   artist_mbids: Array<string>;
-  artists?: Array<MbidMappingAritst>;
+  artists?: Array<MbidMappingArtist>;
   caa_id?: number;
   caa_release_mbid?: string;
 };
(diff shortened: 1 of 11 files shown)


Aerozol:
Looks great!!

re. 1. (if I’m understanding this correctly) Imo always using mapped credits is a great ‘feature’. Based off my Last.fm experience nobody will mind having their listens tidied. If a credit is wrong and the user is detail orientated enough to care they may end up editing MB as well.
The flip side is in situations where the match is wrong, it will be doubly annoying - I don’t know how often this is an issue.

mayhem (approved):
python bits look sane -- this will be nice to have in production!

listenbrainz/webserver/static/js/src/utils/types.d.ts
```diff
@@ -43,10 +43,17 @@ interface AdditionalInfo {
   work_mbids?: Array<string> | null;
 }
 
+declare type MbidMappingAritst = {
```
MonkeyDo, line 46:
```suggestion
declare type MbidMappingArtist = {
```

MonkeyDo (commented):
Thanks for doing this! Great user-facing improvement!

amCap1712:
We decided to do it always until someone complains and then add a user preference for it. Also decided to update the metadata lookup format to match the one used by listens here.

MonkeyDo:
> We decided to do it always until someone complains and then add a user preference for it.

For posterity, I'll say that this would fit with our previous discussions of 1. allowing users to edit listens and 2. allowing users to submit a listen based on a selected MB recording.
In this case, if the match is incorrect the user should be able to edit the offending listen, and select an appropriate one in MB (and perhaps create/modify it if it doesn't exist/is incorrect in MB).

Aerozol:
Excellent - if/when we can rematch listens then this is perfect. Even if someone did complain I would want to think long and hard before adding more complexity to the display. People listen, it gets matched to MBID’s, we display that nicely, end of story… well, let’s see what happens :)

PR #2251, opened 2022-11-16
Fix check for cover art in MetadataViewer

amCap1712:
If recording mbid is not present, do not query CAA for cover art.

### DIFF
--- a/listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
+++ b/listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
@@ -155,19 +155,19 @@ export default function MetadataViewer(props: MetadataViewerProps) {
   const { metadata } = recordingData ?? {};
 
   const artistMBID = first(recordingData?.artist_mbids);
-  const releaseMBID = recordingData?.release_mbid ?? metadata?.release?.mbid;
+  const userSubmittedReleaseMBID =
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/static/js/src/metadata-viewer/MetadataViewer.tsx
```diff
@@ -162,7 +162,7 @@ export default function MetadataViewer(props: MetadataViewerProps) {
   if (CAAReleaseMBID && CAAID) {
     // Bypass the Cover Art Archive redirect since we have the info to directly fetch from archive.org
     coverArtSrc = `https://archive.org/download/mbid-${CAAReleaseMBID}/mbid-${CAAReleaseMBID}-${CAAID}_thumb500.jpg`;
-  } else {
+  } else if (releaseMBID || CAAReleaseMBID) {
     // Backup if we don't have the CAA ID.
     // Try fetching using CAA release MBID and fall back on Release MBID
```
MonkeyDo, line 167: I realized the comment was wrong here, should be:
`// Try fetching using user-submitted MBID, then matched release, then CAA release`

MonkeyDo (approved):
Yep !

mayhem (approved):

MonkeyDo (approved):
We should probably look at how much of this code can be extracted to a helper function, to not have to figure this all out again.

PR #2253, opened 2022-11-16
Add contribution parameter and filter artist credit option to similar recordings

amCap1712:
1. contribution - The maximum contribution a user's listens can make to the similarity score of a recording pair.
2. filter_artist_credit - Whether to filter tracks by same artists in a listening session.


mayhem (approved):
thanks for fixing this!

PR #2254, opened 2022-11-16
Timescale Writer Fixes

amCap1712:
Timescale writer recently crashed due to a listen that was submitted with a very large integer value that is out of bounds of Postgres integer data type. Therefore, add api level checks for limit on the duration field value. While recovering, the writer got stuck again when RMQ redelivered over 10k messages at once. It seems there is some bug or undocumented behavior due to which kombu? is unable to handle this load. Worse it fails silently without any logs. To fix, set the prefetch_count to 500.


mayhem (approved):
Peksy Kombu.

PR #2255, opened 2022-11-17
css: Bump Bootstrap to 3.3.7

chinmaykunkikar:
- The `.less` files are of Bootstrap `v3.3.5` while the bootstrap.min.js is `v3.1.1`.
 - This changes all of that to a common version i.e. `v3.3.7`, which is the last release of Bootstrap 3.
It's a long way towards Bootstrap 5.x. Meanwhile let's update the current bootstrap 3 to it's last release.

 - Bump jquery from `v2.1.0` to `v3.6.0`.
 - Update glyphicons font to match with Bootstrap 3.3.7.

Idea behind removal of duplicate variables -
Currently there are two versions of variables files - `../theme/variables.less` and `../theme/bootstrap/variables.less`. Both of them have duplicate declarations and definitions. It often confuses the VS Code IntelliSense. You will see 2 entries 
 of same variable names with different values on the intellisense dropdown. This duplication can confuse the compiler as well.
The method I used was to remove duplicate variables from the bootstrap's default variables.less file. This was because `../theme/variables.less` has important LB's definitions (mostly colors).
But I see your point, @MonkeyDo. Upgrading bootstrap to future versions can cause problems.
So maybe another method would be to remove duplicate vars from the LB's file and keep the bootstrap's version unchanged?
Or leave the variable files alone?

Also, We should consider upgrading to `v4` instead of going to `v5` directly. BS v4 is a major upgrade to the entire framework and has **loads** of breaking changes, for example they switched to `.scss` from `.less` in v4. We can follow the suit and convert out files to .scss along while upgrading to .scss Read more here - https://getbootstrap.com/docs/4.0/migration/.
And then it will be less tedious to upgrade to BS v5.
That will be a couple of PRs.


MonkeyDo:
I left this on the back-burner too long, and now I think it is going to be tricky to cleanly resolve conflicts and merge.

Regarding an upgrade path, I am of the opinion that we should move to BS v5 without going through v4 first.
I know there will be a lot of breaking changes, but we don't have that many pages and component types that we use; it will be a bit tedious, but probably save some time.

We would be moving to using the exported assets from the repo https://github.com/metabrainz/design-system/ , so I am going to close this PR since it mostly updates BS less files that will eventually become obsolete.

Happy to reopen if you still want to finish the PR, and we can discuss it.

PR #2258, opened 2022-11-17
Add msid updater

amCap1712:
When a listen has not been mapped and a user adds feedback to it, pins it or recommends it; the recording msid is used for it. Now if later the msid gets mapped, the recording mbid is not added with the msid in the original table. Add a job to regularly add mapped msids to msids in various tables.


mayhem (approved):
I bet this sort of stuff has become "how fast can I type" kinda code. :)

mayhem:
Oh, we should really add some tests here.

PR #2259, opened 2022-11-18
Revert "Use artist_credit_id instead of artist mbids for artist credit filtering"

amCap1712:
Use artist credit mbids for filtering because we want to filter artists if any of the artist on two tracks matches. For instance, consider two tracks having artist credit as `Daft Punk feat. Panda Bear` and `Daft Punk feat. Todd Edwards`, if we compare on artist_credit_id these won't be filtered because the feat. part differs. However, if filter on overlapping artist credit mbids then we can filter both.


mayhem (approved):
Yeah, this does seem to be the better approach.

PR #2261, opened 2022-11-21
Add do-not-recommend backend

amCap1712:
Allow users to mark artists/releases/recordings/release groups that should not be recommended to them. Optionally, users can also provide an `until` timestamp till which the do not recommend entry is valid. When that time has elapsed, the entity can again be recommended to users. Frontend will be added in a separate PR.


admin/sql/create_tables.sql
```diff
@@ -276,6 +276,15 @@ ALTER TABLE user_setting
     ADD CONSTRAINT user_setting_timezone_name_check
     CHECK ( timezone_name IS NULL OR now() AT TIME ZONE timezone_name IS NOT NULL );
 
+CREATE TABLE recommendation.do_not_recommend (
+    id              INTEGER GENERATED BY DEFAULT AS IDENTITY NOT NULL,
+    user_id         INTEGER NOT NULL,
+    entity          do_not_recommend_entity_type NOT NULL,
+    entity_mbid     UUID NOT NULL,
+    until           TIMESTAMP WITH TIME ZONE,
```
mayhem, line 284: should this be DEFAULT NULL?

docker/services/cron/crontab
```diff
@@ -36,3 +36,6 @@ MAILTO=""
 0 4 1,15 * * root flock -x -n /var/lock/lb-dumps.lock /code/listenbrainz/admin/create-dumps.sh full >> /logs/dumps.log 2>&1
 ## Around 24 hours later, trigger a full import into the spark cluster, and this time wait for the lock, in case the dump hasn't finished.
 0 4 2,16 * * root flock -x /var/lock/lb-dumps.lock /usr/local/bin/python /code/listenbrainz/manage.py spark request_import_full >> /logs/dumps.log 2>&1
+
+# run weekly cron job to remove expired do not recommends
```
mayhem, line 40: Normally I would say that weekly is not precise enough, but in this case who is ever going to know? :laughing:

listenbrainz/db/tests/test_do_not_recommend.py
```diff
@@ -0,0 +1,51 @@
+from datetime import datetime, timedelta
```
mayhem, line 1: get_total_count and delete are not tested here.

listenbrainz/webserver/views/do_not_recommend_api.py
```diff
@@ -0,0 +1,149 @@
...
+            "user_id": "lucifer"
+        }
+
+    :param count: Optional, number of do not recommend items to return.
+        Default: :data:`~webserver.views.api.DEFAULT_ITEMS_PER_GET` Max: :data:`~webserver.views.api.MAX_ITEMS_PER_GET`.
+    :type count: ``int``
+    :param offset: Optional, number of feedback items to skip from the beginning, for pagination.
+        Ex. An offset of 5 means the top 5 feedback will be skipped, defaults to 0.
+    :type offset: ``int``
+    :statuscode 200: Yay, you have data!
```
mayhem, line 46: I love how this status gets copied and paste over and over and over...

listenbrainz/webserver/views/do_not_recommend_api.py
```diff
@@ -0,0 +1,149 @@
...
+
+        {
+            "entity": "<entity type>",
+            "entity_mbid": "<mbid of the entity>",
+            "until": 1669022475
+        }
+
+    Currently, the supported entity types are ``artist``, ``release``, ``release_group`` and ``recording``. The ``until``
+    key in the json body is optional. If absent, the don't recommend entry will be added forever. If present, it should
+    be an epoch timestamp in seconds denoting the time till which the entity should not be recommended.
```
mayhem, line 86: Does epoch imply UTC? If so, all good here.

listenbrainz/db/tests/test_do_not_recommend.py
mayhem: OK, delete is tested on the API level, but a test for get_total_count would still be good.

mayhem (approved):

admin/sql/create_tables.sql
amCap1712: Yup leaving out a default is implicitly default null.

listenbrainz/webserver/views/do_not_recommend_api.py
amCap1712: Yup. will update text.

docker/services/cron/crontab
amCap1712: The retrieve query already checks until field against current datetime so it will always return precise entries. This job is only there to remove expired entries periodically.

PR #2265, opened 2022-11-24
Fix export playlist to spotify

amCap1712:
This has been broken since we added additional metadata to JSPF format. Fix by updating the correct field to access.


mayhem (approved):
Ahh thanks, not sure if the failure was a fluke or not..

PR #2269, opened 2022-11-28
package.json: Update TypeScript to v4.9

chinmaykunkikar:
Keeping TypeScript updated to latest versions so we don't have much hassle while upgrading to the [upcoming TS5.0 release](https://github.com/microsoft/TypeScript/issues/51362).

TS4.9 announcement - https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/


MonkeyDo (approved):
Thanks ! 💯

PR #2272, opened 2022-11-30
Options missing from ListenCard options dropdown

jivteshsingh:
# Problem

We currently have some very inconsistent drop-down options depending on where/what kind of 'listen'/recording panel type is displayed.

It seems like these options should be the same across all of them:

Open in MusicBrainz
Recommend to my followers
Personally recommend
Pin this recording
Write a review

So we should have consistent options across all sections of the project.

# Solution

The probable solution can be to show all options in the dropdown menu and the make the options non accessible which do not perform the current actions and the user will be able to see which options are in working action and which are not for a particular recording :)

# Action

So I made the non accessible options in hovered gray color and the accessible options in already appointed orange color which resulted in consistent options across all sections of the webapp.

### DIFF
--- a/listenbrainz/webserver/static/css/listens-page.less
+++ b/listenbrainz/webserver/static/css/listens-page.less
@@ -422,6 +422,12 @@
             background-color: #eb743b;
           }
         }
+        .un-active {
+          &:hover {
+            color: #ffffff;
+            background-color: lightgray;
+          }
+        }
       }
     }
     .recommendation-controls {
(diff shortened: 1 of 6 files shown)


karthiknadar1204:
please assign this issue to me,i would like to contribute to this issue

MonkeyDo (commented):
I don't think this is the right solution to our issue.
Making a lot of buttons available but grayed out isn't very helpful.

The real core of the issue is that some features require modals (for example PinRecordingModal or CBReviewModal) as well as some code that currently lives in certain pages and not in others.
This is a historical issue that needs to be solved by a change in our modal usage.

I've been thinking about the issue and I think what we need is a higher order component (HOC) or a React hook that allows us to call up a modal and all its state and logic from anywhere in the codebase.
Luckily that tool exists! https://github.com/eBay/nice-modal-react

This will not be an easy task and requires a fair amount of React knowledge, so perhaps not adapted.

MonkeyDo:
For the reasons laid above I will close this PR, the solution is much more complex and touches pretty much all the front-end codebase, so is not adapted for an entry-level task.

PR #2284, opened 2022-12-10
Update __init__.py and Add new file ../hdfs/utils/__init__.py

vibhoooo:
Ticket ->

LB-678

1 -> Problem ->

There was a file present at this location ->
listenbrainz-server/listenbrainz_spark/utils/__init__.py
The file had some functions related to HDFS which were modifying the local system files as well as the Handoop file system.

2 -> My task ->

My task was to refractor the code from this file into different file by moving all the functions related to HDFS to a new file.

3 -> What I did ? /  Solution ->

1 -> I created a new file __init__.py at this location ->
listenbrainz-server/listenbrainz_spark/hdfs/utils/__init__.py
2 -> I moved all the functions related to HDFS to this new file from the previous file.
3 -> I removed the functions related to HDFS from the previous file.

### DIFF
--- a/listenbrainz_spark/hdfs/utils/__init__.py
+++ b/listenbrainz_spark/hdfs/utils/__init__.py
@@ -1,0 +1,274 @@
+# all functions related to hdfs are transferred from ->
+# listenbrainz-server/listenbrainz_spark/utils/__init__.py to ->
+# listenbrainz-server/listenbrainz_spark/hdfs/utils/__init__.py
+
+import errno
+import logging
+import os
+from datetime import datetime
+from typing import List
+
+from py4j.protocol import Py4JJavaError
+from pyspark.sql import DataFrame, functions
+from pyspark.sql.utils import AnalysisException
+
+import listenbrainz_spark
+from hdfs.util import HdfsError
+from listenbrainz_spark import config, hdfs_connection, path
+from listenbrainz_spark.schema import listens_new_schema
+from listenbrainz_spark.exceptions import (DataFrameNotAppendedException,
+                                           DataFrameNotCreatedException,
+                                           FileNotFetchedException,
+                                           FileNotSavedException,
+                                           HDFSDirectoryNotDeletedException,
+                                           PathNotFoundException,
+                                           ViewNotRegisteredException)
+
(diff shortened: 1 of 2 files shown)


listenbrainz_spark/hdfs/utils/__init__.py
```diff
@@ -0,0 +1,274 @@
+# all functions related to hdfs are transferred from ->
```
amCap1712, line 1: This comment can be removed.

listenbrainz_spark/hdfs/utils/__init__.py
```diff
@@ -0,0 +1,274 @@
...
+        Args:
+            df (dataframe): Dataframe to save.
+            path (str): Path in HDFS to save the dataframe.
+            mode (str): The mode with which to write the paquet.
+    """
+    try:
+        df.write.format('parquet').save(config.HDFS_CLUSTER_URI + path, mode=mode)
+    except Py4JJavaError as err:
+        raise FileNotSavedException(err.java_exception, path)
+
```
amCap1712, line 167: The functions till here and `read_json` later in the file should not be moved because those do more than just hdfs stuff. Those functions should remain in the original file.

amCap1712 (changes requested):
Hi! Thanks for the PR.

1) Another 2 comments in addition to above ones, the new file should be `hdfs/utils.py` not `hdfs/utils/__init__.py`.
2) You will also need to update the imports in other files that are using these functions. The tests are failing because of this reason, because the functions were moved but imports not updated.

vibhoooo:
hi there 
I have made all the changes except the last point you mentioned.
I have one question..
If I am not wrong the files inside the utils folder must be using the functions inside __init__.py.So I need to make changes to the import of the those file only which are present in the utils folder because some functions were moved from __init__.py which are used in those files.
Please correct me if I wrong.

amCap1712:
There can be other files as well which are using these functions and thus may need updated imports. One way to check this would be running Find in your IDE/Text Editor on the function's name in the codebase  and see if any usages come up.

vibhoooo:
Okay got it.
I will try doing this.

PR #2288, opened 2022-12-13
Remove lb_conn option from canonical musicbrainz data dump

alastair:
# Problem
This dump performs joins between generated tables and tables in the musicbrainz schema, so it's impossible to write to the lb database connection, as the join to generate the derived data won't be able to be performed

# Solution
Remove `--use-lb-conn` flag and code path

### DIFF
--- a/mbid_mapping/manage.py
+++ b/mbid_mapping/manage.py
@@ -28,17 +28,16 @@ def create_all():
     """
         Create all canonical data in one go. First mb canonical data, then its typesense index.
     """
-    create_canonical_musicbrainz_data(True)
+    create_canonical_musicbrainz_data()
     action_build_index()
 
 
 @cli.command()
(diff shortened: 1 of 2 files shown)


alastair:
I misunderstood some of how this dump works. Specifically, https://github.com/metabrainz/listenbrainz-server/blob/d9c99ed2552b23e74f4b2593e8c4bb7b606496cf/mbid_mapping/mapping/canonical_musicbrainz_data.py#L153 generates this table in the musicbrainz connection regardless of the value of the flag (`lb_conn` isn't passed in). Therefore, any query which joins between the musicbrainz schema and the `mapping.canonical_musicbrainz_data_release_tmp` table will always work, even with the target being the lb connection.

@amCap1712 is currently trying a dump in prod with the existing setup, if that works fine then this PR can be closed

alastair:
This change was incorrect, we do need to write some tables specifically to the MB connection only

PR #2289, opened 2022-12-15
Enhance release stats with album artist name, cover art ids

amCap1712:
This improves the release stats by aggregating on album artist's name if available (by virtue of user submitted release mbid/mapper supplied mbid) instead of the track artist's name. caa_ids, caa_release_mbids are also added to release data to allow the frontend to fetch cover art with as few API requests as possible. Further, the caa_id of the release group cover art is added a fallback to the release.

Finally, to bring this data in spark cluster we setup a connection to MB DB on pink. We cannot use just mb_metadata_cache (although that has its own problems while trying to bring into spark) because user submitted mbids might not always be present in mb_metadata_cache as those can differ from the canonical release mbids chosen by us.


listenbrainz_spark/postgres/release.py
```diff
@@ -0,0 +1,86 @@
+from listenbrainz_spark.path import RELEASE_METADATA_CACHE_DATAFRAME
+from listenbrainz_spark.postgres.utils import save_pg_table_to_hdfs
+
+
+def create_release_metadata_cache():
+    """ Import release metadata from postgres to HDFS for use in stats calculation. """
```
mayhem, line 6: It is rather kind of scary how much part of this query are just "ah, query bah for X. then meh for Y..." but this 80+ lime query is composed of things I've recently written.... heh.

mayhem (approved):
That was a lot less intimidating than I expected. :)

listenbrainz_spark/postgres/release.py
amCap1712: :)

PR #2290, opened 2022-12-17
Add total listening time and new discovered artists count for YIM

amCap1712:
Total listening time is the sum of duration of all listens that a user listened to in a given year, if the duration is not available from MB data fallback to using 180s.

Newly discovered artist count is the number of artists to whom the user listened to for the first time in the year.

### DIFF
--- a/listenbrainz_spark/year_in_music/listening_time.py
+++ b/listenbrainz_spark/year_in_music/listening_time.py
@@ -1,0 +1,41 @@
+import listenbrainz_spark
+from listenbrainz_spark import config
+from listenbrainz_spark.stats import run_query
+from listenbrainz_spark.year_in_music.utils import setup_listens_for_year
+
+
+def get_listening_time(year):
+    """ Calculate the total listening time in seconds of the user for the given year. """
+    setup_listens_for_year(year)
+    metadata_table = "mb_metadata_cache"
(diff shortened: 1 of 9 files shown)


listenbrainz_spark/year_in_music/listening_time.py
```diff
@@ -0,0 +1,39 @@
+import listenbrainz_spark
+from listenbrainz_spark import config
+from listenbrainz_spark.stats import run_query
+from listenbrainz_spark.year_in_music.utils import setup_listens_for_year
+
+
+def get_listening_time(year):
```
mayhem, line 7: docstrings

listenbrainz_spark/year_in_music/new_artists_discovered.py
```diff
@@ -0,0 +1,44 @@
+from listenbrainz_spark.stats import run_query
+from listenbrainz_spark.utils import get_all_listens_from_new_dump
+
+
+def get_new_artists_discovered_count(year):
```
mayhem, line 5: docstrings

mayhem (approved):

listenbrainz_spark/year_in_music/listening_time.py
amCap1712: Done.

listenbrainz_spark/year_in_music/new_artists_discovered.py
amCap1712: Done.

PR #2294, opened 2022-12-17
CoverArtGenerator improvements and fixes

amCap1712:
1. The custom_name check was invalid at one place because of a missing ,. Due to this the check was instead of doing substring of the type instead of member of tuple.

2. ValueError was passed as message to APIBadRequest which in turn tried to serialize the error as json causing a 500 error. Stringify the error before passing to APIBadRequest.

3. Instead of making an api request to query user stats, fetch stats directly from couchdb. There were ratelimit errors for the user stats fetch in sentry. (Probably because of multiple requests from the same server ip in a short span.) Appropriate place to apply rate limit is art api where we already have it.

### DIFF
--- a/listenbrainz/webserver/views/art_api.py
+++ b/listenbrainz/webserver/views/art_api.py
@@ -140,7 +140,7 @@ def cover_art_grid_stats(user_name, time_range, dimension, layout, image_size):
         if images is None:
             raise APIInternalServerError("Failed to grid cover art SVG")
     except ValueError as error:
-        raise APIBadRequest(error)
+        raise APIBadRequest(str(error))
 
     return render_template("art/svg-templates/simple-grid.svg",
                            background=cac.background,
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/art_api.py
```diff
@@ -180,13 +180,13 @@ def cover_art_custom_stats(custom_name, user_name, time_range, image_size):
     if err is not None:
         raise APIBadRequest(err)
 
-    if custom_name in ("designer-top-5"):
+    if custom_name in ("designer-top-5",):
```
mayhem, line 183: Oh oops. :)

mayhem (approved):

listenbrainz/webserver/views/art_api.py
amCap1712: :)

PR #2299, opened 2022-12-19
Manual mbid mapping

alastair:
# Problem

Sometimes despite our best efforts, the mapper can't find a good mapping, but the user knows what recording it is.

# Solution

Add a new table which allows users to create manual mappings between msids and mbids

# Action

Remaining tasks:

- [ ] Add interface to listen card menu to allow users to enter this mapping
- [ ] Use the mapping when getting mbids for listens


admin/timescale/create_tables.sql
```diff
@@ -60,6 +60,13 @@ CREATE TABLE playlist.playlist_collaborator (
 
 -- MBID Mapping
 
+CREATE TABLE mbid_manual_mapping(
+    recording_msid UUID NOT NULL,
```
amCap1712, line 64: Probably add a GENERATED IDENTITY COLUMN for id and PK.

listenbrainz/db/mbid_manual_mapping.py
```diff
@@ -0,0 +1,87 @@
...
+from sqlalchemy import text
+
+from listenbrainz.db.model.mbid_manual_mapping import MbidManualMapping
+from listenbrainz.db import timescale as ts
+
+def create_mbid_manual_mapping(mapping: MbidManualMapping):
+    """Save a user mapping to the database. """
+    query = """
+        INSERT INTO mbid_manual_mapping(recording_msid, recording_mbid, user_id)
+             VALUES (:recording_msid, :recording_mbid, :user_id)
```
amCap1712, line 13: I think ON CONFLICT DO UPDATE would be better to allow users to update their mappings in case of error or just if a new recording was added to MB etc. You already have the required UNIQUE index in place.

listenbrainz/webserver/views/metadata_api.py
```diff
@@ -151,3 +153,81 @@ def get_mbid_mapping():
...
+
+        create_mbid_manual_mapping(mapping)
+
+        return jsonify({"status": "ok"})
+
+
+@metadata_bp.route("/get_manual_mapping/", methods=["GET", "OPTIONS"])
+@crossdomain
+@ratelimit()
+def get_manual_mapping():
```
amCap1712, line 209: Not sure if this endpoint is useful enough because it supports only 1 recording msid. All use cases I can think of need a bulk lookup endpoint. We probably also want another endpoint to display all of the user's submitted mappings.

listenbrainz/webserver/views/metadata_api.py
```diff
@@ -151,3 +153,81 @@ def get_mbid_mapping():
...
+    user = validate_auth_header()
+
+    recording_msid = request.json.get("recording_msid")
+    recording_mbid = request.json.get("recording_mbid")
+    if not recording_msid or not is_valid_uuid(recording_msid):
+        raise APIBadRequest("recording_msid is invalid or not present in arguments")
+    if not recording_mbid or not is_valid_uuid(recording_mbid):
+        raise APIBadRequest("recording_mbid is invalid or not present in arguments")
+
+    existing_mapping = get_mbid_manual_mapping(recording_msid=recording_msid, user_id=user["id"])
```
amCap1712, line 190: If you go the ON CONFLICT route, this query could be removed.

amCap1712 (approved):
Thanks! Some suggestions inline but otherwise LGTM.

listenbrainz/db/mbid_manual_mapping.py
alastair: good idea, it also answers my question about how we should catch/return postgres exceptions in the case of one being added twice

listenbrainz/webserver/views/metadata_api.py
alastair: yeah, it's more here as a proof of concept. I'm still not sure what data we'll need on the frontend. When showing a list of listens, perhaps it makes more sense to bulk query by msid (e.g. if we have 20 items on a page but only 3 are missing mbid but the user has 1000 mappings, we only want to get those 3, not 1000 and then page through them in JS).

alastair: to get the mapped mbids into the list listens endpoint, I'm not sure what the best approach is, I suspect we could add this table to the existing joins which already look at the regular mapping table? @amCap1712 are you able to take a look at this part?

amCap1712: Sure, will do. Yup, adding another join to listens fetch is what I had in mind.

frontend/js/src/recent/RecentListens.tsx
```diff
@@ -248,6 +255,18 @@ export default class RecentListens extends React.Component<
                       />
                     );
                   }
+                  additionalMenuItems.push(
```
alastair, line 258: This version is half-implemented. I initially didn't realise that there was a separate implementation of the listen card/menu in RecentListens. My recommendation is that we remove this.

frontend/js/src/user/Listens.tsx
```diff
@@ -708,6 +715,7 @@ export default class Listens extends React.Component<
       Boolean(releaseGroupMBID);
     const canPin = Boolean(recordingMBID) || Boolean(recordingMSID);
     const canPersonallyRecommend = Boolean(recordingMSID);
+    const canManuallyMap = recordingMSID && !recordingMBID;
```
alastair, line 718: small point of interest, I noticed that in `Listens`, these fields are called `canDoX`, but in `RecentListens` they are `isListenVerb`

alastair: Currently can only add a mapping to something that doesn't have a recording MBID. People might also want to overwrite a bad mapping

frontend/js/src/user/Listens.tsx
```diff
@@ -744,6 +752,17 @@ export default class Listens extends React.Component<
         />
       );
     }
+    if (canManuallyMap) {
+      additionalMenuItems.push(
+        <ListenControl
+          text="Map to a MusicBrainz Recording"
```
alastair, line 758: This text is too long and goes over two lines. I removed "Manually" from it, but it could be made shorter

frontend/js/src/user/Listens.tsx
```diff
@@ -744,6 +752,17 @@ export default class Listens extends React.Component<
...
+          icon={faPencilAlt}
```
alastair, line 759: Same icon as "Write a review", although we only show this if there is an MBID, so perhaps only one would show

frontend/js/src/user/Listens.tsx
```diff
@@ -995,6 +1015,13 @@ export default class Listens extends React.Component<
                         this.handlePinnedRecording(pinnedListen)
                       }
                     />
+                    <MbidMappingModal
+                      recordingToMap={recordingToMapToMusicbrainz}
+                      newAlert={newAlert}
+                      onSuccessfulPin={(pinnedListen) =>
+                        this.handlePinnedRecording(pinnedListen)
+                      }
```
alastair, line 1023: just left this here for now. Is there an action that we'd want to make once a mapping is made? Should we look up this mapping in the API and add the mbid to the listen?
Note that when performing a "get listens" from the API, we'll include a manual mapping mbid in the response in the normal mbid location if it's set, so after a refresh this will correctly show

frontend/js/src/user/Listens.tsx
```diff
@@ -995,6 +1015,13 @@ export default class Listens extends React.Component<
                         this.handlePinnedRecording(pinnedListen)
                       }
                     />
+                    <MbidMappingModal
```
alastair, line 1018: No test for this, I'm not sure how to do this, or if we have detailed tests for other modals. happy to change this based on feedback

frontend/js/src/utils/APIService.ts
```diff
@@ -931,6 +931,27 @@ export default class APIService {
     return response.json();
   };
 
+  submitMbidMapping = async (
+    userToken: string,
+    recordingMSID: string,
+    recordingMBID: string
+  ): Promise<{ status: string; data: PinnedRecording }> => {
```
alastair, line 938: wrong type, but if we're not using the returned data then we could just remove this from the returned value

frontend/js/src/mbid-mapping/MbidMappingModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { getArtistName, getTrackName } from "../utils/utils";
+
+export type MbidMappingModalProps = {
+  recordingToMap?: Listen;
+  newAlert: (
+    alertType: AlertType,
+    title: string,
+    message: string | JSX.Element
+  ) => void;
+  onSuccessfulPin?: (pinnedrecording: PinnedRecording) => void;
```
alastair, line 13: todo

frontend/js/src/mbid-mapping/MbidMappingModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+      // if (onSuccessfulPin) {
+      //   onSuccessfulPin(newPin);
+      // }
+    }
+  };
+
+  handleMbidInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    event.preventDefault();
+    const input = event.target.value;
```
alastair, line 70: currently no validation on this input. We should do a regex to ensure that it's a valid uuid, and only allow submit to be set if so

frontend/js/src/mbid-mapping/MbidMappingModal.tsx
```diff
@@ -0,0 +1,151 @@
...
+            <div className="modal-body">
+              <div className="form-group">
+                <input
+                  type="text"
+                  value={recordingMBID}
+                  className="form-control"
+                  id="recording-mbid"
+                  name="recording-mbid"
+                  onChange={this.handleMbidInputChange}
+                />
```
alastair, line 126: Future addition to this could be to integrate the artist/track search that we have on the playlist create page so that users can search directly from here

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -183,8 +183,6 @@ def insert(self, listens):
 
         query = """INSERT INTO listen (listened_at, track_name, user_name, user_id, data)
                         VALUES %s
-                   ON CONFLICT (listened_at, track_name, user_id)
```
alastair, line 186: urgh, I thought I undid this from the commit, this is from some other stuff I did, should definitely not be here

listenbrainz/mbid_mapping/evaluate_cache_update.py
```diff
@@ -0,0 +1,44 @@
+import uuid
```
alastair, line 1: again, another thing that I accidentally added :/

listenbrainz/webserver/views/metadata_api.py
alastair: @amCap1712 do we even need an API endpoint to get mappings?

docker/services/cron/crontab
```diff
@@ -45,3 +45,6 @@ MAILTO=""
 
 # run weekly cron job to remove expired do not recommends
 0 8 * * 0 root /usr/local/bin/python /code/listenbrainz/manage.py clear-expired-do-not-recommends >> /logs/do_not_recommend.log 2>&1
+
+# run daily cron job to refresh top msid-mbid mappings
```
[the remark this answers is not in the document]
amCap1712: Uh right, I'll fix.

admin/timescale/create_views.sql
```diff
@@ -0,0 +1,20 @@
+BEGIN;
+
+-- create a materialized view of the top mappings of a recording msid. top mappings are chosen
+-- by the highest number of users that have added it. if multiple mappings have same count, break
+-- ties by preferring the mapping that was created most recently.
+CREATE MATERIALIZED VIEW mbid_manual_mapping_top AS (
```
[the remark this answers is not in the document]
amCap1712: The same user will see the effect because of the way the select listens query is written. Other users won't see the effect until view refresh. Yes, will add to update script.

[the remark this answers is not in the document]
alastair: oh cool, I didn't notice that in the sql query. that's a neat idea

frontend/js/src/mbid-mapping/MbidMappingModal.tsx
alastair: added to LB-1179

frontend/js/src/user/Listens.tsx
alastair: changed to "Link to MusicBrainz" with a chain icon

frontend/js/src/user/Listens.tsx
alastair: changed to "Link to MusicBrainz" with a chain icon

frontend/js/src/user/Listens.tsx
alastair: Added to LB-1179

frontend/js/src/recent/RecentListens.tsx
alastair: Added to LB-1179

frontend/js/src/user/Listens.tsx
alastair: added to LB-1179

listenbrainz/webserver/views/metadata_api.py
amCap1712: @alastair I think we need one if we want to show users what all mappings they have created with an option to delete/update those. Can come later I think.

MonkeyDo:
Some initial interface feedback:

- The text input in the modal should be auto selected when it opens, saving a click.
- It's cumbersome to add a whole album; not sure what a good way to select a whole album would be, but that would improve the process drastically. I'm thinking checkboxes next to/in each listen, then the ability to paste a release MBID?
- validation state colors (green/red) on the text input
- maybe a preview of the selected recording (title+artist)? would require a call to MB API, but would help prevent bad copypasta (since MBIDs all look alike)

PR #2301, opened 2022-12-22
Fix caps and improve wording in music_services

reosarevok:
We had several things such as spotify lowercased, Youtube without the T.

Also, the second part of the Spotify text is fairy confusing and seems to apply only to the playing music bit. I changed it a bit to hopefully make it more clear, although it could benefit from being moved to a more specific section I think?


alastair (approved):
love good to-matoes, thanks

PR #2302, opened 2022-12-23
Year in music 2022 frontend

MonkeyDo:
Big PR with all of the updated frontend for your year in music page.

A couple of elements are still missing which I am figuring out at the moment, but can be added as a separate PR:

- [ ] Cover art composite image component with zoom & pan and clickable (will probably be a separate page for simplicity)
- [ ] Make the "This is fine" bubble in the logo hidden by default, shown with animation on hover
- [ ] Add credits for inspiration of the logo image
- [ ] Manage graphs size on mobile (stats section)


frontend/js/src/year-in-music/2022/MagicShareButton.tsx
```diff
@@ -0,0 +1,197 @@
+import {
```
alastair, line 1: not to be a party-pooper, but the exported function is `ShareButton` and the file is `MagicShareButton.tsx`

alastair:
I'm a little unsure about including the cover art images in the repo - 2MB isn't much on its own, but if we add another 2mb file for every new feature then in a few years we're going to have hundreds of MB of images lying around in the repo. Is there any way of hosting it in a static location? Is the large image hosted somewhere, or is it svg?

amCap1712:
Testing the shareable images, 
image

The size of the image needs to be adjusted and the background and header images in the SVG don't seem to work. Those render fine when calling the API endpoint directly.

MonkeyDo:
> Testing the shareable images, image
> 
> The size of the image needs to be adjusted and the background and header images in the SVG don't seem to work. Those render fine when calling the API endpoint directly.

Yep, I noticed that @amCap1712 . I had a long think about it, and I think it must be related to the use of `<defs>` one way or another, since the cover art images load fine.
Putting the `<defs>` at the top of the file (see https://github.com/canvg/canvg/issues/526#issuecomment-450537693 ) didn't work for me, but we can put the images directly in the main body of the SVG rather than in defs, like we do for the lps-on-the-floor for example: https://api.listenbrainz.org/1/art/lps-on-the-floor/mr_monkey/year/750

I tried that and it does work out well, see https://codesandbox.io/s/serene-wildflower-xel8yd?file=/svg2.svg

For the fonts, I am a bit more surprised. I did find a trick in the codesandbox above that I had to use originally (create an invisible span using the font in question).
Earlier I had added the Inter font to our static files for faster and easier loading, so I'm not sure if we should keep the `@import` font rule. I guess we would need it for standalone use maybe? But is that going to happen? Not sure what the best wya to go is for this one.
In any case, I'm adding a commit to use that span trick in the shareable image component, hopefully that will solve the font loading issue in the SVG renderer

MonkeyDo:
> I'm a little unsure about including the cover art images in the repo - 2MB isn't much on its own, but if we add another 2mb file for every new feature then in a few years we're going to have hundreds of MB of images lying around in the repo. Is there any way of hosting it in a static location? Is the large image hosted somewhere, or is it svg?

Yes I had the same worry @alastair. It's an image rather than an SVG, and the full definition one is 28Mb so it is indeed very relevant.
I'm already worried that I've added and then removed some of those files in commits of this PR, so I might have to do some doctoring to ensure I don't add a lot of weight to the repo history.

I don't know how to resolve this one. Not sure what our usual method is for dealing with this sort of assets.
StaticBrainz? Would appreciate some help with that if you've got a moment.

amCap1712:
@MonkeyDo Yeah, adding to static brainz sounds good to me. Can you share the link to images you want to put there? I can try adding.

amCap1712:
> Yep, I noticed that @amCap1712 . I had a long think about it, and I think it must be related to the use of `<defs>` one way or another, since the cover art images load fine. Putting the `<defs>` at the top of the file (see [canvg/canvg#526 (comment)](https://github.com/canvg/canvg/issues/526#issuecomment-450537693) ) didn't work for me, but we can put the images directly in the main body of the SVG rather than in defs, like we do for the lps-on-the-floor for example: https://api.listenbrainz.org/1/art/lps-on-the-floor/mr_monkey/year/750
> 
> I tried that and it does work out well, see https://codesandbox.io/s/serene-wildflower-xel8yd?file=/svg2.svg
> 
> For the fonts, I am a bit more surprised. I did find a trick in the codesandbox above that I had to use originally (create an invisible span using the font in question). Earlier I had added the Inter font to our static files for faster and easier loading, so I'm not sure if we should keep the `@import` font rule. I guess we would need it for standalone use maybe? But is that going to happen? Not sure what the best wya to go is for this one. In any case, I'm adding a commit to use that span trick in the shareable image component, hopefully that will solve the font loading issue in the SVG renderer

Sounds good to me. Are any changes to the backend needed or did you already make the relevant changes?

MonkeyDo:
> Sounds good to me. Are any changes to the backend needed or did you already make the relevant changes?

Changes are required in the backend to the SVGs.
I can have a look at those today and open a separate PR.

amCap1712:
Cool. Let me know if you need a hand.

PR #2304, opened 2022-12-25
Send track duration in BrainzPlayer

staticssleever668:
# Problem

Currently when you listen to something through BrainzPlayer track duration is not submitted.

# Solution

If the playing track has duration data available, send it. 

# Action

There are two tests that use the modified function, conveniently one of them sets the `duration` field and the other doesn't. I'm not sure if another test might be need to check just this specifically.
Also, maybe it would be possible to also send track number and etc here, but that would involve using the `currentListen` variable, which sounds kind of complicated.


MonkeyDo (approved):
Thanks for catching that !

PR #2305, opened 2022-12-25
Add cover art to playlists

amCap1712:
Lookup cover art and recording metadata using mapping.mb_metadata_cache directly to include caa_id and caa_release_mbid in additional_metadata of playlist tracks. Also, update recording lookup labs api endpoint to use mb_metadata_cache for consistency.

There is an issue with the layout of playlists when viewing one's own playlists:
image

But cover art seems to work fine when viewing playlists of other people:
Screenshot from 2022-12-25 03-22-13


amCap1712:
Oh about the layout I just realised it might not have to do with my or other user's playlist but rather the timestamp displayed taking more space.

MonkeyDo:
Yes, the layout issue is probably related to the timestamp, so a bigger issue i reckon.

I also think we can improve the interplay between the drag handle and the cover art, but I don't have a good suggestion yet.
At the very least less white space around the drag handle, and matching left/right white space would be better.

amCap1712:
@MonkeyDo I tried to adjust the drag icon spacing but was unable to come up with something working. Could you please look at that whenever you have time?

listenbrainz/messybrainz/__init__.py
```diff
@@ -151,8 +152,10 @@ def load_recordings_from_msids(connection, messybrainz_ids: Iterable[str]) -> di
...
-          FROM messybrainz.submissions
-         WHERE gid IN :msids 
+          FROM (VALUES %s) AS t (msid)
+          JOIN messybrainz.submissions s
+            ON s.gid = t.msid::UUID
     """
-    result = connection.execute(text(query), {"msids": tuple(messybrainz_ids)})
-    return {row["msid"]: row for row in result.mappings().all()}
+    results = execute_values(ts_curs, query, [(msid,) for msid in messybrainz_ids], fetch=True)
+    print(results)
```
mayhem, line 160: remove

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -190,37 +196,32 @@ def fetch_playlist_recording_metadata(playlist: Playlist):
...
         rec.artist_credit = row.get("artist_credit_name", "")
-        if "[artist_credit_mbids]" in row and not row["[artist_credit_mbids]"] is None:
+        if "[artist_credit_mbids]" in row and row["[artist_credit_mbids]"] is not None:
             rec.artist_mbids = [UUID(mbid) for mbid in row["[artist_credit_mbids]"]]
         rec.title = row.get("recording_name", "")
 
+        caa_id = row.get("caa_id")
+        caa_release_mbid = row.get("caa_release_mbid")
+        if caa_id and caa_release_mbid:
+            # TODO: Discuss image vs additional_metadata, what to permit caa.org link archive.org link etc?
```
mayhem, line 222: Remove or ticket?

mayhem (approved):
Overall looks good, but there is a YIM SVG in there that doesn't belong, so careful with the merge. A couple of minor bits, but looking good!

listenbrainz/webserver/views/playlist_api.py
amCap1712: Removed for now.

listenbrainz/messybrainz/__init__.py
amCap1712: Done

PR #2308, opened 2022-12-27
SpotifyPlayer: catch failed calls to refresh token

MonkeyDo:
When calling the LB API to refresh a spotify token, the response may throw an error. Currently that error bubbles up and is caught by the page's ErrorBoundary, but that means the page is entirely broken. Instead we catch it, display an error message and try again in a second.


amCap1712:
LGTM, Thanks

amCap1712 (approved):

PR #2312, opened 2022-12-29
Year in music '22: cover art collage

MonkeyDo:
We wanted to have this awesome zoom/pan-able and clickable cover art collage image, which I had a hard time shoehorning into the YIM page itself.
However it's pretty easy to create a new page just for it, so that's what I did.
`/explore/cover-art-collage` will now go to this new page.
I put a warning on there to make sure we don't take users on limited mobile plans/roaming off guard and load a 35Mb image without a user action first:

image

image


amCap1712 (approved):
LGTM, Thanks!

PR #2314, opened 2022-12-31
Add artist map to YIM stats

amCap1712:
Add the usual artist map we have in stats to YIM report. One difference in implementation is that the area containment data is copied over to spark and the countries are resolved in spark query itself because that seems more practical than calling the labs api for all users and their all this years' artists. We also resolve the ISO 3166 2 letter to 3 letter codes spark side.

# TODO
- [ ] The color scheme and layout of the artist map on YIM page needs to be fixed. 

image


mayhem (approved):
I'm very glad this got done -- thank you!

MonkeyDo (approved):
Thank you! Good addition to the page IMO

PR #2317, opened 2023-01-02
LB-1168: Improve Youtube rate limit notification

amCap1712:
LB-1168

### DIFF
--- a/frontend/js/src/utils/utils.tsx
+++ b/frontend/js/src/utils/utils.tsx
@@ -115,7 +115,7 @@ const searchForYoutubeTrack = async (
 
   if (response.status === 429) {
     throw new Error(
-      "The request cannot be completed because ListenBrainz has exceeded its quota. Please try again later."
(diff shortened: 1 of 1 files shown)


frontend/js/src/utils/utils.tsx
```diff
@@ -113,6 +113,12 @@ const searchForYoutubeTrack = async (
     }
   }
 
+  if (response.status === 429) {
+    throw new Error(
+      "The request cannot be completed because ListenBrainz has exceeded its quota. Please try again later."
```
MonkeyDo, line 118: We could make it a bit less technical, maybe like 
"We couldn't play this track because we ran out of Youtube quota, sorry about the inconvenience."

For the longer term plan, I think we should revive the Youtube user credentials from #1361 , and use this 429 error message to encourage users to log into Youtube to improve the situation.

amCap1712: We had got out rate limit increased by google some time ago but due to some reason it seems that didn't come into effect :(. Will need some effort to get that done again but things should probably work without login then.

PR #2318, opened 2023-01-03
Add Link to MusicBrainz button on Missing MB data page

amCap1712:
We need the recording_msid field in the dumps so that it can be included in the missing MB data sent back to LB webserver. Once we have the recording msid in the missing MB data in LB, we can show the Link to MB button on that page.


listenbrainz/db/mbid_manual_mapping.py
```diff
@@ -87,3 +89,24 @@ def get_mbid_manual_mappings(recording_msid: uuid.UUID) -> List[MbidManualMappin
...
+    """Check if a user has a mapping for a list of recordings
+
+    Arguments:
+        user_id: LB user id of a user
+        recording_msids: the msids of the recordings to check
+
+    Returns:
+        A set of msids for which the user has a mapping
+    """
+    query = SQL("""
```
mayhem, line 103: Do you bother to open your eyes when writing mundane queries like this or does it just flow from your soul? (or all the souls you've stolen?) :laughing:

mayhem (approved):

frontend/js/src/missing-mb-data/MissingMBData.tsx
```diff
@@ -198,14 +214,29 @@ export default class MissingMBDataPage extends React.Component<
                     )}
                   />
                 );
+                const additionalMenuItems = [
+                  <ListenControl
+                    text="Link with MusicBrainz"
```
MonkeyDo, line 219: Reo pointed out in the IRC channel that any logged in user can link a listen from someone else's page.
IIRC the outcome of the discussion is that this shouldn't be an issue as there is a form of consensus on the user-provided mappings.

I just wanted to make sure we don't want to limit this to the currently-logged-in-user's own page.

MonkeyDo: In any case, we should either not show this option or show it as disabled when no user is logged in (`!currentUser`)  since the modal will be missing (= non-working option with no feedback on click)

MonkeyDo (commented):
Looks good !

MonkeyDo:
Also one front-end snapshot failing :p

frontend/js/src/missing-mb-data/MissingMBData.tsx
```diff
@@ -198,14 +214,29 @@ export default class MissingMBDataPage extends React.Component<
...
+                  />,
+                ];
                 return (
                   <ListenCard
                     key={`${data.recording_name}-${data.artist_name}-${data.listened_at}`}
                     showTimestamp
                     showUsername={false}
                     newAlert={newAlert}
+                    // eslint-disable-next-line react/jsx-no-useless-fragment
+                    customThumbnail={<></>}
```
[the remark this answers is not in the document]
amCap1712: Yup, we only have recording_msid here anyway which is not info to display a cover art.

frontend/js/src/missing-mb-data/MissingMBData.tsx
amCap1712: As pointed out in LB-1198, it is fine to show the Link with MusicBrainz button on other people's profile. I'll fix the logged in thing.

amCap1712:
A few more UI improvements came to my mind while updating the PR, let me see if I can get those in this one or its enough to split off.

amCap1712:
Updated UI:

Screenshot from 2023-01-13 11-10-52

frontend/js/src/missing-mb-data/MissingMBData.tsx
```diff
@@ -185,27 +201,63 @@ export default class MissingMBDataPage extends React.Component<
...
-                      this,
-                      missingMBDataAsListen[index]
-                    )}
-                  />
-                );
+                let additionalActions;
+                if (currentUser?.auth_token) {
+                  const addToMB = (
+                    <ListenControl
+                      buttonClassName="btn btn-sm btn-success"
```
MonkeyDo, line 208: I think we should encourage users to use the "link with…" button first rather than the add button, which we generally fear could lead to creating a lot of duplicates and bad data.
As such, maybe the `btn-success` class should instead go to the other button below?

MonkeyDo: And for posterity, I'll add that in order to make the link-to-mb modal actually user friendly (and especially relevant on this page perhaps), we'll want to allow users to search for MB recordings rather than only expecting pasted MBIDs/MB URLs.
I guess we can revisit that on the back of #2331 which tackles some of the same questions about searching for a recording.

MonkeyDo (commented):
Looking good !
Small suggestion to push the use of one button over the other

PR #2322, opened 2023-01-04
Final YIM 2022 updates

amCap1712:
Remove stubs users and generate playlists, emails for all users. Switch default YIM page to year 2022.


mayhem (approved):
\ö/

PR #2325, opened 2023-01-04
fix yim playlists

alastair:
(creating a PR based on @amCap1712's work to be able to add a comment to it)

# Problem

Playlists were created with 2023 in the title/description instead of 2022

# Solution

Find the playlists and fix them

# Action


alastair:
I ran this locally on a copy of yim/playlists and it correctly updated the relevant data.

I checked and there are no users with '2023' in their name, so we're not going to change their name accidentally. there are no other cases of '2023' in the text that shouldn't be changed.

PR #2326, opened 2023-01-05
Add a note about not having a report for a user.

mayhem:
We accidentally sent out email to people who didn't actually have a report, so adding a small apology to the landing page is probable the best we can do for now.


MonkeyDo (approved):
Sure

PR #2328, opened 2023-01-05
Fix artist country query

amCap1712:
The existing query doesn't work if the artist area is already set to a country. Add another query for that case.


listenbrainz_spark/postgres/artist.py
```diff
@@ -13,6 +14,14 @@ def create_artist_country_cache():
             ON ac.descendant = a.area
           JOIN musicbrainz.iso_3166_1 iso
             ON iso.area = ac.parent
+         UNION
+         -- for the case where artist area is a country itself
```
alastair, line 18: nice comment!

alastair (approved):

PR #2330, opened 2023-01-06
Fallback to RG cover art if user submitted release doesn't have cover art

amCap1712:
Currently, we don't try to use the release group cover art fallback at all if the user provided a release mbid irrespective of whether that release has a cover art or not.


MonkeyDo (approved):
Fanks !

PR #2335, opened 2023-01-13
Playlist: search for user component

jivteshsingh:
Playlist add collaborator UI was not good as discussed with @hrik2001 and no api to search users registered on LB. So worked on this issue and implemented the changes.


frontend/js/src/playlists/SearchDropDown.tsx
```diff
@@ -0,0 +1,37 @@
...
+import ListenControl from "../listens/ListenControl";
+
+export type SearchDropDownProps = {
+  addCollaborator: (event: string) => void;
+  userSearchResults: Array<any>;
+};
+
+export type SearchDropDownState = {};
+
+export default class SearchDropDown extends React.Component<
```
MonkeyDo, line 13: I think this component should also be made more generic so that it can be used in other places where we will need to search for a user, and it should handle all the logic for searching for  and selecting a user.

I suggest to name the component `UserSearch` and have it handle both the API search *and* rendering the results (which you already do here).
It should have a callback prop `onUserSelect` instead of `addCollaborator` so that it is more generic.

I think you will make your life a lot easier by using the react-select library which we already use in other parts of the codebase.
It renders an input and the dropdown and everything else you need.
Have a look here for example: https://github.com/metabrainz/listenbrainz-server/blob/4b3fa44f7555c5ba8dc28e7d7cbf5b2a6c5b51c8/frontend/js/src/playlists/Playlist.tsx#L807-L821

frontend/js/src/playlists/CreateOrEditPlaylistModal.tsx
```diff
@@ -113,27 +123,53 @@ export default class CreateOrEditPlaylistModal extends React.Component<
...
+        url.searchParams.append("search_term", newCollaborator);
+        const response = await fetch(url.toString(), {
+          method: "GET",
+          headers: {
+            Authorization: `Token ${currentUser.auth_token}`,
+            "Content-Type": "application/json;charset=UTF-8",
+          },
+        });
+
+        const parsedResponse = await response.json();
```
MonkeyDo, line 154: This could be extracted as a separate API utility in the [APIService file](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/utils/APIService.ts).
It has the advantage for example that you can use `this. APIBaseURI` instead of hardcoding a url (looking at you, `http://localhost:8100` on line 144 😛  ), and it also has some error handling helpers.
You can use any of the methods in there as a starting point: https://github.com/metabrainz/listenbrainz-server/blob/4b3fa44f7555c5ba8dc28e7d7cbf5b2a6c5b51c8/frontend/js/src/utils/APIService.ts#L182-L202

MonkeyDo (changes requested):
Good start! Thanks for working on this feature!

If you have the patience and curiosity, together we can make this a much more powerful and useful component that will be reusable across the codebase.
In the future anyone needing a user search component will just be able to import your generic component and voilà ! This is clearly a useful tool that we are currently missing.
How does that sound?

The general theme of this review is: let's first make a generic UserSearch component completely separate from anything playlist related, which will manage it own state for user name input, search results, etc.
Once satisfied with it you will be the first to use the new component in the playlist modal.

listenbrainz/webserver/views/api.py
```diff
@@ -31,6 +31,26 @@
...
+
+@api_bp.route('/playlist/search/users/', methods=['GET', 'OPTIONS'])
+@crossdomain
+@ratelimit()
+def search_user():
+    search_term = request.args.get("search_term")
+    user = validate_auth_header()
+    user_id = user.id
+    if search_term:
+        users = db_user.search(search_term, SEARCH_USER_LIMIT, user_id)
```
hrik2001, line 44: Let us do this, as there can be need in the frontend/client to get more users than the default count value
```suggestion
        count = get_non_negative_param('count', SEARCH_USER_LIMIT)
        users = db_user.search(search_term, count, user_id)
```

frontend/js/src/playlists/UserSearch.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const value =
+      target.type === "checkbox"
+        ? (target as HTMLInputElement).checked
+        : target.value;
+    const { name } = target;
+    // @ts-ignore
+    this.setState({
+      [name]: value,
+    });
+  };
```
MonkeyDo, line 52: I see this is copied from another component, where multiple inputs and checkboxes use the same function.
In the case of this component, this handler can be rewritten like so:
```suggestion
  handleInputChange = (
    event: React.ChangeEvent<HTMLInputElement>
  ) => {
    this.setState({
      newUser: event.target.value
    });
  };
```

With that being said, we probably want to enhance this method so as not to call it on every keyboard stroke, which could slow down the application. We can use the throttle utility function from the lodash library (already installed) like it is done here:
https://github.com/metabrainz/listenbrainz-server/blob/e04252cc5b5213853991528de6254486f51b586c/frontend/js/src/brainzplayer/BrainzPlayer.tsx#L610-L613

In short: for any method `myFunction` on your class, define a new `throttledMyFunction` which will be the output of calling [lodahs's throttle](https://lodash.com/docs/4.17.15#throttle) with `myFunction` as the first argument and a number of milliseconds to wait between calls as the second argument (let's set this delay at 300ms).
Then your input will take `onChange={this.throttledMyFunction}`, or in our case `throttledHandleInputChange`.

frontend/js/src/playlists/CreateOrEditPlaylistModal.tsx
```diff
@@ -42,13 +39,15 @@ export default class CreateOrEditPlaylistModal extends React.Component<
...
     };
   }
 
   // We make the component reusable by updating the state
   // when props change (when we pass another playlist)
-  componentDidUpdate(prevProps: CreateOrEditPlaylistModalProps) {
+  componentDidUpdate(
+    prevProps: CreateOrEditPlaylistModalProps,
+    prevState: CreateOrEditPlaylistModalState
+  ) {
```
MonkeyDo, line 50: I believe this change can be reverted.

frontend/js/src/playlists/CreateOrEditPlaylistModal.tsx
```diff
@@ -200,66 +188,29 @@ export default class CreateOrEditPlaylistModal extends React.Component<
...
-                      className="btn btn-default"
-                      type="submit"
-                      onClick={this.addCollaborator}
-                      disabled={
-                        !newCollaborator ||
-                        (isEdit
-                          ? playlist?.creator.toLowerCase() ===
-                            newCollaborator.toLowerCase()
-                          : currentUser.name.toLowerCase() ===
-                            newCollaborator.toLowerCase())
```
MonkeyDo, line 256: This bit of logic would now be removed, and needs to be implemented in `addCollaborator`.
It prevents adding to the collaborators the creator of the playlist (for an existing playlist, see how `isEdit` is defined above), *or* the currently logged-in user (for a new playlist being created).

The reason for that is that a user should not be able to be in the collaborators if they are the owner of the playlist.

frontend/js/src/playlists/CreateOrEditPlaylistModal.tsx
```diff
@@ -200,66 +188,29 @@ export default class CreateOrEditPlaylistModal extends React.Component<
...
-                      }
-                    >
-                      <FontAwesomeIcon icon={faPlusCircle as IconProp} /> Add
-                    </button>
-                  </span>
+                <div>
+                  <label
+                    htmlFor="playlistcollaborators"
```
MonkeyDo, line 193: This refers to the name of an element that is not on the page anymore, so let's just remove it.

frontend/js/src/playlists/UserSearch.tsx
```diff
@@ -0,0 +1,99 @@
...
+    };
+  }
+
+  componentDidUpdate(prevProps: UserSearchProps, prevState: UserSearchState) {
+    const { newUser } = this.state;
+
+    if (prevState.newUser !== newUser) {
+      this.searchUsers();
+    }
+  }
```
MonkeyDo, line 37: Instead of reacting to a state change this way, we can instead directly call this.searchUsers after setting the state in handleInputChange.
The setState function takes an optional second argument which is a function to be called once the state has been set.
So in the input change handler, you'll be able to simply do this: `this.setState({newUser: event.target.value}, this.searchUsers);` and remove the componentDidUpdate entirely

MonkeyDo (changes requested):
This is looking much better, we're making great progress !
Nice work !

And thanks a lot to @hrik2001 for the assist :+1: ! We did previously make this NamePill and SearchDropdown components more generic so that they could be reused, specifically for this type of use, but I forgot about it 😅 

I have another round of improvements and optimizations requests for the front-end:

jivteshsingh:
@MonkeyDo Any changes u can suggest to improve the PR further

frontend/js/src/playlists/UserSearch.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState(
+      {
+        newUser: event.target.value,
+      },
+      () => {
+        this.searchUsers();
+      }
```
MonkeyDo, line 40:
```suggestion
      this.searchUsers
```

frontend/js/src/playlists/UserSearch.tsx
```diff
@@ -0,0 +1,96 @@
...
+    );
+  };
+
+  // eslint-disable-next-line react/sort-comp
+  throttledHandleInputChange = _throttle(this.handleInputChange, 300);
```
MonkeyDo, line 45: I think similarly to PR #2331 I gave you wrong instructions and this will not work well if you type quickly.
Instead of throttling `handleInputChange` we need to throttle the searchUsers and use `throttledSearchUsers` as the callback to setState in `handleInputChange`

frontend/js/src/playlists/UserSearch.tsx
```diff
@@ -0,0 +1,96 @@
+import * as React from "react";
+import { faUser } from "@fortawesome/free-solid-svg-icons";
+import { throttle as _throttle } from "lodash";
+import debounceAsync from "debounce-async";
+import GlobalAppContext from "../utils/GlobalAppContext";
+import ListenControl from "../listens/ListenControl";
+
+export type UserSearchProps = {
+  userClick: (event: string) => void;
```
MonkeyDo, line 9: What does this do? Is it this?
```suggestion
  onSelectUser: (userName: string) => void;
```

frontend/js/src/playlists/CreateOrEditPlaylistModal.tsx
```diff
@@ -113,28 +107,27 @@ export default class CreateOrEditPlaylistModal extends React.Component<
...
+    const { playlist } = this.props;
+    const isEdit = Boolean(getPlaylistId(playlist));
+    const { collaborators } = this.state;
+    const { currentUser } = this.context;
+    const disabled =
+      !user ||
+      (isEdit
+        ? playlist?.creator.toLowerCase() === user.toLowerCase()
+        : currentUser.name.toLowerCase() === user.toLowerCase());
+    if (collaborators.indexOf(user) !== -1) {
```
MonkeyDo, line 120: The whole if else block below can be rewritten into:
```
    if (!disabled && collaborators.indexOf(user) === -1) {
      this.setState({
        collaborators: [...collaborators, user],
      });
    }
```

MonkeyDo (changes requested):

package.json
```diff
@@ -70,7 +70,7 @@
     "rc-slider": "^10.1.0",
     "react": "^18.2.0",
     "react-bs-notifier": "^7.0.0",
-    "react-date-picker": "^9.1.0",
+    "react-date-picker": "^9.2.0",
```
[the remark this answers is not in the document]
jivteshsingh: some error occured should I revert it back :)

listenbrainz/db/user.py
```diff
@@ -572,3 +572,29 @@ def search(search_term: str, limit: int, searcher_id: int = None) -> List[Tuple[
...
+
+
+def search_user_name(search_term: str, limit: int, searcher_id: int = None) -> List[object]:
+
+    with db.engine.connect() as connection:
+        result = connection.execute(sqlalchemy.text("""
+            SELECT musicbrainz_id
+              FROM "user"
+             WHERE musicbrainz_id <% :search_term
+             LIMIT :limit
```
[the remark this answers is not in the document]
MonkeyDo: Answer: no, SQLAlchemy will sanitize the string.

listenbrainz/db/user.py
```diff
@@ -572,3 +572,30 @@ def search(search_term: str, limit: int, searcher_id: int = None) -> List[Tuple[
...
+           SELECT musicbrainz_id, similarity(musicbrainz_id, :search_term) AS query_similarity
+              FROM "user"
+             WHERE musicbrainz_id <% :search_term
+          ORDER BY query_similarity DESC
+             LIMIT :limit
+            """), {
+            "search_term": search_term,
+            "limit": limit
+        })
+
```
MonkeyDo, line 590: Looking at this now again now that I understand what is happening better and where the code comes from, can't we just use the existing `search` defined above? This code seems to just duplicate it for the most part (ignoring, unless i'm missing something.

MonkeyDo (commented):
The UI looks very good (tested on test.LB).
We had some back and forth about the SQL search, one more comment below:

listenbrainz/webserver/views/api.py
```diff
@@ -31,6 +31,25 @@
...
+def search_user():
+    search_term = request.args.get("search_term")
+    validate_auth_header()
+    if search_term:
+        users = db_user.search_user_name(search_term, SEARCH_USER_LIMIT)
+    else:
+        users = []
+    return jsonify(
+        {
+            'status': 'ok',
```
amCap1712, line 48: No need for status: ok.

listenbrainz/webserver/views/api.py
```diff
@@ -31,6 +31,25 @@
...
 DEFAULT_NUMBER_OF_PLAYLISTS_PER_CALL = 25
 
+SEARCH_USER_LIMIT = 10
+
+@api_bp.route('/search/users/', methods=['GET', 'OPTIONS'])
+@crossdomain
+@ratelimit()
+def search_user():
+    search_term = request.args.get("search_term")
+    validate_auth_header()
```
amCap1712, line 41: Normal search endpoint is public so no need for auth here either.

listenbrainz/db/user.py
amCap1712: Sounds good, to refactor the common part in a separate function.

MonkeyDo (approved):
Looks great, works great, it's good to go 🎉  !

Thank you for your hard work !

PR #2336, opened 2023-01-13
YIM22: coverflow: load lazy image

jahnvisahni31:
trying to fix issue with one image not lazy-loading

# Problem

there are variable with same name having different values and also having the error in the same properties in the line 114?

# Solution

i resolve the error by  giving the variable different names and also by using super properties in 114 line to describe the props variable 

# Action

 could you please provide also the  resources that are suitable for this?

### DIFF
--- a/frontend/js/src/year-in-music/2022/YearInMusic.tsx
+++ b/frontend/js/src/year-in-music/2022/YearInMusic.tsx
@@ -535,7 +535,7 @@ export default class YearInMusic extends React.Component<
                     lazy={{
                       enabled: true,
                       loadPrevNext: true,
-                      loadPrevNextAmount: 4,
+                      loadPrevNextAmount: 2,
                     }}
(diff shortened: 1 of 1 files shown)


MonkeyDo:
Hello @jahnvisahni31 !
Thanks for opening a PR, but I think something might have go wrong. You either forgot to commit you changes, or committed them on the wrong branch, or something along those lines, because I only see 1 single commit in this PR, and it is from myself :D
https://github.com/metabrainz/listenbrainz-server/pull/2336/commits

MonkeyDo:
Can I help you figure this out @jahnvisahni31 ?

MonkeyDo:
Unless I misunderstood something, this is most likely an error. The PR description refers to code and lines that are not in the one commit (which is one of my own commits).
Let me know if you need help opening a PR with your code.

PR #2338, opened 2023-01-16
Generate canonical release redirect only in the mb schema

alastair:
# Problem

When we have a separate mb and lb connection, some queries join with the musicbrainz tables, and therefore always need to be written to the musicbrainz connection, not the lb connection

# Solution

always write CanonicalReleaseRedirect to the mb database, like we already do with CanonicalMusicBrainzDataRelease

# Action

re-release mapper container and run a weekly update


alastair:
failing test appears to be unrelated, but we should check what dependency upgrade is causing it

mayhem (approved):
these PRs are getting a bit long.... :laughing:

PR #2340, opened 2023-01-17
Add option to select the database connection to dump canonical data from

alastair:
# Problem

In https://github.com/metabrainz/listenbrainz-server/pull/2157 we had one pending item:

- [x] Table generation scripts can dump into MB or LB database connection, if we want to keep this, we should modify the dump script to also be able to dump from either.

The initial version of the dump script only dumped from the musicbrainz connection, however when we create the canonical tables in the mapping container we add them to the listenbrainz connection (with the exception of the canonical release table which always goes in the musicbrainz connection, as we need to join against the release/release group tables to create it)

# Solution

duplicate the --use-lb-connection flag which is in the canonical creation command. Tables need to know if they should dump from the connection specified in the flag, or always from the mb connection

# Action

~~I don't have a setup with 2 separate db connections, so we might need to test this in prod with a testing container~~ pending the config problem below, this works and correctly generates a dump

Another pending item is to configure this management command in the cron infrastructure, probably adding it as a valid argument to `create-dumps.sh`, and then adding it to the crontab

### DIFF
--- a/admin/config.sh.ctmpl
+++ b/admin/config.sh.ctmpl
@@ -29,9 +29,11 @@ RSYNC_FULLEXPORT_PORT="{{template "KEY" "rsync_fullexport_port"}}"
 RSYNC_FULLEXPORT_DIR="$FTP_DIR/fullexport"
 RSYNC_INCREMENTAL_DIR="$FTP_DIR/incremental"
 RSYNC_SPARK_DIR="$FTP_DIR/spark"
+RSYNC_MAPPING_DIR="$FTP_DIR/mapping"
 RSYNC_FULLEXPORT_KEY='{{template "KEY" "user_home"}}/.ssh/rsync-listenbrainz-dumps-full'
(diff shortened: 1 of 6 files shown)


alastair:
The `MB_DATABASE_URI` config can't work for performing a dump of the canonical release table, as we currently connect using the `musicbrainz_ro` user, which doesn't have permission to read from the `mapping` schema.

The dumps are made by a "mbid_mapper" user to the same server, which is configured in the mbid_mapping consul_config file.

For the case of this dump, it's probably enough to just call

```py
musicbrainz_db.init_db_engine()
```

again before we make the dumps: https://github.com/metabrainz/listenbrainz-server/blob/4b49ecc241211a022aaf253afc015b75581229c5/listenbrainz/db/mapping_dump.py#L186

But there is a question as to how we get this config item, as it's not in the current config file. I see our options as 
1. take config.MB_DATABASE_URI and do `.replace("musicbrainz_ro", "mbid_mapping")`
2. Add yet another musicbrainz config parameter to the config file for dumps

amCap1712 (approved):
LGTM, Thanks!

PR #2343, opened 2023-01-24
Combine pins and love/hate tabs

MonkeyDo:
These are the first steps of the Great ListenBrainz Redesign, some structural improvements in which we combine multiple pages that belong together ina  single page.

On today's menu, The Pins and Love/Hate tabs are combined into a 'Taste' tab:
image

We create a new UserTaste component and corresponding webserver endpoint to combine the previous functionality into one.

Also found a bug (LB-1221) and fixed that


mayhem (approved):
While this is less gooblegegook to me now, I can't say I've done an adequate job of reviewing your code. But it doesn't look bad! Perhaps best to get another review.

mayhem (commented):
I'm really struggling to follow all of this, TBH. I take it that there is very little new code, yes? mostly just shuffling bit about? If so, then I'm included to accept this and help you test the code to make sure its all good.

MonkeyDo:
> I'm really struggling to follow all of this, TBH. I take it that there is very little new code, yes? mostly just shuffling bit about? If so, then I'm included to accept this and help you test the code to make sure its all good.

Indeed no new code, just consolidation. It's well removed from the React stuff you've been doing.
I mainly asked for review for help with testing, and also so that you can follow along at your leisure/get some insight into the current codebase.

However we have a more general UX question that deserves a discussion: as you can see in the screenshot, at the bottom, we now have two sets of navigation buttons (prev/next). The UserFeedback and UserPins components both handle navigation, API calls, browser history/URL  updates independently. I'm pretty sure this is going to interact badly.

For example in the current love/hate page we update the URL with `?page=2` upon navigation to be able to share specific pages, and the back-end ~~accepts these query parameters to send the right data to the front-end~~ is supposed to load page 2 but that looks broken at the moment (it expects `offset` and `count` parameter instead of `page`).
But does that mean page 2 for the love/hate or page 2 for the pins?

I propose to keep only one set of navigation buttons, define a fixed number of items per page (users won't be able to change that) and use a `?page=X` query parameter to be able to share a specific page.
That seems like the obvious approach to me.
Now what happens if I only have a few pins but many loved tracks and I load page 2? Considering there's no page 2 of pins, what do we want to show? Page 1 of pins anyway? Nothing/a message?

On a side note, these are problems that would be solved by using an infinite scroll mechanism, but is that what we want? Users wouldn't be able to share a specific page for example.

Paging @Aerozol in this discussion since this is a UX question.

mayhem:
Not sure I have good feedback on the UX question. But aside that, lets get this PR put up on test.lb.org for testing, shall we?

Aerozol:
This is an improvement no matter what imo. I am quite into an infinite load here now that you’ve mentioned it…
Is sharing a set/page of pins or whatever that much better anyway? If anything someone would want to share a specific pin I imagine. We should let them do _that_ eventually! (and generate a shareable image and tweet and so forth #social)

MonkeyDo:
I think infinite scroll could be a good option for this page, but for now I kept it simpler with a simple "load more…" button instead of the pagination buttons:
image

MonkeyDo:
I'm seeing a small issue (unrelated to this PR) with the API endpoint to fetch feedback: The total_count returned by the API endpoint is for both love and hate feedback, even if only 'love' feedback is requested:

Asking for 100 items, get only the 66 existing items returned, total_count = 171…
https://test-api.listenbrainz.org/1/feedback/user/rob/get-feedback?offset=0&score=1&count=100
image

mayhem (approved):

PR #2344, opened 2023-01-25
Combine playlists and collaborations tabs

MonkeyDo:
Part two of the Great ListenBrainz Redesign, with another obvious improvement combining two separate tabs of playlists into one (user's playlists and collaborative playlists).

~~I'm not sure how the pagination is going to work in this case, where we use the `count` and `offset` query parameters on the back-end and the front-end (i.e. browser navigation triggers API calls with the query parameters in question):~~
- ~~do we want separate query parameters for offset and count for each list?~~
- should we remove the count parameter, and impose a set number of results we want to show (this probably makes sense with two lists on the same page) <- went for this solution
- ~~do we want to send *all* the playlists for a user in one go in the props, and deal with pagination entirely on the front-end?~~  <- maybe this would be another solution to look at

I ended up using pills at the top of the page to switch from owned and collaborative playlists.
That seemed the easiest way to keep a clean page (not stacking two paginated views on top of each other) while maintaining the design choices of current and redesigned pages:
image


MonkeyDo:
Pending resolution of discussion about pagination, which I opened in #2343 instead of here

listenbrainz/webserver/views/user.py
```diff
@@ -295,70 +288,18 @@ def recommendation_playlists(user_name: str):
...
-        raise BadRequest("Incorrect int argument count: %s" %
-                         request.args.get("count"))
-
-    user = _get_user(user_name)
-    user_data = {
-        "name": user.musicbrainz_id,
-        "id": user.id,
-    }
-
-    include_private = current_user.is_authenticated and current_user.id == user.id
```
mayhem, line 336: This line was not re-added and will cause failures.

mayhem (changes requested):
See comment, will currently crash and burn.

listenbrainz/webserver/views/user.py
MonkeyDo: It's in there, just hidden behind code folding:
https://github.com/metabrainz/listenbrainz-server/pull/2344/files#diff-e91083bcd314135e0a5ef06f4ab7e4121bf1f2ad9360b570e1e2dd6033a75bf3R220

mayhem: Shit, totally missed the fold. Sorry.

mayhem:
Looking at this feature on test, the playlist page has a problem that the user has no idea that collaborations are at the bottom of the page. We need to add some sort of link or hint before we release this.

mayhem (approved):

PR #2345, opened 2023-01-27
LB-1201: YIM 2022 - wrong artist in top tracks image

amCap1712:
:facepalm:


mayhem (approved):
Took me a moment find that!

PR #2347, opened 2023-01-31
Various mapping fixes

mayhem:
This PR fixes two separate issues:

- Make stop word removal the same for mbid mapping and mapping explain
- Remove known BS from recording names before looking them up. First up: "- YYYY remaster" at the end of recording names.


amCap1712 (approved):
LGTM, Thanks!

PR #2348, opened 2023-01-31
Return artist mbids as comma separated instead of postgres array format

alastair:
Without the cast multiple MBIDs will be returned from the query as

    {mbid,mbid}

which means that you would need to strip the {} before using it. This change makes it display as

    "mbid,mbid"

and the field is correctly quoted even when dumping to CSV


amCap1712 (approved):
LGTM.

PR #2349, opened 2023-01-31
Update to python 3.11 and required dependencies

alastair:
# Problem

There's a new python out!

# Solution

Upgrade it!

# Action

I only updated the packages that actively failed during the build process. I've used a python 3.11 image to run the data dumps/import (and saw a ~10% speed increase), but I haven't run the website or all of our management commands to check that they work fully. This could be something to put into beta first.


amCap1712 (approved):
LGTM, Thanks

PR #2350, opened 2023-01-31
Upgrade actions versions

alastair:
# Problem

Some actions show warnings when they are run. Checkout v2 says that it's running on an old version of node which is no longer supported.
The docker actions show a number of warnings about features that will be disabled some time in 2023.

# Solution

Upgrade to the latest version of actions, according to their documentation

# Action

The changes from checkout v2 to v3 should be fine - we already have some actions using checkout v3 anyway.
We'll need to run a prod build to test the upgrades to the docker actions - I didn't check to see if any of the parameters have changed or not between versions

### DIFF
--- a/.github/workflows/build-prod-image.yml
+++ b/.github/workflows/build-prod-image.yml
@@ -28,7 +28,7 @@ jobs:
       if: github.event.pull_request.draft == false
 
       steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Login to Docker Hub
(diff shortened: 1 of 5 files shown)


alastair:
Looking at the test logs, it looks like there's a failure when building uwsgi when running tests. 
I'm not sure why that failure happens here on github and not on my local machine. We could try and upgrade it, though I don't know if that'll have any effect...

amCap1712:
Error resolved itself after rebase.

amCap1712:
LGTM, Thanks!

PR #2351, opened 2023-02-01
LB-1192: Improve artist mbid stats

amCap1712:
Instead of aggregating artist stats on artist credit, aggregate on artist mbid. As of now "A ft. B" is counted as 1, going forward it
will be counted as A - 1 and B - 1. With this change, we can use artist data from MB to improve stats. We replace the artist's name/aliases with the canonical name from MB.


mayhem (approved):
I'm excited to see this get done! Our data foo is deffo getting stronger! Well done!

PR #2352, opened 2023-02-02
Use artist credits for artist similarity

amCap1712:
Earlier we used to explode artist credit mbids and then calculate similarity based on occurrences of individual mbids. Changing that to calculate similarity based on occurrences of entire array. Once similarity is calculated explode to individual mbids.


mayhem (approved):
Heh, you caught me mid-review. :) Can't wait to see this in action -- can we do a release soon and generate updated date?

PR #2353, opened 2023-02-03
Create mbcanonical data dumps twice a month

alastair:
# Problem

We have newly created mb canonical data dumps. We decided to generate them twice a month.
I chose the 3rd/17th so that it doesn't conflict with the full dumps, however this dataset takes less than 10 minutes to generate.

The cron job to re-generate our mapping tables runs in the mbid_mapping container, and happens each day at 4am. In my experience this only takes a few hours to run, so I set this job to run at 8am. If we think that this isn't enough time to get "today's" dump in place, we could delay it further.

Alternatively we could just run it at 1am, and use the data from the day before.


mayhem (approved):
I think this approach makes sense. If not, we can always revise it.

amCap1712 (approved):
LGTM, Thanks!

PR #2354, opened 2023-02-03
LB-1220: Add links to user pages when listing users in /admin

amCap1712:
flask-admin exposes column formatters property to customize display of column data. Use this to render a link to user's page.


mayhem (approved):
Thanks for knocking that out!

PR #2362, opened 2023-02-20
Youtube player: remove autoplay and minimize features

MonkeyDo:
As requested by the Youtube team for approval of our quota extension, we need to remove the functionality to minimize the player as well as require a user interaction (read: user must click the YT player) before playing any video (no autoplay mechanisms).

This is unfortunate for the functionality of BrainzPlayer as it probably means the end of continuous playback where YT is concerned.

That might also mean that if a user is logged into Spotify, we might not want the fallback to Youtube if the track can't be found in Spotify, because that would interrupt the continuous playback


MonkeyDo:
Deployed on test.LB , but we are currnetly out of quota which makes it hard to test.

PR #2363, opened 2023-02-20
Taste and playlists pages: remove sub nav text

MonkeyDo:
What triggered this change is the new Taste page, where the header wasn't very clear. It's not obvious that it should be read "Tracks rob loved". We discussed adding a colon to make it clearer, but a) it won't work in all pages and b) do we even need this bit of text anyway considering the user name is written just above the tabs?
Before:
image

After:
image

As for the playlists page:

Before:
image

After:
image


Aerozol:
Looks good, +1 to taking out the username.
Perhaps we still want to leave ‘Playlists’ (even though it’s duplicating the tabs) to also be consistent with pages where we have multiple sections on the page. But on reviewing the pages the only outlier for that might be Taste > Tracks you [loved] [hated] (reports and charts both just have the buttons, no text)
So perhaps yes to taking it all out (and take out the header on the Recommendations page as well?)

MonkeyDo:
Thanks, I'll keep the recommendations page as is for now, I'm not 100% convinced it doesn't bring anything. Same for the home page and its "Recent listens".
But I'll leave that for another PR, with more page redesigning to come.

PR #2364, opened 2023-02-21
Add lookups to allow choosing a particular release for mapping

amCap1712:
This PR adds all necessary tables and labs API queries to support choosing a particular release while mapping listens to recordings. There are 2 steps pending to enable full release mapping support:
1) Use the new endpoints in the mapper and figure a way to store the mapped release mbid
2) Figure out the best way to retrieve the correct release metadata for a match (currently we only store canonical release data in mb_metadata_cache table).


mayhem (approved):
The mapping code is getting a bit complex -- I hope the detuner will allows us to make it flow a little better. But, I am very glad this is now done, thanks!

PR #2366, opened 2023-02-22
Improve MbidMappingModal

MonkeyDo:
The new-ish MBID mapping modal could do with a few improvements.

In particular I wasn't convinced by the way the name of the track-to-be-linked was displayed, both its styles and the fact that it was not clear if that was the track to link or the selected MB recording:
image

I clarified the situation a bit and added a lookup from MusicBrainz API and used our trusty ListenCard component to display the selected recording:

image


MonkeyDo:
Paging dr. @Aerozol for a look-see at the screenshots in the PR description :)

In particular, I'm not a huge fan of the "Linking: Track name — Artist Name" but I didn't have a better idea on the spot.
I think it is useful information to make sure that you are linking the right unmatched Listen to the right MB recording.

Aerozol:
The dr. is in!
Not sure if this is out of scope for what you’re wanting to do here, but I’ve made some mockups that align more with the ‘add-listen’ modal, which we already spent a bit of time on.

1.
image

2.
image
I don’t know about what kind of information you want to include here, but this looks okay to me. Date would be very useful to me but I think it will mainly confuse people (who will be expecting the original date).
I wasn’t able to fully align this with the add-listen modal, we can make changes there to align it with this, some day - thinking out loud here.

I also gave the explanatory text a bit of an overhaul while I was ‘in there’, because it would be gibberish to a new user imo. Here’s what I imagined the (?) text to be:
image

p.s. The add mapping button is a pixel or two below cancel :P

MonkeyDo:
Thanks for the review !
I went about halfway to what you proposed, mainly kept your layout.
I flipped the listen-to-link and the selected MB recording, that made more sense to me.

I'm also leaving out the extra release and label information because I don't want to introduce a new format of Listen card with extra details before we've figured out if that's how we want to display extra info in other parts of the website (and also because it means extra API calls to get label info for example).

Here's some screenshots:
Initial view:
image

Form validation:
image

Recording selection:
image




I wasn't 100% sure we want to use the full blue listen card, since this is how we show that a listen is pinned (on user page), but I made a screencap anyway so we can compare:
image

Aerozol:
Beautiful! Default listen card is much nicer and clear imo, nice one.?
My only concern is that the cover art may not match what people are adding (even though it’s the correct recording). But maybe let’s leave it for now, it looks nice!

It will be interesting to revisit add-listens some day and see what we want to align with this.

PR #2368, opened 2023-02-27
LB-1003 : Playlist page html title isn't descriptive

Shelly011s:
# Problem
The html <title> for a user's playlist page says
User "alastairp" - ListenBrainz
which doesn't make sense, it should have at least some mention of playlists.

# Solution

I would like to propose the title to be "<User name> - Your Playlists ". And the same pattern  follows through for all the tabs in the profile.

### DIFF
--- a/listenbrainz/webserver/templates/playlists/playlists.html
+++ b/listenbrainz/webserver/templates/playlists/playlists.html
@@ -1,5 +1,5 @@
 {% extends 'user/base.html' %}
-
+{%- block title -%}{{ user.musicbrainz_id }} - Your Playlists{%- endblock -%}
 {% block profile_content %}
 
     <div id="react-container"></div>
--- a/listenbrainz/webserver/templates/explore/huesound.html
+++ b/listenbrainz/webserver/templates/explore/huesound.html
@@ -1,5 +1,5 @@
 {%- extends 'base-react.html' -%}
-
+{%- block title -%}Hue sound - ListenBrainz{%- endblock -%}
 {% block content %}
 
     <div id="react-container"></div>
--- a/listenbrainz/webserver/templates/playlists/recommendations.html
+++ b/listenbrainz/webserver/templates/playlists/recommendations.html
@@ -1,5 +1,5 @@
 {% extends 'user/base.html' %}
-
+{%- block title -%}{{ user.musicbrainz_id }} - Your Recommendations{%- endblock -%}
 {% block profile_content %}
 
     <div id="react-container"></div>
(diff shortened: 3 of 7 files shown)


MonkeyDo:
Thanks for opening this PR !

I think it makes sense to have the user name when I am not looking at my own playlists (or other pages following the same format) rather than 'user name - Your Playlists'.

So I would suggest 'Your playlists - ListenBrainz' If I'm logged in and looking at my own page, and 'Shelly's playlists - ListenBrainz' if I'm looking at your playlists.

That's just my opinion though, @mayhem @amCap1712 @Aerozol  what do you think?

mayhem:
Sounds sensible to me, @MonkeyDo

Aerozol:
Great ticket shelly011s.

Looks good to me too monkey, so then we would follow this format:
‘Your charts - ListenBrainz’, ‘Your listens - ListenBrainz’…
‘Mayhem’s charts - ListenBrainz’, ‘Mayhem’s listens - ListenBrainz’...

listenbrainz/webserver/templates/playlists/playlists.html
```diff
@@ -1,4 +1,12 @@
 {% extends 'user/base.html' %}
+{%- block title -%}
+  {% if current_user.is_authenticated %}
+  Your 
+  {% else %}
+  {{ user.musicbrainz_id }}'s
+  {%- endif -%}
```
MonkeyDo, line 7: This isn't exactly what we need.
`current_user.is_authenticated` will be true if a user is logged in. What we want this condition to check is if `current_user.musicbrainz_id == user.musicbrainz_id`

Of course this will apply to other pages as well.

MonkeyDo (changes requested):
OK, we're getting there!
This won't exactly do what you want. I know it's a bit tricky to test that stuff locally, be here is what I can suggest:
you can create a test account on MusicBrainz that you use to have a second user in your local setup. (for example, [mr_monkey_test_account](https://musicbrainz.org/user/mr_monkey_test_account))
That way, you can create another account in local dev and navigate to the test account's page and check if the page title shows you what you expect. In this specific case, it would currently show your own account name even when looking at the test account's pages.

listenbrainz/webserver/templates/playlists/playlists.html
Shelly011s: Got it. I will implement the suggested. Thank you for the help.

MonkeyDo:
I took the liberty to apply a couple of changes following some restructuring/renaming I did in #2378 ('reports' and 'charts' are now both called 'stats')

MonkeyDo (approved):

PR #2369, opened 2023-02-28
LB-1229: Removed Double-escaped HTML entities in Blog

Cronus1007:
# Problem
Double-escaped HTML entities in "News and Updates" section on front page
# Solution
Added a JacaScript function to decode HTML Entities.
# ScreenShots
Screenshot from 2023-02-28 08-06-22


mayhem:
@Cronus1007 We see your patch and thank you for submitting it. But pestering us to review it isn't very helpful.

mayhem (commented):
Thank you for pasting this stack overflow answer into a PR:

https://stackoverflow.com/questions/5796718/html-entity-decode

mayhem:
This is a very good example of how not to write code -- you understood the problem to hand, but didn't actually look to the shortest way to implement it: Namely using native Javascript functions (or react functions or some NPM module) to do this. Instead you opted to write copy and paste code from someone else, passed it off as your own work. This code could be buggy on top of all of this. To say the least, I am not very impressed by this PR.

MonkeyDo:
For posterity, there is no native javascript function do do this, and the recommended solution is to use this library: https://github.com/mathiasbynens/he

Cronus1007:
Hi @mayhem , I sincerely apologise for taking help from the available sources to solve this issue so as to both save time and achieve the desired result. The issue was quite vague as in it did not talk about using any particular native Javascript functions (or react functions or some NPM module) to achieve the desired result. It was, therefore, obvious and rational to use a method that would be concise and readily available and understandable. 

I've implemented the solution [here](https://github.com/metabrainz/listenbrainz-server/pull/2374) as recommended by @MonkeyDo. Let me know how does this look to you. Thanks for the much appreciated constructive feedback.

elomatreb:
I feel like the proper solution here would be to make the `/blog-data` endpoint not return HTML-escaped strings in the raw data.

PR #2372, opened 2023-02-28
Explore page

mayhem:
This PR implements the explore page, which is a simple sign-posting web page that shows an element for each of our explore features as per the figma mock-up.

The mock-up doesn't quite make it clear how the page should behave responsively, so I've implemented a simple system that works on mobile and desktop. The next goal is to get feedback on this PR on the test server so that we can improve the UI to @Aerozol's satisfaction.

Also, given that we have at most 5 explore features, all of the user interface elements for starring/filtering/searching have not been implemented since they would just confuse the page at this point.


Aerozol:
Looks great! Doesn’t need to be any more complex at this stage.

My notes are:
- Could use some more interesting images. Maybe they don’t have to be representative? What do you think, shall I do some custom images that look inviting and aren’t screenshots (like a splash of colour for hue sound)
- The image overlay makes things look low res doesn’t it? The one used in the mockup is complex but looks nice, with a bit of texture. You could try overlaying the svg: https://drive.google.com/file/d/1qZzZhgdzFAjhztABYyGatBrXurnLthqw/view?usp=sharing. Otherwise perhaps we get rid of the ‘fuzz’.
- Sorry I didn’t do responsive mockups, too much to do… I think it’s worth getting these right early. Here’s my mockup at different screen sizes, let me know what you think: https://www.figma.com/file/sy8cAelERQTybS2Hh0YckG/ListenBrainz-Explore—Responsive?node-id=0%3A1&t=JYxGSueoV2OJrf17-0 (two comments in there for you, atm)
- I went for ‘Roboto Black’ (max weight) for the titles, let’s try that. Monkey may feel this differs from existing styles too much but let’s see
- Can we make them a touch more compact vertically

mayhem:
OK, I've updated the code with your feedback -- please take another look and see what else needs changing.

Aerozol:
The size breakdowns don’t match the sizes in the mockup - it’s 3 column down to 900px width, then 2 column down to 600px
Testing it on my phone now as well, I think it will be nicer with those columns/sizes.

New images: https://drive.google.com/drive/folders/1StDKqhj8vNGe4WplK6LbwtZCUe2s9zew?usp=share_link
Happy to workshop these more, some I am very happy with, other less so.

We’ll need a button for YIM 2022 (image included in the link above)

FYI on a very narrow window the gradient overlay drops down over the text, but I don’t know if any modern phones go that small so it possibly doesn’t need fixing (< 350px width)

Aerozol:
Forgot to mention - Do we want people to still be able to visit YIM 2021? I wasn’t sure if that’s possible, which is why I supplied both 2021 & 2022 icons.
While I’ve got you, can we centre the images? Looks great otherwise!

mayhem:
OK, now with centered images and correct breakpoints!

frontend/js/src/explore/ExploreIndex.tsx
```diff
@@ -0,0 +1,125 @@
...
+import BrainzPlayer from "../brainzplayer/BrainzPlayer";
+
+type ExplorePageProps = {
+  name: string;
+  desc: string;
+  img_name: string;
+  url: string;
+};
+
+function ExploreIndex(props: ExplorePageProps) {
```
MonkeyDo, line 21: There's a small structural refactoring that I'd like here, to follow our style in the rest of the page components:

This component should be called something like `ExploreCard` with its `ExploreCardProps`, and you should create another component `ExploreIndex` or `ExplorePage` which contains what is currently at lines 68->121 (https://github.com/metabrainz/listenbrainz-server/pull/2372/files#diff-9de970bd4956e4d3e4f0a65bd19527cb4baea5250dd466c3325abbfbf520cbbaR68-R121)

You'll end up with this structure:
```
function ExploreCard(props: ExploreCardProps){
…
}

function ExplorePage(){
…
[this renders a bunch of ExploreCards]
}

[...]
renderRoot.render(
    <ErrorBoundary>
      <GlobalAppContext.Provider value={globalProps}>
        <ExplorePage />
      </GlobalAppContext.Provider>
    </ErrorBoundary>
)
```

That keeps the part in the DOMContentLoaded listener mostly separate from the implementation details of that page, allowing us to reuse that page component later on without the need to refactor it.

MonkeyDo (changes requested):
Looks good, nice job !
Small change requested below

frontend/css/explore.less
```diff
@@ -0,0 +1,61 @@
...
+  float: left;
+  padding: 0 15px;
+
+  @media (min-width: 600px) {
+    width: 50%;
+  }
+  @media (min-width: 900px) {
+    width: 33%;
+  }
+}
```
MonkeyDo, line 12: For posterity, I think if we ever revisit this, we should look into CSS grids instead, which seem to be an ideal candidate for this type of layout.
Not something we should do right now though !

frontend/js/src/explore/ExploreCard.tsx
```diff
@@ -0,0 +1,133 @@
...
+  img_name: string;
+  url: string;
+};
+
+function ExploreCard(props: ExploreCardProps) {
+  const { name, desc, img_name, url } = props;
+  return (
+    <div className="explore-card-container">
+      <div className="explore-card">
+        <a href={`${url}`}>
```
MonkeyDo, line 27: We can skip the template string extrapolation here and pass a string directly.
Same goes for line 39 below
```suggestion
        <a href={url}>
```

MonkeyDo (approved):
Looks a lot better, thanks ! I missed a small spot last review, soze !

Apart from that, consider it ready to merge !

PR #2373, opened 2023-02-28
The Great ListenBrainz Redesign

MonkeyDo:
Following the [redesign Figma mockups](figma.com/file/YRbCOtFHBez8XmMdCKbGta/ListenBrainz-redesign), this is the first big redesign block which addresses the navbar to the side and the subnav tabs at the top.

image

image
image


MonkeyDo:
Pending one small modification to add follow button and MB profile button on users' profile page

PR #2376, opened 2023-03-01
Fixed Documentation of User Timeline APIs (user_timeline_event_api.py)

07jasjeet:
# Problem

In the following API documentation link, there seems to be lack of information about request headers.
Also the `POST /1/user/(user_name)/timeline-event/create/recommend-personal` API had its example request data glitched (not represented as code blocks as expected).
https://listenbrainz.readthedocs.io/en/latest/users/api/social.html

# Solution

FIXEM

# Action

Added request header fields wherever 401 is thrown, added appropriate spaces and made empty line deletions to normalize documentation format.

### DIFF
--- a/listenbrainz/webserver/views/user_timeline_event_api.py
+++ b/listenbrainz/webserver/views/user_timeline_event_api.py
@@ -73,9 +73,10 @@ def create_user_recording_recommendation_event(user_name):
             }
         }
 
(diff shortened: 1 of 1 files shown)


mayhem (approved):
Thanks for fixing our docs, much appreciated!

PR #2378, opened 2023-03-02
Remove "Charts" tab from user page

MonkeyDo:
Following in the footsteps of #2344 and #2343 in preparation for the Grand Redesing, losing another not-so-useful tab

Users (and the team as well) have been reporting that the charts tab is confusing as it seems to duplicate some of the information from the "Reports" tab.
It's also a not-very-usable page whoe sole reasons for existing are: 1. pagination and 2. playable listen cards

This PR tackles 2. by making the top entities in the "Reports" tab playable.
We need to keep the charts page because we currently don't have a good solution for 1., but it can be a hidden page with no direct access other than the "View more…" buttons/links on the Reports tab:

Removed another tab:
image

In the process, I had to make a couple of CSS changes to improve our compact ListenCard

The Reports page before:
image

The Reports page after:
image


MonkeyDo:
Updated screenshot with the user and global pills in the sub-nav, to the right, and renamed "statistics" tab:
image

Yes, I know, the top artists card is not the same height as the other ones. But it's not so easy to solve and required making each card a flex item, etc.
Will keep this for another PR.

mayhem (approved):
Great job! I can't wait to see this in production!

MonkeyDo:
Just noticed that the tracks in the top tracks are not playable… something else to fix.

mayhem:
We'll hotfix both at the same time. :)

PR #2379, opened 2023-03-03
Only show "Add listen" button on user's own page

MonkeyDo:
We missed that on the first pass, and users reported their confusion on Discord and IRC.

Small refactor to keep it DRY.


amCap1712 (approved):
LGTM, Thanks

PR #2391, opened 2023-03-09
Add & to detunings.

mayhem:
Track artists may include feat artist and the mapping detuner was used to remove those, but we are not removing " & " which would help "Peter Gabriel & Kate Bush" " Don't Give Up" to get matched when looking up "Peter Gabriel" "Don't Give Up".

### DIFF
--- a/listenbrainz/mbid_mapping_writer/mbid_mapper.py
+++ b/listenbrainz/mbid_mapping_writer/mbid_mapper.py
@@ -86,7 +86,7 @@ class MBIDMapper:
             after it is removed.
         """
 
-        strings = ["(", "[", " ft ", " ft. ", " feat ", " feat. ", " featuring ", " - "]
+        strings = ["(", "[", " ft ", " ft. ", " feat ", " feat. ", " featuring ", " - ", " & "]
         if is_artist_credit:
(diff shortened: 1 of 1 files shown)


listenbrainz/mbid_mapping_writer/mbid_mapper.py
```diff
@@ -86,7 +86,7 @@ def detune_query_string(self, query, is_artist_credit):
             after it is removed.
         """
 
-        strings = ["(", "[", " ft ", " ft. ", " feat ", " feat. ", " featuring ", " - "]
+        strings = ["(", "[", " ft ", " ft. ", " feat ", " feat. ", " featuring ", " - ", " & "]
         if is_artist_credit:
             strings.insert(0, ",")
```
amCap1712, line 91: I think we only want to do this detuning for artist credits so it makes sense to to put it inside the if. otherwise seems fine,

amCap1712 (approved):

listenbrainz/mbid_mapping_writer/mbid_mapper.py
mayhem: Ah good catch.

PR #2394, opened 2023-03-10
Increase listen queue and increase nuber of running clients

mayhem:
This morning listenbrainz was down for a while because:

  *** uWSGI listen queue of socket "0.0.0.0:3031" (fd: 3) full !!! (101/100) ***

So I increased the number of listening connections to 1024 (recommended) from the default 100. Also increased running clients to 100.


zas (approved):
LGTM, it makes sense with the increasing traffic.

PR #2398, opened 2023-03-13
LB-1170 (IV): Replace SimpleModal with ListenPayloadModal

MonkeyDo:
in #2240 I created a reusable modal component (SimpleModal), that would sit alongside the top-level components and be callable via a reference in the global context.
I was trying to fix this issue we have with modals needing to be defined/created in each top-level component in order to be used, resulting in missing functionality throughout the website.
As described in more detail in #2386 the NiceModal library aims to solve this exact problem, and does it in a much better way than my naïve attempt at the same issue.

And so, we say goodby to SimpleModal, so far used only for syntax highlighting of a listen's JSON, and never reused for anything else partly because it was still too cumbersome to use. You won't be missed.


MonkeyDo:
The meat of this PR is in frontend/js/src/listens/ListenPayloadModal.tsx and frontend/js/src/listens/ListenCard.tsx
The changes in the rest of the files are just about removing the SimpleModal component.

amCap1712 (approved):
LGTM, Thanks!

PR #2399, opened 2023-03-13
LB-1170 (V): Refactor MBIDMappingModal with NiceModal

MonkeyDo:
Continuing the series of reimplementing modals properly (#2386 #2387 #2390 #2398) with the MBIDMappingModal.
Like the other ones, the job starts by refactoring the modal class component into a functional component.

This is also relevant to ticket LB-1148 and the idea of making the ListenCard cover art placeholder open the linking modal for unlinked listens and show a "link" icon instead of the default placeholder (CAA logo)

We need to be able to open a mapping modal everywhere; currently the modal is only created in the "Listens" (user profile) and "Missing MusicBrainz Data" pages.


amCap1712 (approved):
LGTM, Thanks

PR #2400, opened 2023-03-14
Fix lint-to-MB in RecentListens page

MonkeyDo:
Fixup of PR #2399 where I forgot to modify the RecentListens page to remove the old SimpleModal system.
Instead, the button is always available in the ListenCard menu options *if* user is logged in and the listen has a recording MSID.
While doing testing however, I realized the latest listens we get (from redis?) to populate the RecentListens page never have a recording MSID. I think it is added later in the pipeline.
Consequently, we will basically never have the option to map listens in RecentListen to an MB recording it looks like.


amCap1712 (approved):
LGTM, Thanks

PR #2402, opened 2023-03-14
Rework listen metadata maintenance

amCap1712:
Currently, an hourly cron job scans all listens interested since its last run and updates the metadata (min_listened_at, max_listened_at, count) for all users. At the same time, listen deletion runs and updates the metadata accordingly. These jobs are a bit slow and these queries also showed up in recent postgres errors causing LB downtime in past week.

Therefore, replacing half of the mechanism with update at time of listen insertion. Listen deletion still happens with an hourly cron job. Hoping this does not affect overall speed of listen insertion by much but brings down cronjob runtime.

### DIFF
--- a/docker/services/cron/crontab
+++ b/docker/services/cron/crontab
@@ -3,7 +3,7 @@ MAILTO=""
 # NOTE: This file is organized in chronological order of the start hour of each job, keeping jobs spread over the night and not overlapping.
 
 # Delete pending listens and update our listen counts, timestamps hourly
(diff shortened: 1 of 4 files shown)


mayhem (approved):
I have slight reservations about this PR since it add more work to the most common operation (insert listens) as opposed to adding more work to the least used operation (deleting listens). 

But, this PR is done, so lets give it a shot. But, please lets keep monitoring this for a while.

PR #2403, opened 2023-03-15
Return 204 status in case of insufficient data to create svg

amCap1712:
Currently, there is a 500 error in case of insufficient data. Handle it and return a 204 No Content instead.

### DIFF
--- a/listenbrainz/webserver/views/art_api.py
+++ b/listenbrainz/webserver/views/art_api.py
@@ -225,6 +225,11 @@ def cover_art_custom_stats(custom_name, user_name, time_range, image_size):
 
 def _cover_art_yim_stats(user_name, stats):
     """ Create the SVG using YIM statistics for the given year. """
+    if stats.get("day_of_week") is None or stats.get("most_listened_year") is None or \
+        stats.get("total_listen_count") is None or stats.get("total_new_artists_discovered") or \
+        stats.get("total_artists_count") is None:
+        return None
+
     match stats["day_of_week"]:
         case "Monday": most_played_day_message = 'I SURVIVED <tspan class="user-stat">MONDAYS</tspan> WITH MUSIC'
         case "Tuesday": most_played_day_message = 'I CHILLED WITH MUSIC ON <tspan class="user-stat">TUESDAY</tspan>'
@@ -254,6 +259,10 @@ def _cover_art_yim_albums(user_name, stats):
     cac = CoverArtGenerator(current_app.config["MB_DATABASE_URI"], 3, 250)
     image_urls = []
     selected_urls = set()
+
+    if stats.get("top_releases") is None:
+        return None
+
     for item in stats["top_releases"]:
         if "caa_id" in item and "caa_release_mbid" in item:
             url = cac.resolve_cover_art(item["caa_id"], item["caa_release_mbid"], 250)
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/views/art_api.py
```diff
@@ -336,26 +349,18 @@ def cover_art_yim_2022(user_name):
...
+    match image:
+        case "stats": svg = _cover_art_yim_stats(user_name, stats)
+        case "albums": svg = _cover_art_yim_albums(user_name, stats)
+        case "tracks": svg = _cover_art_yim_tracks(user_name, stats)
+        case "artists": svg = _cover_art_yim_artists(user_name, stats)
+        case "discovery-playlist": svg = _cover_art_yim_playlist(user_name, stats, "playlist-top-discoveries-for-year")
+        case "missed-playlist": svg = _cover_art_yim_playlist(user_name, stats, "playlist-top-missed-recordings-for-year")
+        case other: raise APIBadRequest(f"Invalid image type {other}. Image type should be one of (stats, artists, albums, tracks, discovery-playlist, missed-playlist)")
+
+    status_code = 200 if svg is not None else 204
```
alastair, line 365: is this necessary? since it'll be caught by the `case other`, right? This is something I've not used before, so I don't know if this is the best way to do things...

listenbrainz/webserver/views/art_api.py
```diff
@@ -254,6 +254,10 @@ def _cover_art_yim_albums(user_name, stats):
     cac = CoverArtGenerator(current_app.config["MB_DATABASE_URI"], 3, 250)
     image_urls = []
     selected_urls = set()
+
+    if stats.get("top_releases") is None:
```
alastair, line 258: I'm not too worried either way but this could also have been `if not stats.get("x")` (unless there are falsy items?), or even `if "x" not in stats` (which I'm a bit more of a fan of, but not enough to force you to change it)

listenbrainz/webserver/views/art_api.py
```diff
@@ -225,7 +225,7 @@ def cover_art_custom_stats(custom_name, user_name, time_range, image_size):
 
 def _cover_art_yim_stats(user_name, stats):
     """ Create the SVG using YIM statistics for the given year. """
-    match stats["day_of_week"]:
+    match stats.get("day_of_week"):
```
alastair, line 228: this means that if there are no items for `day_of_week` then it'll return `"I CRANKED TUNES ON <tspan class="user-stat">None</tspan>"`, is that what we want?

alastair (changes requested):

listenbrainz/webserver/views/art_api.py
amCap1712: This is for the case where the image type is correct but we didn't have enough data in the stats to generate a svg. The case other is for when the image type in the request was invalid.

listenbrainz/webserver/views/art_api.py
amCap1712: Hmm, good point. Let me think about it.

listenbrainz/webserver/views/art_api.py
amCap1712: Makes sense but `"x" not in stats` doesn't handle the case of null items in dict.

listenbrainz/webserver/views/art_api.py
amCap1712: Fixed with return a 204 No Content in that case too.

PR #2411, opened 2023-03-16
Redo schema of listen table

amCap1712:
1. Move the track_name field to inside track_metadata and recording_msid to a top level column because we want to deduplicate on recording_msid instead of just track_name for consistency.

2. Remove user_name column from the listen table.

3. In the data field, currently data is stored as {"track_metadata": ..data}. Change this so that ...data is directly stored at top level inside jsonb field.

### DIFF
--- a/admin/timescale/create_tables.sql
+++ b/admin/timescale/create_tables.sql
@@ -10,7 +10,7 @@ CREATE TABLE listen (
 );
 
 CREATE TABLE listen_new (
-    listened_at     BIGINT                   NOT NULL,
+    listened_at     TIMESTAMP WITH TIME ZONE NOT NULL,
     created         TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
     user_id         INTEGER                  NOT NULL,
     recording_msid  UUID                     NOT NULL,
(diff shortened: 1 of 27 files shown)


admin/timescale/create_tables.sql
```diff
@@ -9,6 +9,14 @@ CREATE TABLE listen (
         data            JSONB                    NOT NULL
 );
 
+CREATE TABLE listen_new (
```
mayhem, line 12: I dislike the name of this table, it isn't very future proof, IMHO.
listen_v2? (or is it v3 already?)

listenbrainz/listenstore/timescale_listens_migrate.py
```diff
@@ -0,0 +1,89 @@
+"""
+Utility tool for migrating listens to new schema table. It can be run multiple times.
```
mayhem, line 2: Great! :+1:

mayhem (approved):
Please change the table name, otherwise good. Do you plan to run this in one go and restart if it gets interrupted or do you plan to run it in stages?

admin/timescale/create_tables.sql
amCap1712: the intent is to do an atomic swap once the migration is complete to rename this to listen and the existing listen table to listen_old.

PR #2412, opened 2023-03-16
Reset completed page number on new import

TheLastZombie:
# Problem

When a user runs the Last.fm importer, resets the import timestamp in a new tab (i. e. without leaving the importer itself or reloading its tab) and then runs the importer again, the number of completed pages would not be reset, instead continuing to increase from the value of the last import.

image

# Solution

This PR simply resets the `numCompleted` value to `0` before starting an import.

# Action

N/A


MonkeyDo (approved):
Simple enough improvement, but a good one, thank you !
:rocket:

PR #2414, opened 2023-03-18
Remove filter_artist_credit option from similarity datasets

amCap1712:
We always want to filter so need to keep this as configurable.

### DIFF
--- a/listenbrainz_spark/year_in_music/listening_time.py
+++ b/listenbrainz_spark/year_in_music/listening_time.py
@@ -1,5 +1,6 @@
 import listenbrainz_spark
 from listenbrainz_spark import config
+from listenbrainz_spark.path import RECORDING_LENGTH_DATAFRAME
 from listenbrainz_spark.stats import run_query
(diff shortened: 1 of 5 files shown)


listenbrainz_spark/year_in_music/listening_time.py
```diff
@@ -20,14 +21,14 @@ def get_listening_time(year):
 
 
 def _get_total_listening_time():
-    # get recording length from mb_metadata_cache, if listen is unmapped default to 3 minutes.
+    # get recording length from recording table, if listen is unmapped default to 3 minutes.
```
mayhem, line 24: Is this because we have direct access to the main servers now?

mayhem (approved):

listenbrainz_spark/year_in_music/listening_time.py
amCap1712: Yes.

amCap1712: And we already use this table elsewhere.

PR #2415, opened 2023-03-18
Daily Jams Fixes (yet again)

amCap1712:
Increase rate limit for troi auth token, update troi to bring in ratelimit fixes and add a not None case LB side for error handling.


mayhem (approved):
Daily jams, our problem child. :/ Let's hope this gets things more stable.

PR #2416, opened 2023-03-18
Codebase Refactor

HighnessAtharva:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=BTQwRFm)

# Problem

Majority of the codebase has legacy ways to format strings using interpolation and usage of the `+` operator which is old-fashioned and slow. Also, some of the code is not all that readable and the flow-control or readability can be improved.

# Solution

I have refactored a lot of files to fix this issue. 
- Replaced string interpolation with modern f-strings
- Flow control and conditional logic improved.
- List Comprehension, Dictionary Comprehension utilized instead of length and complex-looking for loops and generators.

# Action

There is still plenty of room for readability improvement. Once this PR gets merged, I will get started working on the rest. Don't want to burden the reviewers.

### DIFF
--- a/admin/cron_lock.py
+++ b/admin/cron_lock.py
@@ -40,7 +40,7 @@ def lock_cron(slug, msg):
        identified by slug."""
     sanity_check()
 
-    cron_lock_file = os.path.join(CRON_LOGS_DIR, "cron-%s.lock" % slug)
+    cron_lock_file = os.path.join(CRON_LOGS_DIR, f"cron-{slug}.lock")
     if os.path.exists(cron_lock_file):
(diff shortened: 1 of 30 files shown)


staticssleever668:
By the way, there are more percent formatting instances and .format calls, if it's ever decided to migrate from them to f-strings, they can easily be found with pyupgrade rules in [ruff](https://github.com/charliermarsh/ruff): `ruff check --select UP031,UP032 .`.

mayhem (commented):
I really dislike this PR. In some cases the "old fashioned" strings are actually simpler than the "newfangled way". Overall,  I think this PR makes the code worse than before.

PR #2418, opened 2023-03-21
Hide "Add to MB" button from missing MB links page

MonkeyDo:
Aerozol reported that he is seeing some bad edits in MusicBrainz, a bunch of standalone recordings being added from ListenBrainz one by one instead of as a cohesive release.
For now, we are going to comment out the "add to MB" button until we have a good way to group listens together to create releases, or a better way to guide users on how to add data.
Currently this is a dangerous tool that pushes users to add bad data.


MonkeyDo:
Pinging @Aerozol since he was the one to report this concern

MonkeyDo:
Also pinging @mayhem to let you know we're removing this feature for now, as it is opening the door to a lot of bad data in MB.
There are already some incomplete additions, considering we don't currently have a good way to seed an entire release.

mayhem:
Yeah, that is fine for now.

PR #2419, opened 2023-03-21
LB-1148: Distinguish "no cover art" and "not mapped" listens

MonkeyDo:
Redesign the ListenCard thumbnail to differentiate:
* when we have a mapping but the release has no cover art (when clicked, go to MB release cover art page)
* when we don't have a mapping, show link icon (when clicked, open the link-to-MB modal)

image


amCap1712 (approved):
LGTM, Thanks

Aerozol:
Beautiful, gracias

MonkeyDo:
After seeing @Aerozol 's comment on the ticket, I've added a component that tries to load the cover art but shows an icon as a fallback in case that fails (i.e. if CAA is down).
For an example, see the last listen in this gauntlet showing all the possible cases:
image


And in 'compact' mode: 
image



On long hover, you get useful messages:
image

image
 
image

PR #2423, opened 2023-03-22
LB-1251: Store spotify user id in external_service_oauth

meycroft:
# Problem
There is a need for spotify user ids to export playlists from troi to spotify. Currently, the system queries the spotify user id each time before doing so. This solution stores this data in the external service oauth tables itself when the user connects their account.

Related JIRA Ticket: LB-1251

This is the implementation of a the first part of the original ticket.

# Solution

This creates a new column in the `external_services_oauth` table, adds a part to call the Spotify API and get user id in the `add_new_user` function (`listenbrainz/domain/spotify.py`). This ensures that the user ids are added whenever a new user connects their account (as mentioned in the JIRA ticket). Also added an a new entry corresponding to the user id in `save_token` (`listenbrainz/db/external_service_oauth.py`) function to add the external service user id (Spotify in this case) to the row being added in the `external_service_oauth` table.

### DIFF
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -124,6 +124,7 @@ CREATE TABLE spotify_auth (
 CREATE TABLE external_service_oauth (
     id                      SERIAL,  -- PK
     user_id                 INTEGER NOT NULL,  -- FK to "user".id
+    external_user_id        TEXT,
     service                 external_service_oauth_type NOT NULL,
     access_token            TEXT NOT NULL,
     refresh_token           TEXT,
--- a/listenbrainz/db/external_service_oauth.py
+++ b/listenbrainz/db/external_service_oauth.py
@@ -1,4 +1,4 @@
-from typing import List, Union
+from typing import List, Optional, Union
 
 from data.model.external_service import ExternalServiceType
 from listenbrainz import db, utils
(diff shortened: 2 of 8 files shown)


listenbrainz/db/external_service_oauth.py
```diff
@@ -6,7 +6,7 @@
 
 
 def save_token(user_id: int, service: ExternalServiceType, access_token: str, refresh_token: str,
-               token_expires_ts: int, record_listens: bool, scopes: List[str]):
+               token_expires_ts: int, record_listens: bool, scopes: List[str], external_user_id: str = None):
```
staticssleever668, line 9: Passing-by review :)
the type hint is wrong, it should be an option since the default argument is `None`
```suggestion
               token_expires_ts: int, record_listens: bool, scopes: List[str], external_user_id: Optional[str] = None):
```
The type is not in the scope, so you have to update imports at the top as well:
`from typing import List, Optional, Union`

admin/sql/create_tables.sql
```diff
@@ -124,6 +124,7 @@ CREATE TABLE spotify_auth (
 CREATE TABLE external_service_oauth (
     id                      SERIAL,  -- PK
     user_id                 INTEGER NOT NULL,  -- FK to "user".id
+    external_user_id        TEXT,
```
amCap1712, line 127: We also need a new SQL update script under admin/sql/updates to update the tables in production. See https://github.com/metabrainz/listenbrainz-server/blob/master/admin/sql/updates/2022-10-12-add-troi-user-setting.sql for an example.

amCap1712 (commented):
Hi! Thanks for the PR. The changes LGTM but also need to update tests. I'll look at the failures and try to help in those.

meycroft:
> Hi! Thanks for the PR. The changes LGTM but also need to update tests. I'll look at the failures and try to help in those.

Working on tests. Will push them soon.

PR #2424, opened 2023-03-23
bugfix: scrobbling on now_playing (type mismatch)

vjeranc:
# Problem

People are experiencing multiple scrobbles using `cmus-status-scrobbler` https://github.com/vjeranc/cmus-status-scrobbler/issues/2

`cmus-status-scrobbler` sends a `track.updateNowPlaying` event whenever the player resumes from a pause.

Listenbrainz server interprets that as a scrobble (`track.scrobble` is not sent).

# Solution

This looks like a type mismatch. `insert_payload` expects an integer enumeration for the `listen_type` but the code that sends the `listen_type` passes `"playing_now"` and `"listens"`.

# Action

This should be tested. I am unsure if there's a need for the `LISTEN_TYPE_IMPORT` type to be used.

### DIFF
--- a/listenbrainz/webserver/views/api_compat.py
+++ b/listenbrainz/webserver/views/api_compat.py
@@ -12,7 +12,8 @@ from listenbrainz.webserver.errors import InvalidAPIUsage, CompatError, ListenVa
 import xmltodict
 
 from listenbrainz.webserver.models import SubmitListenUserMetadata
-from listenbrainz.webserver.views.api_tools import insert_payload, validate_listen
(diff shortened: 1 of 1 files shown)


amCap1712:
Hi! @vjeranc Thanks for the PR. A quick question: would you be open to accepting a PR to use native ListenBrainz API for submitting listens to LB than using the api_compat.

vjeranc:
@amCap1712 Of course I'm open, I believe it wouldn't be a hard thing to add.

On the other hand, it would add additional code and configuration options.

I did notice when I started using `api_compact`, that there were runtime bugs at Listenbrainz that were eventually fixed.

I guess this is one of the last that is affecting the scrobbler.

amCap1712 (approved):

PR #2425, opened 2023-03-25
LB-1005: Replace newAlert prop with toast notification library in BrainzPlayer

Shelly011s:
# Problem
We are currently passing a 'newAlert' method in BrainzPlayer props from its parent.
The method belongs to a special alert component created at the root, used to display toast notifications on error, warning, success.
We are going to want to manage these notifications directly in BrainzPlayer instead, and remove the newAlert prop.
It is proposed to use the [react-toastify library](https://fkhadra.github.io/react-toastify/introduction) which seems a bit nicer than the one we are currently using and has more features.

# Solution

This is a draft PR which needs more changes.
These are mockups of the implemented react-toastify library.
error
warn

# Action

Suggestions on how to deal with overlapping of toasts from react-bs-notifier and react-toastify.


amCap1712:
Hi! Just taking a quick look at the code changes, it seems you have forgotten to commit some files. Currently, I can only see changes adding the dependency but not replacing the usage of newAlert component with components from the added dependency.

Shelly011s:
> Hi! Just taking a quick look at the code changes, it seems you have forgotten to commit some files. Currently, I can only see changes adding the dependency but not replacing the usage of newAlert component with components from the added dependency.

frontend/js/src/brainzplayer/BrainzPlayer.tsx
```diff
@@ -118,6 +118,14 @@ function isListenFromDatasource(
...
 }
 
+export function ToastMsg({ title, message }: ToastMsgProps) {
+  return (
+    <div>
+      <b>{title}</b>
+      <p>{message}</p>
+    </div>
+  );
+}
```
MonkeyDo, line 128: This isn't the place to export this component.
We are eventually going to use it across the codebase, so we want to import it from a more "neutral" place.

Maybe its own component/file in the `js/src/notifications` folder?

MonkeyDo (changes requested):
Thanks, a helpful PR !

I would like to request as a follow-up, if you have the time, to replace the current `react-bs-notifier` library with your new library across the codebase.
Now that you have the experience of replacing it in BrainzPlayer, it would be great to replace it all over and get rid of the old library.
It will modify a lot of files, so better to be a separate PR.

Please let me know if you have time for it; I can create a ticket to go along with it

frontend/js/src/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -12,6 +12,8 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { IconDefinition } from "@fortawesome/fontawesome-common-types"; // eslint-disable-line import/no-unresolved
 import { IconProp } from "@fortawesome/fontawesome-svg-core";
 import { isNaN as _isNaN } from "lodash";
+import { ToastContainer, toast } from "react-toastify";
```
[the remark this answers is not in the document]
Shelly011s: That's a mistake on my end. I should have removed it since its not used.

frontend/js/src/brainzplayer/BrainzPlayer.tsx
Shelly011s: Sounds good to me :+1:

MonkeyDo:
> I would be glad to work on it even though it would be a very time consuming task. I would really appreciate if you could help me along the way to achieve this task. Correct me if I am wrong but it would require changes in alertNotificationsHOC and then I can work on removing newAlerts as props in all the components that are using it?

Yes, that's exactly it !

Shelly011s:
@MonkeyDo I apologise for that, I thought those changes were to be done with a separate PR in context of replacing `react-bs-notifier` library all over.

MonkeyDo (approved):
Thanks for your work!

Can't wait to be rid of this newAlert prop everywhere :)
BrainzPlayer is a great start!

PR #2427, opened 2023-03-25
LB-1240: Add an API to list external services user has connected

vardansaini:
# Problem

The Android App needed a way to know whether the user has connected their CritiqueBrainz account with ListenBrainz.  Therefore, I added a new API endpoint to return the list of services a user has connected to their account.
 https://tickets.metabrainz.org/browse/LB-1240

# Solution

  I added a new API endpoint to return the list of services a user has connected to their account.

# Action

    I am slightly confused about tests so I might need help with that.

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -18,7 +18,7 @@ from listenbrainz.webserver import timescale_connection
 from listenbrainz.webserver.decorators import api_listenstore_needed
 from listenbrainz.webserver.decorators import crossdomain
 from listenbrainz.webserver.errors import APIBadRequest, APIInternalServerError, APINotFound, APIServiceUnavailable, \
-    APIUnauthorized, ListenValidationError
+    APIUnauthorized, ListenValidationError, APIForbidden
 from listenbrainz.webserver.models import SubmitListenUserMetadata
 from listenbrainz.webserver.utils import REJECT_LISTENS_WITHOUT_EMAIL_ERROR
 from listenbrainz.webserver.views.api_tools import insert_payload, log_raise_400, validate_listen, \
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/api.py
```diff
@@ -635,6 +636,33 @@ def get_playlists_collaborated_on_for_user(playlist_user_name):
     return jsonify(serialize_playlists(playlists, playlist_count, count, offset))
 
 
+@api_bp.route("/user/<user_name>/services", methods=['GET', 'OPTIONS'])
+@crossdomain
+@ratelimit()
+def get_service_details(user_name):
```
amCap1712, line 642: I think this API endpoint should be authenticated and only the current user should be able to access its details.

listenbrainz/webserver/views/api.py
```diff
@@ -635,6 +636,34 @@ def get_playlists_collaborated_on_for_user(playlist_user_name):
...
+
+        {
+            "user_name": "hwnrwx",
+            "services": ["spotify"]
+        }
+
+    :param user_name: the MusicBrainz ID of the user whose similar users are being requested.
+    :resheader Content-Type: *application/json*
+    :statuscode 200: Yay, you have data!
+    :statuscode 403: Forbidden, you do not have permissions to view this user's information.
```
vardansaini, line 656: https://github.com/metabrainz/listenbrainz-server/blob/b28e14f2843f221cd5cbe555c9eabdc50c1a6e36/listenbrainz/webserver/views/user_timeline_event_api.py#L253

saw this from the link above
But we should give 403 on wrong token but you guys are giving 401.

amCap1712: Ah. Good catch. I'll open a ticket for fixing this.

amCap1712 (approved):
Thanks for the PR and pointing out the HTTP Status code issue. LGTM.

PR #2429, opened 2023-03-28
LB-1190: Broken Youtube Link and long error notifications

Shelly011s:
# Problem
The youtube link in brainzplayer is broken as it leads to a "page not found" error. Also when working on local server, there were annoying lengthy error notifications for listens page and fresh releases page due to internal server error.

# Solution

I have fixed the youtube links which now leads to the destined youtube video and I have also fixed the lengthy errors.
Before:
ugly error
After:
good error

# Action


frontend/js/src/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -149,6 +149,15 @@ function BrainzPlayerUI(props: React.PropsWithChildren<BrainzPlayerUIProps>) {
...
   }
+
+  function getURLFromID(trackID: string | undefined): string | undefined {
+    if (currentDataSourceName === "youtube") {
+      return `https://www.youtube.com/watch?v=${trackID}`;
+    }
+
+    return undefined;
+  }
+
```
MonkeyDo, line 160: This isn't what we want to do. Currently, this completely overwrites any valid track url if the listening service is not youtube.
This is also not the right place to fix the issue with the trackUrl. You need to follow the prop up along the river to find why we are sending a videoID instead of a URL.

If you look in the YoutubePlayer component, you'll see we do the right thing here: https://github.com/metabrainz/listenbrainz-server/blob/frontend-redesign/frontend/js/src/brainzplayer/YoutubePlayer.tsx#L146
But we do the wrong thing here: https://github.com/metabrainz/listenbrainz-server/blob/frontend-redesign/frontend/js/src/brainzplayer/YoutubePlayer.tsx#L186

Ideally, this piece of code would be refactored in a static function (getURLFromVideoId)  so that it can be used for both, like we do with getURLFromListen
https://github.com/metabrainz/listenbrainz-server/blob/frontend-redesign/frontend/js/src/brainzplayer/YoutubePlayer.tsx#L100

frontend/js/src/user/Listens.tsx
```diff
@@ -254,7 +254,7 @@ export default class Listens extends React.Component<
         newAlert(
           "danger",
           "We could not load data for the now playing listen",
-          typeof error === "object" ? error.message : error
+          error.status ?? error.toString()
```
MonkeyDo, line 257: I don't think this is a good idea the way it is implemented. There could be an existing `error.message` on some errors, which is a lot more informative than just the status number (which doesn't mean anything to end users)

MonkeyDo (changes requested):

MonkeyDo:
The PR title could do with some love too, it's very vague at the moment :)

frontend/js/src/user/Listens.tsx
Shelly011s: Most of the times error.message and error.toString() produce similar results so maybe replacing it with error.message ?? error.status should do

Shelly011s: Is it possible to give me some examples of those errors, so I can come up with solutions keeping them in mind and also test accordingly.

MonkeyDo: The call to the API method `getFeedbackForUserForRecordings` in `loadFeedbackForNowPlaying` that you are catching passes through this utility to check requests: https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/utils/APIService.ts#L489-L512

Shelly011s:
I have updated the PR. However I need your help to resolve the failed tests. :)

frontend/js/src/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -219,7 +220,7 @@ function BrainzPlayerUI(props: React.PropsWithChildren<BrainzPlayerUIProps>) {
       <div className="actions">
         {isPlayingATrack && currentDataSourceName && (
           <a
-            href={trackUrl ?? "#"}
+            href={trackUrl || "#"}
```
MonkeyDo, line 223: Out of curiosity, what was the need for this change?
Is it for when `trackUrl` is an empty string?

frontend/js/src/brainzplayer/YoutubePlayer.tsx
```diff
@@ -97,6 +97,13 @@ export default class YoutubePlayer
     return images;
   }
 
+  static getURLFromVideoID(trackID: string): string {
```
MonkeyDo, line 100: This would make more sense with the function name:
```suggestion
  static getURLFromVideoID(videoID: string): string {
```

MonkeyDo (changes requested):
Thanks, that's looking a lot better !
Since last review I had forgotten about this PR and opened my own (#2449), but will close it now XD
However, do have a look at that PR as you will need to apply the same change in the test file: https://github.com/metabrainz/listenbrainz-server/pull/2449/files#diff-974e5a52659639895ebc433a796948b0067cba584f4b11ae18cae0a823d0b5b5L119
I also wonder if this is a useful change to add: https://github.com/metabrainz/listenbrainz-server/pull/2449/files#diff-1eab2b5d4f3682037df5a1a53536a8e03ad5fefee3d4828c7f4d75c9ccc0d207L180


I do have a couple of extra small comments below.

For what it's worth, for next time, it's almost always better to separate unrelated changes into separate PRs. In this case the Youtube URL issue is a separate problem from the error messages issue, would have been better as two PRs.

frontend/js/src/brainzplayer/BrainzPlayerUI.tsx
Shelly011s: Yup you got that right, it is for handling the empty string. I will also keep in mind to create separate PR for manage issues.

MonkeyDo (changes requested):
Code looks good now, thanks !
Tests are passing so that's you done for the coding parts of this PR !

However it looks like you commited the .vscode/settings.json file by mistake.
At the moment I don't think we have decided to commit this type of files, nor what sort of default settings we would want, so the file is unwanted at the moment.

MonkeyDo (approved):
Thanks a lot!

PR #2435, opened 2023-03-30
Tooltip

Shelly011s:
# Problem

This is an implementation of thumb scroll as per the figma designs provided.

# Solution
I have taken the liberty to implement the thumb scroll for horizontal scrollbar as well.
vertical
horiz

# Action
I would appreciate feedback on whatever I can improve or work upon.


mayhem:
Confirmed looks/works as in the screenshot. I'll close this for now, but we have it for future reference. Thanks!

PR #2436, opened 2023-03-31
LB-1011: Don't load album covers if Save-Data HTTP header is on

faraz16iqbal:
# Problem

[LB-1011](https://tickets.metabrainz.org/browse/LB-1011):
If a user's browser has set the header "Save-Data: on", we should detect that and set a javascript property in global_props accordingly.

Then in the React pages, we can avoid loading and showing cover artwork on ListenCards.

# Solution

Check for the saveData header on the client side and add it to the global props object. This way we can detect if the user has save data mode on and we can not load album covers in the ListenCard component.

# Action

Updated the GlobalContextType and added the user preferences object to be able to store the user's save-data preference globally. The network-information-types are necessary to use the navigator object in Typescript.

### DIFF
--- a/frontend/js/src/listens/ListenCard.tsx
+++ b/frontend/js/src/listens/ListenCard.tsx
@@ -91,7 +91,6 @@ export type ListenCardProps = {
 export type ListenCardState = {
   isCurrentlyPlaying: boolean;
   thumbnailSrc?: string; // Full URL to the CoverArtArchive thumbnail
-  saveDataMode?: boolean;
 };
 
 export default class ListenCard extends React.Component<
--- a/frontend/js/src/listens/ListenCard.tsx
+++ b/frontend/js/src/listens/ListenCard.tsx
@@ -91,6 +91,7 @@ export type ListenCardProps = {
 export type ListenCardState = {
   isCurrentlyPlaying: boolean;
   thumbnailSrc?: string; // Full URL to the CoverArtArchive thumbnail
+  saveDataMode?: boolean;
 };
 
 export default class ListenCard extends React.Component<
@@ -106,20 +107,28 @@ export default class ListenCard extends React.Component<
 
     this.state = {
       isCurrentlyPlaying: false,
+      saveDataMode: false,
     };
   }
 
   async componentDidMount() {
     window.addEventListener("message", this.receiveBrainzPlayerMessage);
-    await this.getCoverArt();
+    const { userPreferences } = this.context;
+    if (userPreferences?.saveDate) {
+      this.setState({ saveDataMode: true });
+    } else {
+      await this.getCoverArt();
+    }
   }
 
   async componentDidUpdate(oldProps: ListenCardProps) {
     const { listen, customThumbnail } = this.props;
+    const { saveDataMode } = this.state;
     if (
       !customThumbnail &&
       Boolean(listen) &&
-      !isEqual(listen, oldProps.listen)
+      !isEqual(listen, oldProps.listen) &&
+      saveDataMode === true
     ) {
       await this.getCoverArt();
     }
--- a/frontend/js/src/listens/ListenCard.tsx
+++ b/frontend/js/src/listens/ListenCard.tsx
@@ -114,9 +114,7 @@ export default class ListenCard extends React.Component<
   async componentDidMount() {
     window.addEventListener("message", this.receiveBrainzPlayerMessage);
     const { userPreferences } = this.context;
-    if (userPreferences?.saveDate) {
-      this.setState({ saveDataMode: true });
-    } else {
+    if (userPreferences?.saveData !== true) {
       await this.getCoverArt();
     }
   }
--- a/frontend/js/src/listens/ListenCard.tsx
+++ b/frontend/js/src/listens/ListenCard.tsx
@@ -107,7 +107,6 @@ export default class ListenCard extends React.Component<
 
     this.state = {
       isCurrentlyPlaying: false,
-      saveDataMode: false,
     };
   }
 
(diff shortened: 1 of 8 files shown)


faraz16iqbal:
@MonkeyDo has not added the logic to not load covers. Will update this PR once you go through and approve these changes!

MonkeyDo:
> @MonkeyDo has not added the logic to not load covers. Will update this PR once you go through and approve these changes!

Hi!
Please do complete the work required for the PR before asking for review, that way I can review once instead of twice :)

faraz16iqbal:
@MonkeyDo Hi, yes alright

faraz16iqbal:
@MonkeyDo I have made the necessary changes. Kindly review the PR whenever you can

faraz16iqbal:
@MonkeyDo

frontend/js/src/utils/utils.tsx
```diff
@@ -465,8 +466,15 @@ const getPageProps = (): {
       critiquebrainz,
       sentry_traces_sample_rate,
       sentry_dsn,
+      user_preferences,
     } = globalReactProps;
 
+    if ("connection" in navigator) {
+      if (navigator.connection?.saveData === true) {
+        user_preferences!.saveDate = true;
```
MonkeyDo, line 474: I don't like the use of `!` here, it's usually a bad code smell to force typescript types like that.
Can it be avoided? In this case, what was the reason for using this that user_preferences is possibly undefined (looking at GlobalAppProps type)?

It should be possible to instead ensure that it is an existing object.
Maybe something like:
```
user_preferences = Object.assign({}, user_preferences, {saveDate: true})
```
This ensures user_preferences is an existing empty array at least, and concatenate it with the user_preferences (if any) present in the page props, and also with an object that sets `saveData`.

This will require you to change the `const {` on line 461 to a `let {`, otherwise typescript will complain that you `cannot assign to user_preferences because it is a constant`

frontend/js/src/utils/types.d.ts
```diff
@@ -682,3 +682,7 @@ type UserFreshReleasesResponse = {
 declare type SearchUser = {
   user_name: string;
 };
+
+declare type UserPreferences = {
+  saveDate?: boolean
```
MonkeyDo, line 687: This is misnamed in various places in the PR
```suggestion
  saveData?: boolean
```

frontend/js/src/listens/ListenCard.tsx
```diff
@@ -106,20 +107,28 @@ export default class ListenCard extends React.Component<
...
+      saveDataMode: false,
     };
   }
 
   async componentDidMount() {
     window.addEventListener("message", this.receiveBrainzPlayerMessage);
-    await this.getCoverArt();
+    const { userPreferences } = this.context;
+    if (userPreferences?.saveDate) {
+      this.setState({ saveDataMode: true });
```
MonkeyDo, line 118: There should be no need to set this value in the state.
In `componentDidUpdate` below you can extract the property using `const { userPreferences } = this.context;` just like you do here.

MonkeyDo (changes requested):
Looking a lot better, thanks for setting down this stub user_preferences, they will be very handy soon !

Some comments about the code:

frontend/js/src/utils/utils.tsx
faraz16iqbal: Extracted the user property as let and the other's are left as const since eslint was throwing an error otherwise.
  `userPreferences` object is now reassigned as `user_preferences = { ...user_preferences, saveData: true };`

frontend/js/src/utils/types.d.ts
faraz16iqbal: Fixed in all places

frontend/js/src/listens/ListenCard.tsx
faraz16iqbal: Fixed

frontend/js/src/listens/ListenCard.tsx
```diff
@@ -91,6 +91,7 @@ export type ListenCardProps = {
 export type ListenCardState = {
   isCurrentlyPlaying: boolean;
   thumbnailSrc?: string; // Full URL to the CoverArtArchive thumbnail
+  saveDataMode?: boolean;
```
MonkeyDo, line 94: This should be removed

frontend/js/src/utils/utils.tsx
```diff
@@ -467,6 +468,16 @@ const getPageProps = (): {
       sentry_dsn,
     } = globalReactProps;
 
+    let { user_preferences } = globalReactProps;
+
+    user_preferences = { ...user_preferences, saveData: true };
```
MonkeyDo, line 473: I don't think this should be true by default. Did you mean it to be this instead?
```suggestion
    user_preferences = { ...user_preferences, saveData: false };
```

MonkeyDo (changes requested):
Almost there !


I don't know exactly what is happening with the test suite, it looks like typescript is unhappy with the spotify and youtube types, which you didn't touch directly.
Maybe something to do with the addition of the network-information-types in tsconfig.json?
https://github.com/metabrainz/listenbrainz-server/actions/runs/4699023761/jobs/8332861336?pr=2436

frontend/js/src/utils/utils.tsx
faraz16iqbal: Yeah, this should've been false by default! Thanks for that

frontend/js/src/listens/ListenCard.tsx
faraz16iqbal: Removed

faraz16iqbal:
@MonkeyDo Yeah this failure seems to be caused by the additional of that extra information type. On removing it and running the type check, I get this error `Property 'saveData' does not exist on type '{}'.` which persisted earlier

faraz16iqbal:
@MonkeyDo I have just added a ts-ignore for that one particular instance where `navigator.connection` is being used. Let me know if it works fine or should be changed

MonkeyDo (approved):
Yes, I think the ts-ignore is acceptable in this case.
I did want to say that you did the right thing at first in installing the missing types :)

Thanks for the PR !

faraz16iqbal:
@MonkeyDo thank you! Looking forward to contributing more to metabrainz!

PR #2437, opened 2023-03-31
Add a flag to generate daily jams for all users

amCap1712:
create_all flag defaults to False and creates daily jams for users according to the timezones only. If it is passed as true, then daily jams for all users irrespective of the timezone are created.


mayhem (approved):
Thanks!

PR #2439, opened 2023-03-31
Updated Tooltip

jivteshsingh:
The tooltip had some issues like the single date was single digit (1 Apr) instead of (01 Apr) as in the mockup designs which was affecting the overall design of the updated design. So this pr is a update for the issue.


jivteshsingh:
For 2 digit dates like 27 Mar'
Screenshot (7)

jivteshsingh:
For single digit dates like 01 Apr
Screenshot (8)

PR #2440, opened 2023-04-01
Hide user actions when user is not logged in

arsh331:
# Problem
1. When a user is not logged in, value of currentUser is `{}` and `Boolean({})` is true, leading to unintended results.
2. Love, hate, pin, recommend and review track options are shown even when the user is not logged in.

# Solution
1. Use `isEmpty()` to properly check if currentUser is empty.
2. To improve the UX, these options should be hidden from the user until the user has logged in.

# Action
N/A


amCap1712 (changes requested):
Hi!

I tried out the PR locally and it seems to work correctly. Good catch on the `Boolean({})` is `true` thing. There are 2 things that need to worked before this PR is ready. 

1. The commits from documentation stuff need to be removed from this branch. If you are unsure on how to do this, ping me in IRC and I'll assist in it.
2. The test snapshots need to be updated. You can run `./test.sh fe -u` locally for the same, then commit and push the changes.

MonkeyDo (approved):
Thank you, this all makes sense !
:rocket:

PR #2446, opened 2023-04-11
LB-834: New listens are added to top of wrong Recent Listens page

KasukabeDefenceForce:
# Problem
New listens are added to the My Listens page dynamically via websockets. That's great in most cases, but if  navigated to page 2, new listens that the user is listening to now are going to be added to the top of that page, which makes them appear out of order until navigated again.

# Solution

Made a new list "webSocketListens"  in the "receiveNewListen" function to store all the listens which have been played since the page is open. This list is displayed to the user in a similar manner as the "recent listens" list.

Screenshot from 2023-04-11 17-40-44

# Action
Some changes need to be made to the margins of the listen count card.


MonkeyDo:
I have been mulling this over, I think as-is the user experience won't be great: if my page is open for a long time, I could end up with a very long list of listens that would take over the page entirely.
I think the list of new listens should be limited to the most recent 5 or so (maybe?) listens.
Visually, I think that should be represented by a gradual fade to transparency of the new listens container, and should also provide a link to navigate to those new listens (using a timestamp saved when we first load the page).

I threw this mockup together to show you what I was thinking, happy to discuss it with you.

image

KasukabeDefenceForce:
@MonkeyDo  Okay, I am working on it. Also what to do when the user clicks on "See all your new listens"?

MonkeyDo:
> @MonkeyDo Okay, I am working on it. Also what to do when the user clicks on "See all your new listens"?

This should be the same mechanism as the "newest" button in the pager.
(the one on the left here 
image
)

https://github.com/metabrainz/listenbrainz-server/blob/b78eede9e68621eed1eded66faeb1c8491cf802a/frontend/js/src/user/Listens.tsx#L940-L954

KasukabeDefenceForce:
https://github.com/metabrainz/listenbrainz-server/pull/2446/files#diff-757635a0c8763dba0b425cc53ad61b9a0d6fa1bf7781d9e41c1c37a898e45c6aR909
@MonkeyDo  Can you please tell me how to fix this error?

MonkeyDo:
> https://github.com/metabrainz/listenbrainz-server/pull/2446/files#diff-757635a0c8763dba0b425cc53ad61b9a0d6fa1bf7781d9e41c1c37a898e45c6aR909 @MonkeyDo Can you please tell me how to fix this error?

I went in and pushed the change :)
Basically, we recently refactored our notifications and the `newAlert` prop that we used to pass around doesn't exist anymore, so it needed to be removed from the call to open the modal.

MonkeyDo:
I hope you don't mind, I took over to polish things up a little. I deployed it to test and it's looking great !
Thanks a lot, I've been wanting to see this for a long while :)

PR #2448, opened 2023-04-12
LB-1258: Listening now not showing cover art even if correct release is found.

KasukabeDefenceForce:
# Problem
Listening now is not displaying the cover art even if it has identified the release and the release has cover art.

# Solution
Made the necessary changes in the API query. The type of the metadata variable was set to " MetaDataLookup | null" and then the cca_id and cca_release_mbid were destructured. Also made changes to interface so that it allows null as a value for both the variables. 

# Action


phw:
Without knowing much about the sources overall the code changes in isolation definitely look good. I used the same general approach of setting the `track_metadata.mbid_mapping.caa_id` (and corresponding release field) from the deeper nested values in the implementation for the discourse plugin.

MonkeyDo (commented):
A coupe of questions/ details below.
Overall this looks good, thank you for the improvement !

frontend/js/src/user/Listens.tsx
```diff
@@ -240,12 +240,16 @@ export default class Listens extends React.Component<
...
     const { APIService } = this.context;
     try {
-      const metadata = await APIService.lookupRecordingMetadata(
+      const metadata: MetadataLookup | null = await APIService.lookupRecordingMetadata(
         playingNow.track_metadata.track_name,
         playingNow.track_metadata.artist_name,
-        false
+        true
       );
       playingNow.track_metadata.mbid_mapping = metadata as MBIDMapping;
```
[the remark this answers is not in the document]
KasukabeDefenceForce: I created a new variable "response" to store the receive the data from the API call and then assigned it to "playing.track_metadata.mbid_mapping". Assigning "metadata.metadata" displayed the artist name and song name without hyperlinks. The "metadata" value was destructured from the response variable in order to get the "cca_id" and "cca_release_mbid", if the response variable was not null.

frontend/js/src/user/Listens.tsx
```diff
@@ -240,12 +240,16 @@ export default class Listens extends React.Component<
     const playingNow = newPlayingNow;
     const { APIService } = this.context;
     try {
-      const metadata = await APIService.lookupRecordingMetadata(
+      const metadata: MetadataLookup | null = await APIService.lookupRecordingMetadata(
```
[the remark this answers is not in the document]
KasukabeDefenceForce: Yes, I think it is better to rename it.

frontend/js/src/user/Listens.tsx
```diff
@@ -240,12 +240,19 @@ export default class Listens extends React.Component<
...
+      const response = await APIService.lookupRecordingMetadata(
         playingNow.track_metadata.track_name,
         playingNow.track_metadata.artist_name,
-        false
+        true
       );
-      playingNow.track_metadata.mbid_mapping = metadata as MBIDMapping;
+      if (response) {
+        const { metadata } = response;
+        playingNow.track_metadata.mbid_mapping = response as MBIDMapping;
```
MonkeyDo, line 250: This is currently going to result in a messy listen with lots of unexpected keys, considering the response has a lot more information than what goes into `track_metadata.mbid_mapping`.

Your approach in the following lines with picking properties one by one seems like the best approach.

So I'll ask you to compare the output of the api call ([example](https://api.listenbrainz.org/1/metadata/lookup/?recording_name=Magister+Blues&artist_name=Lemurian+Folk+Songs&metadata=true&inc=release)) with the expected keys in the listen `track_metadata.mbid_mapping` from the `MBIDMapping` type [here](https://github.com/metabrainz/listenbrainz-server/blob/662e847fa78c6623a0ed402143c3481fe843444e/frontend/js/src/utils/types.d.ts#L54) and pick each property one by one, instead of attaching the whole response json.

You'll most likely also be able to drop the hacky `as MBIDMapping` that was there as a bonus :)

MonkeyDo (commented):
Thanks for refactoring, I've looked into it a bit more and suggest this solution:

frontend/js/src/user/Listens.tsx
KasukabeDefenceForce: I have individually assigned the properties as you asked for, but I could not find the 'artists' field in the sample output you gave, so I omitted that field.

MonkeyDo: My bad @KasukabeDefenceForce !

Here is an example of the same API call with artist and tag also passed in the `inc=` query parameter:
https://api.listenbrainz.org/1/metadata/lookup/?recording_name=Magister+Blues&artist_name=Lemurian+Folk+Songs&metadata=true&inc=release%20artist%20tag

MonkeyDo (changes requested):
Circled back to this looking at the API response more in detail :)

frontend/js/src/utils/types.d.ts
```diff
@@ -56,8 +56,8 @@ declare type MBIDMapping = {
   release_mbid: string;
   artist_mbids: Array<string>;
   artists?: Array<MBIDMappingArtist>;
-  caa_id?: number;
-  caa_release_mbid?: string;
+  caa_id?: number | null;
```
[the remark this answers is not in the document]
KasukabeDefenceForce: caa_id was declared with the type"?number" which alllowed null. I changed the declaration by shifting the "?" before the colon which will disallow the null.

frontend/js/src/metadata-viewer/types.d.ts
```diff
@@ -81,7 +81,7 @@ declare type ListenMetadata = {
     duration?: number;
   };
   release?: {
-    caa_id: ?number;
+    caa_id?: number;
```
MonkeyDo, line 84: Yes, that looks right, now I get it :)

frontend/js/src/metadata-viewer/types.d.ts
```diff
@@ -81,7 +81,7 @@ declare type ListenMetadata = {
...
     caa_release_mbid: ?string;
```
MonkeyDo, line 85: Same fix here as with the caa_id above?

frontend/js/src/utils/types.d.ts
```diff
@@ -56,7 +56,7 @@ declare type MBIDMapping = {
   release_mbid: string;
   artist_mbids: Array<string>;
   artists?: Array<MBIDMappingArtist>;
-  caa_id?: number | null;
+  caa_id?: number;
   caa_release_mbid?: string | null;
```
MonkeyDo, line 60: And along with the comment I made above, perhaps `caa_release_mbid` too should  have the `| null` removed.

MonkeyDo (commented):
Looks good now ! Thanks !

MonkeyDo:
Also please try to update your branch with master and resolve the conflicts.
Don't hesitate if you need help with that!

MonkeyDo (approved):
Looking good, thank you for your work ! :rocket:

PR #2450, opened 2023-04-14
LB-1250: user-submitted `{recording, artist}_mbid` does not match artist display in listens

amCap1712:
Prefer user-submitted recording mbids over mapped mbids for listen metadata. We already do this for spark dumps and as a result for statistics.

https://github.com/metabrainz/listenbrainz-server/blob/78e37b78ff9e1a246d24adb2a0a0108a1462b284/listenbrainz/listenstore/dump_listenstore.py#L351-L353

Fix the omission in this case as well.

For consistency on the frontend, try to use all data from MB. If the user submitted a recording_mbid, its title will be shown. Otherwise the mapped recording's. If no mapping was made then the user submitted track name will be used.

### DIFF
--- a/frontend/js/src/utils/types.d.ts
+++ b/frontend/js/src/utils/types.d.ts
@@ -52,6 +52,7 @@ declare type MBIDMappingArtist = {
 };
 
 declare type MBIDMapping = {
+  recording_name?: string;
   recording_mbid: string;
   release_mbid: string;
   artist_mbids: Array<string>;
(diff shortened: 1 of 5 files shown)


mayhem (approved):
SQL/python bits look fine to me. thanks for tackling this!

MonkeyDo (approved):
Front-end looks good.

I guess we are solving the ambiguity described in the ticket, but only doing so by ignoring user-submitted data even harder :D
I'm not sure I have a better proposition at the moment, and we probably want to fix mapping issues higher up in the stream in any case.

PR #2453, opened 2023-04-17
LB-1042: Add release group entity stats

amCap1712:
Just like we have top release stats, add top release group stats. release stats haven't been removed and are still generated regularly in case someone wants to use those. Those are just now not being shown on the frontend. Also, some fixes to sitewide stats which were not taking benefit of MB metadata.


listenbrainz/webserver/views/stats_api.py
```diff
@@ -163,6 +163,78 @@ def get_release(user_name):
...
+    :param range: Optional, time interval for which statistics should be returned, possible values are
+        :data:`~data.model.common_stat.ALLOWED_STATISTICS_RANGE`, defaults to ``all_time``
+    :type range: ``str``
+    :statuscode 200: Successful query, you have data!
+    :statuscode 204: Statistics for the user haven't been calculated, empty response will be returned
+    :statuscode 400: Bad request, check ``response['error']`` for more details
+    :statuscode 404: User not found
+    :resheader Content-Type: *application/json*
+    """
+    return _get_entity_stats(user_name, "release_groups", "total_release_group_count")
```
mayhem, line 235: 98% comments, 2% code. :) :)

mayhem (approved):
You and Spark still amaze me. :)

listenbrainz/webserver/views/stats_api.py
amCap1712: :)

PR #2456, opened 2023-04-20
Store MusicBrainz access/refresh token in db

amCap1712:
We need MusicBrainz access/refresh token in ListenBrainz database to submit tags to MusicBrainz.

Note: This only works for new users. Even just logging out users won't work. The access to ListenBrainz application needs to be removed from MB as well in order to make this work. Alternatively, we can add refresh tokens to MB oauth tables manually for existing users and then copy them over to LB.

### DIFF
--- a/listenbrainz/domain/critiquebrainz.py
+++ b/listenbrainz/domain/critiquebrainz.py
@@ -1,19 +1,11 @@
-import time
-
-import json
-import requests
-from requests_oauthlib import OAuth2Session
-from oauthlib.oauth2.rfc6749.errors import InvalidGrantError
 from typing import Iterable
+
+import requests
+from flask import current_app
 
 from data.model.external_service import ExternalServiceType
 from listenbrainz.db.model.review import CBReviewMetadata
-from listenbrainz.db import external_service_oauth
-
-from flask import current_app
-
-from listenbrainz.domain.external_service import ExternalService, ExternalServiceInvalidGrantError
-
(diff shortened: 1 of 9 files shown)


listenbrainz/domain/critiquebrainz.py
```diff
@@ -25,76 +17,18 @@
...
-class CritiqueBrainzService(ExternalService):
+class CritiqueBrainzService(BaseBrainzService):
 
     def __init__(self):
-        super(CritiqueBrainzService, self).__init__(ExternalServiceType.CRITIQUEBRAINZ)
-        self.client_id = current_app.config["CRITIQUEBRAINZ_CLIENT_ID"]
-        self.client_secret = current_app.config["CRITIQUEBRAINZ_CLIENT_SECRET"]
-        self.redirect_uri = current_app.config["CRITIQUEBRAINZ_REDIRECT_URI"]
-
-    def add_new_user(self, user_id: int, token: dict) -> bool:
```
mayhem, line 36: Why is all this code no longer needed?

mayhem (approved):
Looks good, but I don't really grok all the Oauth stuff very well, so....

listenbrainz/domain/critiquebrainz.py
amCap1712: Moved to brainz_service.py as it is commonly used between CB and MB auth.

PR #2457, opened 2023-04-21
Prevent redundant requests to CAA from taste page

amCap1712:
While working on an unrelated task I noticed a barrage of API requests being made to CAA on taste page.

image

The MsidMbidModel is used to add track metadata to feedback and pinned recordings. It sets the mapping metadata in track_metadata.additional_info. However, some fields like caa_id and caa_release_mbid belong in mbid_mapping because the frontend doesn't check additional_info field for these. 

Also, if a release_mbid is present in additional_info, it will be treated as user submitted mbid by the frontend. The caa_id/caa_release_mbid fields will be ignored and a redundant API call will be made to CAA with the "user-submitted" release mbid. Therefore, move all fields to mbid_mapping instead. Logically also mbid_mapping makes more sense than additional_info.


mayhem (approved):

mayhem:
Good catch on this one!!

PR #2458, opened 2023-04-21
Use mapping data in recommendation events

amCap1712:
Instead of having the user to submit artist/track/release etc. fields for a recommendation, only accept a msid and mbid (atleast one required). While retrieving the events, add metadata using mapping tables. This has the benefit that we can have cover art and all other metadata for free. Also, ensures a more consistent experience across the website.

There is also a bugfix for when recommendation created using only mbids without msids fails to load.

Before:
image

After:
image


mayhem (approved):
LGTM!

MonkeyDo (approved):
Good improvement, nice job !
:rocket:

PR #2461, opened 2023-04-25
Recording multisearch component

MonkeyDo:
This PR creates a new multimodal search component, combining the ACRM search (track/artist name fuzzy search) and the MusicBrainz lookup by MB URL or recording MBID.

We had one modal (AddListen) use text search only and another one (MBIDMapping) use lookup MBID/URL only which made both a bit less useful.

Now the new combined multisearch should be used whenever we want a very convenient way for users to search for tracks but also provide exact recordings.


mayhem:
Nitpick: "ACRM search (autocomplete track/artist name) "

 is Artist Credit Recording Mapping. Not a great name, but not something I didn't expect to expose to the world.

frontend/js/src/utils/SearchTrackOrMBID.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const throttledHandleValidMBID = useMemo(
+    () =>
+      throttle(
+        async (input: string) => {
+          const newRecordingMBID = RECORDING_MBID_REGEXP.exec(
+            input
+          )![2].toLowerCase();
+
+          try {
+            const lookupResult: MusicBrainzRecordingWithReleases = (await lookupMBRecording(
```
amCap1712, line 85: Alternatively,  you can use [the LB api](https://api.listenbrainz.org/1/metadata/recording/?recording_mbids=78194fbe-3bcd-4afe-b4a6-7a97711c325f&inc=artist%20release%20tag). The one benefit is that it will give you the canonical release metadata.

amCap1712: The benefit is that it will be consistent with labs api which also returns the canonical recording.

MonkeyDo:
I am deciding against using the `https://api.listenbrainz.org/1/metadata/lookup/` endpoint for the text search, because it requires us 1. to know if the user typed a track or artist name and 2. it doesn't do the type of fuzzy search we're looking for in this case.

PR #2462, opened 2023-04-26
Take 2: Improve ListenCard cover art fallback

MonkeyDo:
I've been struggling with the front-end test suite on #2447 to no avail.
Somehow, the commit 368d9d2 is wreaking havoc with Enzyme.

It adds a fallback, in case there is no mbid_mapping info in the listen metadata but a release_mbid is provided.
In that case, it tries to fetch the cover art for said release using the CAA API.

I don't know why, but in CI this results in some tests randomly breaking the test suite (not just failing, but breaking, thus preventing from knowing which test was running at the time).
This is all related to how Enzyme is abandoned and only support React <= 16 while adapters for newer versions are not official packages and have some weird issues.
Time to refactor the entire front-end test suite using a new testing library ! Joy !

In any case, this PR has cherry-picked commits that implement one part of the solution, which makes it less confusing for users: when a listen is mapped but there's no CAA info with it, display coverart placeholder instead of "link with musicbrainz" button (the listen being already linked).


MonkeyDo:
(Code from this PR approved in #2447)

PR #2463, opened 2023-04-27
LB-1259: Deleting a playlist doesn't refresh the playlist page

faraz16iqbal:
# Problem

[LB-1259](https://tickets.metabrainz.org/browse/LB-1259):
While creating, editing, or deleting a new playlist, the Playlists page does not re-render to display the changed state. This occurs due to the props not being loaded by the _PlaylistList.tsx_ file.

# Solution

Load the latest props dynamically in the  _PlaylistList.tsx_ file to display the modified state.

# Action

Implemented the getDerivedStateFromProps function in the _PlaylistList.tsx_ file to fetch the latest props from the parent component. Also added a unique prop to each Playlist being rendered.

### DIFF
--- a/frontend/js/src/playlists/PlaylistsList.tsx
+++ b/frontend/js/src/playlists/PlaylistsList.tsx
@@ -44,10 +44,22 @@ export default class PlaylistsList extends React.Component<
     };
   }
 
+  static getDerivedStateFromProps(
+    nextProps: React.PropsWithChildren<PlaylistsListProps>,
+    prevState: PlaylistsListState
+  ) {
+    if (nextProps.playlists !== prevState.playlists) {
+      return {
+        playlists: nextProps.playlists,
+      };
+    }
+    return prevState;
+  }
+
   async componentDidUpdate(
     prevProps: React.PropsWithChildren<PlaylistsListProps>
   ): Promise<void> {
-    const { user, activeSection, newAlert } = this.props;
+    const { user, activeSection, newAlert, playlists } = this.props;
     const { currentUser } = this.context;
     if (prevProps.activeSection !== activeSection) {
       await this.fetchPlaylists(0);
@@ -178,6 +190,7 @@ export default class PlaylistsList extends React.Component<
                 onSuccessfulCopy={this.onCopiedPlaylist}
                 newAlert={newAlert}
                 selectPlaylistForEdit={selectPlaylistForEdit}
+                key={playlist.identifier}
               />
             );
           })}
--- a/frontend/js/src/playlists/PlaylistsList.tsx
+++ b/frontend/js/src/playlists/PlaylistsList.tsx
@@ -53,7 +53,7 @@ export default class PlaylistsList extends React.Component<
         playlists: nextProps.playlists,
       };
     }
-    return prevState;
+    return null;
   }
 
   async componentDidUpdate(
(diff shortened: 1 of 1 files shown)


faraz16iqbal:
@MonkeyDo can you please review this?

faraz16iqbal:
@MonkeyDo can you please review this?

MonkeyDo:
> @MonkeyDo can you please review this?

Will do.
FYI, there's a way to request a review on github, at the top of the right-hand side column:
image

Better that way than direct mentions :)

faraz16iqbal:
> > @MonkeyDo can you please review this?
> 
> Will do. FYI, there's a way to request a review on github, at the top of the right-hand side column: image
> 
> Better that way than direct mentions :)

Haha yeah I know, I wasn't getting an option to request a review so I pinged you instead.

frontend/js/src/playlists/PlaylistsList.tsx
```diff
@@ -44,10 +44,22 @@ export default class PlaylistsList extends React.Component<
...
+      };
+    }
+    return null;
+  }
+
   async componentDidUpdate(
     prevProps: React.PropsWithChildren<PlaylistsListProps>
   ): Promise<void> {
-    const { user, activeSection, newAlert } = this.props;
+    const { user, activeSection, newAlert, playlists } = this.props;
```
MonkeyDo, line 62: Unnecessary change, maybe left over from trying something else?

MonkeyDo (approved):
Thanks for looking into this !
This will resolve the issue described in the ticket with minimum changes. :+1:

However, the way I see it the separation of concerns between the components at play is flawed. That is usually the conclusion when we need to use `getDerivedStateFromProps` to solve a state issue.

What I mean is that we are updating the state in PlaylistList because we handle pagination there, but also updating the props passed down by the parent (creating/deleting playlists).
I'm not sure why I did it that way when i last refactored these components, but that's not a good idea!

We will need to rethink the implementation of these pages and refactor accordingly to avoid splitting state management across components.

We will soon be revisiting the recommendations page, so we might simply not need a separate reusable PlaylistList component anymore, in which case w'll be able to hoist PlaylistList back into the Playlist component.

MonkeyDo:
Sorry, I meant to set this review as a comment or change request, please see the comment [above ^](https://github.com/metabrainz/listenbrainz-server/pull/2463#discussion_r1184848763)

frontend/js/src/playlists/PlaylistsList.tsx
faraz16iqbal: resolved

faraz16iqbal:
@MonkeyDo I understand what you're trying to say. I also had to wrap my head a little around the implementation while I was trying to fix it. Would love to help out in refactoring this whenever you plan on doing so!

MonkeyDo:
Welp, looks like we broke pagination on the playlist page !
I merged this too quickly without doing enough testing.

That being said, we have a plan for the recommendations tab which does not involve the PlaylistList component, so we're going to be able to get rid of the complexity we have with reusability.

faraz16iqbal:
Okay, is there an issue which can be picked up to help solve this?

MonkeyDo:
> Okay, is there an issue which can be picked up to help solve this?

I already went and fixed it, I hope you don't mind !

I've been thinking about these playlist and recommendation tabs a fair amount lately, so my head was already in it…

PR #2464, opened 2023-05-01
LB-1263: release-group cover art fallback for user submitted release_mbids

amCap1712:
Currently, we use the cover art from the user submitted release mbid if available. Otherwise, we fallback to using MBID mapper specified cover art. Add another case of fallback - try using cover art from the release group of the release mbid submitted by the user before going for the mapper's cover art.


MonkeyDo (approved):
Yes please !

PR #2465, opened 2023-05-01
LB-1252: Create a script to retrieve spotify user ids for existing users

meycroft:
# Problem

A mechanism for retrieving the Spotify user id of new users has been implemented in #2423. Following this, the next task is to retrieve Spotify user ids for existing users, by adding a script and a new click command to invoke it. The script should query Spotify APIs for all the connected users, fetch their user ids and insert it in the database.

Related JIRA ticket: [LB-1252](https://tickets.metabrainz.org/browse/LB-1252)

# Solution
Added a script to get the Spotify user ids for users existing in the database. Added a click command to invoke the same.

### DIFF
--- a/listenbrainz/domain/spotify_fill_user_id.py
+++ b/listenbrainz/domain/spotify_fill_user_id.py
@@ -1,0 +1,80 @@
+from flask import current_app
+from psycopg2.extras import execute_values
+from spotipy import Spotify
+from sqlalchemy import text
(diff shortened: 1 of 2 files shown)


amCap1712 (approved):
Thanks for the PR. I made one change to exclude users for whom we have previously recorded errors and ran this in production. Works nicely!

PR #2468, opened 2023-05-02
Make the navbar logo always go to the home page.

mayhem:
When we released the redesign the logo always went the the users home feed. We agreed that it should go the home page, at least for the time being.


amCap1712 (approved):
LGTM, Thanks

PR #2469, opened 2023-05-03
Fix URL on stats details previous page button

Maxr1998:
# Problem

The missing link on the previous page button will cause the button to reload the page instead of going to the previous one if JavaScript isn't enabled.

Originally reported by o3 in [the MetaBrainz Community](https://community.metabrainz.org/t/listenbrainz-redesign-beta-release/631498/66?u=maxr1998).

# Solution

Copied the link template from the next button and adapted it accordingly.

# Action

Not applicable.

### DIFF
--- a/frontend/js/src/stats/UserEntityChart.tsx
+++ b/frontend/js/src/stats/UserEntityChart.tsx
@@ -636,7 +636,7 @@ export default class UserEntityChart extends React.Component<
                         }`}
                       >
                         <a
-                          href=""
+                          href={this.buildURLParams(prevPage, range, entity)}
                           role="button"
                           onClick={(e) => {
                             this.handleClickEvent(e, () => {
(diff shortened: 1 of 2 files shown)


Maxr1998:
Looks like I'll have to fix the tests as well.

MonkeyDo (approved):
Thank you for jumping on this fix !

Looks like we missed this regression two years ago 😵‍💫 https://github.com/metabrainz/listenbrainz-server/commit/a60c82e4e3397250d0106b9637c451958827d831#diff-41db651878c91d9f859270388837a887e30b48c9bab8b94fabed3e6f0722309bR587

MonkeyDo:
Looks like the test snapshots need to be updated.
Do you have a local installation of this repo? If so you can run the command `./test.sh fe -u ` and commit to modified `.snap` files.

If you don't, no worries at all, let me know and I'll update the snapshots for you !

Maxr1998:
Not sure if my installation still works, so it might be preferable that you update them. Thanks!

Maxr1998:
Thanks!

PR #2471, opened 2023-05-03
Update add-data.html

hvanderlinde:
Added an entry for SmashTunes, a standalone Mac program, that can submit listens for Apple Music and Spotify.

# Problem

I added support for ListenBrainz to my project. To let other Mac users know this new method of submitting listens exists, a link to my project should be included to the add-data page of ListenBrainz.
# Solution

Another URL entry to the add-data page.

# Action


mayhem (approved):
Thank you for adding this -- much appreciated. We'll do a server update in the next few days when this should go live.

PR #2472, opened 2023-05-04
Refresh similarity datasets weekly

amCap1712:
When we start receiving the production table, we create a temporary table to store the data. Once the dataset has been received completely, we switch the temporary table with the production tables.  We do not store the production dataset in the development tables because production dataset is generated periodically, and we do not want leftovers from earlier runs to mixup the latst runs' results.

We cannot use BulkInsertTable because it is defined in mbid_mapping directory not in listenbrainz. Also, it would need certain changes to support data coming through RMQ.

### DIFF
--- a/admin/timescale/create_tables.sql
+++ b/admin/timescale/create_tables.sql
@@ -202,6 +202,12 @@ CREATE TABLE similarity.recording (
     metadata JSONB NOT NULL
 );
 
+CREATE TABLE similarity.recording_prod (
+    mbid0 UUID NOT NULL,
+    mbid1 UUID NOT NULL,
+    score INT NOT NULL
+);
+
 CREATE TABLE similarity.artist_credit_mbids (
     mbid0 UUID NOT NULL,
     mbid1 UUID NOT NULL,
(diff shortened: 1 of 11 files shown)


admin/timescale/create_tables.sql
```diff
@@ -202,10 +202,23 @@ CREATE TABLE similarity.recording (
     metadata JSONB NOT NULL
 );
 
+CREATE TABLE similarity.recording_prod (
```
mayhem, line 205: One set of tables is for production use and another is for use continuing development, yes? If that is right I think the development tables should have a _dev postfix and the _prod tables should not have a suffix. Having production tables without suffixes makes those tables consistent with the other production tables that do not have _prod appending. What do you think?

mayhem (approved):
One issue about table naming, but I dont feel strongly about it.

admin/timescale/create_tables.sql
amCap1712: Yes. Renaming makes sense, will do.

PR #2473, opened 2023-05-05
LB-1262: Create POST endpoint for get-feedback-for-recordings

faraz16iqbal:
[LB-1262]:(https://tickets.metabrainz.org/browse/LB-1262)

# Problem

Currently, the endpoint /user/<user_name>/get-feedback-for-recordings accepts GET requests and a list of recording MBIDs in the URL.

This is creating some issues for example for playlists with over 80 tracks.

# Solution

Add a method to accept POST requests as well as the corresponding body containing the list of recording MBIDs.

# Action

Added a POST method for /user/<user_name>/get-feedback-for-recordings and updated the frontend API service to make a POST instead of a GET request.

### DIFF
--- a/frontend/js/src/utils/APIService.ts
+++ b/frontend/js/src/utils/APIService.ts
@@ -588,8 +588,17 @@ export default class APIService {
       throw new SyntaxError("Username missing");
     }
 
-    const url = `${this.APIBaseURI}/feedback/user/${userName}/get-feedback-for-recordings?recording_msids=${recording_msids}&recording_mbids=${recording_mbids}`;
-    const response = await fetch(url);
+    const url = `${this.APIBaseURI}/feedback/user/${userName}/get-feedback-for-recordings`;
+    const response = await fetch(url, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json;charset=UTF-8",
+      },
+      body: JSON.stringify({
+        recording_mbids,
+        recording_msids,
+      }),
+    });
     await this.checkStatus(response);
     return response.json();
   };
@@ -601,8 +610,16 @@ export default class APIService {
     if (!userName) {
       throw new SyntaxError("Username missing");
     }
-    const url = `${this.APIBaseURI}/feedback/user/${userName}/get-feedback-for-recordings?recording_mbids=${recording_mbids}`;
-    const response = await fetch(url);
+    const url = `${this.APIBaseURI}/feedback/user/${userName}/get-feedback-for-recordings`;
+    const response = await fetch(url, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json;charset=UTF-8",
+      },
+      body: JSON.stringify({
+        recording_mbids,
+      }),
+    });
     await this.checkStatus(response);
     return response.json();
   };
@@ -692,9 +709,8 @@ export default class APIService {
       };
     }
 
-    const url = `${this.APIBaseURI}/user/${userName}/playlists${
-      createdFor ? "/createdfor" : ""
-    }${collaborator ? "/collaborator" : ""}?offset=${offset}&count=${count}`;
+    const url = `${this.APIBaseURI}/user/${userName}/playlists${createdFor ? "/createdfor" : ""
+      }${collaborator ? "/collaborator" : ""}?offset=${offset}&count=${count}`;
 
     const response = await fetch(url, {
       method: "GET",
--- a/listenbrainz/webserver/views/feedback_api.py
+++ b/listenbrainz/webserver/views/feedback_api.py
@@ -196,7 +196,7 @@ def _get_feedback_for_recording(recording_type, recording):
     })
 
 
-@feedback_api_bp.route("/user/<user_name>/get-feedback-for-recordings", methods=["GET"])
+@feedback_api_bp.route("/user/<user_name>/get-feedback-for-recordings", methods=["GET", "POST"])
 @crossdomain
 @ratelimit()
 def get_feedback_for_recordings_for_user(user_name):
@@ -225,13 +225,21 @@ def get_feedback_for_recordings_for_user(user_name):
     :statuscode 200: Yay, you have data!
     :resheader Content-Type: *application/json*
     """
-
-    msids_unparsed = request.args.get("recording_msids")
-    if msids_unparsed is None:
-        msids_unparsed = request.args.get("recordings")
-    mbids_unparsed = request.args.get("recording_mbids")
+    msids_unparsed = None
+    mbids_unparsed = None
 
     recording_msids, recording_mbids = [], []
+    if request.method == "GET":
+        msids_unparsed = request.args.get("recording_msids")
+        if msids_unparsed is None:
+            msids_unparsed = request.args.get("recordings")
+        mbids_unparsed = request.args.get("recording_mbids")
+    elif request.method == "POST":
+        msids_unparsed = request.json.get(recording_msids)
+        if msids_unparsed is None:
+            msids_unparsed = request.json.get("recordings")
+        mbids_unparsed = request.json.get("recording_mbids")
+        
     if msids_unparsed:
         recording_msids = parse_param_list(msids_unparsed)
     if mbids_unparsed:


faraz16iqbal:
@MonkeyDo, I think I'll have to update the tests for this route too, right?

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -225,13 +225,21 @@ def get_feedback_for_recordings_for_user(user_name):
...
+    mbids_unparsed = None
 
     recording_msids, recording_mbids = [], []
+    if request.method == "GET":
+        msids_unparsed = request.args.get("recording_msids")
+        if msids_unparsed is None:
+            msids_unparsed = request.args.get("recordings")
+        mbids_unparsed = request.args.get("recording_mbids")
+    elif request.method == "POST":
+        msids_unparsed = request.json.get(recording_msids)
```
amCap1712, line 238:
```suggestion
        msids_unparsed = request.json.get("recording_msids")
```

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -225,13 +225,21 @@ def get_feedback_for_recordings_for_user(user_name):
...
+        if msids_unparsed is None:
```
amCap1712, line 239: You can avoid this fallback for the POST case because it is a new addition and no one is using it.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -196,7 +196,7 @@ def _get_feedback_for_recording(recording_type, recording):
     })
 
 
-@feedback_api_bp.route("/user/<user_name>/get-feedback-for-recordings", methods=["GET"])
+@feedback_api_bp.route("/user/<user_name>/get-feedback-for-recordings", methods=["GET", "POST"])
```
amCap1712, line 199: I think you may need to add `OPTIONS` here as an additional method.

amCap1712:
Yes, please add a few more tests for the POST method and also update the documentation of the endpoint to mention how to use it with POST.

listenbrainz/webserver/views/feedback_api.py
amCap1712: No, the GET is under use. I just meant that you can remove the check for `recordings` field from POST. That check is only needed in GET.

listenbrainz/tests/integration/test_feedback_api.py
```diff
@@ -1162,6 +1162,123 @@ def test_get_feedback_for_recordings_for_user_invalid_recording(self):
                                    query_string={"recording_mbids": recordings})  # recording_mbids has invalid recording_msid
         self.assert400(response)
         self.assertEqual(response.json["code"], 400)
+    # Started here
+    def test_get_feedback_for_recordings_for_post_method_for_user_valid_mbids(self):
+        inserted_rows = self.insert_test_data_with_mbid(self.user["id"])
+
+        # recording_msids for which feedback records are inserted
+        recordings = inserted_rows[0]["recording_msid"] + "," + inserted_rows[1]["recording_msid"]
```
amCap1712, line 1170: These should be arrays and not comma separated string in case of POST.

amCap1712: Same comment for all other newly added tests.

frontend/js/tests/utils/APIService.test.ts
```diff
@@ -709,7 +709,17 @@ describe("getFeedbackForUserForRecordings", () => {
...
     expect(window.fetch).toHaveBeenCalledWith(
-      "foobar/1/feedback/user/foo/get-feedback-for-recordings?recording_msids=bar,baz&recording_mbids=new,old"
+      "foobar/1/feedback/user/foo/get-feedback-for-recordings",
+      {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json;charset=UTF-8",
+        },
+        body: JSON.stringify({
+          recording_msids: "bar,baz",
```
amCap1712, line 719: Same comment as for backend tests. This should be array like `["bar", "baz"] instead.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -208,30 +208,47 @@ def get_feedback_for_recordings_for_user(user_name):
...
+    msids_unparsed = None
+    mbids_unparsed = None
 
     recording_msids, recording_mbids = [], []
+    if request.method == "GET":
+        msids_unparsed = request.args.get("recording_msids")
+        if msids_unparsed is None:
+            msids_unparsed = request.args.get("recordings")
+        mbids_unparsed = request.args.get("recording_mbids")
+    elif request.method == "POST":
```
amCap1712, line 248: I just noticed that this is accepting a comma separated string for POST as well. In case of POST we want a proper array, so the msids/mbids will not need to parsed.

frontend/js/src/utils/APIService.ts
```diff
@@ -588,8 +588,17 @@ export default class APIService {
...
 
-    const url = `${this.APIBaseURI}/feedback/user/${userName}/get-feedback-for-recordings?recording_msids=${recording_msids}&recording_mbids=${recording_mbids}`;
-    const response = await fetch(url);
+    const url = `${this.APIBaseURI}/feedback/user/${userName}/get-feedback-for-recordings`;
+    const response = await fetch(url, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json;charset=UTF-8",
+      },
+      body: JSON.stringify({
```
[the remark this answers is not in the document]
faraz16iqbal: I see there are a few different places where this method is being called. I have two approaches in mind. 
1. Create a util function to modify the callers of this APIService method to pass an array instead of a string.
2. Create a util function in the APIService method itself to convert this list of strings into an array and then make a request to the server.

Which approach should I go ahead with?

listenbrainz/webserver/views/feedback_api.py
faraz16iqbal: Resolved

frontend/js/src/utils/APIService.ts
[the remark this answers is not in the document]
amCap1712: As far as I know, we manually convert arrays to comma separated string before passing to this method. So you should change every place its called to not create that string and instead pass an array: For one example:

https://github.com/metabrainz/listenbrainz-server/blob/b8c86e290bb7427f005e441eba951ce0cbd2e6f5/frontend/js/src/pins/UserPins.tsx#L131-L148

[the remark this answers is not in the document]
faraz16iqbal: That makes sense, thanks

frontend/js/tests/utils/APIService.test.ts
faraz16iqbal: Resolved

listenbrainz/tests/integration/test_feedback_api.py
faraz16iqbal: Resolved

faraz16iqbal:
@amCap1712 this is ready to be reviewed

frontend/js/src/user/UserFeedback.tsx
```diff
@@ -175,25 +175,23 @@ export default class UserFeedback extends React.Component<
...
     } = this.state;
 
-    let recording_msids = "";
-    let recording_mbids = "";
+    let recording_msids: string[] = [];
+    let recording_mbids: string[] = [];
     if (feedback?.length && currentUser?.name) {
       recording_msids = feedback
-        .map((item) => item.recording_msid)
+        .map((item) => item.recording_msid!)
```
MonkeyDo, line 182: I'm not a big fan of those typescript `!` to resolve type issues, they tend to hide the problem rather than fix it explicitly.
I do understand that in this case it's typescript not understanding that our map + filter operation results in an array of strings (and not string |undefined | null), but I suggest this more declarative approach:

```typescript
    if (feedback?.length && currentUser?.name) {
      const recording_msids = feedback
        .map((item) => item.recording_msid)
        // Only request non-undefined and non-empty string
        .filter(Boolean)
        // Only request feedback we don't already have
        .filter((msid) => !has(recordingMsidFeedbackMap, msid));
      const recording_mbids = feedback
        .map((item) => item.recording_mbid)
        // Only request non-undefined and non-empty string
        .filter(Boolean)
        // Only request feedback we don't already have
        // @ts-ignore
        .filter((mbid) => !has(recordingMbidFeedbackMap, mbid)) as string[];
        […]
`
``

Note that the MSID array doesn't need anything done to it and already has the expected type, and we add the simple `as string[];` at the end of the MBID array declaration.

I think it's a bit easier to read and future developers have less chance of missing the `!`.

File: frontend/js/tests/utils/APIService.test.ts [hunk 9]
```diff
@@ -697,19 +697,29 @@ describe("getFeedbackForUserForRecordings", () => {
       return Promise.resolve({
         ok: true,
         status: 200,
-        json: () => Promise.resolve({ feedbacl: [] }),
```
MonkeyDo, line 700: Nice catch :p

frontend/js/src/utils/APIService.ts
```diff
@@ -581,15 +581,24 @@ export default class APIService {
 
   getFeedbackForUserForRecordings = async (
```
MonkeyDo, line 582: I think this method and `getFeedbackForUserForMBIDs` can now be merged into a single method.

We would need to make the MSIDs array optional to allow for what getFeedbackForUserForMBIDs does (fetching by MBIDs only):
```typescript
getFeedbackForUserForRecordings = async (
    userName: string,
    recording_mbids: string[], // note that I moved the order of this required parameter
    recording_msids?: string[] // while this one is last since it is optional
  ) => {
```

I'm not certain of the server API endpoint will like `recording_msids: undefined` in the case it's not passed to the method, but you can always add a check if it doesn't like it, something along the lines of:
```
if(recording_msids?.length){
  body.recording_msids = recording_msids;
}
```

MonkeyDo (changes requested):
The front-end parts look good, but I do have a suggestion to refactor code that now look duplicated:

frontend/js/tests/utils/APIService.test.ts
faraz16iqbal: Hahahha thanks xD

frontend/js/src/user/UserFeedback.tsx
faraz16iqbal: Oh yes! Learnt something new. Changing these

frontend/js/src/utils/APIService.ts
faraz16iqbal: Okay cool I'll go ahead and merge these two and refactor the endpoints and tests as well for these

faraz16iqbal:
@MonkeyDo can you review this once again?

frontend/js/src/user/UserFeedback.tsx
```diff
@@ -175,32 +175,30 @@ export default class UserFeedback extends React.Component<
...
         .map((item) => item.recording_msid)
         // Only request non-undefined and non-empty string
         .filter(Boolean)
         // Only request feedback we don't already have
-        .filter((msid) => !has(recordingMsidFeedbackMap, msid))
-        .join(",");
+        .filter((msid) => !has(recordingMsidFeedbackMap, msid)) as string[];
       recording_mbids = feedback
-        .map((item) => item.recording_mbid)
+        .map((item) => item.recording_mbid!)
```
MonkeyDo, line 188: You should be able to remove this now:
```suggestion
        .map((item) => item.recording_mbid)
```

MonkeyDo (changes requested):
Looking good! I'll deploy this to test.listenbrainz to do some in-situ testing later today

MonkeyDo:
I tested this by deploying it to test.listenbrainz.org, I see everything working absolutely fine, including for the original page I created the ticket ofr (a very long playlist with lots of tracks). Nice job !

frontend/js/src/user/UserFeedback.tsx
MonkeyDo: Actually, such a small change, I'm going to commit the suggestion straight away, hope you don't mind :)
I did some testing live on test.LB and it's working fine, so I'm keen to move it forward.

MonkeyDo (approved):

regorxxx:
Would want to note the current [docs](https://listenbrainz.readthedocs.io/en/latest/users/api/recordings.html?highlight=get-feedback#post--1-feedback-user-(user_name)-get-feedback-for-recordings) have the get and post versions mixed, with some comments about the post version which don't apply anymore.

MonkeyDo:
@faraz16iqbal Care to look a the docs for these methods in a separate PR?

faraz16iqbal:
Sure @MonkeyDo, starting the work on it

PR #2474, opened 2023-05-05
Switch listen tables in production

amCap1712:
For the purpose of testing, we had two sets of listen tables, one with the existing schema and another with the new schema. Now that we are satisfied with the testing of the new schema, switch the tables in prod. Rename the existing tables by adding an old suffix and remove the new suffix from the new tables.


mayhem (approved):
:crossed_fingers:

PR #2475, opened 2023-05-05
Listening now: allow users to vote on and add tags

MonkeyDo:
This PR takes over from #2072 which had become very much out of date.

This new branch is rebased on fresh master, and implements the missing pieces of this new feature (auth token, refresh and retry mechanism).

User can now vote on and add tags from the front-end !
For now, until we finesse a few details* this is only available on the Listening Now page as a test bed.
image

\*  < sarcasm > small details like how the tags will look, where we show them, which entities do we want to tag and other trivial points < /sarcasm >


frontend/js/src/metadata-viewer/MetadataViewerPageWrapper.tsx
```diff
@@ -128,6 +128,20 @@ document.addEventListener("DOMContentLoaded", () => {
...
+  try {
+    const response = await fetch(
+      "https://musicbrainz.org/ws/2/genre/all?fmt=txt"
+    );
+    const genresList = await response.text();
+    globalAppContext.musicbrainzGenres = Array.from(genresList.split("\n"));
+  } catch (error) {
+    console.error(error);
+  }
+
```
MonkeyDo, line 144: In a follow-up PR, we should probably load those as props from the back-end, and cache them aggressively?

frontend/js/src/utils/APIService.ts
```diff
@@ -1354,4 +1359,96 @@ export default class APIService {
     await this.checkStatus(response);
     return response.json();
   };
+  /** MusicBrainz */
+
+  getMusicBrainzAccessToken = async (): Promise<string | null> => {
+    const response = await fetch(`${this.APIBaseURI}/profile/mbtoken`);
```
amCap1712, line 1365: This seems like an old leftover?

frontend/js/src/metadata-viewer/MetadataViewer.tsx
```diff
@@ -373,7 +394,11 @@ export default function MetadataViewer(props: MetadataViewerProps) {
               aria-labelledby="headingTwo"
             >
               <div className="panel-body">
-                <TagsComponent tags={metadata?.tag?.release_group} />
+                <TagsComponent
```
amCap1712, line 397: The other TagComponents have a `{Boolean(metadata?...) &&` before them, one is needed here as well probably.

frontend/js/src/metadata-viewer/MetadataViewer.tsx
```diff
@@ -286,7 +310,13 @@ export default function MetadataViewer(props: MetadataViewerProps) {
             aria-labelledby="headingOne"
           >
             <div className="panel-body">
-              <TagsComponent tags={metadata?.tag?.recording} />
+              {Boolean(metadata?.recording) && (
```
amCap1712, line 313: I am not sure what's the purpose of this check. Is it to check if we have recording mbid to vote on? In that case, it would make more sense to do `Boolean(recordingMBID)` instead.

frontend/js/src/metadata-viewer/MetadataViewerPageWrapper.tsx
amCap1712: Sounds good to me, Maybe a ticket or ping me tomorrow and I'll add it.

frontend/js/src/tags/TagsComponent.tsx
```diff
@@ -0,0 +1,210 @@
...
+  entityMBID?: string;
+  tags?: Array<ArtistTag | RecordingTag | ReleaseGroupTag>;
+}) {
+  const { tags, entityType, entityMBID } = props;
+  const [selected, setSelected] = useState<TagOptionType[]>(
+    tags?.length
+      ? tags.map((tag) => getOptionFromTag(tag, entityType, entityMBID))
+      : []
+  );
+  useEffect(() => {
```
amCap1712, line 94: This useEffect seems unnecessary.

amCap1712: I see my understanding was incorrect but you can still avoid an effect it seems: https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes

frontend/js/src/metadata-viewer/MetadataViewer.tsx
MonkeyDo: Yep, I've got that currently in my local branch but not yet pushed, encountered some issues with stale data, still finishing that up.

frontend/js/src/tags/TagsComponent.tsx
MonkeyDo: Ah yes, thank you!
I regularly forget `keys` is how it's done properly, I was scratching my head yesterday trying to figure out what I was doing wrong…

UltimateRiff:
I wonder if we could add support for tagging works here as well? made a Jira ticket as well [LB-1326](https://tickets.metabrainz.org/browse/LB-1326) as a gentle reminder more broadly

mayhem:
Yes, that has been in the works for a while, but the PR is not quite ready for merging yet, IIRC.

mayhem:
I stand corrected -- the PR was merged, but due to an unrelated bug, this feature didn't appear in the Metadata Viewer. If you're using the metadata viewer and you can't add tags, log out from MB on your desktop browser and log in again. Then restart the metadata viewer.

PR #2476, opened 2023-05-05
Small improvements of the Fresh Releases page

MonkeyDo:
Until such a day that we redesign the page, users have been complaining about a coupe of small things that are easy to fix:

- The default page should be the currently logged in user's custom page rather than sitewide
- Clicking on an album should take you to a page where you can play the music (rather than to MB)
- The automatic scrolling to today's date in sitewide page in disturbing and doesn't really add much.


mayhem (approved):
THANK YOU!

PR #2477, opened 2023-05-05
Finish migrating listens table

amCap1712:
1. Remove new- prefix redis cache keys. Was only needed for listens table migration.
2. Use datetime instead of timestamp in mbid_mapper. The listened_at column in listen table is of TIMESTAMPTZ type instead of BIGINT now. So use datetime objects instead of epoch timestamps where needed.
3. Fix reference to track_metadata in data JSONB. track_metadata used to be a key of data JSON column in listen table
earlier but in the new schema data column itself is track_metadata. Therefore, remove erroneous references to ->'track_metadata'.


mayhem (approved):
That was a lot of work to read for something that was mostly a giant rename. programming is weird!

PR #2480, opened 2023-05-10
LB-1271: Fix MBID mapping modal

MonkeyDo:
At some point in #2461 I broke the functionality of the MBID mapping modal without realizing.
Everything appeared fine, but the API call for the mapping was not being triggered, or was cancelled before completion.

This PR fixes the functionality, and also improves the user experience.
For the fix: instead of using bootstrap/jquery `data-*` attributes to handle the modal and backdrop events (click on button => show/hide backdrop and modal) I'm using the modal state instead. This ensure the modal will not close unless there is an actual successful API call (previously the two were decoupled).

As for UX improvements: once the user has selected a recording and after successful API call, the associated ListenCard component on the user page will automatically be updated to reflect the change, without requiring them to refresh the page.
Obviously this is cosmetic, and if the user does refresh the page that won't be kept in memory. However by that point the server should be sending metadata for the listen so it should look smooth and the UX should be significantly improved.


frontend/js/src/mbid-mapping/MBIDMappingModal.tsx
```diff
@@ -66,40 +66,57 @@ export default NiceModal.create(
...
-      closeModal,
-      APIService,
-      selectedRecording,
-      handleError,
-    ]);
+        const recordingMBID =
+          selectedRecordingToListen &&
+          getRecordingMBID(selectedRecordingToListen);
+        if (recordingMBID) {
+          const recordingMSID = _get(
```
amCap1712, line 84: IIRC, we have a getRecordingMSID function somewhere. best to reuse that imo.

amCap1712 (approved):
LGTM, Thanks

PR #2483, opened 2023-05-10
LB-1269: Separate follow listens from feed

07jasjeet:
# Problem

1) [LB-1269](https://tickets.metabrainz.org/browse/LB-1269)
According to the ListenBrainz Redesign, listen events (from followed users) are going to be separated from other feed events. This is to avoid shadowing due to high number of listens events when compared to all other events combined. 

2) Documentation was outdated.

# Solution

Procedure mentioned in ticket's description.

# Action

1) Created a new endpoint that just gives us listen events from followed users. No limit to how many listen events are given per user. Max items per get is 25 iff 25 listens are available. Else, all the listen in span of 7 days will be returned.
2) Updated documentation.


listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -254,6 +259,7 @@ def user_feed(user_name: str):
 
     users_following = db_user_relationship.get_following_for_user(user['id'])
 
+    # TODO: Remove these listen events from event list after listen events endpoint is active.
```
amCap1712, line 262: Do you mean when the feed page UI has been updated instead?

07jasjeet: Yes

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -338,6 +344,53 @@ def user_feed(user_name: str):
...
+
+    :param user_name: The MusicBrainz ID of the user whose timeline is being requested.
+    :type user_name: ``str``
+    :param max_ts: If you specify a ``max_ts`` timestamp, events with timestamps less than the value will be returned
+    :param min_ts: If you specify a ``min_ts`` timestamp, events with timestamps greater than the value will be returned
+    :reqheader Authorization: Token <user token>
+    :reqheader Content-Type: *application/json*
+    :statuscode 200: Successful query, you have feed listen-events!
+    :statuscode 400: Bad request, check ``response['error']`` for more details.
+    :statuscode 401: Unauthorized, you do not have permission to view this user's feed.
```
amCap1712, line 362: 401 is when the credentials are missing or invalid. 403 when the creds are valid but do not have access to the resource. So this needs to be fixed like:

https://github.com/metabrainz/listenbrainz-server/blob/3e4e2c216d4f000c0ef21d8fe24fc3d940b7dc38/listenbrainz/webserver/views/api.py#L656-L657

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -338,6 +344,53 @@ def user_feed(user_name: str):
...
+    :statuscode 404: User not found
+    :resheader Content-Type: *application/json*
+    '''
+
+    user = validate_auth_header()
+    if user_name != user['musicbrainz_id']:
+        raise APIUnauthorized("You don't have permissions to view this user's timeline.")
```
amCap1712, line 369: This should return a APIForbidden instead.

amCap1712 (changes requested):
Thanks for the PR. mostly looks good. A few changes as suggested in the comments and some tests for the new endpoint. See existing [tests](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/tests/integration/test_feed_api.py) for feed api for reference.

listenbrainz/webserver/views/user_timeline_event_api.py
07jasjeet: Done

listenbrainz/webserver/views/user_timeline_event_api.py
07jasjeet: Done

07jasjeet:
> Thanks for the PR. mostly looks good. A few changes as suggested in the comments and some tests for the new endpoint. See existing [tests](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/tests/integration/test_feed_api.py) for feed api for reference.

Just added the test. Most of the work was already done just had to change the URL for the request and make a new function.

07jasjeet:
@amCap1712 test added, did changes to limit and per_user_limit by supplying count arg itself. Default = 25 (As decided) and Max = 100 (Should we decrease it?)

amCap1712:
According to previous IRC discussions, we rather not have any limits on per user listens. Therefore, it would be a better option to paginate based on timestamps instead of count and offsets.

amCap1712:
To do so, you will need to add a new method analogous to:
https://github.com/metabrainz/listenbrainz-server/blob/f1c70fb4510039eda55d6a6dcbc64f0551f74329/listenbrainz/listenstore/timescale_listenstore.py#L353

1. The method should accept users, min_ts: datetime, max_ts: datetime as arguments.
2. Build the filters_list in the same way as in that method.
3. In the SQL query itself use `ORDER listened_at DESC` to get newer listens first.
4. For testing you can use some values of min_ts and max_ts from other tests.

Once this method is done, the API can be updated to use it instead of the existing limited one.
1. Accept min_ts and max_ts as params in the API.
2. Impose a constraint that min_ts cannot be more than say `7 days` in the past. We want recent listens afterall.
3. Default min_ts to the `7 days` or a shorter time frame depending on how recent you want to consider.

For the client calling the API,
1. For the first batch of listens, call the API without any params.
2. For subsequent batches, use the **minimum timestamp** you received in the previous batch as the value of **max_ts** in this API call.

07jasjeet:
@amCap1712 
Did the requested changes: 
1) Now, only min_ts and max_ts are used to get listens.
2) Search space limited to 7 days.
3) Max items per get = 25
4) Test updated with one listen out of search window.

07jasjeet:
Not sure why tests are failing, all of them passed on local.

MonkeyDo:
@07jasjeet I see that one of the tests is failing; something to look at when you have the time :)
https://github.com/metabrainz/listenbrainz-server/actions/runs/5235373522/jobs/9452173761?pr=2483#step:6:1892

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -454,6 +454,111 @@ def fetch_recent_listens_for_users(self, users, min_ts: datetime = None, max_ts:
...
+        args = {"user_ids": tuple(user_id_map.keys()), "limit": max(limit, 100)}
+
+        # min_ts and max_ts must exist.
+        if min_ts and max_ts:
+            args["min_ts"] = min_ts
+            args["max_ts"] = max_ts
+        else: 
+            raise ValueError()
+
+        filters = """user_id IN :user_ids
```
amCap1712, line 477: Since the fliters are always the same, just write in the query directly.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -454,6 +454,111 @@ def fetch_recent_listens_for_users(self, users, min_ts: datetime = None, max_ts:
...
+        user_id_map = {user["id"]: user["musicbrainz_id"] for user in users}
+
+        args = {"user_ids": tuple(user_id_map.keys()), "limit": max(limit, 100)}
+
+        # min_ts and max_ts must exist.
+        if min_ts and max_ts:
+            args["min_ts"] = min_ts
+            args["max_ts"] = max_ts
+        else: 
+            raise ValueError()
```
amCap1712, line 475: add an informative message to the ValueError if you want to use it. Alternatively, ensure wherever this function is called that specifies the max_ts/min_ts always. Since, it will be only 1 place. This alternative solution should be fine.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -454,6 +454,111 @@ def fetch_recent_listens_for_users(self, users, min_ts: datetime = None, max_ts:
                 ))
         return listens
 
+
+    def fetch_recent_listens_for_users_new(self, users, min_ts: datetime, max_ts: datetime, limit=25):
```
amCap1712, line 458: Instead of the new suffix, use some more informative name.

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -622,6 +675,64 @@ def get_listen_events(
...
+    if min_ts and max_ts:
+        temp = datetime.utcfromtimestamp(min_ts)
+        max_ts = datetime.utcfromtimestamp(max_ts)
+        if max_ts - temp < DEFAULT_LISTEN_EVENT_WINDOW_NEW:
+            min_ts = temp 
+        else:
+            # If the given interval for search is greater than :DEFAULT_LISTEN_EVENT_WINDOW_NEW:,
+            # then we must limit the search interval to :DEFAULT_LISTEN_EVENT_WINDOW_NEW:.
+            min_ts = max_ts - DEFAULT_LISTEN_EVENT_WINDOW_NEW
+    else:
```
amCap1712, line 700: You can remove this nesting and directly use elif here.

listenbrainz/tests/integration/test_feed_api.py
```diff
@@ -126,6 +128,80 @@ def test_it_sends_listens_for_users_that_are_being_followed(self):
...
+            self.assertEqual(response.json['status'], 'ok')
+
+        # Send 3 listens with lower timestamps for following_user_2
+        for i in range(3):
+            payload['payload'][0]['listened_at'] = ts - 10 - i
+            response = self.send_data(payload, user=self.following_user_2)
+            self.assert200(response)
+            self.assertEqual(response.json['status'], 'ok')
+
+        from datetime import timedelta
```
amCap1712, line 150: `listened_at` field should be in seconds not milliseconds. I think that's why the test is failing.

listenbrainz/listenstore/timescale_listenstore.py
07jasjeet: Since this function is eventually going to replace the old `fetch_recent_listens_for_users` function, i thought a temporary new suffix should do. I'm not sure what to name it any suggestions?

listenbrainz/listenstore/timescale_listenstore.py
07jasjeet: Alright!

listenbrainz/tests/integration/test_feed_api.py
07jasjeet: Done!

listenbrainz/webserver/views/user_timeline_event_api.py
07jasjeet: Done

listenbrainz/listenstore/timescale_listenstore.py
07jasjeet: So, i have already ensured that min_ts and max_ts are defined in above abstractions but added this ValueError() if this function is used somewhere else someday.

07jasjeet: Removed them anyways

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -454,6 +454,105 @@ def fetch_recent_listens_for_users(self, users, min_ts: datetime = None, max_ts:
...
+        """ Fetch recent listens for a list of users.
+
+            users: A list containing the users for which you'd like to retrieve recent listens.
+            min_ts: Only return listens with listened_at after this timestamp, required.
+            max_ts: Only return listens with listened_at before this timestamp, required.
+        """
+
+        user_id_map = {user["id"]: user["musicbrainz_id"] for user in users}
+
+        args = {"user_ids": tuple(user_id_map.keys()), "limit": max(limit, 100)}
```
amCap1712, line 468: The `max(limit, 100)` here seems wrong. You probably want to do a min instead, either way it should be in the caller not here.

07jasjeet: Oopzz will fix it

07jasjeet: Done

amCap1712: Don't calculate min here, instead do it in the function calling this.

07jasjeet: Luckily its already being in validate_get_endpoint_params just had this as failsafe if the functions was used anywhere else. Done

listenbrainz/listenstore/timescale_listenstore.py
07jasjeet: Changed function name and injected "all" in naming schema since the original function returned limited listens.

amCap1712 (approved):
LGTM, Thanks!

PR #2486, opened 2023-05-14
LB-1268: Placeholder for cover art not occupying full size in `Listening Now` unlike normal cover arts

faraz16iqbal:
# Problem

Placeholder for cover art not occupying full size in `Listening Now` unlike regular cover arts. This occurs because the picture icon is actually an img HTML element while the other icons are SVG (FontAwesome)

# Action

Added CSS to make the img occupy full size.

### DIFF
--- a/frontend/css/listens-page.less
+++ b/frontend/css/listens-page.less
@@ -285,11 +285,7 @@
           max-width: @cover-art-thumbnail-size / 2;
           aspect-ratio: unset;
           border-radius: unset;
-        }
-      }
-      .add-cover-art {
-        .img {
-          max-height: 25px;
+          height: auto;
         }
       }
       img {
--- a/frontend/css/listens-page.less
+++ b/frontend/css/listens-page.less
@@ -276,7 +276,6 @@
       .cover-art-fallback {
         background-color: @gray-lighter;
         min-width: @cover-art-thumbnail-size;
-        height: inherit;
         display: flex;
         align-items: center;
         color: #cbcbcb;
@@ -286,6 +285,11 @@
           max-width: @cover-art-thumbnail-size / 2;
           aspect-ratio: unset;
           border-radius: unset;
+        }
+      }
+      .add-cover-art {
+        .img {
+          max-height: 25px;
         }
       }
       img {


frontend/css/listens-page.less
```diff
@@ -288,6 +287,11 @@
           border-radius: unset;
         }
       }
+      .add-cover-art {
+        .img {
+          max-height: 25px;
+        }
+      }
```
MonkeyDo, line 294: I had a closer look after deploying the PR to test.listenbrainz.org

I found that we're not covering all cases with this css, and we end up with another cover art fallback image stretched like this:
image

So I dug into the css a little bit and found a more accurate solution that will cover all our fallback cases:
Instead of applying the 25px height to only the .add-cover-art image child, we need to apply `height: auto` to images in all three of the fallback classes.

The reason is somewhere else in the css definition meant for regular cover art, we set `height: inherit;` which is why we're seeing the issue. Resetting it with `auto` solves that issue without needing to add arbitrary pixel number like I initially suggested :)

it should go like this:
```less
      .add-cover-art,
      .not-mapped,
      .cover-art-fallback {
        [...]
        // We still want to remove "height: inherit;", keep that change
        > * {
          font-size: @cover-art-thumbnail-size / 2;
          max-width: @cover-art-thumbnail-size / 2;
          aspect-ratio: unset;
          border-radius: unset;
          height: auto; // Reset height here
        }
      }
```

And that covers all our cases!
image

MonkeyDo (changes requested):
Thanks for working on this :)
I dug in and found a more elegant solution after seeing an issue

frontend/css/listens-page.less
faraz16iqbal: That was cool! Going ahead and patching this up

MonkeyDo: I think you might have forgotten to push :)

faraz16iqbal: I did xD. Pushed!

MonkeyDo (approved):
Boom! 🚀 

Thanks for the patching !

PR #2488, opened 2023-05-22
Add update interval docs

mayhem:
Users have been asking why some bits of data on the lb.org site are not updating as quickly as expected. To address this I've write a page that summarizes all of the latencies and other update intervals we have. We still need to add a link from the stats (and possibly other pages) to the new page:

  https://listenbrainz.readthedocs.io/en/latest/general/data-update-intervals.html

The ticket for adding a link is: https://tickets.metabrainz.org/browse/LB-1285


mayhem:
@amCap1712 This should be trivial to  review, since we've already reviewed the actual text.

amCap1712 (approved):
LGTM, Thanks

PR #2489, opened 2023-05-22
Pedants being pedants

mayhem:
Not specifying a timezone for "top of the next hour" is not universally clear, because it would not be true if you are in IST. Thanks India. :)


amCap1712 (approved):
LGTM, Thanks

PR #2491, opened 2023-05-23
Add more details on mapping updates and don't freak out!

mayhem:
Based on user feedback, I improve the section about the mapper and its timings as well as a general note about not freaking out and contacting us when in doubt.

### DIFF
--- a/docs/general/data-update-intervals.rst
+++ b/docs/general/data-update-intervals.rst
@@ -25,8 +25,8 @@ This complicated schedule is caused by ListenBrainz having a lot of interconnect
 at different scales. We also urge our users to keep calm when you find that some listens have not been
 recorded -- there are various reasons for that -- see below.
 
-However, sometimes things break -- its just a fact of life. If you think something is broken, please let us know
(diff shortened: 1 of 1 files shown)


docs/general/data-update-intervals.rst
```diff
@@ -22,7 +22,15 @@ Incremental dumps					            Daily
 =============================================== =========================================
 
 This complicated schedule is caused by ListenBrainz having a lot of interconnected parts that work
-at different scales. For more details, read on!
+at different scales. We also urge our users to keep calm when you find that some listens have not been
+recorded -- there are various reasons for that -- see below.
+
+However, sometimes things break -- its just a fact of life. If you think something is broken, please let us know
+either in our forums or on IRC. ( see https://metabrainz.org/contact ). We are very much engaged in
```
kellnerd, line 29:
```suggestion
However, sometimes things break -- it's just a fact of life. If you think something is broken, please let us know
either in our forums or on IRC ( see https://metabrainz.org/contact ). We are very much engaged in
```
:nerd_face:

PR #2495, opened 2023-05-25
LB-1285: Add link to data intervals documentation

MonkeyDo:
Add a link at the top of the user stats page:
image


mayhem (approved):
Great -- can we also please get a link in the Link with MusicBrainz modal?

PR #2497, opened 2023-05-26
LB-1291: Link to Update Intervals doc page

Aerozol:
# Problem

We have a great new ‘expected data update intervals’ doc, but people might not see it.

# Solution

Added a link to the doc from the LB About > Status page


mayhem (approved):
Perfect, thanks!

MonkeyDo (approved):
Thanks!
We might find other pages in the future where a link to that doc would make sense.
I found one yesterday: the LastFM import page; I'm sure we'll find more little by little.

 :rocket:

PR #2499, opened 2023-05-28
LB-1256 Fix uses of APIUnauthorized error

vardansaini:
# Problem

At some places in the api endpoints, we return 401 Unauthorized error instead of 403 Forbidden for wrong auth token.
For instance, see https://github.com/metabrainz/listenbrainz-server/blob/b28e14f2843f221cd5cbe555c9eabdc50c1a6e36/listenbrainz/webserver/views/user_timeline_event_api.py#L252-L253 .

https://tickets.metabrainz.org/browse/LB-1256

# Solution

 401 should be returned for missing or invalid token. If the token is valid but not authorized to access a resource, 403 should be returned. Therefore, the above case should be updated to return a 403 error instead.
So I made changes to 403 at couple of places and added docs.

# Action


mayhem (approved):
Thanks, looks good to me!

amCap1712:
Looks good, the two tests failures need to be resolved by changing `assert401` to `assert403`.

PR #2500, opened 2023-05-29
Fix Spotify playlist export message

MonkeyDo:
Currently exporting a playlist to Spotify does work, but the resulting confirmation message breaks the page.
We were trying to get json from the response, which already returns json.
A very simple change indeed.


mayhem (approved):
Thanks!

PR #2501, opened 2023-05-29
LB-1235: Add ability to delete listens in missing MB data page

MonkeyDo:
One side-effect of the missing mb data page is that it allows to easily identify "wrong" listens —for example scrobbled videos that aren't actually music— without having to go page by page to look for outliers.

A user requested the addition of the "delete listen" menu option we have on the listens page, which makes a lot of sense.
We have the information required for this, so why not?

In order for that to happen, we also remove the `compact` property on the ListenCards in order to show the "…" option menu, which is something I wanted to do anyway so that I (and others) can inspect the listens and take other actions.

I found a better way to store "deleted" listens so that we can hide them across pagination (contrarily to the listens page currently…) but I'm still not satisfied because it doesn't survive page refresh/page navigation.

For that I think I'm going to use the browser's local storage, which will probably need to be garbage-collected regularly, hence wanting to do it in a separate PR.

image


amCap1712:
The solution is fine but if a user views this page from different devices or browsers, these entries will reappear. So may need a disclaimer.

MonkeyDo:
When we first talked about implementing this, we also discussed the idea of fetching the list of listens waiting deletion.
We could do that on every page load (and maybe still store that in local storage).

Do we have an endpoint for that?

MonkeyDo:
Created [LB-1296](https://tickets.metabrainz.org/browse/LB-1296) for hiding staged-for-deletion listens as a follow-up improvement both for the listens page and the missing MB data page

PR #2503, opened 2023-06-09
Redesign Recommendations page

MonkeyDo:
Now that we have pretty solid music recommendations, we really need to improve the recommendations page/tab which is currently very very dull and looks exactly like the playlists page.
This also required a new endpoint `/user/<user>/playlists/recommendations`.

**Before:**
image

**After:**
image


amCap1712 (commented):
Currently doesn't work on test.lb.
image

PR #2504, opened 2023-06-12
Bulk generate playlists for all users using Spark

amCap1712:
Troi is good for experimenting with playlist patches but is too slow for generating those for all users. Most of the data troi needs to generate its playlists is already available in spark, making it simple to do the bulk generation there.


listenbrainz_spark/troi/periodic_jams.py
```diff
@@ -0,0 +1,53 @@
...
+    elif slug == "weekly-exploration":
+        time_filter = "latest_listened_at IS NULL"
+    else:
+        return []
+
+    query = f"""
+        WITH recommendations AS (
+            SELECT user_id
+                 , recording_mbid
+                 , rank() OVER (PARTITION BY user_id ORDER BY score DESC) AS position
```
mayhem, line 28: the recommendation tracks should be sorted randomly, not by order. In troi the playlist gets shuffled before being truncated, but that is hidden in the playlist generator.

mayhem (changes requested):
One minor sort change and that should be it!

listenbrainz_spark/troi/periodic_jams.py
amCap1712: Shuffle before truncating in all cases?

mayhem: I actually went back and looked at exactly what troi does and its the other way around. Sort by ranking, truncate THEN shuffle. Not how I had originally intended, but it works well, so let's continue to do that.

amCap1712: Fixed.

mayhem (approved):
OK, lets get this merged. We'll still need a crontab entry, yes? I think we should run this daily for a while until it is clear that the process is stable. Then we drop back to weekly. How does that sound?

PR #2506, opened 2023-06-20
Support searching Soundcloud for playback in BrainzPlayer

amCap1712:
BrainzPlayer already supported SoundCloud playback if a soundcloud id was present in the original listen. Now we have access to SoundCloud search APIs, so adding search support to SoundCloud Player in BP.


frontend/js/src/user/UserTaste.tsx
```diff
@@ -86,6 +86,7 @@ export default class UserTaste extends React.Component<UserTasteProps> {
           listenBrainzAPIBaseURI={APIService.APIBaseURI}
           refreshSpotifyToken={APIService.refreshSpotifyToken}
           refreshYoutubeToken={APIService.refreshYoutubeToken}
+          refreshSoundcloudToken={APIService.refreshSoundcloudToken}
```
MonkeyDo, line 89: This is getting really annoying, isn't it?
Since each of those functions is just calling the APIService method with the name of the service as a string, I think we should be able to refactor it (separate PR, I don't mind getting in there and doing it) and just pass the `refreshAccessToken` method to BrainzPlayer.

Then in BrainzPlayer we can pass the prop along, for example 
```
<SoundcloudPlayer
  ...
  refreshToken={refreshAccessToken}
```

And finally, in SoundcloudPlayer, we can call `this.props.refreshToken(this.name)` <- the name string is already stored in the class.


Anyway, I'll have a look at refactoring that once this PR is merge, so as not to be disruptive :)

frontend/js/src/utils/utils.tsx
```diff
@@ -131,6 +131,61 @@ const searchForYoutubeTrack = async (
   return null;
 };
 
+const searchForSoundcloudTrack = async (
+  soundcloudToken?: string,
```
MonkeyDo, line 135: This should probably not be allowed to be undefined

frontend/js/src/utils/utils.tsx
```diff
@@ -131,6 +131,61 @@ const searchForYoutubeTrack = async (
...
+  artistName?: string,
+  releaseName?: string
+): Promise<string | null> => {
+  if (!soundcloudToken) return null;
+  let query = trackName ?? "";
+  if (artistName) {
+    query += ` ${artistName}`;
+  }
+  // Considering we cannot tell the Soundcloud API that this should match only an album title,
+  // results are paradoxically sometimes worse if we add it to the query
```
MonkeyDo, line 146: These comments were for the youtube search. I haven't evaluated the SoundCloud search, but maybe this doesn't hold true for it.

frontend/js/src/utils/utils.tsx
```diff
@@ -131,6 +131,61 @@ const searchForYoutubeTrack = async (
...
+    headers: {
+      "Content-Type": "application/json",
+      Authorization: `OAuth ${soundcloudToken}`,
+    },
+  });
+  const responseBody2 = await response2.json();
+  if (!response.ok) {
+    throw responseBody;
+  }
+  return responseBody2?.http_mp3_128_url;
```
MonkeyDo, line 186: Just checking: is this prop fully stable in the response?
The live API docs do seem to say so.
In any case, this code would just safely return undefined otherwise.

MonkeyDo (commented):
Looking good !

Hope you don't mind, I jumped in for a preliminary review of the react bits.
You can safely ignore my rambling about refactoring the refreshToken methods, they're mainly notes for myself.

Let me know if you need any UI work that I can help with.

frontend/js/src/utils/utils.tsx
amCap1712: Yup, this one should be stable.

frontend/js/src/utils/utils.tsx
amCap1712: Fixed

frontend/js/src/user/UserTaste.tsx
amCap1712: Makes sense

frontend/js/src/brainzplayer/SoundcloudPlayer.tsx
```diff
@@ -180,28 +192,124 @@ export default class SoundcloudPlayer
...
-    return false;
+    return true;
   };
 
   datasourceRecordsListens = (): boolean => {
     return false;
   };
 
+  searchAndPlayTrack = async (listen: Listen | JSPFTrack): Promise<void> => {
+    // TODO: Implement token refresh for SoundCloud
```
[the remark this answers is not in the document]
amCap1712: fixed by adding a check in the constructor instead (we do the same for spotify player).

MonkeyDo (approved):
Looks good, thanks !

MonkeyDo:
Server tests are failing, but apart from that I think it is ready to go into beta.

PR #2509, opened 2023-06-26
Integrate Apple Music playback into ListenBrainz 

vardansaini:
# Problem

ListenBrainz currently supports integration with Spotify, but there is a growing demand for integration with other music services. Apple Music is a popular music streaming service, and integrating it into ListenBrainz will provide more options for users to access personalized recommendations.

# Solution

Integrating Apple Music into ListenBrainz involves several steps, starting with OAuth configuration on the frontend. The OAuth process facilitates user authentication and access to their Apple Music account. This integration aims to enhance the ListenBrainz experience by allowing users to play music seamlessly from Apple Music through the BrainzPlayer.

1. OAuth Integration Frontend: To initiate the integration of Apple Music into ListenBrainz, the first step is configuring OAuth on the frontend. This is accomplished using the MusicKit JavaScript library, which offers authentication through the OAuth protocol. To begin, include the MusicKit library in the webpage with the following script:
htmlCopy code
<script src="https://js-cdn.music.apple.com/musickit/v3/musickit.js" data-web-components async></script> 

2. Developer Token and MusicKit Configuration: To configure MusicKit on the frontend, a developer token needs to be generated on the backend and sent to the frontend. After including the MusicKit library, you can configure the MusicKit instance using the generated developer token:
javascriptCopy code
document.addEventListener('musickitloaded', async function () { try { await MusicKit.configure({ developerToken: 'DEVELOPER-TOKEN', app: { name: 'ListenBrainz', build: 'GIT-COMMIT-SHA', } }); const music = MusicKit.getInstance(); } catch (err) { // Handle configuration error } }); 

3. User Authentication and Token Submission: The authorize method provided by MusicKit can be utilized to authenticate the user and gain access to their Apple Music account. Upon successful authorization, the obtained Music-User-Token is sent to the server's callback URL. However, it's important to note that Apple Music does not require a client secret for OAuth; instead, the process is carried out client-side using MusicKit.

4. Backend Processing: The callback URL endpoint for music services receives the authorization code for other services. In the case of Apple Music, the OAuth process is slightly different due to the absence of a client secret. MusicKit is used to handle authorization client-side, and once the user grants access, the music.authorize().then(musicUserToken => {...}) block is executed. The retrieved Music-User-Token is then submitted to the server at the callback URL.

5. Storing Token and Storefront: To ensure a seamless experience, the obtained token and the user's storefront (region) should be stored in the database for future use. This data will be crucial for accessing and customizing the user's Apple Music content.

6. Integrate Apple Music in BrainzPlayer: The BrainzPlayer, which is a web-based music player used in ListenBrainz, can be extended to support Apple Music playback. MusicKit is integrated into the player in a manner similar to its inclusion in the OAuth frontend setup.

7. Player Controls and Event Handling: Once the BrainzPlayer is loaded with MusicKit, various playback controls like play, pause, etc., can be accessed using the MusicKitInstance. Event callbacks can also be added to respond to changes in playback state through the playbackStateWillChange and playbackStateDidChange events.

8. User Authorization Check: Before implementing event callbacks, it's essential to verify whether the user has authorized Apple Music. This step ensures that only authorized users can interact with the Apple Music integration in the player.

9. Track Enqueue and Playback: To play tracks from Apple Music, the MusicKit player allows tracks to be enqueued using the setQueue or playNext method. The Apple Music Identifier of the desired track is used to facilitate playback.

10. Search and Playback: In cases where a specific track lacks an Apple Music Identifier in the ListenBrainz database, the search API can be employed to retrieve relevant tracks. Subsequently, tracks can be enqueued from the search results into the player for seamless playback within the BrainzPlayer interface.

# Action

### BEFORE MIDTEM EVAL
Video of integration so far :- 
https://github.com/metabrainz/listenbrainz-server/assets/61253200/0851bc57-b9f3-4fda-84e2-7a496376fa9d


frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -0,0 +1,365 @@
...
+
+    this.debouncedOnTrackEnd = _debounce(props.onTrackEnd, 700, {
+      leading: true,
+      trailing: false,
+    });
+
+    // Do an initial check of the AppleMusic token permissions (scopes) before loading the SDK library
+    if (AppleMusicPlayer.hasPermissions(props.appleMusicUser)) {
+      window.addEventListener("musickitloaded", this.connectAppleMusicPlayer);
+      loadScriptAsync(
```
hrik2001, line 80: I remember you telling that you had issues with installing new modules. Although using CDN is fine, but leads to security concerns and consistency issues (since we don't use this approach anywhere else)

so why don't you try
- installing musickit with npm, then 
- rebuilding `static_builder` with `--no-cache` flag

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -0,0 +1,365 @@
...
+      "playbackDurationDidChange",
+      this._boundOnPlaybackDurationChange
+    );
+    this.appleMusicPlayer.addEventListener(
+      "nowPlayingItemDidChange",
+      this._boundOnNowPlayingItemChange
+    );
+  };
+
+  onPlaybackStateChange = ({ state: currentState }: any) => {
```
hrik2001, line 281: Rename this as `handlePlaybackStateChange` to make it consistent with other code, also it would be preferable to not use `any` here

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -0,0 +1,365 @@
...
+  }
+
+  public name = "Apple Music";
+  public domainName = "music.apple.com";
+  public icon = faApple;
+  // Saving the access token outside of React state , we do not need it for any rendering purposes
+  // and it simplifies some of the closure issues we've had with old tokens.
+  private accessToken = "";
+
+  private readonly _boundOnPlaybackStateChange: (event: any) => any;
```
hrik2001, line 55: Remove `any` from here too

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -0,0 +1,365 @@
...
+
+    this.accessToken = props.appleMusicUser?.music_user_token || "";
+
+    this.debouncedOnTrackEnd = _debounce(props.onTrackEnd, 700, {
+      leading: true,
+      trailing: false,
+    });
+
+    // Do an initial check of the AppleMusic token permissions (scopes) before loading the SDK library
+    if (AppleMusicPlayer.hasPermissions(props.appleMusicUser)) {
```
hrik2001, line 78: So, not to sound icky, but do we need `hasPermissions`? or can we just do a check if accessToken exists or not. And can we use `musickit.authorize` here? Looks like it's declared at the bottom, but if you think you can use it in the constructor itself for this check then that would be great

hrik2001 (changes requested):

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
vardansaini: same permission has been used in other players and I thought in future we might add some check here

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
vardansaini: done

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
vardansaini: nowhere in apple music kit and in this file they are using handles in any name and hence I am not using it either. But if you still insist I change the name.

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -0,0 +1,365 @@
...
+    if (this.appleMusicPlayer) {
+      this.appleMusicPlayer.pause();
+    }
+  };
+
+  handleAccountError = (): void => {
+    const errorMessage = (
+      <p>
+        In order to play music with AppleMusic, you will need a AppleMusic
+        Premium account linked to your ListenBrainz account.
```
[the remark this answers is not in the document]
vardansaini: I will check if listen recording is possible.

frontend/js/src/brainzplayer/BrainzPlayer.tsx
```diff
@@ -112,7 +116,7 @@ function isListenFromDatasource(
   if (datasource instanceof SoundcloudPlayer) {
     return SoundcloudPlayer.isListenFromThisService(listen);
   }
-  return undefined;
+  return AppleMusicPlayer.isListenFromThisService(listen);
```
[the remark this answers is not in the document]
vardansaini: I will do this. I am still going through the error handling of music kit.

amCap1712:
While testing on test.lb I found a couple of issues:
1. Apple Music Player asks for authorization as soon as the page loads, this seems rather annoying. We should not try to authenticate until the user is actually trying to play with apple music's player.
2. The buttons in the screenshot work correctly but the current selection is not highlighted. This is confusing for the user and needs to be fixed.
image

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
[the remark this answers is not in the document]
amCap1712: The endpoints are available but the listen data provided by those lacks listened_at timestamp so we cannout support it at this time.

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
amCap1712: I checked and now this method is used to check if the user has linked their Apple Music account so this is fine.

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -0,0 +1,372 @@
...
+      await this.connectAppleMusicPlayer();
+      return;
+    }
+    try {
+      await this.appleMusicPlayer.setQueue({
+        song: appleMusicId,
+        startPlaying: true,
+      });
+    } catch (error) {
+      handleError(error.message, "Error playing on Apple Music");
```
MonkeyDo, line 135: Something should happen after catching this error. handleError only shows the message in a toast notification. 
Not sure what should happen though, it should probably depend on the type of error if we can diferentiate that.
Should it retry? should it call `handleAccountError`? `onTrackNotFound` if the song can't be found (unlikely I guess)?
I would again point to the Spotify player as an example for this: 
https://github.com/metabrainz/listenbrainz-server/blob/93c392f9756ff5b688610ecc460bf604afcddc1a/frontend/js/src/brainzplayer/SpotifyPlayer.tsx#L259-L286

frontend/js/src/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -0,0 +1,372 @@
...
+  searchAndPlayTrack = async (listen: Listen | JSPFTrack): Promise<void> => {
+    if (!this.appleMusicPlayer) {
+      return;
+    }
+    const trackName = getTrackName(listen);
+    const artistName = getArtistName(listen);
+    const releaseName = _get(listen, "track_metadata.release_name");
+    const searchTerm = `${trackName} ${artistName} ${releaseName}`;
+    if (!searchTerm) {
+      return;
```
MonkeyDo, line 152: Should call `onTrackNotFound` from this.props.

MonkeyDo (dismissed):
I can't test the music playing because I'm not a premium user, but I have some findings as a non-premium user :p

MonkeyDo:
Hello @vardansaini , what is the status on this PR?

vardansaini:
> Hello @vardansaini , what is the status on this PR?

Hi @MonkeyDo I got super busy with University and some family stuff, I will be back on it as soon as possible. My apologies for the delay!

MonkeyDo:
Resolved merge conflicts, deploying to test.LB to ensure I didn't break anything

listenbrainz/webserver/views/settings.py
```diff
@@ -274,32 +278,38 @@ def music_services_details():
     soundcloud_user = soundcloud_service.get_user(current_user.id)
     current_soundcloud_permissions = "listen" if soundcloud_user else "disable"
 
+    apple_service = AppleService()
+    apple_user = apple_service.get_user(current_user.id)
+    current_apple_permissions = "listen" if apple_user["refresh_token"] else "disable"
```
anshg1214, line 283: `apple_user` can be empty in this case, which is throwing an error.

```suggestion
    current_apple_permissions = "listen" if apple_user and apple_user["refresh_token"] else "disable"
```

anshg1214 (changes requested):
I have tested this PR on test.lb, and have found a couple of issues as discussed on chat. LMK if I can help with anything.

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -92,6 +101,66 @@ export default function MusicServices() {
...
+            title="Success"
+            message="You are now logged in to Apple Music."
+          />
+        );
+      }
+
+      setPermissions((prevState) => ({
+        ...prevState,
+        appleMusic: action,
+      }));
```
anshg1214, line 152: You also need to update the context similar to how it is done with other external services

MonkeyDo: That's already dona above on lines 126 and 139 (depending on the action)

anshg1214 (approved):
LGTM 🚀

PR #2510, opened 2023-06-28
Set the descriptions of the periodic playlists and update troi version.

mayhem:
Update troi version and set the playlist descriptions of the periodic playlists as defined in troi.

TBD: How do I test this??


listenbrainz/db/playlist.py
```diff
@@ -747,7 +750,7 @@ def bulk_insert(slug, playlists):
     playlist_template = SQL("""({creator_id}, %s, {description}, 't', %s, jsonb_build_object('algorithm_metadata', jsonb_build_object('source_patch', {slug})))""")\
         .format(
             creator_id=Literal(TROI_BOT_USER_ID),
-            description=Literal(f"{jam_name} Playlist!"),
+            description=description,
```
amCap1712, line 753: This needs to be enclosed in a `Literal`

mayhem: whops.

PR #2511, opened 2023-06-30
Fixed Error thrown in Stats tab

Shelly011s:
# Problem
When using local server api (localhost:8100) the stats page gives ugly error.
Stats erro

# Solution
StatsError

# Action


amCap1712 (approved):
LGTM, Thanks!

PR #2517, opened 2023-07-06
First commit for artist graph

arsh331:
# Problem

First commit for the interactive similarity graph project

# Solution

# Action


amCap1712 (changes requested):
Thanks for the PR. @mayhem will do a detailed review methinks but just a few steps to avoid existing builds and tests from failing.

1. You only need to retain the files under the `src/` folder. Copy those `.tsx` files so that they are directly under `frontend/js/src/artist-similarity`.
2. The index.css file needs to go in `frontend/css` with an appropriate file name.
3. If there are any new dependencies in the `package.json` file here, add those to the `package.json` file at the root of the repository. Then remove this `package.json` file.
4. Remove all other files/directories.

mayhem:
@amCap1712 : Are your comments valid even if we don't really want to move the code to LB yet? Remember, this is for eval purposes only and the work will continue in codepen (or whatever) for a while longer. I suspect your comments are great for when we want to actually integrate this code with LB.

amCap1712:
@mayhem Ah cool! My understanding was that the code is intended to be merged in the repo but not integrated with the LB site yet. If the code isn't going to be merged, then feel free to ignore the previous comments.

mayhem:
Yeah, no need to merge the code for the time being.

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+        red: color.red,
+        green: color.green,
+        blue: color.blue,
+        opacity: opacity,
+        mixed: "rgba(" + color.red + "," + color.green + "," + color.blue + "," + opacity + ")",
+    }
+}
+
+const colorGenerator = (): ColorType => {
+    var red = Math.floor(Math.random() * 200);
```
mayhem, line 57: We'll get results that are more pleasing to the human eye if we use the hsv_to_rgb function here:

import colorsys
import random

red, green, blue = hsv_to_rgb(random.random(), 1.0, .9)
then convert each from 0->1.0 to 0 -> 255.

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+    var red = Math.floor((color2.red - color1.red) * weight + color1.red);
+    var green = Math.floor((color2.green - color1.green) * weight + color1.green);
+    var blue = Math.floor((color2.blue - color1.blue) * weight + color1.blue);          
+    var opacity = Math.floor((color2.opacity - color1.opacity) * weight + color1.opacity);
+
+    return Color(red, green, blue);
+}
+
+const Data = () => {
+    const LIMIT_VALUE = 18;
```
mayhem, line 74: We should use a better name here. What does it limit? Lets say that it limits the number of beers you're allowed to have, so it would better be BEER_LIMIT_VALUE or some such. :)

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+    const BASE_URL = "https://labs.api.listenbrainz.org/similar-artists/json?algorithm=session_based_days_7500_session_300_contribution_5_threshold_10_limit_100_filter_True_skip_30&artist_mbid=";
+    
+    var artist_mbid = "8f6bd1e4-fbe1-4f50-aa9b-94c450ec0f11";
+    var color1 = colorGenerator();
+    var color2 = colorGenerator();
+
+    const [similarArtists, setSimilarArtists] = useState({} as DatasetResponseType);
+    const [artist, setArtist] = useState({} as DatasetResponseType);
+    const [limit, setLimit] = useState(LIMIT_VALUE);
```
mayhem, line 83: the above comment applies here too. it would be nice to know what is being limited.

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+        .then((data) => processData(data))   
+    }
+    
+    const processData = (data: Array<ApiResponseType>): void => {
+        let mainArtist = data[1] as DatasetResponseType;
+        setArtist(mainArtist);
+
+        let similarArtists = data[3] as DatasetResponseType;
+        setSimilarArtists(similarArtists);
+        setColors([mixColor(colors[0], colors[1], 0.3), color2]);
```
mayhem, line 104: What does .3 imply here? should we defined a constant here?

frontend/js/src/artist-similarity/src/Graph.tsx
```diff
@@ -0,0 +1,112 @@
+import React, {useState} from 'react';
```
mayhem, line 1: I think this file/module should be called SimilarArtistsGraph

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+            
+            else{
+                computedColor = colors[0];
+                similarArtistsList.pop();
+            }
+
+            return {
+                "id": artist.name,
+                "artist_mbid": artist.artist_mbid,
+                "size": artist.artist_mbid === mainArtist.artist_mbid ? 150 : 85,
```
mayhem, line 140: These constants should also be defined, not literals.

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+    // Calculating minScore for normalization which is always the last element of the array (because it's sorted)
+    minScore = similarArtistsList && similarArtistsList[LIMIT_VALUE - 1].score as number;
+    minScore = Math.sqrt(minScore);
+
+    transformedArtists = similarArtistsList && {
+        
+        "nodes": similarArtistsList.map((artist: ArtistType, index: number) => {
+
+            if(artist !== mainArtist) {
```
mayhem, line 126: artist and mainArtist are also not good names. How about newArtist and currentArtist?

frontend/js/src/artist-similarity/src/Graph.tsx
```diff
@@ -0,0 +1,112 @@
...
+        <animated.text 
+        textAnchor="middle"
+        dominantBaseline="alphabetic"
+        fontSize={to([animatedProps.size], size => size / 6)}
+        style={{ pointerEvents: 'none', fill: 'white'}}
+        >
+            {
+            node.id.split(" ").map((word, index) => (
+                index < 2 ?
+                <animated.tspan x={0} dy={index === 0 ? "0" : to([animatedProps.size], size => size / 6)}>
```
mayhem, line 50: Several constants that should be defined here.

mayhem (changes requested):
Overall, looking pretty good from the general functioning side of things -- please be more descriptive in your variable naming and anytime that a constant isn't 0, or 1, please define a well named constant for it. That will make adjustments later much easier.

frontend/js/src/artist-similarity/src/Data.tsx
mayhem: Wow, what a dumbass. @MonkeyDo pointed out that this is javascript, not python. Lol, sorry. Too focused on my current project, sorry. :)

MonkeyDo: But this ain't python :D

However we do have some color-manipulation library already installed in the repo called tinycolor2: https://bgrins.github.io/TinyColor/

mayhem: Great -- tinycolor("hsv(Math.random(), 100%, 90%)"); should give the results we're looking for.

frontend/js/src/artist-similarity/src/Input.tsx
```diff
@@ -0,0 +1,40 @@
...
+interface InputProps {
+    fetchData: Function;
+    setLimit: Function;
+}
+
+const Input = (props: InputProps) => {
+
+    const handleInput = (event: React.FormEvent<HTMLFormElement>): void => {
+        event.preventDefault();
+        const form = event.target as HTMLFormElement;
```
MonkeyDo, line 12: I see another `as` band-aid :)
Could this possibly be undefined? then we should make sure we plan for that rather than shut Typescript up :)

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+}
+
+const mixColor = (color1: ColorType, color2: ColorType, weight: number): ColorType => {
+    var red = Math.floor((color2.red - color1.red) * weight + color1.red);
+    var green = Math.floor((color2.green - color1.green) * weight + color1.green);
+    var blue = Math.floor((color2.blue - color1.blue) * weight + color1.blue);          
+    var opacity = Math.floor((color2.opacity - color1.opacity) * weight + color1.opacity);
+
+    return Color(red, green, blue);
+}
```
MonkeyDo, line 71: https://github.com/bgrins/TinyColor#color-utilities
has the useful `tinycolor.mix(color1, color2, amount = 50)`

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+const Data = () => {
+    const LIMIT_VALUE = 18;
+    const BASE_URL = "https://labs.api.listenbrainz.org/similar-artists/json?algorithm=session_based_days_7500_session_300_contribution_5_threshold_10_limit_100_filter_True_skip_30&artist_mbid=";
+    
+    var artist_mbid = "8f6bd1e4-fbe1-4f50-aa9b-94c450ec0f11";
```
MonkeyDo, line 77: I think it would make sense to store the artist MBID in this component's state, and pass the `setArtistMBID` function down to the children components instead of passing fetchData. Then you can react to changes to `artistMBID` in this parent component.

In short:
```
const [artistMBID, setArtistMBID] = useState("8f6bd1e4-fbe1-4f50-aa9b-94c450ec0f11");
[…]
// On line 107
useEffect(() => {
        fetchData(artistMBID);
}, [artistMBID]);
[…]
// In the return condition
<Input onArtistChange={setArtistMBID} …
<Graph onArtistChange={setArtistMBID} …
```

MonkeyDo (changes requested):
I know this looks like a lot of feedback in one go, but to be honest I think you did a really good job so far!

The result is already working quite well.
I see no significant mistakes in the code, and it's mostly just small improvements, particularly in TypeScript, which I wouldn't expect you to be very familiar with. It's learning time ! :)

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+    // Calculating minScore for normalization which is always the last element of the array (because it's sorted)
+    minScore = similarArtistsList && similarArtistsList[LIMIT_VALUE - 1].score as number;
+    minScore = Math.sqrt(minScore);
+
+    transformedArtists = similarArtistsList && {
+        
+        "nodes": similarArtistsList.map((artist: ArtistType, index: number) => {
+
+            if(artist !== mainArtist) {
+                var computedScore = minScore / Math.sqrt(artist.score as number);
```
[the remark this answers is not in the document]
MonkeyDo: Ah, as a matter of fact I just hit that type of error on your prototype :D
image

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Renamed to similarArtist and mainArtist

frontend/js/src/artist-similarity/src/Graph.tsx
arsh331: Renamed to SimilarArtistsGraph

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Renamed limit to similarArtistsLimit

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Changed to SIMILAR_ARTISTS_LIMIT_VALUE

frontend/js/src/artist-similarity/src/Graph.tsx
```diff
@@ -0,0 +1,112 @@
...
+interface GraphProps {
+    data: any;
+    fetchData: Function;
+    backgroundColor: string;
+}
+
+const Graph = (props: GraphProps) => {
+    type Node = (typeof props.data)['nodes'][number];
+    type Link = (typeof props.data)['links'][number];
+    const CustomNodeComponent = <Node extends InputNode>({
```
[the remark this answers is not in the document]
arsh331: Yes, you are right. I need it to show the name of the artists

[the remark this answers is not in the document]
arsh331: I tried looking it for a lot but found no other way than defining a CustomNode

frontend/js/src/artist-similarity/src/Graph.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const CustomNodeTooltipComponent = ({ node }: NodeTooltipProps<Node>) => (
+        <div
+            style={{
+                background: node.color,
+                color: '#ffffff',
+                padding: '9px 12px',
+                borderRadius: '3px'
+            }}
+        >
+            <strong>Name: {node.id}</strong>
```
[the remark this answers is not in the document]
arsh331: Fixed

frontend/js/src/artist-similarity/src/Graph.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { InputNode } from '@nivo/network';
+interface GraphProps {
+    data: any;
+    fetchData: Function;
+    backgroundColor: string;
+}
+
+const Graph = (props: GraphProps) => {
+    type Node = (typeof props.data)['nodes'][number];
+    type Link = (typeof props.data)['links'][number];
```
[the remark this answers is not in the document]
arsh331: I cannot think of a better way. If there is one please let me know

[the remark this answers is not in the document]
arsh331: So I did try something out, basically I defined my own NodeType and LinkType. I believe it also solves using any with Data and replace with GraphDataType
```
type NodeType = {
    id: string;
    artist_mbid: string;
    size: number;
    color: string;
    seed: boolean;
    score: number;
}

type LinkType = {
    source: string;
    target: string;
    distance: number;
    strength: number;
}

type GraphDataType = {
    nodes: Array<NodeType>;
    links: Array<LinkType>;
}

interface GraphProps {
    data: GraphDataType;
    fetchData: Function;
    backgroundColor: string;
}
```

frontend/js/src/artist-similarity/src/Graph.tsx
```diff
@@ -0,0 +1,112 @@
+import React, {useState} from 'react';
+import { ResponsiveNetwork, NodeProps, LinkProps, NodeTooltipProps, NetworkSvgProps } from '@nivo/network';
+import { animated, to } from '@react-spring/web';
+import { InputNode } from '@nivo/network';
+interface GraphProps {
+    data: any;
```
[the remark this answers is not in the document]
arsh331: Fixed below

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+    var scoreList: Array<number> = [];
+    var similarArtistsList: Array<ArtistType> = [];
+    var mainArtist: ArtistType = {} as ArtistType;
+    var minScore = 0;
+    
+    const fetchData = (artist_mbid: string): void => {
+        fetch(BASE_URL + artist_mbid)
+        .then((response) => response.json())
+        .then((data) => processData(data))   
+    }
```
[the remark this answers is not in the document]
arsh331: Fixed

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+const Data = () => {
+    const LIMIT_VALUE = 18;
+    const BASE_URL = "https://labs.api.listenbrainz.org/similar-artists/json?algorithm=session_based_days_7500_session_300_contribution_5_threshold_10_limit_100_filter_True_skip_30&artist_mbid=";
+    
+    var artist_mbid = "8f6bd1e4-fbe1-4f50-aa9b-94c450ec0f11";
+    var color1 = colorGenerator();
+    var color2 = colorGenerator();
+
+    const [similarArtists, setSimilarArtists] = useState({} as DatasetResponseType);
+    const [artist, setArtist] = useState({} as DatasetResponseType);
```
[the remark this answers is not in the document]
arsh331: Fixed

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+        setSimilarArtists(similarArtists);
+        setColors([mixColor(colors[0], colors[1], 0.3), color2]);
+    }
+    
+    useEffect(() => {
+        fetchData(artist_mbid);
+        setColors([color1, color2]);
+    }, []);
+
+    similarArtistsList = similarArtists && similarArtists.data && (similarArtists.data.map((artist: any) => artist));
```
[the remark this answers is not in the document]
arsh331: I am sorry the mapping function was left behind from old code.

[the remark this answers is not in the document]
arsh331: Fixed

frontend/js/src/artist-similarity/src/Graph.tsx
arsh331: Fixed.
```
const MAX_LINES = 2;
const MAX_WORD_LENGTH = 10;
```

frontend/js/src/artist-similarity/src/Graph.tsx
```diff
@@ -0,0 +1,112 @@
...
+    fetchData: Function;
```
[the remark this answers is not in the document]
arsh331: fixed

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+            return {
+                "source": mainArtist.name,
+                "target": artist.name,
+                "distance": (artist.artist_mbid != mainArtist.artist_mbid ? scoreList[index] * 250 : 0),
+                "strength": artist.score as number < 5000 ? 2 : artist.score as number < 6000 ? 4 : 8,
+                };
+        }),
+    }
+
+    var backgroundColor = `linear-gradient(` + changeOpacity(colors[1], 0.2).mixed + `,` + changeOpacity(colors[0], 0.2).mixed + `)`;
```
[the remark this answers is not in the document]
arsh331: Fixed

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+                "score": artist.score
+            };
+        }),
+
+        "links": similarArtistsList.map((artist: ArtistType, index: number) => {
+            return {
+                "source": mainArtist.name,
+                "target": artist.name,
+                "distance": (artist.artist_mbid != mainArtist.artist_mbid ? scoreList[index] * 250 : 0),
+                "strength": artist.score as number < 5000 ? 2 : artist.score as number < 6000 ? 4 : 8,
```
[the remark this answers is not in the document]
arsh331: Is this the right way, 
```
strength: similarArtist.score! < 5000 ? 2 : similarArtist.score! < 6000 ? 4 : 8,
```

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+    mainArtist = artist && artist.data && artist.data[0];
+    similarArtistsList && similarArtistsList.push(mainArtist);
+
+    // Calculating minScore for normalization which is always the last element of the array (because it's sorted)
+    minScore = similarArtistsList && similarArtistsList[LIMIT_VALUE - 1].score as number;
+    minScore = Math.sqrt(minScore);
+
+    transformedArtists = similarArtistsList && {
+        
+        "nodes": similarArtistsList.map((artist: ArtistType, index: number) => {
```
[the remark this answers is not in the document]
arsh331: I guess I worked with parsing JSON before and followed the same here. But, it's fixed now.

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+        setArtist(mainArtist);
+
+        let similarArtists = data[3] as DatasetResponseType;
+        setSimilarArtists(similarArtists);
+        setColors([mixColor(colors[0], colors[1], 0.3), color2]);
+    }
+    
+    useEffect(() => {
+        fetchData(artist_mbid);
+        setColors([color1, color2]);
```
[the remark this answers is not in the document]
arsh331: True, I just noticed I can omit setColors in useEffect. I use it again in fetchData to make a new color for the main node so the colors keep changing by a tiny bit after each click. Hope it makes sense.

frontend/js/src/artist-similarity/src/Data.tsx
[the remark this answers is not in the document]
arsh331: I do want to learn a little better how to handle undefined. I tried some changes, basically checking if it's defined and if not setting it to null (i.e. 1 in case of score).

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Implemented your suggestion

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+    var minScore = 0;
+    
+    const fetchData = (artist_mbid: string): void => {
+        fetch(BASE_URL + artist_mbid)
+        .then((response) => response.json())
+        .then((data) => processData(data))   
+    }
+    
+    const processData = (data: Array<ApiResponseType>): void => {
+        let mainArtist = data[1] as DatasetResponseType;
```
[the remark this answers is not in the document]
arsh331: Fixed

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Do we need to multiply random with 360 to generate random color.
```
tinycolor("hsv(" + Math.random() * 360 + ", 100%, 90%)");
```

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Fixed

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: I am not sure if I understand correctly, but I believe removing quotes should do the trick. Please correct me if I am wrong

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+const changeOpacity = (color: ColorType, opacity: number): ColorType => {
+    return {
+        red: color.red,
+        green: color.green,
+        blue: color.blue,
+        opacity: opacity,
+        mixed: "rgba(" + color.red + "," + color.green + "," + color.blue + "," + opacity + ")",
+    }
+}
```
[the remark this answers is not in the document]
arsh331: Fixed

frontend/js/src/artist-similarity/src/Input.tsx
arsh331: Fixed, using event.currentTarget gets rid of the as

frontend/js/src/artist-similarity/src/Data.tsx
mayhem: Yes, quite likely.

frontend/js/src/artist-similarity/src/Data.tsx
mayhem: No, I was talking about the values 150 and 85. What are they?

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,167 @@
...
+          console.log(data);
+          processData(data);
+        }
+        catch (error){
+          //Error message goes here.
+          alert("Something went wrong while loading information, please try again");
+        }
+    }
+    
+    const processData = (data: Array<ApiResponseType>): void => {
```
MonkeyDo, line 97: I see the type of `data` in this function is defined as `Array<ApiResponseType>`, unless I'm missing something I think that might be incorrect. Should it not be `ApiResponseType` instead, which itself contains a `data` property?

I would recommend to choose more descriptive names to avoid confusions like this, maybe: `similarArtistResponse: ApiResponseType`

After you change this type I think you'll find some of the code below will need changing, something like
```
if(similarArtistResponse.type === "dataset"){
// deconstructive assignment
  const [mainArtist, ...similarArtists] = similarArtistResponse.data;
setArtist(mainArtist);
setSimilarArtists(similarArtists);
}
```
For deconstructive assignment, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

frontend/js/src/artist-similarity/src/Graph.tsx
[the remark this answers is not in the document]
MonkeyDo: OK, makes sense.

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Ok. 150 and 85 are the node sizes. I have replaced them with appropriate constants MAIN_NODE_SIZE and SIMILAR_NODE_SIZE

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Actually API response contains is an array. It either contains elements which are markup response or dataset response. The data attribute exists only on the dataset responses and not directly on ApiResponseType. I have modified ApiResponseType to be an array itself which I believe is much clearer.

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,167 @@
...
+    const MIN_LINK_DIST = 50;
+    const MAIN_NODE_SIZE = 150;
+    const SIMILAR_NODE_SIZE = 85;
+    const BACKGROUND_ALPHA = 0.2;
+    const COLOR_MIX_WEIGHT = 0.3;
+    
+    var color1 = colorGenerator();
+    var color2 = colorGenerator();
+
+    const [similarArtists, setSimilarArtists] = useState<DatasetResponseType>();
```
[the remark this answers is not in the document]
arsh331: The way I defined DatasetResponseType, it contains an array of ArtistType as the data field.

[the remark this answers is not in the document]
arsh331: I tried to model it closely to the Api response from the endpoint but I am happy to make changes. You can also take a look at Api response at https://labs.api.listenbrainz.org/similar-artists/json?artist_mbid=1a425bbd-cca4-4b2c-aeb7-71cb176c828a&algorithm=session_based_days_7500_session_300_contribution_5_threshold_10_limit_100_filter_True_skip_30, if that helps to make things clearer.

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,167 @@
...
+          //Error message goes here.
+          alert("Something went wrong while loading information, please try again");
+        }
+    }
+    
+    const processData = (data: Array<ApiResponseType>): void => {
+        let mainArtist = data[1];
+        setArtist(mainArtist.type === "dataset" ? mainArtist : undefined);
+
+        let similarArtists = data[3];
```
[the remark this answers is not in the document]
arsh331: So data[3] is basically the 3rd element from the Api response which contains the list of similar artists.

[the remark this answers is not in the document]
arsh331: Here is the Api endpoint, https://labs.api.listenbrainz.org/similar-artists/json?artist_mbid=1a425bbd-cca4-4b2c-aeb7-71cb176c828a&algorithm=session_based_days_7500_session_300_contribution_5_threshold_10_limit_100_filter_True_skip_30, if that helps

frontend/js/src/artist-similarity/src/Input.tsx
```diff
@@ -0,0 +1,38 @@
...
+        var artist_mbid = form.artist_mbid.value;
+        var limit = form.limit.value;
+        props.setSimilarArtistsLimit(limit);
+        props.onArtistChange(artist_mbid);
+        
+    }
+
+    return (
+        <div>
+            <form method="post" onSubmit={(e) => handleInput(e)}>
```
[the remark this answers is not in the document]
arsh331: Fixed

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,150 @@
...
+            else {
+                computedColor = colors[0];
+            }
+
+            return {
+                id: similarArtist.name,
+                artist_mbid: similarArtist.artist_mbid,
+                size: similarArtist.artist_mbid === mainArtist?.artist_mbid ? MAIN_NODE_SIZE : SIMILAR_NODE_SIZE,
+                color: computedColor.toRgbString(),
+                score: similarArtist.score ?? 0
```
MonkeyDo, line 126: This is indeed a good way to have a default value if score is undefined. :+1:

Just wanted to check that it is expected that the default value is `0` on this line, but it is `1` on line 112 when calculating `computedScore`

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,150 @@
...
+        }
+    }
+    
+    const processData = (dataResponse: ApiResponseType): void => {
+        let mainArtistResponse = dataResponse[1];
+        setMainArtist(mainArtistResponse.type === "dataset" ? mainArtistResponse.data[0] : undefined);
+        
+        let similarArtistsResponce = dataResponse[3];
+        setSimilarArtists(similarArtistsResponce.type === "dataset" ? similarArtistsResponce.data.slice(0, similarArtistsLimit) : undefined);
+
```
MonkeyDo, line 94: Here I would take another approach which might be more readable without having to know the order or format of the API response quite as much. Effectively the same result, but the code is easier to read in my opinion:

```
// Get the datasets out of the API response
const artistsData = dataResponse.filter(response =>
response.type === "dataset");
if(artistsData.length){
  setMainArtist(artistsData[0]);
const similarArtistsResponse = artistsData[1];
  if(similarArtistsResponse?.data?.length){
    setSimilarArtists(similarArtistsResponse.data.slice(0, similarArtistsLimit);
  }
}
```
At each point where i set a value in the state, there is a previous check to ensure it's not undefined, so we don't need the tertiary operator.
What do you think?

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,150 @@
...
+        setColors([tinycolor.mix(color1, color2, COLOR_MIX_WEIGHT), color2]);
+    }
+    
+    useEffect(() => {
+        fetchData(artistMBID);
+    }, [artistMBID]);
+
+    // Calculating minScore for normalization which is always the last element of the array (because it's sorted)
+    var minScore = similarArtists?.[similarArtistsLimit - 1]?.score ?? 0;
```
MonkeyDo, line 103: :+1:

frontend/js/src/artist-similarity/src/Data.tsx
```diff
@@ -0,0 +1,150 @@
...
+    minScore = Math.sqrt(minScore);
+
+    var transformedArtists: GraphDataType = similarArtists! && {
```
MonkeyDo, line 106: Don't like to see those exclamation points ! (here and the line below)
In TypeScript, they mean "ignore the fact that this value could be null or undefined and just go ahead) and should be used only in rare cases.
Here, the `similarArtists` could be null-ish, so we want to first ensure it is not null and not undefined. Same goes for mainArtist.

A simple `if` check is often the most readable, something like this:
```
let transformedArtists: GraphDataType;
if(mainArtist && similarArtists?.length){
  transformedArtists = {…}
}
```

MonkeyDo (changes requested):
Looking a lot better now @arsh331 !
Good work on improving the types and general robustness and readability of the code :)
Some final suggestions for this part of the project

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Fixed.
```
let transformedArtists: GraphDataType = {
        nodes: [],
        links: []
    };
    // Checking if mainArtist and similarArtists are defined
    if(mainArtist && similarArtists?.length) { 
        transformedArtists = {...}
```

MonkeyDo: Right, that will definitely work.
The other option is to only render the graph if `transformedArtists` is not undefined with a
```
{transformedArtists && <SimilarArtistsGraph … }
```

Anyway, that looks solid now 👍

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: I am trying out your suggestion, the only problem I encounter is that for some reason. Typescript detects artistData as Dataset or Markup. So when I try to access its data field it shows that it might be string.
```
Argument of type 'string | ArtistType' is not assignable to parameter of type 'SetStateAction<ArtistType | undefined>'.
  Type 'string' is not assignable to type 'SetStateAction<ArtistType | undefined>'.ts(2345)
```

frontend/js/src/artist-similarity/src/Data.tsx
arsh331: Thinking about if again. I have redefined it as a constant
```
const NULL_SCORE = Infinity;
```
As otherwise it might interfere in case the score is really 1. Because I set up the graph to not render a score of 1 as it's the null case. Moreover, it also covers the case of division by 0.
```
computedScore = minScore / Math.sqrt(similarArtist?.score ?? NULL_SCORE);
``` 
Hope it makes sense.

frontend/js/src/artist-similarity/src/Data.tsx
MonkeyDo: Brilliant! That looks perfect.
And thanks, I learned something today !

MonkeyDo: The article mentions using find() is his example, and that made me realize that would also be ideal for us since we only want the first item.

MonkeyDo (approved):
Love the improvements of the last commit.
Cleaner code, better function names, comments and a great TypeScript solution !
:+1::+1:

We don't want to merge the PR as-is, will all the files and structure of the create-react-app project, but the code looks good to me


I took the liberty of pushing a change (installing the nivo network library in LB's own package.json file) to fix some typescript compilation issues I was seeing in the automated tests

PR #2521, opened 2023-07-08
LB-1270: Similar listens for feed

07jasjeet:
# Description

[LB-1270](https://tickets.metabrainz.org/browse/LB-1270)
New feature! Similar listens are listens from users who are similar to you!

# Action

Created a new function that gets similar users but with extra parameter "id". This "id" is used by get_all_listen_events, so we can create user_id_map. Now instead of deep embedding "similarity" into core functions, a new field is created in APITimelineEvent class that is null by default.

### Repercussions: 
Now every function that uses `APITimelineEvent` class to return responses will have an extra "similarity" dict associated with or in them. We can create a new class inheriting from `APITimelineEvent` to avoid this.


listenbrainz/db/user.py
```diff
@@ -456,6 +456,45 @@ def get_similar_users(user_id: int) -> Optional[SimilarUsers]:
         return SimilarUsers(user_id=user_id, similar_users=users)
 
 
+def get_similar_users_as_list(user_id: int) -> Tuple[List[dict], dict]:
```
amCap1712, line 459: A new method shouldn't be needed for this. There is already an existing method for retrieveing similar users.

amCap1712 (changes requested):
Looks mostly good but needs tests.

listenbrainz/db/user.py
amCap1712: Yup, modifying original method sounds better instead of returning a dict and a list just return the list and let the caller function convert it to a dict where required.

07jasjeet: Alright so that means wherever this function is called rn, I should insert the conversion there. Also old method uses SimilarUsers() class. Should I edit all of its occurrences as well?

amCap1712: Yes. you'll have to edit the other occurrences as well.

07jasjeet:
> Looks mostly good but needs tests.

So since I deleted the new function, I just edited the old tests. If any new tests are required, please let me know.
Also, did what was suggested. Take a look!

listenbrainz/db/user.py
```diff
@@ -437,23 +437,52 @@ def get_users_in_order(user_ids):
...
               FROM recommendation.similar_user r 
               JOIN jsonb_each(r.similar_users) j
                 ON TRUE
               JOIN "user" u
                 ON j.key::int = u.id 
              WHERE user_id = :user_id
-        """), {"user_id": user_id})
-        users = {row.user_name: row.similarity for row in result.fetchall()}
-        return SimilarUsers(user_id=user_id, similar_users=users)
+             {sort_query}
```
amCap1712, line 481: Why not sort always?

amCap1712 (approved):
Thanks for the PR, LGTM!

PR #2522, opened 2023-07-08
Migrate all labs api to have Pydantic support

VishalFTW:
Migrating labs API to have Pydantic model supports

These changes are part of improvements to be made in data set hostler repository. Since these API's use the structure from the repository mentioned, compatibility is required.

Theses changes are followed up from the [issue](https://tickets.metabrainz.org/projects/LB/issues/LB-1243?filter=allopenissues)  and [pr](https://github.com/metabrainz/data-set-hoster/pull/12)


VishalFTW:
@amCap1712 Please take a look and review the changes. Thanks

mayhem:
Will pydantic support be required for all dataset queries? Sometimes I wish to quickly throw up one query and not spend more than 2 minutes hacking together the query for it. It would be nice to have them be optional and then if we merge DSH code into a production setup, we need to socially require Pydantic support. Thoughts?

amCap1712:
We discussed and decided to require pydantic support for all queries.

amCap1712 (changes requested):
I tested this on prod. There are a few issues with this.
1. The signature of the fetch method of all queries needs to be updated to accept a source parameter.
2. Queries recording search are not converting dicts to pydantic models due to which it fails.
3. For similar recordings and similar artists, that accept a list of ids in 1 step and 1 or more normal values otherwise, the design isn't good enough. A little bit of more tinkering is probably need with the web_query_handler in the data-set-hoster.

amCap1712:
I updated the branch with latest DSH changes to test it and it seems to work well. Great Work!

I will discuss with @mayhem once before merging this to ensure we don't break anything else in the process.

amCap1712:
Merging this to a separate feature branch to continue the integration work there.

PR #2528, opened 2023-07-20
Generate weekly playlists once a week

amCap1712:
The cron job is run every hour on Sunday and Monday so that we cover all timezones from UTC-11 to UTC+13. The cronjob invokes the troi playlist generator which then checks the users who just had the monday morning in their timezone. The playlists are then generated for only for these users. Over various runs, all timezones are covered.


mayhem (approved):
Yay!

PR #2529, opened 2023-07-20
Remove candidate set recommendations

amCap1712:
These broke down inadvertently while I was resetting the spark cluster some weeks ago because of the deletion of a file. We haven't used these recs in a while anyway and the file/dataset to generate these has also not been updated in over 2 years. Hence, removing.


mayhem (approved):
And now that I am thinking about it, I can't see myself using top-artist recs either -- I think the whole concept of candidate sets could go as well. If we want something like this in the future, we'd do it in troi.

PR #2531, opened 2023-07-24
Minor fixes to periodic playlists

mayhem:
This PR fixes three things:

- The periodic playlists are currently owned by troi-bot when we would prefer them to be owned by ListenBrainz.
- The periodic playlists didn't set the expires_on additional_metadata in order for the "created for you" page to show the timer for expiring playlists.
- Update the text of the recommended tracks page to remove the top-artist and similar-artist recs.
- Fix auto export playlists to spotify to not mention daily jams, but talk about generated playlists.


amCap1712 (changes requested):
The code for deleting playlists uses TROI_BOT_USER_ID so that code needs to be updated as well to LISTENBRAINZ_USER_ID. I would suggest searching for all usages of TROI_BOT_USER_ID and replacing those with LISTENBRAINZ_USER_ID as appropriate.

mayhem:
Good catch, I just looked at all instances of TROI_BOT_USER_ID in the codebase and checked them -- I fixed one occurrence, however that was not in delete code. Let me look at the troi codebase and check that as well.

mayhem:
Nope, not in troi. I think we're set now unless you remember another place.

mayhem:
You were searching in the wrong branch -- in the re-add-expires branch this has been fixed:

   https://github.com/metabrainz/listenbrainz-server/blob/re-add-expires-at/listenbrainz/troi/weekly_playlists.py

I grepped the whole code base and can't find any other references.

amCap1712:
Oh duh! Yes, that makes sense. Thanks!

amCap1712 (approved):

PR #2534, opened 2023-08-07
Refactor user feedback logic

MonkeyDo:
When we first implemented feedback on Listen Cards, we only did so on one page.
As the ListenCard component was refactored from many different components into a single all-powerful ListenCard, the feedback logic has been copied over to every top-level page.
This meant each top-level page had logic for:
- fetching feedback from the API
- storing the feedback in the component state
- accessing the feedback in the state for a specific recording
- updating/submitting feedback for a recording

I've finally had enough of duplicating the code, and refactored it all into a custom react hook which is only used in one single component (ListenFeedbackComponent).
Each ListenCard, when instantiated, now manages its own feedback automatically using the recording MBID or MSID in its listen.

One of the reason this wasn't done before is that I wanted to avoid pelting our API with 50 separate requests for each page of listens.
Now the useFeedbackMap hook handles fetching feedback in batches (throttled) as well as saving the feedback in a singleton Map object to avoid more API calls.

Note: I completely removed ListenFeedbackComponent.test.tsx test file. I am hitting some issues with Enzyme again which doesn't handle React functional components. I'm working on the tests but they'll probably come in a separate PR as we might need to start running two separate test suites in parallel…


mayhem (approved):
Looks good -- I love how much code is removed!

amCap1712 (approved):
Looks awesome, thanks!!

PR #2537, opened 2023-08-08
Lb radio testing interface

mayhem:
This is a preliminary test UI for LB Radio. It is not intended to be linked in the main site, as it isn't sufficiently user friendly. But in order to get more testing feedback, this UI is needed to make the UI more approachable.

### DIFF
--- a/frontend/js/src/explore/lb-radio/LBRadio.tsx
+++ b/frontend/js/src/explore/lb-radio/LBRadio.tsx
@@ -34,6 +34,7 @@ function LBRadio(props: LBRadioProps) {
   const [isLoading, setLoading] = React.useState(false);
   const [errorMessage, setErrorMessage] = useState<string>("");
   const [title, setTitle] = useState<string>("");
+  const [desc, setDesc] = useState<string>("");
 
   const { APIService, currentUser } = React.useContext(GlobalAppContext);
   const generatePlaylistCallback = React.useCallback(
@@ -90,7 +91,8 @@ function LBRadio(props: LBRadioProps) {
           }
           setJspfPlaylist(payload.jspf);
           setFeedback(payload.feedback);
-          setTitle(payload.jspf?.playlist?.annotation ?? "");
+          setDesc(payload.jspf?.playlist?.annotation ?? "");
+          setTitle(payload.jspf?.playlist?.title ?? "");
         } else {
           const msg = await request.json();
           setErrorMessage(msg?.error);
@@ -239,6 +241,7 @@ function LBRadio(props: LBRadioProps) {
             <Playlist
               playlist={jspfPlaylist?.playlist}
               title={title}
+              desc={desc}
               onSavePlaylist={onSavePlaylist}
               onSaveToSpotify={onSaveToSpotify}
               onExportJSPF={onExportJSPF}
--- a/frontend/js/src/explore/lb-radio/Playlist.tsx
+++ b/frontend/js/src/explore/lb-radio/Playlist.tsx
@@ -15,6 +15,7 @@ type LBRadioFeedbackProps = {
 type PlaylistProps = {
   playlist?: JSPFPlaylist;
   title: string;
+  desc: string;
   onSavePlaylist: () => void;
   onSaveToSpotify: () => void;
   onExportJSPF: () => void;
@@ -43,6 +44,7 @@ export function Playlist(props: PlaylistProps) {
   const {
     playlist,
     title,
+    desc,
     onSavePlaylist,
     onSaveToSpotify,
     onExportJSPF,
@@ -115,6 +117,7 @@ export function Playlist(props: PlaylistProps) {
           </span>
         )}
         <div id="title">{title}</div>
+        <div id="description">{desc}</div>
       </div>
       <div>
         {playlist.track.map((track: JSPFTrack, index) => {
--- a/frontend/css/explore.less
+++ b/frontend/css/explore.less
@@ -98,9 +98,15 @@
 }
 
 #title {
-  font-size: 15pt;
-  font-weight: bold;
-  margin-top: 15px;
+    font-size: 15pt;
+    font-weight: bold;
+    margin-top: 25px;
+}
+
+#desc {
+    font-size: 13pt;
+    color: #444;
+    margin-top: 10px;
 }
 
 #doc-link {


frontend/js/src/explore/lb-radio/Playlist.tsx
```diff
@@ -0,0 +1,135 @@
...
+type PlaylistProps = {
+  playlist?: JSPFPlaylist;
+  title: string;
+  onSavePlaylist: () => void;
+  enableOptions: boolean;
+  onSaveToSpotify: () => void;
+  onExportJSPF: () => void;
+};
+
+export function UserFeedback(props: UserFeedbackProps) {
```
MonkeyDo, line 24: This component's name could probably be more precise

frontend/js/src/explore/lb-radio/LBRadio.tsx
```diff
@@ -0,0 +1,284 @@
...
+  renderRoot.render(
+    <ErrorBoundary>
+      <GlobalAppContext.Provider value={globalAppContext}>
+        <NiceModal.Provider>
+          <LBRadioWithAlertNotifications
+            userArg={user}
+            modeArg={mode}
+            promptArg={prompt}
+            authToken={token}
+            enableOptions={enableOptions}
```
MonkeyDo, line 278: Since this is derived from the presence or absence of token, we don't need a separate prop for it and could instead define it directly in the `Playlist` component using the logged-in user stored in the global context:
```
const { currentUser } = useContext(GlobalAppContext);
const enableOptions = Boolean(currentUser?.auth_token);
`
``

File: frontend/js/src/explore/lb-radio/LBRadio.tsx [hunk 3]
```diff
@@ -0,0 +1,284 @@
...
+  const enableOptions = token !== "";
+  renderRoot.render(
+    <ErrorBoundary>
+      <GlobalAppContext.Provider value={globalAppContext}>
+        <NiceModal.Provider>
+          <LBRadioWithAlertNotifications
+            userArg={user}
+            modeArg={mode}
+            promptArg={prompt}
+            authToken={token}
```
MonkeyDo, line 277: This should already be set in the global context as `currentUser.auth_token`.

frontend/js/src/explore/lb-radio/LBRadio.tsx
```diff
@@ -0,0 +1,284 @@
...
+  const renderRoot = createRoot(domContainer!);
+  const LBRadioWithAlertNotifications = withAlertNotifications(LBRadio);
+
+  const enableOptions = token !== "";
+  renderRoot.render(
+    <ErrorBoundary>
+      <GlobalAppContext.Provider value={globalAppContext}>
+        <NiceModal.Provider>
+          <LBRadioWithAlertNotifications
+            userArg={user}
```
MonkeyDo, line 274: This prop ends up being unused and can be removed (and in the LBRadioProps types as well)

frontend/js/src/explore/lb-radio/Playlist.tsx
```diff
@@ -0,0 +1,135 @@
...
+  feedback: string[];
+};
+
+type PlaylistProps = {
+  playlist?: JSPFPlaylist;
+  title: string;
+  onSavePlaylist: () => void;
+  enableOptions: boolean;
+  onSaveToSpotify: () => void;
+  onExportJSPF: () => void;
```
MonkeyDo, line 21: I personally want to refactor these playlist export functions into a reusable menu component that we can plunk into every page that deals with playlists

frontend/js/src/explore/lb-radio/Playlist.tsx
```diff
@@ -0,0 +1,135 @@
...
+      </div>
+      <div>
+        {playlist.track.map((track: JSPFTrack, index) => {
+          return (
+            <PlaylistItemCard
+              key={`${track.id}-${index.toString()}`}
+              canEdit={false}
+              track={track}
+              currentFeedback={0}
+              updateFeedbackCallback={() => {}}
```
MonkeyDo, line 128: Once  #2534 is finished and merged, these two properties will need to be removed

frontend/js/src/explore/lb-radio/Prompt.tsx
```diff
@@ -0,0 +1,128 @@
...
+    },
+    []
+  );
+
+  React.useEffect(() => {
+    if (prompt.length > 0) {
+      setHideExamples(true);
+      onGenerate(initPrompt, initMode);
+    }
+  }, [initMode, initPrompt, onGenerate, prompt.length]);
```
MonkeyDo, line 49: Since this is only used once when we load the component, we want to use `initPrompt` and not `prompt`:
```suggestion
  React.useEffect(() => {
    if (initPrompt.length > 0) {
      setHideExamples(true);
      onGenerate(initPrompt, initMode);
    }
  }, [initMode, initPrompt, onGenerate]);
```

MonkeyDo (commented):
Front-end bits look good ! Well done !

There is some refactoring that we can do down the line when we revisit this component, hence the comments below, but nothing pressing.

MonkeyDo (approved):
Listening to a couple of playlists on beta right now, it's working a treat!

staticssleever668:
I was just passing by and tried this for fun on https://beta.listenbrainz.org/explore/lb-radio, and I don't see this noted, but it looks like mode (easy/medium/hard) is not defined by default.
It shows as "easy" on the web-site, but when you press "Generate" it shows `The mode parameter must be one of 'easy', 'medium', 'hard'.` and the "mode" parameter in the query string of the GET request is empty (e.g. `https://beta-api.listenbrainz.org/1/explore/lb-radio?prompt=my-prompt&mode=`).

mayhem:
Yes, this was a bug that was introduced when we started cleaning up -- a bit too aggressive that. We'll have this fixed tomorrow. Also, it seems that you can just select the mode before clicking generate and it works ok.

frontend/js/src/explore/lb-radio/LBRadio.tsx
mayhem: And its not even being used yet. I will use it in the future, though, so removing userArg.

frontend/js/src/explore/lb-radio/LBRadio.tsx
mayhem: Fixed.

frontend/js/src/explore/lb-radio/Playlist.tsx
mayhem: Fixed

PR #2539, opened 2023-08-09
Add API endpoint to retrieve artist's popular recordings

amCap1712:
We created the popularity datasets from MLHD dataset, now adding an endpoint to retrieve the most popular recordings for a given artist from the dataset.


mayhem (approved):
Thank you!

PR #2547, opened 2023-08-15
Fix user search when user not logged in

MonkeyDo:
The search endpoint throws an error (`Internal server error // 'NoneType' object is not iterable`) when no user is logged in.
Simple fix to move a couple of lines of code behind the right check that already exists

#LookAtMeIAmThePythonDevNow


amCap1712 (approved):
Thanks for the fix. It was correct but I took the opportunity to do some minor code cleanup while reviewing the file.

PR #2548, opened 2023-08-16
Use listens data in addition to MLHD+ in popularity stats 

amCap1712:
MLHD+ does not have any data from recordings from the past some years, so use it in conjunction with the listen data from LB.


mayhem (approved):
THANK YOU!

PR #2550, opened 2023-08-16
Refactor Spotify Metadata Cache

amCap1712:
We will soon have an Apple Music Metadata cache and a lot of the code can be shared between the two. Hence, refactoring.


amCap1712:
I have not moved the file elsewhere yet because there are no other uses yet, when we have a new use case. We can move it appropriately.

mayhem (approved):
ok on the move. Ideally we should put it into brainzutils -- its quite generic.

PR #2551, opened 2023-08-16
LB-1320: replace cover art fallback image

MonkeyDo:
Replaces the current colorful placeholder (cover art archive logo) with one more in line with our other icons for cover art related issues

**Before:**
image

**After:**
image


MonkeyDo:
@Aerozol I took the screenshots in the description after deploying this PR to beta, so they're the real deal and not mockups.
Let me know if you approve or have some suggestions

Aerozol:
Thanks monkey! I prefer it - do you?

I like the icon you’ve used but it doesn’t match the ‘add image’ icon we currently use (e.g. from LB-1148). Before we put this through I would like to either make a new image for you to use here, or I can adjust the ‘add image’ icon to suit this one. Either is fine by me, so you can make a call. Or flip a coin :)

I don’t know if it’s beyond the scope of this pull/ticket but if it’s easy to add a rollover explaining why this image is appearing that would be very helpful (since it looks like a missing image, but there’s nothing you can do/click/add)

mayhem:
I think it is better, but I think we're wasting an opportunity here. What if the button was nothing more than text that says: "ADD ME!" ? Making it clear that this is a call to action and not a helpless error.

Aerozol:
Mayhem, sadly this one is a helpless error! (monkey knows more about when and why exactly)

In cases where you can add an image the picture icon with a little ‘+’ will appear

mayhem:
Ah, in that case, plz to be ignoring my comment.

MonkeyDo:
@Aerozol Hello!

Right, this uses the FontAwesome icon, I forgot you had made a custom one for the add cover art icon…
I'd rather aim towards the FA icons rather than keep making custom ones, but neither is a big headache.

I was able to create layered icons with the image icon (filling in the sun in the corner) and the plus icon, do you think this would work?

image


https://codepen.io/Monkeyyyy/pen/VwVgpYQ

MonkeyDo:
> I don’t know if it’s beyond the scope of this pull/ticket but if it’s easy to add a rollover explaining why this image is appearing that would be very helpful (since it looks like a missing image, but there’s nothing you can do/click/add)

I'm going to say that for now it's out of scope, because I'm not sure what to do 😵‍💫 
Having a rollover text that says "Could not load cover art" on a link to MB recording doesn't sound quite right, and there is an upcoming improvement that might make this situation much less likely to happen.
Currently in this specific case the cover art section is a link to the MB recording, which means we have a recording MBID available, but somehow no release/RG information in the metadata to load cover art from.
However in this case we should do a couple of MB API calls to get a release— any release —from the recording MBID and display **some** cover art instead of this fallback image.

Aerozol:
Agreed re. using FontAwesome icons wherever possible. My custom image fits the ‘+’ a bit nicer but I’m just patting myself on the back, nobody will notice :P
If you don’t mind updating both, these look great!

I think ‘could not load cover art’ is good for a mouseover - as you say, we can’t load it because there’s no release MBID (since the user can’t change this we shouldn’t tell them the details though)
I think this will be annoying more users than we think, filling in cover art was/is a big last.fm thing, any kind of text feedback to let users know it’s ‘OK’ is important (OK = it’s not okay, but there’s nothing they can do)

re. fixing this, yeah, seems like a great case for canonical metadata! If you do it _right now_ then we don’t have to add/change the mouseover ;) But mouseover seems easier!

MonkeyDo:
> My custom image fits the ‘+’ a bit nicer but I’m just patting myself on the back, nobody will notice :P

I will !
Tried a bit harder to mimic your custom one:
image

PR #2552, opened 2023-08-16
Install and set up React Testing Library, update Jest

MonkeyDo:
There has been a long running issue with our front-end test suite, because the Enzyme framework we use to test React components has been abandoned and does not support modern React (v 17+, with hooks and all that).

That being said, people have been moving to the most popular alternative, React Testing Library.

- **The bad news:** We can't just replace the framework and move on. The testing strategy is very different; where most of our existing tests test the implementation, RTL forces you to test from a user perspective instead (which is good to be honest) and requires a rewrite of our tests.
- **The good news:** Thankfully, we can run both frameworks in parallel without issue, so we can do progressive replacement of the entire test suite file by file. in that same process, we've agreed to drop snapshot tests which aren't particularly useful and are a real PITA.

With this new library installed, I am also converting one single test file (one that has been causing a lot of phantom bugs recently): frontend/js/tests/cb-review/CBReviewModal.test.tsx if you want to get an idea of the testing style (I suggest reviewing side-by-side)

Some notes about Jest: I am also updating the version of Jest (the test runner) which was started in #2173 This requires a few extra changes to adapt to a newer version of jest.
In particular, there are some changes with regards to timer functions and the use of setImmediate/setTimeout.


frontend/js/tests/cb-review/CBReviewModal.test.tsx
```diff
@@ -69,314 +54,269 @@ const apiRefreshSpy = jest
...
   it("renders the modal correctly", async () => {
-    const wrapper = mount(
-      <GlobalAppContext.Provider value={globalProps}>
-        <NiceModal.Provider>
-          <CBReviewModal {...niceModalProps} {...props} />
-        </NiceModal.Provider>
-      </GlobalAppContext.Provider>
+    render(<NiceModal.Provider />);
+    await act(async () => {
+      NiceModal.show(CBReviewModal, { ...props });
```
amCap1712, line 68: Do we need `await` here?

frontend/js/tests/test-utils/rtl-test-utils.tsx
```diff
@@ -0,0 +1,74 @@
...
+          theme="light"
+        />
+      </>
+    );
+  }
+  return render(ui, { wrapper: WithProviders, ...renderOptions });
+};
+
+// re-export everything
+export * from "@testing-library/react";
```
amCap1712, line 71: Is this required, seems a bit counter-intuitive.

frontend/js/tests/test-utils/rtl-test-utils.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+// override render method
+export { customRender as render };
```
amCap1712, line 74: This makes sense but it will also work like magic and will trip us as some point in future. I wonder if we shuold just export it as customRender or lbRender whatever and use it so that its clear that context is being inserted. But then we also use context everywhere so its kind of obvious too. Your call.

amCap1712 (approved):
The new tests look very clear and readable to me. Looks Awesome and thank you for working on this enormous task!

frontend/js/src/cb-review/CBReviewModal.tsx
```diff
@@ -514,7 +514,7 @@ export default NiceModal.create(({ listen }: CBReviewModalProps) => {
         </div>
 
         <div className="checkbox">
-          <label>
+          <label htmlFor="#acceptLicense">
```
MonkeyDo, line 517: The ne testing framework has a strong emphasis on testing for accessibility as we write tests (pushed for targeting elements using accessibility features such as input labels, etc.)
Correcting the label here which was missing the `for` attribute targeting the actual checkbox input.

frontend/js/tests/test-utils/rtl-test-utils.tsx
MonkeyDo: I wondered about this as well.
I copied this code from the setup documentation, it is what they recommend: https://testing-library.com/docs/react-testing-library/setup/#custom-render

However, indeed we could also just export `renderWithContext` or `renderWithProviders` and not re-export the library, then only import this custom render function.

Probably not a bad idea after all.

MonkeyDo: OK, just rewrote this and the one test file that uses RTL to export a `renderWithProviders` util instead, and import the regular methods from the original module.

frontend/js/tests/cb-review/CBReviewModal.test.tsx
MonkeyDo: Yes indeed. I'm still getting used to how to properly deal with asynchronous components and API calls, and running into some issues.
However act does not return a Promise so that would indeed be a big no-no !

PR #2557, opened 2023-08-22
LB-1301:  Inconsistent Playlists dropdown menus

KasukabeDefenceForce:
# Problem
The menu available at the global playlists page doesn't match the one found at the playlist page.

# Solution
Copied the required functions from the class component on the playlist page and rewrote them as functional components in the global playlist page.

# Action


KasukabeDefenceForce:
@MonkeyDo  I made the necessary changes you asked for.

frontend/js/src/playlists/Playlist.tsx
```diff
@@ -392,6 +377,10 @@ export default class PlaylistPage extends React.Component<
     const collaboratorsWithoutOwner = collaborators.filter(
       (username) => username.toLowerCase() !== playlist.creator.toLowerCase()
     );
+    if (isPlaylistOwner(playlist, currentUser)) {
+      this.alertNotAuthorized();
+      return;
+    }
```
MonkeyDo, line 383: This looks like it does the inverse to what was happening previously on line 381 (`if (!this.isOwner())`)

frontend/js/src/playlists/PlaylistCard.tsx
```diff
@@ -31,10 +36,13 @@ export default function PlaylistCard({
   selectPlaylistForEdit,
   showOptions = true,
 }: PlaylistCardProps) {
-  const { APIService, currentUser } = React.useContext(GlobalAppContext);
+  const { APIService, currentUser, spotifyAuth } = React.useContext(
+    GlobalAppContext
+  );
```
MonkeyDo, line 42: As far as I understand, most of the new code added to this file has been implemented in the new PlaylistMenu component, which means it should be removed from this file and should use a `<PlaylistMenu>` instead

MonkeyDo (changes requested):
Thank you for the refactoring work, it's getting there :)

frontend/js/src/playlists/Playlist.tsx
KasukabeDefenceForce: Yes, it must be a typing error I have corrected it.

frontend/js/src/playlists/PlaylistCard.tsx
KasukabeDefenceForce: I removed the redundant code and added the playlist menu component instead.

KasukabeDefenceForce:
@MonkeyDo  Can you tell me why this test is failing? https://github.com/metabrainz/listenbrainz-server/actions/runs/6164409730/job/16730062465?pr=2557#step:6:1012

MonkeyDo:
> @MonkeyDo Can you tell me why this test is failing? https://github.com/metabrainz/listenbrainz-server/actions/runs/6164409730/job/16730062465?pr=2557#step:6:1012

I'm looking at the PR right as we speak :)

There are two tests failing.

One is a failing [snapshot test](https://jestjs.io/docs/snapshot-testing) which you can fix by updating the snapshots.
You can run the command `npm run test:update-snapshots` locally, or the command `./test.sh fe -u` if you want to run the same process in Docker.
Once that has completed you should have a modified file with the extension `.test.tsx.snap`, which you can commit.

The second test is failing because you did not commit or push the code to resolve this issue I pointed out: https://github.com/metabrainz/listenbrainz-server/pull/2557#discussion_r1318572781

MonkeyDo:
@KasukabeDefenceForce I've noticed on your various PRs that you are going way too fast, and not taking the time to make sure you're doing things right. This results in some easily preventable mistakes, some duplicated code, some code changes not related to your PR, and some other mistakes.
It's **always** more important to have higher quality PRs than to have many quick PRs. I mention this because there's been a flurry of PRs from you, and I really do appreciate the enthusiasm :)
However if that means that you didn't give yourself the time to really solve the issue and I have to come up with the solution for you, that's not a great way to spend either of our time :D

Something that could help is to do a full code review of your own PR as if it was someone else's code you, going through all of it and making sure each change does the right thing. I do that for my own PRs every time and catch some mistakes that way (still need someone else's review to catch the other ones though!).

Now for this specific PR, in 797d98a9a in Playlist.tsx you modified the wrong line. This change will need to be reverted, and you'll need to modify the right line as pointed out in [the review comment](https://github.com/metabrainz/listenbrainz-server/pull/2557#discussion_r1318581978).

KasukabeDefenceForce:
@MonkeyDo  Okay I will be more careful from next time.

frontend/js/src/playlists/Playlist.tsx
```diff
@@ -190,56 +185,14 @@ export default class PlaylistPage extends React.Component<
...
-
   deletePlaylist = async (): Promise<void> => {
     const { currentUser } = this.context;
     const { playlist } = this.state;
     if (!currentUser?.auth_token) {
       this.alertMustBeLoggedIn();
       return;
     }
-    if (!this.isOwner()) {
+    if (isPlaylistOwner(playlist, currentUser)) {
```
MonkeyDo, line 195: This is incorrect too, it should be the inverse condition

MonkeyDo (approved):
Thanks, this is looking quite right for the playlist menu!

I just realized there is some more work required to make these new changes work.

Long story short, we are currently rendering some modal components directly [in the top-most page ](https://github.com/metabrainz/listenbrainz-server/blob/81fbdebd50d532878acc933a4f8a913f489fc054/frontend/js/src/playlists/Playlist.tsx#L776-L785)(Playlist.tsx, Playlists.tsx, etc.) and open the modal with bootstrap javascript (with `data-toggle="modal"` and ` data-target`).

But we had one more piece in that mechanism that stored a playlist [in the top page state](https://github.com/search?q=repo%3Ametabrainz%2Flistenbrainz-server%20playlistSelectedForOperation&type=code) in order to edit it. 
Currently, trying to edit one of the playlists results in no action, no modal opened.


We have recently started using the [Nice Modal](https://github.com/eBay/nice-modal-react/) library which has made modals a lot easier to use, and allows us to clean up these hacks we had in place.

With your permission, I would like to take over after merging this PR.
I am going to merge it into a separate feature branch, which will allow me to continue work after you and reimplement those playlist edit and deletion modals for use with NiceModal.

KasukabeDefenceForce:
Okay. Is there any way I can help you on implementing the rest of the part?

MonkeyDo:
> Okay. Is there any way I can help you on implementing the rest of the part?

I don't think so, but thank you :)

PR #2561, opened 2023-08-29
Update missing_data.html

Aerozol:
Updating intro text:
- Added links to MusicBrainz and the ‘how to contribute’ wiki page where helpful
- Clarified that recordings may just need to be linked (aren’t always missing from MB)
- Noted that only the top 200 missing listens are displayed
- Shortened text where possible so the above could be added without making the intro too long

See also: https://community.metabrainz.org/t/missing-data-in-imported-listens/652567/4?u=aerozol


MonkeyDo (approved):
Good improvements :+1:

PR #2562, opened 2023-08-29
Fix unstable sort order in python test

MonkeyDo:
One test keeps having random failures due to the assertion of a list with unstable sort order.
Using `set(…)` to ensure the list elements are equal regardless of sort order.

Recent example of failing test: https://github.com/Aerozol/listenbrainz-server/actions/runs/6008353397/job/16295861342#step:6:1965


amCap1712 (approved):

amCap1712 (changes requested):
Actually thinking more about it we should fix the sort order.

MonkeyDo:
> To be clear, I'll fix it.

Thank you :) I'll close this PR then, it doesn't do anything and the test is failing :D

I looked at the `get_personal_recommendation_events_for_feed` query and to be honest I don't want to touch it lest I break something !

PR #2565, opened 2023-09-02
LB-1280: Manual mbid mapping: Duplicate text/fill search field.

KasukabeDefenceForce:
# Problem
Often when matching listens to MBID's in LB, you're re-typing or copying and pasting the raw text, displayed tantalizingly out of reach a few px above the search field.

# Solution
Added the functionality for a copy button that fills the search field.

# Action


frontend/css/theme/bootstrap/mixins/buttons.less
```diff
@@ -59,6 +59,15 @@
   }
 }
 
+.searchFieldCopyButton{
+    border: none;
+    background: none;
+    padding: 0;
+  &hover{
+    cursor: pointer;
+}
```
MonkeyDo, line 68: I don't think this will work as-is, you're missing a colon:
```suggestion
  &:hover{
    cursor: pointer;
  }
```

frontend/css/theme/bootstrap/mixins/buttons.less
```diff
@@ -59,6 +59,15 @@
   }
 }
 
+.searchFieldCopyButton{
```
MonkeyDo, line 62: I see two small issues here. One is that this is the wrong file to add this CSS: everything in the `frontend/css/theme/bootstrap/` folder is stock Bootstrap files.
All our custom css/less is in the `frontend/css/` folder instead.


Here is what I suggest to match what we currently do for other modals:
1. change the modal's topmost component's `id` to match what we do for other modal components -> `"MBIDMappingModal"`:
https://github.com/metabrainz/listenbrainz-server/blob/7681c8d8e63883b42eea202d4d495ae8221da741/frontend/js/src/personal-recommendations/PersonalRecommendationsModal.tsx#L164
2. create a new `mbid-mapping-modal.less` file
3. Move your less code to the new `mbid-mapping-modal.less` file , under `#MBIDMappingModal` block like we do for the other modals:
https://github.com/metabrainz/listenbrainz-server/blob/7681c8d8e63883b42eea202d4d495ae8221da741/frontend/css/personal-recommendation-modal.less#L3-L4
4. rename `searchFieldCopyButton` to `search-field-copy-button` to match the rest of our css classes names.

MonkeyDo (changes requested):
Thanks for the PR !
Some things to change, but it's nice to see this feature going :)

frontend/js/src/utils/SearchTrackOrMBID.tsx
```diff
@@ -28,6 +30,11 @@ export default function SearchTrackOrMBID({
   const [searchResults, setSearchResults] = useState<Array<ACRMSearchResult>>(
     []
   );
+  useEffect(() => {
+    if (defaultValue) {
+      setInputValue(defaultValue);
+    }
+  }, [defaultValue]);
```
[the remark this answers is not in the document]
KasukabeDefenceForce: This method will only allow us to pass the initial value of the defaultValue prop. If we dont use an useEffect it wont update its value in the state when the value of the prop changes.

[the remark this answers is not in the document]
MonkeyDo: I think this should be resolved by giving the `SearchTrackOrMBID` a key in [MBIDMappingModal.tsx](https://github.com/metabrainz/listenbrainz-server/pull/2565/files#diff-063a292b682bc384ca0e8c6b046ac706d197d3a1372124e309aead502d3ee76eR268)
This ensures the element is recreated when the parent defaultValue changes

```

<SearchTrackOrMBID
  key={defaultValue}
```

MonkeyDo (changes requested):

frontend/js/src/mbid-mapping/MBIDMappingModal.tsx
```diff
@@ -250,9 +266,11 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
               ) : (
                 <div className="card listen-card">
                   <SearchTrackOrMBID
+                    key={defaultValue}
```
MonkeyDo, line 269: Hm, I think maybe I spoke too soon :)

After trying it, the issue is that if I click the button (text is copied as expected), then clear the search input text, then click the button again, nothing happens because the defaultValue hasn't changed in the parent so the `key` here remains the same.

If the value changed in the parent, it would re-render the SearchTrackOrMBID, but as it stands it won't work as we want.
I'm still not comfortable with the useEffect solution you had to react to changes in defaultValue, because it seems like the kind of mechanism that could overwrite the user's input text without a clear action from the user.

One suggestion I can think of si that we can make the `key` change on each button click.
We can add a new state for a number value defaulting to 0, and in `copyTextToSearchField` we increment the value.
Then the `SearchTrackOrMBID` key would be composed from these two bits of data:
```key={`${defaultValue}-${copyTextClickCounter}`}```

What do you think? It seems a bit convoluted compared to the useEffect you used previously, but it ensures we don't update the input unless there is a clear user action…

File: frontend/js/src/mbid-mapping/MBIDMappingModal.tsx [hunk 5]
```diff
@@ -207,16 +215,24 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
...
-                <FontAwesomeIcon
-                  icon={faExchangeAlt}
-                  rotation={90}
-                  size="lg"
-                  color={
-                    selectedRecording ? COLOR_LB_GREEN : COLOR_LB_LIGHT_GRAY
-                  }
-                />
+                <button
+                  className="search-field-copy-button"
```
MonkeyDo, line 221: After testing live, I think we should make it clearer that this is a clickable button. Currently I don't think it's obvious.
I tried with our existing css classes and settled on something.
In the end we don't need the custom `search-field-copy-button` class so I think those .less file changes will need to be removed.

image


The code changes are as follows:
```
<button
  className="btn btn-link btn-rounded"
…
```
And the additional text right under the FontAwesomeIcon:
```
<div class="text-muted">copy text</div>
```

MonkeyDo (changes requested):
Ah, this time it's my turn to have been a bit too hasty with my review :)

I ran the PR and tested it, and came up with a few more suggestions:

frontend/js/src/mbid-mapping/MBIDMappingModal.tsx
KasukabeDefenceForce: Ya, I think this method is also good. I have implemented it and committed the changes.

frontend/js/src/mbid-mapping/MBIDMappingModal.tsx
KasukabeDefenceForce: [Screencast from 13-09-23 11:07:53 PM IST.webm](https://github.com/metabrainz/listenbrainz-server/assets/44491059/8684dfed-b5cc-4d68-99dd-c815a54b9243)
When we hover over the button the text gets underlined and on clicking it the border gets highlighted, is this expected behavior?

MonkeyDo: You're right, we can use `btn-transparent` instead of `btn-link`.
As for the highlighted border, looks like our other buttons have the same style from bootstrap, I'm guessing helpful for visually impared users

MonkeyDo (approved):

MonkeyDo:
Now it looks great, thanks ! :)

PR #2546, opened 2023-08-14
Apple metadata cache

vardansaini:
# Problem

The metadata content resolver is used to store metadata of the entire catalog present in an external service's database. Currently, ListenBrainz has such a resolver for Spotify but not for Apple Music.

# Solution

 As a part of linking their Apple Music account with ListenBrainz, I added a similar content resolver for Apple Music as well.

The code in this pull request defines a Python class `AppleCrawlerHandler` which is a handler for crawling and fetching data related to Apple Music's albums and artists. Here's a summary of the main functionalities and methods in the code:

1. The class is a subclass of `BaseHandler`.

2. In the constructor (`__init__`), the class is initialized with various attributes including the name, external service queue, schema name, and cache key prefix. The constructor also initializes instance variables for the app, discovered albums, discovered artists, and an instance of the `Apple` class.

3. The method `get_seed_albums` retrieves Apple album IDs from new releases for all markets by making requests to the Apple Music API.

4. The method `get_items_from_listen` extracts album IDs from a given "listen" data and creates a list of `JobItem` instances.

5. The method `get_items_from_seeder` takes a message containing Apple album IDs and creates a list of `JobItem` instances.

6. The method `transform_album` transforms raw album data by organizing tracks and artists into appropriate data structures and returns an `Album` object.

7. The method `fetch_albums` fetches album data from the Apple Music API and transforms it using the `transform_album` method. It returns a list of transformed albums and new items to be processed.

8. The method `discover_albums` looks up albums of a given artist to discover more albums for seeding the job queue. It keeps track of discovered artists and albums, and returns a list of new `JobItem` instances.

Overall, the `AppleCrawlerHandler` class is designed to interact with the Apple Music API to fetch and process album and artist data, and generate job items for further processing. The class is part of a larger system or application responsible for crawling and caching Apple Music metadata. 

# Action

PR #2566, opened 2023-09-04
Add apple metadata cache

amCap1712:
closed #2546 mistakenly because the base branch was deleted, so opening a new PR with the changes.


amCap1712:
Uhh, unable to add changes to this. will have to open a new again.

PR #2572, opened 2023-09-08
Increase MAX_ITEMS_PER_GET to 1000

mayhem:
As requested here 

https://community.metabrainz.org/t/allow-getting-more-user-data-through-the-api

Setting the max number of items to get to 1000. Makes sense to me.


amCap1712 (approved):
LGTM, thanks

PR #2573, opened 2023-09-12
LB-1278: Now playing notifications not displayed correctly

KasukabeDefenceForce:
# Problem
When the browser has blocked notifications, the now playing notification was not displayed in the right manner.

# Solution
Adjusted the css in order to make it look better.

Screenshot from 2023-09-12 19-35-23

# Action

### DIFF
--- a/frontend/css/main.less
+++ b/frontend/css/main.less
@@ -316,25 +316,18 @@ pre code.hljs {
 }
 
 // For cover art in alerts
-@alertThumbnailSize: 40px;
+@alertThumbnailSize: 60px;
 
-.alert {
+.brainzplayer-alert {
   display: flex;
   align-items: center;
   justify-content: center;
 }
 
-.alert-text {
-  font-size: 11px;
-  font-weight: bold;
-  margin-left: 9px;
-  color: #353070;
-}
-
 .alert-thumbnail {
   max-height: @alertThumbnailSize;
   margin: -@alert-padding;
-  margin-right: 0;
+  margin-right: @alert-padding;
 }
 
 .search-dropdown {
--- a/frontend/css/main.less
+++ b/frontend/css/main.less
@@ -288,8 +288,19 @@ pre code.hljs {
 }
 
 // For cover art in alerts
-@alertThumbnailSize: 60px;
-.alert .alert-thumbnail {
+@alertThumbnailSize: 40px;
+.alert{
+  display: flex;
+   align-items: center;
+  justify-content: center;
+}
+.alert-text{
+  font-size: 11px;
+  font-weight: bold;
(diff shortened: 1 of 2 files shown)


frontend/css/main.less
```diff
@@ -283,13 +310,28 @@ pre code.hljs {
...
+}
+
+.alert-text {
+  font-size: 11px;
+  font-weight: bold;
+  margin-left: 9px;
+  color: #353070;
+}
+
+.alert-thumbnail {
```
MonkeyDo, line 334: A few details here:

- we should be careful about overloading standard class names. in this case, ".alert" is also a bootstrap class with it own styles, so it would be wise to use something like `.brainzplayer-alert` to avoid confusion and style overlap
- I don't think the modifications to the text styles we necessary. I would remove all the `.alert-text` styles
- the margin between the thumbnail and the text is required however, which we can instead add to `.alert-thumbnail` as `margin-right: @alert-padding;` instead of ` margin-right: 0;`

frontend/css/main.less
```diff
@@ -283,13 +310,28 @@ pre code.hljs {
...
     }
+
     margin-right: 0.7em;
   }
 }
 
 // For cover art in alerts
-@alertThumbnailSize: 60px;
-.alert .alert-thumbnail {
+@alertThumbnailSize: 40px;
```
MonkeyDo, line 319: Was changing the size here required?

frontend/js/src/brainzplayer/BrainzPlayer.tsx
```diff
@@ -592,7 +591,7 @@ export default class BrainzPlayer extends React.Component<
         createNotification(title, artist, album, artwork?.[0]?.src);
       } else {
         const message = (
-          <>
+          <div className="alert">
```
MonkeyDo, line 594: As described above, I suggest:
```suggestion
          <div className="brainzplayer-alert">
```

MonkeyDo (changes requested):
Thanks for working on this, I do have a few questions

frontend/css/main.less
KasukabeDefenceForce: I felt the thumbnail was a bit too big for the alert box and wanted to give more space to the text instead.

frontend/js/src/brainzplayer/BrainzPlayer.tsx
KasukabeDefenceForce: Yes, I think "brainzplayer-alert" is better. I will try to avoid using confusing class names in the future.

MonkeyDo (approved):
Thank you ! :rocket:

PR #2575, opened 2023-09-13
Fix stats error message

mayhem:
The message shown to users when not stats are available wasn't great. It used to say:

  Statistics for the user have not been calculated

even for cases where data would not be calculated. Changed the message to:

  There are no statistics available for this user for this period


mayhem:
OK, I've spent too much time on this silly change. I'm abandoning this PR. If you're bugged by it, close it or fix it!

PR #2586, opened 2023-09-24
docs: list gonic as a listenbrainz compatible music server

sentriz:
not really a problem/solution/action situation as asked by the template, but instead just updating the docs to add [gonic](github.com/sentriz/gonic) as an option here 

image

https://listenbrainz.org/add-data/

thanks!


mayhem (approved):
Thank you for this PR!

PR #2588, opened 2023-09-26
LB-1347: Allow "Export to JSPF" option when user is not logged in on the LB radio page.

KasukabeDefenceForce:
# Problem
Currently on the LB radio page the entire options button is not rendered when the user is not logged in, preventing them from exporting the playlist to JSPF.

# Solution
Removed user authorization check for displaying option menu.

# Action


MonkeyDo:
> I haven't added the "onPlaylistDelete" prop to the playlistmenu component as I don't think we want to provide the users the ability to delete playlist.

I'm not sure I understand… do you mean that we don't want not-logged-in users to delete playlists, or any user? Because we certainly want users to be able to delete their own playlists… In any case the API would throw an error if you tried to delete a playlist that you do not own.

>Also, how can I test this on my local environment?

For that, follow the instructions in our documentation step by step: https://listenbrainz.readthedocs.io/en/latest/developers/devel-env.html

frontend/js/src/playlists/PlaylistMenu.tsx
```diff
@@ -59,21 +64,27 @@ function PlaylistMenu({
...
-        currentUser.auth_token,
-        getPlaylistId(playlist)
-      );
+      let newPlaylistId;
+      if (playlistID) {
+        newPlaylistId = await APIService.copyPlaylist(auth_token, playlistID);
+      } else {
+        newPlaylistId = await APIService.createPlaylist(auth_token, {
+          playlist,
+        });
```
MonkeyDo, line 73: I realized the LBRadio playlists do not have an identifier (they are temporary playlists that don't exist on the server) and trying to "Duplicate" was throwing an error, since `copyPlaylist` requires an existing playlist.

I implemented the mechanism we had in LBRadio.tsx previously, which uses `createPlaylist` instead to explicitly send a fully constructed playlist to be saved.

frontend/js/src/playlists/PlaylistMenu.tsx
```diff
@@ -133,17 +141,8 @@ function PlaylistMenu({
       { toastId: "export-playlist" }
     );
   };
-  const handlePlaylistExport = async (
-    handler: (
-      playlistId: string,
-      playlistTitle: string,
-      auth_token: string
```
MonkeyDo, line 140: I realized we don't need to pass those arguments to the handler functions, since they are defined in the main scope of the component function and we have access to those values from the get go.

frontend/js/src/playlists/PlaylistMenu.tsx
```diff
@@ -155,14 +154,11 @@ function PlaylistMenu({
       );
       return;
     }
-    setLoading(true);
```
MonkeyDo, line 158: The `loading` prop wasn't used anywhere, so I removed it.

MonkeyDo (commented):
Sorry, I was a bit rude and took over your PR, I hope you don't mind too much 😅 

After deploying the PR to test.listenbrainz.org and doing some testing, as I started looking into the details I noticed some inconsistencies and unused code that would have been a bit tricky to explain clearly, so instead I went for it directly…

frontend/js/src/playlists/PlaylistMenu.tsx
```diff
@@ -86,35 +97,32 @@ function PlaylistMenu({
...
-  ) => {
+  const exportToSpotify = async () => {
+    if (!auth_token) {
+      alertMustBeLoggedIn();
+      return;
+    }
     const result = await APIService.exportPlaylistToSpotify(
       auth_token,
-      playlistId
+      playlistID
```
MonkeyDo, line 125: We have the same issue here that I mentioned earlier: this only works if there is an existing playlist on the server.
In the case of LBRadio playlists, they are ephemeral and `playlistID` will be undefined.
Instead, I guess we would first need to create a new playlist [as we do here](https://github.com/metabrainz/listenbrainz-server/pull/2588/files#diff-9789b1b5a9afea5adb31b3f150de1651f3bead3f988a6158a4e4ff4324a81978R71-R73) and use the new playlist ID to export it.

PR #2590, opened 2023-10-03
Make musicbrainz instance url configurable

amCap1712:
This makes it simpler to test the new oauth system by simply changing a url in the config.py.


mayhem (approved):
+1

PR #2594, opened 2023-10-08
Fix API docs link 404

mwiencek:
# Problem

The current API docs link is a 404.

# Solution

Replace it with the link from https://listenbrainz.readthedocs.io/en/latest/index.html.


mayhem:
Thanks!

PR #2595, opened 2023-10-09
Art API: Add links and titles in generated stats art 

MonkeyDo:
This PR aims to add links and a title (the browser will show a popup) to the generated art SVGs we create.

I am not currently able to test this locally, so before review please take this PR with a grain of salt !


MonkeyDo:
I have added the links and titles where I could with the metadata available (for example [`create_grid_stats_cover` returns images and release metadata](https://github.com/metabrainz/listenbrainz-server/pull/2595/files#diff-dfdf3af0964194368adb47408a4ec12f55fd42f97d5733825e7ffbee93f53db3R144)), but I will need help from @mayhem to fetch the releases' metadata here:
https://github.com/metabrainz/listenbrainz-server/blob/6441bcdfe4cf6f592fb63cec8ccaa169dcd007cc/listenbrainz/webserver/views/art_api.py#L92-L98

PR #2596, opened 2023-10-12
LB-1046: "Add to playlist" option in ListenCard menu

MonkeyDo:
We are currently lacking a way to easily add tracks to playlists from anywhere on the website.
We have to go to the playlist page, open the right playlist (or create it if it doesn't exist), then search for the recording before selecting it.

Wouldn't it be really nice to have a simple option on each listen to "add to a playlist", complete with createing a new playlist on the fly?
Well, despair no more ! This new PR adds just that feature!
Here's a new modal component that fetches a user's playlists (storing them in the browser's sessionStorage to avoid re-fetching while the page is open), presents them in a list for the user to click on, and even adds a button to open the playlist creation modal with the track pre-filled !
How convenient !
image
image


MonkeyDo:
P.S: I opted for using sessionStorage instead of localStorage because the former is only relevant while the current tab is open (a new tab will start with an empty sessionStorage).
This means I don't have to set up some cache busting strategy such as storing a timestamp of the last playlist fetch and re-fetching past an arbitrary "freshness" time span.
Instead, we optimize just enough to save data but don't have to think about the caching strategy too much.

frontend/js/src/listens/AddToPlaylist.tsx
```diff
@@ -0,0 +1,200 @@
...
+                    false,
+                    false,
+                );
+                
+                if(!response){
+                    return;
+                }
+                const { playlists: existingPlaylists} = response;
+                setPlaylists(existingPlaylists);
+                window.sessionStorage.setItem("listenbrainz_playlists", JSON.stringify(existingPlaylists));
```
phw, line 46: Does this session storage cache gets purged somehow? What happens if the user e.g. adds or removes playlists after this was set and then uses the "add to playlist" again? Will they end up with stale cache?

frontend/js/src/playlists/CreateOrEditPlaylistModal.tsx
```diff
@@ -25,7 +26,7 @@ export default NiceModal.create((props: CreateOrEditPlaylistModalProps) => {
   }, [modal]);
 
   const { currentUser, APIService } = React.useContext(GlobalAppContext);
-  const { playlist } = props;
+  const { playlist,initialTracks } = props;
```
phw, line 29: Nitpicking, but I guess a space after the comma would be nice

frontend/js/src/listens/AddToPlaylist.tsx
```diff
@@ -0,0 +1,200 @@
...
+        }
+    },[listen,currentUser.auth_token]);
+
+    const createPlaylist = React.useCallback(async()=>{
+        let trackToSend;
+        if(has(listen,"title")){
+            trackToSend = listen as JSPFTrack;
+        } else {
+            trackToSend = listenToJSPFTrack(listen as Listen);
+        }
```
phw, line 126: I don't fully understand how the two cases come to be, but anyway the same code block is needed also above in line 84. Would be good to move this into a helper function.

phw (changes requested):
I'm not so much an expert in react, but overall this looks good. Just the caching of loaded playlists in session storage concerns me a bit, as I don't see how this is invalidated.

frontend/js/src/listens/AddToPlaylist.tsx
phw: Just saw your comment above. Makes some sense, but having to open a new tab still means I will end up with probably stale data after doing changes to playlists and staying in the same tab, right?

MonkeyDo: Yes, I tried it on test.LB and indeed a deleted playlist stays in the cached list and trying to add a track to it results in an error: 
image

I think this is a case of premature optimization, I'm just going to remove the caching.

For posterity, I would otherwise have to update the stored playlist after modifying a playlist through the website, which involves parsing the stored array, finding and modifying the modified playlist or adding the new one, and storing the stringified result again.
But that still wouldn't solve the (arguably edge) case of deleting a playlist through the API.

MonkeyDo:
Thanks a lot for the review @phw !

PR #2604, opened 2023-10-25
Labs api improvements

mayhem:
This new labs endpoint allows the quick lookup of tags and popularity data, which will be needed by ListenBrainz local to create local playlists. The endpoint has been working well, so time to merge it.

Also, return release_name and release_mbid from the recording lookup endpoint for use with the entity pages.


amCap1712 (approved):
LGTM, Thanks

PR #2606, opened 2023-10-26
Misc. Spark Fixes

amCap1712:
1. Update query to change empty release mbids to null to pass validation.
2. Fix troi playlists creation when no users in the timezone.
3. Capture exceptions in sentry for spark reader errors


mayhem (approved):
All good when tests pass.

PR #2607, opened 2023-10-26
Docs: Clarify that listened_at should be set to start of playback

phw:
# Problem

The API docs at https://listenbrainz.readthedocs.io/en/latest/users/json.html are unclear about to which time the value of `listened_at` should be set to. Different tools use different times. Common is either the time when playback started, but in some cases tools set the time when the listen was marked for submission (so usually after half the track or 4 minutes of listening).

In the discussion at https://community.metabrainz.org/t/listened-at-parameter-for-submissions-specification-unclear/659142 it was explained, that playback start time is preferred.

# Solution

Update the API documentation to make clear `listened_at` *should* be time of playback start. This is intentionally formulated as should and not must. For technical reasons clients might not be able to submit the playback start time and instead will have to use the time when the listen gets marked for submission.

### DIFF
--- a/docs/users/json.rst
+++ b/docs/users/json.rst
@@ -156,7 +156,8 @@ A minimal payload must include
 ``artist_name`` and ``track_name`` elements must be simple strings.
 
 The payload should also include the ``listened_at`` element, which must be an integer
-representing the Unix time when the track was listened to. The minimum accepted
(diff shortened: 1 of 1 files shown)


mayhem (approved):
I love how the PR text is much larger than the change itself. <3

PR #2610, opened 2023-10-30
Update troi to the latest release

mayhem:
Update troi to the v-2023-10-30.0 release to include album data into listen submission from LB Radio tracks.


MonkeyDo (approved):
:rocket:

PR #2613, opened 2023-10-31
Fix user similarity

mayhem:
Right now user similarity leaves a lot to be desired:

   https://community.metabrainz.org/t/how-is-user-similarity-score-calculated/662453

That thread made me realize that we're being stupid about scaling the user similarity and that it actually already works fine as percentage, we just need to scale it a bit. More trickly, it changes the format of the similar user data -- we will need to consider this carefully when we merge this. Perhaps keeping it backwards compatible for now might be good; thoughts? @amCap1712 

THIS CODE HAS NOT BEEN TESTED. DO NOT MERGE.

### DIFF
--- a/Dockerfile.spark
+++ b/Dockerfile.spark
@@ -1,5 +1,10 @@
 FROM metabrainz/listenbrainz-spark-base:latest
 
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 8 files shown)


mayhem:
Is there a way to run tests on a draft PR?

mayhem:
@amCap1712 any idea why the spark tests keep failing?

PR #2615, opened 2023-10-31
Improve user similarity

mayhem:
This PR also calculates a simple percentage value for user similarity that seems to work better. So far, we've keeping the old value in the DB, in case this change doesn't improve anything.

This PR recognizes the fact that the cos() of the coefficient is an angle, not a rating. An angle can be simply converted to a percentage to make the use similarity a lot more intuitive.


amCap1712 (approved):
LGTM, Thanks!

PR #2620, opened 2023-11-05
LB-1392: Add missing tables to dumps and ensure that all tables are dumped

mayhem:
Fixing patchy data backups. :(

### DIFF
--- a/listenbrainz/db/dump.py
+++ b/listenbrainz/db/dump.py
@@ -128,7 +128,24 @@ PUBLIC_TABLES_TIMESCALE_DUMP = {
         'recording_mbid',
         'match_type',
         'last_updated'
-    )
+    ),
+    'mbid_manual_mapping': (
+        'id',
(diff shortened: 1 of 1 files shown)


listenbrainz/db/dump.py
```diff
@@ -128,7 +128,24 @@
         'recording_mbid',
         'match_type',
         'last_updated'
```
amCap1712, line 130: The mbid_mapping table is missing a check_again field, can you please also add it here

amCap1712:
Change the PR title so that the ticket is not closed. We still need to create dumps for metadata caches and some mechanism to ensure we don't forget to add new tables to dumps.

amCap1712 (approved):
LGTM, Thanks!

listenbrainz/db/dump.py
mayhem: Done.

PR #2621, opened 2023-11-07
LB-1300: Fix 'following' field in get_following_docstring

azr-arch:
# Problem

The JSON example in the API documentation for the `get_following` function was incorrect. It used the term 'followers' instead of 'following', which could lead to confusion for users trying to understand the API.

# Solution

I corrected the JSON example in the docstring of the `get_following` function to use the correct term 'following'. This should now accurately represent the structure of the returned JSON object and make it clearer for users.

# Action

Please review the changes made to the docstring of the `get_following` function. No other action is required.

### DIFF
--- a/listenbrainz/webserver/views/social_api.py
+++ b/listenbrainz/webserver/views/social_api.py
@@ -53,7 +53,7 @@ def get_following(user_name: str):
     .. code-block:: json
 
         {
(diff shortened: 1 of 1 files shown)


amCap1712 (approved):
Thanks for the contribution! LGTM.

PR #2622, opened 2023-11-08
LB-1389: Ignore database connect strings when running tests

mayhem:
This PR changes our tests suite to IGNORE the database connect strings provided by the user and instead use default strings for the locally created database. This fully isolates tests to the local containers created for testing purposes and prevents an accidental connection to some other DB.  For testing purposes I put invalid connect strings into my config.h and ran the tests and a test DB was created and tests successfully ran.

IMPORTANT: @amCap1712  This PR needs to have a critical review to ensure that I didn't miss any other ways in which the testing code can connect to the DB via the connect strings.


listenbrainz/db/__init__.py
```diff
@@ -63,3 +63,12 @@ def run_sql_script_without_transaction(sql_file_path):
             connection.close()
         return True
 
+
+def create_test_database_connect_strings():
+    return ("postgresql://listenbrainz:listenbrainz@lb_db:5432/listenbrainz", "postgresql://postgres:postgres@lb_db/postgres",
```
MonkeyDo, line 68: FWIW, I still think we should use a separate database name if at all possible, i.e.:
```suggestion
    return ("postgresql://listenbrainz:listenbrainz@lb_db:5432/listenbrainz_test", "postgresql://postgres:postgres@lb_db/postgres",
```

mayhem: Oh, good idea. Let me try that.

mayhem: Implemented

admin/timescale/create_extensions.sql
```diff
@@ -1 +1,2 @@
+CREATE EXTENSION IF NOT EXISTS timescaledb CASCADE;
```
mayhem, line 1: I have no idea why this is suddenly required, but the create_hypertable function was failing without it.  :shrug:

listenbrainz/webserver/__init__.py
```diff
@@ -94,8 +95,17 @@ def create_app(debug=None):
     # Database connections
     from listenbrainz import db
     from listenbrainz.db import timescale as ts
-    db.init_db_connection(app.config['SQLALCHEMY_DATABASE_URI'])
-    ts.init_db_connection(app.config['SQLALCHEMY_TIMESCALE_URI'])
+
+    # If we're running tests, overwrite the given DB configuration from the config and disregard the
+    # configuration, since that configuration could possibly point a different (production) DB.
+    if app.config['TESTING']:
```
amCap1712, line 101: some tests set TESTING=False to test for certain conditions that are disabled when TESTING is true.

amCap1712: need to think how to reconcile the two.

mayhem: FWIW, I've tested with invalid connection strings in my config.py and the tests ran just fine, so there was no attempting to connect to a resource defined by the strings.

amCap1712: Yes, those tests currently don't access the DB.

amCap1712 (commented):
I currently cannot think of any other ways the connect string could be modified without actually modifiying the python code.

LGTM, Thanks!!!

amCap1712 (approved):

PR #2623, opened 2023-11-09
LB-1382: Place private dumps into wholly seperate file system paths

amCap1712:
To avoid accidental leaking of private data, make the separation between public and private dumps more explicit. Often such errors occur when running scripts by hand, so adding a few more safeguards to the dumps running code too.

### DIFF
--- a/listenbrainz/db/dump_manager.py
+++ b/listenbrainz/db/dump_manager.py
@@ -1,6 +1,7 @@
 """ This module contains a click group with commands to
 create and import postgres data dumps.
 """
+from pathlib import PurePath
 
 # listenbrainz-server - Server for the ListenBrainz project
 #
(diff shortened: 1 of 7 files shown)


listenbrainz/db/dump_manager.py
```diff
@@ -143,6 +145,15 @@ def create_full(location, threads, dump_id, do_listen_dump: bool, do_spark_dump:
     """
     app = create_app()
     with app.app_context():
+        if not location_private and (do_db_dump or do_timescale_dump):
+            current_app.logger.error("No location specified for creating private database and timescale dumps")
+            sys.exit(-1)
+        if location_private and os.path.normpath(location_private) == os.path.normpath(location):
```
mayhem, line 151: Great checks!

mayhem (approved):
Thank you!

PR #2625, opened 2023-11-12
LB-1357: Add support for importing local listen dumps into Spark

Maxr1998:
See LB-1357.

### Motivation

Currently, there are two Spark management commands to download either a full or an incremental listen dump from the Metabrainz FTP server and import it into spark.

At the same time, the dump manager within the ListenBrainz server supports creating dumps through the `create_full` or `create_incremental` commands. These dumps are stored inside a `listenbrainz-export` folder that's also mounted into the spark container(s) at `/rec/listenbrainz-export`.
If you want to import those dumps into Spark, e.g. for working on stats computation with your own, known data, you'd have to make them available through an FTP server and let the importer download them first, even if they are available locally.

### Solution

An additional `--use-local` flag on the import Spark commands searches for dumps in the export directory, picks the latest (or specified through ID) dump and imports it without having to copy the archive first.

This was implemented by further generalizing the `ListenBrainzFTPDownloader` to provide functions to list files and pick a specific dump archive. The FTP subclass downloads the archive to a temporary folder and returns its path (like now), the local version directly returns the archive path.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -23,3 +23,6 @@
 
 # Javascript packages
 /node_modules/
+
+# ListenBrainz local dump directory
+/listenbrainz-export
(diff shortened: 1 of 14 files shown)


amCap1712 (approved):
Thank you for the PR. LGTM! 

I tested it locally and it works great. I made a minor change to handle an edge case in creating local dumps but otherwise works great!

PR #2626, opened 2023-11-13
Fix art creator copy image functionality

MonkeyDo:
Last time (in commit daef1b7) I tried to make it more robust (considering the range of browser implementations and behaviors) I broke the feature that allows copying images in the stats art creator. I pooped the deck and broke the feature thoroughly.

Had another closer look at the specs and variations of implementation and fixed the feature.
For some context, you can read https://wolfgangrittner.dev/how-to-use-clipboard-api-in-firefox/

There are still some improvements that can be made, in particular to detect whether we have permission from the browser to use the clipboard API


MonkeyDo:
Currently deployed on https://test.listenbrainz.org/explore/art-creator/

anshg1214 (approved):
LGTM! I tested it out on test.LB and it works fine.

PR #2627, opened 2023-11-15
Deduplicate MSIDs

amCap1712:
Duplicate MSIDs exist because:

1. we made the conditions for a new MSID to be created more stringent.
2. during the recent LB database incident, the table was lost and had to be recovered from other sources in the process of which duplicates were created.

To fixup the issue:

1. Run the 2023-11-11-msb-deduplicate-1.sql script. This will separate the MSIDs into two tables, unique submissions and redirects. The redirects table will be used to fixup MSIDs in all of ListenBrainz one by one.

2. Stop timescale writer.

3. Run the 2023-11-11-msb-deduplicate-2.sql script. This will fixup the MSIDs that have been inserted since the first script ran. Since new duplicates cannot be created, we can use a simple insert. Then the script renames and switches the tables atomically.

4. Restart timescale writer.

### DIFF
--- a/admin/timescale/updates/2023-11-11-msb-deduplicate-1.sql
+++ b/admin/timescale/updates/2023-11-11-msb-deduplicate-1.sql
@@ -48,7 +48,7 @@ with gather_data as (
          from gather_data ctnt
  join lateral unnest(msids[2:]) AS duplicate_msid
            on true
-        where array_length(msids) > 1
+        where array_length(msids, 1) > 1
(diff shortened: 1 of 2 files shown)


mayhem:
Were the recovered rows for messybrainz ids stored in messybrainz.submissions ?

amCap1712:
yes

mayhem (approved):
Thanks for cleaning up!

PR #2628, opened 2023-11-15
Deduplicate listens

amCap1712:
Listens were backed up from dumps before MSIDs since MSIDs are how we distinguish unique listens, a lot of duplicates were created from users importing their listens from last.fm etc. To fix the listens, first run the fix MSIDs scripts then the listens.


mayhem (approved):
LGTM. I wonder how many hours this will need to run. :/

amCap1712:
expecting 2 days almost.

PR #2630, opened 2023-11-15
API: Add oldest listen timestamp to /user/<username>/listens API response

MonkeyDo:
@phw [requested in IRC](https://chatlogs.metabrainz.org/libera/metabrainz/2023-11-15/?msg=5249710&page=1) that we add a way to retrieve the oldest listen timestamp using the API.

Currently, the `/user/<username>/listens` API endpoint returns `latest_listen_ts` and we simply discard the information about the oldest timestamp. I suppose we never thought it was useful information to have in that response?

I've modified the code to use the discarded timestamp and return it as part of the API response as `oldest_listen_ts` (as we do in other places in the code, although none publicly-facing).
Are we happy with the name?

If there are other considerations that I missed please let me know.


mayhem (approved):
hey, look at you being a python coder!

PR #2635, opened 2023-11-19
LB-1398: Show dates in ISO 8601 format

julian45:
# Problem

Not necessarily a problem, but this would implement [LB-1398](https://tickets.metabrainz.org/browse/LB-1398), which would convert date formats in pickers to the ISO 8601 standard format.

# Solution

Since these pickers only appear in two places in the codebase and appear to use the same library, this commit moves the date components around and changes the separators.

# Action

Any feedback on this PR is greatly appreciated. Thank you for your time!


julian45:
The only test that seems to fail is one for an unrelated modal (for entering CritiqueBrainz reviews). I'm not sure how my slight changes to frontend date representation (one which doesn't even change how they're stored on the backend, i.e., Unix time) could have affected this. Is it possible to rerun the test and see if it produces the same result?

julian45:
The [most recent test](https://github.com/metabrainz/listenbrainz-server/actions/runs/6935551546/job/18865903934?pr=2636) in #2636 failed the exact same frontend test, passing all others. Same for [this test](https://github.com/metabrainz/listenbrainz-server/actions/runs/6907759542/job/18795592112?pr=2633) on #2633. Given the differences between all of our PRs, this seems suggestive of either a subtle bug in other components being hit upon each time, or of an issue with the test itself.

MonkeyDo (approved):
Thank you @julian45 !

The failing test is indeed unrelated to your changes, and we are aware of the issue of this unstable test (in fact I am working on fixing it as we speak).
I'l try rerunning the tests; if the issue persists I will merge with that failing test.

MonkeyDo:
Finally fixed the tests in #2641 , thanks for your patience !

PR #2636, opened 2023-11-20
Fix websockets setup for development

amCap1712:
Websockets don't work in development because the code expected it to run on the same host and port as the LB webserver. To configure the url for websockets in development and production properly, add a separate configuration variable.

**TODO: Test on test.lb before merging.**

### DIFF
--- a/frontend/js/src/playlists/Playlist.tsx
+++ b/frontend/js/src/playlists/Playlist.tsx
@@ -108,7 +108,10 @@ export default class PlaylistPage extends React.Component<
   };
 
   createWebsocketsConnection = (): void => {
-    this.socket = io(`${window.location.origin}`, { path: "/socket.io/" });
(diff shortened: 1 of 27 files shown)


frontend/js/src/playlists/Playlist.tsx
```diff
@@ -108,7 +108,10 @@ export default class PlaylistPage extends React.Component<
   };
 
   createWebsocketsConnection = (): void => {
-    this.socket = io(`${window.location.origin}`, { path: "/socket.io/" });
+    const { websocketsUrl } = this.context;
+    this.socket = io(`${websocketsUrl || window.location.origin}`, {
```
MonkeyDo, line 112: I suppose at some point we had a reason to use a template literal, but now we can simply:
```suggestion
    this.socket = io(websocketsUrl || window.location.origin, {
```

PR #2639, opened 2023-11-23
Add Listening queues to BrainzPlayer

anshg1214:
This PR adds Listening Queues to BrainzPlayer


frontend/js/src/brainzplayer/Queue.tsx
```diff
@@ -0,0 +1,121 @@
...
+            key="queue-listening-now-card"
+            className="queue-item-card"
+            listen={currentListen as Listen}
+            showTimestamp={false}
+            showUsername={false}
+          />
+        </>
+      )}
+      <div className="queue-headers">
+        <p>Next Up:</p>
```
MonkeyDo, line 77: Same as above:
```suggestion
        <h4 className="queue-headers">Next Up:</h4>
```
And I moved the queue buttons at the bottom. I realize now that might be an issue if  the queue is long, the button won't be visible without scrolling down.

Alternatively, maybe something like this?
image

frontend/js/src/brainzplayer/Queue.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <div className="queue-buttons">
+          <button
+            className="btn btn-info"
```
MonkeyDo, line 80: Buttons can probably be a bit smaller:
```suggestion
            className="btn btn-info btn-sm"
```

frontend/js/src/brainzplayer/Queue.tsx
```diff
@@ -0,0 +1,121 @@
...
+            data-toggle="modal"
+            data-target="#CreateOrEditPlaylistModal"
+            onClick={addQueueToPlaylist}
+            type="button"
+          >
+            Save to Playlist
```
MonkeyDo, line 86: Maybe a save icon?

frontend/js/src/listens/ListenCard.tsx
```diff
@@ -117,9 +117,12 @@ export default class ListenCard extends React.Component<
...
   ) {
-    const { listen: oldListen } = oldState;
-    const { customThumbnail } = this.props;
-    const { listen } = this.state;
+    const { listen: oldListen } = oldProps;
+    const { customThumbnail, listen } = this.props;
+
+    if (Boolean(listen) && !isEqual(listen, oldListen)) {
+      this.setState({ listen });
+    }
```
MonkeyDo, line 125: Why do we need to save the listen in the state?
If it's to access it below in the add to queue functions, you can use the props directly

frontend/js/src/playlists/utils.tsx
```diff
@@ -130,3 +131,19 @@ export function listenToJSPFTrack(listen: Listen): JSPFTrack {
...
+): BrainzPlayerQueueItem {
+  let listenTrack: Listen;
+  if (has(track, "title")) {
+    listenTrack = JSPFTrackToListen(track as JSPFTrack);
+  } else {
+    listenTrack = cloneDeep(track as BrainzPlayerQueueItem);
+  }
+  const queueItem = {
+    ...listenTrack,
+    id: `${listenTrack.track_metadata.track_name} - ${Date.now().toString()}`,
```
MonkeyDo, line 146: If this util is mainly to add an `id` , I wonder if instead we can directly pass an `id` prop to the QueueItemCard. Do you reckon that would work?

MonkeyDo (commented):
Great work ! It's working quite well already, and the saved queue in localstorage is a nifty feature.

My main comment is about functionality.
Currently, if you start playing tracks on a page, it will play all the listens on that page one after the other.
The new queue removes that crucial feature: after playing the first track on a page it just stops as the queue is empty.
What should happen is that if I start playing a track, all the listens under that ListenCard on that page should be added to the queue (or a temp queue, i'll explain below).

When I navigate to another page, the queue should have the listens from the new page added. That being said we don't want to override items in the queue that the user added themselves, so we might need the concept of a temporary queue (or AmbientQueue? the listens that are on the current page) that lives alongside the user's queue (UserQueue?) as a fallback so that when  the UserQueue is finished we continue with the AmbientQueue.

Think of it like on Spotify, when you finish playing an album of playlist, playback continues by itself with spotify's suggestions. In this case instead of our suggestions, we just want to continue playback with the listens on the page.
Does that make sense?

Happy to discuss all this with you in a call if you want; two queues living side-by-side will inevitably create some complications :)
For example we don't want to save the ambient queue to localstorage I suppose.
And another pickle, maybe we keep that one for later: users should be able to move an item from the ambient queue to the user queue.

frontend/js/src/listens/ListenCard.tsx
anshg1214: The `Link with Musicbrainz` feature updates listen in the state with the new linked track data, so we cannot directly use props everywhere. https://github.com/anshg1214/listenbrainz-server/blob/listening-queues/frontend/js/src/listens/ListenCard.tsx#L413-L425

frontend/js/src/brainzplayer/Queue.tsx
anshg1214: Looks good

frontend/js/src/listens/ListenCard.tsx
anshg1214: The `Link with Musicbrainz` feature updates the listen state on successfully linking. So we cannot directly use props everywhere. https://github.com/anshg1214/listenbrainz-server/blob/listening-queues/frontend/js/src/listens/ListenCard.tsx#L413-L425

MonkeyDo (commented):
I've tested the feature again, thanks for the improvements; definitely working and looking better !
I've got some more feedback after thorough testing; I know we have a date to talk about it tomorrow but I thought I'd put this here so I don't forget everything.
Feel free to ignore for now !

While the last addition to play tracks from props if there is no next song works, I still think we need a separate temporary queue (or “ambient queue”):
- The temp queue is ephemeral and should not be saved in localstorage.
- Whenever a page loads or the user navigates to another page, the temp queue should be emptied and all the listens on the new page added instead. (this is a departure from the current “if the queue is empty, load more listens)
- How to add the listens is another story, but instead of using props I think we should use a browser message similar to "add-track-to-queue”, accepting an array of temp listens. This would allow us to decouple pages from BrainzPlayer (i.e. we don’t need to render a BP component with each page, we just render it once on a general layout component instead.
- If the user’s current queue is finished, start playing from the temp queue.
  - Meanwhile if user adds a track to the queue, place it in the user queue and at the end of current song resume playing until the user’s queue is empty again.
- We might want to add another (smaller) header (for example “on this page”) to separate the user's queue from the temp queue.


Queue clearing:
I’m not sure the clear queue mechanism is working as expected. When I clear the queue and start playing a track, the queue is magically re-populated with some tracks that were in the previous queue
Similarly when I play a single track and click “next”, tracks from the page are added to the queue with the old "now playing” listen replacing the current one, probably because “clear queue” does not remove the old current listen from the queue prop
If playback is not currently active, perhaps clear queue should also remove the queue’s “now playing” track.


Repeat mode icon:
The “ban” icon is probably not ideal for the playback controls.
Spotify uses the repeat icon for off, then repeat icon with active styling (green color and a dot underneath), and the same with the repeat icon + number one for single loop.
Probably a good idea in that case to also add that ‘active’ indicator on the queue button when the queue is open:
image
Not sure we want green specifically, but something along those lines would work better IMO.

frontend/css/brainzplayer.less
```diff
@@ -338,3 +338,58 @@
     }
   }
 }
+
+.queue {
+  position: fixed;
+  padding: 1em;
+  bottom: 60px;
```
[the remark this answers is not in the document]
anshg1214: The animation looks quite smooth now. Instead of adding it in Queues.tsx, I've made the changes in BrainzPlayerUI.tsx since `Queues` is a memorised function, so modifying the state `showQueue` is causing the component to re-render on props change

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -193,8 +235,82 @@ export default class BrainzPlayer extends React.Component<
...
+          });
+        }
+      } catch (e) {
+        // Do nothing, we just fallback gracefully to the default queue.
+        const { listens } = this.props;
+        this.replaceQueue(listens);
+      }
+    } else {
+      const { listens } = this.props;
+      this.replaceQueue(listens);
```
MonkeyDo, line 274: I don't think we want these fallback modes anymore, now that we have the ambient queue to replace it.

Currently this results in a duplicated list of listens in the queue and in the ambientQueue.

frontend/js/src/common/brainzplayer/Queue.tsx
```diff
@@ -0,0 +1,148 @@
...
+            })}
+          </ReactSortable>
+        ) : (
+          <div className="lead text-center">
+            <p>Nothing in this queue yet</p>
+          </div>
+        )}
+      </div>
+      <div className="queue-headers">
+        <h4>Suggested for you:</h4>
```
MonkeyDo, line 128: I think the wording of this could be workshopped, we do'nt want users to think that these are recommendations we generate on the spot.
Maybe "On this page:" or something along those lines?
I don't have a better idea right now.
image


Another option could be to remove the header (but keep the divider, with some padding) and show the ambientQueue with a lower opacity like so:
image

What do you think?

anshg1214: `On this page` looks good to me

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
anshg1214: Fixed

MonkeyDo (commented):
I deployed again to test.LB, and here is my feedback:

After just opening the page, the queue has a "now playing" item at the top that is from days ago. I reckon this is saved in localstorage?
I'm not sure I would want the currently playing song to be saved, probably just the "next up" queue. What do you think?

The "on this page" queue looks and acts as we discussed, it's quite reactive and I think it's pretty easy to understand what's going on :+1:

There is something weird happening with the saved queues.
I tried playing some tracks from the page, but I get weird interference that is hard to describe. I think a video chat might be helpful to nail the final details.

I'll try to describe generally: when I first open the page or if there is a playback error, whatever I successfully played before (which I would consider should be forgotten by now) starts playing (i.e. added as "playing now" at the top of the queue) and some items are added to "next up" automatically.

If I clear the queue, in some cases I basically get stuck in a loop. Clicking next just plays the same song over and over. There is nothing in "next up", expected listens in "on this page", and repeat=off so i don't know what's happening.

I guess in short we probably have to rewrite some of the logic of `playNext`  and move it to the Queue component (which makes sense) so we can do Queue.getNextTrack and isolate the logic there.

anshg1214:
Closing this PR in favor of #2827

PR #2646, opened 2023-11-29
Use unique testing env var try 2

mayhem:
The previous fix for preventing tests from being run in production wasn't fool-proof either, since the user can control the TESTING variable. The test now create a different env var and check for its presence and then create a empty test DB from the strings generated from the app, ignoring settings in config.py.

This PR includes fixes from https://github.com/metabrainz/listenbrainz-server/pull/2637, so we can close that PR now.

### DIFF
--- a/listenbrainz/webserver/testing.py
+++ b/listenbrainz/webserver/testing.py
@@ -239,5 +239,5 @@ class APICompatServerTestCase(ServerTestCase):
     @classmethod
     def create_app(cls):
         app = create_api_compat_app()
-        app.config['TESTING'] = True
+        app.config['PYTHON_TESTS_RUNNING'] = True
         return app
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/testing.py
```diff
@@ -22,7 +22,7 @@ class ServerTestCase(unittest.TestCase):
     @classmethod
     def create_app(cls):
         app = create_web_app(debug=False)
-        app.config['TESTING'] = True
```
amCap1712, line 25: No need to make any change here I think, all of the code except the one for setting up databases in manage.py and for setting connect strings in init_app should not touch PYTHON_TESTS_RUNNING and just use TESTING.

amCap1712 (approved):

PR #2648, opened 2023-12-04
LB-1372: Cannot link a release from the stats page

amCap1712:
Make the requirements for opening the MBIDMappingModal on clicking the cover art image same as that of showing the 'Link to MusicBrainz' button in listen dropdown menu. Album stats are based on release groups now so also add a case for opening the release group in MusicBrainz.


MonkeyDo (approved):
That looks good, reading the code, but I haven't tried actually running the PR. None of the modifications break anything and I only see improvements

PR #2650, opened 2023-12-04
LB-1329 (I): Cover art mouseovers don't display in some pages/elements

amCap1712:
Improve the logic for choosing the thumbnail link and title for mouseovers in certain cases. This fixes the two top albums pages.

For playlist pages, we forgot to send the release name/album field in the playlist data. Add that.


MonkeyDo (approved):
Yep, that'll do it !

MonkeyDo:
I'll not that this is set to be merged into the `fix-links` branch, just in case that wasn't on purpose

amCap1712:
Yup that was only to ease PR review.

PR #2651, opened 2023-12-04
LB-1371: "Playing now" does not show cover art on page load

amCap1712:
When the mbid_mapping info in the listen changes, we want the state of the listen to be updated with the new props. The easiest way to make this happen is add all the properties that matter to key on ListenCard. There are over 40 usages of
ListenCard in the project, each with different properties in the key. At some point, we will likely want to unify all of these (or maybe not, unsure). I have created a utility function to use for keys elsewhere as well if needed.

ListenCard does not deepcopy the listen passed to it in props, therefore modifying the playing now listen object elsewhere also changed the object stored inside ListenCard's state even before React can propagate updates. therefore, deep copy
the object.


MonkeyDo (approved):
That's one hefty key! That being said, I'm certain this is still much faster than hashing the whole listen and using the hash as the key
👍 
Thanks for making it easily reusable !

PR #2652, opened 2023-12-04
LB-1315: Attempting to follow a user while not logged in gives error

amCap1712:
Hide follow button when user is not logged in. Also, saw that UserPageHeading component is no longer used so removing it and associated tests.


MonkeyDo (approved):
Yep, makes sense. I originally thought the existing `isCurrentUsersPage` would be sufficient, but the way we use it by inverting the value does create that bug, I totally oversaw that. Thanks !

PR #2653, opened 2023-12-04
LB-1352: Weekly Jams makes playlists of 90% one artist

amCap1712:
There was an error in the existing periodic jams code, we were ordering by score first which would mean choosing the topmost recommendations more frequently. troi doesn't do this step, so remove it.

Also add artist limiting step to have at most 2 recordings per artist. note that it is possible to have more than 2 recordings of artists it appears in combinations with other artists on recordings.


listenbrainz_spark/postgres/recording.py
```diff
@@ -1,13 +1,29 @@
-from listenbrainz_spark.path import RECORDING_LENGTH_DATAFRAME
+from listenbrainz_spark.path import RECORDING_LENGTH_DATAFRAME, RECORDING_ARTIST_DATAFRAME
 from listenbrainz_spark.postgres.utils import save_pg_table_to_hdfs
 
 
 def create_recording_length_cache():
-    """ Import artist country from postgres to HDFS for use in artist map stats calculation. """
+    """ Import recording lengths from postgres to HDFS for use in artist map stats calculation. """
```
mayhem, line 6: this comment isn't quite right is it?

mayhem: And one below too?

mayhem (approved):
I'm so glad this had a simple solution! Well done!

listenbrainz_spark/postgres/recording.py
amCap1712: fixed

PR #2657, opened 2023-12-06
Fix test_fetch_top_similar_users

MonkeyDo:
Following the recent PR #2615 modifying the user similarity, we went from storing in the similarity table two similarity values to a single one. This test was not updated to reflect that and is currently failing in master.


amCap1712 (approved):
LGTM, Thanks!

mayhem (approved):

PR #2658, opened 2023-12-07
LB-1381: replace "Top Similar Users" image

Aerozol:
Adding an 'actual image' for the Top Similar Users Explore feature, as per user feedback:
"I think this should be swapped out for an actual image rather than whatever... this thing is."


mayhem (approved):
The end of an era, there is no way I would approve this!

PR #2659, opened 2023-12-07
Fix fresh releases page

amCap1712:
Currently, the fresh releases page errors with: `RangeError: date value is not finite in DateTimeFormat.formatToParts()` . This was happening because formatReleaseDate was being called on an already formatted date. First call is at:

 https://github.com/metabrainz/listenbrainz-server/blob/206b1528cd6761cdab564f3ac69ad98f59b4b25e/frontend/js/src/explore/fresh-releases/ReleaseCardsGrid.tsx#L18

Removed the second call to fix.


MonkeyDo (changes requested):
I think that entire if... block can be removed actually, considering we already return the key unchanged by defaut a few lines lower.

Thanks for fixing that bug!

amCap1712:
Sorry, i just missed your comment by a second, fixing it now.

PR #2663, opened 2023-12-08
Fresh Fresh Releases explore image

Aerozol:
A fresher image to accompany the Fresh releases update (can be temporary, or permanent)


mayhem (approved):
I like it

PR #2664, opened 2023-12-08
Improve handling of scarce stats in cover art api

amCap1712:
Some of the cover art templates require 5 or 10 artists/albums/images to render the full svg template. Currently, when stats are present but less than the minimum required the number the server 500s because the code assumes enough data will always be available.

To fix this, use two measures:

1. For the templates using cover art images repeat existing urls until the minimum required number is reached.
2. For the templates using text only display items as much as there are in the data.

A small unrelated fix for lastfm loved tracks import and a formatting fix for frontend too.


amCap1712:
deployed on beta for testing.

mayhem (approved):
Glad this is fixed.

PR #2665, opened 2023-12-08
Fetch Top Release Groups for Artist

anshg1214:
This PR adds an endpoint to fetch top release groups for an artist. Also it now also includes `release_colors` in the response.

### DIFF
--- a/listenbrainz/db/color.py
+++ b/listenbrainz/db/color.py
@@ -3,7 +3,7 @@ import psycopg2
 from listenbrainz import db
 from brainzutils import musicbrainz_db as mb_db
 from listenbrainz.db.model.color import ColorResult, ColorCube
-from typing import List
+from typing import List, Dict
 from psycopg2.extensions import adapt, AsIs, register_adapter
 
 # This determines how many releases for a given color are fetched and then randomly chosen from.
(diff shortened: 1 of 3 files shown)


listenbrainz/db/color.py
```diff
@@ -127,3 +127,35 @@ def get_releases_for_color(red: int, green: int, blue: int, count: int) -> List[
                     results[i].rec_metadata = recordings
 
         return results
+
+
+def fetch_color_for_releases(release_mbids: List[str]) -> Dict[str, Dict[str, int]]:
```
amCap1712, line 132: You can accept the cursor here as method argument instead of creating a new connection.

listenbrainz/db/release.py
```diff
@@ -0,0 +1,101 @@
...
+    """ Given a list of mbids return a map with mbid as key and the release info as value.
+
+
+    TODO: Ideally, we should always resolve redirects but that makes testing some LB features difficult
+        because we don't have a MB db in the test environment yet. Once, we do we can merge the two versions.
+    """
+    if not mbids:
+        return {}
+
+    query = """
```
amCap1712, line 17: We are coming up with an artist cache for the entity pages which is much more efficient and complete in #2667. I think we can just use that instead of creating this new endpoint.

anshg1214: Since we're fetching the release data as well, so we don't have the required data in the artist cache. So we'd need to fetch the data from mb_metadata_cache

amCap1712: We have release group data in artist cache, we should show those instead of releases unless absolutely necessary.

amCap1712: Example: https://test-api.listenbrainz.org/1/metadata/artist/?artist_mbids=20244d07-534f-4eff-b4d4-930878889970&inc=release_group

anshg1214: Makes sense, it should work for our needs

anshg1214: Instead can we use the mb_release_group_cache?

amCap1712: If you have an artist mbid and need the artists' release groups, you need to use the artist cache. If you have release group mbids (if you have release mbids even those could be resolved to release groups), you can use the release group cache.

listenbrainz/db/color.py
anshg1214: I don't think its needed here, because we're not using the LB connection in places where this function is being called

amCap1712 (approved):
LGTM, Thanks!

PR #2669, opened 2023-12-13
 Use UNLOGGED tables when creating mapping tables in MB

amCap1712:
Creating mapping tables takes a lot of time and writes a lot of data which in turn creates a lot of WAL. These tables are needed only in MB db temporarily, hence its fine to change these tables to UNLOGGED when using MB connection.

Note that unlogged tables are not replicated, therefore all queries must be run against the primary database.


mayhem (approved):
I dying to see how this improves things!

PR #2671, opened 2023-12-13
LB-1402:While adding a track manually, search menu doesnt close even after selecting some item

pranavkonidena:
# Problem

While opening the add track manually option , even after changing the selected track , it isn't reflected in the card that is rendered afterwards. So i fixed that. Initially a PR was made for the same and reviewed (#2662 ).  Due to git issues, had to open a new PR. 

# Solution

So , now as @MonkeyDo  suggested in #2662 , instead of closing the menu, now the selected song changes when the user changes it in the search menu. I debugged in the add user modal screen but realized that the fault was in the listencard.jsx file. The updated listen data was not being set to the state using set state in the did component update fn. I rectified that.

# Action

I would love a quick review if possible so that I can keep learning and contributing to different issues here.

# Demo of Change

https://github.com/metabrainz/listenbrainz-server/assets/122373207/ab086e8d-0820-495c-92af-09dfb994ac0e


frontend/js/src/listens/ListenCard.tsx
```diff
@@ -118,8 +117,11 @@ export default class ListenCard extends React.Component<
     oldState: ListenCardState
   ) {
     const { listen: oldListen } = oldState;
-    const { customThumbnail } = this.props;
-    const { listen } = this.state;
+    const { listen, customThumbnail } = this.props;
+    if (!isEqual(listen, oldListen)) {
+      this.setState({ listen });
+    }
+    // const { listen } = this.state;
```
MonkeyDo, line 124: Commented out lines should be removed

frontend/js/src/listens/ListenCard.tsx
```diff
@@ -118,8 +117,11 @@ export default class ListenCard extends React.Component<
     oldState: ListenCardState
   ) {
     const { listen: oldListen } = oldState;
-    const { customThumbnail } = this.props;
-    const { listen } = this.state;
+    const { listen, customThumbnail } = this.props;
+    if (!isEqual(listen, oldListen)) {
+      this.setState({ listen });
+    }
```
MonkeyDo, line 123: Looking at this PR that implements a similar solution, we should probably check that the listen is not falsy just in case:
https://github.com/metabrainz/listenbrainz-server/pull/2639/files#diff-58b7d1f4892674a0d59758a5f774056eb2a32909cfcbf6d7444ee47f17d65e36R123-R125

frontend/js/src/listens/ListenCard.tsx
```diff
@@ -98,10 +98,9 @@ export default class ListenCard extends React.Component<
   static coverartPlaceholder = "/static/img/cover-art-placeholder.jpg";
   static contextType = GlobalAppContext;
   declare context: React.ContextType<typeof GlobalAppContext>;
-
   constructor(props: ListenCardProps) {
     super(props);
-
+    console.log("STATE CHANGEF");
```
MonkeyDo, line 103: Console calls should be removed after debugging :)

MonkeyDo (changes requested):
Thank you for opening a PR!

left a couple of comments; it turns out someone in the team has also supplied a solution, but their PR might take longer to merge than this one.
After the following changes I'm happy to merge this fix.

MonkeyDo:
I just merged today some folder refactoring PR, so there are now merge conflicts in this PR.
Let me know if you need any assistance resolving them.

MonkeyDo (approved):
Thank you !

PR #2672, opened 2023-12-14
LB-947: Change entity pills on charts page from button to anchor

piperswe:
# Problem

https://tickets.metabrainz.org/browse/LB-947

Dipping my toes in, looking to contribute something to a project I've been loving lately!

# Solution

I switched the element from a `<button>` to an `<a>`, and also added href properties to their usage in `UserEntityChart`. It defaults to `#` as a hash to preserve the pointer cursor, and for accessibility reasons. Changed validated locally - URL appears in status bar and open-in-new-tab (both middle click and right click -> open in new tab) seems to work fine. I integrated the proposed fix in LB-925, so ctrl/cmd+click also work.

Open question: should it only be an `<a>` if the href property is set? Semantically, `<button>` makes sense in several of the other places `Pill` is used. Not sure how much it matters though.

### DIFF
--- a/frontend/js/src/stats/UserEntityChart.tsx
+++ b/frontend/js/src/stats/UserEntityChart.tsx
@@ -116,6 +116,11 @@ export default class UserEntityChart extends React.Component<
     this.syncStateWithURL();
   };
 
+  entityHref = (newEntity: Entity): string => {
+    const { range } = this.state;
+    return this.buildURLParams(1, range, newEntity);
+  };
+
   getInitData = async (
     range: UserStatsAPIRange,
     entity: Entity
@@ -239,7 +244,7 @@ export default class UserEntityChart extends React.Component<
             count: elem.listen_count,
             caaID: elem.caa_id,
             caaReleaseMBID: elem.caa_release_mbid,
-            artists: elem.artists
+            artists: elem.artists,
           };
         })
         .reverse();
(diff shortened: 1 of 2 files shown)


frontend/js/src/stats/UserEntityChart.tsx
```diff
@@ -239,7 +244,7 @@ export default class UserEntityChart extends React.Component<
             count: elem.listen_count,
             caaID: elem.caa_id,
             caaReleaseMBID: elem.caa_release_mbid,
-            artists: elem.artists
+            artists: elem.artists,
```
piperswe, line 247: this change was made by prettier, looks like there are a few outstanding lint issues? the dev environment automatically runs prettier, so there's no good way to get around including this diff in the PR

MonkeyDo (changes requested):
Thank you for opening this PR, and welcome ! :)

I've got a suggestion below to apply this fix more generally:

MonkeyDo:
> should it only be an <a> if the href property is set? Semantically, <button> makes sense in several of the other places Pill is used. Not sure how much it matters though.

A good question. I think the answer for now is "don't worry about it", as we started talking about moving to a single page application which would change the way we handle navigation.

MonkeyDo:
I had to manually accept for our CI build to be run, and now that that is done there are a few failing tests regarding this change from button to anchor.

Let me know if you need anything !

MonkeyDo:
Hi @piperswe ! 
Let me know if you have any time to finish this PR up.
No pressure; if you would rather pass the baton on to someone else that's also fine, just let us know :)

MonkeyDo:
@piperswe Ping!
Hello, checking in to see if I should take over this PR or if you'd like to finish it. Please let me know!

PR #2676, opened 2023-12-18
 Store yim playlists in json data too 

amCap1712:
Refactor serialize_jspf and fetch_playlist_recordings_metadata. Move these functions to playlist model and database layer respectively to avoid circular import.

Store yim playlists in data too so that frontend doesn't need to make extra api calls.


mayhem (approved):
LGTM

PR #2682, opened 2023-12-26
Fix broken readthedocs links

mwiencek:
I noticed the "data dumps" link on the homepage was broken in the same way as https://github.com/metabrainz/listenbrainz-server/pull/2594.

Then I searched for other instances of "en/latest/dev" and found even more broken links. I guess something changed with your readthedocs setup at some point (?).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -122,7 +122,7 @@ html_theme = 'sphinx_rtd_theme'
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
-    'canonical_url': "https://listenbrainz.readthedocs.io/en/production/",
+    'canonical_url': "https://listenbrainz.readthedocs.io/en/latest/",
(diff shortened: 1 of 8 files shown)


docs/conf.py
```diff
@@ -122,7 +122,7 @@
 # further.  For a list of options available for each theme, see the
 # documentation.
 html_theme_options = {
-    'canonical_url': "https://listenbrainz.readthedocs.io/en/production/",
+    'canonical_url': "https://listenbrainz.readthedocs.io/en/latest/",
```
mwiencek, line 125: I'm not actually sure what this is used for, but based on the other changes here, I'm guessing this is correct (let me know if not, and I'll revert).

amCap1712: Looks right

amCap1712 (approved):
LGTM, Thanks!

PR #2683, opened 2023-12-26
LB-1407 (I): Listening stats don't add up

amCap1712:
Instead of partitioning on user ids, need to partition on entity mbid because we want top listeners per entity. Also, add the correct total listeners count.


MonkeyDo (commented):
Sweet ! Thanks for looking into this

mayhem (approved):

PR #2686, opened 2023-12-27
Album page: fix listen formatting for BrainzPlayer

MonkeyDo:
Currently, with #2639 not yet merged, the album page can only play a single listen at a time and then stops once at the end of the track.
This is because of how I implemented the translation from the tracks we get in props into the Listen format BrainzPlayer expects and relies on to figure out which track is playing.

I reused the utility function I already refactored when working on the Artist page, but Ideally we would send the same track format to both pages to avoid this game.


amCap1712 (approved):
LGTM, Thanks!

PR #2688, opened 2023-12-28
LB-1264: Releases by "Various Artists" should not be included on "Fresh releases" page

anshg1214:
This PR 
1. Adds filtering option for Various Artists in the frontend
2. Caches the API calls for fresh release page

### DIFF
--- a/frontend/js/src/explore/fresh-releases/FreshReleases.tsx
+++ b/frontend/js/src/explore/fresh-releases/FreshReleases.tsx
@@ -248,6 +248,13 @@ export default function FreshReleases() {
           errorMessage: "",
         };
       } catch (error) {
+        toast.error(
+          <ToastMsg
+            title="Couldn't fetch fresh releases"
+            message={error.message}
+          />,
+          { toastId: "fetch-error" }
+        );
         return {
           data: {
             releases: [],
@@ -274,18 +281,6 @@ export default function FreshReleases() {
   const { releases, releaseTypes, releaseTags } = rawData;
 
   React.useEffect(() => {
-    if (hasError) {
-      toast.error(
-        <ToastMsg
-          title="Couldn't fetch fresh releases"
-          message={errorMessage}
-        />,
-        { toastId: "fetch-error" }
-      );
-    }
-  }, [hasError, errorMessage]);
-
-  React.useEffect(() => {
     if (releases.length > 0) {
       setFilteredList(releases);
     } else if (releases.length === 0 && filteredList.length > 0) {
@@ -301,13 +296,18 @@ export default function FreshReleases() {
     });
   }, [releaseTypes, releaseTags]);
 
-  React.useEffect(() => {
-    if (releases.length > 0) {
-      setFilteredList(releases);
-    } else if (releases.length === 0 && filteredList.length > 0) {
-      setFilteredList([]);
-    }
-  }, [releases]);
+  let alt;
+  let message;
+  if (hasError) {
+    alt = "Error fetching releases";
+    message = `Error fetching releases: ${errorMessage}`;
+  } else if (releases.length === 0) {
+    alt = "No releases";
+    message = "No releases";
+  } else {
+    alt = "No filtered releases";
+    message = `0/${releases.length} releases match your filters.`;
+  }
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


amCap1712:
This is not the right way to check for various artists. Various artists releases have the special artist_credit_id of 1.

amCap1712:
I would suggest doing this filtering on frontend instead so that users have the option to show/hide va releases.

MonkeyDo (commented):
Preliminary review/questions.

Maybe I'm misunderstanding some aspects here, let me know if I got it all wrong.

Haven't tried running the PR yet, I'm waiting until we merge the search PR to deploy it to test

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
```diff
@@ -217,29 +238,76 @@ export default function FreshReleases() {
...
+        />,
+        { toastId: "fetch-error" }
+      );
+    }
+  }, [hasError, errorMessage]);
+
+  React.useEffect(() => {
+    if (releases.length > 0) {
+      setFilteredList(releases);
+    } else if (releases.length === 0 && filteredList.length > 0) {
```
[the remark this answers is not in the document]
anshg1214: I ran into a weird warning.
```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```

The issue was that while fetching, the dependencies of this useEffect releaseTypes and releaseTags are empty arrays, and they were being replaced by empty arrays. I added a debugger log this change, and it says [] is being replaced by [].

This is why I added a special case to handle this change

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
```diff
@@ -217,29 +238,76 @@ export default function FreshReleases() {
...
+    }
+  }, [releases]);
+
+  React.useEffect(() => {
+    if (releaseTypes.length === 0 && releaseTags.length === 0) return;
+    setAllFilters({
+      releaseTypes,
+      releaseTags,
+    });
+  }, [releaseTypes, releaseTags]);
```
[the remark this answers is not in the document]
anshg1214: Correct.
I believe I found the issue I was facing. When I was creating a map for `allFilters` and sending to the ReleaseFilters component, But the issue was that, on every re-render, this map was being created again and again. And since it's a `map`, it is compared by reference and not by value like other non-primitive type data types, causing the useEffect to trigger.

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
```diff
@@ -217,29 +231,66 @@ export default function FreshReleases() {
...
+    errorMessage = "",
+  } = loaderData || {};
+
+  const { releases, releaseTypes, releaseTags } = rawData;
+
+  const releasesString = JSON.stringify(releases);
+  React.useEffect(() => {
+    const newReleases = JSON.parse(releasesString);
+    setFilteredList(newReleases);
+  }, [releasesString]);
```
MonkeyDo, line 280: I'm not sure how to test this locally properly, but I wonder if you need the stringification here.
Does this result int he same re-rendering loop issue still?
If so, I think something smells a bit fishy with the way we manipulate the state between parent and child components...
```
React.useEffect(() => {
    setFilteredList(releases);
  }, [releases]);
```

MonkeyDo (commented):
Deployed to test.LB for testing, and working perfectly, nice work !

I do have a question below regarding the odd re-render loop you ran into previously, but feel free to merge if I'm just being thick :P

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
anshg1214: Yes, If i don't add a stringification, I get a warning on the browser console

`FreshReleases.tsx:283 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`

MonkeyDo: I found the culprit !
I did say something was smelling fishy :D

In cases of circular dependency/render issues like this, my usual first step is to look at the dependency arrays and see if anything is missing in them.
In this case that turned out o be the issue, so an easy fix.
Disabling the linting rule `react-hooks/exhaustive-deps` is a dangerous game, it turns out :P

In our case, since we had `if (!_.isEqual(filteredReleases, filteredList)) {` and `filteredList` was never updated in the effect in ReleaseFilters, the condition was always true and was calling the parent's `setFilteredList`, which in turn re-rendered the child with the new `filteredList`, and so on, hence the infinite re-render loop.

MonkeyDo (approved):
With the circular dependency fixed, I think we are ready to go !
Nicely done :)

I will let you merge it yourself, in case there was something else you wanted ti look at or add.

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
anshg1214: Got it. Thank you so much for fixing this!

PR #2690, opened 2023-12-29
Rank Singles/EPs before Albums with secondary types in mapping

amCap1712:
TBA

### DIFF
--- a/mbid_mapping/mapping/custom_sorts.py
+++ b/mbid_mapping/mapping/custom_sorts.py
@@ -1,5 +1,13 @@
 import psycopg2
 
+RELEASE_GROUP_PRIMARY_TYPES = [
+    (1, "Album"),
+    (2, "Single"),
+    (3, "EP"),
+    (11, "Other"),
+    (12, "Broadcast")
+]
+
 RELEASE_GROUP_SECONDARY_TYPES = [
     (2, "Soundtrack"),
     (9, "Mixtape/Street"),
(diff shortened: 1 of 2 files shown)


mbid_mapping/mapping/custom_sorts.py
```diff
@@ -106,6 +114,37 @@
 ]
 
 
+def get_combined_release_group_types_sort():
```
mayhem, line 117: This has got to be one of the weirder functions you've written in your life, I am sure. :)

mayhem (commented):
I'm very curious to see how this goes; however, I am a bit hesitant given that no tests were added/updated in this PR -- do you have items to add to the tests?

amCap1712:
I am not yet sure how to test this. I'll think more about it. 

But I checked the tables manually, and of the 24M canonical recordings in the canonical_musicbrainz_data table, only 200k rows change. Even out of that, only half are because of this change (the rest are because of a previously known issue). So only 0.5% of the canonical recordings are changed by this PR.

mayhem:
Have you run this test? https://github.com/metabrainz/listenbrainz-server/blob/master/mbid_mapping/manage.py#L50 If anything fails, we need to examine it.

amCap1712:
27 passed, 0 failed. I'll think of more cases to add to it.

mayhem:
Great, please do that, then feel free to merge!

mayhem (approved):

mbid_mapping/mapping/custom_sorts.py
amCap1712: Haha yes.

PR #2693, opened 2024-01-01
LB-1417: Add recording artists to release group cache

amCap1712:


frontend/js/src/entity-pages/AlbumPage.tsx
```diff
@@ -164,13 +168,16 @@ export default function AlbumPage(props: AlbumPageProps): JSX.Element {
...
               merge(track, {
-                artist_name: artistName,
-                artist_mbids:
-                  artist?.artists?.map((ar) => ar.artist_mbid) ?? [],
+                artist_name: track.artists
+                  .map(
+                    (track_artist) =>
+                      `${track_artist.artist_credit_name}${track_artist.join_phrase}`
+                  )
+                  .join(""),
```
MonkeyDo, line 176: Let's do this for now; in the future I think we can refactor the [`getArtistName` utility](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/utils/utils.tsx#L231) to accept more input formats.

We probably want to keep the artist_mbids in there, as they can be used for other purposes down the line such as CB reviews

frontend/js/src/entity-pages/AlbumPage.tsx
```diff
@@ -164,13 +168,16 @@ export default function AlbumPage(props: AlbumPageProps): JSX.Element {
...
                 caa_id: Number(caa_id) || undefined,
                 caa_release_mbid,
                 recording_name: track.name,
-                release_mbid: caa_release_mbid ?? "",
+                release_mbid: recordings_release_mbid ?? "",
```
MonkeyDo, line 180: Would this make sense, just to ensure a cover art fallback?
```suggestion
                release_mbid: recordings_release_mbid ?? caa_release_mbid ?? "",
```

MonkeyDo (commented):
Good improvement

MonkeyDo:
Tested live on test.LB, working marvelously ! 💯

mayhem (approved):
Yeah, good improvement!

frontend/js/src/entity-pages/AlbumPage.tsx
amCap1712: Sounds good.

PR #2695, opened 2024-01-04
Migrate Profile pages to SPA

anshg1214:
This PR migrates Profile pages to SPA. Also the `/profile` route has been migrated to `/settings`


frontend/js/src/settings/missing-data/MissingMBData.tsx
```diff
@@ -266,137 +287,120 @@ export default class MissingMBDataPage extends React.Component<
...
-            user={user}
-          />
-        </NiceModal.Provider>
-      </GlobalAppContext.Provider>
-    </ErrorBoundary>
-  );
-});
+export const MissingMBDataPageLoader = async () => {
+  const response = await fetch("/profile/missing-data/", {
+    method: "POST",
```
MonkeyDo, line 399: I don't understand why these requests to fetch data are POST rather than GET requests.
Could you walk me through it?

frontend/js/src/settings/missing-data/MissingMBData.tsx
```diff
@@ -266,137 +287,120 @@ export default class MissingMBDataPage extends React.Component<
...
-      integrations: [new Integrations.BrowserTracing()],
-      tracesSampleRate: sentry_traces_sample_rate,
-    });
-  }
-
-  const { missingData, user } = reactProps;
+export function MissingMBDataPageWrapper() {
+  const data = useLoaderData() as MissingMBDataProps;
+  const { currentUser: user } = React.useContext(GlobalAppContext);
+  return <MissingMBDataPage {...data} user={user} />;
```
MonkeyDo, line 394: I think this might need to be: 
```suggestion
  return <MissingMBDataPage missingData={data} user={user} />;
```

frontend/js/src/settings/missing-data/MissingMBData.tsx
```diff
@@ -266,137 +287,120 @@ export default class MissingMBDataPage extends React.Component<
...
-});
+export const MissingMBDataPageLoader = async () => {
+  const response = await fetch("/profile/missing-data/", {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+  });
+  const data = await response.json();
+  return { ...data };
```
MonkeyDo, line 405: Is there a specific reason to be returning a clone of the object here?

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -0,0 +1,257 @@
...
+          ...prevState,
+          [serviceName]: newValue,
+        }));
+        return;
+      }
+
+      const data = await response.json();
+      const { url } = data;
+
+      window.location.href = url;
```
MonkeyDo, line 60: Here's the culprit for the page refresh. Probably unwanted.

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -0,0 +1,134 @@
...
+};
+
+export default function Import() {
+  const { currentUser } = React.useContext(GlobalAppContext);
+  const { name } = currentUser;
+  const data = useLoaderData() as ImportLoaderData;
+  const {
+    user_has_email: userHasEmail,
+    profile_url: profileUrl,
+    apiUrl,
```
MonkeyDo, line 24: apiURL is missing from props sent by the server, but we should probably be using the one from the GlobalContext.APIService instead.

MonkeyDo (commented):
I deployed this PR on test, seems to be working nicely!

Found a couple of issues, but in general it's clean and more functional than before.

The missing data page isn't working for me. Nothing is shown.

When I change my spotify permissions it refreshes the page, which was intended before but not necessary now.

Left comments for what I think the culprits are.

frontend/js/src/settings/music-services/details/MusicServices.tsx
anshg1214: The website does not refresh, what happens in background is that the `/disconnect/` returns the `url` which is the link to the external service provider which redirects to the callback url in the backend which then reloads the page.

Right now you would be signed in to the respective service, you would have observed a reload. If you try it on incognito, then you can test the entire flow.

frontend/js/src/settings/import/ImportListens.tsx
anshg1214: Fixed

frontend/js/src/settings/missing-data/MissingMBData.tsx
anshg1214: Fixed

frontend/js/src/settings/missing-data/MissingMBData.tsx
anshg1214: Since the routing is now being handled by react-router-dom, the GET request to the URL now return the page, and the POST requests now return the initial props which are essential for the page.

anshg1214:
@MonkeyDo Should we migrate the routes from `/profile` to `/settings`?

MonkeyDo:
> @MonkeyDo Should we migrate the routes from `/profile` to `/settings`?

Yes please !

frontend/js/src/settings/music-services/details/MusicServices.tsx
MonkeyDo: OK, I read the code in profile.py, thanks for the explanation.

Now I remember why we had to set `window.location.href`: that is the only way we can show the spotify login/authorization windows.

I haven't tried it, but I wonder if there is any chance of us improving things with `redirect: 'follow'` in the fetch options. Not sure that will do anything better than whet we are currently doing.
Anyway, that's a problem for another day.

listenbrainz/webserver/views/settings.py
```diff
@@ -354,17 +269,16 @@ def music_services_details():
...
+    data = {
+        "current_spotify_permissions": current_spotify_permissions,
+        "current_critiquebrainz_permissions": current_critiquebrainz_permissions,
+        "current_soundcloud_permissions": current_soundcloud_permissions,
+    }
+
+    return jsonify(data)
 
 
+@settings_bp.route('/music-services/<service_name>/callback/')
```
anshg1214, line 281: Since the routes have migrated, I've enabled callback URLs for both the `/profile` and `/settings` routes. Do we want to keep both of these callback URLs or just one? @MonkeyDo

MonkeyDo (approved):
Looking great ! :rocket:

PR #2696, opened 2024-01-05
Restructure Year in Music tables

amCap1712:
Writing to year in music table generates a lot of WAL which in combination with other issues caused an outage on MB and other projects. To avoid this in future,

1. Move the YIM tables to gaga.
2. Always create a new unlogged table at start of YIM process.
3. Insert the data in the unlogged table.
4. Once all inserts are complete, run vacuum and swap the table in prod.

I have already migrated the data for older years to the new format as well.


listenbrainz/db/year_in_music.py
```diff
@@ -41,111 +39,129 @@
...
+        }
+        for user_id, similar_users in data.items()
+    ]
+
+    insert_heavy("similar_users", year, similar_user_data)
+
+
+def insert_top_stats(entity, year, data):
+    insert_heavy(f"top_{entity}", year, data)
+    insert(f"total_{entity}_count", year, [(user["user_id"], user["count"]) for user in data], False)
```
mayhem, line 126: I think we can improve this by using on INSERT statement, but lets skip it for this second.

amCap1712: Sorry I am not sure what you mean?

amCap1712: ah okay, I see. Yes that is possible, can do later.

PR #2698, opened 2024-01-05
Entity pages: Fix listening stats

MonkeyDo:
Listening stats values can apparently be undefined, which wasn't planned for, and this results in a NaN when we try to format them:

image

Adding a check to ensure they are a number.


kellnerd:
It also affects album pages, which are not handled here as far as I can see?

MonkeyDo:
> It also affects album pages, which are not handled here as far as I can see?

Yep, forgot to push commit, thanks :)

mayhem (approved):

PR #2700, opened 2024-01-06
Fix get_user_details for YIM playlists

amCap1712:
The existing query excluded users who had performed a last.fm import or linked MB but didn't connect spotify. Fixed it and added a test as well.


mayhem (approved):
Great!

PR #2704, opened 2024-01-06
YIM Frontend fixes - batch 3

amCap1712:
1. LB-1429: TypeError on Year in Music 2023
2. LB-1430: YIM 2023 do not show follow button while logged out
3. Do not show empty sections for users with no YIM and improve the css for the same case.


mayhem (approved):
I'm sure its ok. :laughing:

PR #2708, opened 2024-01-11
Don't submit Spotify listens if account is already set to record them

MonkeyDo:
We had previously (in commit 4f9d218) removed this functionality because Spotify had stopped recording the history of listens when playing using BrainzPlayer and Spotify SDK.
This appears to have been fixed recently and we should revert the change to avoid duplicates

### DIFF
--- a/frontend/js/src/brainzplayer/SpotifyPlayer.tsx
+++ b/frontend/js/src/brainzplayer/SpotifyPlayer.tsx
@@ -7,6 +7,7 @@ import {
   debounce as _debounce,
   isString,
   has,
+  difference,
 } from "lodash";
 import { faSpotify } from "@fortawesome/free-brands-svg-icons";
 import {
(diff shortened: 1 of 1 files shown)


MonkeyDo:
Deployed on test.LB and seems to be working as intended.

amCap1712 (approved):
LGTM but I would suggest that we do submit now playing listens from BP so that listening-now page shows upto date data.

MonkeyDo:
> LGTM but I would suggest that we do submit now playing listens from BP so that listening-now page shows upto date data.

:+1: We already do that and submit a playing_now always:
https://github.com/metabrainz/listenbrainz-server/blob/50cbc1574798d156b790a7e768c4a2fa94d52060/frontend/js/src/brainzplayer/BrainzPlayer.tsx#L615

PR #2713, opened 2024-01-15
Remove whitespace in SVG title elements

kellnerd:
# Problem

While whitespace seems to have no effect in Chromium-based browsers, Firefox includes it in its tooltip rendering, which looks weird:

screenshot

# Solution

Slurp leading and trailing whitespace with the appropriate template tags. I found these after having a brief look at the Jinja docs, so let me know if you prefer a different solution (such as inlining the template tag).

# Action

I haven't set up a development server to test this, so a sanity check would be appreciated.

### DIFF
--- a/listenbrainz/webserver/templates/art/svg-templates/lps-on-the-floor.svg
+++ b/listenbrainz/webserver/templates/art/svg-templates/lps-on-the-floor.svg
@@ -13,7 +13,7 @@
     >
         {% if image.title and image.artist %}
         <title>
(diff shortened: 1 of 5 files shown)


amCap1712 (approved):
LGTM, Thanks! I can deploy to test.lb later to confirm it works.

MonkeyDo (approved):
Thank you !
Deployed to test.LB and tested, it's definitely much better on Firefox !

PR #2714, opened 2024-01-15
Create separate MB service type for each LB deployment

amCap1712:
LB has 3 deployments and each deployment is a different MB OAuth application. As of now, we store all the access tokens and refresh tokens in one 'musicbrainz' service entry which creates problems. Logging in on test/beta/prod overrides the tokens generated on other deployments which essentially means refreshing tokens will fail. Further, MB doesn't return refresh tokens on each login so once broken the tokens will forever remain broken.

To fix, create 3 separate musicbrainz service types, one matching each LB deployment.

### DIFF
--- a/admin/sql/create_types.sql
+++ b/admin/sql/create_types.sql
@@ -10,7 +10,7 @@ CREATE TYPE user_timeline_event_type_enum AS ENUM('recording_recommendation', 'n
 
 CREATE TYPE hide_user_timeline_event_type_enum AS ENUM('recording_recommendation', 'recording_pin');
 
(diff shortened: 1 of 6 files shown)


amCap1712:
Further, we'll need to revoke all user's approval to LB in MB and logout all users in LB.

mayhem (approved):
Will this fix the tagging issues on the site?

amCap1712:
Yup, this is for fixing tagging issues indeed.

mayhem:
Excellent, log everyone out if it fixes the problem!

PR #2715, opened 2024-01-16
Tags component improvements

MonkeyDo:
Users have been asking to see non-genre tags in the entity pages.
We initially filtered them out because the tags component wasn't suited for showing a lot of tags comfortably.

This PR addresses this issue by showing all the tags in one line (prevent wrapping of the component) with a scroll indicator to hint at more content.

image


amCap1712 (approved):
LGTM, Thanks!

PR #2716, opened 2024-01-16
LB-1420: Use smaller cover art on listening now page

MonkeyDo:
In case the user provides a release MBID, we are currently fetching the front image (original size) instead of a 500px version like we do below when no MBID is provided.
This is wasteful of users' resources and is only due to my lack of knowledge of the cover art archive's API.


amCap1712 (approved):
LGTM, Thanks!

PR #2717, opened 2024-01-17
LB-1366: Updated POST /grid/ endpoint

07jasjeet:
# Problem

### [LB-1366](https://tickets.metabrainz.org/browse/LB-1366)

# Action

1) The endpoint now accepts release_group_mbids array. 
2) Capped the number of mbids to process in a request to 100.


07jasjeet:
How should I go about the tests? Create new or edit pre-existing tests.

listenbrainz/webserver/views/art_api.py
```diff
@@ -101,21 +107,49 @@ def cover_art_grid_post():
...
-    images = cac.load_images(r["release_mbids"], tile_addrs=tiles, layout=layout, cover_art_size=cover_art_size)
+    # Get release_mbids
+    release_mbids = []
+    if "release_mbids" in r:
+        if not isinstance(r["release_mbids"], list):
+            raise APIBadRequest("release_mbids must be a list of strings specifying release_mbids")
+
+        release_mbids = list(r["release_mbids"])
+
+    # Get release_group_mbids
```
amCap1712, line 125: I think we don't need to support having both release mbids and release group mbids. If release mbids are specified used them if not try for release group mbids.

amCap1712 (changes requested):
Thanks for the PR. Using both releases and release groups like this will break some links in the SVGs so instead we simplify a few things and only support release or release groups at a time but not both.

listenbrainz/webserver/views/art_api.py
```diff
@@ -101,28 +107,61 @@ def cover_art_grid_post():
...
         cover_art_size = 250
     else:
         cover_art_size = 500
 
-    images = cac.load_images(r["release_mbids"], tile_addrs=tiles, layout=layout, cover_art_size=cover_art_size)
+    # Get release_mbids or release_group_mbids
+    mbids = []
+    entity = "release"
+    
+    def look_for_release_group_mbids():
```
mayhem, line 121: I feel this function is named poorly. Right now it sounds more like a function to lookup some data, but its actually a sanity check.  sanity_check_mbids()? Also, a docstring is really needed here.

listenbrainz/webserver/views/art_api.py
```diff
@@ -101,28 +107,61 @@ def cover_art_grid_post():
...
+        if "release_group_mbids" not in r:
+            return
+        
+        if not isinstance(r["release_group_mbids"], list):
+            raise APIBadRequest("release_group_mbids must be a list of strings specifying release_group_mbids")
+
+        nonlocal mbids
```
mayhem, line 128: The use of nonlocal to return values to the calling function is weird. Why not pass in the values and return them at the end of the function?

mayhem (changes requested):
Pretty good, but I think some minor cleanup would be good.

07jasjeet:
Totally forgot about this one lol. Will do the changes.

listenbrainz/webserver/views/art_api.py
07jasjeet: I usually like this if its an inner function since the scope of an inner function's usage is very limited, meaning there are no ghost usages (usages that devs aren't aware about). I would not have to create new variables again and again, but rather just assign the variables their respective values without much boilerplate, hence looks nicer syntactically. 

Thats just my pov, totally your call on how we should go forward.

listenbrainz/webserver/views/art_api.py
07jasjeet: This function assigns values to variables and throws error if the response is not sane. It looks fine but your call!

Docstring added!

listenbrainz/art/cover_art_generator.py
```diff
@@ -197,27 +197,43 @@ def resolve_cover_art(self, caa_id, caa_release_mbid, cover_art_size=500):
...
+
+    def load_images(self, release_mbids, release_group_mbids=[], tile_addrs=None, layout=None, cover_art_size=500):
+        """ Given a list of release and release group MBIDs and optional tile addresses, resolve all the cover art design,
+            all the cover art to be used and then return the list of images and locations where they should be
             placed. Return an array of dicts containing the image coordinates and the URL of the image. """
 
-        release_mbids = [mbid for mbid in mbids if mbid]
-        results = self.load_caa_ids(release_mbids)
+        results = {}
+        if release_mbids and len(release_mbids) != 0:
```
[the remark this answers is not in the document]
07jasjeet: This was my logic before, but there was a case i encountered where list size 0 would slip out.

listenbrainz/webserver/views/art_api.py
amCap1712: Yeah, let's avoid nonlocal. We don't use it anywhere else in the code and things like ghost usages are not an issue because this is all internal code, no library or direct external use intended. You can prefix the function with an `_` to hint that it is private.

listenbrainz/art/cover_art_generator.py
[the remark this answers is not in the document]
amCap1712: `if release_mbids` would evaluate to false for an empty list as empty lists are falsy values in python.
image

So the `len(release_mbids) != 0` is not needed but I think some other part of code might have triggered the error you saw.

listenbrainz/webserver/views/art_api.py
07jasjeet: Done

listenbrainz/webserver/views/art_api.py
07jasjeet: Done

listenbrainz/art/cover_art_generator.py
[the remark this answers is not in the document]
07jasjeet: Done

mayhem (approved):
Looks good to me now. Any more thoughts @amCap1712 ?

amCap1712 (approved):
LGTM, Thanks!

mayhem:
Thanks Jasje!

PR #2719, opened 2024-01-18
Generify mb_conn and lb_conn in BulkTable

amCap1712:
BulkTable assumed that the select should always happen from mb_conn and the inserts should happen in lb_conn if available else mb_conn. However, we want mapping.canonical_recording_release_direct to be present in both databases. That means one time we will have to read the data from lb_conn and write to mb_conn. Hence, making mb_conn and lb_conn generic by renaming it to select_conn and insert_conn respectively.

### DIFF
--- a/mbid_mapping/mapping/bulk_table.py
+++ b/mbid_mapping/mapping/bulk_table.py
@@ -19,7 +19,7 @@ class BulkInsertTable:
 
     def __init__(self, table_name, select_conn, insert_conn=None, batch_size=None, unlogged=None):
         """
-            This init function expects to be passed a database connection to read data from (usually to MB)
+            This init function expects to be passed a database connection to read data from (usually MB)
             and an optional DB connection to write data to (usually ListenBrainz). If the insert_conn is
             passed in, the newly written table will be written to that database and the select_conn connection
             will only be used to fetch the data. However, if insert_conn is not passed in, the select_conn
@@ -72,7 +72,7 @@ class BulkInsertTable:
             The process_row function will be called for each of the rows resulting from the queries.
 
             Example:
-                ["SELECT * from ...", "SELECT * from ... ")]
+                ["SELECT * from ...", "SELECT * from ... "]
         """
         return []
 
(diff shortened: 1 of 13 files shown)


mbid_mapping/mapping/bulk_table.py
```diff
@@ -17,21 +17,22 @@ class BulkInsertTable:
...
     """
 
-    def __init__(self, table_name, mb_conn, lb_conn=None, batch_size=None):
+    def __init__(self, table_name, select_conn, insert_conn=None, batch_size=None, unlogged=None):
         """
-            This init function expects to be passed a database connection (usually to MB)
-            and an optional DB connection to ListenBrainz. If the lb_conn is passed in,
-            the newly written table will be written to that database and the mb_conn connection
-            will only be used to fetch the data.
+            This init function expects to be passed a database connection to read data from (usually to MB)
```
kellnerd, line 22:
```suggestion
            This init function expects to be passed a database connection to read data from (usually MB)
```

mbid_mapping/mapping/bulk_table.py
```diff
@@ -65,12 +68,11 @@ def get_create_table_columns(self):
...
     def get_insert_queries(self):
         """
-            Returns a list of tuples with DB nick to connect to and data insert queries to run.
+            Returns a list of data insert queries to run.
             The process_row function will be called for each of the rows resulting from the queries.
 
             Example:
-                [("MB", "SELECT * from ..."),    # Select data from MB
-                 ("LB", "SELECT * from ... ")]   # Select data from LB
+                ["SELECT * from ...", "SELECT * from ... ")]
```
kellnerd, line 75:
```suggestion
                ["SELECT * from ...", "SELECT * from ... "]
```

kellnerd (commented):
Two minor nitpicks

mbid_mapping/mapping/bulk_table.py
amCap1712: Thanks, fixed.

mbid_mapping/mapping/bulk_table.py
amCap1712: Thanks, fixed.

PR #2720, opened 2024-01-18
Remove mb_metadata_cache's dependency on canonical_musicbrainz_data

amCap1712:
mb_metadata_cache requires only year field from canonical_musicbrainz_data but for this reason we need to build the table in both LB and MB database which is expensive. It is better to read the year from release_first_release_date table.


mayhem (approved):
Good insight!

PR #2725, opened 2024-01-18
Refactor Folder Structure for multiple pages

anshg1214:
This PR refactors the folder structure for
1. User Dashboard
2. Entity Pages
3. Playlist Page
4. Metadata Viewer Page
5. Player Page


MonkeyDo (changes requested):
Thanks for working on this ! It's going to be a lot clearer now.
I think some of these components are not quite in the right place though, my thoughts:

MBIDMappingModal should be in the 'listens' folder with the ListenCard, it's not related to the missing data page only.

I think UserSearch could be higher up in the tree as well. We use it for example in the ArtCreator component as well.
Should we have a `src/common` folder to put such components that can be used across the website and aren't tied to a specific feature? Maybe the current src/listens folder, which is already sort of used for that purpose…

PlaylistCard, PlaylistMenu and the two modals CreateOrEditPlaylistModal and DeletePlaylistConfirmationModal all probably belong to src/playlist

I wonder if src/user/charts should be merged into user/stats. After all the charts page is only an extension of the stats page…

SimilarityScore should go in user/follow/, it's only used in UserListModalEntry

And finally, more generally I think you should also mirror the new folder structure for the tests folder so that it is as easy as possible to find the corresponding test file when working on a component.

anshg1214:
Thanks for the feedback. I've restructured common files, and the other files you mentioned. Along with that, the test folder now is the mirror of the new folder structure.

According to me, lets not merge src/user/charts into /user/stats now. I'll merge the folder and refactor the files while migrating the user dashboard to SPA.

MonkeyDo (approved):
Good to go, thank you !

PR #2727, opened 2024-01-20
TagComponent: Fix vertical alignment

kellnerd:
# Problem

The tag vote buttons were shifted downwards by a few pixels with the previous styling:

before

# Solution

069f1330acfcb78671d208480441a3deaf62c9f3 recently removed the vertical alignment setting, presumably to fix the alignment of the new tag count, but removing the inline-block property is the better solution.
Tag label and count are now aligned to the middle instead of the baseline:

after

# Action

The changes have only been tested by manipulating the live CSS with the Firefox developer tools, but I assume that I have found the right place in the Less source files.

### DIFF
--- a/frontend/css/tags.less
+++ b/frontend/css/tags.less
@@ -15,8 +15,8 @@
   }
   > * {
     height: 100%;
-    display: inline-block;
     padding: 3px 7px;
+    vertical-align: middle;
     &:first-child {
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Thanks ! Didn't even see that after looking at those tags for so long :p

PR #2728, opened 2024-01-22
LB-1237: ListenBrainz Android app button in footer

07jasjeet:
# Problem

[LB-1237](https://tickets.metabrainz.org/browse/LB-1237)

# Action

Added app button to footer by generating badge from https://play.google.com/intl/en_us/badges/.
Do read the mentioned website before proceeding.

### DIFF
--- a/listenbrainz/webserver/templates/base.html
+++ b/listenbrainz/webserver/templates/base.html
@@ -62,7 +62,7 @@
           </ul>
         </div>
         <br/>
-        <div class="col-xs-12 col-sm-6 col-md-3">
+        <div class="col-xs-12 col-sm-6 col-md-3 col-lg-2">
           <h3 class="w-title-a text-brand">Useful Links</h3>
           <ul class="list-unstyled">
             <li>
@@ -91,7 +91,7 @@
             </li>
           </ul>
         </div>
-        <div class="col-xs-12 col-sm-6 col-md-3">
+        <div class="col-xs-12 col-sm-6 col-md-3 col-lg-2">
           <h3 class="w-title-a text-brand">Fellow Projects</h3>
           <ul class="list-unstyled">
             <li>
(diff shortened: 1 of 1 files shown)


MonkeyDo:
I took the liberty of continuing this PR.
I think since the PR was opened, the LB app has improved significantly, and while I might have had hesitations about pushing it in the past I think we are well past the point we should be doing so.

I also took the liberty to refresh the look of the footer a bit.
@Aerozol I would love your opinion on those changes.
I went back to the redesign figma and made the footer a bit closer to the OG design.
I also rephrased the blurb in the footer which I thought had a lot of repetitions.
Let me know what you think !

### Before
image

### After

image
image
image

MonkeyDo:
@07jasjeet I would ask you to please check if the link to the app is still correct (I opened it and it looks fine, but there's a lot of stuff in that URL, just wanted to double-check)

I would also note that we might want to change the video used in the app store, it's weird to have a big CritiqueBrainz logo as the header when you first open the page...

07jasjeet:
> @07jasjeet I would ask you to please check if the link to the app is still correct (I opened it and it looks fine, but there's a lot of stuff in that URL, just wanted to double-check)
> 
> 
> 
> I would also note that we might want to change the video used in the app store, it's weird to have a big CritiqueBrainz logo as the header when you first open the page...
> 
> 

URL is good, it just labels the date this marketing campaign was setup. Since we do not need to track what campaign raked in how much clicks, just the link working fine is good enough.

As for videos, I'll update the content (video/photos) on play store soon.

Aerozol:
Looks great!

----------------

Suggested tweaks to the text:

_ListenBrainz tracks what you listen to and turns it into insights. Explore personalized recommendations, discover new artists and share your taste with others through rich visualizations. User listen data and text is made public under the Creative Commons Zero (CCO) license._

In particular:
- I believe "user listen data **and text**" was added after a user complained about us not being clear about reviews and pins etc being made public
- Remove "You can..." from the second para, to turn it into 'active' language.

I personally don't find the line breaks without spaces in between that helpful, I'd rather read it as a distinct little text block, but mileage may vary :)

----------------

Suggested tweaks to the layout:

1. more even spacing above and below the ListenBrainz logo (I think we can afford to be a bit more generous with the whitespace, if needed)
2. Not sure about 'fellow projects' dropping down to the bottom on mobile - but if you tested it sticking to the top and it looked worse then all good
3. More space between these links and the text block on mobile

image

PR #2729, opened 2024-01-23
Show artist-recording and artist-release_group rels even if marked ended

amCap1712:
Relationships like performer relations are ended by definition because an artist performed something on a particular date. Other recording and release group rels also make sense to be shown to users always. However, ended url rels are only for historical purposes and do not need to be shown to users.

### DIFF
--- a/listenbrainz_spark/stats/listener/release_group.py
+++ b/listenbrainz_spark/stats/listener/release_group.py
@@ -1,10 +1,10 @@
 from typing import Iterator, List
 
-from data.model.entity_listener_stat import ArtistListenerRecord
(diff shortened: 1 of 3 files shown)


mayhem (approved):
A ha! Good detective work!

kellnerd (approved):
Nice one, I have been studying that query myself but my brain somehow failed to make the connection between the observed "rels with dates are missing" and the ended attribute, probably because it is set automatically once you enter a date range on MB.

amCap1712:
@kellnerd Yup, I think its set automatically.

PR #2735, opened 2024-01-29
LB-1485: Album page overflows

huhridge:
# Problem

Fix LB-1485: Album page overflows

https://tickets.metabrainz.org/projects/LB/issues/LB-1485

# Solution

A minor change in the `ListenCard.tsx` clamping the artist-name to 2 lines has fixed the issue.

# Action

No further action required.

### DIFF
--- a/frontend/css/entity-pages.less
+++ b/frontend/css/entity-pages.less
@@ -115,6 +115,7 @@
     .tracks {
       flex: 3 400px;
       min-height: 150px;
+      min-width: 0;
       .artist-page& {
         max-height: 520px;
         overflow-y: hidden;
--- a/frontend/js/src/common/listens/ListenCard.tsx
+++ b/frontend/js/src/common/listens/ListenCard.tsx
@@ -476,7 +476,7 @@ export default class ListenCard extends React.Component<
                 )}
               </div>
               <div
-                className="small text-muted ellipsis-2-lines"
+                className="small text-muted ellipsis"
                 title={artistName}
               >
                 {getArtistLink(listen)}
--- a/frontend/js/src/common/listens/ListenCard.tsx
+++ b/frontend/js/src/common/listens/ListenCard.tsx
@@ -475,7 +475,10 @@ export default class ListenCard extends React.Component<
                   </div>
                 )}
               </div>
-              <div className="small text-muted ellipsis" title={artistName}>
+              <div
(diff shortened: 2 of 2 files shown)


MonkeyDo (changes requested):
Thanks for opening a PR :)
Boy did I go for a ride while reviewing it !

I think your solution was really smart!
In theory, this ListenCard component should behave.
It is only in the album page that this issue was seen, which makes me think the solution must lie there, and I am averse to changing an element that is used across the entire website to fix an issue that we see only in one page...

To be sure, I checked what the change would mean in other circumstances, for example It would make the cards taller on small screens:
original:
image
with `ellipsis-2-lines`
image
See how the middle card is now taller than the others.

I had to dig much, much deeper to understand why specifically the album page was not behaving properly.
Contrarily to other pages, on this one we use flexbox to wrap the columns fluidly, but obviously in this case the tracks container doesn't shrink the way it should to fit the page.
I dived into stackoverflow, as one does, looking for an explanation of why it wasn't shrinking in width. See [this post](https://stackoverflow.com/a/36247448) for an answer.

It turns out setting `min-width: 0;` on the tracks container solves the issue.
Basically, by default the flexbox algorithm does not know what is the minimum width expected of the element, and just doesn't try to shrink it at all!
The class we want to add this min-width is here:
https://github.com/metabrainz/listenbrainz-server/blob/66bfa7823f2de95979b5c5307c9ace63bbb7329a/frontend/css/entity-pages.less#L115-L117

You could also set `overflow: auto;` instead of the min-width (see the stackoverflow answer) but I think that is not as clear to me.

Phew! What a journey! :)

MonkeyDo:
> I did not realise this would turn into such a in-depth incursion. I was thinking why the `flexbox` was behaving improperly, but I just thought it would be better to just clamp the artist name to 2 lines, since that seemed to have caused the content overflow. This is a better solution though definitely, I will commit it right now.

Truth be told I went back and forth myself between "oh, that's not the solution I had in mind, but it's definitely the cause of the problem!" and "but wait, why is that not happening on any other page" a couple of times before I gave up and asked The Internet.
I have to say it's a very unintuitive part of the flexbox spec…

MonkeyDo (approved):
Thank you for opening this PR, and for giving us an oportunity to learn something useful :) !

PR #2736, opened 2024-01-29
LB-1289: Fix XSPF playlist exporter bug and Enable XML support

yuktea:
# Problems

- The `serialize_xspf` method in `listenbrainz/webserver/views/playlist_api.py` has a bug which causes the `additional_metadata` field to be improperly constructed which leads to a crash.
- The response wasn't correctly formatted. 

# Solution
The following changes have been made:
- Fixed `serialize_xspf` to handle `additional_metadata` correctly. 
- Changed the response's content type in `get_playlist_xspf` to xml

### DIFF
--- a/listenbrainz/tests/integration/test_playlist_api.py
+++ b/listenbrainz/tests/integration/test_playlist_api.py
@@ -233,14 +233,19 @@ class PlaylistAPITestCase(IntegrationTestCase):
                     }
                 }
             }
-        
-        response = self.client.post(
+        response_post = self.client.post(
             url_for("playlist_api_v1.create_playlist"),
             json=playlist,
             headers={"Authorization": "Token {}".format(self.user["auth_token"])}
         )
-        print(response)
-        self.assert200(response)
+        self.assert200(response_post)
+        playlist_mbid = response_post.json["playlist_mbid"]
+        print(playlist_mbid)
+        r = self.client.get(
+            url_for("playlist_api_v1.get_playlist_xspf", playlist_mbid=playlist_mbid),
+            headers={"Authorization": "Token {}".format(self.user["auth_token"])}
+        )
+        self.assert200(r)
 
     def test_playlist_json_with_missing_keys(self):
         """ Test for checking that submitting JSON with missing keys returns 400 """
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/playlist_api.py
```diff
@@ -143,11 +143,19 @@ def serialize_xspf(playlist: Playlist):
             playlist_extension_collaborator = ET.SubElement(playlist_extension_collaborators, "collaborator")
             playlist_extension_collaborator.text = collaborator
 
-    if playlist.additional_metadata:
-        playlist_extension_additional_metadata = ET.SubElement(playlist_extension, "additional_metadata")
-        for meta_key, meta_value in playlist.additional_metadata:
-            additional_metadata_item = ET.SubElement(playlist_extension_additional_metadata, "item", key=meta_key)
-            additional_metadata_item.text = meta_value
+        if playlist.additional_metadata:
```
amCap1712, line 146: While testing the PR locally, the additional_metadata block wasn't showing up. Then I saw this entire block is over indented. It should be on the same indentation as other if blocks.

amCap1712 (changes requested):
Thanks for the PR. I tested locally and it works. Let's add a test to ensure it remains that way.

amCap1712:
While reviewing I also came across issue, LB-1489. Feel free to assign it yourself if you'd like to work on it next.

amCap1712 (approved):
LGTM, Thanks!

listenbrainz/tests/integration/test_playlist_api.py
```diff
@@ -212,6 +212,35 @@ def test_playlist_create_empty(self):
...
+                    }
+                }
+            }
+        
+        response = self.client.post(
+            url_for("playlist_api_v1.create_playlist"),
+            json=playlist,
+            headers={"Authorization": "Token {}".format(self.user["auth_token"])}
+        )
+        self.assert200(response)
```
amCap1712, line 242: You also need to make a call to the serialize_xspf endpoint after this, and ensure that it responds with a 200 response too.

yuktea: Thanks for pointing that out

amCap1712 (approved):
LGTM, Thanks!

PR #2740, opened 2024-01-30
Retain existing MB refresh token when user logs in and logs out

amCap1712:
There was a special case to preserve refresh tokens for MusicBrainz service for this case but when I created 3 deployments to fix another auth related bug, that special case broke. I have now fixed it and replaced it with a more general fix so hopefully it won't break again.


amCap1712:
I'll co-ordinate with bitmap to resolve the issue for the existing users for whom this has broken.

mayhem (approved):
I think this PR might also fix the metadata viewer instabilities -- would be good to release this soon. Thanks!

PR #2741, opened 2024-01-30
LB-1456: Can't change back to other user's stats after having selected global view

SatyarajRana:
# Problem

There are two pills on the right side to change between the current user's stats and global. 
When changing to global view the user stats button changes to that of the currently logged in user instead of user whose stats were viewed

    https://tickets.metabrainz.org/projects/LB/issues/LB-1456?filter=allopenissues

# Solution

There is an error in logic in the UserReports.tsx component at 'listenbrainz/listenbrainz-server/frontend/js/src/stats/UserReports.tsx', where clicking on the Global button resets the value of the user state to null, when the user clicks on the User button again, because the value of user state was reset, the stats for the current user are loaded instead.
This problem was solved by storing the value of the user to a variable called 'selectedUser', which can be used when user wants to go back to the user stats.

# Action

No further actions are required.

### DIFF
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -106,8 +106,8 @@ SQLALCHEMY_BINDS = {
    'timescale': SQLALCHEMY_TIMESCALE_URI
 }
 
-{{if service "typesense-listenbrainz"}}
-{{with index (service "typesense-listenbrainz") 0}}
+{{if service "typesense-listenbrainz-new"}}
+{{with index (service "typesense-listenbrainz-new") 0}}
 TYPESENSE_HOST = "{{.Address}}"
 TYPESENSE_PORT = {{.Port}}
 {{end}}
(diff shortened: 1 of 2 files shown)


frontend/js/src/stats/UserReports.tsx
```diff
@@ -97,7 +99,7 @@ export default class UserReports extends React.Component<
   };
 
   render() {
-    const { range, user } = this.state;
+    const { range, user, selectedUser } = this.state;
     const { apiUrl } = this.props;
```
MonkeyDo, line 103: I do not think we need this new state variable.
Instead, the issue we have is that we are using the `user` state (which we modify when clicking the buttons) instead of `user` from the props, which does not change.

```suggestion
    const { range, user } = this.state;
    const { apiUrl, user: initialUser } = this.props;
```

With this, `initialUser` is the same as what you wanted to do with the `selectedUser` state and you can use it as you did below for the button.

The `user: initialUser` syntax is to rename the variable when destructuring `this.props.user` so that the variable name does not collide with `this.state.user` declared just above.
For reference, see the "Assigning to new variable names" section in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

MonkeyDo (changes requested):
Thank you for opening this PR!

Your solution does work, well done :)
I have a suggestion to simplify it a little bit.

I had not at all realized my mistake when I implemented this, and thank you for pointing out the issue !

MonkeyDo:
P.S: we just merged yesterday a PR that refactored a lot of the front-end code, and it appears you now have a merge conflict on the PR. Let me know if you need any assistance fixing it.

PR #2742, opened 2024-01-30
Fix flask hot reload in development

amCap1712:
The latest versions of flask require the --debug flag to be passed on command line to enable hot reload of code.


mayhem (approved):
Ah ha!

PR #2743, opened 2024-01-30
LB-1452: "Missing data" link on Add data page gives 404 error when not logged in

SatyarajRana:
# Problem

When you visit the ListenBrainz page without logging in, then follow the navigation sequence "About" → "Submitting data", which opens https://listenbrainz.org/add-data/, clicking on the "Missing data" link brings you to a 404 error page with the URL https://listenbrainz.org/user//missing-data/

It looks like the username was intended to be substituted into this URL, but that's not possible when the user is not logged in.

   https://tickets.metabrainz.org/projects/LB/issues/LB-1452?filter=allopenissues

# Solution

Adding conditional rendering to the 'Missing data' button based on the availability of user data solved the issue.

# Action

No further action required

### DIFF
--- a/frontend/css/entity-pages.less
+++ b/frontend/css/entity-pages.less
@@ -198,7 +198,6 @@
 
 .review {
   position: relative;
-  margin: 2em 0;
   .text {
     max-height: 10em;
     overflow: hidden;
--- a/frontend/css/entity-pages.less
+++ b/frontend/css/entity-pages.less
@@ -198,6 +198,7 @@
 
 .review {
   position: relative;
+  margin: 2em 0;
   .text {
     max-height: 10em;
     overflow: hidden;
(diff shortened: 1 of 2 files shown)


MonkeyDo (commented):
I am very sorry for the timing of this, but we just refactored this whole section of the website as a single page application ([merged](https://github.com/metabrainz/listenbrainz-server/pull/2695) only yesterday !)
As a consequence:
1. this code and routing details are now handled on the front-end side
2. we removed the link to missing data from the about page (it didn't make much sense to have it there and accessible to non-logged-in users)

That being said, the solution you provided would indeed have worked, and thank you for taking the time to fix it !

MonkeyDo:
Unfortunately, this issue has been fixed in #2695 which was merged just yesterday.
Closing the PR consequently.

SatyarajRana:
Alright, no problem :)

PR #2745, opened 2024-01-31
Make uwsgi.ini for web containers configurable

amCap1712:
We want a different number of processes for prod/beta/test environments so need to make uwsgi.ini populated by consul.


mayhem (approved):
Cool -- glad this was easy afterall.

PR #2747, opened 2024-01-31
LB-1489: Fix content type of errors in `get_playlist_xspf` Endpoint

yuktea:
# Problem
The `/1/<playlist_mbid>/xspf` endpoint in the ListenBrainz API was returning error messages in JSON format, which is inconsistent with its standard XML response format.
# Solution
It looks like this now. (used to be a json before)
image

This change introduces a new exception class, `PlaylistAPIXMLError` ensuring error messages are formatted in XML. The `get_playlist_xspf` function has been updated to utilize the new class.

### DIFF
--- a/listenbrainz/webserver/views/playlist_api.py
+++ b/listenbrainz/webserver/views/playlist_api.py
@@ -15,7 +15,7 @@ from listenbrainz.webserver import db_conn, ts_conn
 
 from listenbrainz.webserver.utils import parse_boolean_arg
 from listenbrainz.webserver.decorators import crossdomain, api_listenstore_needed
-from listenbrainz.webserver.errors import APIBadRequest, APIInternalServerError, APINotFound, APIForbidden, APIError
+from listenbrainz.webserver.errors import APIBadRequest, APIInternalServerError, APINotFound, APIForbidden, APIError, PlaylistAPIXMLError
 from brainzutils.ratelimit import ratelimit
 from listenbrainz.webserver.views.api_tools import log_raise_400, is_valid_uuid, validate_auth_header, \
     _filter_description_html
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/views/playlist_api.py
```diff
@@ -510,20 +510,23 @@ def get_playlist_xspf(playlist_mbid):
...
-
     playlist = db_playlist.get_by_mbid(playlist_mbid, True)
+
     if playlist is None:
-        raise APINotFound("Cannot find playlist: %s" % playlist_mbid)
+        raise PlaylistAPIXMLError("Cannot find playlist: %s" % playlist_mbid, status_code=404)
+
+    try:
+        user = validate_auth_header(optional=True)
+    except Exception as e:
```
amCap1712, line 523: You can catch the exact exceptions that `validate_auth_header` raises here instead of a broad `Exception` clause.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -510,20 +510,23 @@ def get_playlist_xspf(playlist_mbid):
     """
 
     if not is_valid_uuid(playlist_mbid):
```
amCap1712, line 512: There is one more case of exceptions that needs to handled, internal server errors. I think the easiest way to go about that would be to wrap the entire code in this function inside a try/except block and raise a PlaylistsAPIXMLError exception from there. Also, before you re-raise it log the exception using `current_app.logger.error("Failed to export playlist XSPF:", exc_info=True)`.

One common way this exception can happen is database connectivity issues, which is what you can do to test this case. Mock `db_playlist.get_by_mbid` and make it raise an exception.

amCap1712 (changes requested):
Thank you for the PR! The changes and tests look good mostly. One more case to be handled and one nit and then the PR should be ready to merge.

listenbrainz/webserver/views/playlist_api.py
yuktea: Thanks for the suggestion.

yuktea:
Thanks @amCap1712 for the reviews and help with this PR. I believe it should be good to merge now. 
PS: apologies, I was away for a week or so.

amCap1712 (approved):
LGTM, Thanks!

listenbrainz/webserver/views/playlist_api.py
amCap1712: Sorry for the misleading comment, the new way is indeed the correct way and simplest way to do it.

PR #2748, opened 2024-01-31
Do not push test_request_context in API tests

amCap1712:
As a part of upcoming changes to reduce connection churn, I will refactor LB to reuse a single connection for all database queries during a request. These connections are cleaned up when teardown functions of the request are called at the end of the API request. Pushing a request/app context at the start of the request and then making multiple API requests in a test (as we currently do) does not work nicely with this pattern.

The teardown functions would never be called on the request during the running of the test, this will cause uncommitted changes to persist across API requests. So even if you forgot a commit in the server code, the corresponding test would pass.

For the majority of tests, we only need the context to use `url_for` to generate endpoint URLs. I have added a simple variant of `url_for` that can work in the absence of context. Another alternative would have been to manually write the full url in the test, it is also what the example tests in flask's official documentation. I don't have any preference between the two and went with the first one for now.


mayhem (approved):
I like this plan of reducing DB connections!

PR #2749, opened 2024-01-31
Reuse the same database connection for duration of a request

amCap1712:
Currently if we make multiple database queries in a request, a new connection is created for each query, with a few exceptions where we pass around connection objects. We should extend to that entire project and reuse connection objects during the lifetime of a request at least. Flask-SQLAlchemy follows a similar pattern by scoping its sessions to request. Connection are created upon first access i.e. if a request does not try to access a database connection, none will be created for that request. The connection created during the request will be cleaned up during teardown phase of the request.

### DIFF
--- a/listenbrainz/art/cover_art_generator.py
+++ b/listenbrainz/art/cover_art_generator.py
@@ -9,6 +9,7 @@ from data.model.common_stat import StatisticsRange
 from data.model.user_entity import EntityRecord
 
 from listenbrainz.db.cover_art import get_caa_ids_for_release_mbids
+from listenbrainz.webserver import db_conn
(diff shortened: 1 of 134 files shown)


amCap1712:
The changes in this PR are incomplete, essentially all database functions will need to be modified. I have modified one entire endpoint to test the approach out and gather feedback before modifying the entire codebase.

Also, the PR will be much easier to review with `Hide whitespaces` option turned on.

mayhem (approved):

MonkeyDo (commented):
Didn't review the code, but everything I tested on test.LB is working fine 👌

PR #2752, opened 2024-02-01
[LB-1484] Document confusing error during server setup

rimma-kubanova:
# Problem
[LB-1484] Document confusing error during server setup

# Solution
Added clarification about server setup


rimma-kubanova:
So sorry, just noticed that someone have already fixed the issue

PR #2756, opened 2024-02-03
LB-1490: Review previews don't respect markdown formatting

rimma-kubanova:
# Problem
[LB-1490](https://tickets.metabrainz.org/browse/LB-1490)

In ListenBrainz album and artist pages, the preview of CB reviews doesn't display markdown formatting, just plain text, leading to unreadable reviews like:
"The Lord is back with his second proper album. [Silver Dragons](https://musicbrainz.org/recording/43df1fdd-55ed-443b-ab32-f8aac948abc1) and [Moonrise Over Isengard](https://musicbrainz.org/recording/3d440f1d-049e-469d-814e-210c267f..."
We should not display all of the markdown formatting, in particular large titles/headers. We could strip it all out, but it might be nice to keep some in, like italics and bold.

# Solution

- added remove-markdown package that removes all markdown formatting

image


MonkeyDo:
In my opinion, it would be nice to keep the Markdown formatting, and even in the near future enable it for some other components (for example the one that allows writing CritiqueBrainz reviews from ListenBrainz).

Instead of stripping out all the markdown formatting, I would suggest using a library such as [react-markdown](https://remarkjs.github.io/react-markdown/) or similar and specifically disallow or replace some elements like the headings as the ticket suggests (for example maybe with [this plugin](https://github.com/mrzmmr/remark-behead)?)

rimma-kubanova:
Oh sure, I will look for it. I will keep the pull request open and push new changes (will probably work on disabling headings then)

MonkeyDo:
@rimma-kubanova regarding the module import issue I see in the automated tests, I'm not exactly sure what is happening but you might want to try this import style:
`import * as Markdown from 'react-markdown'`  <- note the `* as`

rimma-kubanova:
Update: i almost solved the issue, but now code is not working for some other packages (pls ignore all of this commits T_T)

rimma-kubanova:
OMG I DID IT

MonkeyDo (approved):
Nice work @rimma-kubanova !

I had to struggle with how to import this package too myself. After running the PR and loading the page it crashed because a sub-dependency expected a nodeJS `process` to be available (which is not the case in the browser).

Took a bit of reading issues adn trying a few things but I too got there in the end :)

Anyway, it's all working fine now and the styled output is quite nice, well done !

rimma-kubanova:
Thank you very much! You gave me a very good guidance,  appreciate your help!

PR #2761, opened 2024-02-06
LB-1462: Listen controls dropdown is too narrow / hard to click

rimma-kubanova:
# Problem

[LB-1462](https://tickets.metabrainz.org/browse/LB-1462)

"it's the three vertical dots to the right of a listen) **difficult to click on**, because it seems to be only 5 pixels wide, so I always have to hesitate ever so slightly to line up my cursor with it precisely."

# Solution

I propose the following solution:

- changing the dropdown menu icon to a wider one to make it easier to click on
- increasing the size of the icons for heart, heart-crack, and dropdown-menu
- also improved the gap and padding between the icons, so the play button has a space on the right side

# Before
Screenshot 2024-02-06 at 11 19 07

# After
Screenshot 2024-02-06 at 11 17 45

### DIFF
--- a/frontend/css/listens-page.less
+++ b/frontend/css/listens-page.less
@@ -414,7 +414,7 @@
 
       .fa-heart,
       .fa-heart-crack,
-      .fa-bars{
+      .fa-bars {
         font-size: 21px;
         margin: 0% 7%;
         width: 1em;
--- a/frontend/css/listens-page.less
+++ b/frontend/css/listens-page.less
@@ -246,7 +246,7 @@
       .right-section {
         flex: 0;
         .listen-controls {
-          flex-grow: 0;
+          // flex-grow: 0;
           margin: unset;
         }
       }
@@ -386,6 +386,7 @@
 
       .listen-controls {
         margin-left: 1.5em;
+        margin-right: 1em;
       }
       > *:not(.dropdown-menu) {
         min-width: 1.5em;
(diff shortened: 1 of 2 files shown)


redactedscribe:
I'm in favour of widening the clickable area of the dropdown, but perhaps all of the UI ellipses should be modified in the same way (e.g. the BrainzPlayer's ellipsis -- bottom-right) to remain consistent?

I've been using this 100% hacky CSS modification script (it only achieves a doubling of width, but it's better than the default at least):

```css
#listenControlsDropdown.svg-inline--fa.fa-ellipsis-vertical,
#brainz-player svg.svg-inline--fa.fa-ellipsis-vertical {
  stroke: rgb(255, 0, 0, 0); /* Debug: Set alpha to 1. */
  stroke-width: 12.5em; /* Fix: Add clickable width to ellipsis. */
}
```

Aerozol:
Great! My mockups here are finished btw: https://tickets.metabrainz.org/browse/LB-991

I think it is up to you if you want to tackle some of the bigger changes there - regardless though, I think this probably solves LB-991 (@MonkeyDo), and I could then move the more complicated stuff (pinning options etc) into a new ticket.

rimma-kubanova:
@Aerozol  oh great! I love the idea of leaving just the menu, and just letting users customize with pins if needed. I have a question tho, will the same changes will be applied to the desktop/wider screen versions?

Esp this ones:
mockup alternate

mockup-01a

Aerozol:
Yup, all changes should be applied to mobile and desktop sizes.

I was initially imagining the drop down menu for desktop, and the side 'drawer' for mobile. But someone made a good comment on LB-991 about the drawer causing a lot of 'reaching' - where the drop down would have all the links in close reach right after tapping the menu icon.

So potentially the drop down for both, but it's not set in stone.

For wide screens the drop down should probably be right aligned and have its width limited - sorry, I'm very time poor so my mockups don't always capture every use-case, like this one. Let me know any time if you'd like something expanded upon!

rimma-kubanova:
sure no problem! I will try to come up with some mockups for desktop version and will let you know in IRC

MonkeyDo:
I ended up refactoring the buttons a bit in #2806 which is now merged, and should be used as a base for any continuation.

From the last discussion on this PR, I don't think the proposed changes really answer the issue, and we rather want to implement one of the solutions presented in [LB-991](https://tickets.metabrainz.org/browse/LB-991)

One way or another, I don't think we can build from this PR in its current state, and would probably require starting from scratch with one of the designs from the ticket.

Don't hesitate if you have questions or if I can help with something !

PR #2763, opened 2024-02-06
Fix export listens button on delete pages

amCap1712:
The Export Listens button on delete pages was incorrectly calling the delete listens/account endpoint.


mayhem (commented):
Should these downloadFile() have the same argument?

amCap1712:
@mayhem Yes, according to the current code. That should be the case. It is possible to simplify this is a bit or make it more generic for use elsewhere but I didn't want to make more modifications than necessary. I leave that to @MonkeyDo .

PR #2764, opened 2024-02-07
LB-1494: Music player bar in BrainzPlayer have to move smoothly

rimma-kubanova:
# Problem

[LB-1494](https://tickets.metabrainz.org/browse/LB-1494)

- The music progress bar moving every 1-2 seconds was caused by rounding. The progress bar was getting integer values from 0 to 100, so for a few seconds it had the same value and didn't move (this is fine, but looks a bit weird on wider screens).

# Solution

- just fixed it, so the value rounds to 2 decimal places 
- now the music player bar moves smoothly!

# Demo

https://github.com/metabrainz/listenbrainz-server/assets/115300909/e6628048-edea-4c45-8e0c-61e6a3e2d0d7

### DIFF
--- a/frontend/js/src/common/brainzplayer/ProgressBar.tsx
+++ b/frontend/js/src/common/brainzplayer/ProgressBar.tsx
@@ -25,7 +25,7 @@ const TOOLTIP_TOP_OFFSET: number = 102;
 function ProgressBar(props: ProgressBarProps) {
   const { durationMs, progressMs, seekToPositionMs } = props;
   const [tipContent, setTipContent] = React.useState(TOOLTIP_INITIAL_CONTENT);
-  const progressPercentage = Math.round((Number(progressMs * 100) / durationMs) * 100) / 100;
+  const progressPercentage = Number(progressMs * 100 / durationMs).toFixed(2);
   const hideProgressBar = isNaN(progressPercentage) || progressPercentage <= 0;
 
   // Originally by ford04 - https://stackoverflow.com/a/62017005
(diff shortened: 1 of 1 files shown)


frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -25,9 +25,7 @@ const TOOLTIP_TOP_OFFSET: number = 102;
 function ProgressBar(props: ProgressBarProps) {
   const { durationMs, progressMs, seekToPositionMs } = props;
   const [tipContent, setTipContent] = React.useState(TOOLTIP_INITIAL_CONTENT);
-  const progressPercentage = Math.round(
-    Number((progressMs * 100) / durationMs)
-  );
+  const progressPercentage = Math.round((Number(progressMs * 100) / durationMs) * 100) / 100;
```
MonkeyDo, line 28: That seems a little bit of a roundabout way to round up with more precision; try this :)

```suggestion
  const progressPercentage = Number(progressMs * 100 / durationMs).toFixed(2);
```

MonkeyDo (commented):
Thanks for opening this PR!
I hadn't realized it was the rounding that was causing this issue, thanks for that !

I have a code suggestion below to make it slightly more readable (in my opinion at least).

rimma-kubanova:
Just checked the code, it works, thanks for suggestion !

rimma-kubanova:
Changed the code a little, since toFixed() returns strings, while we need a number variable. Cool, now it works ! Thanks for suggestion !

MonkeyDo (approved):
Perfect, thank you!

PR #2765, opened 2024-02-07
LB-1456: Can't change back to other user's stats after having selecte…

SatyarajRana:
# Problem

There are two pills on the right side to change between the current user's stats and global.
When changing to global view the user stats button changes to that of the currently logged in user instead of user whose stats were viewed

   https://tickets.metabrainz.org/projects/LB/issues/LB-1456?filter=allopenissues

# Solution

Implemented @MonkeyDo 's suggestion and simplified the solution by using user as initialUser from the props which does not change on clicking on the Global button

# Action

No further action needed


MonkeyDo (approved):
Looks great, thank you for resubmitting the PR !

PR #2772, opened 2024-02-12
LB-1496: shorten time before removal when closing modals

MonkeyDo:
We can't currently use the built-in Bootstrap modals from NiceModal since we are using an old version of Bootstrap (v3).

This means that we had to manually remove the modal components from the application tree after closing them.
This requires a delay to allow the fade animation to finish.
Previously this was set way high (3 seconds), but this crated issues if you close a modal and straight away open another one: the removal timer kicks in and removes that newly opened modal.

Until we can use the built-in Bootstrap modals that already take all this into account, I set the timeout to 200ms (from the previous 3000ms) which should be plenty as the fade animation duration is 150ms.

### DIFF
--- a/frontend/js/src/cb-review/CBReviewModal.tsx
+++ b/frontend/js/src/cb-review/CBReviewModal.tsx
@@ -43,7 +43,7 @@ export default NiceModal.create(({ listen }: CBReviewModalProps) => {
 
   const closeModal = React.useCallback(() => {
     modal.hide();
-    setTimeout(modal.remove, 3000);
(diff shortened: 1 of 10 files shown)


MonkeyDo:
@07jasjeet This PR should fix the issue you reported with the modals closing on you.
FWIW the issue was with opening a new modal less than three seconds after closing another modal, because that previous modal had a delayed call to remove the modal from the DOM tree.

anshg1214 (approved):
LGTM

PR #2775, opened 2024-02-13
LB-1506: log_raise_400 type hint added

nikitashrivastava29:
# Problem
We do capture or log 400 (BadRequest) client errors by default. However, sometimes we wish to do that. Hence, we have a helper function called [log_raise_400](https://github.com/metabrainz/listenbrainz-server/blob/265ada1c2047648a54737bcb6a9aeb3f0c019a05/listenbrainz/webserver/views/api_tools.py#L251) which logs an error message and then raises a 400 error. Note that this function always raises and exception, however pycharm or other type checkers cannot detect this and give lots of false positives like.

# Solution

Added a type hint to the log_raise_400 function (NoReturn) so that these false positive warnings can be eliminated.

# Action

### DIFF
--- a/listenbrainz/webserver/views/api_tools.py
+++ b/listenbrainz/webserver/views/api_tools.py
@@ -13,6 +13,8 @@ import orjson
 import uuid
 import sentry_sdk
 
+from typing import NoReturn
+
 from flask import current_app, request
 
 from listenbrainz.listenstore import LISTEN_MINIMUM_TS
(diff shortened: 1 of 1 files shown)


mayhem:
Please give this PR a description to let us know what this PR does. Thanks!

nikitashrivastava29:
Hey, edited the PR with details of the problem and the solution

amCap1712 (approved):
LGTM, Thanks!

PR #2781, opened 2024-02-14
LB-1143 : Import playlist manually on the front-end

SatyarajRana:
# New Feature

Playlists can be imported by uploading a JSON or JSPF file

# Solution

A new Card component was added in the Playlists.tsx file which when clicked opens the ImportPlaylistModal. This modal lets the user upload a JSON or JSPF file. Extension checks and validation of JSON is done on the frontend, but further validation will be done on the backend.

# Action

Please check if any change needs to be made regarding the design of the Import playlist button and the ImportPlaylistModal

### DIFF
--- a/frontend/js/src/playlists/components/ImportPlaylistModal.tsx
+++ b/frontend/js/src/playlists/components/ImportPlaylistModal.tsx
@@ -1,0 +1,204 @@
+import * as React from "react";
+import NiceModal, { useModal } from "@ebay/nice-modal-react";
+import { toast } from "react-toastify";
+
+import GlobalAppContext from "../../utils/GlobalAppContext";
+import { ToastMsg } from "../../notifications/Notifications";
+
+type ImportPlaylistModalProps = {
+  playlist?: JSPFPlaylist;
+  initialTracks?: JSPFTrack[];
+};
+
+export default NiceModal.create((props: ImportPlaylistModalProps) => {
+  const modal = useModal();
+  const closeModal = React.useCallback(() => {
+    modal.hide();
+    setTimeout(modal.remove, 3000);
+  }, [modal]);
+
+  const { currentUser, APIService } = React.useContext(GlobalAppContext);
+  const { playlist, initialTracks } = props;
+
+  const [fileError, setFileError] = React.useState<string | null>(null);
+  const [newPlaylist, setNewPlaylist] = React.useState<JSPFObject | null>(null);
+
+  const createPlaylist = React.useCallback(async (): Promise<
+    JSPFPlaylist | undefined
+  > => {
+    // Creating a new playlist
+    if (!currentUser?.auth_token) {
+      toast.error(
+        <ToastMsg
+          title="Error"
+          message="You must be logged in for this operation"
+        />,
+        { toastId: "auth-error" }
+      );
+      return undefined;
+    }
+    try {
+      if (newPlaylist) {
+        const newPlaylistId = await APIService.createPlaylist(
+          currentUser?.auth_token,
+          newPlaylist
+        );
+
+        toast.success(
+          <ToastMsg
+            title="Created playlist"
+            message={
+              <>
+                Created a new playlist with ID:
+                <a href={`/playlist/${newPlaylistId}`}>{newPlaylistId}</a>
+              </>
+            }
+          />,
+          { toastId: "create-playlist-success" }
+        );
+        try {
+          // Fetch the newly created playlist and return it
+          const response = await APIService.getPlaylist(
+            newPlaylistId,
+            currentUser.auth_token
+          );
+          const JSPFObject: JSPFObject = await response.json();
+          return JSPFObject.playlist;
+        } catch (error) {
+          console.error(error);
+          return newPlaylist.playlist;
+        }
+      } else {
+        return undefined;
+      }
+    } catch (error) {
+      toast.error(
+        <ToastMsg
+          title="Could not create playlist"
+          message={`Something went wrong: ${error.toString()}`}
+        />,
+        { toastId: "create-playlist-error" }
+      );
+      return undefined;
+    }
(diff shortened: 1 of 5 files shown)


frontend/js/src/playlists/components/ImportPlaylistModal.tsx
```diff
@@ -0,0 +1,204 @@
+import * as React from "react";
+import NiceModal, { useModal } from "@ebay/nice-modal-react";
+import { toast } from "react-toastify";
+
+import GlobalAppContext from "../../utils/GlobalAppContext";
+import { ToastMsg } from "../../notifications/Notifications";
+
+type ImportPlaylistModalProps = {
+  playlist?: JSPFPlaylist;
+  initialTracks?: JSPFTrack[];
```
MonkeyDo, line 10: This is just copy pasted from another modal, both of these props are not needed.
Will also need to be removed from line 21.

frontend/js/src/playlists/components/ImportPlaylistModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+};
+
+export default NiceModal.create((props: ImportPlaylistModalProps) => {
```
MonkeyDo, line 13: I think we don't need any props for this component, you can remove the ImportPlaylistModalProps type above completely.
```suggestion
export default NiceModal.create() => {
```

frontend/js/src/user/playlists/Playlists.tsx
```diff
@@ -175,8 +178,26 @@ export default class UserPlaylists extends React.Component<
...
-          )}
+            </Card>,
+            <Card
+              key="import-playlist"
+              className="import-playlist"
+              data-toggle="modal"
+              data-target="#ImportPlaylistModal"
+              onClick={() => {
+                NiceModal.show(ImportPlaylistModal)
+                  // @ts-ignore
```
MonkeyDo, line 189: Making a note here, which can definitely be pushed back for a future refactor if you don't want to touch it:
We can improve the types here and avoid forcefully ignoring the typescript error like so:

```tsx
NiceModal.show<
  JSPFPlaylist, // <- this is the expected return type
  NiceModalHocProps
>(ImportPlaylistModal)
.then((playlist) => { // note the removed type here, no longer needed
  this.onPlaylistCreated(playlist); // <- playlist here has the expected type of JSPFPlaylist that we set above
});
```

frontend/js/src/playlists/components/ImportPlaylistModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+  const closeModal = React.useCallback(() => {
+    modal.hide();
+    setTimeout(modal.remove, 3000);
+  }, [modal]);
+
+  const { currentUser, APIService } = React.useContext(GlobalAppContext);
+  const { playlist, initialTracks } = props;
+
+  const [fileError, setFileError] = React.useState<string | null>(null);
+  const [newPlaylist, setNewPlaylist] = React.useState<JSPFObject | null>(null);
```
MonkeyDo, line 24: I think the names here are a bit confusing, it took me a while to understand that you're not setting the newly created playlist in the state, but rather the content of the uploaded file if I understood correctly.
Maybe `fileContent, setFileContent` or something like that?

frontend/js/src/playlists/components/ImportPlaylistModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+      }
+      // Check if the file content is in valid JSON format and set the new playlist state
+      checkFileContent(selectedFile);
+    }
+  };
+
+  const onSubmit = async (event: React.SyntheticEvent) => {
+    try {
+      createPlaylist();
+      modal.resolve(newPlaylist);
```
MonkeyDo, line 129: Here you're returning the uploaded JSON directly rather than the actual playlist that we fetch from the server on lines 61-65.
I suppose `createPlaylist` should really return the playlist fetched from the server, and you can use that here instead of the uploaded file.
```
const newPlaylist = await createPlaylist();
```
^currently `newPlaylist` is taken but I commented above about renaming it, so it should be free after that

MonkeyDo (changes requested):
Thanks a lot for working on this!
PR looks very good, although I do have some suggestions for small improvements in the wording, and a couple of small code ones too.

SatyarajRana:
@MonkeyDo Sure thing, I'll implement all these changes by tomorrow.

MonkeyDo:
> @MonkeyDo Sure thing, I'll implement all these changes by tomorrow.

No rush :)

frontend/js/src/playlists/components/ImportPlaylistModal.tsx
```diff
@@ -0,0 +1,204 @@
...
+          currentUser?.auth_token,
+          newPlaylist
+        );
+
+        toast.success(
+          <ToastMsg
+            title="Created playlist"
+            message={
+              <>
+                Created a new playlist with ID:
```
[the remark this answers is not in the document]
SatyarajRana: This will add a delay (of around 2-3s) in the display of the toast message, will that be okay?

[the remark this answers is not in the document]
MonkeyDo: Ah, I see that could be an issue.
In that case, I would suggest adding a state variable `isLoading` or something like that and setting it to true right before the try…catch block.
Then we can set it to false both if the process succeeds and also if it fails.

Then in the page markup, you can add a loading indicator and make it visible when `isLoading` is true like so:
https://github.com/metabrainz/listenbrainz-server/blob/39095293509d82a03581d1795d175e404c0ab02a/frontend/js/src/explore/lb-radio/LBRadio.tsx#L115-L119

SatyarajRana:
@MonkeyDo Hey, I'm having some issues with my laptop, I'll finish this as soon as it gets fixed

PR #2782, opened 2024-02-14
Add missing indexes to popularity data

mayhem:
The popularity data tables had no indexes, so this PR adds them.


mayhem:
@pep8speaks thats fine, no need to be overly pedantic, thanks.

amCap1712 (approved):
I noticed a couple of issues with the index name format so I fixed that but otherwise LGTM, Thanks!

mayhem:
AH, thanks for fixing.

PR #2787, opened 2024-02-15
Homepage: Fix number counter

MonkeyDo:
The homepage is showing the number of listens and artists in reverse order for some reason, possibly from a previous attempt at the implementation.
Removed a couple of lines that reverse the number, and we're good to go.

### DIFF
--- a/frontend/js/src/home/NumberCounter.tsx
+++ b/frontend/js/src/home/NumberCounter.tsx
@@ -6,7 +6,7 @@ type NumberCounterProps = {
 
 function NumberCounter({ count }: NumberCounterProps) {
   const countString = count?.toString();
-  const groupsOfThree = countString?.match(/\d{1,3}/g) ?? [];
+  const groupsOfThree = countString?.match(/(\d+?)(?=(\d{3})+(?!\d)|$)/g) ?? [];
 
   return (
     <div className="number-count">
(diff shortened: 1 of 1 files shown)


mayhem (approved):
So happy we're not missing tons of listens!

frontend/js/src/home/NumberCounter.tsx
```diff
@@ -6,13 +6,11 @@ type NumberCounterProps = {
 
 function NumberCounter({ count }: NumberCounterProps) {
   const countString = count?.toString();
-  const reversedStr = countString?.split("").reverse().join("");
-  const groupsOfThree = reversedStr?.match(/\d{1,3}/g);
-  const reversedGroupsOfCount = groupsOfThree ? groupsOfThree.reverse() : [];
+  const groupsOfThree = countString?.match(/\d{1,3}/g) ?? [];
```
anshg1214, line 9: This does not work well if you have the count = `1234567890`. It breaks it into `['123', '456', '789', '0']`. 

Instead you can try out this new regex which breaks it into `['1', '234', '567', '890']`

```suggestion
  const groupsOfThree = countString?.match(/(\d+?)(?=(\d{3})+(?!\d)|$)/g) ?? [];
```

anshg1214 (changes requested):

PR #2788, opened 2024-02-19
Avoid exhausting resources in metadata api lookup

amCap1712:
1. Restrict query length in metadata api lookups

    Typesense performance will degrade the query is huge, for instance a 1000 character query can take 15-20s for the request to complete. This also keeps the worker and other resources occupied making it easy to run out of resources to serve other requests. We limit the length of the lookup being indexed in the database to 500 characters, so it makes sense to apply at the least the same limit on the metadata API lookup.

2. Do not retry on typesense search timeout in metadata API

    The timeout for typesense requests is pretty generous IMO, it might make sense to still retry it for the mapper which runs as a background process but for the API we should probably just fail the request.

### DIFF
--- a/listenbrainz/webserver/views/metadata_api.py
+++ b/listenbrainz/webserver/views/metadata_api.py
@@ -236,7 +236,7 @@ def get_mbid_mapping():
         if exact_results:
             return process_results(exact_results[0], metadata, incs)
 
-        q = MBIDMapper(timeout=10, remove_stop_words=True, debug=False)
+        q = MBIDMapper(timeout=10, remove_stop_words=True, debug=False, retry_on_timeout=False)
         fuzzy_result = q.search(artist_name, recording_name, release_name)
         if fuzzy_result:
             return process_results(fuzzy_result, metadata, incs)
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/metadata_api.py
```diff
@@ -16,6 +16,9 @@
 
 metadata_bp = Blueprint('metadata', __name__)
 
+#: The maximum length of the query permitted for a mapping search
+MAX_MAPPING_QUERY_LENGTH = 500
```
mayhem, line 20: That seems rather permissive still. Do you have an idea of what a typical query length is?

mayhem (approved):

listenbrainz/webserver/views/metadata_api.py
amCap1712: The typical query length is likely less than a 100 characters, 95%+ of the data in that range. There are about 600k recordings in the 100-250 characters range. I think I'll lower the limit to 250, there are just 10-15k recordings out of 24 million there.

amCap1712: The typical query length is likely less than a 100 characters, 95%+ of the data in that range. There are about 600k recordings in the 100-250 characters range. I think I'll lower the limit to 250, there are just 10-15k recordings out of 24 million there.

PR #2789, opened 2024-02-19
Update add-data.html with YTMusic2listenbrainz script

Aerozol:
Add new section, Scripts, and a link to YTMusic2listenbrainz.py (Python script to submit your YouTube Music watch history to Listenbrainz)


Aerozol:
Note: I'm not sure if we want to add scripts to this page? But this seems useful enough to at least consider for inclusion.

mayhem (approved):
I think this is fine for now -- thanks!

PR #2791, opened 2024-02-20
LB-1516: Prevent None output rows in top artist recordings

mayhem:
LB-1516 shows that sometimes we have rows with all None values in the output. The popularity data does not have these bad rows from what I can see, but if the matching recording has been deleted from the database, then the ancillary data we load for the recordings will have None values.


amCap1712 (approved):
LGTM, Thanks!

PR #2793, opened 2024-02-20
Add a threshold to fresh releases calculation

amCap1712:
We want to generate the customized user feed of fresh releases using the user's entire listening history, therefore it makes sense to have a threshold parameter to reduce the noise of artists listened very few times.


mayhem (approved):
Thanks!

PR #2795, opened 2024-02-24
 Separate popularity calculation for MLHD and listens 

amCap1712:
The MLHD dataset is static so we only need to compute it once (occasionally), whereas the listens dataset is regularly updated and hence the popularity data based off it needs to be regenerated regularly too. Separating the two reduces redundant calculation in Spark cluster every time we update popularity data, we get regular spark alerts when we process MLHD popularity so it would be helpful somewhat.


mayhem (approved):
Great

PR #2797, opened 2024-02-26
Disable VA artist entity page

amCap1712:
VA artist has 200k+ release groups. Trying to load popularity data for the same causes massive load on gaga/LB.


mayhem (approved):
thanks!

PR #2799, opened 2024-02-27
LB-1230: Added ability to edit comment after pinning a Listen

lzzy12:
# Problem

Fixes this Ticket: 
https://tickets.metabrainz.org/browse/LB-1230

# Solution

Added an extra action for pinned listen: 
image


frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -111,7 +118,44 @@ export default NiceModal.create(
...
+      async (event: React.MouseEvent<HTMLButtonElement>) => {
+        event.preventDefault();
+        try {
+          if (rowId && recordingToPin && currentUser?.auth_token) {
+            const response = await APIService.updatePinRecordingBlurbContent(
+              currentUser.auth_token,
+              rowId,
+              blurbContent
+            );
+            if (!response.status) {
```
MonkeyDo, line 131: `status` should be an integer of the HTML status code (200, 404, etc) so this condition will always be true.
If the API call fails it will throw an error (see the implementation of `checkStatus` in the APIService file)

frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -111,7 +118,44 @@ export default NiceModal.create(
...
+              toast.error(
+                <ToastMsg
+                  title="Comment update failed"
+                  message="Something went wrong"
```
MonkeyDo, line 135: It would be nice to show the user the actual error if the server returns one, to help with issue reports.

frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -111,7 +118,44 @@ export default NiceModal.create(
...
+                />,
+                {
+                  toastId: "pin-update-failed",
+                }
+              );
+            } else {
+              toast.success(
+                <ToastMsg
+                  title="Comment updated"
+                  message={`Comment has been updated: ${blurbContent}`}
```
MonkeyDo, line 145: This content could potentially be very long, so I don't think we want to add it into the toast here.
I think we could even drop the message here and let the title do all the work.

frontend/js/src/utils/APIService.ts
```diff
@@ -1003,6 +1003,26 @@ export default class APIService {
     return response.json();
   };
 
+  updatePinRecordingBlurbContent = async (
+    userToken: string,
+    rowId: number,
+    blurbContent: string
+  ): Promise<{ status: string }> => {
```
MonkeyDo, line 1010: See comment below about the return value of the API endpoint.

frontend/js/src/user/components/PinnedRecordingCard.tsx
```diff
@@ -148,6 +151,22 @@ export default class PinnedRecordingCard extends React.Component<
...
       );
+      additionalMenuItems.push(
+        <ListenControl
+          text="Edit Comment"
+          key="Edit Comment"
+          icon={faPencilAlt}
+          action={() => {
+            NiceModal.show(PinRecordingModal, {
+              recordingToPin: listen,
+              isUpdate: true,
```
MonkeyDo, line 162: I wonder if `isUpdate` is actually required here.
Wouldn't the presence of the `rowId` be enough to determine it?
Something like this in PinRecordingModal.tsx:
`const isUpdate = Boolean(rowId)`

MonkeyDo (changes requested):
Thanks for opening a PR! It's looking quite good !

I did a first round of review without running the code, but it's looking functional already.
Some questions about return value and statuses below:

lzzy12:
Hi, thanks foor your feedbacks. I will work on them

frontend/js/src/utils/APIService.ts
lzzy12: Hi, could you please elaborate the issue? I don't see a comment related to return value of API endpoint

frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -28,8 +29,13 @@ export const maxBlurbContentLength = 280;
...
 
 export default NiceModal.create(
-  ({ recordingToPin, onSuccessfulPin }: PinRecordingModalProps) => {
+  ({
+    recordingToPin,
+    onSuccessfulPin,
+    rowId,
+  }: PinRecordingModalProps) => {
     const modal = useModal();
+    const isUpdate = Boolean(rowId == null);
```
MonkeyDo, line 38: I think this should be sufficient, since all numbers apart from 0 will return true:
```suggestion
    const isUpdate = Boolean(rowId);
```

In comparison, even now the code seems unreliable, as `rowId` should eitgher be a number or undefined, but not null.
If for some reason in the future rowId ends up being undefined instead of null or vice versa, the check could simply break.

I'd also point out more generally that `==` and `===` don't mean the same thing in javascript; you would probably want "strict equality" here

frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -188,10 +217,12 @@ export default NiceModal.create(
...
                 type="submit"
                 className="btn btn-success"
-                onClick={submitPinRecording}
+                onClick={
+                  isUpdate ? updatePinnedRecordingComment : submitPinRecording
+                }
                 data-dismiss="modal"
               >
-                Pin track
+                {isUpdate ? "Update Comment" : "Pin track"}
```
MonkeyDo, line 225:
```suggestion
                {isUpdate ? "Update comment" : "Pin track"}
```

frontend/js/src/utils/APIService.ts
MonkeyDo: My bad! I don't know what I did but that was probably left in-between some comments I modified

MonkeyDo (changes requested):
Thanks for the modifications, I have a couple more comments but it's looking nice !

frontend/js/src/pins/PinRecordingModal.tsx
lzzy12: Also it seems the `==` work with undefined as well as null, mentioned here: https://stackoverflow.com/a/2647888

frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -14,6 +14,7 @@ import { ToastMsg } from "../notifications/Notifications";
 export type PinRecordingModalProps = {
   recordingToPin: Listen;
   onSuccessfulPin?: (pinnedrecording: PinnedRecording) => void;
+  rowId?: number;
 };
```
MonkeyDo, line 18: new `initialBlurbContent?: string;`  prop required here

frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -28,8 +29,13 @@ export const maxBlurbContentLength = 280;
...
 export default NiceModal.create(
-  ({ recordingToPin, onSuccessfulPin }: PinRecordingModalProps) => {
+  ({
+    recordingToPin,
+    onSuccessfulPin,
+    rowId,
+  }: PinRecordingModalProps) => {
     const modal = useModal();
+    const isUpdate = Boolean(rowId);
     const [blurbContent, setBlurbContent] = React.useState("");
```
MonkeyDo, line 39: Set initial blurb content here, fallback to the empty string as before:
```suggestion
    const [blurbContent, setBlurbContent] = React.useState(initialBlurbContent ?? "");
```

With the other related changes that should mean you have the existing content in the textarea when you open the modal

frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -111,7 +117,30 @@ export default NiceModal.create(
...
+              rowId,
+              blurbContent
+            );
+            toast.success(
+              <ToastMsg
+                title="Comment updated" message=''/>,
+              {
+                toastId: "pin-update-success",
+              }
+            );
```
MonkeyDo, line 136: Here we can have the modal return the updated text so we can update the UI accordingly:
```suggestion
            );
            modal.resolve(blurbContent);
```

I'll leave a comment in PinnedRecordingCard for how to use the returned value.

I don't know if we need to return anything when there is an error, in the catch block below.

frontend/js/src/user/components/PinnedRecordingCard.tsx
```diff
@@ -148,6 +151,21 @@ export default class PinnedRecordingCard extends React.Component<
...
+      additionalMenuItems.push(
+        <ListenControl
+          text="Edit Comment"
+          key="Edit Comment"
+          icon={faPencilAlt}
+          action={() => {
+            NiceModal.show(PinRecordingModal, {
+              recordingToPin: listen,
+              rowId: pinnedRecording.row_id,
+            });
```
MonkeyDo, line 163: There is a good example in the codebase of how to use that return value in the PinnedRecordingCard component: 
https://github.com/metabrainz/listenbrainz-server/blob/c8b5abf56ba77b3b1c81465b01e51b0e144ac6e1/frontend/js/src/user/playlists/Playlists.tsx#L171-L175

That also means you will have to store `pinnedRecording` in the component state of PinnedRecordingCard rather than using it from the props.

You'll need a new class method `updatePinnedRecording` or something like that where you update the state accordingly, similar to what we do here:
https://github.com/metabrainz/listenbrainz-server/blob/c8b5abf56ba77b3b1c81465b01e51b0e144ac6e1/frontend/js/src/user/playlists/Playlists.tsx#L96-L101

Except since pinnedRecording is an object we want to merge properties, something like:
```
    this.setState( currentState => ({
      pinnedRecording: {
        ...currentState.pinnedRecording,
        blurb_content: updatedBlurbContent},
    }));
```

MonkeyDo (changes requested):
Thanks for those changes, the code looks great now !

I deployed the PR to our test server, and it's working nicely! I was able to update a comment without any issue, nice work !

Through testing I do have two more pieces of feedback from a user perspective. Sorry about not realizing sooner !
- It would be useful to have the current pin comment in the textarea when I open the modal to edit the pin. Currently to update a typo I would have to copy the comment first. Left some comments about that below
- I would want the UI to update so that the new text content is shown on the pin card on the page. Also left some comments to that effect, but basically we can consider the call `modal.show(` as a promise and make the modal return a value when the API call succeeds (in this case the updated text).

lzzy12:
Actually the current behavior was inlined to what happens when we pin a track. It only shows up in pin when the user refreshes. I have observed this behavior at multiple places in the webapp. That was the reason I tried to keep it inlined to that behavior. 
I will work on it implement the suggested changes

MonkeyDo:
> Actually the current behavior was inlined to what happens when we pin a track. It only shows up in pin when the user refreshes. I have observed this behavior at multiple places in the webapp. That was the reason I tried to keep it inlined to that behavior.
> I will work on it to implement the suggested changes

Yes, I didn't want to muddy the waters by mentioning it, but I did realize the pin is not updated on the page when pinning a track, which is definitely not great UX :)

I think since you are working on that feature, and since you noticed it yourself, you will be in a great position to fix that too in your next PR, if you are interested in doing that.

MonkeyDo:
Hi @lzzy12 !
Just checking in to see if you have some time to finish this PR, or if you would rather pass the baton to someone else.
Either way is fine by me, please let me know :)

lzzy12:
Sorry got caught up with something. Will be finishing it this weekend

MonkeyDo (approved):

MonkeyDo:
I took the liberty to finish this off, hope you don't mind :)

PR #2804, opened 2024-03-02
Fix similarity datasets message name

amCap1712:
Spark reader expects the names to start with similarity instead of similar.


amCap1712:
Also, I just found out that the development version of similarity data is broken. Do we still need it? If so, I can fix it otherwise I'll just remove it and make labs api point to the production version as well.

mayhem:
I think it is ok to not fix the dev data right now -- we're not in a phase to be working on this data, but once we work on recording data, we'll probably want it back for testing.

mayhem (approved):

PR #2805, opened 2024-03-06
LB-1143: Import playlist manually on the front-end

SatyarajRana:
See the previous PR for this feature #2781 for previous feedback

# New Feature

Playlists can be imported by uploading a JSON or JSPF file

# Solution

A new Card component was added in the Playlists.tsx file which when clicked opens the ImportPlaylistModal. This modal lets the user upload a JSON or JSPF file. Extension checks and validation of JSON is done on the frontend, but further validation will be done on the backend.

# Action

Please check if any change needs to be made regarding the design of the Import playlist button and the ImportPlaylistModal

### DIFF
--- a/frontend/css/playlists.less
+++ b/frontend/css/playlists.less
@@ -71,7 +71,8 @@
   }
 }
 
-.new-playlist {
+.new-playlist,
+.import-playlist {
   margin: 0.5em;
   width: 15em;
   cursor: pointer;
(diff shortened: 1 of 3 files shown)


SatyarajRana:
@MonkeyDo Please review this PR. I made all the suggested changes

MonkeyDo (changes requested):
Good job! it works really nicely, and the code is clean.

I did find an issue with error catching. To replicate, create a json file with names of collaborators that don't exist (example here: https://pastebin.com/cffX69NX).
You'll see that when you submit the playlist the back-end throws an error (which is expected), but the error isn't caught and crashes the page:
image
This part can be improved, but other than that I think it's working as intended, nice job!

PR #2807, opened 2024-03-11
LB-476: Use https on demos when endpoint is `api.listenbrainz.org`

ykchong45:
# Problem

[Ticket LB-476](https://tickets.metabrainz.org/browse/LB-476): Use https on demos when endpoint is `api.listenbrainz.org`.

# Solution

1. add a boolean flag `DEBUG` to indicate whether to call API on the local server, or the real server. 
2. given a `DEBUG` value, the correct API url is selected and assigned to `ROOT`. `DEBUG = True` will select the local dev server.

### DIFF
--- a/docs/users/api_usage_examples/get_latest_import.py
+++ b/docs/users/api_usage_examples/get_latest_import.py
@@ -1,6 +1,10 @@
 import requests
 
-ROOT = '127.0.0.1'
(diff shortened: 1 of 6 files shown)


amCap1712 (approved):
LGTM, Thanks!

PR #2809, opened 2024-03-13
Update add-data.html

Aerozol:
Update add-data.html with a link to the TIDAL Hi-Fi music streaming app.
These are starting to get nicher, but MeB loves niche stuff, right? :)


mayhem (approved):
thanks!

PR #2810, opened 2024-03-13
Tidy up the Import Listens text

Aerozol:
A general tidy-up of the Import Listens page text.

- Added links to 'connect services' and 'import data' pages, for people who want to submit new listens. These pages aren't in the side menu (because they are under  Settings, not About)
- Changed the notes to bullet points, to be more readable and consistent
- Removed some duplicate text, tidied and shortened some text, fixed some errors

### DIFF
--- a/frontend/js/src/settings/import/ImportListens.tsx
+++ b/frontend/js/src/settings/import/ImportListens.tsx
@@ -49,47 +49,35 @@ export default function Import() {
         </div>
       )}
       <p>
-        Most users will want to import from Last.fm directly.
-        <br />
+        Import your existing listen history from other databasess. To submit <em>new</em>
+        listens, please visit the{" "} <a href="https://listenbrainz.org/add-data/">Connect services</a>
+        and{" "} <a href="https://listenbrainz.org/settings/music-services/details/">Submitting data</a>
+        pages.
+      </br>
         Fun Fact: The term <strong>scrobble</strong> is a trademarked term by
         Last.fm, and we cannot use it. Instead, we use the term{" "}
         <strong>listen</strong> for our data.
       </p>
 
-      <h3>Direct import from Last.fm and Libre.fm</h3>
+      <h3>Import from Last.fm and Libre.fm</h3>
       <p>
         The importer manually steps through your listen history and imports the
-        listens one page at a time. Should it fail for whatever reason, it is
-        safe to restart the import process. Running the import process multiple
-        times <strong>does not</strong> create duplicates in your ListenBrainz
-        listen history.
+        listens one page at a time.
       </p>
+
+      <ul>
+        <li>In order for this to work, you must disable the &#34;Hide recent listening information&#34; setting in your Last.fm{" "} <a href="https://www.last.fm/settings/privacy" target="_blank">Privacy Settings</a>.</li>
+        <li>Should it fail for whatever reason, it is safe to restart the import process.</li>
+        <li>Running the import process multiple times <strong>does not</strong> create duplicates in your ListenBrainz listen history.</li>
+        <li>Clicking the &quot;Import listens&quot; button will import your profile without the need to open LastFM.</li>
+        <li>Your listen counts may not be accurate for up to 24 hours after you import your data.{" "} <a href="https://listenbrainz.readthedocs.io/en/latest/general/data-update-intervals.html" target="_blank" rel="noopener noreferrer">See more details here.</a></li>
+      </ul>
+        
       <p>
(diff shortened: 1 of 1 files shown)


Aerozol:
I was trying to get a test server set up on my Win system before entering these changes, because there are a few fiddly format changes, and I'm sure I broke something. But setting that up is an ongoing headache, and this page has bugged me too much in the meantime.
It would be much appreciated if someone could test it for me :)

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -49,47 +49,35 @@ export default function Import() {
         </div>
       )}
       <p>
-        Most users will want to import from Last.fm directly.
-        <br />
+        Import your existing listen history from other databasess. To submit <em>new</em>
```
MonkeyDo, line 52:
```suggestion
        Import your existing listen history from other databases. To submit <em>new</em>
```

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -49,47 +49,35 @@ export default function Import() {
...
+        listens, please visit the{" "} <a href="https://listenbrainz.org/add-data/">Connect services</a>
+        and{" "} <a href="https://listenbrainz.org/settings/music-services/details/">Submitting data</a>
+        pages.
+      </br>
```
MonkeyDo, line 56: These are almost good, but they're making the linters freak out!
Linters are tools that check the code has no issues.

In this case, it's pointing out below that it's expecting closing tags and whatnot, but the real issue is this: self-closing HTML tags need the slash to be at the end like so:
```suggestion
      <br/>
```

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -105,15 +93,16 @@ export default function Import() {
           librefmApiKey={librefmApiKey}
         />
       )}
+
+      </br>
```
MonkeyDo, line 97: Same issue here:
```suggestion
      <br/>
```

MonkeyDo (changes requested):
Thanks for the improvement :)


There might be another step needed after applying the changes for this review, specifically to format the code according to the linter's feedback.
It's simple stuff like trying to keep the lines short, the indentation consistent, and other small code layout feedback.

We have a tool that will try to resolve the formatting issues automatically, I see three ways of doing it:
1. run it in docker; I can look for the right incantation to make it work
2. run it directly on your machine (will require installing nodeJS)
3. I can run it for you :)

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -49,47 +49,35 @@ export default function Import() {
...
-        Clicking the &quot;Import listens&quot; button will import your profile
-        now without the need to open LastFM.
-        <br />
-        You need to keep this page open for the tool to work, it might take a
-        while to complete. Though, you can continue doing your work. :)
+        listens one page at a time.
       </p>
+
+      <ul>
+        <li>In order for this to work, you must disable the &#34;Hide recent listening information&#34; setting in your Last.fm{" "} <a href="https://www.last.fm/settings/privacy" target="_blank">Privacy Settings</a>.</li>
```
MonkeyDo, line 69: The [automated tests](https://github.com/metabrainz/listenbrainz-server/pull/2810/checks#step:4:1943) are complaining about this, apparently to make links safer for older browsers:

```suggestion
        <li>In order for this to work, you must disable the &#34;Hide recent listening information&#34; setting in your Last.fm{" "} <a href="https://www.last.fm/settings/privacy" target="_blank" rel="noreferrer">Privacy Settings</a>.</li>
```

MonkeyDo (commented):
One last little annoyance to fix and we're good to merge :+1:

frontend/js/src/settings/import/ImportListens.tsx
Aerozol: Thanks monkey - is it policy to have this on all external links?

Aerozol:
Hey monkey, thanks!!

Do you want to meet sometime to go over the formatting stuff? Maybe walk me through the theory of it, and then see if it's worth trying to get the tool working for me, or if it's best left to someone with the skillz (I did always wonder how you all got the line widths even etc, I assume this tool does that...)

MonkeyDo (approved):

MonkeyDo:
Yes absolutely!
It shouldn't be very tricky one way or another and as you figured out, the tools do it all for us (OK, maybe not all but 95%)

PR #2812, opened 2024-03-14
Add LB Radio Artist Endpoint

mayhem:
Troi used to make a lot of calls to implement the artist radio. This was slow and cumbersome, but now that we know how we want this to work, I've managed to convert that entire process into one SQL query to make the process fast. Now troi only needs to massage the data from the call to this new endpoint.

This PR is safe to merge and does not depend on other PRs. The data table created in this PR already exists in production, no further action needed.


listenbrainz/db/lb_radio_artist.py
```diff
@@ -0,0 +1,151 @@
...
+        mode: LB radio mode, must be one of: easy, medium, hard.
+        seed_artist: artist mbid of the seed artist for similar artists
+        num_recordings_per_artist: Return up to this many recordings for each artist.
+        pop_begin: Popularity range percentage lower bound. A popularity range is given to narrow down
+                   the recordings into a smaller target group. The most popular track(s) on
+                   LB have a pop percent of 100. The least popular tracks have a score of 0.
+        pop_end: Popularity range percentage upper bound. See above.
+    """
+
+    query = """WITH mbids(mbid, score) AS (
```
amCap1712, line 39: It would be best to pass `UUID`s to this query and then just keep using `UUID`s till the end and in the last select cast it to `TEXT`. This will ensure that we don't miss index use because of TEXT to UUID or vice versa conversion at some step.

listenbrainz/db/lb_radio_artist.py
```diff
@@ -0,0 +1,151 @@
...
+                                    , SUM(total_listen_count) AS total_listen_count
+                                    , SUM(total_user_count) AS total_user_count
+                                 FROM combine_similarity
+                             GROUP BY recording_mbid, artist_mbid, similar_artist_mbid
+                           ), top_recordings AS (
+                               SELECT sa.similar_artist_mbid
+                                    , gs.recording_mbid
+                                    , total_listen_count
+                                    , PERCENT_RANK() OVER (PARTITION BY sa.similar_artist_mbid
+                                                               ORDER BY sa.similar_artist_mbid, total_listen_count ) AS rank
```
amCap1712, line 101: `sa.similar_artist_mbid` is redundant in the `ORDER BY` because you are partitioning by it, so the value will be same inside a partition and the order by in a window function only performs ordering per partition.

admin/timescale/create_tables.sql
```diff
@@ -256,6 +256,12 @@ CREATE TABLE similarity.artist_credit_mbids (
     score INT NOT NULL
 );
 
+CREATE TABLE similarity.overhyped_artists (
+    id                      SERIAL,
```
amCap1712, line 260: Use `INTEGER GENERATED ALWAYS AS IDENTITY NOT NULL` instead of `SERIAL`

admin/timescale/insert_default_data.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
+
+-- Insert the overhyped artists table
+-- TODO: Should this be added to table dumps or do we wait until the other similarity tables get dumped as well?
```
amCap1712, line 4: I think we can postpone this to until similarity data is dumped.

listenbrainz/db/lb_radio_artist.py
mayhem: That sounds like a good policy. will do.

admin/timescale/create_tables.sql
mayhem: Done and done. Updated in production.

listenbrainz/db/lb_radio_artist.py
```diff
@@ -0,0 +1,151 @@
...
+        mode: LB radio mode, must be one of: easy, medium, hard.
+        seed_artist: artist mbid of the seed artist for similar artists
+        num_recordings_per_artist: Return up to this many recordings for each artist.
+        pop_begin: Popularity range percentage lower bound. A popularity range is given to narrow down
+                   the recordings into a smaller target group. The most popular track(s) on
+                   LB have a pop percent of 100. The least popular tracks have a score of 0.
+        pop_end: Popularity range percentage upper bound. See above.
+    """
+
+    query = """WITH mbids(mbid, score) AS (
```
amCap1712, line 40: `score` can be removed from here (also from the parameters being passed to this query).

listenbrainz/webserver/views/api.py
```diff
@@ -777,3 +778,60 @@ def _get_listen_type(listen_type):
...
+    :param max_recordings_per_artist: The maximum number of recordings to return for each artist. If there are aren't enough recordings, all available recordings will be returned.
+    :param pop_begin: Popularity range percentage lower bound. A popularity range is given to narrow down the recordings into a smaller target group. The most popular recording(s) on LB have a pop percent of 100. The least popular recordings have a score of 0. This range is not coupled to the specified mode, but the mode would often determine the popularity range, so that less popular recordings can be returned on the medium and harder modes.
+    :param pop_end: Popularity range percentage upper bound. See above.
+    :resheader Content-Type: *application/json*
+    :statuscode 200: Yay, you have data!
+    :statuscode 400: Invalid or missing param in request, see error message for details.
+    """
+    max_similar_artists = _parse_int_arg("max_similar_artists")
+    max_recordings_per_artist = _parse_int_arg("max_recordings_per_artist")
+
```
amCap1712, line 810: Add validation for the mbid as well otherwise and invalid mbid will cause a 500 error when tried to cast to uuid in the database functions. We have a `is_valid_uuid` function in `api_tools.py`.

mayhem: Done.

PR #2814, opened 2024-03-15
Update add-data.html

matildepark:
Adds OngakuKiroku for those of us on Swinsian. Is it niche? Yes. Do I use it? Yes. Does it work well? Yeah.


mayhem (approved):
LGTM, thanks!

PR #2815, opened 2024-03-18
LB-1495: lb-radio endpoint is affected by letter case

ericd23:
# Problem
Basically, upper case tag won't do anything. For example: https://api.listenbrainz.org/1/lb-radio/tags?tag=Jazz&begin_percent=0&end_percent=50&count=50 gives 0 results. It's because PG's WHERE clause is case-sensitive.

# Solution
Convert all tags into lower case when loading the query params. Also add a integration test for this endpoint.


listenbrainz/webserver/views/api.py
```diff
@@ -722,6 +722,7 @@ def get_tags_dataset():
     tag = request.args.getlist("tag")
     if tag is None:
         raise APIBadRequest("tag param is missing")
+    tag = [t.lower() for t in tag]
```
amCap1712, line 725: I am not sure if all tag names in the database are lower cased. I'll check if that is the case, then the fix would work. If not, a case insensitive match should be performed in the SQL query. Will update.

MonkeyDo: Ping @amCap1712

mayhem: It appears that all tags are in fact lowercase:

```
musicbrainz_db=# select count(distinct(lower(name))) from tag;
 count  
--------
 213811
(1 row)

musicbrainz_db=# select count(*) from tag;
 count  
--------
 213811
(1 row)
```

mayhem:
I feel that this isn't a problem that should be fixed in ListenBrainz, its a troi problem. I've fixed this in troi, so closing this PR. Thanks for submitting the PR, much appreciated!

PR #2816, opened 2024-03-18
Add background tasks container

amCap1712:
Many tasks in LB take a long time to complete and cannot complete during the duration of a request. Add a background tasks service to process such tasks. At the moment, deleting a user account and entire listening history are two such tasks. In the future, support can be added for export listens and other user data in the same way.


frontend/js/src/settings/delete/DeleteAccount.tsx
```diff
@@ -67,10 +67,11 @@ export default function DeleteAccount() {
       toast.success(
         <ToastMsg
           title="Success"
-          message={`Successfully deleted acount for ${name}.`}
+          message={`Successfully enqueued account deletion for ${name}.`}
         />
       );
 
+      // TODO: logout user properly
```
mayhem, line 74: This should probably be done before merging. And also will a user be prevented from logging in again? I bet people will do that "just to check".

listenbrainz/background/background_tasks.py
```diff
@@ -0,0 +1,67 @@
...
+        return True
+
+    def start(self):
+        current_app.logger.info("Background tasks processor started.")
+        while True:
+            try:
+                task = get_task()
+                current_app.logger.info(f"Processing task: {task}")
+                if task is None:
+                    current_app.logger.info("No pending tasks found, sleeping 5 seconds")
```
mayhem, line 49: remove. this will make the logs uselessly noisy.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -621,27 +621,31 @@ def import_listens_dump(self, archive_path: str, threads: int = DUMP_DEFAULT_THR
...
 
         Args:
             musicbrainz_id: the MusicBrainz ID of the user
             user_id: the listenbrainz row id of the user
+            created: delete listens created before this timestamp
 
         Raises: Exception if unable to delete the user in 5 retries
         """
-        query = """
+        if created:
```
mayhem, line 636: is this right? If it is not null (should be 'is not None'?) then the value is overwritten?

mayhem (approved):
A couple of things to address, but overall great!

frontend/js/src/settings/delete/DeleteAccount.tsx
amCap1712: Yes, this is already broken. I just noticed it when testing the new changes. @MonkeyDo, can you please help with this? I am not sure and will have to lookup the correct way to logout users with react-router-dom.

We aren't preventing users from logging in again. I think we should clarify that in the error message. I did update it to say the account has been enqueued for deletion as opposed to the account has been deleted. But maybe we can improve it more.

listenbrainz/listenstore/timescale_listenstore.py
amCap1712: :facepalm: , that is wrong. Thanks for the catch. I'll fix.

listenbrainz/background/background_tasks.py
amCap1712: done

listenbrainz/listenstore/timescale_listenstore.py
amCap1712: done

frontend/js/src/settings/delete/DeleteAccount.tsx
amCap1712: done

MonkeyDo: Sorry just saw the message now. Do you still need any help?

PR #2818, opened 2024-03-20
LB-1369: Adding next track to playlist should clear the Input

SatyarajRana:
# Problem

When adding another track to a playlist the text entry input field contains the text from the previous track that was added. It would be great if instead this was cleared out so adding the next track wouldn't require the user to clear the input themselves.

# Solution

Calling the reset() function which sets the input value back to "" after the track has been added.

# Action

No further actions required.


SatyarajRana:
@MonkeyDo Please review

MonkeyDo (approved):
Yep, that works, thanks ! :rocket:

PR #2819, opened 2024-03-21
LB-1550: Fix inconsistent play status between Youtube widget and the brainzPlayer

ykchong45:
# Problem

[Ticket 1550](https://tickets.metabrainz.org/browse/LB-1550)

When streaming Youtube music in Recommendations, the brainzPlayer seems to play the next song automatically, indicated by the progress bar, whereas the Youtube widget does not start playing automatically.

# Solution

Use `loadVideoByXXX` in Youtube API to start playing next song by default. Referenced from [official docs](https://developers.google.com/youtube/iframe_api_reference#Video_Queueing_Functions).

### DIFF
--- a/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
@@ -281,9 +281,9 @@ export default class YoutubePlayer
       return;
     }
     if (videoId.startsWith("http")) {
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Glorious! Great find!
I have no idea how I missed such a simple fix, but it solves an issue that has been very, very annoying, thank you!

PR #2820, opened 2024-03-21
Fix serialize_jspf to add duration to playlist tracks

amCap1712:
serialize_jspf was missing a line to include the duration in the JSPF.


MonkeyDo (approved):
Thank you !

mayhem (approved):
I just had to make this change in troi as well. Sorry for not doing that here (yet). I think it is clear now that Troi is all things playlists and troi has this function as well. When I upgrade LB radio to the latest and greatest, I'll update the JSPF import/export functions in troi and have LB use that single version.

amCap1712:
Makes sense, thanks!

PR #2822, opened 2024-03-21
Integrating React Query to improve performance

anshg1214:
This PR adds `React-Query` to the loader to enable caching of the results.


frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -41,21 +42,26 @@ export type ArtistPageProps = {
...
   const {
-    artist,
-    popularRecordings,
-    releaseGroups,
-    similarArtists,
-    listeningStats,
-    coverArt: coverArtSVG,
-  } = useLoaderData() as ArtistPageProps;
+    data: {
+      artist,
```
MonkeyDo, line 49: I deployed to test.LB but found that this line was throwing an error when I moved from one artist to another.
To be clear, loading an artist page works, but if I click on one of the similar artists at the bottom of the page an error is thrown because  "cannot read property artist of undefined" since `data` is apparently undefined.

frontend/js/src/utils/QueryClient.ts
```diff
@@ -0,0 +1,13 @@
+import { QueryClient } from "@tanstack/react-query";
+
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
+      retry: 1,
+      staleTime: 1000 * 60 * 5,
```
MonkeyDo, line 8: Sounds reasonable for here.

For pages like the user dashboard, as soon as we have a timestamp (i.e not on the first page of my listens) we can cache the results for a loooong time since the API results should be the same (barring deleted listens, the deletion process is run once an hour I think?)

MonkeyDo (commented):
Good job on working out how to do this!
I'm still seeing an issue with initial data being undefined, see below

frontend/js/src/artist/ArtistPage.tsx
anshg1214: Fixed now

frontend/js/src/release/Release.tsx
```diff
@@ -1,11 +1,19 @@
...
   releaseGroupMBID: string;
 };
 
 export default function Release() {
-  const { releaseGroupMBID } = useLoaderData() as ReleaseLoaderData;
+  const location = useLocation();
+  const { releaseMBID } = useParams() as { releaseMBID: string };
+  const {
+    data: { releaseGroupMBID },
+  } = useQuery(RouteQuery(["release", releaseMBID], location.pathname)) as {
```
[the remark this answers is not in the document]
anshg1214: Yes, we're fetching the releaseGroupMBID from releaseMBID, and then redirecting the user to the album page using this endpoint https://github.com/metabrainz/listenbrainz-server/blob/3ba3d2d08772fff2f4c121f799bb1d7570cc5bdd/listenbrainz/webserver/views/entity_pages.py#L84-L103

frontend/js/src/utils/utils.tsx
```diff
@@ -1058,6 +1058,16 @@ export function getPersonalRecommendationEventContent(
...
 }
 
+export function getObjectForURLSearchParams(
+  urlSearchParams: URLSearchParams
+): Record<string, string> {
+  const object: Record<string, string> = {};
+  urlSearchParams.forEach((value, key) => {
+    object[key] = value;
+  });
+  return object;
```
MonkeyDo, line 1068: Not sure how relevant, but if you like one-liners:

```suggestion
  return Object.fromEntries(urlSearchParams);
```

anshg1214: My code editor is throwing an error 
```Property 'fromEntries' does not exist on type 'ObjectConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later.ts(2550)
```
Strange. We're using `es2022`

MonkeyDo: We're using es2022 _modules_ definition, our compilation target is es5, and this would still be a separate configuration option under `lib` : https://www.typescriptlang.org/tsconfig#lib

Forget I said anything, your solution is fine  :sweat_smile:

frontend/js/src/user/Dashboard.tsx
```diff
@@ -46,175 +48,79 @@ export type ListensProps = {
...
+  const [listenCount, setListenCount] = React.useState<number | undefined>();
+  const [dateTimePickerValue, setDateTimePickerValue] = React.useState<Date>(
+    nextListenTs ? new Date(nextListenTs * 1000) : new Date(Date.now())
+  );
+
+  React.useEffect(() => {
+    setListens(initialListens || []);
+  }, [initialListens]);
+
+  const createWebsocketsConnection = React.useCallback((): void => {
```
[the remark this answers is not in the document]
anshg1214: Correct, that the plan for the next PR, since socket is being used by multiple components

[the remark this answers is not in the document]
MonkeyDo: I mentioned this in our video chat, I'm not so sure anymore that we ant to do that :sweat_smile: 
Basically, we use websocket connections for two pages only at the moment: the user listens page and the playlist page,
In both cases, once the user navigates away to another tab or page, there's really no reason to keep using resources by keeping the websocket connection open.

I can see a future in which we get website notifications show in the UI, or a Playing Now conmponent visible across the website, or something like that, but it might be best to wait until we have that feature in mind.

We also talked with lucifer a little while back about possibly replacing websockets with server-sent events, so maybe we should also revisit that idea before refactoring WS on the frontend.

MonkeyDo (approved):
Let's goooooo !

I took the liberty to fix up the typescript types of the useQuery calls, as the data can be undefined we were potentially creating error scenarios where forcing the type hid the underlying issue.

Also updated the branch with latest single-page-app so that it's ready to merge (there was one conflict)

PR #2823, opened 2024-03-22
LB-620: Document way to run specific integration tests

ykchong45:
# Problem

[LB-620](https://tickets.metabrainz.org/browse/LB-620?filter=12410)

The documentation lacks information on running a single test file. 
It has been updated to provide clear instructions on executing a specific test file using the `./test.sh` script.

# Solution

1. Correct the function `is_unit_db_running` by fixing the container name queried in `docker ps`.
2. Remove the function `is_unit_db_exists`.
3. Adjust the logic so that when the test container is running, executing `./test.sh [path-to-test-file.py]` only runs the specified test and then exits, while retaining the test container for later runs.

### DIFF
--- a/docs/developers/devel-env.rst
+++ b/docs/developers/devel-env.rst
@@ -273,10 +273,10 @@ once so that subsequent running of the tests is faster:
 
 .. code-block:: bash
 
-   ./test.sh -u                      # build unit test containers, start up and initialise the database
-   ./test.sh [path-to-test-file.py]  # run tests, do this as often as you need to
-   ./test.sh -s                      # stop test containers, but don't remove them
-   ./test.sh -d                      # stop and remove all test containers
+   ./test.sh -u                                 # build unit test containers, start up and initialise the database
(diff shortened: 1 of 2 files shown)


MonkeyDo (changes requested):
Hi @ykchong45 !
Thanks for opening a PR.

I am not clear as to what the changes to the three non-documentation files achieve, looking at the PR title.

The only required change is adding the optional `[path-to-test-file.py]` in the documentation as well as possibly in the comments explaining the usage in the test.sh file.
None of the other changes make sense in the context of the problem stated. Could you please explain what they do, and why they are part of the same PR?

amCap1712 (approved):

amCap1712:
LGTM, Thanks!

ykchong45:
Thank you very much for helping me closing this pull request. 

For future's reference, this pull request updated the docs about testing, but in the meantime, I fixed a problem in the test script where it fails to execute a single test.

PR #2824, opened 2024-03-24
LB-1554: Fresh Releases should allow changing sort order

ericd23:
# Problem
As described in https://tickets.metabrainz.org/browse/LB-1554.

# Solution
Add a sorting option to let users choose ascending or descending sort.
image

### DIFF
--- a/frontend/js/src/explore/fresh-releases/components/ReleaseCardsGrid.tsx
+++ b/frontend/js/src/explore/fresh-releases/components/ReleaseCardsGrid.tsx
@@ -1,13 +1,13 @@
 import * as React from "react";
 import ReleaseCard from "./ReleaseCard";
 import { formatReleaseDate } from "../utils";
-import type { DisplaySettings } from "../FreshReleases";
+import type { DisplaySettings, SortDirection } from "../FreshReleases";
 
 type ReleaseCardReleaseProps = {
   filteredList: Array<FreshReleaseItem>;
   displaySettings: DisplaySettings;
   order: string;
-  direction: string;
+  direction: SortDirection;
 };
 
 const getKeyForOrder = (
@@ -63,10 +63,10 @@ export default function ReleaseCardsGrid(props: ReleaseCardReleaseProps) {
     return releaseKey;
   };
 
-  const mappedEntries =
-    direction === "descend"
-      ? Array.from(releaseMapping?.entries()).reverse()
-      : Array.from(releaseMapping?.entries());
+  const mappedEntries = Array.from(releaseMapping?.entries());
+  if (direction === "descend") {
+    mappedEntries.reverse();
+  }
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


frontend/js/src/explore/fresh-releases/components/ReleaseCardsGrid.tsx
```diff
@@ -62,9 +63,14 @@ export default function ReleaseCardsGrid(props: ReleaseCardReleaseProps) {
     return releaseKey;
   };
 
+  const mappedEntries =
+    direction === "descend"
+      ? Array.from(releaseMapping?.entries()).reverse()
+      : Array.from(releaseMapping?.entries());
+
```
MonkeyDo, line 70: I'd say this is a tad more readable, but there's nothing wrong with your code
```suggestion
  const mappedEntries = Array.from(releaseMapping?.entries());
  if(direction === "descend") {
    mappedEntries.reverse()
  }
```

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -6,10 +6,11 @@ import { formatReleaseDate, useMediaQuery } from "../utils";
 type ReleaseTimelineProps = {
   releases: Array<FreshReleaseItem>;
   order: string;
+  direction: string;
```
MonkeyDo, line 9: Same comment as above:
```suggestion
  direction: SortDirection;
```

MonkeyDo (changes requested):
Thanks for working on this!
Looking forward to being able to sort my lists the way I like it :)

ericd23:
Now that #2825 is merged, I will add support for sorting by confidence.

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -41,6 +43,10 @@ export default function ReleaseTimeline(props: ReleaseTimelineProps) {
         (date) => releasesPerDate[date] >= minReleasesThreshold
       );
 
+      if (sortDirection === "descend") {
+        filteredDates.reverse();
+      }
+
```
MonkeyDo, line 49: Here instead of repeating the reverse 4 times in the same functiuon, it would be better to do it a single time right before `return zipObject(percentArr, dataArr);`

MonkeyDo (changes requested):
Nice work so far! Thanks for adding the confidence sorting too!

I did notice that the sort order is inverted when using the confidence:
image

Other attributes are sorting as expected though, and it's working nicely!

One other detail that could be improved it the mobile-size view, where the new direction select is taking space, and maybe some reorganization would help: 
image
However if you're not sure how to attack this, you can leave it to me.

ericd23:
Thanks! I will have another look. Maybe my local test cases for confidence weren't thorough enough.

ericd23:
I will have a look at mobile view :)

MonkeyDo (approved):
Nice work, thank you for the improvement !

PR #2825, opened 2024-03-24
LB-1552: Add timeline to Fresh Releases 'For you' section

Xeitor:
# Problem
[LB-1552](https://tickets.metabrainz.org/browse/LB-1552)
The fresh releases timeline in only available in 'All' section

# Solution
Removed a conditional that was rendering the timeline only for the `PAGE_TYPE_SITEWIDE` section.

# Action
Maybe investigate why the conditional was there in the first place.


MonkeyDo (approved):
Thank you!
Such a simple change :D

I git-blame'd back in time to see what was the reason for hiding it in the "for you" view, but didn't find a reason. I think the for you section was just not appropriate for the timeline but it has been refactored completely since.

PR #2826, opened 2024-04-01
LB-1459 Add integration tests for websockets backend

munish0838:
# Problem

LB1459 - https://tickets.metabrainz.org/browse/LB-1459

# Solution
Added websockets container to test docker compose image and tets.sh. Create test_websockets.py to connect to websockets and try to send a valid single json

### DIFF
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -75,6 +75,8 @@ services:
     volumes:
       - ..:/code/listenbrainz:z
     command: python manage.py run_websockets -h 0.0.0.0 -p 8102
+    expose:
+      - "8102"
     ports:
       - "8102:8102"
     depends_on:
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -4,6 +4,7 @@ version: "3.4"
 # directory that this file is in (`docker/`) and so probably need to start with
 # `../` to refer to a directory in the main code checkout
 
+
 services:
 
   lb_db:
@@ -20,11 +21,14 @@ services:
     volumes:
       - ./couchdb_test.ini:/opt/couchdb/etc/local.ini
 
+
   redis:
     image: redis:6.2.2
 
+
   rabbitmq:
     image: rabbitmq:3.8.16
+
 
   listenbrainz:
     build:
@@ -43,6 +47,7 @@ services:
       - lb_db
       - rabbitmq
 
+
   timescale_writer:
     image: listenbrainz
     command: python3 -m "listenbrainz.timescale_writer.timescale_writer"
@@ -53,6 +58,7 @@ services:
     environment:
       PYTHON_TESTS_RUNNING: 1
 
+
   background_tasks:
     image: listenbrainz
     command: python3 -m "listenbrainz.background.background_tasks"
@@ -62,6 +68,7 @@ services:
     environment:
       PYTHON_TESTS_RUNNING: 1
 
+
   frontend_tester:
     build:
       context: ..
@@ -69,6 +76,7 @@ services:
       target: listenbrainz-frontend-dev
     volumes:
       - ../frontend:/code/frontend:z
+
 
   websockets:
     image: listenbrainz
@@ -82,4 +90,5 @@ services:
     depends_on:
       - redis
       - rabbitmq
+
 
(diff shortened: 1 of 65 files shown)


mayhem:
I tried this PR locally, but I get:

E           socketio.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=8102): Max retries exceeded with url: /socket.io/?transport=polling&EIO=4&t=1712056391.9411511 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7c5b87d24450>: Failed to establish a new connection: [Errno 111] Connection refused'))

When the test runs. Can you please take another look at it?

munish0838:
thanks @mayhem, will looks into this

mayhem:
Any progress on this? I can't stress how important this PR is -- without a successful PR merged, we won't consider your application further.

docker/docker-compose.test.yml
```diff
@@ -69,3 +69,15 @@ services:
       target: listenbrainz-frontend-dev
     volumes:
       - ../frontend:/code/frontend:z
+
+  websockets:
+    image: web
```
amCap1712, line 74: the image here needs to be listenbrainz

listenbrainz/tests/integration/test_websockets.py
```diff
@@ -0,0 +1,52 @@
...
+from listenbrainz import db
+from listenbrainz.tests.integration import ListenAPIIntegrationTestCase
+from listenbrainz.webserver.views.api_tools import is_valid_uuid
+import listenbrainz.db.external_service_oauth as db_oauth
+
+class WebSocketTests(ListenAPIIntegrationTestCase):
+    def setUp(self):
+        super(WebSocketTests, self).setUp()
+        self.sio = socketio.Client(logger=True, engineio_logger=True)
+        self.sio.connect('http://localhost:8102',wait_timeout = 20)
```
amCap1712, line 22: url here needs to be http://websockets:8102

listenbrainz/tests/integration/test_websockets.py
```diff
@@ -0,0 +1,52 @@
...
+        # Send a listen
+        print(payload)
+        ts = int(time.time())
+        payload['payload'][0]['listened_at'] = ts
+        response = self.send_data(payload)
+        print(response)
+        self.assert200(response)
+        self.assertEqual(response.json['status'], 'ok')
+
+    def send_data(self, payload, user):
```
amCap1712, line 39: You can delete this method and just call the one from base class.

requirements_development.txt
```diff
@@ -5,3 +5,5 @@ pytest==7.1.0
 pytest-cov==3.0.0
 requests-mock==1.9.3
 pytest-subtests==0.8.0
+python-socketio[client]
+python-socketio[asyncio_client]
```
amCap1712, line 9: We only need the first client methinks

listenbrainz/tests/integration/test_websockets.py
```diff
@@ -0,0 +1,40 @@
+import socketio
+import json
+import time
+
+from listenbrainz.tests.integration import ListenAPIIntegrationTestCase
+
+
+class WebSocketTests(ListenAPIIntegrationTestCase):
```
amCap1712, line 8: I ran this locally and the test now works fine. I suggest you to update the asserts in the listen event to compare some fields in the listen sent and received.

Also, add another test for playing now.

amCap1712:
Hi! The test was hanging up because you had called `self.sio.wait()` in the setup which blocks the thread until disconnect completes so it should be instead be called in the tearDown step.

listenbrainz/tests/integration/test_websockets.py
munish0838: Relieved to know that the tests finally work now. Will work on the assert and playing test now

munish0838:
@amCap1712 Added test for playing now. Both the tests work for me now. Working on adding the comparison of sent and received data since adding those asserts to listen function led to failing of test

munish0838:
@amCap1712, I have updated the tests to include more asset statements for comparing sent and received data

munish0838:
@mayhem Could you please once check if it passes the tests now

munish0838:
> Any progress on this? I can't stress how important this PR is -- without a successful PR merged, we won't consider your application further.
Hey @mayhem , I have added the tests and only pending a review from lucifer now for merging. Please let me know if there is anything else I would need to do too. I hope my application could be considered now.

PR #2829, opened 2024-04-07
LB-1380: add alt text support to the Art Creator

huhridge:
# Problem

This PR adds alt text support to the Art Creator.

# Solution

After reading some conventions for accessibility, the best way seemed to add a `<title>` and `<desc>` tag to the main `<svg>` tag.
Apart from that, after talking to aerozol on the IRC, this layout was agreed upon:
preview
This also uniforms the button size and shape for the copy-link button, which was previously smaller and square.
 
# Action
No further action required

### DIFF
--- a/frontend/css/stats-art-creator.less
+++ b/frontend/css/stats-art-creator.less
@@ -201,6 +201,10 @@
       flex: 1;
       flex-basis: 300px;
       justify-content: left;
+      > .btn-link > span {
+        margin-right: 0.2em;
+        font-size: 14px;
+      }
     }
     .profile {
       text-align: left;
@@ -209,6 +213,7 @@
       min-width: 150px;
       height: 3em;
       margin-left: 0.6em;
+      margin-right: auto;
       > input {
         text-overflow: ellipsis;
         border-radius: 5px 0 0 5px;
--- a/frontend/css/stats-art-creator.less
+++ b/frontend/css/stats-art-creator.less
@@ -200,14 +200,30 @@
     .share-buttons {
       flex: 1;
       flex-basis: 300px;
+      justify-content: left;
     }
     .profile {
       text-align: left;
     }
     .link-container {
       min-width: 150px;
+      height: 3em;
+      margin-left: 0.6em;
       > input {
         text-overflow: ellipsis;
+        border-radius: 5px 0 0 5px;
+        height: inherit;
+      }
+      > span {
+        height: inherit;
+        > button {
+          height: inherit;
+          border-radius: 0 5px 5px 0;
+          line-height: 1em;
+          > svg {
+            font-size: 1.5em;
+          }
+        }
       }
     }
   }
--- a/frontend/js/src/explore/art-creator/components/IconTray.tsx
+++ b/frontend/js/src/explore/art-creator/components/IconTray.tsx
@@ -89,24 +89,11 @@ function IconTray(props: IconTrayProps) {
         {browserHasClipboardAPI && (
           <button
             type="button"
-            className="btn btn-icon btn-info"
+            className="btn btn-icon btn-link"
             onClick={onClickCopyAlt}
-            style={{
-              marginLeft: "auto",
-              background: "transparent",
-              color: "darkgray",
-            }}
           >
-            <text
-              style={{
-                textDecoration: "underline",
-                fontSize: "14px",
-                marginRight: "0.2em",
-              }}
-            >
-              alt text
-            </text>
-            <FontAwesomeIcon icon={faClone} fixedWidth />
+            <span className="text-muted">alt text</span>
+            <FontAwesomeIcon className="text-muted" icon={faClone} fixedWidth />
           </button>
         )}
       </div>
--- a/frontend/js/src/explore/art-creator/components/IconTray.tsx
+++ b/frontend/js/src/explore/art-creator/components/IconTray.tsx
@@ -2,6 +2,7 @@ import {
   faLink,
   faCode,
   faCopy,
+  faClone,
   faDownload,
   faUser,
 } from "@fortawesome/free-solid-svg-icons";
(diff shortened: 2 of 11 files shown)


Aerozol:
I don't have a LB test server, but the screenshot looks excellent!
May I ask what the output/text looks like that is copied into the clipboard?

huhridge:
On copying the top 5 one:

Top 5 Artists this month for huhridge
1. Frank Ocean 
2. brakence 
3. Khruangbin 
4. Charli XCX 
5. ¥$, Kanye West, Ty Dolla $ign 

The releases one:

Top 10 releases this month for huhridge 
1. Channel Orange by Frank Ocean 
2. hypochondriac by brakence 
3. Blonde by Frank Ocean 
4. A LA SALA by Khruangbin 
5. Von dutch by Charli XCX 
6. Superache by Conan Gray 
7. Snow by Angus &amp; Julia Stone 
8. Scum Fuck Flower Boy by Tyler, the Creator 
9. SUGA by Kyle Dion 
10. SMITHEREENS by Joji

huhridge:
i somehow broke my lb server as well, looking into it 😭😭

frontend/js/src/explore/art-creator/components/IconTray.tsx
```diff
@@ -83,6 +86,29 @@ function IconTray(props: IconTrayProps) {
             </span>
           )}
         </div>
+        {browserHasClipboardAPI && (
+          <button
+            type="button"
+            className="btn btn-icon btn-info"
+            onClick={onClickCopyAlt}
+            style={{
+              marginLeft: "auto",
```
MonkeyDo, line 95: this marginleft can be replaced with adding a `margin-right: auto` to the `.link-container` class, convenient because it already exists...

frontend/js/src/explore/art-creator/components/IconTray.tsx
```diff
@@ -83,6 +86,29 @@ function IconTray(props: IconTrayProps) {
...
+              background: "transparent",
+              color: "darkgray",
```
MonkeyDo, line 97: In the same spirit as above if you don't want to create a new class for this button, these two style can be replaced by using these css classes for the button component : `btn btn-sm btn-link`

and then adding the class `text-muted` to both children elements (the text and the icon)

frontend/js/src/explore/art-creator/components/IconTray.tsx
```diff
@@ -83,6 +86,29 @@ function IconTray(props: IconTrayProps) {
...
+            }}
+          >
+            <text
```
MonkeyDo, line 100: Why this text element? It's an SVG tag that does not belong in this HTML.
Maybe you want a `<span>` instead ?

frontend/js/src/explore/art-creator/components/IconTray.tsx
```diff
@@ -83,6 +86,29 @@ function IconTray(props: IconTrayProps) {
...
+              style={{
+                textDecoration: "underline",
+                fontSize: "14px",
+                marginRight: "0.2em",
```
MonkeyDo, line 104: Suggestion here:
 Somehow the `btn-xs` class will give us the margin we need (not ideal semantically, but it does the trick?) and as described above the `text-muted` class takes care of the color.
 
 Not sure we want the underline as it does not match any other button style anywhere on the website.
 
 
 
 Along with the other suggestions above the result is this:
image
It's not exactly the style you had before, so if you're unhappy with the result feel free to give this button its own CSS class and defined the CSS styles in the .less file.
Hope it makes sense !

frontend/css/stats-art-creator.less
```diff
@@ -200,14 +200,30 @@
     .share-buttons {
       flex: 1;
       flex-basis: 300px;
+      justify-content: left;
     }
     .profile {
       text-align: left;
     }
     .link-container {
```
MonkeyDo, line 208: Button is looking much better, thanks !

MonkeyDo (commented):
Nice improvement, thank you!
PR looks quite nice already, nice work!
I do have some suggestions and opinions below:

I added some suggestions to replace the `style` attributes. It's preferable to either add the styles in the CSS (.less) files or to use existing CSS classes. See comments for more details.

Since you are in there improving things with titles... I think there is another easy improvement by adding a title to each item artist/album title *text* in the SVGs.
The cover art ones already have titles, but the text ones don't.
So for example on the "top 5" SVG hovering over the artist names currently does nothing, but we can easily add a `<title>` element inside the `tspan` or `text` element so that hovering over the text or cover art makes it 1. reader accessible and 2. hover to see the name:
image
  
  
  Also: Khruangbin <3

MonkeyDo:
> `./test.sh` is passing on my system no issues, just warnings, don't know why it's failing here.

Ignore the one failing server test, it's not your code and it's a know issue we introduced last week (I think) that we'll resolve separately.

listenbrainz/webserver/views/art_api.py
```diff
@@ -238,12 +252,21 @@ def cover_art_custom_stats(custom_name, user_name, time_range, image_size):
                 images = _repeat_images(images, 5)
         except ValueError as error:
             raise APIBadRequest(str(error))
+        
+        #implicit string concatenation to conform to PEP regulations
+        title = f"Top {5 if custom_name == 'lps-on-the-floor' else 10} releases" \
```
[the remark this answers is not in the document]
huhridge: fixed

frontend/js/src/explore/art-creator/components/IconTray.tsx
huhridge: ended up just making a selector for that span, looks better this way imo

frontend/css/stats-art-creator.less
huhridge: thanks!

MonkeyDo (approved):
Looking good, thanks for the improvements !

huhridge:
thank you for reviewing! learned a lot!!

Aerozol:
Pictures look great - thank you both, excellent work!!

PR #2830, opened 2024-04-09
Use a separate key for the spotify cache

mayhem:
[ NOTE: https://github.com/metabrainz/docker-server-configs/pull/302 must be merged before this is deployed! ]

In theory this adds support for using the new spotify cache key. Also, not tested, since production.


amCap1712 (approved):
LGTM, Thanks!

PR #2831, opened 2024-04-10
Allow post get metadata

mayhem:
The current GET implementation for the recording metadata cache severely limits the number of recordings that can be fetched since the URL length exceeds our gateways max URL limit quickly. In troi I am refactoring things so that we use the metadata cache, rather than other custom endpoints. This PR allows up the MAX_ITEMS_TO_GET items to be fetched from the cache.


mayhem:
Test failures unrelated.

amCap1712 (approved):
LGTM, Thanks!

PR #2828, opened 2024-04-04
BrainzPlayer settings page / disable music sources

MonkeyDo:
We have wanted for some time now a way to disable specific music sources in BrainzPlayer, but it required a form of settings for BP which we didn't have.

This PR adds both features: a new settings page just for BP as well as the first set of preferences (saved to DB) to disable or enable music sources for playback
image

Also show some indications in BP as well as an info message when the user tries to play a track with everything disabled:
image

PR #2832, opened 2024-04-11
Update global context when disconnecting music service

MonkeyDo:
Following from #2828 we found the right way to directly update global context values.
In this case, when a user disconnects one of their connected music services we can remove the auth tokens, permissions and whatnot form the global context to avoid having to refresh the page or have outdated tokens.

We can't do the same for when we connect to a music service, because with oauth we get redirected to the oauth provider and then back to the callback and finally back to the music services page (with a fresh global context with token etc.)


anshg1214 (approved):
LGTM 🚀

PR #2834, opened 2024-04-12
Add similar_artist_name to LB Radio artist element

mayhem:
Having the artist name available in the output makes everything downstream easier and avoids further API calls from troi.


listenbrainz/webserver/views/api.py
```diff
@@ -793,6 +793,7 @@ def get_artist_radio_recordings(seed_artist_mbid):
             {
               "recording_mbid": "401c1a5d-56e7-434d-b07e-a14d4e7eb83c",
               "similar_artist_mbid": "cb67438a-7f50-4f2b-a6f1-2bb2729fd538",
+              "similar_artist_name": "Boo Hoo Boys",
```
MonkeyDo, line 796: Boo Hoo Boys ?

MonkeyDo (approved):
LGTM, but I have not run it locally to check

listenbrainz/webserver/views/api.py
mayhem: Sorry, getting silly in here.

PR #2837, opened 2024-04-13
LB-1558: user similarity stats on user pages

huhridge:
# Problem
This PR adds user similarity stats on user pages.

Ticket: [LB-1558](https://tickets.metabrainz.org/browse/LB-1558)

# Solution

Added a `compatibilityModal` component that is included in the `UserSocialNetwork` component if the user page is other than that of the current user's. It displays upto 5 common artists.

Mockup

# Action

No further action required!

### DIFF
--- a/frontend/js/src/user/components/follow/CompatibilityCard.tsx
+++ b/frontend/js/src/user/components/follow/CompatibilityCard.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import { includes as _includes } from "lodash";
 
-import { faCircleInfo } from "@fortawesome/free-solid-svg-icons";
+import { faCircleInfo, faSquarePlus } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { Link } from "react-router-dom";
 import ReactTooltip from "react-tooltip";
@@ -55,7 +55,8 @@ function CompatibilityCard(props: CompatibilityCardProps) {
         {hasMoreThanFive && (
           <>
             <span data-tip data-for="more-artists-tooltip">
-              , and more.
+              , and more.{" "}
+              <FontAwesomeIcon icon={faSquarePlus} size="sm" color="gray" />
             </span>
             <ReactTooltip id="more-artists-tooltip" place="top">
               {otherArtists.map((artist, index) => {
--- a/frontend/js/src/user/components/follow/CompatibilityCard.tsx
+++ b/frontend/js/src/user/components/follow/CompatibilityCard.tsx
@@ -19,34 +19,65 @@ export type CompatibilityCardProps = {
 };
 
 function CompatibilityCard(props: CompatibilityCardProps) {
-  const { user, similarityScore, similarArtists } = props;
+  const { user, similarityScore } = props;
+  let { similarArtists } = props;
 
   let content;
+  let isSliced = false;
+
+  if (similarArtists.length > 25) {
+    similarArtists = similarArtists.slice(0, 26);
+    isSliced = true;
+  }
 
   if (similarArtists.length > 0) {
     content = (
       <div className="text-center">
         {"You both listen to "}
-        {similarArtists.map((artist, index) => {
-          return (
-            <span>
-              {index > 0 && ", "}
-              {index > 0 && index === similarArtists.length - 1 ? "and " : ""}
-              {artist.artist_mbid !== null ? (
-                <Link
-                  to={`/artist/${artist.artist_mbid}`}
-                  title={artist.artist_name}
-                  target="_blank"
-                  rel="noopener noreferrer"
-                >
-                  {artist.artist_name}
-                </Link>
-              ) : (
-                `${artist.artist_name}`
-              )}
+        {similarArtists
+          .slice(...(similarArtists.length > 5 ? [0, 5] : [0]))
+          .map((artist, index) => {
+            return (
+              <span>
+                {index > 0 && ", "}
+                {index > 0 && index === similarArtists.length - 1 ? "and " : ""}
+                {artist.artist_mbid !== null ? (
+                  <Link
+                    to={`/artist/${artist.artist_mbid}`}
+                    title={artist.artist_name}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                  >
+                    {artist.artist_name}
+                  </Link>
+                ) : (
+                  `${artist.artist_name}`
+                )}
+              </span>
+            );
+          })}
+        {similarArtists.length > 5 && (
+          <>
+            <span data-tip data-for="more-artists-tooltip">
+              , and more.
             </span>
-          );
-        })}
+            <ReactTooltip id="more-artists-tooltip" place="top">
+              {similarArtists.slice(5, -1).map((artist, index) => {
+                return (
+                  <span>
+                    {index > 0 && ", "}
+                    {index > 0 && index === similarArtists.length - 1
+                      ? "and "
+                      : ""}
+                    {index > 0 && index % 5 === 0 && <br />}
+                    {`${artist.artist_name}`}
+                  </span>
+                );
+              })}
+              {isSliced && <span> and even more.</span>}
+            </ReactTooltip>
+          </>
+        )}
       </div>
     );
   } else {
--- a/frontend/js/src/user/components/follow/CompatibilityCard.tsx
+++ b/frontend/js/src/user/components/follow/CompatibilityCard.tsx
@@ -59,7 +59,7 @@ function CompatibilityCard(props: CompatibilityCardProps) {
     <Card id="compatibility-card" data-testid="compatibility-card">
       <div className="info-icon" data-tip data-for="info-tooltip">
         <Link
-          to={`/user/${currentUser.name}/stats/?range=all_time`}
+          to={`/user/${currentUser?.name}/stats/?range=all_time`}
           target="_blank"
           rel="noopener noreferrer"
         >
--- a/frontend/js/src/user/components/follow/CompatibilityCard.tsx
+++ b/frontend/js/src/user/components/follow/CompatibilityCard.tsx
@@ -1,9 +1,13 @@
 import * as React from "react";
 import { includes as _includes } from "lodash";
 
+import { faCircleInfo } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { Link } from "react-router-dom";
+import ReactTooltip from "react-tooltip";
 import Card from "../../../components/Card";
 import SimilarityScore from "./SimilarityScore";
+import GlobalAppContext from "../../../utils/GlobalAppContext";
 
 export type CompatibilityCardProps = {
   user: ListenBrainzUser;
@@ -17,6 +21,7 @@ export type CompatibilityCardProps = {
 
 function CompatibilityCard(props: CompatibilityCardProps) {
   const { user, similarityScore, similarArtists } = props;
+  const { currentUser } = React.useContext(GlobalAppContext);
 
   let content;
 
@@ -52,6 +57,15 @@ function CompatibilityCard(props: CompatibilityCardProps) {
 
   return (
     <Card id="compatibility-card" data-testid="compatibility-card">
+      <div className="info-icon" data-tip data-for="info-tooltip">
+        <Link
+          to={`/user/${currentUser.name}/stats/?range=all_time`}
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          <FontAwesomeIcon icon={faCircleInfo} />
+        </Link>
+      </div>
       <span>Your compatibility with {user.name} :</span>
       <SimilarityScore
         similarityScore={similarityScore}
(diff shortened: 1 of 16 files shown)


Aerozol:
The screenshot looks good to me!
Curious to see all the colours it goes through, when it's in production, and if I can find anyone similar enough to me :D

huhridge:
It's just the regular ones in the similarityScore component.

image

frontend/js/src/utils/APIService.ts
```diff
@@ -497,6 +497,43 @@ export default class APIService {
...
+    const response = await fetch(url);
+    await this.checkStatus(response);
+    if (response.status === 204) {
+      const error = new APIError(`HTTP Error ${response.statusText}`);
+      error.status = response.statusText;
+      error.response = response;
+      throw error;
+    }
+    return response.json();
+  };
```
MonkeyDo, line 535: You don't need this new method, you can use the existing `getUserEntity`:
```
getUserEntity(user.name, "artist", "all_time", 0, 100)
```

MonkeyDo (changes requested):
Looking good so far ! Thanks for the PR.

The only thing I would suggest adding is a link to the user's stats right after the list of artists in the compatibility card, it's a good bait for users to discover more artists while comparing stats.
Maybe a simple link at the end of the sentence like:
"You both listened to A, B, C and D (<ins>all time stats<ins/>)"

Aerozol:
A link to all stats would be cool, but it could be a very short text imo - "(stats)" or "(all)", for instance

MonkeyDo:
> A link to all stats would be cool, but it could be a very short text imo - "(stats)" or "(all)", for instance

I've had another think through this, and I think we should show a tiny bit more data to give some context about these artists in common.
In this case, we calculate the overlap between both users' top 100 artists of all time.

I think we should somehow mention this in a succint manner, and that would be the ideal place for a link to the stats.

Maybe something on the basis of "(based on the top 100 artists)" which is too long IMO.

Could it be simply "(<ins>top 100</ins>)" as a link ?

frontend/js/src/user/components/follow/CompatibilityModal.tsx
```diff
@@ -0,0 +1,70 @@
...
+  user: ListenBrainzUser;
+  similarityScore: number;
+  similarArtists: Array<{
+    artist_name: string;
+    artist_mbid: string | null;
+    listen_count: number;
+  }>;
+};
+
+function CompatibilityModal(props: CompatibilityModalProps) {
```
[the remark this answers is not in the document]
huhridge: I was thinking this also, but the `FollowerFollowingModal` and `SimilarUsersModal` are named the same way in the SocialNetwork component.

frontend/js/src/utils/APIService.ts
huhridge: Ahh of course, i was sure there must be some method! will change

huhridge:
> > A link to all stats would be cool, but it could be a very short text imo - "(stats)" or "(all)", for instance
> 
> I've had another think through this, and I think we should show a tiny bit more data to give some context about these artists in common. In this case, we calculate the overlap between both users' top 100 artists of all time.
> 
> I think we should somehow mention this in a succint manner, and that would be the ideal place for a link to the stats.
> 
> Maybe something on the basis of "(based on the top 100 artists)" which is too long IMO.
> 
> Could it be simply "(top 100)" as a link ?

i have done the changes required!

can you please expand a bit more on what kind of tiny data we can show?

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
```diff
@@ -69,6 +80,72 @@ export default class UserSocialNetwork extends React.Component<
...
+      const userArtists = payload.artists;
+      response = await getUserArtists(currentUser.name, "all_time", 100);
+      ({ payload } = response);
+      const currentUserArtists = payload.artists;
+      let similarArtists = currentUserArtists.filter((currArtist: any) =>
+        userArtists.some(
+          (artist: { artist_name: any }) =>
+            artist.artist_name === currArtist.artist_name
+        )
+      );
```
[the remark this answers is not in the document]
huhridge: yeah sometimes it is null, so it's better by `artist_name` for now

Aerozol:
I think adding more context would be awesome - why not go the whole way and have a (?) icon to rollover/click on that explains how the feature works? I've been meaning to make a ticket to add more of those anyway (or add them myself if I get the time) - imo they should be on most LB features, except for the very obvious ones.
If it's just plain text/a link I still lean towards keeping it short. (top 100) works as a middle ground - e.g. a little bit short and a little bit informative :P

frontend/js/src/user/components/follow/CompatibilityModal.tsx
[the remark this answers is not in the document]
MonkeyDo: Yeah, we should probably rename them eventually :)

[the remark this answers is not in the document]
huhridge: i changed this one for now anyways!

huhridge:
Added the info tooltip. Take a look:
Screenshot 2024-04-16 at 18 39 06
Clicking it also opens the stats page of the user for all time!

huhridge:
also any ideas how to fix the basename error in the run? the app is wrapped in the router on the dashboard page correctly! i don't know why this is happening

Aerozol:
I like the popup, let's see what monkey thinks.

Maybe the text could be:
"Artists displayed are the top matches, comparing the top 100 most listened artists of all time for both users. Click here for more statistics."
Assuming this is actually how it works :P

Personally, I'm not a fan of exclamation marks in UI text!!!!
I'm also not sure about 'click', assuming we show this on mobile too (tap?), but I'm not sure what's better.

MonkeyDo:
> also any ideas how to fix the basename error in the run? the app is wrapped in the router on the dashboard page correctly! i don't know why this is happening

I took the liberty to fix that for you. Basically a <Link> component from react-router won't work unless you have a browser context it can use, which needed to be added to the test suite

MonkeyDo:
> I like the popup, let's see what monkey thinks.
> 
> Maybe the text could be: "Artists displayed are the top matches, comparing the top 100 most listened artists of all time for both users. Click here for more statistics." Assuming this is actually how it works :P
> 
> Personally, I'm not a fan of exclamation marks in UI text!!!! I'm also not sure about 'click', assuming we show this on mobile too (tap?), but I'm not sure what's better.

I agree with everything :) Text is a bit better, and the !!!! aren't needed in this context.
I think "click here" works fine probably but I haven't deployed this yet to test on mobile, doing that now.

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
```diff
@@ -69,6 +80,82 @@ export default class UserSocialNetwork extends React.Component<
...
+      let response = await getUserEntity(
+        user.name,
+        "artist",
+        "all_time",
+        0,
+        100
+      );
+      const userArtists = (response as UserArtistsResponse).payload.artists;
+      response = await getUserEntity(
+        currentUser.name,
```
MonkeyDo, line 131: This currently fetches all time global stats when a user is not logged in, so you end up with whatever intersection there is between the user whose page we're looking at and the global stats.
We should exit early if there is no currentUser, as commented above.

frontend/js/src/user/components/follow/CompatibilityCard.tsx
```diff
@@ -0,0 +1,86 @@
...
+        <Link
+          to={`/user/${currentUser?.name}/stats/?range=all_time`}
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          <FontAwesomeIcon icon={faCircleInfo} />
+        </Link>
+      </div>
+      <span>Your compatibility with {user.name} :</span>
+      <SimilarityScore
```
MonkeyDo, line 70: I think instead of showing 0% when that's the value we have (i.e most likely not calculated or insufficient overlap) we should have a bit of text instead of the progress bar and all the rest, 

Maybe something along the lines of  "You and {userName} do not share any of your respective top 100 artists."?
With no hr element underneath, but we can keep the tooltip though, it's still useful. 
Hopefully @Aerozol will have some input on the phrasing :)

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
```diff
@@ -69,6 +80,82 @@ export default class UserSocialNetwork extends React.Component<
...
+      );
+      const currentUserArtists = (response as UserArtistsResponse).payload
+        .artists;
+      let similarArtists = intersectionBy(
+        userArtists,
+        currentUserArtists,
+        "artist_name"
+      );
+      if (similarArtists.length > 5) {
+        similarArtists = similarArtists.slice(0, 5);
```
MonkeyDo, line 145: OK, hear me out... We already have the full list of similar artists, I wonder if we could show only the top 5 (like we do currently) as text, and have the whole list show when we hover over that section of the CompatibilityCard if there are more than 5.
Could be a good way to show more info straight away for those who want, while staying out of the way.
What do you think?

MonkeyDo (changes requested):
Deployed the PR to test.LB for more thorough testing, have some more feedback :)

Overall, a really nice addition that I'm looking forward to seeing in action :)

frontend/js/src/user/components/follow/CompatibilityCard.tsx
huhridge: the issue is there are cases where there is 0% and still common artists, like yours and mine. the score is 0%, but we still share Khruangbin as an artist.

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
huhridge: i was thinking about this too, but sometimes there are a lot, i have seen upto 40 with a friend of mine. don't know what the best way to display those will be. lemme know if you have any ideas, @Aerozol too

frontend/js/src/user/components/follow/CompatibilityCard.tsx
```diff
@@ -0,0 +1,86 @@
...
+    );
+  } else {
+    content = <div className="text-center">You have no common artists.</div>;
+  }
+
+  return (
+    <Card id="compatibility-card" data-testid="compatibility-card">
+      <div className="info-icon" data-tip data-for="info-tooltip">
+        <Link
+          to={`/user/${currentUser?.name}/stats/?range=all_time`}
```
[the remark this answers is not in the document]
huhridge: i'll also change the text from "Click here for more statistics" to "Click here for more of their statistics".
makes it more clear i think

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
```diff
@@ -69,6 +80,82 @@ export default class UserSocialNetwork extends React.Component<
...
+    const { user } = this.props;
+    const { APIService, currentUser } = this.context;
+    if (currentUser?.name === user?.name) {
+      return;
+    }
+    const { getSimilarityBetweenUsers } = APIService;
+    try {
+      const response = await getSimilarityBetweenUsers(
+        user.name,
+        currentUser.name
```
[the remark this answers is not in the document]
huhridge: huh weird, shouldn't it both return the same thing? i'll change it anyhow.

frontend/js/src/user/components/follow/CompatibilityCard.tsx
MonkeyDo: Huh, that's interesting...
I am trying to find out if that's an error with the calculated similarity or something else.

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
huhridge: so display them as a tool tip when you over the list?

huhridge:
Added the tooltip which shows the whole artists (upto 20) list on hovering over the "and more." text
Screenshot 2024-04-18 at 02 39 15
The above is the maximum, when there were about 40 common artists.
Screenshot 2024-04-18 at 02 47 16
This one is the one more likely to happen.

Still renders like before when there are less than 5 common artists. I have kept the links for now, can change if necessary!

frontend/js/src/user/components/follow/CompatibilityCard.tsx
```diff
@@ -20,35 +19,65 @@ export type CompatibilityCardProps = {
...
+  const { user, similarityScore } = props;
+  let { similarArtists } = props;
 
   let content;
+  let isSliced = false;
+
+  if (similarArtists.length > 25) {
+    similarArtists = similarArtists.slice(0, 26);
+    isSliced = true;
+  }
```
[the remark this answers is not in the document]
huhridge: that does work better! thanks!

MonkeyDo (approved):

MonkeyDo:
Looking great! Nice to see this component on the website as we have not been surfacing this similarity data very much so far.

It's already showing there are some issues with similarity calculations :D

UltimateRiff:
wanted to mention it here, the (i) link is not currently mobile friendly imho (tapping it just links to stats without showing the pop-up, as might be expected on mobile). that said, looks good on desktop~ (I can open a Jira ticket for this issue, if need be~)

PR #2838, opened 2024-04-15
BrainzPlayer search fixes

MonkeyDo:
Rob reported a few issues with the spotify player and BP more generally.

After some debugging, we found out why some search queries find no matches despite the song being available in their catalog. Turns out we were not encoding the colons in the search query...

Second issue with spotify search API is due to the featuring artist, which can sometimes cause some issues.

Finally, show a warning toast when we are skipping a track because we found no result, as it is pretty poor UX for the user otherwise for track to be skipped without any explanation, makes it look like a bug.


mayhem (approved):
Looks good -- thanks so much!

PR #2841, opened 2024-04-18
Fix alt-text in art creator when stats are few

amCap1712:
If a user has less than 5/10 artists or releases, the alt text generation fails with an IndexError resulting in an internal server error. Make iteration smarter to fix it. Also, move the alt text creation to template for keeping it with the rest of the template.


mayhem (approved):
Hard to review, but looks ok...

PR #2842, opened 2024-04-18
Fix compatibility card when not logged in

huhridge:
# Problem
Compatibility card still shows when no user is logged in.

# Solution
Changed the checking as currentUser is apparently not empty when a user isn't logged in.

# Action

No further action required!


MonkeyDo (approved):
Thanks for the quick fixing :rocket: !

PR #2846, opened 2024-04-23
Fix secondary navbar CSS layout

MonkeyDo:
@Aerozol reported that the layout for the secondary nav (top navbar) was broken recently (during our first single-page-app implementation PRs).
We didn't look carefully enough at the changes in the CSS regarding the main containers, and this resulted in the secondary nav being part of the main content with its width restrictions:
image

It was not as trivial to fix as I originally thought, hence the delay, but here we are with everything in its right place :
image


Aerozol:
Thank you monkey, sorry for the annoyance!!

PR #2844, opened 2024-04-22
Refactor UserFeed component 

MonkeyDo:
This PR rewrites our UserFeed class component into a functional component, and uses a new /feed POST endpoint to send the required data, as with other endpoints using react-query.

Rewrote the tests as well and we needed a good way to test react-query components. Spent some time to figure everything out but it should server as a good base for other components.

PR #2853, opened 2024-04-25
Refactor HueSound into a functional component

MonkeyDo:
Based on the userfeed-refactor branch, wait for #2844 to be merged please


frontend/js/src/explore/huesound/HueSound.tsx
```diff
@@ -0,0 +1,205 @@
...
+      .catch((err) => {
+        toast.error(
+          <ToastMsg
+            title="Error"
+            message={err.message ? err.message.toString() : err.toString()}
+          />,
+          { toastId: "error" }
+        );
+      });
+  }, [selectedColorString, APIService]);
```
anshg1214, line 55: It is not a good idea to call the API when the component mounts. Whenever the page mounts, the releases corresponding to hex `000000` are loaded on the screen. Instead the behaviour should be similar to how it is right now on prod.

image

anshg1214, line 55: Set the loading state to false once the API call have been made
```suggestion
    setLoading(false);
```

frontend/js/tests/explore/huesound/HueSound.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+describe("HueSound", () => {
+  let server: SetupServerApi;
+  beforeAll(async () => {
+    // Mock the server responses
+    const handlers = [
+      http.get("/1/explore/color/000000", () =>
+        HttpResponse.json({ payload: { releases: [release] } })
+      ),
+    ];
```
anshg1214, line 51: Again, I don't believe that it should call hex `000000` on mount, so the test in this case needs to be refactored accordingly

frontend/js/tests/explore/huesound/HueSound.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    renderWithProviders(<HueSound />);
+    const canvasElement = screen.getByTestId("colour-picker");
+    await user.click(canvasElement);
+    const releaseButton = screen.getByRole("button");
+    await user.click(releaseButton);
+    const links = screen.getAllByRole("link");
+    expect(links.at(0)).toHaveTextContent("Prelude to Heliogabalus");
+    expect(links.at(1)).toHaveTextContent("Rorcal & Music For The Space");
+    
+    const brainzPlayerMessages = messages.filter(
```
anshg1214, line 97: Some whitespace here
```suggestion

```

anshg1214 (changes requested):
Thanks! The PR looks really good, just some small changes left.

frontend/js/src/explore/huesound/HueSound.tsx
MonkeyDo: That was...not on purpose. Didn't realize.

Now that I look at it, I wonder if this should be a random (or pseudo-random) color on every start?
Anyway, will fix it as on prod

MonkeyDo:
Note that this component is refactored to play nice with react-router, but has not been refactored to be used with the new BrainzPlayer with queues

anshg1214 (approved):
LGTM 🚀

PR #2854, opened 2024-04-26
Limit Spotify connection attempts

MonkeyDo:
We have a retry count to limit attempts to play a track, but not to connect to Spotify at the beginning.

Since we've been having rate-limiting issues with Spotify API, it has resulted in endless connection retries until the refresh token is returned correctly, which can't be good for the initial rate-limiting issue...

This PR limits the number of connection retries to 6 before abandoning and marking the music service as unavailable.

### DIFF
--- a/frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
@@ -219,14 +219,16 @@ export default class SpotifyPlayer
     retryCount = 0
   ): Promise<void> => {
     const { device_id } = this.state;
-    const { handleError } = this.props;
+    const { handleError, onTrackNotFound } = this.props;
     if (retryCount > 5) {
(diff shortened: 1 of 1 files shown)


frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
```diff
@@ -219,14 +219,16 @@ export default class SpotifyPlayer
     retryCount = 0
   ): Promise<void> => {
     const { device_id } = this.state;
-    const { handleError } = this.props;
+    const { handleError, onTrackNotFound } = this.props;
     if (retryCount > 5) {
       handleError("Could not play Spotify track", "Playback error");
+      onTrackNotFound();
```
MonkeyDo, line 225: Unrelated to this PR, but I think this function call was missing and resulted in BrainzPlayer being stuck if there was a failure playing a track, instead of trying another source or skipping the track.

PR #2857, opened 2024-04-29
Fix tests

amCap1712:
1. Fix troi test due to change in imports.
2. Fix websockets tests as asserts from other threads don't work properly. move asserts to main thread.
3. Fix settings views test.
    Trying to rollback the transaction in the except clause can be problematic if it fails and raises an exception. As there is no exception handler for it, it will lead to the container stopping. In production, the container would restart because it is run with unless-stopped but that doesn't happen for tests. so move rollback to inside try as the first step.

### DIFF
--- a/listenbrainz/tests/integration/test_settings_views.py
+++ b/listenbrainz/tests/integration/test_settings_views.py
@@ -102,9 +102,6 @@ class SettingsViewsTestCase(IntegrationTestCase):
         resp = self.client.post(self.custom_url_for('settings.delete_listens'))
         self.assertEqual(resp.status_code, 200)
 
-        # listen counts are cached for 5 min, so delete key otherwise cached will be returned
-        cache.delete(REDIS_USER_LISTEN_COUNT + str(self.user['id']))
-
         with self.app.app_context():
             task = get_task()
             self.assertIsNotNone(task)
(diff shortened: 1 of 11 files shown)


MonkeyDo:
I'm rooting for you @amCap1712 ! :rocket:

amCap1712:
Per aspera ad astra.

listenbrainz/tests/integration/test_settings_views.py
```diff
@@ -102,30 +102,30 @@ def test_delete_listens(self):
...
             self.assert200(resp)
             if json.loads(resp.data)['payload']['count'] == 0:
-                continue
-            else:
-                ok = True
+                break
 
-        if not ok:
-            self.assertEqual(json.loads(resp.data)['payload']['count'], 0)
+        print("listen count:", cache.get(REDIS_USER_LISTEN_COUNT + str(self.user['id'])))
```
mayhem, line 127: keep? delete?

mayhem (approved):

listenbrainz/tests/integration/test_settings_views.py
amCap1712: removed, thanks

PR #2858, opened 2024-04-30
Fix playing-now listen on user dashboard

MonkeyDo:
Reported by Chaban in https://community.metabrainz.org/t/listenbrainz-single-page-application/692873/2
The playing-now listen disappears if you navigate away from the dashboard and then come back to it, and there might be some issues with updating it after that too (didn't fin done upon testing but...)

With the new query cache, whatever we were doing previously (plunk t he playing now listen at the top of the listens list sent by the server) is not adapted anymore.
Send the playing now listen separately in the page props so we don't have to manipulate the listens array in the query cache, and also update the cache when a new playing-now listen comes in through websockets so that it persists updated once you navigate back to the Dashboard


anshg1214 (commented):
The PR does fix that the Playing now component was static sometimes, but if you try navigating and comping back to the dashboard, you'll see that the playing now component disappears. This is because the websocket connection dispatches the latest listen, but the endpoint does not. I believe this is because of the queues we use to process the new listens.

MonkeyDo:
Hmmm, it was working running locally, unless I'm missing something. Will test again more thoroughly.

MonkeyDo:
Deployed to test.LB, but the websockets update only works on prod.

However you can try playing something and load your dashboard, the playing now listen is sent by the server.
Then you can test navigating around and back to the dashboard, it seems to be working fine for me.

It also resolves another issue, where the playing now listen was not changing if you navigated from your dashboard to another user's dashboard and was showing your own playing now.

anshg1214 (approved):
I've tested it thoroughly, and it works perfectly now. (Maybe something wrong with my local setup)

Thanks for fixing this!!

PR #2859, opened 2024-04-30
Send the correct playing_now listen metadata

MonkeyDo:
...and not the metadata of the previous track *facepalm*
The issue is that even though we call `setState` earlier in the function, the state values are not updated by the time we call `submitNowPlayingToListenBrainz` and the track metadata is still that of the previous track.

Thankfully this was only an issue with playing_now listens which are by nature transient.


amCap1712 (approved):
LGTM, Thanks!

PR #2871, opened 2024-05-10
Speedup stats processing in Spark cluster

amCap1712:
1. Write a copy of the listens to HDFS on import of a full dump, this makes speeds up filtering of listens and increases the speed of processing in many cases.
2. Remove Pydantic validation in places where it seemed redundant or of not much use.

Before this PR, an entire stats run took about 9 hours. With step 2, it went down to 6.25 hours and then with step 1 on top of it, it goes down to 5.75 hours.


listenbrainz_spark/stats/user/entity.py
```diff
@@ -103,26 +102,26 @@ def calculate_entity_stats(from_date: datetime, to_date: datetime, table: str, c
...
             logger.error("Invalid entry in entity stats", exc_info=True)
             total_entity_count -= 1
 
     try:
-        return UserEntityRecords(
-            user_id=_dict["user_id"],
-            data=entity_list,
-            count=total_entity_count
-        )
+        return {
```
mayhem, line 120: I agree, this step isn't needed here.

mayhem (approved):
Great speedup!

PR #2872, opened 2024-05-15
Jspf cleanup

mayhem:
This PR fixes some problems with JSPF serialization:

1. The created_for element should be part of the extension.
2. track identifier should be a list not a string.

This PR corrects the problems above, but still allows playlists in the old format to be submitted.  All code fragments for this have been marked with a date, allowing us to clean this code up in a year.

This PR must be released WITH the troi PR of the same branch name!

TODO: Update the Troi release before merging.

### DIFF
--- a/listenbrainz/db/model/playlist.py
+++ b/listenbrainz/db/model/playlist.py
@@ -151,7 +151,7 @@ class Playlist(BaseModel):
 
         tracks = []
         for rec in self.recordings:
-            tr = {"identifier": [PLAYLIST_TRACK_URI_PREFIX + str(rec.mbid)]}
+            tr = {"identifier": PLAYLIST_TRACK_URI_PREFIX + str(rec.mbid)}
             if rec.artist_credit:
                 tr["creator"] = rec.artist_credit
 
--- a/listenbrainz/db/model/playlist.py
+++ b/listenbrainz/db/model/playlist.py
@@ -151,7 +151,7 @@ class Playlist(BaseModel):
 
         tracks = []
         for rec in self.recordings:
-            tr = {"identifier": PLAYLIST_TRACK_URI_PREFIX + str(rec.mbid)}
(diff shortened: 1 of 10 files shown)


listenbrainz/db/model/playlist.py
```diff
@@ -152,7 +151,7 @@ def serialize_jspf(self):
 
         tracks = []
         for rec in self.recordings:
-            tr = {"identifier": PLAYLIST_TRACK_URI_PREFIX + str(rec.mbid)}
+            tr = {"identifier": [PLAYLIST_TRACK_URI_PREFIX + str(rec.mbid)]}
```
amCap1712, line 154: Doesn't the JSPF spec mandate this to be a single identifier? I suggest using an additional_identifiers or mbids or some one well named field in the extension to have multiple MBIDs.

mayhem: Good catch -- multiple IDs are not really needed, so easy fix.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -75,9 +75,16 @@ def validate_playlist(jspf):
...
-        recording_uri = track.get("identifier")
-        if not recording_uri:
-            log_raise_400("JSPF playlist track %d must contain an identifier element with recording MBID." % i)
+        recording_uris = track.get("identifier")
+        if not recording_uris:
+            log_raise_400("JSPF playlist track %d must contain an identifier list with at least one recording MBID." % i)
+
+        # This allows identifier to be a list, tuple or string. The string support is a leftover and should be removed
+        # after 2025-06, which marks a year or backward compatibility. Only the first option of the if statement should remain.
+        if isinstance(recording_uris, list) or isinstance(recording_uris, tuple):
```
amCap1712, line 84: This should be removed as well since we are only doing 1 identifier?

mayhem: yep

amCap1712:
Sorry for the initial confusion. fixed now.

PR #2874, opened 2024-05-17
Improve listens and feedback download pages

MonkeyDo:
- Adds a loader indicator and an error alert
- Disables buttons while a download is in progress to avoid repeated queries
- Refactors buttons+mechanism into separate component to be reused in various places instead of reimplementing it on other pages (delete listens, delete account)
- Review warnings and phrasing in delete page

image

image

### DIFF
--- a/frontend/js/src/settings/delete-listens/DeleteListens.tsx
+++ b/frontend/js/src/settings/delete-listens/DeleteListens.tsx
@@ -1,36 +1,17 @@
 import * as React from "react";
 
-import { redirect, useLocation, Link } from "react-router-dom";
+import { useLocation, Link, useNavigate } from "react-router-dom";
 import { toast } from "react-toastify";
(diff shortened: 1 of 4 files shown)


frontend/js/src/settings/delete-listens/DeleteListens.tsx
```diff
@@ -43,27 +24,42 @@ export default function DeleteListens() {
...
+        navigate(`/user/${name}/`);
       }, 3000);
     } catch (error) {
       toast.error(
         <ToastMsg
           title="Error"
-          message={`Error while deleting listens for user ${name}`}
+          message={
+            <>
+              Error while deleting listens for user {name}:{error.toString()}
```
kellnerd, line 48: Just a minor thing which caught my eye while briefly skimming this PR:
```suggestion
              Error while deleting listens for user {name}: {error.toString()}
```

PR #2878, opened 2024-05-22
Don't hit delete routes by accident...

MonkeyDo:
With #2874 and refactoring the DL buttons into a reusable component for multiple pages, I did not realize our use of `window.location.href` in the existing code for the *download* button functionality was a big old bug that needed fixing.
It meant that on the delete listens/delete account pages the *download* button would hit the current route (i.e `/settings/delete/` or `/settings/delete-listens/`) endpoint instead which is the ones we use for deleting account/listens.
"Want to download your listens? Well, I'm deleting your account."

Hardcoding those danger URLs instead to ensure this doesn't happen again by mistake.

### DIFF
--- a/frontend/js/src/settings/delete-listens/DeleteListens.tsx
+++ b/frontend/js/src/settings/delete-listens/DeleteListens.tsx
@@ -18,7 +18,7 @@ export default function DeleteListens() {
     e.preventDefault();
 
     try {
-      const response = await fetch(location.pathname, {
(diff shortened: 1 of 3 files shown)


mayhem (approved):
for once, hard-coded seems better.

MonkeyDo:
I'm thinking of adding other mechanisms to protect against accidentally hitting the route. For starters a confirmation modal.
But let's get this merged and fix beta first.

PR #2879, opened 2024-05-22
Show a confirmation modal before deletion

MonkeyDo:
Adds a confirmation modal on the 'delete listens' and 'delete account' pages, requiring an extra user click before the deletion  goes through.
Helps prevent accidental clicks, considering there is no way to cancel those queries.

image

image

image


mayhem (approved):
yay!

PR #2881, opened 2024-05-25
Fix user stats navigation for special case usernames

MonkeyDo:
Issue reported [in IRC](https://chatlogs.metabrainz.org/libera/metabrainz/2024-05-24/?msg=5339030&page=3)

The code that matches the URL to determine what entity type we navigated to breaks if the user has a space or other special character in their name, because the value of the user's name in the URL will not match the value in global props.
For spaces for example`my username` will become `my%20username` in the URL, and this code will not match anything, defaulting to "artist" entity: https://github.com/metabrainz/listenbrainz-server/blob/212d4da1ce1547cc8fd7f4117aca0e752e2c35a5/frontend/js/src/user/charts/UserEntityChart.tsx#L352-L357

While I was in there modifying this code, I also removed an unnecessary duplicated API 

Deployed to test.LB for testing, note we are successfully on the the "tracks" tab, previously not working and defaulting to artists:
image

### DIFF
--- a/frontend/js/src/user/charts/UserEntityChart.tsx
+++ b/frontend/js/src/user/charts/UserEntityChart.tsx
@@ -350,7 +350,9 @@ export const UserEntityChartLoader = async ({
   const range: UserStatsAPIRange =
     (currentURL.searchParams.get("range") as UserStatsAPIRange) ?? "all_time";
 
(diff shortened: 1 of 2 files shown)


anshg1214 (approved):
LGTM! Thanks

PR #2882, opened 2024-05-26
Dump improvements

amCap1712:
1. Some more user data tables were missing from data dumps. Add pinned_recording and user_relationship to public dumps as anyone can view the data for all users's from listenbrainz.org. reported_users, user_setting, user_timeline_event, and hide_user_timeline_event are only visible to the user they belong to and hence adding to private dumps.
2. When private dumps were separated, I forgot to update scripts that deleted temporary files and old backups. Fix that now.


mayhem (approved):
thanks!

PR #2884, opened 2024-05-26
Cache User Stats

anshg1214:
This PR adds caching in the User Stats.


MonkeyDo:
I have two preliminary comments:

There is an initial query done when there is no range in the URL, and a subsequent one with the default rang of all_time.
Is there anyway we can set a default earlier on to avoid the initial empty range API call?

I don't understand why changing the metric in the artist map should reload the page, but I see it's a conscious decision.
Maybe we can invalidate a specific part of the data instead, but not sure about that

anshg1214:
Thanks for the review. 
I've fixed the reloading issue on changing the metric on artist map. I believe the PR on test was of the older commit.

However, I found another issue. When you click on the artist names on the map, although these links are using the `Link` component, it does not use the router to navigate: 
image

frontend/js/src/user/stats/components/UserArtistMap.tsx
```diff
@@ -126,99 +73,98 @@ export default class UserArtistMap extends React.Component<
...
+  React.useEffect(() => {
+    if (rawData && rawData?.payload) {
+      const newProcessedData = processData(selectedMetric, rawData);
+      setData(newProcessedData);
+    }
+  }, [rawData, selectedMetric]);
+
+  const changeSelectedMetric = (newSelectedMetric: "artist" | "listen") => {
+    setSelectedMetric(newSelectedMetric);
+  };
```
MonkeyDo, line 91: You can delete this function and use `setSelectedMetric` directly in the pill buttons below

frontend/js/tests/user/stats/UserArtistMap.test.tsx
```diff
@@ -1,247 +1,145 @@
...
-      // When stats haven't been calculated, processData is called with an empty object
-      const result = instance.processData(
-        {} as UserArtistMapResponse,
-        "listen"
-      );
-
-      expect(result).toEqual([]);
-    });
+    expect(screen.getByTestId("user-stats-map")).toBeInTheDocument();
+    expect(screen.getByTestId("Choropleth")).toBeInTheDocument();
```
MonkeyDo, line 104: I think you're testing the same thing on line 100, can probably remove this line.

frontend/js/tests/user/stats/UserArtistMap.test.tsx
```diff
@@ -1,247 +1,145 @@
...
-      expect(instance.rawData).toMatchObject(userArtistMapResponse);
-
-      expect(wrapper.state()).toMatchObject({
-        data: userArtistMapProcessedDataArtist,
-        loading: false,
-      });
-    });
+    // Check if the choropleth is rendered correctly
+    const choropleth = screen.getByTestId("Choropleth");
+    expect(choropleth).toBeInTheDocument();
```
MonkeyDo, line 141: Same comment as above, I think you can remove this line as it duplicates the assertion on line 136

frontend/js/tests/user/stats/UserDailyActivity.test.tsx
```diff
@@ -1,201 +1,128 @@
...
-      expect(wrapper.state().loading).toEqual(false);
-      const { data } = wrapper.state();
-      // Sort results for stable comparison with test fixture
-      data.forEach((day) => {
-        day.data.sort((a, b) => (a.x as number) - (b.x as number));
-      });
-      expect(data).toEqual(userDailyActivityProcessedData);
-    });
+    const heatmap = screen.getByTestId("heatmap");
+    expect(heatmap).toBeInTheDocument();
```
MonkeyDo, line 124: Duplicates assertion on line 120

frontend/js/tests/user/stats/UserReports.test.tsx
```diff
@@ -1,173 +1,144 @@
...
 
-        window.location = {
-          href: "https://foobar.org?range=year",
-        } as Window["location"];
-        const range = instance.getURLParams();
-
-        expect(range).toEqual("year");
-      });
+      // Check if the URL is updated
+      expect(mockSearchParam).toEqual({ range: "month" });
```
MonkeyDo, line 126: Here I wonder if you could also test  that the query client has received a query.
For example using [queryClient.getQueryState](https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientgetquerystate) with the right queryKey, which has a `dataUpdatedAt` or `dataUpdateCount` properties that could be used to ensure there is something reacting to the pill being clicked

MonkeyDo (commented):
Works great after some testing, another good chunk of refactoring and I'm here for it :)
The test suite looks a lot cleaner too.

frontend/js/src/user/routes/userRoutes.tsx
```diff
@@ -29,7 +29,7 @@ const getUserRoutes = (): RouteObject[] => {
               index: true,
               lazy: async () => {
                 const UserReports = await import("../stats/UserReports");
-                return { Component: UserReports.UserReportsWrapper };
+                return { Component: UserReports.default };
               },
               loader: RouteLoader,
```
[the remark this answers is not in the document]
anshg1214: Fixed

frontend/js/tests/user/stats/UserReports.test.tsx
anshg1214: Fixed

PR #2885, opened 2024-05-28
Fix get_personal_recommendation_events_for_feed

amCap1712:
users can be null if the users to whom the recording was recommended have deleted their account, coalesce to empty array in this case.


mayhem (approved):
Thanks for the fix!

PR #2886, opened 2024-05-29
Add Entity Search

anshg1214:
This PR adds an entity search feature on the search page.

Earlier, only the searching user feature was supported on the website, but now, users can now search Artist, Album, Songs, Playlists and Users.


frontend/css/search.less
```diff
@@ -0,0 +1,23 @@
+.search-bar {
+  position: relative;
+  margin-top: 1em;
+  input {
```
MonkeyDo, line 4: We can truncate the text otherwise it appears under the magnifying glass icon:
```css
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: nowrap;
    display: initial;
    padding-right: calc(45px + 10px + .5em);
```

And possibly make the search bar wider (on the search results page only) to match the mockup with (inline styles for .search-bar element):
```css
    max-width: 400px;
    width: 45%;
```
or something  along those lines.

The result:
image

frontend/js/src/utils/APIService.ts
```diff
@@ -1635,4 +1661,27 @@ export default class APIService {
     await this.checkStatus(response);
     return response.json();
   };
+
+  searchPlaylistForUser = async (
```
MonkeyDo, line 1665: For consistency with the other method below:
```suggestion
  searchPlaylistsForUser = async (
```

frontend/js/src/utils/APIService.ts
```diff
@@ -1635,4 +1661,27 @@ export default class APIService {
...
+    searchQuery: string,
+    musicbrainzID: string,
+    count: number = 25,
+    offset: number = 0
+  ): Promise<any> => {
```
MonkeyDo, line 1670: This method anbd the one below should be typed more accurately.
I think you've created the right type for it:
```suggestion
  ): Promise<PlaylistTypeSearchResult> => {
```

frontend/js/src/search/SongSearch.tsx
```diff
@@ -0,0 +1,166 @@
...
+const RECORDING_COUNT_PER_PAGE = 50;
+
+type SongSearchProps = {
+  searchQuery: string;
+};
+
+type SongTypeSearchResult = {
+  count: number;
+  offset: number;
+  recordings: {
```
MonkeyDo, line 20: You might be able to use the existing MusicBrainzRecording, MusicBrainzRelease, etc. types that exist

frontend/js/src/search/SongSearch.tsx
```diff
@@ -0,0 +1,166 @@
...
+        listen={{
+          listened_at: 0,
+          track_metadata: {
+            mbid_mapping: {
+              artists,
+              artist_mbids: artistMBIDs,
+              release_mbid:
+                recording.releases?.length > 0 ? recording.releases[0].id : "",
+              recording_mbid: recording.id,
+            },
```
MonkeyDo, line 129: I don't know for sure that we want to supply mbid_mapping here, considering it is usually provided by the server rather than manually added.
Semantically, this data we have is not from our MBID mapping, so i would avoid it.

frontend/js/src/search/SongSearch.tsx
```diff
@@ -0,0 +1,166 @@
...
+        join_phrase: ac.joinphrase ?? "",
+      });
+      artistMBIDs.push(ac.artist.id);
+      return null;
+    });
+
+    return (
+      <ListenCard
+        key={recording.id}
+        listen={{
```
MonkeyDo, line 120: For readability, perhaps best to defined the listen above like you did for artistCredit

frontend/js/src/search/UserSearch.tsx
```diff
@@ -1,87 +1,92 @@
...
-          </span>
-        </div>
-      </form>
+  const username = currentUser ? currentUser.name : null;
 
-      <table className="table table-striped">
-        <thead>
+  return (
+    <table className="table table-striped">
+      <thead>
```
MonkeyDo, line 55: Following the mockup, this and other pages might need a `border-bottom: 2px solid;`
Maybe best to have a `.table-strong-header > thead` class or something we can apply more generally to tables like we do `table-striped` for example?

MonkeyDo (changes requested):
Soooo fast!! :rocket: :bullettrain_side:

A couple of comments, but it's already a fully workable page, nice work!
I rebuilt and deployed to test, but I see that we have some sql update script to run in order for pg_trgm similarity to work so haven't tested playlist search.

I have a general naming comment: in most places, we use the terminology "track" over "song". It would probably be best to standardize on one for continuity.

frontend/js/src/search/Search.tsx
```diff
@@ -0,0 +1,174 @@
...
+      </div>
+      <div role="main">
+        <form className="form-group" onSubmit={search}>
+          <h2
+            className="header-with-line"
+            style={{ alignItems: "center", flexWrap: "wrap" }}
+          >
+            <div>
+              Search Results for
+              <strong style={{ marginLeft: "5px" }}>{searchTerm}</strong>
```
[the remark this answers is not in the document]
MonkeyDo: This is what it looks like with the `flex-wrap: nowrap` removed and the following added to the dive that contianse "search results" + the search term:
```css
    text-overflow: ellipsis;
    white-space: nowrap;
    overflow: hidden;
```
image

frontend/js/src/search/PlaylistSearch.tsx
```diff
@@ -0,0 +1,179 @@
...
+                    {(currPageNo - 1) * PLAYLIST_COUNT_PER_PAGE + index + 1}
+                  </td>
+                  <td>
+                    <Link to={`/playlist/${playListID}/`}>
+                      {playlist.title}
+                    </Link>
+                  </td>
+                  <td>{playlist?.annotation}</td>
+                  <td>
+                    <a
```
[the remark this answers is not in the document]
anshg1214: Fixed

frontend/js/src/search/SongSearch.tsx
anshg1214: I did checkout the types which exist right now, but I don't think we can re-use any of those.

frontend/js/src/search/UserSearch.tsx
anshg1214: Sure! I've added this class

frontend/js/src/search/SongSearch.tsx
anshg1214: In case a track has multiple artists, We would need to supply mbid_mapping here. 

Without mbid_mapping: 
image
With mbid_mapping:
image

frontend/js/src/search/PlaylistSearch.tsx
```diff
@@ -0,0 +1,179 @@
...
+            PLAYLIST_COUNT_PER_PAGE,
+            offset
+          );
+        } else {
+          queryData = await APIService.searchPlaylists(
+            searchQuery,
+            PLAYLIST_COUNT_PER_PAGE,
+            offset
+          );
+        }
```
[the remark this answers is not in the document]
anshg1214: I believe having a pill would create confusion among logged in users.
Right now, we have two endpoints
1. Searching All the playlists a user has access to (all public playlists + private playlists by the user / the user has collaborator access on)
2. Searching only public playlists.

If we want to have pills, it'll be better if we replace the first endpoint to search only for private playlists.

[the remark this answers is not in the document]
MonkeyDo: Ah, I see, now I understand what the endpoints actually do, I had misunderstood it as 1. search only my playlists and 2. search only public playlists.

I think then that there is nothing to change, and your implementation is fine :+1:

frontend/js/src/search/SongSearch.tsx
MonkeyDo: I see, thanks for the clarification.
Indeed that feature relies entirely on the mbid_mapping data, which is not ideal but also probably not something we should try to change in this PR,
Objection withdrawn! :D

MonkeyDo (approved):

admin/timescale/updates/2024-05-28-add-pg_trgm-extension.sql
```diff
@@ -0,0 +1,2 @@
+CREATE EXTENSION IF NOT EXISTS "pg_trgm";
+ALTER DATABASE listenbrainz_ts SET pg_trgm.word_similarity_threshold = 0.1;
```
MonkeyDo, line 2: Do you need to copy this configuration line anywhere else? I guess maybe create_extensuions.sql?
See for example to other create_extensions file:
https://github.com/metabrainz/listenbrainz-server/blob/4afff49fbf1d4b57883d9449d2c68714461dec41/admin/sql/create_extensions.sql#L2-L5

listenbrainz/db/playlist.py
```diff
@@ -359,6 +366,174 @@ def get_playlists_collaborated_on(db_conn, ts_conn, user_id: int, include_privat
...
+    )
+    SELECT COUNT(*)
+      FROM playlist_similarities
+     WHERE name_similarity > 0.1
+        OR description_similarity > 0.1
+    """)
+    result = ts_conn.execute(query, params)
+    row = result.fetchone()
+    if row:
+        total_count = row[0]
```
MonkeyDo, line 450: I deployed the PR to test again and ran the required SQL update script so that I could test at scale.
The query is quite slow with the few playlists we have, so I think we should try to  improve the situation a little bit.

In particular, I'm not convinced we need to do a second DB query to get the total count.
I do understand it is currently used for pagination, but there are a couple ways we can achieve a similar UX without doing two queries each time.

1. don't fetch a total count, and instead expect that you will eventually have a request return a set of results with less than the 25 results requested. That would be a cue to disable the "next" button. In practice that means we are settings totalPageCount in the state to currentpage+1, and if we get not enough results we set it to = currentPage which will disable the "next" button.
The drawback is that the user has to click on next only to find out it is now disabled and we've arrived at the end,
2. Similar to 1, we can request 26 results instead of 25 for each query, and if we get < 26 results we know we have run out of pages and we are on the last one, and we can disable the "next" button. Fetching one extra results removes the drawback of solution 1.
3. Add a new search param to the API endpoints to specifically request a total count. If we only pass this parameter when the query term changes, we avoid getting the count for subsequent pagination requests.

Out of all these, I think # 2 is probably the most straightforward but with a good UX, and should reduce the time these queries take with no visible  drawback for the user.

let me know your thoughts.

admin/timescale/updates/2024-05-28-add-pg_trgm-extension.sql
anshg1214: I've already added this line in the create extension [file](https://github.com/metabrainz/listenbrainz-server/pull/2886/files#diff-745f27b3b8521e66884ca62ced68b7307d92f0f5ba37d8154b426f4335337a4cR3), and have added the second command in the [manage.py](https://github.com/metabrainz/listenbrainz-server/pull/2886/files#diff-da5414eec775a578bf2c026b63b632f4de08079068683a672bd9d810199cc48eR187-R191), so I guess this is enough. I haven't tested it, tho

listenbrainz/db/playlist.py
anshg1214: I have tried optimising the queries by the three methods, but the there isn't much difference in the time taken by the endpoint in this [branch](https://github.com/metabrainz/listenbrainz-server/tree/search-page-improvement). As discussed on IRC, let's keep this part for v2

MonkeyDo:
I reckon this is ready to merge and deploy, but I didn't want to press the button in case you had anything that you needed to add on...

anshg1214:
Nope, I don't have anything for this PR, I'll try optimising the SQL query in a separate PR.

PR #2888, opened 2024-05-29
Fix daily jams

amCap1712:
patch api changed in troi breaking daily jams.


mayhem (approved):
Damni, how did I miss this. I looked so hard. :( Thanks for fixing.

PR #2889, opened 2024-05-30
LB-1585: Add Syndication Feeds to ListenBrainz

ericd23:
Tracking PR for https://summerofcode.withgoogle.com/programs/2024/projects/k4Iq0Pnx :smile_cat: 

- [ ] Feed authentication (MB OAuth)
- [ ] Feed UI (including a summary page in the settings?)

Endpoints:
- [ ] My feed (auth needed): `/user/<username>/feed/events`
- [x] User listens:` /user/<username>/listens`
- [x] User stats: `/stats/user/<username>/*`
- [ ] Playlist (updates): `/playlist/<playlist_mbid>`
- [ ] Recommendations (Weekly jams and explorations)
- [x] Fresh releases: `/explore/fresh-releases and /user/<username>/fresh_releases`
- [ ] Stats art


ericd23:
Not much on the PR right now. With the help of https://feedgen.kiesow.be/index.html, I think currently there's no need for a dedicated class to format and generate the feed (which is on my proposal plan). Will need to see if any common patterns for different feeds shows up so that we can abstract them away in a utility class.

ericd23:
So... Will first try to add some feeds that don't need auth. And the first thing to do is to come up with the schema of the feed XML file. It would be great to embed a bunch of related MB links in the feed entry so that ideally a single click takes user to the MB page for that entity.

listenbrainz/webserver/__init__.py
```diff
@@ -407,3 +407,6 @@ def _register_blueprints(app):
     app.register_blueprint(album_bp, url_prefix='/album')
     app.register_blueprint(release_bp, url_prefix='/release')
     app.register_blueprint(release_group_bp, url_prefix='/release-group')
+    
+    from listenbrainz.webserver.views.atom import atom_bp
+    app.register_blueprint(atom_bp, url_prefix='/atom')
```
mayhem, line 412: Lets call the endpoint /feed rather than /atom -- atom is a technology name and it doesn't really inform the user that there are feeds available at this URL.

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,29 @@
+from feedgen.feed import FeedGenerator
```
mayhem, line 1: Nice to have a test! However, I think this should be moved to an actual python test, rather than a function in the views.

mayhem:
What feeds are going to need auth? Because we dont have a lot of private data in LB, so not sure what needs to be authed....

mayhem:
OK, I just had  a quick chat with @amCap1712 on the auth stuff and we need to make a small tweak to your proposal -- namely we can mostly skip authentication stuff. We're in the middle of working on improved OAuth support in MetaBrainz and that should complete sometime during your project. The good news is that that auth now becomes super easy for you -- sometime later on, we will add a python decorator that requires OAuth Authentication for the given endpoint. And all the OAuth work will already be done for you. This shrinks your project a bit, so we should look for other feeds we can create instead. Ping me on IRC (or matrix soon) when you're around and lets coordinate about these changes, ok?

listenbrainz/webserver/__init__.py
ericd23: Good point! I guess most of my feed subscriptions are technical writer and /atom isn't uncommon.

mayhem: Ah, I see, yes /feed can't work. How about /syndication-feed instead?

The reason to not use /atom is because that isn't considere to be RESTful:

  https://restfulapi.net/rest-api-design-tutorial-with-example/

A RESTful API structures its URLs in a way that describes the resources available, not the technology being used to generate the data or the format of the data being returned.

mayhem:
Heya!

I see you've been busily chugging along -- looking good!

The mid-term eval is coming up, so I would like you to come to a stopping point and prepare this PR for a complete review. I would love to have this PR ready for merging before the the evaluation deadline.

I'll start looking over the code in more detail in the meantime and start leaving some comments.

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,242 @@
...
+
+@atom_bp.route("/user/<user_name>/listens", methods=["GET"])
+@crossdomain
+@ratelimit()
+@api_listenstore_needed
+def get_listens(user_name):
+    """
+    Get listens feed for a user.
+
+    :param interval: The time interval in minutes from current time to fetch listens for. For example, if interval=60, listens from the last hour will be fetched. Default is 60.
```
mayhem, line 28: We will need to set an upper limit on this call as well. Giving an argument of 10000000 would take a looong time to execute and would result in a DoS attack on our database server. What would be a good upper limit? A week?

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,242 @@
...
+    if interval < 1:
+        return Response(status=400)
+
+    # Construct UTC timestamp for interval
+    from_ts = datetime.now() - timedelta(minutes=interval)
+
+    listens, _, _ = timescale_connection._ts.fetch_listens(user, from_ts=from_ts)
+
+    fg = FeedGenerator()
+    # TODO: use {base_url} instead of hardcoded url
```
mayhem, line 55: Please take care of TODOs for the mid-term eval.

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,242 @@
...
+    )
+    fg.logo("https://listenbrainz.org/static/img/listenbrainz_logo_icon.svg")
+    fg.language("en")
+
+    # newer listen comes first
+    for listen in reversed(listens):
+        fe = fg.add_entry()
+        # according to spec, ids don't have to be deferencable.
+        fe.id(
+            f"https://listenbrainz.org/syndication-feed/user/{user_name}/listens/{listen.ts_since_epoch}/{listen.data['track_name']}"
```
mayhem, line 72: It would be good to avoid hard coding both the base url as well as the URL path here. Lets use url_for() to abstract these out.

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+@atom_bp.route("/fresh_releases", methods=["GET"])
+@crossdomain
+@ratelimit()
+def get_fresh_releases():
+    """
+    Get site-wide fresh releases.
+
+    :param days: The number of days of fresh releases to show. Max 90 days.
```
mayhem, line 106: For docstrings, it is better to avoid the constants and add a reference to DEFAULT_NUMBER_OF_FRESH_RELEASE_DAYS instead.

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,242 @@
...
+        return Response(status=404)
+
+    past = _parse_bool_arg("past", True)
+    future = _parse_bool_arg("future", True)
+
+    data = db_get_fresh_releases(user["id"])
+    releases = data["releases"] if data else []
+    releases = sorted(releases, key=lambda k: k.get("release_date", ""))
+
+    if not past:
```
mayhem, line 193: Date is possibly parsed twice here. I would parse it once and then use the result in both places.

listenbrainz/webserver/views/atom.py
ericd23: Thanks for point it out. A little math: a song ~ 3 minutes; listening to music nonstop for 8 hours a day => 160 songs. A week gives us 1120 songs. Similar to 
https://github.com/metabrainz/listenbrainz-server/blob/f82daad555a5e26af02bb0b73a7c64cc7767d35f/listenbrainz/webserver/views/api_tools.py#L44
So I think we can handle a week. And for feed reading, a week is more than enough. So I think setting the upper limit to a week is good!

listenbrainz/webserver/views/atom.py
mayhem: Great!

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,570 @@
...
+        recording_mbid = listen.data["additional_info"].get("recording_mbid")
+        artist_name = listen.data["artist_name"]
+        artist_mbid = (
+            listen.data["additional_info"].get("artist_mbids")[0]
+            if listen.data["additional_info"].get("artist_mbids")
+            else None
+        )
+
+        fe = fg.add_entry()
+        # according to spec, ids don't have to be deferencable.
```
mayhem, line 151: What do you mean with deferencable? Thats not an english word I know. :p

mayhem (changes requested):
Starting to look pretty good. Lets try this on our test server.

listenbrainz/webserver/views/atom.py
ericd23: In this case, it basically means the id string doesn't have to be a valid webpage URL, though I'm using HTTP scheme. It just needs to uniquelly identify a feed or feed entry.

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,570 @@
...
+DEFAULT_NUMBER_OF_FRESH_RELEASE_DAYS = 14
+MAX_NUMBER_OF_FRESH_RELEASE_DAYS = 90
+DEFAULT_STATS_ITEMS_PER_GET = 10
+MAX_STATS_ITEMS_PER_GET = 1000
+
+
+atom_bp = Blueprint("atom", __name__)
+
+
+def _external_url_for(endpoint, **values):
```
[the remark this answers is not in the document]
ericd23: OK!

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,570 @@
...
+        )
+
+        fe = fg.add_entry()
+        # according to spec, ids don't have to be deferencable.
+        fe.id(
+            f"{_external_url_for('.get_listens', user_name=user_name)}/{listen.ts_since_epoch}/{track_name}"
+        )
+        fe.title(f"{track_name} - {artist_name}")
+
+        _content = render_template(
```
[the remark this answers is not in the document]
ericd23: Yeah that's a bit weird. 😹

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,570 @@
...
+
+        fe.published(listen.timestamp)
+        fe.updated(listen.timestamp)
+
+    atomfeed = fg.atom_str(pretty=True)
+
+    return Response(atomfeed, mimetype="application/atom+xml")
+
+
+@atom_bp.route("/fresh_releases", methods=["GET"])
```
mayhem, line 185: Please change the URLs to use -, rather than _ for all URLs. More consistent with the rest of our URLs.

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,570 @@
...
+    :param minutes: The time interval in minutes from current time to fetch listens for.
+                    For example, if minutes=60, listens from the last hour will be fetched. Default is 60.
+    :statuscode 200: The feed was successfully generated.
+    :statuscode 400: Bad request.
+    :statuscode 404: The user does not exist.
+    :resheader Content-Type: *application/atom+xml*
+    """
+    user = db_user.get_by_mb_id(db_conn, user_name)
+    if user is None:
+        return Response(status=404)
```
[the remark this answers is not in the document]
ericd23: I don't think feed readers will display the error msg. 
image
But it's good practice and helps with debugging so I include them anyway.

listenbrainz/webserver/templates/atom/top_albums.html
```diff
@@ -0,0 +1,22 @@
+<p>
+    <ul>
+        {% for release in releases %}
+        <li>
+            {% if release.release_group_mbid %}
+                <a href="https://musicbrainz.org/release/{{ release.release_group_mbid }}">{{ release.release_group_name }}</a>
```
amCap1712, line 6: should release-group instead of release

listenbrainz/webserver/views/atom.py
```diff
@@ -0,0 +1,592 @@
...
+    :statuscode 200: The feed was successfully generated.
+    :statuscode 400: Bad request.
+    :statuscode 404: The user does not exist.
+    :resheader Content-Type: *application/atom+xml*
+    """
+    user = db_user.get_by_mb_id(db_conn, user_name)
+    if user is None:
+        return NotFound("User not found")
+
+    minutes = request.args.get("minutes", DEFAULT_MINUTES_OF_LISTENS)
```
amCap1712, line 132: Not sure minutes is a good idea, probably just accept a `from_ts` and forward it to fetch_listens. Maybe also accept `to_ts` (default to now), and add a check that from_ts and to_ts are not more than a week apart or something.

listenbrainz/webserver/templates/atom/top_albums.html
mayhem: Good catch!

listenbrainz/webserver/views/atom.py
amCap1712: Ah right, makes sense in that case.

amCap1712: So instead of accepting `from_ts` just calculate it from `to_ts` and `minutes`.

mayhem: That sounds good.

mayhem:
@ericd23 we're pretty close -- two minor things and we'll merge this!

listenbrainz/webserver/views/atom.py
ericd23: Thanks! But I'm quite sure the supposed change here. For feeds, to_ts is always now(), which is in the code.

amCap1712: Yes, to_ts does not need to change. 

But fetch_listens also accepts a `from_ts`, 

https://github.com/metabrainz/listenbrainz-server/blob/150d2bc59f381a25788892fe172bc2fea6afaed9/listenbrainz/listenstore/timescale_listenstore.py#L178

You should calculate `from_ts = to_ts - timedelta(minutes=minutes)` and then pass it to fetch_listens instead of passing `limit = minutes`.

ericd23: That's what I did at first. But it's very likely that we won't be able to fetch all listens within a time range with the default limit. Is it?

amCap1712: If you want to use minutes that way, then just rename minutes argument to `limit`. But I think using `from_ts` with a default of past 24 hours would be a better option, performance wise.

ericd23: Ah right. Is there a way to only pass `from_ts` and `to_ts` to `fetch_listens` and get all the listens within that range? I see the `LIMIT :limit` in the query and thought that we always have to pass some reasonable limit number or we will be capped by default.

mayhem:
Looking great -- thanks for the last two bits. I'm going to merge this now and we'll just release that into production with the next release. for those of us in the know, we can start using the new features. but we won't actually make noise about them until we finish merging the rest of your project.

PR #2890, opened 2024-05-31
LB-1370: Add accessibility while searching for tracks

anshg1214:
This PR adds keyboard accessibility to the dropdown while searching for tracks on the playlist page

### DIFF
--- a/frontend/css/search-track.less
+++ b/frontend/css/search-track.less
@@ -10,7 +10,6 @@
   border: 1px solid @gray-light; // IE8 fallback
   border: 1px solid rgba(0, 0, 0, 0.15);
   background-color: @white-background;
-  max-height: 102px;
   overflow-y: auto;
   border-radius: 0;
   padding: 0;
--- a/frontend/css/search-track.less
+++ b/frontend/css/search-track.less
@@ -11,24 +11,23 @@
   border: 1px solid rgba(0, 0, 0, 0.15);
   background-color: @white-background;
   max-height: 102px;
-  /*overflow: hidden;*/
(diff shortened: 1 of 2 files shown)


frontend/js/src/utils/SearchTrackOrMBID.tsx
```diff
@@ -185,22 +211,35 @@ export default function SearchTrackOrMBID({
...
-                  onClick={() => {
-                    selectSearchResult(track);
-                    reset();
-                  }}
-                >
-                  {`${track.recording_name} - ${track.artist_credit_name}`}
-                </button>
+          <select
+            className="track-search-dropdown"
+            size={Math.min(searchResults.length, 5)}
```
MonkeyDo, line 216: Let's be a bit more generous with the height of this dropdown
```suggestion
            size={Math.min(searchResults.length, 8)}
```

frontend/css/search-track.less
```diff
@@ -11,24 +11,23 @@
   border: 1px solid rgba(0, 0, 0, 0.15);
   background-color: @white-background;
   max-height: 102px;
```
MonkeyDo, line 13: While in here, could you please remove this line?
Turns out limiting the height of this element is pretty bad UX.

MonkeyDo (commented):
Works great, thanks !

I figured the UI for this results dropdown was pretty poor, showing only a small sliver.
Let's instead be  abit more generous with it, since you're in here making changes.

MonkeyDo (approved):
Good to go, thank you !

PR #2861, opened 2024-05-01
Upgrade dataset hoster

amCap1712:
The new dataset hoster adds typing support but also changes the data format. Hence, queries need to be refactored.

PR #2893, opened 2024-05-31
fix: Music Neighbourhood

anshg1214:
This PR fixes the Music Neighborhood page which broke after the labs endpoint was updated in #2861


MonkeyDo (approved):
Looks good, thanks !

PR #2894, opened 2024-05-31
LB-1401: Manually submit album

MonkeyDo:
This new component allows users to submit an album (or selection of tracks within) at once.
Like the single track submission, it allows searching by text or pasting a release or release group MBID.

Search:
image

After selection:
image

Works for releases with multiple media:
image

The PR also adds the ability to select a specific release for the existing single listen submissions, and adds the relevant MBIDs to the listen metadata:
Collapsed by default:
image
Expanded:
image


MonkeyDo:
@anshg1214 I deployed this to test.LB, but in order to see it the submission actually work you will need to change the api url in the APIService in the globalcontext.
For testing the looks and other things without fearinngto submit a bunch of crappy listens you can just test normally on test.LB.
It will say the submission went through but test.LB does not ingest listens so they won't appear anywhere.

Aerozol:
Amazing <3

One question, that I can't figure out without the data turning up, is how it handles releases without track times? The Picard listen submission plugin submits every 0:00 track in the release at the same time stamp. I would rather default to 1 minute track length, or similar.

Without track lengths it might also be useful to display question marks instead of the track times (currently a blank space), perhaps with a rollover/tooltip saying "no track length data, listen will be submitted as if 1 minute long", or similar.

Screenshot - oh and my search results look different to yours! Fun:
image

Side-note: I will be really curious to see some stats eventually about how many users track vinyl/cassette/etc listens!

anshg1214 (changes requested):
The PR works pretty good!

Can we have a checkbox at the top of the releases to check/uncheck all the tracks? In case the user wants to select 5/20 tracks, it makes the process cumbersome.
image

MonkeyDo:
Updated test.LB with everything mentioned fixed/improved (I think that's everything!)

MonkeyDo:
After testing it again on test.LB, I think maybe the way I set the time is not right.
Currently setting a time and date for submitting an album, I set the timestamp of the *last* track with that time and work backwards for the rest of the album.
It made sense in my head when I code it, but after using it it feels like a weird choice. Maybe I can add a choice for the custom time between  "starts at" and "ends at"?

Aerozol:
1. Hmm being able to set the 'ended' time is pretty cool imo! But not at all what I expected from the UI.

I think it should default to when the album started, but a toggle to set it to ended (for people who are really on the ball with their submissions) would be excellent.

2. The reason I would go for a shorter time, like 1 minute, isn't because I think that's the average track length, but because I think it is less likely to cause problems. I would potentially even go shorter.

e.g. If I know I've listened to album A at 1:00 and album B at 1:30. I don't _really_ care if LB thinks album A finished at 1:10. But I do care if it thinks album A went to 2:00, and now it looks like I was listening to both albums at once.

Really any short time (that makes the listens go in order, rather than all at once) will do for me :)

3. I love the clock popup!! Even if I keep trying to click on it :P

MonkeyDo:
Thanks for all the useful feedback!

I clarified the default duration a bit, changed it to 1 minute and in the modal replaced the question mark with the default duration and displayed it differently, with a little caption at the top of the column. On hover that caption popover text is "When no duration is available a default of 60 seconds will be used"
This default duration caption is not shown unless there is at least one track in one medium missing a length.
Hopefully that's a lot clearer, although I'm not 100% on the phrasing of "default 60 seconds" :
image

I also added the total length for each medium next to the medium name with a little clock icon, with the default duration applied for tracks that don't have a duration.
I think it also adds a bit of information regarding what you said in point 2, helping users calculate time:
image

And finally, the "starts at" and "finishes at" options (defaulting to "starts at" as discussed):
image

MonkeyDo:
OK, I think there is one last issue I want to resolve before considering this done:
How am I supposed to differentiate between these releases? :zany_face: 
image

anshg1214 (changes requested):
Some final changes

Aerozol:
Full date, country, label, looks great! Is it an easy change to 50% tint that information (maybe italics as well, but might be too much)?

MonkeyDo:
> Full date, country, label, looks great! Is it an easy change to 50% tint that information (maybe italics as well, but might be too much)?
> 
> edit: I don't love the country flags everywhere on MB tbh, can be distracting

~~I tried to do something like that, make it smaller and more subdued colour, but unfortunately inside an HTML select element I can style the text for the entire line, but not for portions of each line separately. It's all or nothing :/~~

Scratch that, I found a fancy workaround !
Looks much nicer and much more readable, thanks for pushing me to explore alternatives again !
image

anshg1214 (approved):
LGTM

MonkeyDo:
Thanks for the help both of you !

PR #2895, opened 2024-05-31
Bulk recording metadata lookup

amCap1712:
POST endpoint for looking up multiple recording mbids at once, useful in troi for resolving mbids for imported playlist data.

### DIFF
--- a/listenbrainz/webserver/views/metadata_api.py
+++ b/listenbrainz/webserver/views/metadata_api.py
@@ -5,9 +5,11 @@ from flask import Blueprint, request, jsonify, current_app
 from listenbrainz.db.mbid_manual_mapping import create_mbid_manual_mapping, get_mbid_manual_mapping
 from listenbrainz.db.metadata import get_metadata_for_recording, get_metadata_for_artist, get_metadata_for_release_group
 from listenbrainz.db.model.mbid_manual_mapping import MbidManualMapping
-from listenbrainz.labs_api.labs.api.artist_credit_recording_lookup import ArtistCreditRecordingLookupQuery
+from listenbrainz.labs_api.labs.api.artist_credit_recording_lookup import ArtistCreditRecordingLookupQuery, \
+    ArtistCreditRecordingLookupInput
 from listenbrainz.labs_api.labs.api.artist_credit_recording_release_lookup import \
     ArtistCreditRecordingReleaseLookupQuery, ArtistCreditRecordingReleaseLookupInput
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/views/metadata_api.py
```diff
@@ -295,6 +297,137 @@ def get_mbid_mapping():
     return jsonify({})
 
 
+def make_acrr_input(recording):
+    if "release_name" in recording:
```
mayhem, line 301: docstrings

listenbrainz/webserver/views/metadata_api.py
```diff
@@ -295,6 +297,137 @@ def get_mbid_mapping():
...
+    return all_results, all_params
+
+
+@metadata_bp.route("/lookup/", methods=["POST"])
+@crossdomain
+@ratelimit()
+def get_mbid_mapping_post():
+    """
+    This endpoint is the POST version for looking up recording mbids and associated MusicBrainz data. It allows up to
+    max number of items allowed. (:data:`~webserver.views.api.MAX_ITEMS_PER_GET` items)
```
mayhem, line 360: I think for this endpoint we need to lower the max number of items per call. Sending us 1,000 items to look up and then doing so in quick succession is basically a DoS attack. :(

listenbrainz/webserver/views/metadata_api.py
```diff
@@ -295,6 +297,137 @@ def get_mbid_mapping():
...
+
+        all_params[idx] = recording
+        all_results.append({
+            "artist_name_arg": artist_name,
+            "recording_name_arg": recording_name,
+            "release_name_arg": release_name,
+            "index": idx
+        })
+
+    try:
```
mayhem, line 415: The scalability of this really scares me. :(

mayhem (approved):

listenbrainz/webserver/views/metadata_api.py
amCap1712: The aim is to support import of playlists using Troi, an alternative would be to have import logic in LB codebase instead but that would be less ideal. I have reduced the limit to 50 items per call.

listenbrainz/webserver/views/metadata_api.py
amCap1712: done

listenbrainz/webserver/views/metadata_api.py
amCap1712: True, we'll have to work out something better eventually.

PR #2898, opened 2024-06-05
Fix importing spotify playlists

amCap1712:
Currently, looking up spotify tracks in mapping errors out due to rate limiting. Lookup tracks for import using the POST bulk lookup API instead.


mayhem (approved):
LGTM!

PR #2901, opened 2024-06-07
Reinstate "report user" button

MonkeyDo:
We removed it previously, but I do not remember the reason.
It's needed back now, I ran into a suspected bot user and was looking for the button.
I ended up using react devtools to access the APIService in the global context in order to report the user, in other words not exactly ideal for a regular user...

### DIFF
--- a/frontend/js/src/user/Dashboard.tsx
+++ b/frontend/js/src/user/Dashboard.tsx
@@ -39,6 +39,7 @@ import {
 } from "../utils/utils";
 import FollowButton from "./components/follow/FollowButton";
 import { RouteQuery } from "../utils/Loader";
+import ReportUserButton from "../report-user/ReportUser";
 
 export type ListensProps = {
   latestListenTs: number;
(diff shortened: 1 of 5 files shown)


frontend/js/src/user/Dashboard.tsx
```diff
@@ -468,6 +471,12 @@ export default function Listen() {
               />{" "}
               MusicBrainz
             </Link>
+            {user && !isCurrentUsersPage && (
+              <ReportUserButton
+                user={user}
+                alreadyReported={already_reported_user}
+              />
+            )}
```
anshg1214, line 479: We should check if the user is logged in or not, then only they should be allowed to report a user. 
The best way is to check if the user is logged in the modal, and accordingly redirect user to the login page first if they're not logged in.

anshg1214 (changes requested):

frontend/js/src/user/Dashboard.tsx
MonkeyDo: I decided to show a link to the user in a toast message, in case they were doing something else on the page they didn't want to stop, like playing music.
Also made the link redirect back to the current page, which is something we should probably do in other cases in the website where we require login for actions, but that's for another day.

anshg1214 (approved):
LGTM

PR #2902, opened 2024-06-08
LB-1565: Add playlist collaborators feature broken

anshg1214:
LB-1565

### DIFF
--- a/frontend/js/src/common/UserSearch.tsx
+++ b/frontend/js/src/common/UserSearch.tsx
@@ -42,12 +42,8 @@ export default function UserSearch(props: UserSearchProps) {
     [APIService]
   );
 
-  const throttledSearchUsers = React.useCallback(
-    (newUserName: string) => {
-      _throttle(async () => {
-        await searchUsers(newUserName);
-      }, 300)();
-    },
+  const throttledSearchUsers = React.useMemo(
+    () => _throttle(searchUsers, 800),
     [searchUsers]
   );
 
--- a/frontend/js/src/common/UserSearch.tsx
+++ b/frontend/js/src/common/UserSearch.tsx
@@ -1,10 +1,7 @@
 import * as React from "react";
-import { faUser } from "@fortawesome/free-solid-svg-icons";
 import { throttle as _throttle } from "lodash";
 import { toast } from "react-toastify";
-import debounceAsync from "debounce-async";
 import GlobalAppContext from "../utils/GlobalAppContext";
-import ListenControl from "./listens/ListenControl";
 import { ToastMsg } from "../notifications/Notifications";
 
 export type UserSearchProps = {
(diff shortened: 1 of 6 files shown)


frontend/js/src/common/UserSearch.tsx
```diff
@@ -14,97 +11,131 @@ export type UserSearchProps = {
...
-  throttledSearchUsers = _throttle(async () => {
-    await this.searchUsers();
-  }, 300);
+  // Refs
+  const dropdownRef = React.useRef<HTMLSelectElement>(null);
 
-  searchUsers = async () => {
-    const { APIService } = this.context;
-    const { newUser } = this.state;
+  const searchUsers = async () => {
```
MonkeyDo, line 31: We can memoize this function and pass it the user name to search as an argument:
```tsx
  const searchUsersFunction = React.useCallback(
    async (username: string) => {
      try {
        const response = await searchUsers(username);
        setUserSearchResults(response.users);
      } catch (error) {
        toast.error(<ToastMsg title="Error" message={error.message} />, {
          toastId: "error",
        });
      }
    },
    [searchUsers]
  );
```

And below also memoize the throttled function:
```jsx
const throttledSearchUsers = React.useMemo(
    () => _throttle(searchUsersFunction, 300),
    [searchUsersFunction]
  );
```

Note that I deconstructed `searchUsers` from APIService and as a consequence I renamed the function to `searchUsersFunction`. Naming can probably be improved

frontend/js/src/common/UserSearch.tsx
```diff
@@ -14,97 +11,131 @@ export type UserSearchProps = {
...
+      reset();
+    }
+  };
+
+  React.useEffect(() => {
+    if (!newUser) {
+      return;
+    }
+    throttledSearchUsers();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
MonkeyDo, line 77: Probably want to remove this eslint-disable and add the relevant dependencies (throttledSearchUsers) in the array, especially now that it's all memoized and won't create unnecessary rerenders.

frontend/js/src/common/UserSearch.tsx
```diff
@@ -14,97 +11,131 @@ export type UserSearchProps = {
...
-          className="form-control"
-          name="newUser"
-          onChange={this.handleInputChange}
-          placeholder={placeholder}
-          value={newUser}
-          aria-haspopup={Boolean(userSearchResults?.length)}
-        />
-        <div className="search-dropdown">
-          {userSearchResults?.map((user) => {
+  const handleKeyDown = (event: React.KeyboardEvent) => {
```
MonkeyDo, line 57: Thanks for adding the accessibility improvements here too !

If we run into it again it might make sense to create a custom component for selects that implement these mechanisms, to be reused.

Or maybe something along the lines of this hook? https://github.com/dealroom/react-arrow-key-navigation-hook

frontend/js/src/common/UserSearch.tsx
```diff
@@ -14,97 +11,131 @@ export type UserSearchProps = {
...
+      option.scrollIntoView({ block: "nearest" });
+    }
+  }, [selectedIndex]);
+
+  return (
+    <div
+      className="input-group user-search"
+      style={{
+        width: "100%",
+      }}
```
MonkeyDo, line 92: The css class `.input-group-flex` is helpful here:
```suggestion
      className="input-group input-group-flex user-search"
```

frontend/js/src/playlists/components/CreateOrEditPlaylistModal.tsx
```diff
@@ -246,7 +246,7 @@ export default NiceModal.create((props: CreateOrEditPlaylistModalProps) => {
       (isEdit
         ? playlist?.creator.toLowerCase() === user.toLowerCase()
         : currentUser.name.toLowerCase() === user.toLowerCase());
-    if (!disabled && collaborators.includes(user)) {
+    if (!disabled && !collaborators.includes(user)) {
```
MonkeyDo, line 249: Also sums up my reaction seeing this bug: '!'

MonkeyDo (commented):
Looking good, thanks for fixing the main bug and for the rest of the improvements !

frontend/js/src/common/UserSearch.tsx
MonkeyDo: This article might be interesting: https://css-tricks.com/striking-a-balance-between-native-and-custom-select-elements/

I ended up there while trying to figure out something about selects while working on #2894

MonkeyDo (commented):
One last optimization and we're good to go!

frontend/js/src/common/UserSearch.tsx
```diff
@@ -14,97 +12,96 @@ export type UserSearchProps = {
...
-    );
-  };
+  const throttledSearchUsers = React.useCallback(
+    (newUserName: string) => {
+      _throttle(async () => {
+        await searchUsers(newUserName);
+      }, 300)();
+    },
+    [searchUsers]
+  );
```
MonkeyDo, line 52: As far as I can tell, this does not throttle anything currently, because you are recreating the throttled function every call I think.
Basically, throttle returns a function, so currently every time `throttledSearchUsers` is called you call `throttle` and executed the function it returns, meaning no throttling is happening. See the network requests for each letter I typed:
image

Instead, I suggest using `useMemo` to memoize the throttled function with arbitrary arguments, rather than `useCallback` which requires you to pass at least one argument and complicates everything.

```suggestion
  const throttledSearchUsers = React.useMemo(
    () => _throttle(searchUsers, 800),
    [searchUsers]
  );
```
* Note that I also increased the throttle wait time to 800ms, 300 was just too low

The throttled result, typing really fast:
image

anshg1214: Yes, this does fix the throttling issue. Thanks for the fix.

MonkeyDo (approved):
Great, thanks for fixing this!

PR #2905, opened 2024-06-11
Improve Apple Music search with fuzzy matching

MonkeyDo:
A user reported in the forums that the apple player sometimes plays the wrong track even though they have the right track in the catalog.
While testing with good examples I identified two issues:
1. having the album title in the search terms can lead to worse results because there is no way to specify it's the title; adding a bunch of words to the search makes results unpredictable. Take for example an album with named one of the songs that was a radio hit; if you are looking for a less know song, having the popular song name (album title) in the search terms will result in the popular song being the first result.
2. Considering the above and that apple music tends to push the music they want you to listen to rather than the best match (I tested that), ;I added a small fuzzy matching library to try to find the best possible match in the API results.

Regarding 1 it's no surprise: we have already removed the album title from spotify search for example, for the same reasons, even though spotify allows specifying parts for track title/album title/artist name.

Regarding 2: I will very likely be adding the fuzzy matching strategy to youtube player, whose API results are all over the place

### DIFF
--- a/frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
@@ -1,7 +1,8 @@
 import * as React from "react";
-import { get as _get, isString } from "lodash";
+import { get as _get, escapeRegExp, isString } from "lodash";
 import { faApple } from "@fortawesome/free-brands-svg-icons";
 import { Link } from "react-router-dom";
+import fuzzysort from "fuzzysort";
 import {
   getArtistName,
   getTrackName,
@@ -190,11 +191,27 @@ export default class AppleMusicPlayer
         `/v1/catalog/{{storefrontId}}/search`,
         { term: searchTerm, types: "songs" }
       );
-      const apple_music_id = response?.data?.results?.songs?.data?.[0]?.id;
-      if (apple_music_id) {
-        await this.playAppleMusicId(apple_music_id);
+      const candidateMatches = response?.data?.results?.songs?.data;
+      // Check if the first API result is a match
+      if (
+        new RegExp(escapeRegExp(`${trackName}`), "ig").test(
+          candidateMatches?.[0]?.attributes.name
+        )
+      ) {
+        // First result matches track title, assume it's the correct result
+        await this.playAppleMusicId(candidateMatches[0].id);
         return;
       }
+      // Fallback to best fuzzy match based on track title
+      const fruzzyMatches = fuzzysort.go(trackName, candidateMatches, {
+        key: "attributes.name",
+        limit: 1,
+      });
+      if (fruzzyMatches[0]) {
+        await this.playAppleMusicId(fruzzyMatches[0].obj.id);
+        return;
+      }
+      // No good match, onTrackNotFound will be called in the code block below
     } catch (error) {
       console.debug("Apple Music API request failed:", error);
     }
--- a/frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
@@ -178,8 +178,9 @@ export default class AppleMusicPlayer
     const { onTrackNotFound } = this.props;
     const trackName = getTrackName(listen);
     const artistName = getArtistName(listen);
-    const releaseName = _get(listen, "track_metadata.release_name");
-    const searchTerm = `${trackName} ${artistName} ${releaseName}`;
+    // Album name can give worse results, re;oving it from search terms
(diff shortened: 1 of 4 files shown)


regorxxx:
Just a comment about this:
https://tickets.metabrainz.org/projects/LB/issues/LB-1592

i.e. if changes are being made for matching against Apple Music and the logic is similar to the one used at Spotify, the latter has some errors with non latin chars. Maybe it is not related at all, but now that you are revising it, just to be sure.

MonkeyDo:
> Just a comment about this: https://tickets.metabrainz.org/projects/LB/issues/LB-1592
> 
> i.e. if changes are being made for matching against Apple Music and the logic is similar to the one used at Spotify, the latter has some errors with non latin chars. Maybe it is not related at all, but now that you are revising it, just to be sure.

Thanks for pointing this out !
I will look into the (lack of) support of accented and non-latin characters in this fuzzy search library.

regorxxx:
Also note current fuzzy search per last commit uses the track name as-is. 
While tracks on spotify/Apple may have both Latin and non Latin titles (for ex. Japanese tracks). The issue linked above has such examples. 

That means:
- The fuzzy search and deburr funcs must be compatible with such non ascii chars to work. (just to be sure on your tests)
- Is still required to not only match the search results against the track name but also against the track name variations, found at Music Brainz (which usually have both original and romanized versions of the title). So a loop of regexps.

MonkeyDo:
Thanks for being on the lookout @regorxxx !
Both the fuzzy search and deburr functions do support unicode and will behave properly with non-latin character sets.

However the second part (matching on MB aliases/variations) we will need more work on a different part of the codebase.
Just to clarify that I am not ignoring the issue with potential mismatch in latin/non-latin track names, but it's a wider effort across how we deal with recording metadata and I won't be able to address it in this PR.

regorxxx:
@MonkeyDo Understandable. The issue is here https://tickets.metabrainz.org/projects/LB/issues/LB-1592). 

Back to this PR , related only to fuzzy search.

Current methods may not properly match variations of title which have things like feat., etc. For ex:
- This is a song feat. Deep Purple
- This is a song (feat. Deep Purple)
- This is a song with Deep Purple
- This is a song

It depends on the score for the fuzzy search, which may properly match or not in some cases.

```const fuzzy = require('..\\helpers-external\\fuzzysort\\fuzzysort');
// This will match both
let candidates = ['My Title feat. Deep Purple', 'My Title (feat. Deep Purple)'];
console.log(fuzzy .go('My title', candidates).map(o => o.target + ': ' + o.score));

// This will match only the second one
candidates = ['My Title', 'My Title (feat. Deep Purple)'];
console.log(fuzzy .go('My title feat. Deep Purple', candidates).map(o => o.target + ': ' + o.score));

// This will match both and prefer the first one
candidates = ['My Title', 'My Title (feat. Deep Purple)'];
console.log(fuzzy .go('My Title', candidates).map(o => o.target + ': ' + o.score));
```

Note how the second example totally misses a candidate. This kind of error is the norm between Spotify/Apple/Listenbrainz since they handle titles in different ways.

To avoid it, you may either clean these situations with Regexp and removing feat., parentheses, etc. or just perform an inverse fuzzy. i.e. perform a fuzzy search for every candidate against the original one. That's the third example, where My Title does match 'My Title (feat. Deep Purple)' but the opposite is not true.

Since the fuzzy library already outputs the target matches, you may simply skip all results and any target not passing the threshold would need an inverse fuzzy.

regorxxx:
Ok, found one of the main problems which may have gone unnoticed:
https://github.com/farzher/fuzzysort/issues/55

The fuzzy library DOES NOT support accents. Please perform some tests with the library since it's not working as intended.

```
// This will match only the second one, note how 'titlé' does not match 'title'
candidates = ['My Titlé', 'My Title (feat. Deep Purple)'];
console.log(fuzzy .go('My Title', candidates).map(o => o.target + ': ' + o.score));
```

MonkeyDo:
> Ok, found one of the main problems which may have gone unnoticed: [farzher/fuzzysort#55](https://github.com/farzher/fuzzysort/issues/55)
> 
> The fuzzy library DOES NOT support accents. Please perform some tests with the library since it's not working as intended.
> 
> ```
> // This will match only the second one, note how 'titlé' does not match 'title'
> candidates = ['My Titlé', 'My Title (feat. Deep Purple)'];
> console.log(fuzzy .go('My Title', candidates).map(o => o.target + ': ' + o.score));
> ```

Indeed accents are not supported, but we use[ lodash's deburr method](https://lodash.com/docs/4.17.15#deburr) which removes accents.

I will look into adding cleanup regexps to remove "feat." and such.

PR #2906, opened 2024-06-12
Fix search bar icon position in navigation tray

MonkeyDo:
Reported by outsidecontext [in the community forums](https://community.metabrainz.org/t/search-for-artists-albums-and-tracks-on-listenbrainz/699675/4).

The absolute positioning did not work in this case on Firefox, so I reworked the CSS for this element to use flexbox instead, and a couple of changes to make it look and act like before (especially the inset shadows).

Before:
image
After:
image

### DIFF
--- a/frontend/css/search.less
+++ b/frontend/css/search.less
@@ -4,28 +4,29 @@
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
-  display: initial;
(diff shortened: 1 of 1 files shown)


anshg1214:
@MonkeyDo Since this CSS is also being used in the search page, this change broke the search bar there 

image

PR #2913, opened 2024-06-20
Update chat link in footer

MonkeyDo:
Point to the wiki page for all things comunications
image

### DIFF
--- a/frontend/js/src/components/Footer.tsx
+++ b/frontend/js/src/components/Footer.tsx
@@ -29,7 +29,7 @@ export default function Footer() {
               <li className="color-a">
                 <span className="color-gray">Chat with us: </span>{" "}
                 <a
-                  href="https://wiki.musicbrainz.org/Communication/ChatBrainz"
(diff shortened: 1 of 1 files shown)


phw:
This is also more inviting then the previous "IRC: #metabrainz", which likely a lot of users didn't understand what it is about :D

frontend/js/src/components/Footer.tsx
```diff
@@ -27,14 +27,13 @@ export default function Footer() {
             </p>
             <ul className="list-unstyled">
               <li className="color-a">
-                <span className="color-gray">IRC: </span>{" "}
+                <span className="color-gray">Chat with us: </span>{" "}
                 <a
-                  href="https://kiwiirc.com/nextclient/irc.libera.chat/?#metabrainz"
+                  href="https://wiki.musicbrainz.org/Communication/ChatBrainz"
```
phw, line 32: I think the link should go to the doc page, not directly the wiki.

https://musicbrainz.org/doc/Communication/ChatBrainz

PR #2914, opened 2024-06-20
LB-1448: Manually submit a queue listens

MonkeyDo:
### Based on #2912, wait until it is merge first, or if this one is merged sooner,  please close the other one
We have been asked to add a way to queue up listens in the manual submission modal, to make it easier for users and help calculate the listening time properly based on the tracks' duration,
Build up on the work in #2894, still allow users to select a specific release for each listen in the queue.

image


anshg1214 (approved):
LGTM!!

PR #2918, opened 2024-06-24
Add a background job processor for Spark Reader

amCap1712:
To handle long-running tasks without skipping heartbeats in the main thread, a background message processor that runs in a separate thread.

This setup is useful when processing a message (even rarely) can take over a minute. Processing such tasks in the main thread can cause missed heartbeats which in turn lead to the connection being restarted and the same messages being redelivered to the consumer setting up a loop. The rabbitmq queue essentially gets stuck and the consumer keeps processing the same messages again and again.

This processor runs in its own thread. However, that presents another challenge because messages need to be ack-ed from the same thread on which they were received (the restriction is about channels ack-ing from the same channel but channels are not thread-safe, hence same thread). Here's a brief explanation of how the overall flow works:

1. The RabbitMQ consumer listens for messages on the main thread.
2. Upon receiving a message, the RabbitMQ callback enqueues the message into the processor's message queue.
3. The processor handles the message and, upon completion, enqueues the message into its acknowledgment queue.
4. The main thread periodically checks the processor's acknowledgment queue for completed messages and acknowledges them.

### DIFF
--- a/listenbrainz/spark/background.py
+++ b/listenbrainz/spark/background.py
@@ -1,0 +1,200 @@
+import json
+import logging
+from queue import Queue, Empty
+from threading import Thread
+
+import orjson
+import sentry_sdk
+
+from listenbrainz.spark.handlers import (
+    handle_candidate_sets,
+    handle_dataframes,
+    handle_dump_imported,
+    handle_model,
+    handle_recommendations,
+    handle_user_daily_activity,
+    handle_user_entity,
+    handle_user_listening_activity,
+    handle_sitewide_entity,
+    notify_artist_relation_import,
+    notify_mapping_import,
+    handle_missing_musicbrainz_data,
+    cf_recording_recommendations_complete,
+    handle_sitewide_listening_activity,
+    handle_similar_users,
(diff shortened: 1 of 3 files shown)


listenbrainz/spark/background.py
```diff
@@ -0,0 +1,200 @@
...
+         ack queue if successfully processed.
+        """
+        with self.app.app_context():
+            while not self.done:
+                try:
+                    message = self.internal_message_queue.get(block=True, timeout=5)
+                    self.process_message(message)
+                    self.internal_message_ack_queue.put(message)
+                except Empty:
+                    self.app.logger.debug("Empty internal message queue")
```
mayhem, line 114: You probably don't want this in production long term...

mayhem (approved):
Skimmed, but looks reasonable!

listenbrainz/spark/background.py
amCap1712: Yeah, that's why I set it to debug log level and we log by default at INFO level. Its only for debugging in future if something hangs.

PR #2920, opened 2024-06-24
LB-1597: 'New listens since you arrived' carries over between profiles

MonkeyDo:
# Problem

When you load a user's dashboard and they are listening to music, you will see new listens arrive in a separate list.
Now if you navigate to another user's page the list will stay, showing the first user's listens.

# Solution
Reset the webSocketListens array when user name changes

Also did some minimal cleanup of some react dependencies warnings and memoized a function.

### DIFF
--- a/frontend/js/src/user/Dashboard.tsx
+++ b/frontend/js/src/user/Dashboard.tsx
@@ -106,31 +106,34 @@ export default function Listen() {
 
   const queryClient = useQueryClient();
 
-  const receiveNewListen = (newListen: string): void => {
-    let json;
-    try {
-      json = JSON.parse(newListen);
-    } catch (error) {
-      toast.error(
-        <ToastMsg
-          title="Coudn't parse the new listen as JSON: "
-          message={error?.toString()}
-        />,
-        { toastId: "parse-listen-error" }
-      );
-      return;
-    }
-    const listen = formatWSMessageToListen(json);
+  const receiveNewListen = React.useCallback(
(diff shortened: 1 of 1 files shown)


frontend/js/src/user/Dashboard.tsx
```diff
@@ -241,6 +246,8 @@ export default function Listen() {
   React.useEffect(() => {
     // On first load, run the function to load the metadata for the playing_now listen
     if (playingNow) updatePlayingNowMutation(playingNow);
+    // no exhaustive-deps because we only want ot run this on initial start
```
kellnerd, line 249:
```suggestion
    // no exhaustive-deps because we only want to run this on initial start
```
:grin:

frontend/js/src/user/Dashboard.tsx
```diff
@@ -106,31 +106,34 @@ export default function Listen() {
...
-  };
+  const receiveNewListen = React.useCallback(
+    (newListen: string): void => {
+      let json;
+      try {
+        json = JSON.parse(newListen);
+      } catch (error) {
+        toast.error(
+          <ToastMsg
+            title="Coudn't parse the new listen as JSON: "
```
kellnerd, line 117:
```suggestion
            title="Couldn't parse the new listen as JSON: "
```
Was already there before, but still

kellnerd (changes requested):

anshg1214 (approved):

PR #2921, opened 2024-06-24
LB-1512: Display total duration of playlist

MonkeyDo:
Calculate the total duration of a playlist using track metadata, sum it up and display it in a human-friendly manner:
image

### DIFF
--- a/frontend/js/src/playlists/Playlist.tsx
+++ b/frontend/js/src/playlists/Playlist.tsx
@@ -389,8 +389,10 @@ export default class PlaylistPage extends React.Component<
               </h1>
               <div className="info">
                 <div>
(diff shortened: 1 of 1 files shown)


frontend/js/src/playlists/Playlist.tsx
```diff
@@ -378,7 +388,10 @@ export default class PlaylistPage extends React.Component<
                 </small>
               </h1>
               <div className="info">
-                <div>{playlist.track?.length} tracks</div>
+                <div>
+                  {playlist.track?.length} tracks&nbsp;-&nbsp;
+                  {totalDurationForDisplay}
+                </div>
```
anshg1214, line 394: You should hide the hyphen symbol if the playlist is empty.

anshg1214 (changes requested):

PR #2927, opened 2024-07-03
Manual submission modal improvements and bug fixes

MonkeyDo:
- Sending a `null` or non-integer value for duration_ms in the listen payload results in an error and the listens not submitted.
Reported in the forums https://community.metabrainz.org/t/add-manual-album-listens-offline/696894/19
Thanks Nicknickcdf!
- default durations were not scrolling with the rest of the medium content because of position:absolute without position:relative on their parent element.
- send tracknumber string instead of forcing to integer; the API docs say they expect an integer but track.number in MB is a string and the DB model also expects optional string
- if track duration is not available, use associated recording's duration.
- set the release MBID correclty in `track_metadata.additional_info.release_mbid` instead of `track_metadata.release_mbid`
- display MB `track.number` string instead of `track.position` integer, better for some mediums (cassettes, etc.)
- while doing that, fixed the API docs for `tracknumber` which expected an optional integer, to an optional string as the DB model expects

All of this feedback has been reported on https://tickets.metabrainz.org/browse/LB-1401 or in the forums.

### DIFF
--- a/docs/users/json.rst
+++ b/docs/users/json.rst
@@ -206,7 +206,7 @@ The following optional elements may also be included in the ``additional_info`` 
      - array of strings
      - A list of MusicBrainz Work IDs that may be associated with this recording.
    * - ``tracknumber``
-     - integer
+     - string
      - The tracknumber of the recording. This first recording on a release is tracknumber 1.
    * - ``isrc``
      - string
--- a/frontend/js/src/user/components/AddAlbumListens.tsx
+++ b/frontend/js/src/user/components/AddAlbumListens.tsx
@@ -38,9 +38,7 @@ function TrackRow({ track, isChecked, onClickCheckbox }: TrackRowProps) {
         }}
         checked={isChecked}
       />
-      <strong className="small track-number">
-        {padStart(track.position.toString(), 2, "0")}
-      </strong>
+      <strong className="small track-number">{track.number}</strong>
       <span>{track.title}</span>
       <span className={`duration ${!track.length ? "default-duration" : ""}`}>
         {millisecondsToStr(track.length ?? DEFAULT_TRACK_LENGTH_SECONDS * 1000)}
--- a/frontend/js/src/user/components/AddAlbumListens.tsx
+++ b/frontend/js/src/user/components/AddAlbumListens.tsx
@@ -213,7 +213,7 @@ export default function AddAlbumListens({
   );
 
   const allDurations = selectedAlbum?.media.flatMap((medium) =>
-    medium.tracks.map((track) => track.length)
+    medium.tracks.map((track) => track.length ?? track.recording.length)
   );
   const showDefaultDuration = !allDurations?.every(Boolean);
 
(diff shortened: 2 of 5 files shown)


docs/users/json.rst
```diff
@@ -206,7 +206,7 @@ The following optional elements may also be included in the ``additional_info``
      - array of strings
      - A list of MusicBrainz Work IDs that may be associated with this recording.
    * - ``tracknumber``
-     - integer
+     - string
```
kellnerd, line 209: Looks good, listen submissions with alphanumeric track numbers are accepted and displayed fine.
I already updated my submission client to make use of this 😎

frontend/js/src/user/components/AddListenModal.tsx
```diff
@@ -65,15 +68,21 @@ export const getListenFromTrack = (
       additional_info: {
         recording_mbid: track.recording.id,
         submission_client: "listenbrainz web",
-        duration_ms: track.length,
         track_mbid: track.id,
-        tracknumber: parseInt(track.number, 10),
+        tracknumber: track.number,
```
kellnerd, line 72: That was the issue with the `null` values I was describing in the comments on Jira.
I guess I was testing with alphanumeric track numbers for which `parseInt` returns `NaN`.

kellnerd (approved):
Thanks a lot for these improvements!

frontend/js/src/user/components/AddListenModal.tsx
MonkeyDo: I figured that's what it was, but wasn't sure.

I ended up treating tracknumbers that look like actual numbers differently for display (I wanted to pad the numeric track numbers) in 59ac4dc43

MonkeyDo:
@kellnerd I also added the shift key multi-checkbox selection thing you mentioned, it's pretty neat and made me do some interesting coding :) cf9389518
Building the image now to redeploy to test.LB

frontend/js/src/user/components/AddAlbumListens.tsx
```diff
@@ -22,24 +38,49 @@ type MBReleaseWithMetadata = MusicBrainzRelease &
...
         checked={isChecked}
       />
       <strong className="small track-number">
-        {padStart(track.position.toString(), 2, "0")}
+        {isFinite(Number(track.number))
+          ? padStart(track.position.toString(), 2, "0")
+          : track.number}
       </strong>
       <span>{track.title}</span>
       <span className={`duration ${!track.length ? "default-duration" : ""}`}>
```
kellnerd, line 86: The recording length fallback is still missing in the display component.

MonkeyDo: Do you have a release at hand where a track doesn't have a length but its associated recording does?

kellnerd: [This cassette release](https://musicbrainz.org/release/f44999ba-2b55-44f2-b45e-6d956c4116c0) does not have track lengths, for example.

MonkeyDo: Thanks, that allowed me to find another couple of things to fix, looking good now!
image

anshg1214 (approved):
LGTM

PR #2930, opened 2024-07-05
LB-1600: Require login for /listening-now page

MonkeyDo:
This page and associated POST endpoint (to get data for the front-end) require a logged-in user. Otherwise the POST request returns a redirect response to the login page, and the RouteLoader in the front-end throws an error trying to JSON parse an HTML body.


MonkeyDo:
Big PR, the review is going to take forever...

anshg1214 (changes requested):
Please also add the listening-now route in the react-router config to protected routes so that any link navigation to the `listening-now` route will be handled.

https://github.com/metabrainz/listenbrainz-server/blob/2760e68d16454f3e983c8fe1a425846a50457f8a/frontend/js/src/routes/index.tsx#L88-L97

MonkeyDo:
I hope this is what you had in mind !
Moved the route definition to a separate file and added it to the definition that had protected routes for the settings pages.

anshg1214 (approved):

PR #2931, opened 2024-07-05
LB-1399: Choose specific release in manual MBID mapping

MonkeyDo:
Allows users to select a specific release when they link an unmatched listen, as well as allowing users to correct failures of the automatic MBID mapper which sometimes maps to a recording and a release that don't match (the recording is not on that release).

### DIFF
--- a/admin/timescale/create_indexes.sql
+++ b/admin/timescale/create_indexes.sql
@@ -81,7 +81,7 @@ CREATE UNIQUE INDEX similar_artist_credit_mbids_reverse_uniq_idx ON similarity.a
 
 CREATE INDEX similarity_overhyped_artists_artist_mbid_idx ON similarity.overhyped_artists(artist_mbid) INCLUDE (factor);
 
-CREATE INDEX mbid_manual_mapping_top_idx ON mbid_manual_mapping_top (recording_msid) INCLUDE (recording_mbid);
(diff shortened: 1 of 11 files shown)


qappell:
@MonkeyDo Any reason for why this is still a draft? A ton of my listens (which have release_mbid already) are incorrectly mapped, so being able to remap them manually would be awesome. Currently 18 votes and 14 watchers on https://tickets.metabrainz.org/browse/LB-1399

MonkeyDo:
> @MonkeyDo Any reason for why this is still a draft? A ton of my listens (which have release_mbid already) are incorrectly mapped, so being able to remap them manually would be awesome. Currently 18 votes and 14 watchers on https://tickets.metabrainz.org/browse/LB-1399

Hi @qappell sorry for the delayed response.
In short, we all agree that this feature is missing, and really impairs or capacity to match to the right recording.
Fixing this (and other issues) required fully rewriting our mapper tool, which has been a *huge* undertaking, compounded with some unrelated personal issues.
Our team has made good progress on this, and we are currently testing out our v2 which you can try out for yourself at https://mapper.listenbrainz.org/
Once we are satisfied with this replacement (and it is looking very good so far), I can finish off this PR which only modifies the interface for mapping.

Help with testing the new mapper is very welcome; if you encounter any issues please inform us here: https://community.metabrainz.org/t/mbid-mapper-version-2-0-preview-has-finally-arrived/813908/21

PR #2932, opened 2024-07-06
fix: Encoded Apple musicUserToken getting saved in the database

anshg1214:
This PR fixes three things
1. Decodes the request data before saving it in the database.
2. Sets the `music_user_token` in the global context.
3. Checks if the user has an active subscription while mounting

### DIFF
--- a/frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
@@ -9,10 +9,9 @@ import {
   loadScriptAsync,
 } from "../../utils/utils";
 import { DataSourceProps, DataSourceType } from "./BrainzPlayer";
+import GlobalAppContext from "../../utils/GlobalAppContext";
 
-export type AppleMusicPlayerProps = DataSourceProps & {
-  appleMusicUser?: AppleMusicUser;
-};
+export type AppleMusicPlayerProps = DataSourceProps;
 
 export type AppleMusicPlayerState = {
   currentAppleMusicTrack?: MusicKit.MediaItem;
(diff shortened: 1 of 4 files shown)


amCap1712:
As discussed on IRC, a few more changes are needed to get this to work.

1. When configuring MusicKit, pass `suppressDialog: true` like:
    ```js
    await MusicKit.configure({
      developerToken,
      app: {
        name: "My Cool Web App",
        build: "0.0.1",
      },
      suppressErrorDialog: true,
    });
    ```
2. When setting the music user token on the MusicKitInstance, also set `restrictedEnabled`
    ```js
    const music = MusicKit.getInstance();
    music.musicUserToken = newMut;
    music.restrictedEnabled = false;
    ```
3. Before loading any media item for playback for the first time or as one time setup,
   ```js
   try {
      await music._services.apiManager.store.storekit.me();
   } catch (error) {
      // token is expired, do token cleanup and redo login
   }
   ```
This should get stuff working for now, in future I also want to explore the refresh token stuff but that should be implemented on the backend so let's leave it out of this PR.

MonkeyDo (commented):
Code looks good!
I can't exactly test the whole workflow as I don't have a paying apple music account, but the changes look reasonable.

For future PR, it would be helpful to have a PR description; just reading the code, it's not clear to me how it resolves the issue of  the token not being saved in the DB.
Even looking at the specific commit a29bdb52 I can see you use the global context instead of a  prop passed from the parent. That makes total sense in terms of updating the token for use in other parts of the website, but I don't see where the DB is concerned.
Is it just the small changes in settings.py about decoding the user token?

anshg1214:
@MonkeyDo, I've thoroughly tested the PR. The encoded user token was saved in the database, which was fixed by decoding the token.

MonkeyDo (approved):
Thanks for checking and for updating the PR!
Now I'm caught up :)

PR #2933, opened 2024-07-07
Update playlist_api.py

rain0r:
The docs state the parameter name is `q` while it acutally is `query`.


mayhem (approved):
Thanks for the fix!

PR #2939, opened 2024-07-14
Update Token in context after being refreshed

anshg1214:
Right now, the access token for Spotify and Soundcloud is being refreshed but is not being updated in the context. This PR fixes this issue.

### DIFF
--- a/frontend/js/src/common/brainzplayer/SoundcloudPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/SoundcloudPlayer.tsx
@@ -8,6 +8,7 @@ import {
   getTrackName,
   searchForSoundcloudTrack,
 } from "../../utils/utils";
+import GlobalAppContext from "../../utils/GlobalAppContext";
 
 require("../../../lib/soundcloud-player-api");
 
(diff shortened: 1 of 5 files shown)


frontend/js/src/common/brainzplayer/SoundcloudPlayer.tsx
```diff
@@ -264,6 +268,7 @@ export default class SoundcloudPlayer
     callbackFunction: () => void
   ): Promise<void> => {
     const { refreshSoundcloudToken, onTrackNotFound, handleError } = this.props;
```
MonkeyDo, line 270: If you feel so inclined, since you are accessing the global context here you could also use the APIService directly, and get rid of this `refreshSoundcloudToken` prop which is just `APIService.refreshSoundcloudToken`.
And the same goes for the other services.
We previously made it this way with the idea of possibly deploying brainzplayer as a separate JS package, hence no relying on global context.
However it's been a few years and there has been no need for a separate package, so let's make it cleaner...

Anyway, could be in this PR or later if you're busy with other stuff, it's nothing crucial.

frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
```diff
@@ -436,6 +444,9 @@ export default class SpotifyPlayer
           const userToken = await refreshSpotifyToken();
           this.accessToken = userToken;
           this.authenticationRetries = 0;
+          if (spotifyUser) {
+            spotifyUser.access_token = userToken;
+          }
```
MonkeyDo, line 449: Do we need an else condition here, where we populate the `spotifyAuth` object if it doesn't exist, with the new access token ?
I guess we won't have the `permissions` though

MonkeyDo (approved):
Thanks for updating this !

frontend/js/src/common/brainzplayer/SoundcloudPlayer.tsx
anshg1214: Yes, we can get rid of the props, but I've already made these changes in #2827 PR, so it does not make sense to make these changes here.

PR #2940, opened 2024-07-14
Add Music Player UI for Mobile Devices

anshg1214:
This PR adds a Music Player UI for Mobile Devices


MonkeyDo:
Looks awesome already! Much, much better UI on mobile, and I think the desktop version is also going to be jealous :P

Early feedback based on this: 
image

- I added a padding-bottom of the height of the brainzplayer bar (`@brainzplayer-height + @progress-bar-height` variables), otherwise the bottom buttons are all hidden behind it: 
image
- You can't really read the track duration text; you might need a tool to decide which color is most readable, and we have tinycolor2 installed. You can generate  complementary colors for example, and use this utility to automatically choose one with default to black/white if it's not readable: https://github.com/bgrins/TinyColor?tab=readme-ov-file#mostreadable   I was just looking at that for #2937 
- I think the top header might be a bit too tall; it looks quite roomy, and (it's hard to see) the cover art is cropped at the bottom, suggesting there is not enough space available (because of the bottom padding I added)
- Regarding the sizing of everything I would recommend the following:
  - Have a padding all around the player of 1em that will serve to align all the elements inside left and right
  - Remove the padding from `.cover-art-scroll-wrapper` and add the following to make sure the cover art stays at max dimensions: ` flex-shrink: 0; aspect-ratio: 1;`
  - Remove horizontal padding for the other elements
  - Don't constrain the height of the `.info` and `.header` classes. Basically, we set a full size for the cover art and let the rest automatically take up the available space
- I think the `cover-art-wrapper` class doesn't need to be flexbox (or maybe you do for swiping??). 
- The feedback icons stroke looks too thick next to the rest of the UI. `stroke-width: 20px;` (or maybe 1em) (down from 50px) should work: 
image
- Finally, for `.player-buttons` you can remove the padding and the font-size, and set `margin-top: auto; margin-bottom: auto;` to automatically center it vertically in the available space. Each button inside can have the class `.btn-icon` added to have a larger button target for my fat fingers :P

The result is this, with flexbox highlighted for convenience

image

frontend/css/brainzplayer.less
```diff
@@ -412,3 +412,144 @@
...
+  width: 100%;
+  display: flex !important;
+  flex-direction: column;
+  right: 0;
+  bottom: -100%;
+  transition: bottom 1s ease-in-out;
+
+  &.show {
+    bottom: 0;
+  }
```
MonkeyDo, line 430: I saw that the .show class has a `display:... !important` which comes from Bootstrap, forcing you to use `!important` as well here.
But you can use a different class name and avoid that easily:

```suggestion
  display: flex;
  flex-direction: column;
  right: 0;
  bottom: -100%;
  transition: bottom 1s ease-in-out;

  &.open {
    bottom: 0;
  }
```

anshg1214: fixed

frontend/css/brainzplayer.less
```diff
@@ -373,6 +373,12 @@
     margin-top: 0;
   }
 
+  .queue-list > ul, li {
```
MonkeyDo, line 376: `li` here targets all list item elements. Is that what you were aiming for, or are you looking for `.queue-list > ul, .queue-list > li`?

anshg1214: Good catch! Yes, I meant `.queue-list > ul, .queue-list > li`.

MonkeyDo (commented):
A few more little pieces of feedback as I've been testing on my phone.
The new player UI is looking sick !

MonkeyDo:
Looks like after I finish my queue, the player starts playing from the top despite the loop/repeat option being off

frontend/css/brainzplayer.less
```diff
@@ -412,3 +420,150 @@
...
+
+.music-player {
+  position: fixed;
+  z-index: 108;
+  height: 100vh;
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  right: 0;
+  bottom: -100%;
```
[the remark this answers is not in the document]
anshg1214: Fixed

MonkeyDo:
I will let you resolve the conflict if you don't mind, and I'm ready for finalizing the PR.
Will do extensive testing locally and deploy on test as well to get aerozol's feedback (and maybe wider community after that?)

MonkeyDo:
One thing I forgot is that we now have a volume control button to fit in.
The button bar was getting busy and losing its symmetry, not looking great.
So I reorganized extra buttons as a secondary button bar (moving the like/dislike buttons there), and tweaked everything a bit to fit nicely.
What do you think?
image

And with the volume bar open (slides in from offscreen to the right)
image

MonkeyDo:
Aaaand finally a bit more cleanup/reorganization on the bottom player bar, because I wasn't fully satisfied.
For example, moved the "..." more actions menu button to the top-right of the new player UI.
Before playing anything, clean look, centered:
image

After pressing play:
image

MonkeyDo (approved):
As far as I am concerned this is ready and looking fire! :fire: 

Currently deployed to test.listenbrainz.org, but who knows how long, so there are screenshots at the bottom of this page to look at.

Would love to hear @Aerozol's opinion on all this, but I think we could already open up to beta testing and community feedback.

MonkeyDo:
Mmm, spoke too soon, the progress bar in the desktop view doesn't seem to update, so I need to fix that.

Aerozol:
Looks great!! Love the simplicity! I don't listen to music much from my phone, and only have YouTube/the default connected, so FYI this have not done any "deep" testing :)

- The button to make the player full screen (the disc) is not obvious to me. I think an "up chevron", or a "full screen" icon might be clearer.
- If I full screen the player, I think the LB player bar at the bottom should be hidden. I don't see that it has any buttons that aren't also in the full screen mode. I know we need to show the YouTube logo, we can put it into the corner of the "cover art square" instead.
- Could we make full screen appear if you swipe up from the player bar, and hide it by swiping down again?
- Had the same thought as Rusty Nova - when the full screen player is activated, could we lock the YouTube player to the "cover art square"? Should keep everybody happy.
- (maybe not for another ticket but while I'm here...) Could we swipe things out of the queue, instead of having to go through the menu each time

These notes are visual but I think would make a big difference:
- It would be cool to add a slight gradient to the bottom, from our secondary huesound colour? I think it would make everything really pop - and avoid having the occasional gross colour hanging out on its own.
- @MonkeyDo, if you had fun with the blob-ules on the home page, and they are not CPU-heavy, we could even have some of them subtly blobbling about in the bg? Happy to do a mockup if you want to see what I mean :) (I was thinking about integrating them into entity pages as well)

MonkeyDo:
Thank you @RustyNova016 and @Aerozol for the good feedback!
I am going to try to address as many of these as I can, but for some I fear we would need a complete rewrite of the player, which would be for another day.
I will make tickets for the remaining improvements that I can't fix in this PR.

MonkeyDo:
Feedback I could not address directly:
- When changing the music, the player doesn't change information all at once
  - I don't think there's much we can do about this I'm afraid
- When the title scrolls, it jitters
  - I haven't seen that happen, but nonetheless I tweaked the animation timings, hopefully that solves the issue you were seeing
- when changing music, the scrollbar disappear, making the css jitter
  - I *think* this issue is resolved—the progress bar no longer disappears between songs—but maybe I misidentified it
- Managed to make the title disappear. No idea how :/
  - No idea either :D One way or another, I assume it is not the result of changes in this PR
- having the original title of the track would be great
  - Not sure about this one. I do understand the reasoning, but I think in the overwhelming majority of cases it will just duplicate info on screen, while we are trying to keep everything as tidy and uncluttered as possible. However,showing the play queue
- swipe up from the player bar to show fullscreen, and hide it by swiping down again
  - This is definitely doable, but I think ansh had some plans regarding swipe/touch actions, so I want to wait until we have a better overall plan for what happens for which actions, and how we will implement them.
  - In the meantime, I also made the cover art thumbnail in the bottom player bar expand the full player, just like the expand icon.
- lock the YouTube player to the "cover art square"
  - It's the obvious way to handle this for the youtube player, showing the video instead of the thumbnail. Unfortunately, the video iframe is not in the same component as the mobile UI, so this would require a pretty consequential rewrite
- "from our secondary huesound colour"
  - Soooo, we are not using that information. Prior to yesterday we were just using the average color of the loaded cover art image. I improved this by using the [Vibrant library](https://vibrant.dev/) which does smarter detection of popping colors in images, and use that to create a gradient. It is much better than the 'average color' approach we use in huesound IMHO
- we could even have some blob-ules subtly blobbling about in the bg
  - This could be pretty cool! However the animation consumes extra resources, and making the screen more busy might not be ideal. I propose to make this experiment in a separate PR for testing.
- bringing some secondary, slightly duller, colours into the player UI
  - as mentioned we don't use huesound. However the vibrant library does detect a couple of "muted" colours for each image, so we could potentially use those. I am currently using for the text and icons the colours selected by the library as most readable (white or black) for each colour of the background gradient. If we take another approach we need to ensure readability. This can be very difficult to make it work for every image IMO
  
  Feedback for other components:
- Could we swipe things out of the queue, instead of having to go through the menu each time
  - I also very much want this, or a little garbage can icon or something I can use as a shortcut

RustyNova016:
If you need me to retest the jitter and progress bar disappearing, just put the build on test server I'll retry.

MonkeyDo:
> If you need me to retest the jitter and progress bar disappearing, just put the build on test server I'll retry.

Yes please! I just redeployed on test with latest changes.

Aerozol:
WHAT THE HECK.

This is rad. And you already glued the youtube player into the center, which was going to be my suggestion! Colours are good, no notes (again, I'm not a massive mobile LB user, so not a detailed look, but the design looks great).

re. blobules, I agree that we should avoid anything that makes the screen busy. I was thinking it could be quite subtle, almost merging with the background. But if it's battery/resource intensive I am happy to do a 180 on that idea.

P.S. Turns out I'm a bastard because I did find some more notes haha:

- The "playing a track" notification that pops up really mucks the visuals, and it's a bit unnecessary when the whole screen is already dedicated to telling you what's playing
- When I open up the queue and scroll down to play something, it's not immediately clear how to hide the queue again. You have to swipe to the top of the queue again to find the "hide" chevron. That chevron should be sticky (It could also be in purple & white and sit above the queue square, rather than inside, but opinions may vary)
- Now that everything is nice and round, could we get a slight round-ness on the album art/youtube square as well?

MonkeyDo:
> WHAT THE HECK.
> 
> This is rad. And you already glued the youtube player into the center, which was going to be my suggestion! Colours are good, no notes (again, I'm not a massive mobile LB user, so not a detailed look, but the design looks great).
> 
> re. blobules, I agree that we should avoid anything that makes the screen busy. I was thinking it could be quite subtle, almost merging with the background. But if it's battery/resource intensive I am happy to do a 180 on that idea.

I think the blobules are going to be too much, but I can always try something out at a later date and evaluate resource consumption.

> P.S. Turns out I'm a bastard because I did find some more notes haha:

<3

> * The "playing a track" notification that pops up really mucks the visuals, and it's a bit unnecessary when the whole screen is already dedicated to telling you what's playing

Point taken, but I don't think I can address it in this PR.

> * When I open up the queue and scroll down to play something, it's not immediately clear how to hide the queue again. You have to swipe to the top of the queue again to find the "hide" chevron. That chevron should be sticky (It could also be in purple & white and sit above the queue square, rather than inside, but opinions may vary)

OK, I will fix the queue button placement to make it sticky to the queue

> * Now that everything is nice and round, could we get a slight round-ness on the album art/youtube square as well?

Sure can do!

RustyNova016:
Found two new bugs in the current test.listenbrainz version:
- You can open up the player while no song is loaded... Which is a bit useless. 
- If it's intented, I feel like the CAA icon could be changed to something less jarringly placeholder. Maybe a transparent record icon? Like what blackplayer does
- On firefox, the search bar hide the top infos until you scroll up. Not sure what's the extent of the possibilities here
- At 0:00 a tiny little bit of progress bar is visible. Is this a nipick? Yes. Does it trigger my OCD? Absolutely. 
- There's no youtube video at the center anymore. Is it intented?

RustyNova016:
Forgot the images. My bad

MonkeyDo:
I have not put the youtube video anywhere, I just made sure the mobile player is rendered above it so that the youtube video does not cover the player.
To put the video in the player will be for another day and will require a significant rewrite, so just not possible at the moment.

Will look into the other improvements, thank you !

Aerozol:
> To put the video in the player will be for another day and will require a significant rewrite, so just not possible at the 

Sorry if I'm missing something, but my thinking is leave everything as-is, with the youtube player on top. And "simply" shift the youtube frame to be the same size and position as where the cover art sits, behind it?

MonkeyDo:
> > To put the video in the player will be for another day and will require a significant rewrite, so just not possible at the
> 
> Sorry if I'm missing something, but my thinking is leave everything as-is, with the youtube player on top. And "simply" shift the youtube frame to be the same size and position as where the cover art sits, behind it?

Yes, totally understand, but at the moment I can't do that. The youtube video is rendered by another component entirely and its position is free floating. I can't attach it to the mobile player UI.
That's what I meant by "significant rewrite", it will require re-coding a bunch of components (not sure how yet) to allow that to happen.

PR #2943, opened 2024-07-19
LB-1606, LB-1601: Standalone recordings cannot be manually added as listen

anshg1214:
Ticket: LB-1606


MonkeyDo (approved):
Working nicely, thanks for fixing my mistakes ! :p

Checked both examples from LB-1606 and LB-1601 , both related to the same issue

PR #2944, opened 2024-07-19
feat: Upgrade sentry to use React Router Integration

anshg1214:
This PR Upgrades Sentry to use Browser Tracing compatible with React Router V6.

Docs: [Sentry](https://docs.sentry.io/platforms/javascript/guides/react/features/react-router/)


MonkeyDo (approved):
From what I can tell from the docs this is looking fine. I guess we will see how sentry likes it after deployment !

PR #2958, opened 2024-08-10
fix: YIM Graph Overflow Issue

anshg1214:
This PR fixes the changes done in 3f96befd1afee70c0e9268e322ea62b144e5b58a. 

Issue Right Now:

https://github.com/user-attachments/assets/99b347e5-ec98-413c-8fa7-90637a7fa696

### DIFF
--- a/frontend/css/year-in-music.less
+++ b/frontend/css/year-in-music.less
@@ -569,6 +569,7 @@
       .graph {
         color: var(--selectedColor);
         overflow: scroll;
+        overflow-y: hidden;
         > * {
           min-width: 550px;
         }
--- a/frontend/css/year-in-music.less
+++ b/frontend/css/year-in-music.less
@@ -568,7 +568,7 @@
       overflow: hidden;
       .graph {
         color: var(--selectedColor);
-        overflow: scroll;
+        overflow: visible;
         > * {
           min-width: 550px;
         }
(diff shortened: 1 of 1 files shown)


frontend/css/year-in-music.less
```diff
@@ -568,7 +568,7 @@
       overflow: hidden;
       .graph {
         color: var(--selectedColor);
-        overflow: auto;
+        overflow: scroll;
```
MonkeyDo, line 571: This looks better IMO, and doesn't hide the tooltips:
```suggestion
        overflow: visible;
```

Before:
image


After:
image

MonkeyDo (changes requested):
Thanks for finding the issue with this!
Suggestion to improve the result:

frontend/css/year-in-music.less
anshg1214: If we set overflow to visible, it breaks on smaller screens 

image

MonkeyDo: Maybe`overflow-y: hidden;` then, I think that resolves both issues

anshg1214: Yep! It does the job

frontend/css/year-in-music.less
```diff
@@ -568,7 +568,8 @@
...
```
MonkeyDo, line 571: I think this should allow  scrolling horizontally when needed (mobile sizes), while not showing the scrollbar when unnecessary:
```suggestion
        overflow-x: auto;
```

anshg1214: Done!

PR #2959, opened 2024-08-11
LB-1254: Graph for month view in statistics may cut of days at the end due to varying length of months

anshg1214:
LB-1254

This PR fixes the listening activity for `last_month` which gets cut out because of using length of previous month instead of the maximum of both months.


MonkeyDo (approved):
The logic seems solid, although I wasn't sure exactly how to test this in real-life.

PR #2961, opened 2024-08-12
Minor text updates

Aerozol:
**Import listens page:**
Swapped around the target url's for 'connect services' and 'submitting data'
Moved the Fun fact into a tool tip.

**Delete listens page:**
Separate 'download listens' and 'delete listens' more, by moving relevant text to the corresponding button.

**Delete account page:**
Separate 'download listens/feedback' and 'delete account' more, by moving relevant text to the corresponding button.

### DIFF
--- a/frontend/js/src/settings/delete-listens/DeleteListens.tsx
+++ b/frontend/js/src/settings/delete-listens/DeleteListens.tsx
@@ -81,7 +81,7 @@ export default function DeleteListens() {
         <title>Delete Listens</title>
       </Helmet>
       <h3 className="page-title">Delete listens: {name}</h3>
-            <p>
+      <p>
         <b>Deleted listens are not recoverable.</b> Consider exporting your
         ListenBrainz data before deleting your account.
       </p>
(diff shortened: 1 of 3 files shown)


Aerozol:
@MonkeyDo do you think you could have a look at my weak attempt to move some text into a React tooltip in ImportListens.tsx? I just kind of plonked it in there, and I can't test to see if that will put it inline with the text... I feel like it won't, but who knows. xo

MonkeyDo:
Overall good!
There was one thing missing for the tooltip, a trigger. Basically, you add some special incantation to any element on the page (`data-tip` and `data-for="id-of-tooltip-element"`), and the tooltip will appear when you hover over that element. Otherwise the tooltip is never shown.

That being said, there need to be an element to hover over. Did you mean to have the tooltip show when you hover over the wor "listen" (what I did in the screenshot below, with "listen" in bold) or over the whole sentence, or something else?

image

Aerozol:
Hah, I ripped the code from this page:
image

Guess I missed the piece that actually adds that rollover item! The intent was to get the little (i) icon at the end of the line - but it's pretty cool to have a secret rollover on the word listen... I'm happy with that if you are.

PR #2962, opened 2024-08-13
LB-1585: Add Syndication Feeds to ListenBrainz

ericd23:


ericd23:
Notes on stats art feed: I tested a few readers and found out that they don't render plain `<svg>` very well. But if we wrap the data in `<object>` as described [here](https://listenbrainz.readthedocs.io/en/latest/users/api/art.html), it works great. So instead of generate the SVG in the feed endpoint itself, I just generate an URL poniting to our API.

mayhem:
@MonkeyDo : Can you please have a look at the UI bits of this PR?

mayhem:
@ericd23 : For the URLs that are being used for the feeds, please use the API_URL from config -- on my dev box right now API_URL is set to a localhost address, but the URL goes to listenbrainz.org

mayhem:
Where can I find the UI buttons for the other feeds (I know where the listens button is), but where are the others?

frontend/js/src/components/SyndicationFeedModal.tsx
```diff
@@ -0,0 +1,161 @@
+import NiceModal, { useModal } from "@ebay/nice-modal-react";
+import * as React from "react";
+import { useState } from "react";
+
+export type SyndicationFeedModalProps = {
+  feedTitle: string;
+  options: {
+    label: string;
+    key: string;
+    type: "dropdown" | "number" | string;
```
MonkeyDo, line 10: I learned recently that React has a type for this specific use case:
```suggestion
    type: React.HTMLInputTypeAttribute;
```

In this case, it won't prevent the type to be an arbitrary string with a typo, but at least autocompletion will provide common use-cases.


A quick note with more advanced typescript, there is a way to conditionally define some props based on another prop, for example requiring `min` and `max` only if `type` = `number`
I don't think it is required to do that here but thought that might be of interest to you:
https://github.com/metabrainz/listenbrainz-server/blob/f346f6aa994acfbcc616c233f8a581ef303ae6a8/frontend/js/src/utils/SearchTrackOrMBID.tsx#L22-L40

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -416,6 +420,24 @@ export default function ArtCreator() {
               Check out the #ListenbrainzMonday tag on your social platform of
               choice!
             </p>
+            <FontAwesomeIcon
```
MonkeyDo, line 423: I'm not convinced of the placement here when we have a big toolbar on the page specifically for this sort of thing:
image

I imagine it was done this way to make it the same as some other pages with the sidebar (i.e the fresh releases page)?

frontend/js/src/explore/fresh-releases/components/ReleaseFilters.tsx
```diff
@@ -32,6 +42,7 @@ type ReleaseFiltersProps = {
   setShowFutureReleases: React.Dispatch<React.SetStateAction<boolean>>;
   releaseCardGridRef: React.RefObject<HTMLDivElement>;
   pageType: string;
+  currentUser: ListenBrainzUser;
```
MonkeyDo, line 45: This should be acquired directly using a React Context hook that we have set up for it, instead of being passed by the parent as a prop.

As an example of that:
https://github.com/metabrainz/listenbrainz-server/blob/f346f6aa994acfbcc616c233f8a581ef303ae6a8/frontend/js/src/components/Navbar.tsx#L8

frontend/js/tests/explore/fresh-releases/FreshReleases.test.tsx
```diff
@@ -168,6 +168,7 @@ describe("FreshReleases", () => {
         setShowFutureReleases={setShowFutureReleases}
         releaseCardGridRef={releaseCardGridRef}
         pageType="sitewide"
+        currentUser={user}
```
MonkeyDo, line 171: With the comment above, don't forget to remove the prop here too.

frontend/js/src/explore/fresh-releases/components/ReleaseFilters.tsx
```diff
@@ -221,6 +233,42 @@ export default function ReleaseFilters(props: ReleaseFiltersProps) {
...
+                    defaultIndex: 0, // Default to 3 days
+                  },
+                ],
+                baseUrl: `https://listenbrainz.org/syndication-feed/fresh-releases`,
+              });
+            } else if (pageType === PAGE_TYPE_USER) {
+              NiceModal.show(SyndicationFeedModal, {
+                feedTitle: `User-specific Fresh Releases`,
+                options: [],
+                baseUrl: `https://listenbrainz.org/syndication-feed/user/${currentUser.name}/fresh-releases`,
```
MonkeyDo, line 267: I'm worried here that `currentUser` might be undefined.
In some other cases you used `currentUser?.name` with the question mark making it safe.
This part of the page shouldn't be accessible by a user that is not logged in, but better make sure it is accessed safely.

frontend/css/main.less
```diff
@@ -399,3 +399,8 @@ pre code.hljs {
   padding: 1rem;
   border-radius: 5px;
 }
+
+.feed-button {
+  cursor: pointer;
+  color: #353070;
+}
```
MonkeyDo, line 406: This seems unused, and I think you overwrite the text color on the component using `style`, making it less useful.

I would suggest wrapping the SVG icon inside a parent `div` element with classes `btn btn-icon feed-button`.
The `btn` class handles curser pointer and all that, `btn-icon` makes a button with just the icon as content, and repurposing the `feed-button` class to adjust the padding and size:
```suggestion
.feed-button {
  padding: 0;
  height: auto;
  font-size: 1em;
}
```

MonkeyDo (commented):
UI looks pretty good!
Seems to be aligned with what you discussed with aerozol, although I will leave reviewing the finer details to him as I did not follow all the details.

A few details and comments below, but nothing that will change the UI drastically

mayhem:
I've got your code running on beta.listenbrainz.org right now and the Weekly Jams and Weekly Exploration Feeds have some issues:

1. the feed URL has a trailing ? that causes problems adding the feed to the reader.
2. the feed has one entry, but that entry seems to have no body. see below.

mayhem:
The title of the "Feed: Site-wide Fresh Releases" feed says: "Fresh Releases - ListenBrainz" and the per user one says: "Fresh Releases for rob -- ListenBrainz". Looking at this, it isn't quite clear what the site-wide feed is for -- I think we should say "All Fresh Releases -- ListenBrainz". What do you think?

mayhem:
The "Feed: Stats Art" also has a trailing ? in the URL.

mayhem:
Also unclear is when an entry gets added to the stats art generator -- my feed is empty. (but then again, I have not generated feed art as of late -- should it be automatic?)

ericd23:
Ahhh. I will have a look.

ericd23:
For the art creator, I think it's because this beta API [URL](https://beta.listenbrainz.org/1/art/designer-top-5/ercd/this_month/750) is not open for consumption. For art creator feed to work, I embed such API URL in the feed so that feed readers could display the SVG correctly.

ericd23:
image
image

ericd23:
And yes, the trailing `?` is a problem. I will fix it in the front end.

mayhem:
OK, the trailing ? has been fixed and the recommendations feed is fixed -- great. My stats are feed is still empty -- what actions cause new entries to appear in the feed there?

ericd23:
Here:
https://github.com/metabrainz/listenbrainz-server/pull/2962#issuecomment-2310228700
https://github.com/metabrainz/listenbrainz-server/pull/2962#issuecomment-2310230740

Simply put I embed an API call in the feed. And the base URL for this call is dependent on the env. Since the beta API cannot be used, the feed reader cannot retrieve the SVG. I guess we could temporarily patch the code to use api.listenbrainz.org in the feed content and see if it works ok.

mayhem:
Ah, got it. That makes sense. OK, then I think we're pretty much ready to go merge this, save for one thing: Commenting out the Feed atom feed, since that isn't properly protected right now. Lets hide the feed button in the UI and comment out the API endpoint that corresponds to it. Also, please leave docs as to how to undo the comments once the OAuth stuff is done and rolled out. Thanks!

ericd23:
Great!

ericd23:
I think there's more to be done than just uncommenting the code once the oauth stuff is done. Once we are there, I will add necessary integration. Just let me know when it's time. I will keep an eye on the chat!

listenbrainz/webserver/views/atom.py
```diff
@@ -171,12 +218,14 @@ def get_listens(user_name):
 
         content = render_template(
             "atom/listens.html",
-            user_page_url=_external_url_for("user.index", path="", user_name=user_name),
+            user_page_url=_external_url_for(
+                "user.index", path="", user_name=user_name),
             user_name=user_name,
             recording_mb_page_base_url="https://musicbrainz.org/recording/",
             track_name=track_name,
             recording_mbid=recording_mbid,
```
amCap1712, line 226: Unable to comment a few lines above so doing it here, I think the listens should fallback to using mapped mbid data if the additional_info doesn't have the required fields? But not a blocker for merging, we can do this later as well.

ericd23:
Yes I can work on this next week and I will talk to aerozol.

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
```diff
@@ -5,13 +5,19 @@ import { toast } from "react-toastify";
 import { Helmet } from "react-helmet";
 import { useQuery } from "@tanstack/react-query";
 import { useNavigate } from "react-router-dom";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faRss, faSquareRss } from "@fortawesome/free-solid-svg-icons";
```
MonkeyDo, line 9: faSquareRss seems unused here

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
```diff
@@ -370,6 +376,53 @@ export default function FreshReleases() {
...
+                          defaultIndex: 0,
+                          tooltip:
+                            "Select how many days of past releases to include in the feed, starting from today. Only releases that have already been published will be included.",
+                        },
+                      ],
+                      baseUrl: `${getBaseUrl()}/syndication-feed/fresh-releases`,
+                    });
+                  } else if (pageType === PAGE_TYPE_USER) {
+                    NiceModal.show(SyndicationFeedModal, {
+                      feedTitle: `User-specific Fresh Releases`,
```
MonkeyDo, line 415: Should this be customized?

```
`${currentUser.name}'s Fresh Releases`
```

mayhem:
The following UI improvements are still needed before release: 

https://tickets.metabrainz.org/browse/LB-1678

mayhem (approved):
Lets coordinate with @Aerozol to release this.

MonkeyDo:
Deployed on beta.listenbrainz.org

Dashboard: https://beta.listenbrainz.org/syndication-feed/user/MY_USERNAME/listens?minutes=240
Entity stats: https://beta.listenbrainz.org/syndication-feed/user/MY_USERNAME/stats/top-tracks?range=this_week&count=10
Fresh releases: https://beta.listenbrainz.org/syndication-feed/user/MY_USERNAME/fresh-releases
Stats art: https://beta.listenbrainz.org/syndication-feed/user/MY_USERNAME/stats/art/custom
Recommendation playlists: https://beta.listenbrainz.org/syndication-feed/user/MY_USERNAME/recommendations?recommendation_type=weekly-jams
Playlists: https://beta.listenbrainz.org/syndication-feed/playlist/PLAYLIST_ID

Aerozol:
Looking really nice, thanks so much @MonkeyDo! Those buttons look like they always belonged. I couldn't ask for more.

I am struggling to review the RSS feed updates themselves, because my reader doesn't want to refresh them, but I trust you made the coupla changes you marked as done in the doc.

My only note is that the (?) mouseover in the stats feeds (top artist/album/tracks) references 30 minutes, instead of weeks/month/year etc, which are the available options.

Is there a reason there's no RSS feed/icon in the global feed yet? Is that giving some trouble?

MonkeyDo:
There is no global stats feed, if that's what you meant. Not sure why, or if it was part of the original plan.
In any case it's probably best to work on it separately.

And none for the recent listens page, either, which I'm not exactly certain we can do easily.

MonkeyDo:
I've created the ticket https://tickets.metabrainz.org/browse/LB-1690 to keep track of improvements and other feedback I was given by users

MonkeyDo:
Well done with @ericd23 laying out this cool new feature !

Aerozol:
Well done and thanks all!! <3

PR #2963, opened 2024-08-15
Use multiple threads to insert stats in couchdb

amCap1712:
Use a threadpool to split and insert stats in couchdb. 25% improvement in processing time.

### DIFF
--- a/listenbrainz/spark/spark_dataset.py
+++ b/listenbrainz/spark/spark_dataset.py
@@ -115,6 +115,9 @@ class _StatsDataset(SparkDataset):
 
     def __init__(self, stats_type):
         super().__init__(stats_type)
+        # doing empirical testing for various numbers of workers, no speedup
+        # was observed by raising workers to more than 2 because the bottleneck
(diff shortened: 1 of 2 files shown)


listenbrainz/spark/spark_dataset.py
```diff
@@ -102,6 +111,85 @@ def handle_end(self, message):
 CouchDbDataset = _CouchDbDataset()
 
 
+class _StatsDataset(SparkDataset):
+
+    def __init__(self, stats_type):
+        super().__init__(stats_type)
+        self.workers = 2
```
mayhem, line 118: Should this be a const with comments as to how/when to change this for tuning purposes?

mayhem (approved):

listenbrainz/spark/spark_dataset.py
amCap1712: Sure, I'll add a comment. FWIW, 2 threads is the best at the moment. The bottleneck shifts to spark at that point.

PR #2964, opened 2024-08-19
test: Rewrite disabled Dashboard tests

anshg1214:


frontend/js/tests/user/Dashboard.test.tsx
```diff
@@ -50,773 +48,763 @@ const props: ListensProps = {
...
+      // Wait for data to be successfully loaded
+      const state = queryClient.getQueryState(queryKey);
+      expect(state?.status === "success").toBeTruthy();
+    });
+
+    expect(screen.getAllByTestId("listen")).toHaveLength(26);
+  });
+
+  // eslint-disable-next-line jest/no-disabled-tests
+  xit("fetches the user's listen count", async () => {
```
MonkeyDo, line 142: This test is still x-ed out and skipped. Is this on purpose?
Probably best not to disable eslint in these cases.

EDIT: I un-x-ed the test and it passes, so I do think this in unintentional.

frontend/js/tests/user/Dashboard.test.tsx
```diff
@@ -50,773 +48,763 @@ const props: ListensProps = {
...
+    await userEvent.click(deleteButton);
+
+    expect(listenCards).toHaveLength(25);
+
+    await waitFor(
+      () => {
+        expect(listenCards).toHaveLength(25);
+      },
+      { timeout: 1000 }
+    );
```
MonkeyDo, line 313: The expect condition here is doubled. I assume a single one should do the trick, otherwise it might indicate another issue

MonkeyDo (changes requested):
Looking good!

frontend/js/tests/user/Dashboard.test.tsx
anshg1214: I've had trouble with this test, it passes when I just run a single test, but when I run the whole test suite, it fails. I'm not able to debug this issue

frontend/js/tests/user/Dashboard.test.tsx
anshg1214: Makes sense, I'll fix this one

frontend/js/tests/user/Dashboard.test.tsx
MonkeyDo: Odd. It was passing for me when I ran the Dashboard test suite. I'll check again.

frontend/js/tests/user/Dashboard.test.tsx
```diff
@@ -50,773 +48,755 @@ const props: ListensProps = {
...
+      http.get("/1/stats/user/*/artists", async (path) =>
+        HttpResponse.json({ payload: [] })
+      ),
+    ];
+    server = setupServer(...handlers);
+    server.listen();
+  });
+  afterEach(() => {
+    queryClient.cancelQueries();
+    queryClient.clear();
```
MonkeyDo, line 115: No idea if his is the reason for the weird issue with the test passing on some machines and not others, but `cancelQueries` here returns a Promise and we should probably await it:
```suggestion
  afterEach(async () => {
    await queryClient.cancelQueries();
    queryClient.clear();
```

MonkeyDo (approved):

PR #2966, opened 2024-08-23
LB-1555: Fresh Releases 'For you' toggle redirects to login page, when not logged in

anshg1214:
LB-1555: Fresh Releases 'For you' toggle redirects to login page, when not logged in

### DIFF
--- a/frontend/css/fresh-releases.less
+++ b/frontend/css/fresh-releases.less
@@ -30,11 +30,6 @@
   display: flex;
   justify-content: center;
   align-items: center;
-
-  #user-releases:disabled {
-    color: rgba(141, 141, 141, 0.5);
-    border-color: rgba(141, 141, 141, 0.5);
-  }
 }
 
 .releases-page {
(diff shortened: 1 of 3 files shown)


frontend/css/fresh-releases.less
```diff
@@ -19,13 +19,22 @@
...
+  }
 }
 
 .fresh-releases-row {
   gap: 0.5rem;
   display: flex;
   justify-content: center;
   align-items: center;
+
+  #user-releases:disabled {
```
MonkeyDo, line 34: I think it could be useful in other cases to have a "disabled" style for the pill, buttons.
That would mean moving these css definitions to the pill.less file.

I would also recommend doing like Bootstrap and defining both the disabled attribute and a .disabled css selector, so that someone in the future can use one or the other without requiring specific knowledge.

MonkeyDo (commented):
Totally makes more sense than hiding the pill !

frontend/css/fresh-releases.less
anshg1214: Fixed

PR #2978, opened 2024-09-11
Add backend support for user flair and donations leaderboard

amCap1712:


listenbrainz/config.py.sample
```diff
@@ -22,10 +22,8 @@ SQLALCHEMY_TIMESCALE_URI = "postgresql://listenbrainz_ts:listenbrainz_ts@lb_db/l
 TIMESCALE_ADMIN_URI = "postgresql://postgres:postgres@lb_db/postgres"
 TIMESCALE_ADMIN_LB_URI = "postgresql://postgres:postgres@lb_db/listenbrainz_ts"
 
-# MessyBrainz
-MESSYBRAINZ_SQLALCHEMY_DATABASE_URI = "postgresql://messybrainz:messybrainz@lb_db:5432/messybrainz"
-MESSYBRAINZ_ADMIN_URI = "postgresql://postgres:postgres@lb_db/postgres"
-MESSYBRAINZ_ADMIN_MSB_URI = "postgresql://postgres:postgres@lb_db/messybrainz"
+# MetaBrainz
+SQLALCHEMY_METABRAINZ_URI = ""
```
mayhem, line 26: We should probably add a comment as to why this is where and when someone might need to fill it out.

listenbrainz/db/donation.py
```diff
@@ -0,0 +1,124 @@
...
+
+def init_meb_db_connection(connect_str):
+    """Initializes database connection using the specified Flask app."""
+    global engine
+    while True:
+        try:
+            engine = create_engine(connect_str, poolclass=NullPool)
+            break
+        except psycopg2.OperationalError as e:
+            print("Couldn't establish connection to db: {}".format(str(e)))
```
mayhem, line 20: Should this be a logging call, rather than print?

listenbrainz/db/donation.py
```diff
@@ -0,0 +1,124 @@
...
+            print("Sleeping 2 seconds and trying again...")
+            time.sleep(2)
+
+
+def get_flairs_for_donors(db_conn, donors):
+    """ Get flairs for donors. """
```
mayhem, line 26: What does this function return?

mayhem: Same for other functions

listenbrainz/db/donation.py
```diff
@@ -0,0 +1,124 @@
...
+        OFFSET :offset
+    """
+
+    results = meb_conn.execute(text(query), {"count": count, "offset": offset})
+    donors = results.all()
+
+    return get_flairs_for_donors(db_conn, donors)
+
+
+def is_user_donor(meb_conn, musicbrainz_row_id: int):
```
mayhem, line 119: I think there is a concept that I may not have clearly articulated in the doc -- but the idea that a users donation is time limited. I suggested $5/month, so if a user donates $50, then they get flair for 10 months. I think this function needs to reflect that.

mayhem (changes requested):
some more comments needed and a change to see if a user has recently donated.

listenbrainz/db/donation.py
amCap1712: The logger is likely not initialized at this, because we are still inside create_app. Other db connection init code uses print for the same reason. I think logging might work but there's a detail or two I need to look up to be sure. for another day.

listenbrainz/db/donation.py
amCap1712: Fixed.

listenbrainz/db/donation.py
mayhem: Ah, perhaps just document it then and move on.

listenbrainz/db/donation.py
amCap1712: improved docstrings

listenbrainz/config.py.sample
amCap1712: done

PR #2979, opened 2024-09-11
Add a donation page with support tiers

MonkeyDo:
Following the discussions regarding donations as outlined in  https://docs.google.com/document/d/1Nt8vZa_EZN8qgP0sQmfb5jZyjGepcxe3fbphphA0uF8

Here we have a good starting point for a facescious "support tiers" type of page.
We might want to revisit some of the text, particualrly if we find better "perks" than the ones listed

image

Also adds a link to the page in the side navbar. It is already in the footer as well be we might want a banner of some sort there instead.

Please be aware, currently the buttons to donate redirect to the metabrainz website while we are setting up donations on LB.

### DIFF
--- a/frontend/js/src/about/donations/Donate.tsx
+++ b/frontend/js/src/about/donations/Donate.tsx
@@ -38,9 +38,12 @@ export default function Data() {
                 <b>Free for everyone</b>
               </h2>
             </div>
-            <div className="perk">
-              <FontAwesomeIcon icon={faCheck} color={COLOR_LB_GREEN} />{" "}
-              <b>All website features, for free, forever</b>
+            <div>
+              <FontAwesomeIcon icon={faCheck} color={COLOR_LB_GREEN} />
+              <b style={{ marginLeft: "0.5em" }}>
+                All website features, for free, forever
+              </b>
+              <br />
             </div>
           </div>
           <div className="tier card">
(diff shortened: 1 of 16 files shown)


frontend/js/src/about/donations/Donate.tsx
```diff
@@ -0,0 +1,213 @@
...
+            your personal data isn&apos;t the price you pay.
+            <br />
+            We believe everyone should have access to all features —no paywalls,
+            no “Pro++” features.
+            <br />
+            <b>All features are free for everyone.</b>
+            <br />
+            <br />
+            While it takes real-world money to keep us going, all our data is
+            open-srouce and free for everyone.
```
kellnerd, line 198:
```suggestion
            open-source and free for everyone.
```

kellnerd (commented):
I like what I see so far, but I would probably turn the lists of benefits into HTML lists with a custom list indicator so that the checkmarks are not floated by text when the list item wraps over multiple lines.

Aerozol:
Looks great!

I think the text can be cut back, it's very busy for a simple business proposition/joke where every tier is the same. It will make the remaining jokes hit harder and leave more focus on the donate buttons.

e.g. I think the header joke can be smaller, and I would take out the sub-header, which is ground we cover in 'jokes aside'. I would take out all non-bold text from the feature lists, I think people will get it without the explaining.

My 2 cents :)

MonkeyDo:
@Aerozol I updated the page as we discussed during the summit (I hope), and update the screenshot in the description.
Does that work for you? Any feedback, or better wording for any of the "perks"?
I can also deploy the PR to test.LB when you need, in order to test it in-situ. Let me know.

MonkeyDo:
Merging this into the donations branch, meaning we will have more occasions to review the page if need be, along with the rest of the donations-related pages and features

PR #2980, opened 2024-09-11
LB-1608, LB-1614 :Add flash messages to index page

MonkeyDo:
The flash messages that were shown on the page were mistakenly removed during refactoring to a single-page app and we lost critical messages telling users their email is not validated, for example.

See the test error "bob" at the top here

image


anshg1214 (commented):
Thanks for fixing this!
We can probably remove the margin below the flash message to make it look better

image

PR #2981, opened 2024-09-16
Add Donation Leaderboard Page

anshg1214:
image

### DIFF
--- a/frontend/js/src/components/Navbar.tsx
+++ b/frontend/js/src/components/Navbar.tsx
@@ -97,6 +97,9 @@ function Navbar() {
           <NavLink to="/explore/" onClick={toggleSidebar}>
             Explore
           </NavLink>
+          <NavLink to="/donors/" onClick={toggleSidebar}>
+            Donors
+          </NavLink>
         </div>
 
         <div className="navbar-bottom">
--- a/frontend/js/src/components/Navbar.tsx
+++ b/frontend/js/src/components/Navbar.tsx
@@ -97,9 +97,6 @@ function Navbar() {
           <NavLink to="/explore/" onClick={toggleSidebar}>
             Explore
           </NavLink>
-          <NavLink to="/donors/" onClick={toggleSidebar}>
-            Donors
-          </NavLink>
         </div>
 
         <div className="navbar-bottom">
@@ -118,6 +115,9 @@ function Navbar() {
           )}
           <NavLink to="/about/" onClick={toggleSidebar}>
             About
+          </NavLink>
+          <NavLink to="/donors/" onClick={toggleSidebar}>
+            Donors
           </NavLink>
           <a
             href="https://community.metabrainz.org/c/listenbrainz"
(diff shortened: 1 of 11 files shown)


frontend/js/src/components/Navbar.tsx
```diff
@@ -97,6 +97,9 @@ function Navbar() {
           <NavLink to="/explore/" onClick={toggleSidebar}>
             Explore
           </NavLink>
+          <NavLink to="/donors/" onClick={toggleSidebar}>
+            Donors
+          </NavLink>
```
MonkeyDo, line 102: I think this would probably be better in the navbar-bottom element:

image

MonkeyDo (commented):
Super clean!

I love how we can fairly easily set up new pages without having to worry about data loading too much, with all the tools you put in place with the SPA refactor!

Would like some eyes form @Aerozol and @mayhem .
I just redeployed this PR to https://test.listenbrainz.org/donors/

I wonder if it would make sense to have a bit of a paragraph explaining that this is a list of MetaBrainz donors, not specifically ListenBrainz donors. Opinions welcome.

MonkeyDo (changes requested):

MonkeyDo (approved):

Aerozol:
Is this still on test? What I can see looks good, but it's just the header : D

image

anshg1214:
@Aerozol Thanks for reporting this issue. This issue has been fixed in #2983 

I've also updated the PR description with a screenshot of how the page looks like.

Aerozol:
**Looks great!** So cool to see top donors!
My feedback:

1. Double checking that this respects the privacy of anonymous donors in all cases?
2. Where does the link to the page to donate go? If this page is going live before the donation page we can link to https://metabrainz.org/donors for now
3. It suppose it technically makes sense to use the 'square' edged tiles/cards (ahh I'm forgetting what we call these in LB) styles, which is what we use in the nearby setting pages. However I think users will be expecting the rounded + shadowed style of the main pages here.
4. Love the listens stats!
5. This might need a (?) rollover (or maybe a doc link, if it gets long) that answers questions a lot of users will have (are multiple donations combined into a single count on this page, what currency is this in, does this count metabrainz donations, how do I associate a previous donation with my current/new LB account etc)
6. This is more of a feature request, but I want to check before making a ticket: I think it would be _awesome_ to show pins here, when any of these users have a current pinned track. I imagine this wouldn't harm the $ donation status. If anyone disagrees please say!

Let me know if there's anything I can help with.
Note that I haven't been able to test the interaction or on mobile, I just looked at the screenshot, so I might have more notes or some of these might be cleared up if I can test.

frontend/js/src/donors/Donors.tsx
```diff
@@ -98,13 +98,22 @@ function Donors() {
                 {donor.musicbrainz_id &&
                   (donor.is_listenbrainz_user ? (
                     <Link
-                      to={`/user/${donor.musicbrainz_id}`}
+                      to={`${
+                        !donor.is_listenbrainz_user && "https://musicbrainz.org"
```
MonkeyDo, line 102: I suspect this condition is never truthy,aas you are already inside a conditional check of  `donor.is_listenbrainz_user ? (...`

PR #2983, opened 2024-09-18
fix: error in donation eligibility calculation

anshg1214:
This PR fixes an issue where 
1. the calculation for determining donation eligibility could result in a division by zero when the payment_date is too recent (i.e., 0 days between now() and the payment_date).
2. the donations are set as anonymous


amCap1712 (approved):
LGTM, Thanks!

PR #2984, opened 2024-09-20
Cache Album Art Queries

anshg1214:
This PR caches all Album Art Queries using React Query.

### DIFF
--- a/frontend/js/src/common/listens/ListenCard.tsx
+++ b/frontend/js/src/common/listens/ListenCard.tsx
@@ -28,9 +28,11 @@ import NiceModal from "@ebay/nice-modal-react";
 import { faPlayCircle } from "@fortawesome/free-regular-svg-icons";
 import { toast } from "react-toastify";
 import { Link } from "react-router-dom";
+import { useQuery } from "@tanstack/react-query";
 import {
   fullLocalizedDateFromTimestampOrISODate,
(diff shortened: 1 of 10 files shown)


frontend/js/src/common/listens/ListenCard.tsx
```diff
@@ -761,5 +742,40 @@ export class ListenCard extends React.Component<
...
+        return undefined;
+      }
+      try {
+        return await getAlbumArtFromListenMetadata(
+          listen,
+          spotifyAuth,
+          APIService
+        );
+      } catch (error) {
+        return undefined;
```
MonkeyDo, line 771: Might be useful to print out the rare errors to the console, in case users are having issues or don't understand why cover art doesn't load.
For example a user previously reported issues that were caused by their browser blocking the spotify API calls.

MonkeyDo (commented):
Getting this error when testing locally:

```
Query data cannot be undefined. Please make sure to return a value other than undefined from your query function. Affected query key: ["album-art","spotify:7vFC68CBMKu99p4tIz6yzx",false]
```

frontend/js/src/common/listens/ListenCard.tsx
anshg1214: Fixed

MonkeyDo (approved):
All working fine ! :+1:

PR #2985, opened 2024-09-23
LB-1626: Sort Release Groups on Artist Page

anshg1214:
This PR fixes LB-1626.

But now since we're displaying more release group types, we need to fix [LB-1536](https://tickets.metabrainz.org/browse/LB-1536) on priority.

### DIFF
--- a/frontend/js/src/artist/ArtistPage.tsx
+++ b/frontend/js/src/artist/ArtistPage.tsx
@@ -39,10 +39,14 @@ import { useBrainzPlayerDispatch } from "../common/brainzplayer/BrainzPlayerCont
 import SimilarArtistComponent from "../explore/music-neighborhood/components/SimilarArtist";
 import CBReviewModal from "../cb-review/CBReviewModal";
 
+interface ReleaseGroupWithSecondaryTypes extends ReleaseGroup {
+  secondary_types: string[];
+}
+
 export type ArtistPageProps = {
   popularRecordings: PopularRecording[];
   artist: MusicBrainzArtist;
-  releaseGroups: ReleaseGroup[];
+  releaseGroups: ReleaseGroupWithSecondaryTypes[];
   similarArtists: {
     artists: SimilarArtist[];
     topReleaseGroupColor: ReleaseColor | undefined;
@@ -51,6 +55,8 @@ export type ArtistPageProps = {
   listeningStats: ListeningStats;
   coverArt?: string;
 };
+
+const COVER_ART_SINGLE_ROW_COUNT = 8;
 
 export default function ArtistPage(): JSX.Element {
   const _ = useLoaderData();
@@ -94,14 +100,11 @@ export default function ArtistPage(): JSX.Element {
     ["asc", "desc", "desc", "asc", "asc"]
   );
 
-  const [rgsByThisArtist, alsoAppearsOn] = partition(
+  const [rgWithoutCompilations, rgCompilations] = partition(
     releaseGroupsSorted,
-    (rg) => rg.artists[0].artist_mbid === artist?.artist_mbid
+    (rg) => rg?.secondary_types?.includes("Compilation")
   );
-
-  const rgGroups = chain(rgsByThisArtist).groupBy("type").value();
-  const otherRgs = rgGroups.Other ?? [];
-  otherRgs.push(...(rgGroups.Broadcast ?? []));
+  const rgGroups = chain(rgWithoutCompilations).groupBy("type").value();
 
   React.useEffect(() => {
     async function fetchReviews() {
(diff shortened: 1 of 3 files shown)


MonkeyDo (commented):
From what I can tell after deployment on test.LB, the sorting isn't working well yet.
I think we have made some assumptions a bit too fast, and maybe we are doing some operations in the wrong order.
I'll take this complex page as an example (lot of different types):
https://test.listenbrainz.org/artist/f58384a4-2ad2-4f24-89c5-c7b74ae1cce7/

If you look at the albums and compare to the MB page, there are only a handful of items there and they are not the albums I would expect to see  (there are some demos, compilations, and about 20 albums missing)
I think the main issue is doing the groupBy operation *after* the sortBy. We should group first and apply the sort to each group.

Thanks to the react query devtools I can get the original server to try some things:

Sort by primary type and by first secondary type:
`_.groupBy(releaseGroups, rg => rg.type + (rg.secondary_types?.[0] ? " + " + rg.secondary_types?.[0] : ""))`
We end up with something close to what MB displays:
image
Applying the sortBy (as you coded it in this PR) for each of these groups gives me the expected result.


If we don't want to show all these different sections, we can probably show primary-type-only section followed by a group composed of that same primary type + all other secondary types.
For example based on the screenshot of groups above:
- Group 1: Album
- Group 2: Album+Live, Album+Compilation,Album+Demo,Album+Remix
- Group 3: Single
- Group 4: Single+Live (Single+Anything would go here too)
- Group 5: EP
- Group 6: EP+Live (EP+Anything would go here too)
- etc.

For this, something along those lines should work:
`_.groupBy(releaseGroups, rg => rg.type + (rg.secondary_types?.length ? "+" : ""))`
image

More likely than not, we probably want to select specific secondary types to show separate, and clump some other ones together.
For example, I think 1. Album, 2. Album+Live, and 3. (Album+Compilation, Album+Demo,Album+Remix) could be a good way to split it, i.e. keep live albums separate but clump the rest together.
Let me know if you want to have a chat about this in more details

anshg1214:
@MonkeyDo I've made the respective changes to show all the primary types and secondary types similar to how MusicBrainz shows

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -84,11 +117,52 @@ export default function ArtistPage(): JSX.Element {
...
-  const [albumsByThisArtist, alsoAppearsOn] = partition(
+  const [sort, setSort] = React.useState<"release_date" | "total_listen_count">(
+    "release_date"
+  );
+
+  const rgGroups = groupBy(
     releaseGroups,
-    (rg) => rg.artists[0].artist_mbid === artist?.artist_mbid
+    (rg) =>
+      rg.type + (rg.secondary_types?.[0] ? ` + ${rg.secondary_types?.[0]}` : "")
```
MonkeyDo, line 127: Annoyingly, primary type can be undefined while secondary type isn't (valid in MB) and results in this:
image

Because there is a secondary type, the bit of code where you check `rgGroups.undefined` on line 143 does not apply to this `undefined + SecondaryType`.

So maybe default to primary type to Other to resolve it?
Not sure the check on line 143 will still be useful after that.
```suggestion
      (rg.type ?? "Other") + (rg.secondary_types?.[0] ? ` + ${rg.secondary_types?.[0]}` : "")
```

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -407,20 +481,21 @@ export default function ArtistPage(): JSX.Element {
...
+            <div className="listen-header">
+              <h3 className="header-with-line">{type}</h3>
+              <SortingButtons sort={sort} setSort={setSort} />
+            </div>
+            <div
+              className={`cover-art-container dragscroll ${
+                rgGroup.length <= COVER_ART_SINGLE_ROW_COUNT ? "single-row" : ""
+              }`}
+            >
+              {rgGroup.map(getReleaseCard)}
```
MonkeyDo, line 495: While you are in here and now that you've added a type for secondary_type on the RGs, I wonder if we should add the secondary types in getReleaseCard?
On the one hand we have the info so we could display it, on the other it seems that it would be redundant now that we have the secondary types in the subheaders.
Your call.

MonkeyDo (commented):
It's looking a hundred times better after the last pass !

I agree with your decision to follow the MB way, as we quite clearly don't knnow enough between the two of us to account for all the weird ways items can be split or combined depending on type, nor have a clear idea of what our users would want to see.
At least with type + secondary type it is clear!


Final couple of changes and I think this is ready.

frontend/js/src/artist/ArtistPage.tsx
anshg1214: I don't think we should display the secondary types on the release cards, as that'll be too much information on the card.

MonkeyDo (approved):
Looking mighty fine to me !

PR #2986, opened 2024-09-25
fix: incorrect documentation for `GET /1/metadata/lookup/` request

cgytrus:
# Problem
The documentation for `GET /1/metadata/lookup/` parameters is wrong, `recording_name` doesn't list a type and is duplicated

# Solution
Likely due to hasty copy-paste, the `:type artist_name` doc was duplicated for all 3 params and `release_name` ended up as `recording_name` instead, this PR simply fixes that


MonkeyDo (approved):
Thank you !

PR #2988, opened 2024-09-27
Add an endpoint to get all donors with flairs

amCap1712:


listenbrainz/webserver/views/donor_api.py
```diff
@@ -38,3 +39,19 @@ def biggest_donors():
 
     donors = get_biggest_donors(meb_conn, db_conn, count, offset)
     return jsonify(donors)
+
+
+@donor_api_bp.route("/all-flairs", methods=["GET", "OPTIONS"])
+@crossdomain
+@ratelimit()
+def all_flairs():
+    """ Get flairs (including expired ones) for all users """
```
mayhem, line 48: I think we need more details in this endpoint documentation. What does this endpoint return? Does the caller need to care about expired flairs?

listenbrainz/webserver/views/user_settings_api.py
```diff
@@ -27,6 +27,8 @@ def update_flair():
     if "flair" not in request.json:
         raise APIBadRequest("Missing flair")
 
+    # todo: add flair validation
```
mayhem, line 30: Does this need to be done now or add a ticket?

mayhem (changes requested):

listenbrainz/webserver/views/user_settings_api.py
amCap1712: this needs to be done (if at all) after the structure of the flairs has been decided.

listenbrainz/webserver/views/donor_api.py
amCap1712: fixed.

listenbrainz/webserver/views/user_settings_api.py
mayhem: OK, sounds like no validation will be needed, so lets remove the comment on move on.

mayhem (approved):

PR #2994, opened 2024-10-09
Add Recent Donors Card on Global Feed Page

anshg1214:
This PR adds Recent Donors Card on Global Feed Page

image

### DIFF
--- a/frontend/js/src/recent/RecentListens.tsx
+++ b/frontend/js/src/recent/RecentListens.tsx
@@ -57,14 +57,14 @@ export default class RecentListens extends React.Component<
                 <small className="text-muted">songs played</small>
               </div>
             </Card>
-            <Card id="listen-count-card">
+            <Card id="listen-count-card" className="card-user-sn">
               <div>
                 {globalUserCount ?? "-"}
                 <br />
                 <small className="text-muted">users</small>
               </div>
             </Card>
-            <Card className="card-user-sn hidden-xs hidden-sm">
+            <Card className="hidden-xs hidden-sm">
               <RecentDonorsCard donors={recentDonors} />
             </Card>
           </div>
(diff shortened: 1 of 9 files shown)


MonkeyDo:
Deployed to test.LB for live testing, working fine!

image

frontend/js/src/recent/RecentListens.tsx
```diff
@@ -55,13 +57,16 @@ export default class RecentListens extends React.Component<
...
             </Card>
-            <Card id="listen-count-card">
+            <Card id="listen-count-card" className="card-user-sn">
               <div>
                 {globalUserCount ?? "-"}
                 <br />
                 <small className="text-muted">users</small>
               </div>
             </Card>
+            <Card className="hidden-xs hidden-sm">
```
MonkeyDo, line 67: Not sure I like that it is hidden on some screen sizes, but I suppose if that's what we do with other cards there, we should be coming up with a solution for the other ones too.
Maybe not necessary for sm sizes?
`<Card className="hidden-xs">`

frontend/js/src/recent/components/RecentDonors.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  if (!donors || donors.length === 0) {
+    return null;
+  }
+
+  return (
+    <>
+      <h3 className="text-center" style={{ marginTop: "10px" }}>
+        Recent Donors
+      </h3>
```
MonkeyDo, line 26: We'll want a link to the donors page. Maybe inside the h3 here, with a `<small>` element?

image

MonkeyDo (commented):
Looking good!

I have a couple of suggestions, but nothing that is a deal breaker.

MonkeyDo (approved):
Thanks for the quick changes!

PR #2995, opened 2024-10-09
Make reusable scroll container with indicator arrows

MonkeyDo:
This PR makes a scroll container we are currently using on the "created for you" page reusable so we don't have to reimplement it elsewhere.
This follows a discussion on the LB channel regarding the artist page and its album grid, which is not very usable on desktop if you don't have a trackpad.
Preparations for https://tickets.metabrainz.org/browse/LB-1652, we can reuse this new component for said album grid allowing for arrows and styled scrollbar underneath.

In the meantime, in this PR it's only moving current functionality around and improving customizability with some extra options (optional scrollbar, optional drag-to-scroll, both active by default, and custom css class names for targeted styling).

Example with arrows and scrollbar:

image

### DIFF
--- a/frontend/js/src/components/HorizontalScrollContainer.tsx
+++ b/frontend/js/src/components/HorizontalScrollContainer.tsx
@@ -10,13 +10,13 @@ import { useDraggable } from "react-use-draggable-scroll-safe";
 type HorizontalScrollContainerProps = {
   showScrollbar?: Boolean;
   enableDragScroll?: Boolean;
-  scrollContainerCssClass?: string;
+  className?: string;
 };
 
 export default function HorizontalScrollContainer({
   showScrollbar = true,
   enableDragScroll = true,
-  scrollContainerCssClass,
+  className,
   children,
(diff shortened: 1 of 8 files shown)


frontend/js/src/components/HorizontalScrollContainer.tsx
```diff
@@ -0,0 +1,116 @@
...
+} from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { throttle } from "lodash";
+import React, { PropsWithChildren } from "react";
+import { useDraggable } from "react-use-draggable-scroll-safe";
+
+type HorizontalScrollContainerProps = {
+  showScrollbar?: Boolean;
+  enableDragScroll?: Boolean;
+  scrollContainerCssClass?: string;
```
anshg1214, line 13: Let's change the name of props to className
```suggestion
  className?: string;
```

anshg1214 (changes requested):

anshg1214 (approved):

PR #2997, opened 2024-10-10
LB-1652: Better horizontal scroll on the artist pages album grids

MonkeyDo:
Following up from #2995 whose main goal was to make our existing scroll container with arrows on the side reusable, in order to be used on the artist pages where we show scrollable album grids.
Users have been reporting this issue that was also bugging me, and as we are currently doing some improvements on the artist page albums view (#2985) it seemed like the appropriate time to look into it again.

In the process, I realized we previously used the dragscroll library, but lost that functionality by mistake during the single-page-app migration (the JS file was loaded in an HTML template and we forgot to move that over). That being said, for drag-to-scroll functionality we now use a React hook instead.

Before:
image

After:
image

________________________________________

I will make a note here of an uncertainty I have regarding the album grid itself:
Currently items are laid out in this column order (because of CSS grid peculiarities):
|        |        |        |     |
|--------|--------|--------|-----|
| item 1 | item 3 | item 5 | ... |
| item 2 | item 4 | item 6 | ... |

I dug into the issue for hours, and I do not think there is a way to have a CSS grid restricted to two rows and generating an arbitrary number of columns, which  we would need for a left-to-right order like this;
|        |        |        |     |
|--------|--------|--------|-----|
| item 1 | item 2 | item 3 | ... |
| item 25 | item 26 | item 27 | ... |

So I guess we are stuck with the former solution, which is not necessarily a big issue considering that leaves the higher-sorted elements visible on the page rather than requiring to scroll to see more top items (example above showing items 1-3 and 25-27).

### DIFF
--- a/frontend/css/entity-pages.less
+++ b/frontend/css/entity-pages.less
@@ -117,19 +117,21 @@
       min-height: 150px;
       min-width: 0;
       .artist-page& {
-        max-height: 520px;
+        max-height: 500px;
         overflow-y: hidden;
         border-bottom-style: inset;
         position: relative;
-        &::after {
-          content: "";
-          background-image: @white-gradient;
-          position: absolute;
-          bottom: 0;
-          height: 8em;
-          width: 100%;
+        &.expanded {
+          max-height: initial;
+          padding-bottom: 4em;
         }
       }
+
+      .read-more {
+        bottom: 0;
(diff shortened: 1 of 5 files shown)


anshg1214 (approved):
LGTM 🚀

PR #2998, opened 2024-10-14
Rethink album grouping on artist page

MonkeyDo:
User rustynova had the excellent suggestion regarding the album sorting, following improvements in #2985 :
> Or maybe also add subtypes lists, but for all the types. I don't see any reason to separate single remixes and album remixes

On a technical level, what this means is that instead of grouping like on the MusicBrainz website each primary type followed by each secondary type, we are grouping *first* by secondary type and then by primary type.
We only use the first secondary type even if there are multiple, for extra simplifying.

This means we end up with secondary type categories like "live", "compilation", "demo" etc. regardless of the primary types they contain (so "album + remix", "single + remix" and "other + remix" all end up in the same "remix" category).

This results in a page with fewer categories, but ones that make sense from a user point of view.

Compare for example The Beatles' page: 
image

This PR:
image

### DIFF
--- a/frontend/js/src/artist/ArtistPage.tsx
+++ b/frontend/js/src/artist/ArtistPage.tsx
@@ -437,15 +437,19 @@ export default function ArtistPage(): JSX.Element {
               />
             );
           })}
-          <div className="read-more">
-            <button
-              type="button"
-              className="btn btn-outline"
-              onClick={() => setExpandPopularTracks((prevValue) => !prevValue)}
-            >
-              See {expandPopularTracks ? "less" : "more"}
-            </button>
-          </div>
+          {popularRecordings && popularRecordings?.length > 4 && (
+            <div className="read-more">
+              <button
+                type="button"
+                className="btn btn-outline"
+                onClick={() =>
+                  setExpandPopularTracks((prevValue) => !prevValue)
+                }
+              >
+                See {expandPopularTracks ? "less" : "more"}
+              </button>
+            </div>
+          )}
         </div>
         <div className="stats">
           <div className="listening-stats card flex-center">
--- a/frontend/js/src/artist/ArtistPage.tsx
+++ b/frontend/js/src/artist/ArtistPage.tsx
@@ -265,6 +265,8 @@ export default function ArtistPage(): JSX.Element {
     );
   };
 
+  const releaseGroupTypesNames = Object.entries(groupedReleaseGroups);
+
   return (
     <div id="entity-page" className="artist-page" role="main">
       <Helmet>
@@ -494,7 +496,7 @@ export default function ArtistPage(): JSX.Element {
           )}
         </div>
         <div className={`discography ${expandDiscography ? "expanded" : ""}`}>
-          {Object.entries(groupedReleaseGroups).map(([type, rgGroup]) => (
+          {releaseGroupTypesNames.map(([type, rgGroup]) => (
             <div className="albums">
               <div className="listen-header">
                 <h3 className="header-with-line">{type}</h3>
@@ -511,15 +513,17 @@ export default function ArtistPage(): JSX.Element {
               </HorizontalScrollContainer>
             </div>
           ))}
-          <div className="read-more mb-10">
-            <button
-              type="button"
-              className="btn btn-outline"
-              onClick={() => setExpandDiscography((prevValue) => !prevValue)}
-            >
-              See {expandDiscography ? "less" : "full discography"}
-            </button>
-          </div>
+          {releaseGroupTypesNames.length >= 2 && (
+            <div className="read-more mb-10">
+              <button
+                type="button"
+                className="btn btn-outline"
+                onClick={() => setExpandDiscography((prevValue) => !prevValue)}
+              >
+                See {expandDiscography ? "less" : "full discography"}
+              </button>
+            </div>
+          )}
         </div>
       </div>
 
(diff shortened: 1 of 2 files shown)


reosarevok:
Is it intentional to disregard every secondary type except one?

MonkeyDo:
> Is it intentional to disregard every secondary type except one? If something is both a compilation and live, shouldn't it appear on both sections if you're splitting like this?

I guess I went for the simplest solution, although I'll concede it's probably not ideal.
Do you think it would be better to duplicate the item and show it in both categories? I thought that would be confusing and went for the simpler method (code-wise) that would also limit the complexity of the page.

It doesn't make the selection of the category wrong per se (i.e. if compilation+live and it ends up in complication only or in live only, the selection will still be "correct" although arguably incomplete), but maybe duplication would be better?
From my understanding you would probably have to be coming from the MB page to realize something is different, but maybe there will be cases that are confusing from the get-go?

I can definitely be possible to duplicate the items in multiple categories, but not with the current lodash utility as far as I know.

reosarevok:
My only worry is whether users will be confused that they're looking for all live releases and miss some because they're also under something else, for example. But admittedly, the default MB presentation is not significantly better for that issue - so it might be fine like this. Something for you and @Aerozol to decide, I was just bringing up the issue to make sure it was a decision and not just a mistake.

MonkeyDo:
> My only worry is whether users will be confused that they're looking for all live releases and miss some because they're also under something else, for example. But admittedly, the default MB presentation is not significantly better for that issue - so it might be fine like this. Something for you and @Aerozol to decide, I was just bringing up the issue to make sure it was a decision and not just a mistake.

It was a decision, but motivated by lazyness rather than a deep calculation, so I think the point stands, thank you :)

We could end up prioritizing certain secondary types over others depending on how relevant we think they are for users.
Example code shared by reo from MB webservice backwards compatibility:
https://github.com/metabrainz/musicbrainz-server/blob/de898cc4e8fede8c459c8eee67e626a72c3a6326/lib/MusicBrainz/Server/WebService/XMLSerializer.pm#L281-L288

Aerozol:
I haven't looked super closely, but I think we can use these early days of LB to try out changes like this without worrying too much - as long as you are okay to stay flexible and change it back if there is some disaster that we haven't foreseen, monkey!

P.S. I couldn't find where the 'Album + Demo' release "Anthology Sessions" goes in the new layout? Either I've missed it or it's gotten itself disappeared.

P.P.S It might also free up that space on the album card where we display stuff like "Album", which is redundant if the sections is already called Album (or maybe the repetition is good?)

I will also paste in what I said to monkey on Matrix, cut for relevance:
_I think that page is getting too long/tall for LB’s target audience tbh. Us data nerds want to see every single, sure, but I think the pleb listener (said with love/lols) wants an overview and then to see what else we offer - reviews, similar artist map, similar artists, etc.
I think it could be a bit weird to mix remix albums and remix singles for some artists, but honestly, anything that shortens the page is a +1 from me!_

reosarevok:
> It might also free up that space on the album card where we display stuff like "Album"

If anything that seems more important now that the Live section displays different things and you need a way to tell if it's Single, EP or Album, doesn't it?


> I think that page is getting too long/tall for LB’s target audience tbh

Any reason not to just show the discography collapsed at first, showing just a couple rows of albums with a "see all" link that shows the whole thing?

Aerozol:
> If anything that seems more important now that the Live section displays different things and you need a way to tell if it's Single, EP or Album, doesn't it?

I wasn't very clear - where it's not useful is just where it just repeats the name of the section (ALBUM: ALBUM, EP: EP, SINGLE: SINGLE). Maybe there is something else we could display there. But it's also confusing to only change it in some sections, I can't say I have fully thought about a solution. It's isn't a blocker anyway.

> Any reason not to just show the discography collapsed at first, showing just a couple rows of albums with a "see all" link that shows the whole thing?

From my point of view - no :)
I think monkey has talked about this option on Matrix before, so it could be on the cards for a future update?

MonkeyDo:
Thanks for the feedback, how about this?
Now added a "Full discography" button to expand a limited initial view. It's up on test.LB if you want to try it yourself:
image

reosarevok:
> Now added a "Full discography" button to expand a limited initial view. It's up on test.LB if you want to try it yourself:

Seems fine to me - dunno if "See full discography" would be better to go with the "See more" above, especially if the opposite is still "See less", but that's nitpicking :)

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -125,12 +125,15 @@ export default function ArtistPage(): JSX.Element {
   const [expandPopularTracks, setExpandPopularTracks] = React.useState<boolean>(
     false
   );
+  const [expandDiscography, setExpandDiscography] = React.useState<boolean>(
```
anshg1214, line 128: I believe we should reset the states when the user redirects from one artist page to another. This issue is also there in the state `expandPopularTracks`.

Right now, when you navigate from one page to other, the expandDiscography state retain their present values.

anshg1214 (changes requested):
This PR looks good!
Along with this PR, can we also hide the popular tracks - hide / show button when the popular tracks are [null](https://test.listenbrainz.org/artist/5be435a8-e698-4a0a-8589-5a9845ac4fc0/) or [less](https://test.listenbrainz.org/artist/a0e4d5cd-019e-4158-b1db-9dfa22832939/) ?

anshg1214 (approved):

PR #2999, opened 2024-10-17
fix: Release Group Artist Credits are duplicated

anshg1214:
This PR fixes the Release Group Artist Credits which are duplicated due to the second join to the artist_credit_name.

### DIFF
--- a/mbid_mapping/mapping/mb_artist_metadata_cache.py
+++ b/mbid_mapping/mapping/mb_artist_metadata_cache.py
@@ -85,7 +85,7 @@ class MusicBrainzArtistMetadataCache(MusicBrainzEntityMetadataCache):
                     "name": release_group_name,
                     "mbid": release_group_mbid,
                     "artist_credit_name": artist_credit_name,
-                    "artists": release_group_artists
+                    "artists": sorted(release_group_artists, key=lambda x: x.pop("position", float('inf')))
                 }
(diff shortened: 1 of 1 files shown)


mbid_mapping/mapping/mb_artist_metadata_cache.py
```diff
@@ -85,7 +85,7 @@ def create_json_data(self, row):
                     "name": release_group_name,
                     "mbid": release_group_mbid,
                     "artist_credit_name": artist_credit_name,
-                    "artists": release_group_artists
+                    "artists": sorted(release_group_artists, key=lambda x: x.pop("position", float('inf')))
```
mayhem, line 88: Why are you using pop() here? Did you want to modify the list or simply dereference it>

mayhem (commented):
Have you run this on wolf?

mbid_mapping/mapping/mb_artist_metadata_cache.py
anshg1214: So, we just need the position data to sort the artists, but we don't actually require it anywhere else. So, I'm popping the data before saving it to the cache.

mayhem: Interesting approach, but if it works...

PR #3000, opened 2024-10-19
LB-1635: Navigation history broken when clicking on a user's stats

anshg1214:
This PR fixes LB-1635.

Currently, stat range validation happens within the component, which causes an unwanted history entry due to redirections. As a result, users are unable to navigate back properly. This PR moves the validation logic to the loader, ensuring smoother navigation by avoiding history interference.

### DIFF
--- a/frontend/js/src/search/Search.tsx
+++ b/frontend/js/src/search/Search.tsx
@@ -25,7 +25,10 @@ export default function Search() {
 
   React.useEffect(() => {
     if (invalidSearchTypes(searchType)) {
-      setSearchParams({ search_term: searchTerm, search_type: "artist" });
+      setSearchParams(
+        { search_term: searchTerm, search_type: "artist" },
+        { replace: true }
+      );
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [searchType]);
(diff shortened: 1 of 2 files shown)


MonkeyDo (commented):
That is an interesting side-effect to keep in mind for the future. Do you know if there are any other components where we don validation+redirection that should be fixed as well?

I also wanted to make sure that this can't be solve more simply by using `replace` instead of `redirect` (-> https://reactrouter.com/en/main/fetch/replace)
Did you try that way? IMO the separation of state logic vs. data loading was a bit easier to follow before this PR, so I figured this could be the minimal solution.

MonkeyDo:
P.S: I redeployed the PR to test.LB , since I merged the album grouping PR and going to deploy that to beta.

anshg1214:
> I also wanted to make sure that this can't be solve more simply by using replace instead of redirect (-> https://reactrouter.com/en/main/fetch/replace)
Yes, we can use replace instead of redirect, but again - replace works only in the loader.

BUT, I found a much cleaner and easier way. `setSearchParams` supports replace option for URL updates.

MonkeyDo:
Clean !

MonkeyDo (approved):
Nice find!

PR #3001, opened 2024-10-19
LB-1647: use anchor links in MissingData pagination

MonkeyDo:
On the Missing Data page, the pagination uses anchor links as buttons and does navigation programmatically.
This prevents opening the next page in new tab, for example.

Instead, use the react-router Link component which renders an anchor link.
See on hover that it is indeed a regular anchor link according tho the browser:
image

Also opted for disabling prev/next buttons on first/last page respectively, instead of making them invisible which is not required and changes the UI elements for no good reason.

### DIFF
--- a/frontend/js/src/settings/missing-data/MissingMBData.tsx
+++ b/frontend/js/src/settings/missing-data/MissingMBData.tsx
@@ -3,7 +3,7 @@
 import * as React from "react";
 
 import { faLink, faTrashAlt } from "@fortawesome/free-solid-svg-icons";
-import { useLocation, useSearchParams } from "react-router-dom";
+import { Link, useLocation, useSearchParams } from "react-router-dom";
 import { toast } from "react-toastify";
 import { Helmet } from "react-helmet";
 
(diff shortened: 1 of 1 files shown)


frontend/js/src/settings/missing-data/MissingMBData.tsx
```diff
@@ -172,24 +172,6 @@ export default function MissingMBDataPage() {
     }
   };
 
-  const handleClickPrevious = () => {
-    if (currPage && currPage > 1) {
-      setLoading(true);
-      const updatedPage = currPage - 1;
-      setSearchParams({ page: updatedPage.toString() });
-      afterDisplay();
```
anshg1214, line 180: You can now remove this function `afterDisplay` since it's not being used now.

frontend/js/src/settings/missing-data/MissingMBData.tsx
```diff
@@ -172,24 +172,6 @@ export default function MissingMBDataPage() {
     }
   };
 
-  const handleClickPrevious = () => {
-    if (currPage && currPage > 1) {
-      setLoading(true);
```
anshg1214, line 177: Similarly, the state `loading` and the loader can also be removed now.

anshg1214 (changes requested):

frontend/js/src/settings/missing-data/MissingMBData.tsx
MonkeyDo: Thanks, totally didn't see those :)

PR #3004, opened 2024-10-21
Update AddData.tsx with LB Playlist Tool

Aerozol:
Added the ListenBrainz Playlist Tool to the list.
(Might need a quick test sorry!)


MonkeyDo (approved):
Thanks for adding it to the list :)

Quick formatting, and we're ready to got !

PR #3011, opened 2024-10-26
feat: improve apple music fuzzy matching

anshg1214:


frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -209,38 +207,44 @@ export default class AppleMusicPlayer
...
         candidateMatches,
         {
           key: "attributes.name",
-          limit: 1,
         }
       );
-      if (fruzzyMatches[0]) {
-        await this.playAppleMusicId(fruzzyMatches[0].obj.id);
+
+      const matchWithAlbum = fuzzyMatches.find((match) => {
```
MonkeyDo, line 232: This should probably only be run if releaseName is truthy (I believe it is optional in the listen metadata so could potentially be undefined/null).

I'm also reading the documentation for fuzzysort, and I wonder if instead of a secondary pass, this couldn't all be achieved in a single pass (when calling fuzzysort.go above )by:
- using both keys `attributes.name` and `attribute.albumName`
- passing a custom `scoreFn` that boosts results that match the albumName

(See 'advanced usage' section https://github.com/farzher/fuzzysort?tab=readme-ov-file#advanced-usage )

MonkeyDo (commented):
I haven't done a lot of thorough testing of Apple Music overall, but this seems like an improvement.
However in some cases it appears to be too restrictive and ignoring decent matches.

I went back to [this forum comment](https://community.metabrainz.org/t/music-playback-with-apple-music/692332/5) which has a good example.
The new system does not find an acceptable result to play this track (https://listenbrainz.org/player/?recording_mbids=26a654d4-c528-4d39-90a2-28958d30c828) whereas the existing system does.

I believe this is because by trying to match the album name we might ignore a track with the right name and artist but from a different album (in this case, expecting "Sobredosis de amor y salsa" and ignoring a match from "Greatest Hits" album).

I think the mechanism should *prefer* the correct album but still play a decent match (right title+artist) in case we don't get the perfect match with matching album name.

MonkeyDo:
I'm also seeing that the newest version 3.1.0 automatically handles accents and such (we are currently using 3.0.1) so we might be able to drop the use of deburr
https://github.com/farzher/fuzzysort?tab=readme-ov-file#v310

MonkeyDo:
I thing is clear to me, the regexp hack we have to match a string will not work in a lot of cases (like the different apostrophe character I mentioned).
What are our other options? Some library?

PR #3013, opened 2024-10-28
Update AddData.tsx with ListenBrainz macOS Scrobbler for Music.app

Aerozol:
Updated AddData.tsx with ListenBrainz macOS Scrobbler for Music.app


MonkeyDo (approved):
One more :rocket:

PR #3014, opened 2024-10-29
BrainzPlayer album-aware music search

anshg1214:


frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -945,6 +951,75 @@ export default function BrainzPlayer() {
...
+      currentPlayingListenIndex + 1
+    );
+
+    const newQueueMatchedTracks: Record<string, string> = {};
+    const newAmbientQueueMatchedTracks: Record<string, string> = {};
+
+    // Filter the tracks in the album that are not yet matched
+    const queueTracksInAlbum = currentQueueAfterCurrent.filter((track) => {
+      return (
+        getReleaseMBID(track) === releaseMBID &&
```
anshg1214, line 977: @MonkeyDo Should we exclude tracks without a releaseMBID? In some cases, a listen might lack this mapping data but still have the same trackName we’re searching for.

frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -173,10 +184,15 @@ export default class AppleMusicPlayer
...
     }
     try {
-      await this.appleMusicPlayer.setQueue({
+      const queueData = await this.appleMusicPlayer.setQueue({
         song: appleMusicId,
         startPlaying: true,
       });
+      const albumId =
+        // eslint-disable-next-line no-underscore-dangle
+        queueData._queueItems[0].item.relationships.albums.data[0].id;
```
MonkeyDo, line 193: That's looks a bit unsafe. If any element of this long chain doesn't exist or is null/undefined, it will throw an error.
is this the failure condition we want in this case?

Also not sure about accessing _underscore shadow variables, are you missing a getter or something?

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -945,6 +951,75 @@ export default function BrainzPlayer() {
...
+      trackName: string;
+      uri: string;
+    }[]
+  ): void => {
+    const currentQueue = queueRef.current;
+    const currentAmbientQueue = ambientQueueRef.current;
+
+    // Filter the currentQueue after the current playing track
+    const currentPlayingListenIndex = currentListenIndexRef.current;
+    const currentQueueAfterCurrent = currentQueue.slice(
```
MonkeyDo, line 967: I reckon this needs a more descriptive name.
Maybe something like `upcomingQueue` or `upcomingQueueItems`?

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
MonkeyDo: I'm reading all this for the first time, so i hope I won't say anything too stupid XD

I don't think we should filter out non-mapped tracks, but I also don't understand why you want to do that in the first place.
Considering the mapping is a best-effort sort of feature, assume it won't always be present.

You can additionally use the `releaseName` and do some fuzzy matching on that if you are trying to group all tracks from the same album.

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -945,6 +951,75 @@ export default function BrainzPlayer() {
...
+        newAmbientQueueMatchedTracks[track.id] = matches[0].item.uri;
+      }
+    });
+
+    // Update the matched tracks
+    dispatch({
+      type: "UPDATE_MATCHED_TRACKS",
+      data: {
+        dataSource,
+        queueMatchedTracks: newQueueMatchedTracks,
```
MonkeyDo, line 1017: I wonder if it would be easier, instead of storing this info in the listens in the queue, to have a separate object belonging to the BP component, which can be used to retrieve a URI (for a specific music source) using an MBID or name string.
Something like:
```
albumMappings: {
  [releaseMBID || releaseName]: {
    appleMusic: appleMusicAlbumURI,
    spotify: spotifyAlbumURI,
    etc.
  }
}
```
That way, for example if you have the same album spread over two pages of listens, when you go to page 2 you already have the album info you need without having to do the whole process again.
Also means not updating the whole queue, which might re-render the queue UI needlessly.

On the music components side, basically in playListen you would need to call a BP method to check if a value exists in the albumMappings object for this release (by MBID/releasName), and use that if there is a hit; otherwise do the search as usual.

What do you think?

anshg1214: There is a problem in this implementation. We'll have a album mapping in this case, which might not be relevant. In order to use a album mapping, we'll have to make an API call to the data source provider every time in order to use this information. 

Right now, what I'm doing is that fetching all the tracks of the current album, and then using the track <> URI mapping to update the upcoming tracks in the queue.

So, if we want to keep this information separate from queueItem, we can save all this information like

```
albumMapping: {
  [releaseName]: {
    [trackName]: {
      appleMusic: appleMusicURI,
      spotify: spotifyURI,
    }
  }
}
```

MonkeyDo: Thank you for the explanation, I misunderstood the code at first.

Do you think my suggestion makes sense,as opposed to modifying the queue items?

anshg1214: Yes, this makes sense, as we don't have to re=render the listen cards again.

PR #3016, opened 2024-10-29
feat: Add Donation Flair Settings Page

anshg1214:


frontend/js/src/settings/layout.tsx
```diff
@@ -25,6 +25,7 @@ const sections: Section[] = [
   {
     title: "Account",
     links: [
+      { to: "./", label: "User Settings" },
```
MonkeyDo, line 28:
```suggestion
      { to: "./", label: "User settings" },
```

frontend/js/src/settings/Settings.tsx
```diff
@@ -6,10 +6,23 @@ import { Helmet } from "react-helmet";
...
+];
+
 export default function Settings() {
-  const { currentUser } = React.useContext(GlobalAppContext);
+  const globalContext = React.useContext(GlobalAppContext);
+  const { currentUser, APIService, flair: currentFlair } = globalContext;
+
   const { auth_token: authToken, name } = currentUser;
 
+  const [selectedFlairs, setSelectedFlairs] = React.useState<FlairPreferences>(
```
MonkeyDo, line 23:
```suggestion
  const [selectedFlair, setSelectedFlair] = React.useState<FlairPreferences>(
```

frontend/js/src/settings/Settings.tsx
```diff
@@ -50,13 +63,79 @@ export default function Settings() {
...
+  const submitFlairPreferences = async () => {
+    if (!currentUser?.auth_token) {
+      toast.error("You must be logged in to update your preferences");
+      return;
+    }
+    const response = await APIService.submitFlairPreferences(
+      currentUser?.auth_token,
+      selectedFlairs
+    );
+    toast.success("Flair preferences updated successfully");
```
MonkeyDo, line 75: Needs a try-catch around the API call, and handling failures as well with a toast message.

frontend/js/src/settings/Settings.tsx
```diff
@@ -50,13 +63,79 @@ export default function Settings() {
...
+            />{" "}
+            Edit Profile on MusicBrainz
+          </a>
+        </div>
+
+        <div className="mb-15 donation-flairs-settings">
+          <div className="form-group">
+            <h3>Flair Settings</h3>
+            <p>
+              Choose which flairs do you want to use for donation in
```
MonkeyDo, line 112: Maybe something like this ?
```suggestion
              Choose which flair you want your username to show to let other users know you donated.
```

frontend/js/src/utils/types.d.ts
```diff
@@ -657,6 +657,13 @@ declare type BrainzPlayerSettings = {
   >;
 };
 
+declare type FlairPreferences =
```
MonkeyDo, line 660: I think this might be better handled using an enum like 
```
export enum UserFlair {
  Default = "default",
  Air = "air",
  ...
}
```

Not sure we want to export an enum from this file though, maybe needs to move to....Settings.tsx I guess?

MonkeyDo (commented):
Thanks for setting that up!
Some preliminary comments

frontend/js/src/utils/types.d.ts
anshg1214: I don't think we should export the enum from setings.tsx....

This is just a placeholder for now. Since we've not touched upon the implementation of user flairs for now everywhere, we might need to save additional information.

MonkeyDo:
I guess this is about ready to merge, what do you think?

I've started working on a PR to add some CSS flairs, as you work on the username component (I checked, and it will indeed be required for some of the animations). Then we can tie it all together!

anshg1214:
Yes, this PR is ready for merge. I'll open a followup PR for adding a username component

MonkeyDo (approved):
LGTM

PR #3018, opened 2024-10-30
Remove deprecated manual LastFM importer

MonkeyDo:
After #2991 , Last.FM imports are automatically managed by the back-end, making this front-end manual import a duplicate we want to get rid of. Currently it is causing confusion for users to have two separate import mechanisms.

We keep the front-end components as we still use them for LibreFM imports, and rename the components accordingly, making them work by default and primarily for LibreFM.

This PR also moves the functionality to import loved tracks from LastFM into the connect services page as it would otherwise be missing:
image

Once we have converted LibreFM import as well we can get rid of the front-end importer components


Aerozol:
Related comment in Matrix, in case you miss it:
"metadataismydrug:
Is the new importer going to completely replace the old? If so, does the new one support only importing loved tracks? I stopped importing and submit directly, but still import my loved tracks so the times are synced."

MonkeyDo:
> Related comment in Matrix, in case you miss it: "metadataismydrug: Is the new importer going to completely replace the old? If so, does the new one support only importing loved tracks? I stopped importing and submit directly, but still import my loved tracks so the times are synced."

I did miss it!
Yes, the button to import loved tracks (see screenshot in the description) is a manual one-time process (of course you can re-run it with no issues)

Aerozol:
Oh great, thanks money. metadataismydrug and I both misinterpreted this as _all_ the manual lfm functions being removed, I think. Great stuff.

MonkeyDo:
We should be good to go now.
Fixed small typo, LFM username is now saved and shown in the input when you load the page, the buttons are full width and I changed the text of the connect button to "connected" when connected (will update that for the other services in a separate PR).
image

PR #3019, opened 2024-10-31
LB-1664: Only show "full discography" button when required

MonkeyDo:
### Deployed to test.LB for testing: https://test.listenbrainz.org/artist/2fe63f6c-3c54-4151-a6b1-845d030dcb0d/
On the artist page, adding some logic to show the "see full discography" button only if there are more than 4 rows in total across type categories.
This can show up to 4 categories of 1 row, 2 categories of 2 rows, or 1 of 2 rows + 2 of one rows.
After that, like before, crop the container and show the "see full discography" button


anshg1214 (approved):
🚀

PR #3020, opened 2024-10-31
Show user information in last.fm settings page

amCap1712:


MonkeyDo (approved):

MonkeyDo:
Hope you don't mind, I was trying to see if #3018 was missing anything and I figured I would combine both PRs.
To do that I made the change I suggested above, and realized my other suggestion was wrong, or I tested it on an older version, or something.
In any case, it was ready to merge after that.

PR #3022, opened 2024-11-04
Update AddData.tsx with Discord BrainzBot + tidying

Aerozol:
Added the BrainzBot Discord bot. Additionally:

# Problem

A few of the items on this page/list have long descriptions, with extra links to other services, and lists of features.

If we do that for every item on the page it's going to be 4x as long, hard to keep up to date, and harder for users to browse. Similarly, if we link Foobar2000 then we should link every other service mentioned, which will make the page very hard to read.

I suggest that this page is useful as a reference list of tools and services that integrate with LB, not a description/advertisement of all the features they include. It is more practical to maintain feature lists on the tool's page. From a UX perspective, users are more likely to be visiting this page to find what they can do with ListenBrainz + the player/s etc that they are _already_ using. In other words, if they are not using Foobar2000, they should not need to scroll past long descriptions of Foobar2000 plugins. This applies to all players etc on this page.

# Solution

Shortened some feature descriptions.
Removed the links to Foobar2000.

# Action

Will need to have the tidying script run on it, thank you!

### DIFF
--- a/frontend/js/src/about/add-data/AddData.tsx
+++ b/frontend/js/src/about/add-data/AddData.tsx
@@ -404,8 +404,8 @@ export default function AddData() {
             <a href="https://www.foobar2000.org/">Foobar2000</a>
           </em>{" "}
           plugin for syncing local playlists with ListenBrainz (+ Spotify).
-          Tracks playlists changes and resolves tracks with local content
-          and YouTube links.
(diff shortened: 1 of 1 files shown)


Aerozol:
Thank you @MonkeyDo! Changes accepted.

MonkeyDo (approved):

PR #3023, opened 2024-11-04
Update AddData.tsx with TIDAL + neptune client mod + plugin

Aerozol:
Added links to send listens from TIDAL, via the neptune client mod and a ListenBrainz plugin.

# Action

Will need tidying via script, thank you!
Should be no collisions with today's other PR for AddData.tsx, but let me know if there are any issues.

### DIFF
--- a/frontend/js/src/about/add-data/AddData.tsx
+++ b/frontend/js/src/about/add-data/AddData.tsx
@@ -134,6 +134,19 @@ export default function AddData() {
         </li>
         <li>
           <em>
+            <a href="https://github.com/Mastermindzh/tidal-hifi">TIDAL</a>
+          </em>
+          , a cross-platform music player and streaming service:{" "}
+          <a href="https://github.com/uwu/neptune">
+            <code>neptune client mod</code>
+          </a>
+          +{" "}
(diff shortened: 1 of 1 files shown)


frontend/js/src/about/add-data/AddData.tsx
```diff
@@ -132,6 +132,19 @@ export default function AddData() {
           , the web version of Tidal running in electron with Hi-Fi (High & Max)
           support
         </li>
+        <li>
```
MonkeyDo, line 135: Not sure if it is intentional to duplicate the link to TIDAL Hi-Fi (already there at [lines 128-134](https://github.com/metabrainz/listenbrainz-server/blob/ac3a438c01b3de17023eb98b2c3cb5ded6ed4869/frontend/js/src/about/add-data/AddData.tsx#L128-L134)).

I suspect you wanted to replace that item with the new one, in which case the old one still needs to be removed.

Aerozol: I must have been asleep when I was doing these additions... thanks for the catch. The link should not be the same - as far as I can tell TIDAL Hi-Fi and this new link (TIDAL + a custom client) are two different options for users to send listens via TIDAL. I have updated the new link now!

MonkeyDo (approved):
Ace, thank you!

PR #3024, opened 2024-11-05
LB-1667: Artist collages no longer show top releases, instead showing releases by year

anshg1214:


MonkeyDo (approved):
Yeah, I suppose the best we can do in this case is put it back as it was before #2985

PR #3025, opened 2024-11-05
fix: Improve mobile view

anshg1214:
This PR fixes the case where BrainzPlayer hide the login button on homepage on the mobile

### DIFF
--- a/frontend/css/homepage.less
+++ b/frontend/css/homepage.less
@@ -359,7 +359,7 @@
 
   @media (max-width: @screen-sm) {
     .homepage-info {
-      bottom: 2.5em;
+      bottom: 4.5em;
     }
     .homepage-upper {
       .listen-container {
@@ -459,7 +459,7 @@
         top: 0;
       }
       .homepage-info {
-        bottom: 1.5em;
+        bottom: 4.5em;
         color: @dark-grey;
         max-width: 50%;
       }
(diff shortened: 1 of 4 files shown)


MonkeyDo (commented):
I suppose in this case it would make more sense to hide BrainzPlayer on the homepage, no?
There's no music to play and I think it doesn't improve the homepage.

frontend/css/homepage.less
```diff
@@ -459,7 +459,7 @@
         top: 0;
       }
       .homepage-info {
-        bottom: 1.5em;
+        bottom: 4.5em;
```
MonkeyDo, line 462: I think this should probably be reverted, on mobile phone in horizontal mode the bottom text comes butting in with the top text:
image

With the previous `1.5em` value:
image


Only applies to this `max-height` media query, the other one is fine

MonkeyDo (commented):
Sweet!
One last little detail and we're good to merge :rocket:

frontend/css/homepage.less
anshg1214: But, in the cases where we show the BrainzPlayer on the homepage, it fits perfectly
image

MonkeyDo: Ah yes, I forgot while testing that you can be playing music and then go to the landing page.

MonkeyDo (approved):
:rocket:

PR #3026, opened 2024-11-06
LB-1546: Embedable HTML widgets 

MonkeyDo:
This PR creates two new endpoints that return an HTML widget with the aim to embed it on personal websites using an iframe like so:

`<iframe src="https://test.listenbrainz.org/user/mr_monkey/embed/playing-now" frameborder="0" width="100%"></iframe>` and
`<iframe src="https://test.listenbrainz.org/user/mr_monkey/embed/pin" frameborder="0" width="100%"></iframe>`

This is only part 1, implementing playing-now and pinned recording, as a trial run. We still need to refine the design before moving forward with other potential templates (recent listens, top albums/artists/tracks).

image

image

### DIFF
--- a/listenbrainz/webserver/views/user.py
+++ b/listenbrainz/webserver/views/user.py
@@ -6,7 +6,7 @@ from collections import defaultdict
 import listenbrainz.db.user as db_user
 import listenbrainz.db.user_relationship as db_user_relationship
 
-from flask import Blueprint, Response, render_template, request, url_for, jsonify, current_app, stream_with_context
+from flask import Blueprint, render_template, request, url_for, jsonify, current_app
 from flask_login import current_user, login_required
 import psycopg2
 from psycopg2.extras import DictCursor


MonkeyDo:
Paging Dr. @Aerozol for some design feedback, now that all the technical bits are out of the way:

image
 
 
 Would also be interested in better phrasing and a more enticing look, perhaps, for when a user is not listening to anything:
 
image

Aerozol:
Beautiful!

My visual thoughts are:
- Remove the underline from the links
- Bold the song title
- Use #46433a instead of full black

I don't have any other design feedback, this is mainly aligning with the LB panels. If they get really popular maybe some day we (meaning you haha) can add some more variables like `showplaycount="0"` or whatever, but I like these simple guys _a lot!_

Re. a user not playing anything, interesting. It is such a empty canvas, so I'm just going to brain dump:

- It is tempting to add colour or imagery or something fancy, but in my experience plain is preferred for this kind of thing
- I would remove the repeat of the linked username in the text, it feels a bit mechanical, and we already have it on the right.
- I think light grey for this text would be good, since it's unlikely a user wants it to stand out if they're not listening to anything
- It could be fun to randomize the text?
  - Silly things like:
    - mr_monkey has taken the red pill
    - the corporate music industry is holding mr_monkey hostage
    - mr_monkey is currently enjoying the music of life
    - mr_monkey is taking a break
    - mr_monkey is thinking about music
    - ( ˘ ɜ˘) ♬♪♫
    - mr_monkey is not listening to music right now
    - mr_monkey is cool and nice and uses open source
    - etc
  - Or more code intensive things like:
    - mr_monkey has listened to [123] songs
    - the first song mr_monkey submitted to ListenBrianz is [song]
    - mr_monkey loves [random loved track]
    - mr_monkey has listened to [123] songs this week
    - etc
  - Or mix all the above
- Or, pick one of the above as static text
- Or, give users the option to be random or static
- Static or random, could add a variable to let users set their own 'not listening' text

MonkeyDo:
I addressed what I could easily (the text/links CSS), including some modifications for the "no data" cards with muted text, a bit of colour from the logo, and trying to make the two widgets similar:
image
image
The section on the right is a link to the user's profile

I've also added one easy thing to the pin widget: if there is no currently active pin, show the last pinned track as a fallback (but **not** in blue, like we do for the [user taste page](https://listenbrainz.org/my/taste) on the website).

I love the ideas for further improvements, and I think we will inevitably want to add a bunch of options for users to customize the functionality and look. One suggestion from the ticket for example is a light/dark mode option.
I think we should ship this as a  good v1, get some feedback from users and poll for more opinions from those who use it in the wild.

MonkeyDo:
OK, now I'm definitely stopping here, after more small tweaks to make it perfect...

Aerozol:
Looks great!!
Just the big ListenBrainz logo is a big turn-off for me. I know we want to rep LB, but thinking of this as a non-meb fan I would be *much* more likely to use the previous plain text credit version. I think we can be modest and people will still think "ooh that's cool, I want that too"

Aerozol:
Oh sorry @MonkeyDo! My notification email chopped off your message at a point before the examples, and looked like there wasn't more to read.

Thanks for making so many versions. I hope I'm not being too boring when I say that I prefer V1, "No logo"!

PR #3027, opened 2024-11-07
Add flairs CSS animations

MonkeyDo:
Adding a new .less file with the CSS definitions for flair animations.
We'll want to review if we want to keep all of those, and think about others to add, but this should get us started nicely.

Add the `.flair` CSS class along with one of the animation CSS classes (`.shake`,`.flip-3d`, etc.).

Some of these effects require manipulating the text in one of two ways:
1. easy one: adding the `data-text="username"` HTML attribute to the containing element , with the actual username of course
2. spreading each letter in the username to a separate span HTML element. this allows animating each letter with a delay


MonkeyDo:
I'm working on using an enum  to defined the label and value for each flair, and generate a type from the enum to be used where needed.
Please hold.

MonkeyDo:
I added a little preview to the side:
image

That being said, I might try to use react-select to show a select with each option showing the appropriate css style.

MonkeyDo:
It could be interesting, but I'd love some other opinions. Maybe it doesn't add a whole lot compared to the simpler select with the preview on the side.
What do you think?
Maybe @Aerozol would like to chime in?
image

Aerozol:
Tricky, I agree that it possibly doesn't add much. Where it would come in handy is if it is unclear from the drop down which flair is hover-only, allowing the user to muck with mouseover in the preview.

The thing that bugs me more in the screenshot is the long feature explanation :D
The dropdown with the username + flair preview already says it all better than words can. Can we change it to something like:

Flair
Unlocked by donating*. Some flairs are only visible on hover.

*I'm not sure where this component is, if it's in settings and not on the donator page somewhere, then "donating" can/should link to the donate page.

MonkeyDo:
I opted for both the fancy dropdown with previews of the animations, as well as a preview on the side of the dropdown, and changed the text as suggested
image

frontend/js/src/settings/Settings.tsx
```diff
@@ -111,33 +136,55 @@ export default function Settings() {
...
             >
-              Submit
+              <div style={{ flexBasis: "300px" }}>
+                <Select
+                  id="flairs"
+                  name="flairs"
+                  isMulti={false}
+                  value={{
+                    value: selectedFlair,
+                    label: startCase(selectedFlair) as FlairName,
```
anshg1214, line 156: Can we fix this case? I believe changing the className would be the easiest solution
image

frontend/js/src/settings/Settings.tsx
```diff
@@ -3,25 +3,49 @@ import * as React from "react";
...
+        <Username
+          style={{ textAlign: "right" }}
+          username={data.username}
+          selectedFlair={data.value}
+          hideLink
+          elementType="a"
+        />
+      </div>
+    </components.Option>
+  );
```
anshg1214, line 36: We'll probably have to adjust the flex css such that
1. The Underline is limited to the text
2. The Username for the sliced text is not in the middle, instead it should be right alligned.
image

frontend/js/src/settings/Settings.tsx
anshg1214: Also, selecting None makes it blank. Instead It should show the text "None"

image

frontend/js/src/settings/Settings.tsx
```diff
@@ -73,7 +98,7 @@ export default function Settings() {
         currentUser?.auth_token,
         selectedFlair
       );
-      toast.success("Flair preferences updated successfully");
+      toast.success("Flair saved successfully");
```
anshg1214, line 101: Now, on updaing the flair preference, the cache needs to invalidated so that it gets updated everywhere.

```suggestion
      queryClient.invalidateQueries({ queryKey: ["flair"] });
```

frontend/css/flairs.less
```diff
@@ -3,3 +3,380 @@
...
+    }
+  }
+  &.lb-colors-sweep > span {
+    animation-name: lb-colors-sweep;
+  }
+  @keyframes lb-colors-sweep {
+    0%,
+    9%,
+    100% {
+      color: var(--link-color, inherit);
```
anshg1214, line 111: Can we avoid setting the default color in this case to blue?

frontend/css/flairs.less
```diff
@@ -3,3 +3,380 @@
...
+  /*
+    3D text with support for visited and hover colors 
+  */
+  &.extruded {
+    color: @body-bg;
+    text-shadow: 0px 0px var(--link-color), -1px 1px var(--link-color),
+      -2px 2px var(--link-color), -3px 3px var(--link-color),
+      -4px 4px var(--link-color), -5px 5px var(--link-color),
+      -6px 6px var(--link-color);
+    -webkit-text-stroke: 0.5px var(--link-color);
```
anshg1214, line 241: Same here. Can we use the default text color here instead?

anshg1214 (changes requested):
The PR looks great! Just some minor comments, and then we're ready to merge.

One thing, for all the animations, can we avoid using the link-color blue, and use the default text color instead?

frontend/js/src/settings/Settings.tsx
```diff
@@ -111,33 +136,55 @@ export default function Settings() {
         </div>
 
         <div className="mb-15 donation-flairs-settings">
-          <div className="form-group">
+          <form className="form-group" onSubmit={submitFlairPreferences}>
             <h3>Flair Settings</h3>
             <p>
-              Choose which flair you want your username to show to let other
-              users know you donated. ListenBrainz.
+              Unlocked by <Link to="/donate/">donating</Link>. Some flairs are
```
anshg1214, line 142: Can we also add a modal here explaining how do we calculate the time duration for which the flairs remain?

MonkeyDo: Yes, I think this is crucial and currently missing, I had the same thought.

frontend/js/src/settings/Settings.tsx
MonkeyDo: Good catch, these are showing the css class name instead of the label for them!

frontend/js/src/settings/Settings.tsx
MonkeyDo: Here's what I have, would love to hear @Aerozol's feedback as well:
image

anshg1214 (changes requested):
The PR Looks really good now but some flairs are broken now.

The Flair Options Wave, Flip Horizontal, Flip Vertical, Flip 3D, and Tornado are broken, and I cannot see any animation.

It's probably after you tweeked some css as I can see [here](https://github.com/metabrainz/listenbrainz-server/compare/81fe291cc1005db598eb71ac560734a52b7b4873...957db7df9678f30830b9802b786e360ba000cb64#diff-6c1318d73ceea679edb2261a009ab19d55530335ea9ad9e4cc14ec61d631f301L144-L175).

MonkeyDo:
> The PR Looks really good now but some flairs are broken now.
> 
> The Flair Options Wave, Flip Horizontal, Flip Vertical, Flip 3D, and Tornado are broken, and I cannot see any animation.
> 
> It's probably after you tweeked some css as I can see [here](https://github.com/metabrainz/listenbrainz-server/compare/81fe291cc1005db598eb71ac560734a52b7b4873...957db7df9678f30830b9802b786e360ba000cb64#diff-6c1318d73ceea679edb2261a009ab19d55530335ea9ad9e4cc14ec61d631f301L144-L175).

Yes, I realized too, thanks !
Looking into it, but lunch break first

frontend/js/src/settings/Settings.tsx
Aerozol: I thought this was important too, glad to see you're already onto it!

My usual text changes, take or leave as you please:
"Every $5 donated unlocks flair for 1 month. Larger donations will unlock multiple months of flair. Multiple donations will stack."
(I assume this is all correct now, after lucifer changed the donation calculation system?)

Two additional points (sorry!)

1. We will have to mention that the flair isn't forever on the donations page as well - we can also add the same modal/modal text.
2. Since it's getting a bit confusing we should probably show how many months of flair are left/have been built up :S This seems like the right place for it. e.g. "You have x days of flair remaining". (display nothing if they have not donated)

anshg1214 (approved):
🚀

frontend/js/src/settings/Settings.tsx
MonkeyDo: Well I messed up and pushed commits to the feature branch directly instead of opening  PR...

But here it is, using the existing nag-check endpoint to get the remaining days of unlock, or warn the user otherwise:
image

image

Tooltip is the same message as just above:
image

PR #3028, opened 2024-11-10
Add User Flairs and Username Component

anshg1214:
This PR adds a universal Username Component for showing user flairs.

### DIFF
--- a/frontend/js/src/common/Username.tsx
+++ b/frontend/js/src/common/Username.tsx
@@ -2,23 +2,32 @@ import * as React from "react";
 import { Link } from "react-router-dom";
 import useUserFlairs from "../utils/FlairLoader";
 
-function Username({
-  username,
-  elementType,
-  hideFlair = false,
-  showLink = false,
-  ...otherProps
-}: {
+type BaseUsernameProps = {
   username: string;
-  elementType?: keyof JSX.IntrinsicElements;
   hideFlair?: boolean;
-  showLink?: boolean;
   [key: string]: any;
-}) {
-  const Element = elementType;
+};
+
+type WithLinkProps = BaseUsernameProps & {
+  hideLink?: false;
+};
+
+type WithElementProps = BaseUsernameProps & {
+  hideLink: true;
+  elementType: keyof JSX.IntrinsicElements;
+};
+
+function Username(props: WithLinkProps | WithElementProps) {
+  const {
+    username,
+    elementType,
+    hideFlair = false,
+    hideLink = false,
+    ...otherProps
+  } = props;
   const flairType = useUserFlairs(username);
 
-  if (showLink) {
+  if (!hideLink) {
     return (
       <Link
         to={`/user/${username}/`}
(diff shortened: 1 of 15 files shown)


frontend/js/src/common/Username.tsx
anshg1214, line 1: @MonkeyDo Is this component suitable for your flair designs that require JavaScript manipulation, or do you need any customizations?

frontend/js/src/common/Username.tsx
```diff
@@ -0,0 +1,52 @@
...
+  username,
+  elementType,
+  hideFlair = false,
+  showLink = false,
+  ...otherProps
+}: {
+  username: string;
+  elementType?: keyof JSX.IntrinsicElements;
+  hideFlair?: boolean;
+  showLink?: boolean;
```
MonkeyDo, line 15: I think it might make more sense to have the default be a link, and the option to be `hideLink` instead when we really don't want it to be a link (which according to my understanding will be more rare).

In that case, perhaps there is a way to make elementType required, but only when hideLink is true?
We have an example of that in the codebase:

https://github.com/metabrainz/listenbrainz-server/blob/90453e5ef35c5b1c21051d9f33882c8e2a8d5b09/frontend/js/src/explore/art-creator/ArtCreator.tsx#L25C1-L40C1

frontend/js/src/common/Username.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const Element = elementType;
+  const flairType = useUserFlairs(username);
+
+  if (showLink) {
+    return (
+      <Link
+        to={`/user/${username}/`}
+        {...otherProps}
+        className={`${otherProps?.className || ""} ${
+          !hideFlair ? flairType || "" : ""
```
MonkeyDo, line 27: I made it so you always need to apply the class `flair` as well as the actual flair type, that seemed a more safe option to ensure we don't end up with a wacky css animation somewhere not intended...
See #3027 

That being said, "flair" need to be added here along with flairType/

MonkeyDo (commented):
Looking good !

frontend/js/src/common/Username.tsx
anshg1214: Hmm, yes that's a better approach. Fixed

frontend/js/src/common/Username.tsx
MonkeyDo: I propose to focus on merging this PR first, and I will make the modifications needed in #3027 
That way I can actually test with the CSS animation to ensure I'm not missing anything.

MonkeyDo: I made a small modification to the types and to the component to make it easier to adapt once it is merged

anshg1214: Yep, looks good

MonkeyDo (approved):
I took the liberty of pushing a commit with a couple of changes (I had too dig into `React.HTMLAttributes` and `HTMLElementTagNameMap` and try things out before I could comment anything, so I figures I might as well just commit those changes).

PR #3029, opened 2024-11-11
LB-1419: Fix broken partial dates

mayhem:
The MB data cashes have invalid partial dates for release_group data. The problem is that when any part of a || concat expression in PG is NULL, the entire result is NULL. The solution is to move the partial data logic into python code.

### DIFF
--- a/mbid_mapping/mapping/mb_release_group_cache.py
+++ b/mbid_mapping/mapping/mb_release_group_cache.py
@@ -133,9 +133,16 @@ class MusicBrainzReleaseGroupCache(MusicBrainzEntityMetadataCache):
                 tag["genre_mbid"] = genre_mbid
             release_group_tags.append(tag)
 
+
+        date = str(row["year"] or '')
+        if row["month"] is not None:
+            date += "-%02d" % row["month"]
+        if row["day"] is not None:
+            date += "-%02d" % row["day"]
+
         release_group = {
             "name": row["release_group_name"],
-            "date": row["date"],
+            "date": date,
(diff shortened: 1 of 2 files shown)


mayhem:
This PR is also temporarily deployed, anticipating the monthly rebuild of all of the data.

mbid_mapping/mapping/mb_release_group_cache.py
```diff
@@ -133,9 +133,16 @@ def create_json_data(self, row):
                 tag["genre_mbid"] = genre_mbid
             release_group_tags.append(tag)
 
+
+        date = str(row["year"] or '')
+        if row["month"] is not None:
```
amCap1712, line 138: Is this the same format that MB follows for partial dates because consider a date where the month is unknown but the year and day are, it could be formatted as 2024-01 which is easily misunderstood.

amCap1712 (approved):
I am not sure we are relying on the date to be parsable anywhere but in case something breaks we should keep this change in mind. Also, you could use COALESCE in postgres to generate the partial dates but python is fine too.

mbid_mapping/mapping/mb_release_group_cache.py
mayhem: MB dates are not allowed to have a day without a month. YYYY, YYYY-MM, YYYY-MM-DD are the only valid forms.

amCap1712: Ah cool, awesome!

PR #3030, opened 2024-11-11
LB 1455: Fix missing releases in release-group cache

mayhem:
LB-1455 describes how tracklists sometimes take weeks to update -- the cause is that the canonical release data is only updated when the data dumps are generated. Luckily this query doesn't take long, so we can go this a few times a day. With the PR we update the data every 6 hours.


MonkeyDo (approved):
Looks sane to me.
I suppose one question I could ask is if there was an actual reason we weren't doing that before (resources, timing, ???) or if we just forgot to put it in?

mayhem:
Not aware that it needed to happen.

mayhem:
The failing tests do not seem related, merging.

PR #3032, opened 2024-11-13
LB-1660: Sort by listen count on artist pages doesn't properly sort

anshg1214:
This PR fixes LB-1660

### DIFF
--- a/frontend/js/src/artist/ArtistPage.tsx
+++ b/frontend/js/src/artist/ArtistPage.tsx
@@ -71,14 +71,15 @@ function SortingButtons({
   );
 }
 
-interface ReleaseGroupWithSecondaryTypes extends ReleaseGroup {
+interface ReleaseGroupWithSecondaryTypesAndListenCount extends ReleaseGroup {
   secondary_types: string[];
+  total_listen_count: number | null;
 }
 
 export type ArtistPageProps = {
   popularRecordings: PopularRecording[];
   artist: MusicBrainzArtist;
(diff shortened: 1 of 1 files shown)


frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -139,13 +140,17 @@ export default function ArtistPage(): JSX.Element {
...
+    releaseGroupsInput: ReleaseGroupWithSecondaryTypesAndListenCount[]
   ) =>
     orderBy(
       releaseGroupsInput,
       [
-        sort === "release_date" ? (rg) => rg.date || "" : "total_listen_count",
-        sort === "release_date" ? "total_listen_count" : (rg) => rg.date || "",
+        sort === "release_date"
+          ? (rg) => rg.date || ""
+          : (rg) => rg.total_listen_count ?? 0,
```
MonkeyDo, line 150: Yes, the solution here for nullish total_listen_count defaulting to 0 works fine :+1: 
As I wrote in chat, I don't remember why we do these two sorts like this though, and it reads funny:

- if sort order = date, sort by date, otherwise sort by listen count.
- then if sort order = date,sort by listen count, otherwise sort by name.
- then sort by name as a tiebreaker.

I don't understand why we do the second step that way

anshg1214: As we discussed on chat, here sort is a state variable. So we first sort by the selected metric, and if equal then sort by the other metric.

MonkeyDo (approved):

PR #3033, opened 2024-11-13
LB-1677: timescale robustness improvement, might be futile, lol

mayhem:
Last week during a DB outage, our timescale-writer got stuck and didn't come back once the DB came back. @MonkeyDo  and I wondered if this is a problem with the code, where the missing DB connection gets the code stuck somehow.

I audited the code and the recent timescale-writer logs and I didn't find anything that suggests that the code is at fault. This problem seems that its a consul problem, where no DB is defined in the config, and then when it comes back, something gets stuck before the container is really started.

Next time the timescale-writer gets stuck, we should carefully examine the logs to see if we can spot where the problem is. 

This PR changes that catch-all exception to actually exit the container, rather than just restarting the main loop. It is unclear how helpful this will be since, the error message from this catch-all was not found in the logs. But @MonkeyDo and I discussed that we should exit the container when something funky happens, rather than restarting the main loop.

### DIFF
--- a/listenbrainz/timescale_writer/timescale_writer.py
+++ b/listenbrainz/timescale_writer/timescale_writer.py
@@ -195,7 +195,9 @@ class TimescaleWriterSubscriber(ConsumerProducerMixin):
                 break
             except Exception:
                 current_app.logger.error("Error in Timescale Writer:", exc_info=True)
-                time.sleep(3)
+                current_app.logger.error("Sleeping 3 seconds and exiting...", exc_info=True)
+                time.sleep(self.ERROR_RETRY_DELAY)
+                break
 
 
 if __name__ == "__main__":


MonkeyDo (approved):
I hope this does not have any unforeseen adverse effects, but in general I think this is a good way to deal with the issue at hand.
Exiting and letting consul restart the service is what we do for BookBrainz and works absolutely fine.

Flask n00b question: is the addition of that `break` enough to exit completely, including cleanup, closing connections, etc.?

mayhem:
The break exists the main loop, which then exists the program. All resources will be cleaned up on exist, so should be fine.

MonkeyDo:
I want to point out that we also have LB-1676 ("uWSGI address already in use" error + reboots endlessly) for other services (as described in the ticket I saw that error for the `listenbrainz-labs-api-prod` container) which might be a more general problem with consul or flask or brainzutils.

mayhem:
Yeah, I agree. This also appears to be consul/container start problem. I read the code and there isn't anything under our control for how to start a service. Its all uwsgi, so yeah, not an and LB problem, but more an MBH problem.

listenbrainz/timescale_writer/timescale_writer.py
```diff
@@ -195,7 +195,9 @@ def start(self):
                 break
             except Exception:
                 current_app.logger.error("Error in Timescale Writer:", exc_info=True)
-                time.sleep(3)
+                current_app.logger.error("Sleeping 3 seconds and exiting...", exc_info=True)
```
amCap1712, line 198: Instead of breaking, you can remove the loop entirely but either is fine for now.

PR #3034, opened 2024-11-14
Improve Navigation between protected and unprotected routes

anshg1214:
This PR solves two problems:
1. The navigation history breaks when you try to access a protected route while you're logged out.
2. During this navigation, the protected route is rendered once before going to the login page.


MonkeyDo (approved):
I can't figure out how to test this or how to reproduce the issue described in the description, but the code looks fine and after deploying to test.LB for testing everything is workign fine, so :+1:

PR #3035, opened 2024-11-15
LB-1676: Kill lingering uWSGI processes on restart

MonkeyDo:
# Problem

We have had a long-running but rare issue with uWSGI failing to start, complaining that the address is already in use, and restarting in a loop every 10 seconds.

With a lot of help from zas, we investigated the cause and found that if consul template crashes (for an as-yet-unknown reason*), its child processes do not receive a termination signal and keep running in the background, unmanaged.

When consul-template starts uWSGI again, it finds the address in use as the previous process is still running, but it does not know about it.

* we don't know why consul-template crashes, but it has happened when the whole infrastructure loses connectivity (i.e. during Hetzner maintenance).
* This might be a bug in consul-template, or a "normal" mechanism that we are not expecting (such as terminating consul-template after X failed connection retries, but not killing child processes?)

# Solution

Use the config option to store the PID of the uWSGI main process in a file on startup.
Now on the next startup, if said file contains a PID, send a termination signal (SIGTERM) to it before starting a new process.

# Action

We will need to redeploy listenbrainz-labs-api and listenbrainz-api-compact which are the containers affected by this issue (they rely on uwsgi config changes with consul-template, other containers don't)

The second modification moves the command that restarts uWSGI when its config changes to the specific services that rely on those config changes, rather than all services.
There is no need to restart uWSGI when its config has not changed.

### DIFF
--- a/docker/consul-template.conf
+++ b/docker/consul-template.conf
@@ -1,7 +1,6 @@
 template {
     source = "/code/listenbrainz/consul_config.py.ctmpl"
     destination = "/code/listenbrainz/listenbrainz/config.py"
-    command = "sv hup uwsgi"
 }
 template {
     source = "/code/listenbrainz/admin/config.sh.ctmpl"
--- a/docker/services/api_compat/api_compat.service
+++ b/docker/services/api_compat/api_compat.service
@@ -1,4 +1,13 @@
 #!/bin/bash
 
-sleep 1
+set -x
+
+UWSGI_PID_FILE=/tmp/uwsgi-api-compat.pid
+if [ -e "$UWSGI_PID_FILE" ]; then
+  UWSGI_PID=`cat "$UWSGI_PID_FILE"`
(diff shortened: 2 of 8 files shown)


zas (approved):
LGTM

MonkeyDo:
@zas Might want to look at this commit where the `sleep 1` was added. Might have been necessary after all:
https://github.com/metabrainz/listenbrainz-server/commit/b082cb2f33fe80e9ae4bab343dfec4493efea851

zas:
> @zas Might want to look at this commit where the `sleep 1` was added. Might have been necessary after all:
> https://github.com/metabrainz/listenbrainz-server/commit/b082cb2f33fe80e9ae4bab343dfec4493efea851

Ok, then keep it and decrease the sleep above to 4.

zas (approved):

mayhem (approved):
Amazing work you two!!

PR #3036, opened 2024-11-18
Background tasks container stuck in a rollback loop

mayhem:
The background tasks container failed with a catch-all exception about an open transaction. It is very difficult to see exactly where that might haven't come from and how to fix this properly.

However, this problem is easy to mitigate -- when we have a catch-all exception, exit the container and let consul start the container again. We'll need to continue to look out for the root cause of this problem, but this should be a decent mitigation.


mayhem:
Test failures are unrelated. We'll need to look at those soon.

PR #3038, opened 2024-11-19
Check the listen count every 5 minutes

mayhem:
listen counts should be checked more than once every 30 minutes -- with the PR, it drops to every 5 minutes.


MonkeyDo (approved):
:+1:

PR #3039, opened 2024-11-20
LB-1455 again

mayhem:
The previous fix was writing the canonical release data to the lb_conn, not like the mb_conn like the cache functions do. Tested on gaga, should be working fine.

### DIFF
--- a/mbid_mapping/manage.py
+++ b/mbid_mapping/manage.py
@@ -199,7 +199,9 @@ def cron_build_all_mb_caches(ctx):
 @click.pass_context
 def cron_update_all_mb_caches(ctx):
     """ Update all mb entity metadata cache in ListenBrainz. """
-    ctx.invoke(update_canonical_release_data)
+
+    # In this context we want to use mb_conn, not lb_conn, like the functions that follow
+    update_canonical_release_data(False)
     ctx.invoke(update_mb_metadata_cache)
     ctx.invoke(update_mb_artist_metadata_cache)
     ctx.invoke(update_mb_release_group_cache)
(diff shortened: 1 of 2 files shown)


MonkeyDo (commented):
No idea ho all of this works, but if you tested then i suppose it is fine.

Do we need to both remove the default value of use_lb_conn in the update method, and also call it with False in manage.py?
Does the default value change modify any other part of the code?

mayhem:
We want to keep the code as is, since manual invoking might require choosing to write the resultant table to LB or to MB. Only the update-cron entry point will by default write to MB, because that is what prod requires.

Its all a bit messy and needs to be refactored, clearly. But that is for another day.

mbid_mapping/manage.py
```diff
@@ -199,7 +199,9 @@ def cron_build_all_mb_caches(ctx):
 @click.pass_context
 def cron_update_all_mb_caches(ctx):
     """ Update all mb entity metadata cache in ListenBrainz. """
-    ctx.invoke(update_canonical_release_data)
+
+    # In this context we want to use mb_conn, not lb_conn, like the functions that follow
+    update_canonical_release_data(False)
```
MonkeyDo, line 204: Couldn't this be called like `ctx.invoke(update_canonical_release_data, use_lb_conn=False)` instead or does it need to be called separately from this context?

mayhem: Unsure. click is full of poorly documented black magic, that I didn't even bother trying this.

MonkeyDo: I figured that was possibly the case, looking at https://click.palletsprojects.com/en/stable/advanced/#invoking-other-commands

PR #3048, opened 2024-11-25
Add the StatsExplanationsModal to the created for you page

MonkeyDo:
When we don't have this data calculated, we only show users the image below with some text saying their stats haven't been calculated.
Adding a link to show more information to users about how often stats are calculated, otherwise they may be confused.

image


anshg1214 (changes requested):
Placing the image and button side by side doesn’t look appealing. Could you position the button below the image instead?

image

PR #3051, opened 2024-11-28
LB-1640: Use artist MBID in radio prompt

MonkeyDo:
Currently, on artist and album pages, the "artist radio" button and links currently use the artist name instead of the much more precise MBID, resulting in some very bad results in some cases

Fallback to artist name (as before) in case there is no MBID for some unknown -possibly never happening- reason.

Also move the "artist radio" option to the top, and 'nosim' radio (rephrased to "this artist only") below the —arguably more interesting— artist radio.

### DIFF
--- a/frontend/js/src/album/AlbumPage.tsx
+++ b/frontend/js/src/album/AlbumPage.tsx
@@ -260,6 +260,7 @@ export default function AlbumPage(): JSX.Element {
       </div>
     );
   }
+  const artistForRadio = artist.artists?.[0]?.artist_mbid ?? encodeURIComponent(artistName);
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


frontend/js/src/album/AlbumPage.tsx
```diff
@@ -260,6 +260,7 @@ export default function AlbumPage(): JSX.Element {
       </div>
     );
   }
+  const artistForRadio = artist.artists?.[0]?.artist_mbid ?? encodeURIComponent(artistName);
```
anshg1214, line 263: Instead of playing artist radio for just the first artist, we shuold send all artist uuids to the LB Radio to get the best results

frontend/js/src/album/AlbumPage.tsx
```diff
@@ -334,20 +333,16 @@ export default function AlbumPage(): JSX.Element {
...
               </li>
               <li>
                 <Link
-                  to={`/explore/lb-radio/?prompt=artist:(${encodeURIComponent(
-                    artistName
-                  )})&mode=easy`}
+                  to={`/explore/lb-radio/?prompt=artist:(${artistForRadio})::nosim&mode=easy`}
                 >
-                  Similar artists
+                  This artist only
```
anshg1214, line 345: This should be this / these based on the number of artists.

anshg1214 (changes requested):

anshg1214 (approved):
🚀

PR #3052, opened 2024-11-28
Explore page cards reorganization

MonkeyDo:
As mentioned in ticket [LB-1639](https://tickets.metabrainz.org/browse/LB-1639), it would be good to do a little bit of reorganization of the Explore page tiles by separating the YIM cards into their own section.

While I was in there I also took the oportunity to set up a card for YIM 2024 as a placeholder.

Overall result:

image

### DIFF
--- a/frontend/css/explore.less
+++ b/frontend/css/explore.less
@@ -50,14 +50,16 @@
   padding: 0.5em;
 }
 
-.explore-card-text-name > a {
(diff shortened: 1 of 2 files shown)


anshg1214 (approved):

Aerozol:
Looks good, thanks!

PR #3054, opened 2024-12-01
Add back `inserted_at` field to export

Maxr1998:
This was part of the previous exports and should probably be kept. **Untested**, but should hopefully work.

I also noticed that the `recording_msid` is now part of the `track_metadata`, while previously it was part of both the listen object and the `additional_info`.


Maxr1998:
I'd love to, but it appears that the asserts currently compare to the uploaded JSON, which shouldn't contain insertion timestamps. I don't think I understand the test architecture well enough to handle it. Do you maybe have some pointers on what needs to be done?

MonkeyDo:
Sorry, I started fixing merge conflicts and looking at fixing the test case, but I broke something.
I'll have another look at it tomorrow, sorry for messing up your PR ;)

Maxr1998:
No worries. Thanks for looking into it!

mayhem (approved):

MonkeyDo:
Thanks for getting started on this @Maxr1998 !
I scratched my head for a while to get everything working as expected.

Maxr1998:
Thanks for adding the tests and getting this merged! Now I understand why I struggled with it, I didn't realize there was a column with a different name.

MonkeyDo:
> Thanks for adding the tests and getting this merged! Now I understand why I struggled with it, I didn't realize there was a column with a different name.

Neither did I!
One of the big issues I faced was that the relevant error message was output only in the background_task container logs, which do not show when you run test.sh
After that the issue was that changes to that background.py file required taking down the background_tasks container with `./test.sh -d` for cleanup, then running test.sh again, otherwise the changes were not taken into account.

Armed with error logs and up-to-date code it was just a question of figuring out what was missing in the SQL query, in this case the column name was different and inserted_at needed to be added to the GROUP BY clause.

PR #3056, opened 2024-12-02
LB-1693: Prevent overflow on entity pages titles

MonkeyDo:
On album pages particularly, a very long album title or a very long list of artists will cause an egregious overflow.
First time I run into it (reported by a user) so I don't think it's very common, but the fix is very simple, we just prevent overflowing.

I opted to also give a bit more space for artist names by reducing the line height of the top title and font-size of the artist names.

### Before:
image
image

### After:
image
image


frontend/css/entity-pages.less
```diff
@@ -28,7 +28,11 @@
       display: flex;
       flex-direction: column;
       min-width: 200px;
-      max-height: 410px;
+      max-height: 29em;
```
anshg1214, line 31: Handles cases for tablet size screens.
```suggestion
      min-width: 15em;
      max-height: 29em;
```

anshg1214 (changes requested):
Just a small change. Otherwise LGTM 🚀

PR #3046, opened 2024-11-24
Improve Homepage

anshg1214:
This PR is deployed on [https://test.listenbrainz.org/?redirect=false](https://test.listenbrainz.org/?redirect=false)

PR #3057, opened 2024-12-02
Homepage improvements followup

anshg1214:
Following up #3046 with some more feedback from aerozol


frontend/css/homepage.less
```diff
@@ -348,8 +352,8 @@
           position: absolute;
           left: 0;
           bottom: -10px;
-          width: 100%;
-          border-bottom: 1px solid @gray-light;
+          width: 30%;
+          border-bottom: 1px solid @gray-dark;
```
MonkeyDo, line 356: This was correct before, the comment on figma states this sohuld be the same color as the header-with-line used in other pages which uses gray-light: https://github.com/metabrainz/listenbrainz-server/blob/b368548ec7b52583356ddb3b129a3ebe63bd2fb6/frontend/css/entity-pages.less#L209
```suggestion
          border-bottom: 1px solid @gray-light;
```

MonkeyDo (changes requested):
Looking at aerozol's feedback on the other PR, the gap between columns in .homepage-information-content is too small (his point # 4)
I suggest `gap: 6em;` looking at the original design.
I also think the padding for that element should be `padding: 0;` on desktop sizes; not sure the current padding right and left adds much, and since the columns are now 
narrower it evens things out a bit.

I also noticed the .homepage-information is missing a `padding-right: 50px;` (equal to the padding-left) otherwise the content overflows horizontally.

MonkeyDo (approved):

PR #3058, opened 2024-12-02
Update data-update-intervals.rst

Aerozol:
- Added playlist update intervals
- Added note that stats can take longer on the 1st and the 15th

### DIFF
--- a/docs/general/data-update-intervals.rst
+++ b/docs/general/data-update-intervals.rst
@@ -8,10 +8,12 @@ System                                          Update schedule
 =============================================== =========================================
 Receiving listens, updating listen counts		Immediate*
 Deleting listens					            Removed at the top of the next hour (UTC)
-Updating statistics for new listens			    Daily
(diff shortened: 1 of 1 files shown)


Aerozol:
To check: I couldn't get an answer on matrix re. if the daily playlists generate based on the user's timezone, or at a central time.

If this is correct then all is well:
"Daily playlists: Every morning, based on the users timezone setting"

MonkeyDo:
> To check: I couldn't get an answer on matrix re. if the daily playlists generate based on the user's timezone, or at a central time.
> 
> If this is correct then all is well: "Daily playlists: Every morning, based on the users timezone setting"

Sorry, that's my fault I forgot to answer!
I just re-read the code: the daily jams job is run every hour, and takes into account user timezones.

So it should be run daily at midnight in the user's timezone.
@amCap1712 would be able to confirm or if I can't read code for shit :p

amCap1712 (approved):

PR #3062, opened 2024-12-03
LB-1639: Rename "missing data" to "link listens" and add it to explore page

MonkeyDo:
As discussed in the ticket, we want easier access to this page, hence adding it to the explore page.

While discussing it, I reported that a user suggested the name "missing data" is confusing, all the more since we improved the page's functionality to allow linking whole albums.

Renamed the routes and text, but also the component and variables to match.

image

image

I've also added a small explanation tooltip next to the text "automatically linked":
image


Aerozol:
Oh wow, thanks for jumping on this so quick!

I like it - I like that the text just says "Link listens" and then the full page says "Link with MusicBrainz" (keeping it simple until users are on the page with room for a bigger description/a link to MB). It also matches with the link icon we use for unmatched listens.

Seems good to me!

frontend/js/src/settings/link-listens/LinkListens.tsx
```diff
@@ -198,10 +203,26 @@ export default function LinkListensPage() {
         <title>Link with MusicBrainz</title>
       </Helmet>
       <h2 className="page-title">Link with MusicBrainz</h2>
+      <ReactTooltip id="matching-tooltip" multiline>
+        We try our best to match every listen with a MusicBrainz recording in
+        order to enrich the information about the track, album and artist.
```
Aerozol, line 208: Me with the red pen, as always... suggestions:
"We automatically match listens with MusicBrainz recordings when possible, which provides data like track, album, artist, cover art, and more."

frontend/js/src/settings/link-listens/LinkListens.tsx
```diff
@@ -198,10 +203,26 @@ export default function LinkListensPage() {
...
+        <br />
+        Sometimes the auto-matching fails to recognize tracks, but you can
+        manually link them below.
+        <br />
+        Other times, this data does not exist in MusicBrainz and needs to be
+        added.
```
Aerozol, line 214: Suggestions:
"When a track can't be auto-matched you can manually link them here."
"Recordings may not exist in MusicBrainz, and need to be added there first."

frontend/js/src/settings/link-listens/LinkListens.tsx
MonkeyDo: This is specifically why I request a review from you !

anshg1214 (approved):
🚀

PR #3065, opened 2024-12-04
LB-1698: Switching between different stat timeframes can sometimes break the stats displayed

anshg1214:
[LB-1698](https://tickets.metabrainz.org/browse/LB-1698)


MonkeyDo (approved):
Good to go ! :+1:

PR #3066, opened 2024-12-04
LB-1695: A hopeful attempt at avoiding deadlocks when updating similar_user data

mayhem:
Right now rotating the similar user data into place is done in 3 separate transactions, which leaves open a pile of chances for deadlocks to occur. The first thing we should do is carry out these tasks in one transaction and see if the problem persists. If it does, we'll take a second look at it.


mayhem:
Unless I restart the prod web container I can't test this -- any thoughts on how to test this?

amCap1712 (approved):

PR #3069, opened 2024-12-08
Store sitewide stats in a separate db

amCap1712:
Currently sitewide stats are stored alongside user stats in a couchdb database. When a day's user stats have been generated, the old day's database is removed, deleting the older sitewide stats as well. Between then and the time today's sitewide stats are generated they are unavailable to view on the site. To avoid this issue store sitewide stats in a different db.

### DIFF
--- a/listenbrainz/db/couchdb.py
+++ b/listenbrainz/db/couchdb.py
@@ -130,6 +130,20 @@ def fetch_data(prefix: str, user_id: int):
     return None
 
 
+def fetch_exact_data(database: str, document_id: str):
+    """ Retrieve data from couchdb for the exact given database and document id.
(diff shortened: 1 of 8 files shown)


listenbrainz/db/couchdb.py
```diff
@@ -130,6 +130,20 @@ def fetch_data(prefix: str, user_id: int):
     return None
 
 
+def fetch_exact_data(database: str, document_id: str):
```
mayhem, line 133: Good -- last time I looked at the couchdb code it wasn't very friendly for adding new types of data, so this stuff helps a lot.

mayhem (approved):
LGTM!

PR #3070, opened 2024-12-09
LB-1689: Improve last.fm loved tracks import

amCap1712:
The existing importer only tries to import tracks with track_mbids. Extend it to recognise recording_mbids and for tracks without mbids do a messybrainz lookup.

### DIFF
--- a/listenbrainz/domain/lastfm.py
+++ b/listenbrainz/domain/lastfm.py
@@ -107,6 +107,7 @@ def fetch_lfm_feedback(lfm_user: str):
 
 
 def bulk_get_msids(connection, items):
+    """ Fetch msids for all the specified items (recording, artist_credit) in batches. """
     query = """
         SELECT DISTINCT ON (key)
                lower(s.recording)  || '-' || lower(s.artist_credit) AS key
--- a/listenbrainz/domain/lastfm.py
+++ b/listenbrainz/domain/lastfm.py
@@ -1,8 +1,10 @@
+import logging
 import uuid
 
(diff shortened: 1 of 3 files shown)


listenbrainz/domain/lastfm.py
```diff
@@ -88,27 +89,43 @@ def fetch_lfm_feedback(lfm_user: str):
...
+            except (ValueError, TypeError):
+                item["mbid"] = None
 
-    return track_mbids, track_data, counts
+            items.append(item)
 
+    return items, total_count
+
+
+def bulk_get_msids(connection, items):
```
mayhem, line 109: docstring

listenbrainz/domain/musicbrainz.py
```diff
@@ -41,3 +45,46 @@ def get_user_info(self, token: str):
         )
         response.raise_for_status()
         return response.json()
+
+    def _submit_review_to_CB(self, token: str, review: CBReviewMetadata):
```
mayhem, line 49: This seem unrelated...

mayhem (approved):
Minor cleanup, after that good to go

listenbrainz/domain/lastfm.py
amCap1712: fixed.

listenbrainz/domain/musicbrainz.py
amCap1712: my bad. removed.

PR #3075, opened 2024-12-11
LB-1707: Add sort name and disambiguation in artist search pages

MonkeyDo:
Sort name and disambiguation are easily accessible in the MB API search results so no reason not to show them on the page.
Showing disambiguation after the artist name, and adding the sort name  + disambiguation on hover over the name.

On artist pages and listen cards it's going to be more complicated as we don't necessarily have access to that metadata in the same way.


anshg1214 (approved):
🚀

PR #3078, opened 2024-12-13
Add OpenGraph tags to playlist page head

MonkeyDo:
These were requested [on the forums](https://community.metabrainz.org/t/syndication-feeds-atom-rss-now-out-for-beta-testing/726559/18), now that we have Atom feeds for playlists and the embedded previews that use OG tags look very generic.

Followed docs at https://ogp.me/#type_music

My initial implementation used react-helmet (which we already use for the purpose of modifying head tags), but unfortunately that means the page must first load and run the react app before the tags are modified/added.

The solution required moving back to the flask/jinja template layer as those will be rendered to HTML directly.

I've modified our main index.html jinja template to support known and arbitrary opengraph tags, with sane defaults (same as before this PR) for title, description and type.

Once we have images for playlist thumbnails, we could consider adding that as og:image tags and testing that out.

Here is the resulting head when loading a playlist page:
image

N.B.: Considering LB is now a single-page-app with routing on the client-side, the opengraph tags rendered by flask will stay there even after you navigate to another page where (some of) those tags don't apply. This is not really an issue, as the meta tags are meant to be parsed by a machine once on page load while the client-side navigation is all about users in their browser page.


MonkeyDo:
I am not sure about the `description` tag content. Currently I just put the playlist description, but it might be better to have something like "Playlist by mr_monkey — 123 listens — ListenBrainz" similar to what I ended up doing in #3118 , for consistency (and also because it might be more useful info?)


Any opinions?

frontend/js/src/playlists/Playlist.tsx
```diff
@@ -319,8 +320,29 @@ export default function PlaylistPage() {
 
   return (
     <div role="main">
-      <Helmet>
-        <title>{playlist.title} - Playlist</title>
+      <Helmet defer={false}>
+        <title>
```
anshg1214, line 324: What does defer here do? I was reading through the available documentations, but couldn't find it there.

MonkeyDo: Left over from my previous (failed) attempt at getting the meta tags to be readable directly at page load, will remove.

> {/* (optional) set to false to not use requestAnimationFrame and instead update the DOM as soon as possible.
        Useful if you want to update the title when the tab is out of focus 
    */}
    defer={false}

MonkeyDo:
I have modified the description to show creator name and number of tracks instead of the playlist text description.

> `<meta property="og:description" content="Playlist by mr_monkey — 1 track — ListenBrainz">`

anshg1214 (approved):
🚀

PR #3080, opened 2024-12-14
Update example data for /metadata/ endpoints

samglt:
This updates some out of date example data for some of the /metadata/ endpoints:

/recording: tags for release group, updated artist shape, more information in release
/release_group: the example file was missing
/lookup: example was referred to but not included
/artist: used to be a copy of /recording

Also shortened the lists of tags so they're not super long on the page


MonkeyDo (approved):
Thank you very much for this long overdue cleanup!

PR #3081, opened 2024-12-15
LB-1374: Add sort options for playlists

anshg1214:
LB-1374


MonkeyDo (changes requested):
It's working quite well!
Nice work.

The sort option should be reset to default value when navigating to collab playlists and back in `setPlaylistType`, or alternatively the playlists should be sorted again when you switch back.

MonkeyDo (approved):
Lesssgoooo !!!

PR #3083, opened 2024-12-18
LB-1706: RSS Feed shows < and > properly now

Suvid-Singhal:
# Problem

This PR fixes the issue LB-1706. The RSS Feed shows < and > character properly now.

# Solution

Replaced &lt; and &gt; with < and > in the generated RSS Feed string.

### DIFF
--- a/frontend/js/src/settings/link-listens/MultiTrackMBIDMappingModal.tsx
+++ b/frontend/js/src/settings/link-listens/MultiTrackMBIDMappingModal.tsx
@@ -300,7 +300,8 @@ export default NiceModal.create(
       matchingTracks && Object.entries(matchingTracks);
     const hasMatches = Boolean(matchingTracksEntries?.length);
     const unmatchedItems =
-      unlinkedListens.filter((md) => !matchingTracks?.[md.recording_msid]) ?? [];
+      unlinkedListens.filter((md) => !matchingTracks?.[md.recording_msid]) ??
+      [];
 
     // We may need to escape or replace the Lucene search special characters
     // + - && || ! ( ) { } [ ] ^ " ~ * ? : \ /     as described in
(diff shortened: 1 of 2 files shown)


MonkeyDo (commented):
Hi @Suvid-Singhal and thanks for opening a PR!
It is very much appreciated :)

I have a couple of comments/questions:
The first one is whether this is required, and whether it will break atom readers.
The reason I ask is because these feeds currently seem to be working fine once viewed in an atom reader, with HTML links working as expected.
So, assuming this PR is to make the XML for the feed readable by humans, we need to ensure that it does not break the main purpose of these feeds to be interacted with in a feed reader.

The second comment is whether this encoding issue can be solved more generally for all encoded HTML entities rather than just the `<` and `>` symbols.

And a final coding style comment: if code is repeated in a bunch of places, then it needs to be refactored so that it is defined in one place and called from multiple places, to avoid issues in the future.

MonkeyDo:
EDIT: I did a tiny bit or research, and according to the Atom specifications the content of `type="html"` should be escaped:

> If type="html", then this element contains entity escaped html.
> 
> <title type="html">
>   AT&amp;amp;T bought &amp;lt;b&amp;gt;by SBC&amp;lt;/b&amp;gt;!
> </title>
From https://validator.w3.org/feed/docs/atom.html#text

So unless I misunderstood something, it is actually expected that HTML entities in a feed item's content should be escaped for HTML parsing, rather than be human-readable.
I'm going to close the PR and the associated ticket, but of course if you dig in more and find that I am incorrect feel free to reopen both!

Suvid-Singhal:
I get it. Thanks @MonkeyDo, I also learnt a new thing today 😅

MonkeyDo:
> I get it. Thanks @MonkeyDo, I also learnt a new thing today 😅

You and me both!

PR #3087, opened 2024-12-18
Add Genre Graph for YIM 2024

anshg1214:


MonkeyDo (commented):
Looking nice!

We might want to adjust the colors to use the ones defined in the YIM24 file (the "text" colors), with variations, maybe?
https://github.com/metabrainz/listenbrainz-server/blob/yim-2024/frontend/js/src/user/year-in-music/2024/YearInMusic2024.tsx#L63-L68

Try to tie it up with the rest of the page look.
That being said I'm going to merge this to avoid further merge conflicts.

The tag counts don't look right to me at all, as they all have crazy number 5 times higher than my actual listen count for the year. Not sure there's anything we can do here though.

anshg1214:
>We might want to adjust the colors to use the ones defined in the YIM24 file (the "text" colors), with variations, maybe?

I had that in mind, but if we do it the graph does not look good. 
image

>The tag counts don't look right to me at all, as they all have crazy number 5 times higher than my actual listen count for the year. Not sure there's anything we can do here though.

@amCap1712 As I can see in the [SQL query](https://github.com/metabrainz/listenbrainz-server/blob/b4bc18cf9d343956819b4a3038674e50e2ec4c40/listenbrainz_spark/year_in_music/top_genres.py), instead of merging the genre count from recording, artist and release group, can we have that data separate?

MonkeyDo:
> > We might want to adjust the colors to use the ones defined in the YIM24 file (the "text" colors), with variations, maybe?
> 
> I had that in mind, but if we do it the graph does not look good. image
> 
I meant more using the four colors, regardless of the chosen season. So where other graph use the selected season's textColor, this graph would use all 4 textColor from the seasons. I reckon we might still have the same issue of not enough variety between colours, but maybe it works?

anshg1214:
This looks good! 

image

PR #3088, opened 2024-12-19
Add year to release color

mayhem:
Add year to release_color table in the mapping. This was going to be necessary for the mosaic API endpoint, but we decided against doing this. However, this will still be useful for Huesound when we add the ability to restrict data to a year range.

These tables have been deployed and updated in production.

### DIFF
--- a/mbid_mapping/mapping/release_colors.py
+++ b/mbid_mapping/mapping/release_colors.py
@@ -268,7 +268,7 @@ def incremental_update_release_color_table():
                       ON cat.id = caa.id
                     JOIN musicbrainz.release r
                       ON caa.release = release.id
-                    JOIN release_first_release_date rfrd
+               LEFT JOIN release_first_release_date rfrd
                       ON rfrd.release = r.id
                    WHERE type_id = 1
                      AND caa.date_uploaded > %s


amCap1712 (approved):
LGTM, Thanks!

mbid_mapping/mapping/release_colors.py
```diff
@@ -259,11 +262,14 @@ def incremental_update_release_color_table():
...
                        , mime_type
                        , date_uploaded
+                       , year
                     FROM cover_art_archive.cover_art caa
                     JOIN cover_art_archive.cover_art_type cat
                       ON cat.id = caa.id
-                    JOIN musicbrainz.release
+                    JOIN musicbrainz.release r
                       ON caa.release = release.id
+                    JOIN release_first_release_date rfrd
```
amCap1712, line 271: This will exclude releases that have a cover art but no release year. If you don't intend to do that, use a LEFT JOIN.

amCap1712: I just saw the SQL and I had misread it initially yeah you need to use a LEFT JOIN here to avoid missing albums.

mayhem: Good catch, thanks!

PR #3089, opened 2024-12-20
Update AddData.tsx to include Last.fm

Aerozol:
Added Last.fm to the 'Automatically submit from...' section in add-data.

Thought about adding it under its own header, but the disclaimer text is the same, as well as the link it points to, so I have combined it with the Spotify section.

Now we should be able to link to this one page (e.g. in the posts about YIM that I'm doing) and a user will be able to see *all* their submission options.


amCap1712 (approved):
LGTM, Thanks!

PR #3093, opened 2024-12-23
fix: Fetch Artist info for redirected MBIDs

anshg1214:
As reported on https://community.metabrainz.org/t/empty-neighboor-bubble/732844, the artist information is not being fetched for redirected MBID. 

This PR fixes that by fetching the data for the `redirected_mbids`.


MonkeyDo (approved):
LGTM

PR #3096, opened 2024-12-27
fix: Render Annotations for Playlist on Search page

anshg1214:
**Before:**
image

**After:** 
image

### DIFF
--- a/frontend/js/src/search/PlaylistSearch.tsx
+++ b/frontend/js/src/search/PlaylistSearch.tsx
@@ -137,7 +137,12 @@ export default function PlaylistSearch(props: PlayListSearchProps) {
                       {playlist.title}
                     </Link>
                   </td>
-                  <td>{playlist?.annotation}</td>
+                  <td
+                    // eslint-disable-next-line react/no-danger
+                    dangerouslySetInnerHTML={{
+                      __html: playlist?.annotation || "",
+                    }}
+                  />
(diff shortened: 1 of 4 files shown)


frontend/js/src/search/PlaylistSearch.tsx
```diff
@@ -137,7 +137,12 @@ export default function PlaylistSearch(props: PlayListSearchProps) {
                       {playlist.title}
                     </Link>
                   </td>
-                  <td>{playlist?.annotation}</td>
+                  <td
```
MonkeyDo, line 140: I think while we're in here, maybe we should limit the number of characters/lines that we show here. I can imagine some gigantic descriptions ruining the page...

We have the `ellipsis-2-lines` utility class for this, but I think 2 lines isn't sufficient.

I would suggest creating an `ellipsis-4-lines` CSS class with:
```css
.ellipsis-4-lines {
  display: -webkit-box;
  -webkit-line-clamp: 4;
  -webkit-box-orient: vertical;
  overflow: hidden;
  text-overflow: ellipsis;
}
```

Currently the `ellipsis-2-lines` class is defined in listens-page.less file, but it would be good to move it to.... maybe main.less? Or a new utilities.less file.
In any case, the new `ellipsis-4-lines` class should live alongside 2-lines.


After:
image

Before:
image

PR #3097, opened 2024-12-27
LB-1702: Added Volume Control Slider to BrainzPlayer

Suvid-Singhal:
# Problem

Earlier, ListenBrainz used to play music in max volume by default. This feature aims to let users control the volume of music in ListenBrainz itself.
The feature request ticket was LB-1702.

# Solution
I created an input range component. Also added some hooks and states. The UI is basic as of now and can be improved further I believe.
The demo is attached below:

https://github.com/user-attachments/assets/b7efde23-8880-4e99-9323-cbd2b7c3f25e

# Action
I tested the new feature for YouTube and Spotify but couldn't test it for SoundCloud and Apple Music. I couldn't test it for SoundCloud as they are not accepting new application requests right now so couldn't get the API key. Couldn't test it for Apple Music as I don't have Apple Music premium 😅

If anyone could test it for these two platforms it would be great. Also, please suggest UI changes, if any.

P.S. It's my biggest open-source contribution till date. Thanks @anshg1214 for guiding and supporting me :)


frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
```diff
@@ -617,6 +617,14 @@ export default class SpotifyPlayer
...
-    return <div data-testid="spotify-player">{this.getAlbumArt()}</div>;
+    return (
+      <BrainzPlayerContext.Consumer>
+        {(context) => {
+          const { volume } = context;
+          this.spotifyPlayer?.setVolume(volume / 100);
+          return <div data-testid="spotify-player">{this.getAlbumArt()}</div>;
+        }}
+      </BrainzPlayerContext.Consumer>
+    );
```
anshg1214, line 628: How about instead of creating a consumer for the context, we pass down the volume as props?

It would simplify things and also prevent re-rendering whenever the context values are updated

frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
```diff
@@ -617,6 +617,14 @@ export default class SpotifyPlayer
...
   }
```
anshg1214, line 629: Adding a comment here as I cannot comment on [this code piece ](https://github.com/metabrainz/listenbrainz-server/pull/3097/files#diff-3a174aac201e54fa841962b156d84592132542e10d10f7eacb529c488df2a80eR443-R466). 

While playing a track on spotify you should also change the volume being set here.

anshg1214: You should also add a listener for volume change, so that if someone changes volume on the spotify app, the changes are reflected here.

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -75,6 +76,34 @@ function PlaybackControlButton(props: PlaybackControlButtonProps) {
   );
 }
 
+function VolumeControlButton() {
```
anshg1214, line 79: Let's move this to a new file?

frontend/js/src/common/brainzplayer/BrainzPlayerContext.tsx
```diff
@@ -83,6 +86,7 @@ export type BrainzPlayerActionType = Partial<BrainzPlayerContextT> & {
     | "ADD_LISTEN_TO_BOTTOM_OF_AMBIENT_QUEUE"
     | "ADD_MULTIPLE_LISTEN_TO_BOTTOM_OF_AMBIENT_QUEUE";
   data?: any;
+  payload?: any;
```
anshg1214, line 89: Instead of payload, you can use the data variable.

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -75,6 +76,34 @@ function PlaybackControlButton(props: PlaybackControlButtonProps) {
...
+      ref={volumeRef}
+      onMouseUp={handleVolumeChange}
+      className="volume"
+      type="range"
+      defaultValue="100"
+      max="100"
+      min="0"
+      step="5"
+    />
+  );
```
anshg1214, line 104: The volume slider is not visible on the mobile view.

image

anshg1214 (commented):
Thank you for taking up this ticket. The functionality works as expected. But there are some improvements which can be done here.

frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
anshg1214: This improvement can be made for all the players.

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
Suvid-Singhal: In mobile view, even the previous song button is also not visible. Could you please suggest improvements that can be done in order to accomodate the volume button properly?

MonkeyDo: I would say you probably want to hide it on mobile devices for now, and wait for #2940 to take care of the rest.

To hide on mobile you can use the bootstrap classes `hidden-xs`  or maybe `hidden-sm`, you will have to text different screen sizes.

Suvid-Singhal: Done @MonkeyDo :)
The volume slider is no longer visible on xs and sm viewports.

frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
```diff
@@ -357,7 +358,8 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
   };
 
   render() {
-    const { show } = this.props;
+    const { show, volume } = this.props;
+    this.youtubePlayer?.setVolume(volume ?? 100);
```
MonkeyDo, line 362: Same as for the others, this shouldn't be called in the render function but rather after setting up `this.youtubePlayer`, probably in the `onReady` function

frontend/js/src/common/brainzplayer/SoundcloudPlayer.tsx
```diff
@@ -404,7 +405,9 @@ export default class SoundcloudPlayer
   };
 
   render() {
-    const { show } = this.props;
+    const { show, volume } = this.props;
+    this.soundcloudPlayer?.setVolume((volume ?? 100) / 100);
```
MonkeyDo, line 409: This shouldn't be called here but rather most likely in the `onReady` function, or when setting up soundcloudPlayer on line 145

MonkeyDo (changes requested):
UI looks good, thanks.
I took another look at the code and have a bit more feedback, sorry for trickling the feedback bit by bit!

frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
```diff
@@ -613,7 +612,8 @@ export default class SpotifyPlayer
   };
 
   render() {
-    const { show } = this.props;
+    const { show, volume } = this.props;
+    this.spotifyPlayer?.setVolume((volume ?? 100) / 100);
```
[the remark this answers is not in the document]
Suvid-Singhal: It doesn't change the volume if I only set volume in the new window.Spotify.Player and not update it in render function.
I think that might be because that function is used to set up Spotify player initially with the given initial volume.
That's why I had to put volume state in render function of every player.

frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
Suvid-Singhal: Hi,
I can't think of any way to track changes in volume from the spotify app 🤔
I tried modifying the existing player_state_changed event listener but it doesnt seem to work.
Can you please help me out in this?
Thanks

frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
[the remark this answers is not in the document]
MonkeyDo: according to the documentation you should indeed be able to set the volume in the Spotify.Player constructor: https://developer.spotify.com/documentation/web-playback-sdk/reference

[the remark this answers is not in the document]
Suvid-Singhal: Fixed for all the players :)

frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
Suvid-Singhal: > While playing a track on spotify you should also change the volume being set here.
Fixed this by putting setVolume in has_player_state_changed event handler. The volume update in spotify happens whenever this event is fired now.
I don't think so we can track the volume change in spotify app to update on brainzplayer for now.

MonkeyDo: Here's a question for @anshg1214 : How do you wan to synchronize volume with spotify?
If the volume changes in spotify app and we manage to get the value, should it change the volume for BP?
If you change the volume for BP, should it change the value of spotify app volume?

That seems maybe a bit too intertwined

anshg1214: Yes. If we can have that feature that'll be great. We do support change handler such that if someone changes track from the spotify app, the states are updated in BP.

However, if spotify does not support this, we can drop this for now.

frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
```diff
@@ -148,7 +149,11 @@ export default class AppleMusicPlayer
   }
 
   componentDidUpdate(prevProps: DataSourceProps) {
-    const { show } = this.props;
+    const { show, volume } = this.props;
+    const player = this.appleMusicPlayer;
+    if (player) {
```
MonkeyDo, line 154: For each of these calls in componentDidUpdate, you need to check if the `volume` prop has changed:
```suggestion
    if (prevProps.volume !== volume && player) {
```

frontend/js/src/common/brainzplayer/VolumeControlButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+function VolumeControlButton() {
+  const dispatch = useBrainzPlayerDispatch();
+  const volumeRef = React.useRef<HTMLInputElement>(null);
+  const handleVolumeChange = () => {
+    dispatch({
+      type: "VOLUME_CHANGE",
+      data: volumeRef?.current?.value ?? 100,
+    });
+  };
```
MonkeyDo, line 12: You don't need a ref for this at all:
```suggestion
  const handleVolumeChange = (e: React.MouseEvent<HTMLInputElement>) => {
    dispatch({
      type: "VOLUME_CHANGE",
      data: e.currentTarget?.value ?? 100,
    });
  };
```

Don't forget to remove the ref prop on the input element

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -262,6 +263,7 @@ function BrainzPlayerUI(props: React.PropsWithChildren<BrainzPlayerUIProps>) {
             disabled={disabled}
           />
         </div>
+        <VolumeControlButton />
```
MonkeyDo, line 266: For my proposal to turn this into a volume icon that you click/hover on to show the input, you would need to move this one line below.

frontend/js/src/common/brainzplayer/VolumeControlButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <input
```
MonkeyDo, line 14: Instead of rendering the input directly in here, I suggest rendering a volume icon. When clicking on the icon, the volume input will appear above it.
For CSS styles for the hidden/shown input, you can take inspiration from the player queue (the three horizontal bars in the player UI)
https://github.com/metabrainz/listenbrainz-server/blob/f40c908885229d42c5cb7716ea57a54d39ed2ca4/frontend/css/brainzplayer.less#L354C2-L370C4

MonkeyDo (changes requested):
You found a good way to do volume setting in componentDidUpdate, albeit with a small prop check required. Nice work on that front!

However I think the UI needs some more work.

The input on the player bar does not look very good on desktop sizes. It is very wide for no reason, and its placement leaves to be desired. Not sure if you looked at it in desktop size, but it's massive and bright colored, it makes it look like a progress bar...
image


Usually, what I expect from this type of media player is one of two things:
- a volume button, which on hover or click will show a vertical range input above it
image
- a small horizontal range input like what you see on youtube, spotify, apple music in their respective player controls
image
image
image

Considering we already show the player queue by sliding it up (click the stacked horizontal bars in the player), maybe the first solution makes the most sense in this case. It would also work for smaller screen sizes more easily, as we would only need space for one extra volume icon in the player bar:
image


Do let me know if you have questions or if you are unsure about anything, better to talk first to figure things out before implementing :)

Suvid-Singhal:
Made all the required changes. It now looks like this:

https://github.com/user-attachments/assets/a7e4f1d2-7121-4566-9c71-8847c911828b

frontend/css/brainzplayer.less
```diff
@@ -351,6 +351,23 @@
   }
 }
 
+.volume {
+  position: fixed;
+  width: fit-content;
+  z-index: 109;
```
MonkeyDo, line 357: This will put the volume component on top of the queue if it is open, otherwise users might not see that there is a component at all and think the button does nothing.
```suggestion
  z-index: 110;
```

frontend/css/brainzplayer.less
```diff
@@ -351,6 +351,23 @@
...
+  height: 10em;
+  background-color: #f8f8f8;
+  right: 16em;
+  .volume-input {
```
MonkeyDo, line 361: Let's animate the transition between hidden and shown:
```suggestion
  bottom: -100px;
  transition: bottom 0.5s ease-in-out;
  .volume-input {
```

frontend/css/brainzplayer.less
```diff
@@ -351,6 +351,23 @@
...
+    width: 7em;
```
MonkeyDo, line 362: I think this is a bit wide, I suggest:
```suggestion
    width: 3.5em;
```

frontend/css/brainzplayer.less
```diff
@@ -351,6 +351,23 @@
...
+    margin-top: @brainzplayer-height;
+    transform: rotate(270deg);
```
MonkeyDo, line 364: I have a fun CSS trick for you today!
For vertical range inputs, you can use:
```suggestion
    writing-mode: vertical-rl;
    direction: rtl;
    margin: 1.5em 0.6em;
```

frontend/css/brainzplayer.less
```diff
@@ -351,6 +351,23 @@
...
+  }
+  &.show {
+    bottom: @brainzplayer-height;
+  }
+}
+
```
MonkeyDo, line 370: Let's make some styles common to .volume and .queue.

These are the properties I can tell will help have a consistent look between the two:
```css
  transition: bottom 1s ease-in-out;
  background-color: #f8f8f8;
  border-top: 1px solid #ccc;
  box-shadow: -6px -1px 10px rgba(0, 0, 0, 0.2);
  &.show {
    bottom: @brainzplayer-height;
  }
```

frontend/css/brainzplayer.less
```diff
@@ -351,6 +351,23 @@
   }
 }
 
+.volume {
```
MonkeyDo, line 354: One more style suggestion to be applied to the whole website, so to be added to the frontend/css/main.less file:

```less
input[type='range'] {
    accent-color: @blue;
}
```

This will style the color of the range input:
image

frontend/css/brainzplayer.less
```diff
@@ -351,6 +351,23 @@
...
+  position: fixed;
+  width: fit-content;
+  z-index: 109;
+  height: 10em;
+  background-color: #f8f8f8;
+  right: 16em;
```
MonkeyDo, line 360: to go with my width changes:
```suggestion
  right: 17em;
```

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -264,20 +271,29 @@ function BrainzPlayerUI(props: React.PropsWithChildren<BrainzPlayerUIProps>) {
...
+                  color={currentDataSourceIconColor}
+                />
+              </a>
               <FontAwesomeIcon
-                icon={currentDataSourceIcon!}
-                color={currentDataSourceIconColor}
+                icon={faVolumeUp}
+                style={{ color: showVolume ? "green" : "" }}
+                onClick={() => setShowVolume(!showVolume)}
+                className="hidden-sm hidden-xs"
```
MonkeyDo, line 292: After seeing it in action, I think it should only be
```suggestion
                className="hidden-xs"
```

MonkeyDo (changes requested):
OK, I have a bunch more small suggestions, don't hate me :p
I think after this batch we'll be good, solid and consistent.

Here's where i ended up at:
image

Thanks for your patience!

frontend/css/brainzplayer.less
Suvid-Singhal: I also thought of including transition for 0.5s initially but then dropped it later 😅

MonkeyDo:
OK, I see what I forgot to copy over from my experimentations :)

In the .volume css class, remove the `height` declaration entirely>
Should be left with a good looking input:
image

Suvid-Singhal:
Was having this issue after removing height declaration. But I fixed it by changing `bottom: -100px;` to `bottom: -12em;`

Screenshot from 2025-01-17 17-29-58

Now I feel that it looks perfect on testing. 😅

MonkeyDo (approved):
Thanks for the good work!
Everything looking fine now :rocket:

PR #3099, opened 2024-12-28
LB-1687: Add mapping info to listen exports

amCap1712:
The mapping information is available in the listens api, make it available in listen exports as well.


mayhem (approved):
:ship:

PR #3100, opened 2024-12-28
Improve Playlist Page

anshg1214:


frontend/css/playlists.less
```diff
@@ -40,34 +89,86 @@
...
-          right: 0;
-          width: 1em;
-          height: 1em;
-          margin-top: 0.2em;
-          background-color: inherit;
+      }
+    }
+
+    .playlist-more-info {
+      align-self: center;
```
MonkeyDo, line 153: Add this to prevent shrinking when there is a description:
```suggestion
      align-self: center;
      flex-shrink: 0;
```

Before:
image

After:
image

frontend/css/playlists.less
```diff
@@ -1,11 +1,60 @@
...
+			*/
+  .description {
+    overflow: hidden;
+    position: relative;
+    line-height: 1.2em;
+    max-height: @descriptionLines * 1.2em;
+    text-align: justify;
+    padding-right: 2em;
+    display: -webkit-box;
+    line-clamp: @descriptionLines;
```
MonkeyDo, line 36: The line clamp doesn't combine well with the fact that we have HTML content.
If I use a paragraph, the margin will throw the line height calculation out the window and crop the text awkwardly:
image

You can however remedy this with 
```css
#playlists-container .description p {
    margin-bottom: 0;
}
```

image

frontend/css/playlists.less
```diff
@@ -1,11 +1,60 @@
...
+    }
+    &:after {
+      content: "";
+      position: absolute;
+      right: 0;
+      width: 1em;
+      height: 1em;
+      margin-top: 0.2em;
+      background-color: inherit;
+    }
```
MonkeyDo, line 55: I'm not certain we need these. I believe it is made as a fallback fro the line-clamp property if that's not implemented. I don't think it combines with the line-clamp stuff, and it shows two sets of ellipsis once set up properly, so we probably just want to drop it.
 
If we still want to use them, there are currently some issues with these pseudo elements:
- display:none on :before prevents rendering of the element
- :after is missing `bottom:0` for placement (reading the code comments on the article) and the `right` position needs to be the same as :before
- :after background-color cannot be inherit, which results in transparent. It should be the same color as the card background (`white` in this case)

MonkeyDo (commented):
Looking good!
Some UI suggestions below, but funcionnally speaking it's working perfectly :ok_hand:

frontend/js/src/user/playlists/Playlists.tsx
```diff
@@ -207,141 +230,171 @@ export default class UserPlaylists extends React.Component<
...
-            <option value={SortOption.RANDOM}>Random</option>
-          </select>
+            <div className="playlist-view-controls">
+              <Pill
+                active={view === PlaylistView.GRID}
+                type="secondary"
+                onClick={() => this.setState({ view: PlaylistView.GRID })}
+                title="Grid view"
+              >
+                <FontAwesomeIcon icon={faGrid} />
```
MonkeyDo, line 259: I've been using the `fixedWidth` option for any button/pill that only has an icon, it ensure we have buttons of the same size:
```suggestion
                <FontAwesomeIcon icon={faGrid} fixedWidth />
```

frontend/js/src/user/playlists/Playlists.tsx
```diff
@@ -207,141 +230,171 @@ export default class UserPlaylists extends React.Component<
...
+              </Pill>
+              <Pill
+                active={view === PlaylistView.LIST}
+                type="secondary"
+                onClick={() => this.setState({ view: PlaylistView.LIST })}
+                title="List view"
+              >
+                <FontAwesomeIcon icon={faStacked} />
```
MonkeyDo, line 267: Maybe the regular list view icon here?
In any case, please add `fixedWidth`
```suggestion
                <FontAwesomeIcon icon={faList} fixedWidth/>
```

MonkeyDo (commented):
Looking great!
Last little change and it's ready to merge.

MonkeyDo (approved):
Yussss, thanks !

PR #3101, opened 2024-12-29
fix: Donation Modal not closing on navigating

anshg1214:


MonkeyDo (approved):
Makes sense.
I'm getting really tired of the interractions between NiceModal and old Bootstrap.
I foresee moving to React-Bootstrap in my near future

PR #3103, opened 2024-12-29
LB-1715: BrainzPlayer overlaps filters on Fresh Releases page

Suvid-Singhal:
# Problem

The problem is mentioned in this ticket: [LB-1715](https://tickets.metabrainz.org/browse/LB-1715)
Some of the fresh releases sidebar options were hidden due to the BrainzPlayer.

# Solution

Made a new class and changed the max-height to 85vh instead of 100vh. Applied that class to the sidebar component in fresh releases page.

Ubuntu-2024-12-29-23-50-17

### DIFF
--- a/frontend/css/fresh-releases.less
+++ b/frontend/css/fresh-releases.less
@@ -316,10 +316,10 @@
   }
   .sidebar-fresh-releases {
     max-height: 85vh;
-    
+
     @media screen and (max-width: @screen-md) {
       max-height: calc(100vh - @brainzplayer-height);
-    }    
+    }
   }
 }
 
--- a/frontend/css/fresh-releases.less
+++ b/frontend/css/fresh-releases.less
@@ -314,6 +314,9 @@
   .sidebar {
     flex: 0 0 250px;
   }
+  .sidebar-fresh-releases {
+    max-height: 85vh;
+  }
 }
 
 .sidebar {
--- a/frontend/css/fresh-releases.less
+++ b/frontend/css/fresh-releases.less
@@ -316,6 +316,10 @@
   }
   .sidebar-fresh-releases {
     max-height: 85vh;
+    
+    @media screen and (max-width: @screen-md) {
+      max-height: calc(100vh - @brainzplayer-height);
+    }    
   }
 }
 
--- a/frontend/css/sidebar.less
+++ b/frontend/css/sidebar.less
@@ -74,10 +74,6 @@
   }
 }
 
-.sidebar-fresh-releases {
-  max-height: 85vh;
-}
-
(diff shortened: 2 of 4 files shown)


frontend/css/sidebar.less
```diff
@@ -74,6 +74,10 @@
   }
 }
 
+.sidebar-fresh-releases {
+  max-height: 85vh;
+}
+
```
anshg1214, line 80: Since this css is only relevant for fresh-releases page, you can consider moving it to [that page's css file](https://github.com/suvid-singhal/listenbrainz-server/blob/3734db6c52ad621c16f1dd01346ae4cb420a65f1/frontend/css/fresh-releases.less)

frontend/css/sidebar.less
```diff
@@ -74,6 +74,10 @@
   }
 }
 
+.sidebar-fresh-releases {
+  max-height: 85vh;
```
anshg1214, line 78:
```suggestion
  max-height: calc(100vh - @brainzplayer-height - 20px);
```

I think this looks much better. What do you think?

anshg1214 (changes requested):
Thank you for taking up this ticket.  A few changes, and it should be good to go.

frontend/css/sidebar.less
Suvid-Singhal: Hi
The problem with this is that it works fine when the page loads and gets scrolled a bit down to the first date in the timeline, but when I scroll it up to top and then scroll the sidebar to the bottom, the information filter is hidden again 😅(screenshot attached below).

image


Whereas in my solution, this problem does not occur, so I feel that my approach might be a bit better as it displays correctly in the given situation as well (screenshot attached below).
image

frontend/css/sidebar.less
Suvid-Singhal: Done :)

frontend/css/fresh-releases.less
```diff
@@ -314,6 +314,9 @@
   .sidebar {
     flex: 0 0 250px;
   }
+  .sidebar-fresh-releases {
+    max-height: 85vh;
+  }
```
anshg1214, line 319:
```suggestion
  .sidebar-fresh-releases {
    max-height: 85vh;
    
    @media screen and (max-width: @screen-md) {
      max-height: calc(100vh - @brainzplayer-height);
    }    
  }
```

I believe you should also add media queries to improve the view on smaller screen devices.

image

anshg1214 (changes requested):
One small change, and we're good to go.

anshg1214 (approved):
Thank you!

PR #3107, opened 2025-01-01
fix: Text not visible in YIM 2024 Stats image

anshg1214:
The text color and background color are being set in the backend and are correctly returned by the endpoint. However, when you try downloading the image, the custom styles change the text color, making it difficult to read.

Since these styles are not necessary, I have removed them.

Before:
image

After:
image


MonkeyDo (approved):
Thank you!

PR #3108, opened 2025-01-01
Dockerfile change

fettuccinae:
Screenshot (1)

 When building the docker images, it was giving several warnings about FromAsCasting and LegacyKeyValueFormat

  Capitalized as in "FROM x as y" and added equal sign to "ENV key value"


amCap1712 (approved):
LGTM, Thanks!

PR #3109, opened 2025-01-03
LB-1719: Fix CORS error on Feedback endpoint

anshg1214:


amCap1712 (approved):
LGTM, Thanks!

I looked into this further because this shouldn't be needed, @crossdomain decorator should have been enough. There are a couple of misconfiguration in the decorator and the catch-all index path, I'll open PRs to fix the root cause of the issue separately but for now let's get this fix released.

fettuccinae:
image
im still getting the same error

PR #3112, opened 2025-01-04
Use custom converter for catch-all index path

amCap1712:
#3045 fixed the 404 errors for non-existent urls but the implementation masks CORS issues with other endpoints. The catch-all index_pages endpoint , by default, has provide_automatic_options enabled and as a result it matches OPTIONS requests for non-existent urls or any other endpoints that have CORS incompletely configured. This makes it harder to debug CORS issues. To avoid this instead of throwing a 404 after matching, we need to throw the error during matching itself. Hence, use a custom path converter that matches all paths except those starting with the api prefix.

### DIFF
--- a/listenbrainz/webserver/converters.py
+++ b/listenbrainz/webserver/converters.py
@@ -1,0 +1,10 @@
+from werkzeug.routing import PathConverter, ValidationError
+
+
+class NotApiPathConverter(PathConverter):
+
+    def to_python(self, path):
+        print(path)
+        if path.startswith('1/'):
+            raise ValidationError()
+        return path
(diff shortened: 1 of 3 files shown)


listenbrainz/webserver/converters.py
```diff
@@ -0,0 +1,10 @@
+from werkzeug.routing import PathConverter, ValidationError
+
+
+class NotApiPathConverter(PathConverter):
+
+    def to_python(self, path):
+        print(path)
```
mayhem, line 7: Should this be kept?

mayhem (approved):
Thanks!

PR #3113, opened 2025-01-05
Fix @crossdomain decorator

amCap1712:
When handling OPTIONS requests, Flask can either generate an automatic response internally or route the request to the view function for handling. The request is routed to the view function if the view function has a provides_automatic_options attribute set to False or if the methods list passed to route definition includes OPTIONS. OPTIONS requests handled by Flask internally don't have CORS headers. Hence, API endpoints that need to support CORS should ensure that the OPTIONS request is routed to the view function.

The https://github.com/crossdomain decorator sets the provides_automatic_options attribute on the view function which should have ensured this. However, due to the way decorators work the attribute should have instead been set on the decorator function itself for Flask to pick it up. However, we often list OPTIONS in the methods list of the route anyway when applying https://github.com/crossdomain. As a result, Flask would still route the request as expected. The errors would only happen when the OPTIONS method was not mentioned in the route's methods list.

https://github.com/crossdomain decorator has now been fixed to set the attribute correctly and also add OPTIONS to the methods list automatically. Further, a test has been added to ensure that all API endpoints support CORS.

While at it, also refactor the views to use newer .get/.post decorators insteads of methods list.

### DIFF
--- a/listenbrainz/webserver/decorators.py
+++ b/listenbrainz/webserver/decorators.py
@@ -6,7 +6,11 @@ from listenbrainz.webserver import timescale_connection
 
 
 def crossdomain(f):
-    """ Decorator to add CORS headers to flask endpoints """
+    """ Decorator to add CORS headers to flask endpoints.
+
+    This decorator should be applied just after the route to ensure the provide_automatic_options
+    is set correctly.
+    """
     @wraps(f)
     def decorator(*args, **kwargs):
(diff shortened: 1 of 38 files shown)


amCap1712:
As this touches almost all views, deployed on beta for testing for a while.

mayhem (approved):
Lgtm

PR #3115, opened 2025-01-07
Incremental user stats

amCap1712:
Builds on #3114 and incrementalizing user stats.


mayhem (approved):
Please document the base classes as discussed in IRC. After that, feel free to merge.

I reviewed this PR, but not really line by line. #justbeinghonest

MonkeyDo:
I stopped and re-ran the spark test CI suite, which looked stuck.
It is still stuck after the second try, so I suspect something is actually wrong in the tests.

PR #3117, opened 2025-01-09
Accept potentially invalid dates

MonkeyDo:
A user stumbled upon an issue with a partial date with a year and day but no month ("1993-??-01").
The browser's date parser does not like that at all and throws an error which breaks the page entirely.
We need to check if the date is valid before parsing, in a few places in the code where we might receive such dates.

For reference, the issue was trying to add [this recording](https://musicbrainz.org/recording/fe6bb919-b5d5-4e59-a096-4eda7ba4e504) for [this release](https://musicbrainz.org/release/774d0f07-a554-4be0-b259-f1e492e4cfc9)


anshg1214 (approved):
LGTM

PR #3118, opened 2025-01-09
Add OpenGraph meta tags in artist, album and user pages

MonkeyDo:
### To be merged *after* #3078 

We talked SEO and OpenGraph tags recently with Ansh, and as I have been working on playlist OpenGraph tags  in #3078 (as requested by a user [on the forums](https://community.metabrainz.org/t/syndication-feeds-atom-rss-now-out-for-beta-testing/726559/18)), I thought now was the perfect time to implement OG tags in other relevant pages, while it's all fresh in my head.

#### Artist pages:
I haven't added a custom image yet, not exactly sure what to do.
We can't put a link to the SVGs we use in LB pages, because SVGs are not supported in OG tags.
image
image

#### Album pages:
image
image

#### User pages:
image
image

### DIFF
--- a/listenbrainz/webserver/views/entity_pages.py
+++ b/listenbrainz/webserver/views/entity_pages.py
@@ -113,7 +113,7 @@ def release_redirect(release_mbid):
 
 @artist_bp.route("/",  defaults={'path': ''})
 @artist_bp.route("/<artist_mbid>/", methods=["GET"])
-def artist_page(artist_mbid:str):
+def artist_page(artist_mbid: str):
     og_meta_tags = None
     if is_valid_uuid(artist_mbid) and artist_mbid not in {"89ad4ac3-39f7-470e-963a-56509c546377"}:
         artist_data = get_metadata_for_artist(ts_conn, [artist_mbid])
@@ -125,10 +125,12 @@ def artist_page(artist_mbid:str):
             release_group_data = artist.release_group_data
             release_group_mbids = [rg["mbid"] for rg in release_group_data]
             album_count = len(release_group_mbids)
-            listening_stats = get_entity_listener(db_conn, "artists", artist_mbid, "all_time")
+            listening_stats = get_entity_listener(
+                db_conn, "artists", artist_mbid, "all_time")
             total_listen_count = 0
             if listening_stats and "total_listen_count" in listening_stats:
-                total_listen_count = number_readable(listening_stats["total_listen_count"] or 0)
+                total_listen_count = number_readable(
+                    listening_stats["total_listen_count"] or 0)
 
             og_meta_tags = {
                 "title": f'{artist_name}',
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/entity_pages.py
```diff
@@ -111,14 +112,40 @@ def release_redirect(release_mbid):
...
+                total_listen_count = number_readable(
+                    listening_stats["total_listen_count"] or 0)
+
+            og_meta_tags = {
+                "title": f'{artist_name}',
+                "description": f'Artist — {total_listen_count} listens — {album_count} albums — ListenBrainz',
+                "type": "profile",
+                "profile:username": artist_name,
+                "profile.gender": artist.artist_data.get("gender"),
+                "url": f'{current_app.config["SERVER_ROOT_URL"]}/artist/{artist.artist_mbid}',
```
anshg1214, line 141: The current title that we render through react helmet is of the format `artist_name - Listenbrainz`. Let's choose either one of them to maintain consistancy.

Can we also render these tags through React Helmet?

listenbrainz/webserver/views/entity_pages.py
```diff
@@ -209,14 +236,54 @@ def artist_entity(artist_mbid):
...
+            og_meta_tags = {
+                "title": f'{album_name} — {artist_name}',
+                "description": f'Album — {track_count} tracks — {total_listen_count} listens — ListenBrainz',
+                "type": "music.album",
+                "music:musician": artist_name,
+                "music:release_date": release_group.get("release_group").get("date"),
+                "image": f'https://coverartarchive.org/release-group/{release_group_mbid}/front-500',
+                "image:width": "500",
+                "image:alt": f"Cover art for {album_name}",
+                "url": f'{current_app.config["SERVER_ROOT_URL"]}/album/{release_group_mbid}',
```
anshg1214, line 278: Same as above. 

React Helmet renders the title as `album_name - ListenBrainz`.

anshg1214 (changes requested):

listenbrainz/webserver/views/entity_pages.py
MonkeyDo: Also, the page title is a separate issue from these meta tags IMO.
I've opted to have the `— ListenBrainz` addition in the description rather than the title meta tag as it is for use in previews on other websites (I thought it made more sense) and we don't have a lot of space.

That does not change the html page title which is still managed by Helmet with the `— ListenBrainz` added at the end.

In short, I understand the drive for consistency, but I think in this case they are different use-cases and won't be displayed side by side.

Compare for example
image

image

anshg1214: According to https://developers.google.com/search/docs/crawling-indexing/javascript/javascript-seo-basics, the page is rendered before indexing.

So, on the search engines, the title will contain `-ListenBrainz`. Are we fine with that?

MonkeyDo: Yes, I think for SEO that is fine, inn this specific case I'm more worried about the preview displayed to the user.
I think for SEO and for browser tabs it's good to have LB in the title as we're less limited in space.

anshg1214: Cool then, I think we're good to go now.

anshg1214 (approved):
LGTM!

PR #3119, opened 2025-01-10
Fix cursor style and remove extra spaces on stats page

granth23:
# Problem

1. The cursor on the statistics page was the default pointer instead of a hand pointer when hovering over clickable areas.

image
2. There were unnecessary extra spaces on the info link in the statistics page, which appeared multiple times.

image

# Solution

1. Changed the cursor style to cursor: pointer to improve the user experience for clickable areas.

image
2. Removed the extra spaces from the statistics page content wherever they appeared using a margin instead.

image

# Action

1. Please review the changes to ensure they align with the project's style guide.
2. Verify that the updated cursor and spacing improvements work as intended on the live site.

### DIFF
--- a/frontend/css/stats.less
+++ b/frontend/css/stats.less
@@ -23,6 +23,9 @@
     height: calc(100% * 0.6);
     position: relative;
   }
+  &.user-artist-map path[fill]:not([fill="#efefef"]):not([fill="none"]) {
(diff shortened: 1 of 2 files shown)


anshg1214 (approved):
Thank You!

PR #3121, opened 2025-01-10
Use SPA navigation for link to playlist after adding track

rrrnld:
# Problem

When adding a track to a playlist, a toast is displayed containing a success message and a link to the playlist. The link to the playlist will abort any currently playing track, because it doesn't use JS-based navigation to prevent a full reload.

# Solution

I updated the link to use `<Link />` instead of `<a />`.

# Action

I wasn't able to test this locally, so I'm not sure if I accidentally broke the link completely. It builds fine, but there are some (unrelated I think?) failing tests.

### DIFF
--- a/frontend/js/src/common/listens/AddToPlaylist.tsx
+++ b/frontend/js/src/common/listens/AddToPlaylist.tsx
@@ -6,6 +6,7 @@ import {
   faPlusCircle,
 } from "@fortawesome/free-solid-svg-icons";
 import { toast } from "react-toastify";
+import { Link } from "react-router-dom";
 import NiceModal, { useModal } from "@ebay/nice-modal-react";
 import { ToastMsg } from "../../notifications/Notifications";
 import { PLAYLIST_URI_PREFIX, listenToJSPFTrack } from "../../playlists/utils";
@@ -100,7 +101,7 @@ export default NiceModal.create((props: AddToPlaylistProps) => {
               message={
                 <>
                   Successfully added <i>{trackToSend.title}</i> to playlist{" "}
-                  <a href={playlistIdentifier}>{playlistName}</a>
+                  <Link to={playlistIdentifier}>{playlistName}</Link>
                 </>
               }
             />,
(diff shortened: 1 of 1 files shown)


frontend/js/src/common/listens/AddToPlaylist.tsx
```diff
@@ -100,7 +102,7 @@ export default NiceModal.create((props: AddToPlaylistProps) => {
               message={
                 <>
                   Successfully added <i>{trackToSend.title}</i> to playlist{" "}
-                  <a href={playlistIdentifier}>{playlistName}</a>
+                  <Link to={playlistUrl.pathname}>{playlistName}</Link>
```
MonkeyDo, line 105: I think you should be able to set the identifier directly as it is already a fully qualified URL:
```suggestion
                  <Link to={playlistIdentifier}>{playlistName}</Link>
```
Of course that means we won't need the new URL at line 92.
I will test after the change.

MonkeyDo (changes requested):
Thanks for opening another PR!

frontend/js/src/common/listens/AddToPlaylist.tsx
rrrnld: Oh ok! I actually introduced that in the second commit because I wasn't sure how `react-router` deals with full URLs and couldn't find anything in the docs. I pushed the changes you suggested, thanks!

frontend/js/src/common/listens/AddToPlaylist.tsx
```diff
@@ -100,7 +101,7 @@ export default NiceModal.create((props: AddToPlaylistProps) => {
               message={
                 <>
                   Successfully added <i>{trackToSend.title}</i> to playlist{" "}
-                  <a href={playlistIdentifier}>{playlistName}</a>
+                  <Link to={playlistIdentifier}>{playlistName}</Link>
```
MonkeyDo, line 104: Sorry, I was wrong!
My suggestion would work fine in production, but not if you are on the beta or test websites or developing locally on your machine.
Instead, this is what we want:
```suggestion
                  <Link to={`/playlist/${playlistId}`}>{playlistName}</Link>
```

MonkeyDo (changes requested):
My bad, I was wrong, sorry for the waste of time!

MonkeyDo (approved):
Thank you very much !

PR #3123, opened 2025-01-10
LB-1718: fix top discoveries playlist

amCap1712:
To find tracks discovered in a year correctly, need to use all listens.

### DIFF
--- a/listenbrainz_spark/year_in_music/top_discoveries.py
+++ b/listenbrainz_spark/year_in_music/top_discoveries.py
@@ -2,6 +2,7 @@ from datetime import datetime, date, time
 
 from more_itertools import chunked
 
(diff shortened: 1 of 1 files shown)


mwiencek:
I wondered if a similar issue caused https://community.metabrainz.org/t/year-in-music-2024/735268/6 but you do use `LAST_FM_FOUNDING_YEAR` in `get_new_artists_discovered_count` at least. :shrug: (My YIM also says I have 91% new artists discovered which is definitely wrong.)

amCap1712:
Ah my bad! :facepalm:. I used a partial dump to generate YIM because the full dump takes over a day to generate and import but I forgot there are things that need the full dump.

mayhem (approved):
Oh good, I had some questionable items in that playlist!

PR #3125, opened 2025-01-13
LB-1528: Added shortcut to today's date in Fresh Releases using a Calendar Icon

granth23:
# Problem
[LB-1528](https://tickets.metabrainz.org/browse/LB-1528): Add shortcut to today's date in Fresh Releases

# Solution
Added a calendar icon in the datewise scroll bar as a ticker to directly refer to today's date.

image

# Action

### DIFF
--- a/frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
+++ b/frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
@@ -3,7 +3,7 @@ import Slider from "rc-slider";
 import { countBy, debounce, zipObject } from "lodash";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faCalendarCheck } from "@fortawesome/free-solid-svg-icons";
-import { isToday } from "date-fns";
+import { startOfDay, format, closestTo, parseISO } from "date-fns";
 import { formatReleaseDate, useMediaQuery } from "../utils";
 import { SortDirection } from "../FreshReleases";
 import { COLOR_LB_BLUE } from "../../../utils/constants";
@@ -28,18 +28,28 @@ function createMarks(
       releases,
       (item: FreshReleaseItem) => item.release_date
     );
+
+    const recentDateStr = format(startOfDay(new Date()), "yyyy-MM-dd");
+    const dates = Object.keys(releasesPerDate).map((date) => parseISO(date));
+    const closestDateStr = dates.length
+      ? format(closestTo(new Date(recentDateStr), dates)!, "yyyy-MM-dd")
+      : recentDateStr;
+
     const filteredDates = Object.keys(releasesPerDate).filter((date) =>
-      isToday(new Date(date))
+      date === closestDateStr
         ? releasesPerDate[date] >= 0
         : releasesPerDate[date] > minReleasesThreshold
     );
 
+    const title = closestDateStr === recentDateStr ? "Today" : "Nearest Date";
+
     dataArr = filteredDates.map((date) =>
-      isToday(new Date(date)) ? (
+      date === closestDateStr ? (
         <FontAwesomeIcon
           icon={faCalendarCheck}
           size="2xl"
           color={COLOR_LB_BLUE}
+          title={title}
         />
       ) : (
         formatReleaseDate(date)
@@ -119,7 +129,9 @@ export default function ReleaseTimeline(props: ReleaseTimelineProps) {
   const { releases, order, direction } = props;
 
   const [currentValue, setCurrentValue] = React.useState<number | number[]>();
-  const [marks, setMarks] = React.useState<{ [key: number]: React.ReactNode }>({});
+  const [marks, setMarks] = React.useState<{ [key: number]: React.ReactNode }>(
+    {}
+  );
 
   const screenMd = useMediaQuery("(max-width: 992px)"); // @screen-md
 
--- a/frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
+++ b/frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
@@ -5,6 +5,7 @@ import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faCalendarCheck } from "@fortawesome/free-solid-svg-icons";
 import { formatReleaseDate, useMediaQuery } from "../utils";
 import { SortDirection } from "../FreshReleases";
+import { COLOR_LB_BLUE } from "../../../utils/constants";
 
 type ReleaseTimelineProps = {
   releases: Array<FreshReleaseItem>;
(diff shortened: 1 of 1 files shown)


frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -24,11 +26,26 @@ function createMarks(
...
+        : releasesPerDate[date] > 0
     );
 
-    dataArr = filteredDates.map((item) => formatReleaseDate(item));
+    dataArr = filteredDates.map((item) =>
+      formatReleaseDate(item) === formattedTodayString ? (
+        <FontAwesomeIcon
+          icon={faCalendarCheck}
+          size="2xl"
+          style={{ color: "#353070" }}
```
MonkeyDo, line 43: We have a constant `COLOR_LB_BLUE` , which you can import from frontend/js/src/utils/constants.ts
Also you can directly use the `color` prop instead of style:
`color={COLOR_LB_BLUE}`

MonkeyDo (changes requested):
Looking good, thanks for working on this it will improve the page straight away!

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
granth23: Thanks for the suggestion

granth23:
Made the necessary changes, thanks for the suggestion

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -24,11 +27,26 @@ function createMarks(
...
+      formatReleaseDate(item) === formattedTodayString ? (
+        <FontAwesomeIcon
+          icon={faCalendarCheck}
+          size="2xl"
+          color={COLOR_LB_BLUE}
+        />
+      ) : (
+        formatReleaseDate(item)
+      )
+    );
```
MonkeyDo, line 49: I had another read through the code and I figured we can simplify this quite a lot thanks to a utility function from the date-fns library that we use in LB:
https://date-fns.org/v4.1.0/docs/isToday

From testing in the console on their website, it should be as simple as doing:

image
and removing today, todayString and formattedTodayString.

Of course you'll need to `import { isToday } from "date-fns";`

```ts
    const filteredDates = Object.keys(releasesPerDate).filter((date) =>
      !isToday(new Date(date))
        ? releasesPerDate[date] >= minReleasesThreshold
        : releasesPerDate[date] > 0
    );

    dataArr = filteredDates.map((item) =>
      isToday(new Date(item)) ? (
        <FontAwesomeIcon
          icon={faCalendarCheck}
          size="2xl"
          color={COLOR_LB_BLUE}
        />
      ) : (
        formatReleaseDate(item)
      )
    );
```

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -15,7 +18,7 @@ function createMarks(
   sortDirection: string,
   order: string
 ) {
-  let dataArr: Array<string> = [];
+  let dataArr: Array<React.ReactNode> = [];
```
MonkeyDo, line 21: This might be more appropriate:
```suggestion
  let dataArr: Array<string | JSX.Element> = [];
```

MonkeyDo (changes requested):
Please take the time to check that everything is working as expected, it feels like you rushed that last one through a bit  :)

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -28,16 +29,14 @@ function createMarks(
       (item: FreshReleaseItem) => item.release_date
     );
     const today = new Date();
-    const todayString = today.toISOString().split("T")[0];
-    const formattedTodayString = formatReleaseDate(todayString);
     const filteredDates = Object.keys(releasesPerDate).filter((date) =>
-      date !== todayString
+      isToday(today)
```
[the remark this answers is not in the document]
granth23: Will make sure to recheck once from next time.

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -24,11 +28,23 @@ function createMarks(
       releases,
       (item: FreshReleaseItem) => item.release_date
     );
-    const filteredDates = Object.keys(releasesPerDate).filter(
-      (date) => releasesPerDate[date] >= minReleasesThreshold
+    const filteredDates = Object.keys(releasesPerDate).filter((date) =>
+      isToday(new Date(date))
```
MonkeyDo, line 32: I didn't originally understand why you were doing this filtering here, but now that I've deployed the PR to test.listenbrainz.org, I understand it was to hide the today icon in case there are no albums released today.

I think it is best to always show the calendar icon to have a consistent UI, and have it take you to today's date (or whichever one is closest to it if no albums for the day), even if there are no albums released today it is still useful for users to locate themselves in the timeline.

So I'm guessing this check on line 32-33 should probably just return `true` if isToday is true.

But there is one more "issue" which is that if today's date is not part of the list (i.e. no album released today), today's date needs to be added to the list so that we can show the calendar icon.
For today on my page, for example, there is no 2025-01-17 and it would need to be added:
image

Something like this: (I used date-fns format and startOfDay utilities which I find more readable, but your previous method works too) 
```ts
const today = format(startOfToday(),"yyyy-MM-dd")
if (!(today not in releasesPerDate)) {
  releasesPerDate[today] = 0
}
dataArr = ....
```

granth23: Adding `today` if no releases is a little messy with the pre-existing code for percentage calculation since it defaults to a 0 and it creates an overlap and other such complications, thus I added in the feature of the calendar icon being the date nearest to `today` in case `today` has no releases, with a `title` prop saying if the calendar icon points to `Today` or the `Nearest Date`

MonkeyDo:
PR looks good, thanks for the work!

I am identifying some other issues related to the timeline bar, unrelated to your PR:
image

Because we have a minimum limit on how many releases a day must have in order to appear on the timeline bar, this results in big gaps and some incorrect placement on the timeline:
image
Note the the topmost date on the timeline is Jan.10 (last date with  min. 2 releases) while the top of the page is actually Apr. 4th
That throws off the calculation of the placement on the timeline.
I'll investigate that issue.

granth23:
Thanks! Will look into this as well

MonkeyDo:
> Thanks! Will look into this as well

@granth23 Thanks, but I took the opportunity to implement the changes directly, because I was laying out how I think it should work, and through testing different methods ended up with functional code.

PR #3126, opened 2025-01-14
LB-1724: Improve 'link listens' / 'link with musicbrainz' text

Aerozol:
Clarifying when and how 'Link listens' page data gets updated.
Tightened up the text and text hierarchy while in there, to try avoid the text amount piling up.

Specifically:
- Added Link listens to the Data Update Intervals document table
- Changed the page text to have a succinct header/explanatory paragraph (without any how-to text)
- Moved some how-to text to the small print
- Added when the page updates (Monday) to the small print

### DIFF
--- a/docs/general/data-update-intervals.rst
+++ b/docs/general/data-update-intervals.rst
@@ -13,6 +13,7 @@ Removing deleted listens from stats			    On the 2nd and 16th of each month
 Full dumps						                1st and 15th of each month
 Incremental dumps					            Daily
 Weekly playlists						           Monday morning, based on the users timezone setting
+Link listens						           Monday morning, based on the users timezone setting
 Daily playlists [#f3]_						          Every morning, based on the users timezone setting
 =============================================== =========================================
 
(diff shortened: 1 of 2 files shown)


Aerozol:
Thanks in advance for reviewing! Hopefully this should be pretty straightforward, but I'm sure it will need some formatting magic. Also needs a check in a test environment to make sure I didn't derp up any code.

The other thing that I'm not sure about: @amCap1712 mentioned that the page updates/the cron runs on Mondays. I've put in "Monday morning, based on the users timezone setting" because that's what the other Monday cron has, but I don't actually know when. If that's wrong please correct it, or let me know and I'll change it.

MonkeyDo (approved):
I did a bit of (manual) formatting in the .rst docs file because it was bothering me :D

More importantly, I corrected the time at which it runs, being 2AM UTC rather than depending on user-defined timezone.

Also took a moment to format the date a bit better in the Link Listens page -> 'Jan 13, 2:29 PM' instead of '1/13/2025'  with no time.

Aerozol:
Thank you @MonkeyDo <3

PR #3128, opened 2025-01-16
Add back button and secondary navbar to playlist pages

MonkeyDo:
This was originally done in #2598 when we were still rendering everything in flask.
However it was lost in the move to single-page app architecture and front-end routing.
We did add the back button on artist and album pages (in a0cd3f5), but I am now making this layout component generic to be used for other such pages.

image


anshg1214 (approved):
LG!

PR #3130, opened 2025-01-17
Don't show edit options for other users' pins

MonkeyDo:
I wrote this fix months ago but had apparently forgotten to open a PR...

Currently you can "edit" another users' pinned tracks, or at least have the impression to.
The API rejects the modification, but the front-end acts like you can indeed edit their pins.
Adding a couple of checks to the card and edit modal so that users don't get the impression they can edit another's pins.


frontend/js/src/pins/PinRecordingModal.tsx
```diff
@@ -182,6 +198,7 @@ export default NiceModal.create(
               </p>
               <div className="form-group">
                 <textarea
+                  disabled={!isCurrentUser}
```
anshg1214, line 201: I believe we don't require this checker here. If I go to any user's dashboard and try pinning the listen, the condition will be false here, and we'll not be able to pin the recording.

anshg1214 (changes requested):

frontend/js/src/pins/PinRecordingModal.tsx
MonkeyDo: I think in theory a user shouldn't have an edit button to click, but I prefer to be overly cautious for future breakage, especially if it does not cost any more resources.
In this case we already calculate isCurrentUser so why not use it?
It's a "belt and braces" approach but I think it does no harm.

anshg1214: We're using PinRecordingModal in two places.
1. Editing Comment
2. Pinning a Track

For the second use case, the check prevents me to pin the track if it's not my listen.

MonkeyDo: Ah, thanks for clarifying, I hadn't caught that. Will look into it.

anshg1214 (approved):
LG!

PR #3131, opened 2025-01-17
Add Cache_Control private header to server responses

MonkeyDo:
We have seen issues where it seems that a user's cookies are sent to the wrong user, making it appear like you are logged into another user's account. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#private_caches as well as  https://stackoverflow.com/a/26813182

I created this branch months ago and promptly forgot about it entirely.
Now I anm cleaning old branches, here we are :)

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -177,6 +177,12 @@ def create_app(debug=None):
     @app.after_request
     def after_request_callbacks(response):
         return inject_x_rate_headers(response)
+    
+    @app.after_request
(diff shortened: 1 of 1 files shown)


mayhem:
Seems sensical, but I dont understand caching well enough. I think we should have @amCap1712 also look at this.

MonkeyDo:
Quick ping to @amCap1712

amCap1712 (approved):
I think we should add this for only responses that require authorization, like that determine the user based on the auth token in the request header but it's fine to do that later.

PR #3133, opened 2025-01-20
Added LastFM save button in setting in accordance to LB-1721.

GauravGupta993:
# Problem

This solves a part of the ticket [LB-1721](https://tickets.metabrainz.org/browse/LB-1721), which aims to create a submit button in the LastFM connect services setting page to improve the UI.

# Solution

The submit button and input fields are disabled in case someone has selected the Disable option. Other than that for "Connect to Last.FM" and "Import loved tracks", the submit button works according to the option selected. 
https://github.com/user-attachments/assets/a2a75f54-cedf-42a7-a26e-9ef7bea26544

### DIFF
--- a/frontend/js/src/settings/music-services/details/MusicServices.tsx
+++ b/frontend/js/src/settings/music-services/details/MusicServices.tsx
@@ -57,6 +57,10 @@ export default function MusicServices() {
       : undefined
   );
 
+  const [lastFMSubmit, setlastFMSubmit] = React.useState(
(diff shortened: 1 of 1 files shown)


MonkeyDo:
For reference, you can alway modify your commits locally and force-push to your own branch if you wan to make modifications, instead of closing the PR and opening a new one.
That is a fairly standard workflow, and force-pushing is only an issue when someone else is also working on the same branch.
It's what I do when I commited something I shouldn't have, or realise  a second after pushing a commit that I forgot sonmething.

PR #3134, opened 2025-01-20
LB1721: Added LastFM save button in connect services setting.

GauravGupta993:
# Problem

This solves a part of the ticket [LB-1721](https://tickets.metabrainz.org/browse/LB-1721), which aims to create a submit button in the LastFM connect services setting page to improve the UI. 

# Solution

The submit button and input fields are disabled if someone has selected the Disable option. Other than that for "Connect to Last.FM" and "Import loved tracks," the submit button works according to the option selected. 

https://github.com/user-attachments/assets/95030a0b-4bc1-494a-ba83-a150b513a871

### DIFF
--- a/frontend/js/src/settings/music-services/details/MusicServices.tsx
+++ b/frontend/js/src/settings/music-services/details/MusicServices.tsx
@@ -57,9 +57,7 @@ export default function MusicServices() {
       : undefined
   );
 
-  const [lastFMSubmit, setlastFMSubmit] = React.useState(
-    permissions.lastfm === "import" ? "LastFMConnect" : "disable"
-  );
+  const [lastFMEdit, setLastFMEdit] = React.useState(false);
 
   const handlePermissionChange = async (
     serviceName: string,
@@ -103,9 +101,6 @@ export default function MusicServices() {
             break;
           case "critiquebrainz":
             if (critiquebrainzAuth) critiquebrainzAuth.access_token = undefined;
-            break;
-          case "lastfm":
-            setlastFMSubmit("disable");
             break;
           default:
             break;
@@ -194,6 +189,20 @@ export default function MusicServices() {
   ) => {
     evt.preventDefault();
     try {
+      if (!lastfmUserId) {
+        setLastfmUserId(loaderData.current_lastfm_settings?.external_user_id);
+        setLastfmLatestListenedAt(
+          loaderData.current_lastfm_settings?.latest_listened_at
+            ? format(
+                new Date(
+                  loaderData.current_lastfm_settings?.latest_listened_at
+                ),
+                "yyyy-MM-dd'T'HH:mm:ss"
+              )
+            : undefined
+        );
+        throw Error("LastFM username empty");
+      }
       const response = await fetch(`/settings/music-services/lastfm/connect/`, {
         method: "POST",
         body: JSON.stringify({
@@ -458,12 +467,7 @@ export default function MusicServices() {
                     onChange={(e) => {
                       setLastfmUserId(e.target.value);
                     }}
-                    disabled={
-                      !(
-                        lastFMSubmit === "LastFMConnect" ||
-                        lastFMSubmit === "LastFMLovedTrack"
-                      )
-                    }
+                    readOnly={!lastFMEdit && permissions.lastfm === "import"}
                   />
                 </div>
                 <div>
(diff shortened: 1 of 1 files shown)


MonkeyDo (commented):
Thanks for opening the PR, it's going to be easier to follow for feedback !

I tested the new mechanism and I think it needs some changes.
I don't like the fact that clicking on "connect to lastfm" does nothing by itself when you click it, making it different from the other music services above/below.


My observations after testing:
- Clicking on "connect to LFM" or "import loved tracks" should perform the action on click (rather than the smaller submit button).
- This means we can't have disabled input fields initially because the user needs to input their username (my bad on that one, I didn't think it through)


Here is my updated proposal:
- on page load, if user is not connected, input fields are empty and enabled, "edit" button is disabled and greyed out (`btn-default` class):
image
- After successfully connecting to LFM, the name and import time inputs are set as `readonly`, and the "edit" button becomes enabled (`btn-warning` class):
image
- clicking on the edit button, inputs become enabled and the button text changes color to green (`btn-success` class) and text changes to "save":
image
- after clicking on the "save" button, inputs become `readonly` again and the action to connect to LFM is performed again automatically to connect with the new name
- We can't use native form validation if the inputs are readonly or disabled, which means we need to add a check for empty username and show a warning toast message if it's empty.


How does that sound to you?

GauravGupta993:
> Thanks for opening the PR, it's going to be easier to follow for feedback !
> 
> I tested the new mechanism and I think it needs some changes. I don't like the fact that clicking on "connect to lastfm" does nothing by itself when you click it, making it different from the other music services above/below.
> 
> My observations after testing:
> 
> * Clicking on "connect to LFM" or "import loved tracks" should perform the action on click (rather than the smaller submit button).
> * This means we can't have disabled input fields initially because the user needs to input their username (my bad on that one, I didn't think it through)
> 
> Here is my updated proposal:
> 
> * on page load, if user is not connected, input fields are empty and enabled, "edit" button is disabled and greyed out (`btn-default` class):
>   image
> * After successfully connecting to LFM, the name and import time inputs are set as `readonly`, and the "edit" button becomes enabled (`btn-warning` class):
>   image
> * clicking on the edit button, inputs become enabled and the button text changes color to green (`btn-success` class) and text changes to "save":
>   image
> * after clicking on the "save" button, inputs become `readonly` again and the action to connect to LFM is performed again automatically to connect with the new name
> * We can't use native form validation if the inputs are readonly or disabled, which means we need to add a check for empty username and show a warning toast message if it's empty.
> 
> How does that sound to you?

Sounds great. Thanks for the review. I will do it this way.

MonkeyDo:
Hello, please let me know if there anything I can help you with regarding his PR, or if you have any questions.

GauravGupta993:
> Hello, please let me know if there anything I can help you with regarding his PR, or if you have any questions.

I am really sorry for the delay. My college exams started, and I got busy there. They just ended this week. I will complete this work within 2 days. Wont delay it much further.

MonkeyDo:
No worries, I totally understand. Don't feel rushed, your exams and college life are more important.
Just wanted an update to know if you were blocked by something.

MonkeyDo:
Hello! I'm coming back to ask if you still wan tot work on this PR, or if you prefer I can work on it to push it past the finish line. Please let me know either way :)

GauravGupta993:
@MonkeyDo Once again sorry for the delay. Also, thanks a lot for your reminders and patience. I have made changes as per your instructions. Please review in case some further changes are required. 

https://github.com/user-attachments/assets/9ac2f1c1-3c75-4806-92fd-604e5a9ed781

Also, I realize I dragged this small task for two months, and I'll ensure it doesn’t happen again.

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -469,8 +486,23 @@ export default function MusicServices() {
...
                 </div>
+                <div style={{ flex: 0, alignSelf: "end" }}>
+                  <button
+                    disabled={permissions.lastfm !== "import"}
+                    type={lastFMEdit ? "button" : "submit"}
+                    className={
+                      permissions.lastfm !== "import"
+                        ? "btn-default"
+                        : (lastFMEdit && "btn-success") || "btn-warning"
+                    }
```
MonkeyDo, line 500: Bootstrap expects that buttons should have the `btn` class as well as the styling classes (btn-success, btn-warning, etc.) for proper styling.
Since this line will become a bit messy, I propose to refactor this ternary operator into a separate constant `lastFMEditButtonClass` declared above, then use it like soe:

```suggestion
                    className={`btn ${lastFMEditButtonClass}`}
```

MonkeyDo (changes requested):
Thank you for working on this some more!
I have one final change request for the button styling, after which I think we can merge the PR.
I think it is definitely better than the current setup, and with the other PR #3226 it improves the situation quite a bit, so a green flag for me :green_circle:

frontend/js/src/settings/music-services/details/MusicServices.tsx
GauravGupta993: Ok, thanks for the review. Will make the changes.

MonkeyDo: There was a misunderstanding with the solution I proposed.

I was not suggesting to use react state for this lastFMEditButtonClass, but rather a constant value (`const`) since it is simply derived from the state of permissions.lastfm and lastFMEdit.

Basically it would be equivalent to what you had before (ternary operator in className) except it is refactored so the value is defined higher up in the code to make it easier to read.

GauravGupta993: I misread your text. Thanks for the clarification. I'll fix it immediately.

GauravGupta993: @MonkeyDo, I think the notification didn't go through since I just commented. Please review the changes whenever you are free. No hurries.

MonkeyDo (approved):
Looks good, thank you!

I think this is an improvement over the existing workflow, but there are still some points of confusions I think.
This feature will need a more thorough review in terms of user experience, and a series of mockups to support the next steps.

PR #3135, opened 2025-01-20
Update data-update-intervals.rst

Aerozol:
Added "similarity dataset" update time (Sundays)


MonkeyDo (commented):
A suggestion for clarification.

Thanks for keeping this doc up to date !

PR #3138, opened 2025-01-21
FreshReleases: show more intermediary dates

MonkeyDo:
Hot on the heels of a comment I made while testing #3125 , realizing our timeline looked very empty for me.

We have a system to decide which dates to show on the timeline that depends on the number of releases for each day.
If there is a long stretch of time without the required minimum releases, the timeline is suddenly very empty, less useful and less clear.
image

The modifications in this PR ensure a date is shown in the timeline at minimum every week.

Also small refactor while I'm in there.


MonkeyDo:
This didn't quite resolve the general issues with the timeline and ended up with more jumbled up dates.
Follow discussion in #3142 for some more context.

PR #3139, opened 2025-01-22
Fixed global feed page crashing

fettuccinae:
# Problem

    Global feed kept crashing because donor's database is not created in local environment.

# Solution

   I added a condition which would return an empty donors list if the environment is set to development, this way it wont crash by calling recent_donors function. 

# Action

    We can create a test donor's db so that index renders that donor panel.


MonkeyDo (commented):
Apart from the linting issues (see https://github.com/metabrainz/listenbrainz-server/pull/3139#issuecomment-2606108004), this PR is ready to go, thanks for fixing it!

MonkeyDo (approved):

PR #3140, opened 2025-01-23
Improved Nginx Configuration to Address Critical Issues in Production Environments

Advaitgaur004:
# Problem

The current Nginx configuration lacks essential security features and HTTPS support which are crucial for a production environment. Running a web service without HTTPS in production (😢):
1. Exposes user data to potential man-in-the-middle attacks
2. Reduces trust in the service
3. May affect SEO rankings ( It Affects, I have the experience )

# Solution

Enhanced the Nginx configuration to include:

1. SSL/HTTPS Support:
   - Added HTTPS listener on port 443 (that is default btw)
   - Configured SSL certificate paths
   - Enabled modern TLS protocols (1.2 and 1.3) (Again standard implementation)

2. Security Improvements:
   - Added HTTP to HTTPS redirect
   - Basic security headers
   - Disabled server tokens to prevent version disclosure

The changes are minimal and focused on essential security features while maintaining the original proxy functionality.

Technical Details:
- SSL certificates should be placed in `/etc/nginx/ssl/`
- Existing wsgi proxy settings remain unchanged
- Added HTTP to HTTPS redirect on port 80

# Action

Required actions:
1. Review the SSL configuration
2. Ensure SSL certificates are available at the specified paths
3. Test HTTPS functionality after deployment
4. Verify that the HTTP to HTTPS redirect works as expected ( if any problem occur please let me know)

# Note
I haven't configured the timeout for uwsgi. This is for future enhancements, allowing juniors to solve this issue later. It won't take much time.

### DIFF
--- a/docker/prod/nginx/nginx.conf
+++ b/docker/prod/nginx/nginx.conf
@@ -1,12 +1,58 @@
-server {
-    listen 80;
-    server_name proxy.listenbrainz.org;
-    root /usr/share/nginx/html;
-    location / {
-        try_files $uri @wsgi;
+worker_processes auto;
+worker_rlimit_nofile 65535;
+
+events {
+    worker_connections 65535;
+    multi_accept on;
+}
+
+http {
+    charset utf-8;
+    sendfile on;
+    tcp_nopush on;
+    tcp_nodelay on;
+    server_tokens off;
+    log_not_found off;
+    client_max_body_size 16M;
+    
+    # Logging
+    access_log /var/log/nginx/access.log combined buffer=512k flush=1m;
+    error_log /var/log/nginx/error.log warn;
+
+    # Gzip Settings
+    gzip on;
+    gzip_vary on;
+    gzip_proxied any;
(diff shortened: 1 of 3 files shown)


Advaitgaur004:
@MonkeyDo Take a look; if you allow, I can make further modifications for security reasons.

mayhem:
Thanks for this PR, but the configuration that you've changed lives behind our gateways, so the changes you introduced are entirely redundant because this configuration is never intended to be exposed to the outside word. It should always be behind a reverse proxy that addresses the issues in your PR.

Advaitgaur004:
> Thanks for this PR, but the configuration that you've changed lives behind our gateways, so the changes you introduced are entirely redundant because this configuration is never intended to be exposed to the outside word. It should always be behind a reverse proxy that addresses the issues in your PR.

I'm glad to hear this, as the current configuration makes sense if a reverse proxy hasn't been used. Thank you for the information.

PR #3141, opened 2025-01-23
Fix doc build error

samglt:
# Problem

Docs have been [failing to build](https://app.readthedocs.org/projects/listenbrainz/builds/26799324/) after [this commit](https://github.com/metabrainz/listenbrainz-server/commit/6a86710c2c06c75d5923f3cd55369a30371c3305) since the added converter wasn't registered in the special initializer for RTD


MonkeyDo (approved):
Thank you for consistently fixing the docs we keep breaking !

PR #3142, opened 2025-01-23
Improved marks mapping

granth23:
# Problem

image

Sidebar with Date marks didn't map to the date accurately.

# Solution

image

Modified the method of percentage calculation to accommodate a better mapping to the date searched for by the user.

# Action
1. Please review the changes to ensure they align with the project's style guide.
2. Verify that the updated mapping improvements work as intended on the live site.

### DIFF
--- a/frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
+++ b/frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
@@ -62,10 +62,7 @@ function createMarks(
     let cumulativeSum = 0;
     percentArr = filteredDates.map(
       (item) =>
-        (((cumulativeSum += releasesPerDate[item] || 0) -
-          releasesPerDate[item]) *
-          100) /
-        totalSum
+        (((cumulativeSum += releasesPerDate[item] || 0) - releasesPerDate[item]) * 100) / totalSum
     );
   } else if (order === "artist_credit_name") {
     const artistInitialsCount = countBy(releases, (item: FreshReleaseItem) =>
@@ -136,9 +133,7 @@ export default function ReleaseTimeline(props: ReleaseTimelineProps) {
   const { releases, order, direction } = props;
 
   const [currentValue, setCurrentValue] = React.useState<number | number[]>();
-  const [marks, setMarks] = React.useState<{ [key: number]: React.ReactNode }>(
-    {}
-  );
+  const [marks, setMarks] = React.useState<{ [key: number]: React.ReactNode }>({});
 
   const screenMd = useMediaQuery("(max-width: 992px)"); // @screen-md
 
--- a/frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
+++ b/frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
@@ -55,11 +55,18 @@ function createMarks(
         formatReleaseDate(date)
       )
     );
-    percentArr = filteredDates
-      .map((item) => (releasesPerDate[item] / releases.length) * 100)
-      .map((_, index, arr) =>
-        arr.slice(0, index + 1).reduce((prev, curr) => prev + curr)
-      );
+    const totalSum = Object.values(releasesPerDate).reduce(
+      (acc, value) => acc + value,
+      0
+    );
+    let cumulativeSum = 0;
+    percentArr = filteredDates.map(
+      (item) =>
+        (((cumulativeSum += releasesPerDate[item] || 0) -
+          releasesPerDate[item]) *
+          100) /
+        totalSum
+    );
   } else if (order === "artist_credit_name") {
     const artistInitialsCount = countBy(releases, (item: FreshReleaseItem) =>
       item.artist_credit_name.charAt(0).toUpperCase()
@@ -129,7 +136,9 @@ export default function ReleaseTimeline(props: ReleaseTimelineProps) {
   const { releases, order, direction } = props;
 
   const [currentValue, setCurrentValue] = React.useState<number | number[]>();
-  const [marks, setMarks] = React.useState<{ [key: number]: React.ReactNode }>({});
+  const [marks, setMarks] = React.useState<{ [key: number]: React.ReactNode }>(
+    {}
+  );
 
   const screenMd = useMediaQuery("(max-width: 992px)"); // @screen-md
 


granth23:
This is for the `/explore/fresh-releases/` page.

MonkeyDo:
Hehe, great minds think alike !

I've done the same change in #3138 

However that PR also puts in a date whenever a week has passed, I thought it might be useful to avoid having huge gaps in the timeline, but I think it might make it worse in the end. I'll check your solution and see which one works best.

MonkeyDo:
So this is pretty similar results to what I tried myself, and while it improves the timeline I think there are still issues that I don't know how to solve.
Primarily, clicking on a date in the timeline never gets me to the actual date...
I think part of the issue is that we are trying to make our data fit a library that isn't really meant for this (rc-slider).

I think the solution would instead be to roll out our own component where we can control things a bit better.
In particular, we can turn the date separators on the page into headings (`<h4 id={myDateStr}>` or something like that), and make the dates on the timeline anchor tags (```<a href={`#${myDateStr}`}/>```) and let the browser deal with scrolling you accurately to the desired spot on the page. I tested it manually and it works way nicer, basically for free.

That leaves us with creating the component and styles, and reacting to page scroll/click on the timeline
Those last two points are fairly easy, and I tested that too (scroll event listener + calculating of container height and timeline height + rescaling one to the other).
Considering we already calculate percentages for positioning, overall I think it is pretty achievable.

What do you think about that plan?
And would you be interested in working on this? If so I'll share all my trials in more details.

granth23:
Originally I was thinking with the idea of anchor tags but didn't feel like changing the entire thing but surely I can work on it, do share your trials will get on to it right away.

granth23:
One of the problems with anchor tags is that once the user redirects himself, the scroll will automatically move here and there from the date due to the percentage implementation, any idea on how to resolve that?

MonkeyDo:
> One of the problems with anchor tags is that once the user redirects himself, the scroll will automatically move here and there from the date due to the percentage implementation, any idea on how to resolve that?

I don't think I understand what you mean by "redirects himself".

Do you mean that there will be a mismatch between the vertical placement of the dates in the timeline and the actual position of the "cursor" (the position indicator) on the timeline?
I think that's a possibility. I suppose we could find a hacky way to get the actual Y position on the page for each date-separator, but what I have in mind won't be very React-y.

granth23:
Yes that's what I mean that the placement won't be the same, do you think we could ignore the threshold part and just make it equally distributed than total releases focused, the only problem with that will be the concept behind which date has more releases.

MonkeyDo:
I think you can go ahead without worrying too much about the placement of dates on the timeline.
We can try equal distribution, or we can try later to implement some other mechanism.
Either way, it won't be very different from the current broken placement, so we're not regressing in terms of functionality.

MonkeyDo:
Hello @granth23, let me know if you have any questions or doubts, or if there is anything I can help you with regarding this PR.

granth23:
Hi, sorry for the delay, was a little occupied somewhere. I have worked on the logic and modified it a lot:
1. Insert the first date.
2. Go on adding unique dates which have a cumulative release of more than 10% from the current date.
3. Make sure the last date is included
4. Make sure to include today or the date nearest to today if not already.

Hoping this looks better and based on the tests I did, it is mapping to very much the correct position as well!

image

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -21,45 +21,98 @@ function createMarks(
 ) {
   let dataArr: Array<string | JSX.Element> = [];
   let percentArr: Array<number> = [];
-  // We want to filter out the keys that have less than 1.5% of the total releases
```
MonkeyDo, line 24: You removed the comment here, but the value `minReleasesThreshold` is still in use in the text-based sorting modes.
You could modify the comment to explain where it is used now.

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
MonkeyDo, line 1: Let's make the component more type-safe while we are in here:

Where we defined the expected arguments we can use better types we already have instead of "string" which provides no protection  or code-completion (potential for uncaught typos, like `if (sortDirection === "decend")` with a missing letter).

Line 13:
`order: SortOption;`
Lines 19-20:
```
sortDirection: SortDirection,
order: SortOption
```

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -100,26 +173,16 @@ function createMarks(
...
 
-  /**
-   * We want the timeline dates or marks to start where the grid starts.
-   * So the 0% should always have the first date. Therefore we use unshift(0) here.
-   * With the same logic, we don't want the last date to be at 100% because
-   * that will mean we're at the bottom of the grid.
-   * The last date should start before 100%. That explains the pop().
-   * For descending sort, the reverse computation above possibly already ensures that
-   * the percentArr starts with 0 and ends with a non-100% value, which is desired.
-   * Hence, we add a check to skip the unshift(0) and pop() operations in that case.
```
MonkeyDo, line 118: I think this comment was useful to understand  the last block of code with the unshift and pop.
Did you remove it by accident?

I'm not opposed to rephrasing the comment if you want, but it was there for a reason.

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
```diff
@@ -88,6 +151,16 @@ function createMarks(
...
+
+    if (sortDirection === "descend") {
+      dataArr.reverse();
+      percentArr = percentArr.reverse().map((v) => (v <= 100 ? 100 - v : 0));
+    }
+
+    if (percentArr[0] !== 0) {
+      percentArr.unshift(0);
+      percentArr.pop();
+    }
```
MonkeyDo, line 163: This block of code is now repeated multiple times, which is not great.

Since it applies in all cases apart from order === "release_date", you can simply apply it conditionally at the end of the function, after all the other blocks.

```
if (order !== "release_date") {
    if (sortDirection === "descend") {
      dataArr.reverse();
      percentArr = percentArr.reverse().map((v) => (v <= 100 ? 100 - v : 0));
    }

    if (percentArr[0] !== 0) {
      percentArr.unshift(0);
      percentArr.pop();
    }
  }
```

MonkeyDo (changes requested):
A few minor code cleanup comments below, but apart from that I think this is already a very good improvement of the timeline ! 
Nice work.

frontend/js/src/explore/fresh-releases/components/ReleaseTimeline.tsx
granth23: I have rephrased the comment and added it back.

granth23:
Made the necessary changes

MonkeyDo (approved):
Thanks for the speedy changes.
Good to merge! :rocket:

I do wan tot leave a comment here for posterity:
When clicking on one of the timeline dates, the page scrolls down **near** the top of the section for the date in question, but not quite at the top of it.
I think it's going to be hard to calculate what the actual scroll value should be considering the slider component uses percentages.

The replacement option would be to add ids to the headers and turn timeline dates into anchor links to those ids, and let the browser scroll you down to the exact right place.

Probably requires significant rewrite though, so merging this PR as a very good improvement over the current situation :+1:

PR #3143, opened 2025-01-23
LB-1688 Feature to thank a user for a pin/recommendation

Suvid-Singhal:
# Run the [SQL migration script](https://github.com/metabrainz/listenbrainz-server/pull/3143/files#diff-e9d05c8922ce6343cb56ab1138807c74f05b8263c77c3656218a6e106fa00d06) when releasing this PR 

# Problem

This feature aims to add a functionality to thank users for their pins/recommendations
This feature is a work in progress as of now
Backend is mostly complete but frontend needs some work.

The feature request ticket is: [LB-1688](https://tickets.metabrainz.org/browse/LB-1688)

# Solution

Implemented backend as of now. Didn't have to change the DB schema for it.

# Action

I need some help in some places for frontend in order to complete this.


MonkeyDo (commented):
Thank you for opening the PR!

There are a couple of things missing before I can test the feature:

Currently, I am recieveing an error on the back-end and two on the front-end.
The back-end one is crashing the web container preventing my testing.

Back-end error:
>File "/code/listenbrainz/listenbrainz/webserver/views/user_timeline_event_api.py", line 731
web-1             |     raise APIBadRequest(f"{metadata['event_type']} event with id {
web-1             | SyntaxError: unterminated string literal (detected at line 731)

Front-end errors:

First one is easy and just needs renaming the `thankFeedEvent` method to avoid name collision:
>/code/frontend/js/src/user-feed/UserFeed.tsx
static_builder-1  |   270:15  error  'thankFeedEvent' is already declared in the upper scope on line 268 column 9

Second issue is a  typescript type error:
> ERROR in ./frontend/js/src/user-feed/UserFeed.tsx:277:27
static_builder-1  | Found 1 error in 13396 ms.
static_builder-1  | TS2339: Property 'blurb_content' does not exist on type 'EventMetadata'.
static_builder-1  |   Property 'blurb_content' does not exist on type 'Listen'.

I think at the core of this issue is a misunderstanding of where the thanking logic goes.
The way it is currently set up, clicking on "Thank Event" tries to send a non-existing blurb content via the API (the event you are thanking might not have a blurb content, and if it has one it's not the thank you message).
So instead, clicking on the button should open a modal where the user can write an optional thank you message. The logic for sending the thank-you with the optional blurb content should be implemented in the modal in question instead of the UserFeed page.
A good example of how to do all this is to see how the `PersonalRecommendationModal` is used in the ListenCard component, and basically use that as an example for the action button and the new modal.

Suvid-Singhal:
Hi @MonkeyDo
The backend error was caused by linter 😅
I have fixed it now
Thanks for your detailed inputs for the frontend. Yes, indeed I had misunderstood about where I had to put the logic for thanking logic should have been put. I have now fixed that as well. Made a new file for the ThanksModal.
I think the frontend is fine to test now. Just need some minor cosmetic changes I believe. The backend also seems to be fine now.

The issue I am facing now is with database. I am getting Database exception while testing. I will have to write SQL scripts and run them on my system to test it but I can't figure it out for now. Will try to fix that as well.

Could you please test backend and frontend part and let me know where changes are needed for now?

MonkeyDo:
I think I know where the DB errors are coming from actually: the 'thanks' event type is missing in the `user_timeline_event_type_enum`.
You'll need to modify the create_types.sql file (for new DB creation) as well as creating an sql update file that adds the type in the existing enum (for modifying existing DBs)

Suvid-Singhal:
Hi @MonkeyDo,
So I updated the DB now but still I'm facing 500 internal server error. I'm not able to figure out what might be causing them :(
Could you please help me out?
The error message suggests that it is being caused due to `DatabaseException` only I think.
Thanks

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -611,6 +640,119 @@ def create_personal_recommendation_event(user_name):
...
+        raise APIBadRequest("JSON document must contain metadata", data)
+
+    metadata = data["metadata"]
+
+    if 'original_event_type' not in metadata or 'original_event_id' not in metadata:
+        raise APIBadRequest(
+            "Metadata must contain both original_event_type and original_event_id", metadata)
+
+    try:
+        validate = ThanksMetadata(**metadata)
```
MonkeyDo, line 714: Main issue I had to resolve is that you are not using the class created here when you send it to `db_user_timeline_event.create_thanks_event` which results in some errors.
Overwriting the metadata variable here resolves that issue
```suggestion
        metadata = ThanksMetadata(**metadata)
```

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -611,6 +640,119 @@ def create_personal_recommendation_event(user_name):
...
+        row_id = metadata["original_event_id"]
+        event_type = metadata["original_event_type"]
```
MonkeyDo, line 716: After the change above to use the ThanksMetadata class, you'll need to access properties like so:
```suggestion
        row_id = metadata.original_event_id
        event_type = metadata.original_event_type
```

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -611,6 +640,119 @@ def create_personal_recommendation_event(user_name):
...
+        if event_type in [
+                UserTimelineEventType.RECORDING_RECOMMENDATION.value,
+                UserTimelineEventType.PERSONAL_RECORDING_RECOMMENDATION.value]:
```
MonkeyDo, line 719: Similarly, now armed with the ThanksMetadata class, event_type can be compared directly to the enums:
```suggestion
        if event_type in [
                UserTimelineEventType.RECORDING_RECOMMENDATION,
                UserTimelineEventType.PERSONAL_RECORDING_RECOMMENDATION]:
```

Either that, or use `if event_type.value`

MonkeyDo (commented):
This set of changes should resolve the blocking issues.

For reference, I just put `current_app.logger.info(...` calls at regular intervals to follow where the issue was coming from and isolate the bits of code that were failing, one by one.

While I am here, without doing a full code review of the front-end parts, the UI is looking great. All that is missing is to show the thank events on both users' timelines.

listenbrainz/webserver/views/user_timeline_event_api.py
MonkeyDo: I believe there were a couple of issues resulting from not using the expected class, but you possibly fixed one of them (the UserTimelineEventMetadata union type did not contain the new ThanksMetadata)

If I test now with latest commits and add a logger call to log the error, this is what I get:
```
2025-01-28 15:58:05,802 listenbrainz.webserver ERROR    'dict' object has no attribute 'dict'
```
Thrown by this line: https://github.com/metabrainz/listenbrainz-server/blob/46374f7d66d7d1a21850378cc816a84d53a4e7ed/listenbrainz/db/user_timeline_event.py#L59

Since a `UserTimelineEvent` class is expected in all of these methods, you are likely to run into some issue or other if you pass a dict instead of a class.

Suvid-Singhal:
Hi @MonkeyDo,
It's mostly working now. The event shows up on the feed now but it does not show the user who is thanking and the one being thanked properly. It just shows user thanking only for both. Tried using user_name, currentUser.name and event.name as well. So I tried to include a field in metadata called thanker_username. But after adding it, it does not even seem to fetch any thanks events from the DB.

I'm kinda confused on how to proceed.
Should I event include it in the first place in metadata or there seems to be some other way?
Really sorry for so many problems as it's my first time working on backend and frontend together in such a large codebase. 😓

MonkeyDo:
OK, let's take a step back for a second because it seems you are trying to add stuff to make it work, without a clear plan or knowing what info you are missing where.

Let's think about our issue by analyzing the process step by step.
The scenario is simple: user1 want to thank user2 for a recommendation (event1).
Here's the data we have:
- user1 called the API with their own username: /user/user1/timeline-event/create/thanks so we have the thanker's username
- an event is created in the DB (event2) with the id of user1 (user_id column of the user_timeline_event table), so we have the thanker's user id as well
- the thanks event (event2) metadata contains the original_event_id
- the original event (event1) was created by user2, and in the event1 DB row has the id of user2 in the user_id column
- so far, we have the name and id of user1 from the thanks event, and we have the id of user2 via the original event 
 
The only piece of data we are missing at this point is the name of user2.
There are two options: 1. get user2's name from the original event, using their user_id or 2. store the thankee_username in the thanks event metadata when you create it.

Option 1: To get user2's name, you can retrieve event1 using the original_event_id, using [`get_user_timeline_event_by_id`](https://github.com/metabrainz/listenbrainz-server/blob/da2a980c613dfd68c26eb88c245924a0120f9781/listenbrainz/db/user_timeline_event.py#L359) then get the user name using the user_id of  event1, with the [`get_users_by_id` function](https://github.com/metabrainz/listenbrainz-server/blob/da2a980c613dfd68c26eb88c245924a0120f9781/listenbrainz/db/user.py#L421).

Option 2:  when you create the thanks event, either send the thankee_name from the front-end, or on the back-end use the original_event_id and as described above fetch user2's name using the user_id of the original event, then store that in the thanks event metadata in the database.

I think option 2 will be easier and faster, at least at retrieval time (we don't need to fetch the original event or the username of the original event creator if we already store that in the thanks event metadata)

I think with this limited amount of information, we only want to show a simple UI for now, just the line "user1 thanked user2 for recommendation_type"
This can probably be improved but let's cross that bridge when we come to it.


-------------------------------------------------------

Regarding your second attempt, the front-end does not send any thanker_username as part of the metadata JSON ([here](https://github.com/metabrainz/listenbrainz-server/pull/3143/files#diff-071a195614ed23ed439509e6755c6e9fa7ed216d11c55e28885dcc141ed06205R1354-R1357)), and when it is received by the back-end endpoint, the event metadata uses that JSON as-is ([here](https://github.com/metabrainz/listenbrainz-server/pull/3143/files#diff-68b3baf8382d61b3b29a2894bb9edca6db7a948d6f0794bf28275c8581f859b4R714)), meaning it is created in the DB without the thanker_username.

When you try to retrieve it from the DB, the APIThanksEvent class expects a thanker_username in the metadata and does not find it, and throws a validation error.

MonkeyDo:
One last thing :)

For the UI, it's always easiest to first make a mockup of what you want things to look like, before coding it.
That allows you to iterate quickly over various options, get an idea of whether it satisfies what users expect, and also tells you what data you are going to need in order to achieve it.

I usually make my mockups directly in the browser page, modifying the html and css of the page directly from the browser devtools, then take a screenshot of that.

Suvid-Singhal:
Hi @MonkeyDo,
Thanks a lot for the detailed review. It helped a lot understanding how frontend is communicating with the backend in the codebase. I think its much more clearer to me now.
The backend now works as intended and now it's sending all the 4 pieces of information to the frontend :)

I think now only frontend part is left cuz it looks really bad as of now 😅
Implemented the very basic UI you suggested for now.
Screenshot from 2025-02-01 14-28-07
Also, should this event be listenable?
Please suggest some UI changes :)

MonkeyDo:
> Hi @MonkeyDo, Thanks a lot for the detailed review. It helped a lot understanding how frontend is communicating with the backend in the codebase. I think its much more clearer to me now. The backend now works as intended and now it's sending all the 4 pieces of information to the frontend :)
> I think now only frontend part is left cuz it looks really bad as of now 😅 Implemented the very basic UI you suggested for now. !) Also, should this event be listenable? Please suggest some UI changes :)

I would like you to make a UI mockup as per my last comment.
It is good to learn how to code, but one must also learn to understand how to evaluate the entire feature.

In your case, that means testing the website some more to get a feel for it, and making UI mockups of how you think this feature would best integrate into the feed.

Some things to think about:
- not all thanks will contain a message
- thanks events don't need to be playable (listenable)
- do you want to show the thank event under the original event?
- if not, do you wan to duplicated the original event to show some context?
- if you duplicate the original event, it might make sense to modify its opacity (for example) to give it less importance compared to the thank event

Suvid-Singhal:
Hi @MonkeyDo,
Thanks a lot for the design inspirations first of all. It really made me think about the product and how can I implement the feature in terms of the UI.

> do you want to show the thank event under the original event?

I like this idea as duplicating the event again and again would clutter the user feed. So, this seems to be a better solution.
I tried making this mockup in dev tools:

Screenshot from 2025-02-04 16-36-35

I can be displayed like a sub-event timeline of the original event.
Its a bit rough as of now. Can add a box for the message but I think that will take too much space if there are multiple thankers.

Or we can do it like this as well:

Screenshot from 2025-02-04 16-43-27
Clicking the URL will reveal the contents in like in the above screenshot and clicking it again would hide them again.

What do you think?

MonkeyDo:
> I think we can go ahead with Option B as it seems easier to implement 😛 and also looks good enough. I think it kinda makes more sense as we won't be putting the thanks event twice in the timeline. And yea, I didn't consider the case when someone thanks a really old recommendation/pin. Btw your mockups look so good 🔥 Will try to make better mockups next time :)

Great, thanks for taking the time to look into all this thoroughly. I know it's a lot of moving parts and some things you might not be comfortable with (yet), but it's coming along nicely.

As for the mockups, thanks, I think they always help a lot to visualize not just the look of the UI, but the underlying data and mechanisms required to make it work. They don't necessarily need to be pretty, but it gives you a blueprint to follow once you actually implement it.
There is always a lot of focus on the code in computer science, but when working on websites it's important to take the time to change point of view, and place yourself in the shoes of a user (the UX part).
The UI design, CSS and all that comes after and is IMO not as important to nail as good UX.


> Recommendations and thank you's do look like a really helpful feature ngl, would love to have it on Spotify as well xD Ok, so I think I'll start working on it by tomorrow.

You know what? 10 years ago Spotify had a sidebar with all your friends listed, and you could drag and drop songs directly on their username to share with them.
They got rid of that at some point after striking a deal with Facebook :(|

I've missed that feature ever since... Generally, I think the LB team is trying to recreate the good parts of a few music services from 10 years ago that marked us and helped us discover new music.

listenbrainz/db/user_timeline_event.py
```diff
@@ -152,6 +171,38 @@ def create_personal_recommendation_event(db_conn, user_id: int, metadata: WriteP
     except Exception as e:
         raise DatabaseException(str(e))
 
+'''
```
MonkeyDo, line 174: Please delete all unused commented out code

frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -487,9 +588,93 @@ export default function UserFeedPage() {
...
+  }>({});
+
+  const renderSubEvent = (event_id: number, event_type: EventTypeT) => {
+    const subEvent = events?.find(
+      (event) => event.id === event_id && event.event_type === event_type
+    );
+
+    if (!subEvent) return null;
+
+    subEvent.subevent = true;
```
MonkeyDo, line 615: I'm a bit uncomfortable with this, it's not very React-y to modify the props array like that.
From my understanding it is used to decide whether to render the thanks icon?

I think a better way would be to make `renderEventActionButton` accept an optional second parameter and take decisions based on that.
For example `isSubEvent` (defaults to `false`):

`const renderEventActionButton = (event: TimelineEvent<EventMetadata>, isSubEvent=false) => {`

Then in your `renderSubEvent` method you use `{renderEventActionButton(subEvent, true)}`

MonkeyDo (changes requested):
Looking better and better!

I had a bit more brainstorming about expanding the subevent, it will change a few things in the code but simplify it a lot.

frontend/css/timeline.less
```diff
@@ -79,6 +82,41 @@
         padding-top: 1em;
         padding-left: 2.5em;
       }
+
+      ul {
+        opacity: 0.8;
+        // Horizontal line
+        &::before {
+          content: "";
+          height: 1px;
```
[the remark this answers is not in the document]
Suvid-Singhal: 1.5px looks too thick on my display 🤔
image

I think 1.25px looks better
What do you think?

frontend/js/src/user-feed/UserFeed.tsx
Suvid-Singhal: I see
Fixed it!

frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -709,6 +923,20 @@ export default function UserFeedPage() {
                         </div>
 
                         {renderEventContent(event)}
+
+                        {event.event_type === EventType.THANKS &&
+                        !event.hidden ? (
+                          <ul>
```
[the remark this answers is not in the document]
Suvid-Singhal: Yea, that indeed seems like a better way. Fixed it now!

Suvid-Singhal:
Hi @MonkeyDo,
I have implemented all the changes now! Thanks a lot for suggesting good coding practices :)
Just have to fix the failing tests now 😅

Suvid-Singhal:
Fixed test issues as well!

frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -540,6 +675,34 @@ export default function UserFeedPage() {
...
     }
+    if (event_type === EventType.THANKS) {
+      const {
+        original_event_id,
+        original_event_type,
+        thanker_id,
+        thanker_username,
+        thankee_id,
+        thankee_username,
+        blurb_content,
```
MonkeyDo, line 686: Unused variables here:
```suggestion
        original_event_type,
        thanker_username,
        thankee_username,
```

frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -487,9 +591,40 @@ export default function UserFeedPage() {
...
+          <Card className="listen-card">
+            <div className="main-content">{metadata?.blurb_content}</div>
+          </Card>
+        </div>
+      );
+    }
     return null;
   };
 
+  const renderSubEvent = (subEvent: TimelineEvent<EventMetadata> | undefined) => {
```
MonkeyDo, line 607: This function needs to be moved under the `renderEventText` function, I am getting a warning "'renderEventText' was used before it was defined"

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -628,6 +660,78 @@ def create_personal_recommendation_event(user_name):
     return jsonify(event_data)
 
 
+@user_timeline_event_api_bp.post('/user/<user_name>/timeline-event/create/thanks')
+@crossdomain
+@ratelimit()
+def create_thanks_event(user_name):
+    '''
+    Make the user thank a pin/suggestion. The request should post
```
MonkeyDo, line 668: Is the end of the sentence missing?

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -1049,6 +1175,55 @@ def get_personal_recording_recommendation_events(
...
+        try:
+            thank = APIThanksEvent(
+                user_id=event.user_id,
+                created=event.created.timestamp(),
+                original_event_id=event.metadata.original_event_id,
+                original_event_type=event.metadata.original_event_type.value,
+                blurb_content=event.metadata.blurb_content,
+                thanker_id=event.metadata.thanker_id,
+                thanker_username=event.metadata.thanker_username,
+                thankee_id=event.metadata.thanker_id,
```
MonkeyDo, line 1212: This looks incorrect:
```suggestion
                thankee_id=event.metadata.thankee_id,
```

frontend/css/timeline.less
```diff
@@ -79,6 +82,53 @@
...
+          padding-left: 3em;
+        }
+
+        details .event-icon:before {
+          content: "+";
+          font-weight: bold;
+          background-color: @gray;
+          aspect-ratio: 1;
+          border-radius: 50%;
+          width: 0.7em;
```
MonkeyDo, line 116:
```suggestion
          width: 0.65em;
```

frontend/css/timeline.less
```diff
@@ -79,6 +82,53 @@
...
+          display: block;
+          color: white;
+          font-size: 2em;
+          border: none;
+          line-height: 0.65em;
+          text-align: center;
+          margin-top: 0.65em;
+          margin-left: -0.5em;
```
MonkeyDo, line 124: Adding a small convenience
```suggestion
          margin-left: -0.5em;
          cursor: pointer;
```

frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -419,6 +468,59 @@ export default function UserFeedPage() {
...
+                  original_event_id: event.id!,
+                  original_event_type: event.event_type,
+                });
+              }}
+              dataToggle="modal"
+              dataTarget="#ThanksModal"
+            />
+          ) : (
+            null
+          )}
```
MonkeyDo, line 490: I like this conditional rendering format better, with a boolean check followed by the component, rather than having to return `null` or `undefined` with the ternary operator
```suggestion
          {!isSubEvent && (
            <ListenControl
              title="Thanks"
              text=""
              icon={faHandshake}
              buttonClassName="btn btn-link btn-xs"
              action={() => {
                NiceModal.show(ThanksModal, {
                  original_event_id: event.id!,
                  original_event_type: event.event_type,
                });
              }}
              dataToggle="modal"
              dataTarget="#ThanksModal"
            />
          )} 
```

frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -680,7 +843,16 @@ export default function UserFeedPage() {
...
+                      const {original_event_id, original_event_type} = metadata as ThanksMetadata;
+                      const subEvent = events?.find(
+                        (evt) => evt.id === original_event_id && evt.event_type === original_event_type
+                      );
+                      subEventElement = renderSubEvent(subEvent);
+                    }
+
                     return (
                       <li
                         className="timeline-event"
```
MonkeyDo, line 858: I can't add a comment below, but there is an attribute missing in the icon rendering:
Please add the `fixedWidth` attribute to both FontAwesomeIcons in the `event-description` div.
This will improve the horizontal positioning of the icons (all of them, not just the thank events):

image

image

frontend/css/timeline.less
```diff
@@ -79,6 +82,53 @@
...
+        }
+
+        details[open] .event-icon:before {
+          background-color: @gray-light;
```
MonkeyDo, line 128: I think I suggested the wrong color here, it's a bit too light compared to the non-expanded color.
New suggestion:
```suggestion
          background-color: lighten(@gray, 25%);
```

MonkeyDo (changes requested):
A little bit of cleanup left and a few corrections and improvements, but at this point I'm going at it with a fine-toothed comb.
The feature works and looks great, nice work !

listenbrainz/webserver/views/user_timeline_event_api.py
Suvid-Singhal: Thanks for pointing this out!

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -628,6 +660,78 @@ def create_personal_recommendation_event(user_name):
...
+        else:
+            raise APIBadRequest(
+                "This event type is not supported for thanking")
+
+        if not result:
+            raise APIBadRequest(f"{event_type} event with id {row_id} not found")
+        # result.user_id -> thankee's id
+        # user_name -> thanker's username
+        # get thankee's username from get_users_by_id function
+        # user['id'] -> thanker's id
```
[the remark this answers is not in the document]
Suvid-Singhal: Although it was meant to be left here only but after a second thought that it would be better to remove it as it is not that hard to understand.

frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -378,7 +393,41 @@ export default function UserFeedPage() {
...
+          text=""
+          icon={faEyeSlash}
+          buttonClassName="btn btn-link btn-xs"
+          // eslint-disable-next-line react/jsx-no-bind
+          action={() => {
+            hideEventMutation(event);
+          }}
+        />
+      );
+    }
```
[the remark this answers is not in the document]
Suvid-Singhal: Ohh, I might have resolved it by mistake!

listenbrainz/webserver/views/user_timeline_event_api.py
Suvid-Singhal: Sorry forgot to delete it

Suvid-Singhal:
Implemented the suggested changes now :)

MonkeyDo (approved):
Looking great, thanks !

I took the liberty of doing one final small refactor as the renderEventActionButton was getting a bit messy.

Suvid-Singhal:
Thanks a lot @MonkeyDo for bearing with me 😅 and providing such detailed feedback and helping me inculcate good coding practices. Also, that native HTML solution was really cool and efficient as compared to the React solution. Really liked that one and learnt new things.
It was a nice experience implementing this feature.

MonkeyDo:
> Thanks a lot @MonkeyDo for bearing with me 😅 and providing such detailed feedback and helping me inculcate good coding practices. Also, that native HTML solution was really cool and efficient as compared to the React solution. Really liked that one and learnt new things. It was a nice experience implementing this feature.

I am glad that was helpful to you as it was to us !
Also always happy to share knowledge. In this case, I don't think I ever used the HTML `details`, but I had it in the back of my head. Happy to have been able to put that to use :)

MonkeyDo:
# Run the [SQL migration script](https://github.com/metabrainz/listenbrainz-server/pull/3143/files#diff-e9d05c8922ce6343cb56ab1138807c74f05b8263c77c3656218a6e106fa00d06) when releasing this PR

PR #3145, opened 2025-01-23
LB1727: Filters for link listens page

granth23:
# Problem

[LB1727](https://tickets.metabrainz.org/projects/LB/issues/LB-1727?filter=allopenissues): Filters for link listens page

# Solution

[FIlter.webm](https://github.com/user-attachments/assets/32672e6c-57c9-4b74-8953-094477193d31)

Filtering functionality added

# Action

### DIFF
--- a/frontend/js/src/settings/link-listens/LinkListens.tsx
+++ b/frontend/js/src/settings/link-listens/LinkListens.tsx
@@ -6,6 +6,7 @@ import {
   faLink,
   faQuestionCircle,
   faTrashAlt,
+  faSearch,
 } from "@fortawesome/free-solid-svg-icons";
 import { Link, useLocation, useSearchParams } from "react-router-dom";
 import { toast } from "react-toastify";
@@ -17,6 +18,7 @@ import { groupBy, isNil, isNull, isString, pick, size, sortBy } from "lodash";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { useQuery } from "@tanstack/react-query";
 import ReactTooltip from "react-tooltip";
+import Fuse from "fuse.js";
 import Loader from "../../components/Loader";
 import ListenCard from "../../common/listens/ListenCard";
 import ListenControl from "../../common/listens/ListenControl";
@@ -100,6 +102,41 @@ export default function LinkListensPage() {
   const [unlinkedListens, setUnlinkedListens] = React.useState<
     Array<UnlinkedListens>
   >(unlinkedListensProps);
+  const [filterType, setFilterType] = React.useState("album"); // Default filter type
+  const [searchQuery, setSearchQuery] = React.useState(""); // To store the search input
+  const handleSearch = () => {
+    if (filterType === "album") {
+      const fuzzysearch = new Fuse(unlinkedListensProps, {
+        keys: ["release_name"],
+      });
+      const filtered = fuzzysearch
+        .search(searchQuery)
+        .map((result) => result.item);
+      setUnlinkedListens(filtered);
+    } else if (filterType === "artist") {
+      const fuzzysearch = new Fuse(unlinkedListensProps, {
+        keys: ["artist_name"],
+      });
+      const albumsByArtist = fuzzysearch
+        .search(searchQuery)
+        .map((result) => result.item.release_name);
+      const filtered = unlinkedListensProps.filter((listen) =>
+        albumsByArtist.includes(listen.release_name)
+      );
+      setUnlinkedListens(filtered);
+    } else if (filterType === "song") {
+      const fuzzysearch = new Fuse(unlinkedListensProps, {
+        keys: ["recording_name"],
+      });
+      const albumsByArtist = fuzzysearch
+        .search(searchQuery)
+        .map((result) => result.item.release_name);
+      const filtered = unlinkedListensProps.filter((listen) =>
+        albumsByArtist.includes(listen.release_name)
+      );
+      setUnlinkedListens(filtered);
+    }
+  };
   const unsortedGroupedUnlinkedListens = groupBy(
     unlinkedListens,
     "release_name"
@@ -282,6 +319,35 @@ export default function LinkListensPage() {
           Updates every Monday at 2AM (UTC). Last updated{" "}
           {lastUpdatedHumanReadable}
         </p>
+      )}
+      {unlinkedListensProps.length > 0 && (
+        <div className="filter-div">
+          <div className="dropdown">
+            <select
+              value={filterType}
+              onChange={(e) => setFilterType(e.target.value)}
+              className="filter-btn btn btn-info"
+            >
+              <option value="album">Album</option>
+              <option value="artist">Artist</option>
+              <option value="song">Song</option>
+            </select>
+          </div>
+          <input
+            type="text"
+            placeholder="Search"
+            value={searchQuery}
+            onChange={(e) => setSearchQuery(e.target.value)}
+            className="filter-search"
+          />
+          <button
+            onClick={handleSearch}
+            className="filter-btn btn btn-info"
+            type="submit"
+          >
+            <FontAwesomeIcon icon={faSearch} />
+          </button>
+        </div>
       )}
       <br />
       <div>
--- a/frontend/css/filter-search.less
+++ b/frontend/css/filter-search.less
@@ -1,0 +1,26 @@
+.filter-div {
+    display: flex;
+    align-items: center;
+}
+
+.filter-search {
+    font-size: 12px;   
+    font-family: inherit;
+    padding: 3px 10px;
+    width: 300px;
+    margin: 0;
+}
+
+.dropdown option {
+    color: black;
+    background-color: white;
+    text-align: left;
+    font-family: inherit;
+    font-size: 12px;
+    border: none;
+}
+
+.filter-btn {
+    margin:0;
+    font-size: 12px;
+}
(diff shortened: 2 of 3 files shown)


frontend/js/src/settings/link-listens/LinkListens.tsx
```diff
@@ -100,6 +102,41 @@ export default function LinkListensPage() {
...
+  const [searchQuery, setSearchQuery] = React.useState(""); // To store the search input
+  const handleSearch = () => {
+    if (filterType === "album") {
+      const fuzzysearch = new Fuse(unlinkedListensProps, {
+        keys: ["release_name"],
+      });
+      const filtered = fuzzysearch
+        .search(searchQuery)
+        .map((result) => result.item);
+      setUnlinkedListens(filtered);
```
anshg1214, line 115: In entity search, users should be redirected to the first page of results whenever a new search query is initiated.

anshg1214 (changes requested):
The PR Looks good. However you can perform cleanup by using bootstrap classes.

frontend/js/src/settings/link-listens/LinkListens.tsx
```diff
@@ -283,6 +322,39 @@ export default function LinkListensPage() {
...
+          onSubmit={handleSearch}
+        >
+          <select
+            value={filterType}
+            onChange={(e) => setFilterType(e.target.value)}
+            className="btn btn-info"
+          >
+            <option value="album">Album</option>
+            <option value="artist">Artist</option>
+            <option value="song">Song</option>
```
anshg1214, line 338:
```suggestion
            <option value="track">Track</option>
```

anshg1214 (changes requested):
One final change, and we're good to go.

frontend/js/src/settings/link-listens/LinkListens.tsx
```diff
@@ -100,6 +102,46 @@ export default function LinkListensPage() {
...
+  const [filterType, setFilterType] = React.useState("album"); // Default filter type
+  const [searchQuery, setSearchQuery] = React.useState(""); // To store the search input
+  const handleSearch = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (!searchQuery) {
+      setUnlinkedListens(unlinkedListensProps);
+      return;
+    }
+    if (filterType === "album") {
+      const fuzzysearch = new Fuse(unlinkedListensProps, {
```
MonkeyDo, line 110: We can pass an options object to Fuse, but let's refactor the following code to stay D.R.Y:
This object created at the root of the component:
```ts
const searchOptions: IFuseOptions<UnlinkedListens> = {
    minMatchCharLength: 3,
    threshold: 0.25,
  };
```

And in the handleSearch method: 
```ts
    let key;
    switch (filterType) {
      case "artist":
        key = "artist_name";
        break;
      case "song":
        key = "recording_name";
        break;
      case "album":
      default:
        key = "release_name";
        break;
    }

    const fuzzysearch = new Fuse(unlinkedListensProps, {
      keys: [key],
      ...searchOptions,
    });
    const filtered = fuzzysearch
      .search(searchQuery)
      .map((result) => result.item);
    setUnlinkedListens(filtered);
```

I think the bit of logic that previously mapped and filtered the albumsByArtist is not required.

This all should simplify the code a fair amount.

MonkeyDo (changes requested):
Nice work so far, it's working quite nicely!
I think searching by entity type was the right choice, otherwise it would get quite confusing I think.

Currently missing:

- a button to clear the search (currently emptying the search bar and clicking search results in an empty page)
- another state variable to hold the original unlinkedListensProps, which is what we should use as the searchable list. If you link a track, it is removed from the list of options (see for example in the `openMultiTrackMappingModal` function). We should be able to hold state of the modified list while also beign able to filter it.
- a "no results" message to show users

frontend/js/src/settings/link-listens/LinkListens.tsx
granth23: I dont think I can remove the the logic behind albumsByArtist since we planned to show the entire album even if a single song matches from the album.

I have made changes to incorporate switch case.

MonkeyDo: It still works fine because we return a list of matches (whichever tracks match the search) which then get grouped by album, so it only shows matches, grouped by album:
image

frontend/js/src/settings/link-listens/LinkListens.tsx
```diff
@@ -100,6 +102,46 @@ export default function LinkListensPage() {
...
     Array<UnlinkedListens>
   >(unlinkedListensProps);
+  const [filterType, setFilterType] = React.useState("album"); // Default filter type
+  const [searchQuery, setSearchQuery] = React.useState(""); // To store the search input
+  const handleSearch = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (!searchQuery) {
+      setUnlinkedListens(unlinkedListensProps);
+      return;
+    }
```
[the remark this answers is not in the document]
granth23: The last commit didn't have the below code segment
`if (!searchQuery) {
    setUnlinkedListens(unlinkedListensProps);
    return;
  }`
So that has been catered to already, a clear search button leading to the unfiltered Link Listens is a must so will make the changes.

frontend/js/src/settings/link-listens/LinkListens.tsx
granth23: I feel we should show the entire album with all the tracks of the album instead of the tracks which match the query. If you want me to make it specific to the tracks, I could make that change.

frontend/js/src/settings/link-listens/LinkListens.tsx
```diff
@@ -283,6 +325,39 @@ export default function LinkListensPage() {
           {lastUpdatedHumanReadable}
         </p>
       )}
+      {unlinkedListensProps.length > 0 && (
+        <form
+          className="input-group input-group-flex"
+          style={{ maxWidth: "400px", gap: "10px" }}
```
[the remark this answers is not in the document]
granth23: I’m a bit confused about the second point you mentioned. From my perspective, if a user filters and then links a track, the next search should display only unlinked listens that are still unlinked. Why should the original list be used as the searchable list, considering it might include tracks that have already been linked? Please let me know if I’m misunderstanding something.

granth23:
I’m a bit confused about the second point you mentioned. From my perspective, if a user filters and then links a track, the next search should display only unlinked listens that are still unlinked. Why should the original list be used as the searchable list, considering it might include tracks that have already been linked? Please let me know if I’m misunderstanding something.

frontend/js/src/settings/link-listens/LinkListens.tsx
MonkeyDo: OK, I get your point, that is what we originally discussed and I got confused :0

frontend/js/src/settings/link-listens/LinkListens.tsx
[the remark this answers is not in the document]
granth23: Done

frontend/js/src/settings/link-listens/LinkListens.tsx
MonkeyDo: Let me think about it for a while, after testing it I'm not 100% sure of that decision.
It feels a bit weird to show a bunch of unrelated matches when searching for a specific song.
From a user's perspective, what are they trying to accomplish when searching for a specific track?

Without a way to highlight matches inside an album, you can end up with a confusing experience.
For example, here are my search results for "Bayou". There's a a track in there that matches, but it's hard to find it:
image

I think the options include:
- Show only matched tracks, but still show the album groups as before
- Show only matched tracks, but hide the link-entire-album button (only when searching for a track)
- Show only matched tracks, but not as album groups, only standalone tracks
- Show the whole album but highlight matched tracks (for example different card color or something)

Not sure which of those would be most useful. I think we should continue and finish this PR in the original direction we decided, deploy it, and get some feedback from users.

MonkeyDo:
Deployed to https://test.listenbrainz.org/settings/link-listens/ for further testing

frontend/js/src/settings/link-listens/LinkListens.tsx
```diff
@@ -97,9 +99,58 @@ export default function LinkListensPage() {
...
+    } else {
+      const albumsByArtist = fuzzysearch
+        .search(searchQuery)
+        .map((result) => result.item.release_name);
+      const filtered = originalUnlinkedListens.filter((listen) =>
+        albumsByArtist.includes(listen.release_name)
+      );
+      setUnlinkedListens(filtered);
+    }
+    setSearchQuery("");
```
MonkeyDo, line 146: Not sure I like that the search terms are auto-deleted here.
From my point of view, it gets in my way if I mistype something, or if I want to check if I typed correctly, or if I didn't select the right entity type.

What was your use-case for clearing the input?

granth23: Since we applied Fuzzy Search the spelling mistakes will be taken into consideration but if you want we can surely remove that and let the searched content be there. Anything else that needs to be looked into?

MonkeyDo: That's the only detail I would change, otherwise I think it looks and works great!
Actually quite useful!

Let's wait for @anshg1214 's review, but after that for me it's a green light :+1:

granth23: Cool, made that change for now.

anshg1214 (approved):
LGTM

PR #3149, opened 2025-01-26
LB-1214, LB-1215: Add playlist cover

anshg1214:
LB-1214, LB-1215


frontend/js/src/utils/APIService.ts
```diff
@@ -1822,4 +1822,21 @@ export default class APIService {
     await this.checkStatus(response);
     return response.json();
   };
+
+  getPlaylistCoverArt = async (
```
MonkeyDo, line 1826: This is now unused, with the server endpoint removed.

I think the endpoint could actually be useful. In particular, I'm thinking about the grid view on the playlists page, where we don't have all the playlist metadata but we could show each playlist's cover art.

Also thinking about atom feeds where we sometimes need a URL to the SVG rather than SVG content.

MonkeyDo (commented):
Great work!

I'm loving what I'm seeing.
Those little image layout icons are going to be very useful in the art creator page too!

Couple of comments related to art api endpoints, but already I think this is good for a v1

frontend/js/src/utils/APIService.ts
anshg1214: So, LMK do a thing. I'll remove the frontend function to call the server api, and bring back the server endpoint.

frontend/img/playlist-cover-art/cover-art_3-1.svg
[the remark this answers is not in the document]
anshg1214: Fixed

frontend/js/src/utils/APIService.ts
MonkeyDo: Currently I am getting an error about image size when I try the endpoint on test:
"image size must be between 128 and 1024, inclusive."

URL was POST https://test-api.listenbrainz.org/1/art/playlist/5bb057eb-b38a-4a8d-8218-c5004f052fb2/250/1/

On a side note: why did you want a POST endpoint rather than GET?
My envisioned used would be a simple `` or something along those lines in the PlaylistCard component to get the image directly, but that would require a GET endpoint.

What was your use case?

anshg1214: So the endpoint's structure is `/playlist/<uuid:playlist_mbid>/<int:dimension>/<int:layout>`. The dimension and layout are based on those defined in the `GRID_TILE_DESIGNS`. 

Since, a lot of playlists are private, we cannot remove authentication from the request, so therefore we cannot add it as a src in the image HTML Tag.

MonkeyDo: Ah bummer, I forgot about that. Yeah, I guess we fetch the SVG content with a fetch call then :+1:

Still, the image_size issue was not related to the dimension or layout params, not sure what is happening there. Did you successfully test a POST request with auth on the test-api.lb art endpoint?

anshg1214: Yes. POST https://test-api.listenbrainz.org/1/art/playlist/5bb057eb-b38a-4a8d-8218-c5004f052fb2/4/0

MonkeyDo: It is indeed working. :tada: 
Not sure what I did wrong before ¯\_(ツ)_/¯

anshg1214:
@MonkeyDo Do you know why the tests are failing on this PR?

listenbrainz/webserver/views/playlist.py
```diff
@@ -53,6 +106,41 @@ def load_playlist(playlist_mbid: str):
...
+
+    serialized_playlist = playlist.serialize_jspf()
+
+    selected_cover_art = playlist.additional_metadata.get("cover_art")
+    if not selected_cover_art and len(options) > 0:
+        sorted_options = sorted(
+            options,
+            key=lambda x: (-x["dimension"], x["layout"])
+        )
+        selected_cover_art = sorted_options[0]
```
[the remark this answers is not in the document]
anshg1214: I'm not sure on it. Let's keep it static for now.

PR #3150, opened 2025-01-26
LB-1726: Fixed following status glitching when followed from similar pane

fettuccinae:
# Problem

Follow button unable to correctly reflect the changes made to the following status of a user if followed from similar users pane and showing error if re-clicked.
https://tickets.metabrainz.org/browse/LB-1726 

# Solution

In updateFollowingList method, when a change was being made, it was only making a change in the followingList(the list used for similar users pane), but not the currentUserFollowingList which passed the old list to FollowersFollowingModal and hence the user following status was rendering from old list. So, i updated both currentUserFollowingList and followingList after a change is made.
Also, added verification if the current user.name is same to the profile user.name, and if it is same, then push or slice the usernames from following pane.


anshg1214 (approved):
LGTM!

fettuccinae:
Thanks! Are there any changes I should make?

PR #3151, opened 2025-01-27
LB-1722: Part B, making the link listens page more discoverable. 

granth23:
# Problem

The page `link-listens` is tough to locate for new users.

# Solution

image

Added a button in the `link-listen` popup to make the page more easy to locate.

# Action


frontend/js/src/common/listens/MBIDMappingModal.tsx
```diff
@@ -285,21 +271,81 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
...
+                      justifyContent: "center", // Optional: centers content horizontally
+                      marginTop: "15px",
+                    }}
+                  >
+                    <img
+                      src="/static/img/icons/stacked_link_icon.svg"
+                      width="20"
+                      height="20"
+                      alt="MetaBrainz"
+                    />
```
anshg1214, line 300: Instead of loading the svg using image tag, add the svg in [icon.ts](https://github.com/granth23/listenbrainz-server/blob/LB-1722--Part-B-Making-page-more-discoverable/frontend/js/src/utils/icons.ts), and import it from there.

frontend/js/src/common/listens/MBIDMappingModal.tsx
```diff
@@ -285,21 +271,81 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
...
+                  <button
+                    type="button"
+                    className="btn btn-default"
+                    onClick={massLinkButton}
+                    style={{
+                      backgroundColor: COLOR_LB_BLUE,
+                      display: "flex", // Enables flexbox
+                      alignItems: "center", // Vertically centers the content
+                      justifyContent: "center", // Optional: centers content horizontally
+                      marginTop: "15px",
```
anshg1214, line 292: We usually avoid writing inline css. Can you move it to the css file instead?

anshg1214 (changes requested):
The PR looks good. Just some minor refactoring is required, and we should be good to go.

frontend/js/src/common/listens/MBIDMappingModal.tsx
granth23: There is a slight problem with that, we can't define opacity with `FontAwesome`, a workaround for that is defining the entire path of three chains as one path with different colors instead of opacity and in my opinion using an `svg` is a better option than that.

anshg1214 (changes requested):

frontend/js/src/common/listens/MBIDMappingModal.tsx
anshg1214: Okay, thank you for clarifying.

frontend/js/src/common/listens/MBIDMappingModal.tsx
```diff
@@ -285,21 +267,74 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
...
+                    type="button"
+                    className="btn btn-default modal-mass-link-button"
+                    onClick={massLinkButton}
+                  >
+                    <img
+                      src="/static/img/icons/stacked_link_icon.svg"
+                      width="20"
+                      height="20"
+                      alt="MetaBrainz"
+                    />
```
MonkeyDo, line 289: I am also of the opinion that we might as well use the FontAwesome icons. Here's a working example:
With those changes you can remove the new SVG file.
```
                    <span
                      className="fa-layers fa-fw"
                      style={{ marginRight: "0.5em" }}
                    >
                      <FontAwesomeIcon icon={faLink} />
                      <FontAwesomeIcon
                        icon={faLink}
                        color="#FFFFFFA1"
                        transform="right-5"
                      />
                      <FontAwesomeIcon
                        icon={faLink}
                        color="#ffffff42"
                        transform="right-10"
                      />
                    </span>
```

image

frontend/js/src/common/listens/MBIDMappingModal.tsx
```diff
@@ -285,21 +267,74 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
...
+              <div
+                style={{
+                  display: "flex",
+                  justifyContent: "space-between",
+                  alignItems: "center",
+                }}
               >
-                Add mapping
-              </button>
+                <div style={{ textAlign: "left", marginBottom: "10px" }}>
```
MonkeyDo, line 278: The style here are not required, I'll comment above about adding a class to the parent element for the bottom margin
```suggestion
                <div>
```

frontend/js/src/common/listens/MBIDMappingModal.tsx
```diff
@@ -285,21 +267,74 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
...
-                type="submit"
-                className="btn btn-success"
-                disabled={!selectedRecording}
+            <div className="modal-footer" style={{ textAlign: "left" }}>
+              <div
+                style={{
+                  display: "flex",
+                  justifyContent: "space-between",
+                  alignItems: "center",
+                }}
```
MonkeyDo, line 276: We can use the utility class `mb-10` here to add the same 10px bottom margin you had on the mass link button.
Also don't think we need `alignItems: "center"` here.
```suggestion
              <div
                className="mb-10"
                style={{
                  display: "flex",
                  justifyContent: "space-between",
                }}
```

frontend/css/main.less
```diff
@@ -43,6 +43,7 @@
 @import "scroll-container.less";
 @import "flairs.less";
 @import "utilities.less";
+@import "link-listens-modal.less";
```
MonkeyDo, line 46: With CSS classes suggestions below, you'll be able to remove this line and the new file itself.

frontend/js/src/common/listens/MBIDMappingModal.tsx
```diff
@@ -56,6 +58,11 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
     setTimeout(modal.remove, 200);
   }, [modal]);
 
+  const massLinkButton = React.useCallback(() => {
```
MonkeyDo, line 61: This could be renamed to describe the action, something along the lines of 
```suggestion
  const closeAndGoToMassLink = React.useCallback(() => {
```

MonkeyDo (changes requested):
Nice work!

I've made a bunch of comments below, mainly small things related to how we usually do things in this codebase, but all minor changes and nitpicks.

Looks good!
I think alogn with your other PR this resolves the issue entirely :+1:

frontend/js/src/common/listens/MBIDMappingModal.tsx
```diff
@@ -285,21 +262,85 @@ export default NiceModal.create(({ listenToMap }: MBIDMappingModalProps) => {
...
+                  <button
+                    type="submit"
+                    className="btn btn-success"
+                    disabled={!selectedRecording}
+                  >
+                    Add mapping
+                  </button>
+                </div>
+              </div>
+              <div className="small help-block">
```
MonkeyDo, line 316: All three of these lines need the help-block class removed, for placement:
```suggestion
              <div className="small">
```

MonkeyDo (changes requested):
Final little forgotten change, apart from that it's good to go for me!

MonkeyDo (approved):
Good to go as far as I'm concerned!

PR #3152, opened 2025-01-27
LB-1509: Genre Explorer

anshg1214:
LB-1509


MonkeyDo:
Shall we do another pass on this feature, push it past the finish line?

PR #3153, opened 2025-01-28
LB-1729: Flairs settings shows incorrect time remaining

fettuccinae:
# Problem

Flair section in settings page shows the invalid status of user's flair. It uses both LB API's endpoint (donors/all-flairs) and MB's endpoint(nag-check), MB's endpoint returns the user's donation data without setting the threshold (FLAIR_MONTHLY_DONATION_THRESHOLD = 5) and it overrides the flairUnlocked value.

    https://tickets.metabrainz.org/browse/LB-1729

# Solution

 By discarding the nag value returned by nag-checker and not setting flairUnlocked value to it, we can just use LB's endpoint to verify user's donation status and then use nag-checker's daysLeft value to show the time left.

### DIFF
--- a/frontend/js/src/settings/flairs/FlairsSettings.tsx
+++ b/frontend/js/src/settings/flairs/FlairsSettings.tsx
@@ -67,8 +67,8 @@ export default function FlairsSettings() {
           `https://metabrainz.org/donations/nag-check?editor=${name}`
         );
         const values = await response.text();
-        const [shouldNag, daysLeft] = values.split(",");
-        setFlairUnlocked(!Number(shouldNag));
+        // discard the "shouldNag" value
+        const [_, daysLeft] = values.split(",");
         setUnlockDaysLeft(Math.max(Number(daysLeft), 0));
       } catch (error) {
         // eslint-disable-next-line no-console


MonkeyDo (approved):
Thank you!
I think this is a good stopgap measure to not show confusing information to the user, until we figure out how we want to fix the larger issue (considering the daysLeft is also calculated without the threshold, so is probably not completely correct either)

MonkeyDo:
I will ask however that you describe future commits more clearly than a single word :smiley:

fettuccinae:
> I will ask however that you describe future commits more clearly than a single word 😃

Sorry about that, my git was having some problems yesterday. Will take care from next time 👍

fettuccinae:
> Thank you! I think this is a good stopgap measure to not show confusing information to the user, until we figure out how we want to fix the larger issue (considering the daysLeft is also calculated without the threshold, so is probably not completely correct either)

Thanks! do we have any ideas on how we might approach fixing the larger issue?

MonkeyDo:
I think that one is going to require more team members and a long discussion. I'm currently not sure how to go about it, but all in all I think this will require someone from the MetaBrainz team to solve, and we possibly have to wait until some other project is finished (we are in the —very slow— process of moving user accounts from the  MusicBrainz databse to the MetaBrainz DB).

So I guess don't worry about it, you've done all you can be expected to do on this topic for now :)
Thanks again for the simple solution!

PR #3154, opened 2025-01-28
Dissmisable alerts

fettuccinae:
# Problem

The alert wasnt dismissable and it felt like it reflected the status of user's last.fm connection.
https://tickets.metabrainz.org/browse/LB-1728
    

# Solution

Added a button which onclick disables the visibility of alert.


MonkeyDo (changes requested):
I think  that is a good idea, thanks for opening a PR !

There is a simpler way to dismiss alerts. You don't need a stats variable, just the close button with no onClick param, and a few relevant css classes: you can follow the steps in 
 https://getbootstrap.com/docs/4.0/components/alerts/#dismissing
 
 The button html content is very similar to what you currently have:
 ```
 <button type="button" class="close" data-dismiss="alert" aria-label="Close">
    <span aria-hidden="true">&times;</span>
  </button>
 ```

PR #3156, opened 2025-01-28
Incremental Stats deployment fixes

amCap1712:
Currently, the database for each stat is generated and the previous database for the same stat is deleted daily. However, incremental stats generates messages only for users with listens in incremental dumps, assuming stats for users without new listens are already available on the LB server. To maintain this assumption, each stats database must have a fixed name. Therefore, avoid sending the database name from the LB server to Spark in stat generation requests. During stats generation, Spark determines whether it's a full or incremental generation. For full stats, a new database name is created and sent; for incremental stats, a database prefix is sent for the LB server to look up the existing database and insert the stats accordingly.

### DIFF
--- a/listenbrainz/spark/request_manage.py
+++ b/listenbrainz/spark/request_manage.py
@@ -4,6 +4,7 @@ from datetime import date
 
 import click
 import orjson
+from dateutil.relativedelta import relativedelta, MO
 from kombu import Connection
 from kombu.entity import PERSISTENT_DELIVERY_MODE, Exchange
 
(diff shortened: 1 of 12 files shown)


amCap1712:
For listening activity, a CROSS JOIN is needed to add rows for days on which the user didn't listen to any listens.  I had removed it in a previous PR due a misunderstanding. Its now added back. For the LEFT JOIN, need to pay attention on which table attributes being join-ed on are taken from (user_id and time_range) otherwise they will be NULL for the extra added rows.

mayhem (approved):

PR #3157, opened 2025-01-29
Alerts dismissal without states

fettuccinae:
#Problem

The alert wasnt dismissable and it felt like it reflected the status of user's last.fm connection.
https://tickets.metabrainz.org/browse/LB-1728

#Solution

Added a button which onclick disables the visibility of alert.

(This is #3154 's continuation because i pushed an old branch and that pr got closed)


MonkeyDo (approved):
Thank you! :rocket:

PR #3159, opened 2025-01-30
Link listens modal text cleanup

MonkeyDo:
As discussed in LB-1722, after the changes in #3151 the modal looks a bit too busy.
Aerozol suggested removing the intro text and linking to MB in the title.

While in there I also fixed a couple of small details I noticed (capitalization in title, remove cover art for unlinked listen preview) as well as on the Link Listens page (formatting of help tooltip text)

Before:
image

After:
image


Aerozol:
Beautiful! I apologize in advance if we get complaints and end up adding something back in... but to me this looks very beginner friendly/inviting <3

PR #3160, opened 2025-01-30
Add RTD status badge

samglt:
Since there's no monitoring for the documentation (that I'm aware of), adding this little badge should at least make it less hard to notice when it's broken


MonkeyDo (approved):
Thanks, that should indeed help :)

PR #3162, opened 2025-02-01
Fix: Add missing <ul> element to the template.

troyvassalotti:
# Problem

The Top Artists feed HTML was missing the `<ul>` element to wrap all the generated `<li>` elements.

# Solution

I've added the missing `<ul>` element to the template.


mayhem (approved):
Thanks for catching this!

PR #3163, opened 2025-02-01
Allow admins to "pause" a user's account

fettuccinae:
# Problem

Right now when we see users spamming their LB feeds, the only thing we can do is delete their account, which is pretty harsh. Admins should have the ability to "pause" an account, so that is stops accepting listens and then sends the user an email about this pause.
https://tickets.metabrainz.org/browse/LB-1730

# Solution

1. Added a flag to the user table called is_paused with a default value of false
2. There's 3 ways to import listens: submit-listens api endpoint, lastfm importer and spotify importer(last 2 use the same base class), Added an conditional statement which would check if the flag is set as True, if it is, the user is skipped during processing (for spotify and lastfm importers), for submit-listen api, it raises an APIUnauthorized error.
3. Added a "pause" function to set the "is_paused" flag to true and send an email to the user.
4. Added an "unpause" function to set the "is_paused" flag to false and sends an email to the user.
5. Added relevant test cases


listenbrainz/db/user.py
```diff
@@ -560,3 +567,68 @@ def get_all_usernames():
...
+
+    except sqlalchemy.exc.ProgrammingError as err:
+        logger.error(err)
+        raise DatabaseException("Couldn't pause user: %s" % str(err))
+
+
+def _notify_user_email(db_conn, user_id,paused):
+    user = get(db_conn, user_id, fetch_email=True)
+    if user["email"] is None:
+        return
```
mayhem, line 623: We should log an error here, just in case.

listenbrainz/db/user.py
```diff
@@ -560,3 +567,68 @@ def get_all_usernames():
...
+    url = current_app.config['SERVER_ROOT_URL']
+    template = 'emails/id_paused.txt' if paused else 'emails/id_unpaused.txt'
+    subject = 'ListenBrainz UserID Paused' if paused else 'ListenBrainz UserID Unpaused'
```
mayhem, line 626: I think the strings here should say:

"Your ListenBrainz account has been paused and is not accepting incoming listens"
and
"Your ListenBrainz account has been unpaused and is accepting incoming listens"

Options: Add the account name: "Your ListenBrainz accout ([account_name]) has been..."

listenbrainz/webserver/templates/emails/id_paused.txt
```diff
@@ -0,0 +1,7 @@
+Hi {{ username }}!
+
+We have paused your listenbrainz ID due to sus activities, you wont be able to import new listens.
```
mayhem, line 3: sus, lol. :)

Lets go with:

"We have paused your ListenBrainz account ([user name]) because of a problem with your account. Our community manager will reach out to you with the details as to why we took this step. Feel free to [Contact Us] if you have any questions about this."

and 

"Your ListenBrainz account ([account name]) has been restore and is now accepting listens again."

listenbrainz/webserver/views/api.py
```diff
@@ -76,6 +76,9 @@ def submit_listen():
     if mb_engine and current_app.config["REJECT_LISTENS_WITHOUT_USER_EMAIL"] and not user["email"]:
         raise APIUnauthorized(REJECT_LISTENS_WITHOUT_EMAIL_ERROR)
 
+    if user['is_paused']:
+        raise APIUnauthorized("user_id is paused.")
```
mayhem, line 80: " ... is paused and is currently not accepting listens. Feel free to [Contact Us] if you have any questions about this."

mayhem (changes requested):
Looking pretty good. A few things:

1. I've added some text on how to improve the wording of your messages.
2. You'll need to check the users paused status in the timescale writer as well: https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/timescale_writer/timescale_writer.py
3. How does an admin set/unset the admin flag? Right now that doesn't seem possible in the admin interface, so I think more work needs to be done there.

mayhem:
How can I pause/unpause a user in the admin interface? its not in the edit user page...

admin/sql/updates/2025-02-04-add-type-background-tasks.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+
+ALTER TYPE background_tasks_type ADD VALUE 'pause_user' AFTER 'export_all_user_data';
+
+ALTER TYPE background_tasks_type ADD VALUE 'unpause_user' AFTER 'pause_user';
+
+COMMIT;
```
reosarevok, line 7: Missing end newline

listenbrainz/background/pause.py
```diff
@@ -0,0 +1,23 @@
...
+
+def unpause_user(db_conn,user_id):
+    """ Unpause a user from ListenBrainz. First, changes the is_paused flag 
+    to false and then sends them an e-mail. 
+
+    Args:
+        user_id: the LB row ID of the user
+    """
+    db_user.unpause(db_conn,user_id)
+    db_conn.commit()
```
reosarevok, line 23: Missing end newline

listenbrainz/db/user.py
```diff
@@ -560,3 +567,70 @@ def get_all_usernames():
...
+    subject = ("Your ListenBrainz account %s has been paused and is not accepting incoming listens" % user["musicbrainz_id"] if paused
+                 else "Your ListenBrainz account %s has been unpaused and is accepting incoming listens" % user["musicbrainz_id"])
+    content = render_template(template, username=user["musicbrainz_id"], url=url)
+    send_mail(
+        subject=subject,
+        text=content,
+        recipients=[user["email"]],
+        from_name='ListenBrainz',
+        from_addr='noreply@'+current_app.config['MAIL_FROM_DOMAIN'],
+    )
```
reosarevok, line 636: Missing end newline

listenbrainz/webserver/templates/emails/id_paused.txt
```diff
@@ -0,0 +1,3 @@
+We have paused your ListenBrainz account {{ username }} because of a problem with your account.
+Our community manager will reach out to you with the details as to why we took this step.
+Feel free to Contact Us {{ url }} if you have any questions about this.
```
reosarevok, line 3: `Contact Us {{ url }}` is weird. Ideally "contact us" (lowercase) would be the link, but that might not be supported by the email system, in which case maybe:

```suggestion
Feel free to contact us (see {{ url }}) if you have any questions about this.
```

Also, shouldn't this have a signature like `id_unpaused`? The same seems fine.

listenbrainz/webserver/templates/emails/id_paused.txt
```diff
@@ -0,0 +1,3 @@
+We have paused your ListenBrainz account {{ username }} because of a problem with your account.
```
reosarevok, line 1: I see no reason why this wouldn't also start with `Hi {{ username }}!`

listenbrainz/webserver/views/api.py
```diff
@@ -76,6 +76,9 @@ def submit_listen():
     if mb_engine and current_app.config["REJECT_LISTENS_WITHOUT_USER_EMAIL"] and not user["email"]:
         raise APIUnauthorized(REJECT_LISTENS_WITHOUT_EMAIL_ERROR)
 
+    if user['is_paused']:
+        raise APIUnauthorized("user_id is paused and is currently not accepting listens. Feel free to Contact Us if you have any questions about this.")
```
reosarevok, line 80: I can see no reason why "Contact Us" should be uppercase here in the middle of a sentence. Please lowercase it.

Should this also include a link to the Contact Us page?

Would it make sense to have a constant for this in `utils.py` like for `REJECT_LISTENS_WITHOUT_EMAIL_ERROR` just above? Similarly, does this need to also be implemented separately for last.fm compatibility in https://github.com/metabrainz/listenbrainz-server/blob/d9ef6f3d5916709363946ee61258c8ab1746da29/listenbrainz/webserver/errors.py#L68 ?

reosarevok (changes requested):
Few small changes, plus a couple questions that an actual LB team member should probably check and give you an answer to :)

listenbrainz/webserver/views/api.py
mayhem: Yes, lets fix the case and also add a link to the contact us page. But, I am confused about the comment about the constant.... And yes, compat API also needs to be implemented! I will post another comment for this issue.

mayhem:
The last.fm compatibility API also need to check if the user is paused: https://github.com/metabrainz/listenbrainz-server/blob/d9ef6f3d5916709363946ee61258c8ab1746da29/listenbrainz/webserver/views/api_compat.py#L240

listenbrainz/webserver/views/api.py
reosarevok: Well, the previous error in this page at https://github.com/metabrainz/listenbrainz-server/blob/89684cd853537fa983fb0b04c36f1aea2a7909d8/listenbrainz/webserver/views/api.py#L77 uses a constant for the error text. If this needs to be reused for compat api, then probably double the reason to have a constant rather than reusing the string :)

listenbrainz/webserver/views/api.py
```diff
@@ -76,6 +76,9 @@ def submit_listen():
     if mb_engine and current_app.config["REJECT_LISTENS_WITHOUT_USER_EMAIL"] and not user["email"]:
         raise APIUnauthorized(REJECT_LISTENS_WITHOUT_EMAIL_ERROR)
 
+    if user['is_paused']:
+        raise APIUnauthorized("user_id is paused and is currently not accepting listens. Feel free to contact us if you have any questions about this. https://listenbrainz.org/about/")
```
mayhem, line 80: use http://metabrainz.org/contact for this link

listenbrainz/webserver/views/api_compat.py
```diff
@@ -240,7 +240,8 @@ def record_listens(data):
     user = db_user.get(db_conn, session.user_id, fetch_email=True)
     if mb_engine and current_app.config["REJECT_LISTENS_WITHOUT_USER_EMAIL"] and user["email"] is None:
         raise InvalidAPIUsage(CompatError.NO_EMAIL, output_format=output_format)  # No email available for user in LB
-
+    if user['is_paused']:
+        raise APIUnauthorized("user_id is paused and is currently not accepting listens. Feel free to contact us if you have any questions about this. https://listenbrainz.org/about/")
```
mayhem, line 244: same for this link

mayhem (changes requested):
Two minor changes and then I'm happy with this PR -- many thanks! @amCap1712 -- can you also please take a look and make sure we didn't miss something important?

mayhem (commented):
Thanks for making the last round of changes!

mayhem:
OK, now just a look from @amCap1712 -- I'm happy with this PR. Thanks!

amCap1712:
I'll try to review this today.

listenbrainz/webserver/views/api_compat.py
```diff
@@ -240,7 +240,8 @@ def record_listens(data):
     user = db_user.get(db_conn, session.user_id, fetch_email=True)
     if mb_engine and current_app.config["REJECT_LISTENS_WITHOUT_USER_EMAIL"] and user["email"] is None:
         raise InvalidAPIUsage(CompatError.NO_EMAIL, output_format=output_format)  # No email available for user in LB
-
+    if user['is_paused']:
+        raise APIUnauthorized("user_id is paused and is currently not accepting listens. Feel free to contact us if you have any questions about this. https://metabrainz.org/contact")
```
amCap1712, line 244: This needs to be a InvalidAPIUsage error.

listenbrainz/webserver/views/api_compat_deprecated.py
```diff
@@ -126,6 +126,8 @@ def submit_listens():
         return 'FAILED Invalid data submitted!\n', 400
 
     user = db_user.get(db_conn, session.user_id)
+    if user['is_paused']:
+        raise APIUnauthorized("user_id is paused and is currently not accepting listens. Feel free to contact us if you have any questions about this. https://metabrainz.org/contact")
```
amCap1712, line 130: Like other errors in this api file, this should be a plain text message with the error code.

listenbrainz/db/user.py
```diff
@@ -560,3 +567,70 @@ def get_all_usernames():
...
+        })
+        db_conn.commit()
+        _notify_user_email(db_conn,id,False)
+
+    except sqlalchemy.exc.ProgrammingError as err:
+        logger.error(err)
+        raise DatabaseException("Couldn't unpause user: %s" % str(err))
+
+
+def _notify_user_email(db_conn, user_id,paused):
```
amCap1712, line 620: Rename this to something like _notify_user_paused?

listenbrainz/webserver/views/api.py
```diff
@@ -76,6 +76,9 @@ def submit_listen():
     if mb_engine and current_app.config["REJECT_LISTENS_WITHOUT_USER_EMAIL"] and not user["email"]:
         raise APIUnauthorized(REJECT_LISTENS_WITHOUT_EMAIL_ERROR)
 
+    if user['is_paused']:
```
amCap1712, line 79: I think instead of user_id, the error should say the user account. also same for the error message in the other two api files.

amCap1712 (approved):

PR #3164, opened 2025-02-03
More granular incremental stats

amCap1712:
Currently, the entirety of incremental dumps is checked to filter entities whose stats/popularity needs to be recomputed. This can be further optimized by storing the latest created timestamp of the listens in incremental dumps when a stat is
run. The next time a stat is run, only listens with a higher created value (added through newer incremental dumps) are considered for the filter.

Note that the incremental aggregate is still computed from all of incremental listens dump, only the filter is made more granular.

### DIFF
--- a/listenbrainz_spark/popularity/listens.py
+++ b/listenbrainz_spark/popularity/listens.py
@@ -24,8 +24,8 @@ class PopularityProvider(QueryProvider):
     def get_base_path(self) -> str:
         return LISTENBRAINZ_POPULARITY_DIRECTORY
 
-    def get_filter_aggregate_query(self, existing_aggregate: str, incremental_aggregate: str,
-                                   existing_created: Optional[datetime]) -> str:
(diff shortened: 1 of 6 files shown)


listenbrainz_spark/popularity/listens.py
```diff
@@ -91,11 +94,13 @@ def get_table_prefix(self) -> str:
     def get_base_path(self) -> str:
         return LISTENBRAINZ_POPULARITY_DIRECTORY
 
-    def get_filter_aggregate_query(self, existing_aggregate: str, incremental_aggregate: str) -> str:
+    def get_filter_aggregate_query_coarse(self, existing_aggregate: str, incremental_aggregate: str,
```
mayhem, line 97: this function has the same name as the one above. Intended?

mayhem (approved):
Once the question about the duplicated function is resolved, good to go.

listenbrainz_spark/popularity/listens.py
amCap1712: There are two classes in this file, so once in each.

PR #3166, opened 2025-02-07
LB-1742 Notify user if they're adding a duplicate to a playlist

fettuccinae:
# Problem
  
  When a user adds a track which is already present in the playlist, there is no indication about it. 
  https://tickets.metabrainz.org/browse/LB-1742

# Solution

  Added logic to check if the track is already present in the playlist, and if it is then render DuplicateTrackModal.


fettuccinae:
Duplicate model UI
image

MonkeyDo (commented):
Looks good!

If you want to make all users happy, adding a "don't ask me again until I close this page" checkbox to the modal would ensure you don't make it miserable for creative users that want the same track 30 times in a playlist (not that I know [anyone who would do that...](https://listenbrainz.org/playlist/562a53e7-f2ca-4bfd-abb0-bfb7c875016b/)).

Than you can save that boolean into the playlist page state, and skip the confirmation or not accordingly. Probably ideally passing the saved state back to the modal component to allow users to change their mind by unticking the checkbox.

fettuccinae:
Thanks!
I didnt opt in for toast message because i wanted it to be visually similar to ytm and spotify . 
btw you should checkout this song from [Family Guy](https://www.youtube.com/watch?v=uSlB4eznXoA)

frontend/js/src/playlists/components/DuplicateTrackModal.tsx
```diff
@@ -0,0 +1,90 @@
+import * as React from "react";
+import NiceModal, { useModal } from "@ebay/nice-modal-react";
+
+type DuplicateTrackModalProps = {
+  message: string;
+  dontAskAgain: boolean;
+  setDontAskAgain: (value: boolean) => void;
```
MonkeyDo, line 7: Instead of passing the setDontAskAgain down to the child component, let's instead make this modal return both booleans: the confirmation bool you already have, and also now the localDontAskAgain value.
Then in Playlist.tsx you can expect the modal to return two values (will leave a comment on the appropriate lines)

frontend/js/src/playlists/Playlist.tsx
```diff
@@ -220,6 +222,25 @@ export default function PlaylistPage() {
...
+      if (isDuplicate && !dontAskAgain) {
+        const confirmed = await NiceModal.show(DuplicateTrackModal, {
+          message:
+            "This track is already in the playlist. Do you want to add it anyway?",
+          dontAskAgain,
+          setDontAskAgain,
+        });
+        if (!confirmed) {
+          return;
+        }
```
MonkeyDo, line 241: Along with my other comment below, I suggest a couple of changes:
1. Get back two values from the modal promise
2. Set an expected return type for the modal
3. Don't pass `setDontAskAgain` down, but instead call it with the new returned boolean value (keeping states more separate between components)
```suggestion
        const [confirmed, dontAskAgainValue] = await NiceModal.show<
          [boolean, boolean],
          any
        >(DuplicateTrackModal, {
          message:
            "This track is already in the playlist. Do you want to add it anyway?",
          dontAskAgain
        });
        setDontAskAgain(dontAskAgainValue);
        if (!confirmed) {
          return;
        }
```

frontend/js/src/playlists/components/DuplicateTrackModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const { message, dontAskAgain: initialDontAskAgain, setDontAskAgain } = props;
+
+  // Tried to use dontAskAgain directly, but the checkbox state wasnt updating when toggled.
+  const [localDontAskAgain, setLocalDontAskAgain] = React.useState(
+    initialDontAskAgain
+  );
+
+  const confirm = () => {
+    setDontAskAgain(localDontAskAgain);
+    modal.resolve(true);
```
MonkeyDo, line 21: This will become:
```suggestion
    modal.resolve(true, localDontAskAgain);
```

Same change for `cancel` below

frontend/js/src/playlists/components/DuplicateTrackModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+};
+
+export default NiceModal.create((props: DuplicateTrackModalProps) => {
+  const modal = useModal();
+  const { message, dontAskAgain: initialDontAskAgain, setDontAskAgain } = props;
+
+  // Tried to use dontAskAgain directly, but the checkbox state wasnt updating when toggled.
+  const [localDontAskAgain, setLocalDontAskAgain] = React.useState(
+    initialDontAskAgain
+  );
```
MonkeyDo, line 17: Makes sense that it would not change the controlled state, because ticking the checkbox did not call `setDontAskAgain`, it was only called after confirming or cancelling.

There is another way to do this using an uncontrolled checkbox (meaning we don't save its state), but I do not recommend it as I think keeping a local state like this works fine.
For more context read this: https://react.dev/reference/react-dom/components/input#my-checkbox-doesnt-update-when-i-click-on-it

However I think you can remove the comment.

MonkeyDo (changes requested):
Looking good, thanks for implementing those improvements. For users who regularly add duplicates this will greatly reduce annoyance :)

I do have some suggestions below for state management between components. Always an interesting topic, and there are so many ways to do it!

frontend/js/src/playlists/components/DuplicateTrackModal.tsx
```diff
@@ -4,27 +4,23 @@ import NiceModal, { useModal } from "@ebay/nice-modal-react";
...
-    setDontAskAgain(localDontAskAgain);
-    modal.resolve(true);
+    modal.resolve([true, localDontAskAgain]);
     modal.remove();
   };
 
   const cancel = () => {
-    setDontAskAgain(localDontAskAgain);
-    modal.resolve(false);
+    modal.resolve([false, localDontAskAgain]);
```
MonkeyDo, line 23: I think this will need to return `false` for the dontAskAgain return value, because otherwise if I tick the checkbox and click cancel I won't be asked again despite having said "don't add this track".
I suppose I expect that if I clicked cancel I would indeed be asked again regardless of the ticked boxed.

MonkeyDo (changes requested):
Looking great, nice job!

I did a bit of testing and I really love the result.

I do have one last small change after some more thorough testing, but that's the last thing from me, after that it is ready to merge as far as I am concerned.

frontend/js/src/playlists/components/DuplicateTrackModal.tsx
fettuccinae: Ohh yes, that was an oversight, sorry about that.

PR #3167, opened 2025-02-07
LB-1103: Only show front cover art in ListenCards

MonkeyDo:
Requested by @Aerozol, for cases where the user has provided a release MBID, but there is no front image for that release  (for example a vinyl release with liner notes added but no front cover: https://musicbrainz.org/release/ce5deb72-4bc3-461f-b92e-1637709f70ef/cover-art).

In that case, fallback to another method to get a front cover (from the release group, finds a front image if there is one), in this case: https://musicbrainz.org/release/62d9695f-ac89-474c-b9c8-ee95828e2e58/cover-art

I kept the rest of the mechanism as untouched as possible, the new option is, well, optional, and defaults to the previous logic.

**Before:**
image

**After:**
image


anshg1214 (approved):
The PR looks good! But, I'm not sure on how to test it.

PR #3169, opened 2025-02-10
fix: Playlist Cover Art not updating

anshg1214:
This PR fixes:
* Whenever we're adding / removing tracks from the playlist, we now re-validate the data again through route loaders to update the cover art.
* Fix the Playlist Description css.


MonkeyDo (approved):
Yus!

PR #3170, opened 2025-02-11
LB-1737: Create top artists graph showing album details

granth23:
# Problem

[LB-1737](https://tickets.metabrainz.org/projects/LB/issues/LB-1737?filter=allopenissues): Create top artists graph showing album details

# Solution

image

Using the previous data fetch `api` with some modifications and using `nivo.rocks` for the same to be displayed as a stacked/segmented bar graph.

# Action


listenbrainz/webserver/views/stats_api.py
```diff
@@ -409,6 +409,69 @@ def get_listening_activity(user_name: str):
...
+    :statuscode 200: Successful query, you have data!
+    :statuscode 204: Statistics for the user haven't been calculated, empty response will be returned
+    :statuscode 400: Bad request, check ``response['error']`` for more details
+    :statuscode 404: User not found
+    :resheader Content-Type: *application/json*
+    """
+    data = _get_entity_stats(user_name, "release_groups", "total_release_group_count")
+    result = {}
+    for release_group in data["payload"]["release_groups"]:
+        artist_name = release_group["artist_name"]
```
anshg1214, line 459: I deployed your PR on test.LB, I got an error saying 
```
  File "/code/listenbrainz/listenbrainz/webserver/views/stats_api.py", line 458, in get_artist_activity
    for release_group in data["payload"]["release_groups"]:
                         ~~~~^^^^^^^^^^^
TypeError: 'Response' object is not subscriptable
```
The problem is that `_get_entity_stats` returns a Flask Response object, but you're trying to access it like a dictionary. Instead you should fetch the stats directly from the DB here, and process it further.

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -0,0 +1,147 @@
...
+      ...artist.albums.reduce(
+        (acc, album) => ({ ...acc, [album.name]: album.listen_count }),
+        {} as Record<string, number>
+      ),
+    })) as ChartDataItem[];
+  };
+
+  type ChartDataItem = {
+    label: string;
+  } & Record<string, number>;
```
anshg1214, line 70: Don't define the type defination inside the react component, instead define it outside the component.

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -0,0 +1,147 @@
...
+
```
anshg1214, line 71: The type can be improved as:

```suggestion
  type ChartDataItem = {
    label: string;
    [albumName: string]: number | string;
  };
```

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -0,0 +1,147 @@
...
+    <Card className="user-stats-card" data-testid="user-artist-activity">
+      <div className="row">
+        <div className="col-xs-10">
+          <h3 className="capitalize-bold" style={{ marginLeft: 20 }}>
+            Artist Activity
+          </h3>
+        </div>
+        <div className="col-xs-2 text-right">
+          <h4 style={{ marginTop: 20 }}>
+            <a href="#artist-activity">
```
anshg1214, line 91: this should be same as the id ussed in the section defined in the UserReports.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -409,6 +409,69 @@ def get_listening_activity(user_name: str):
...
+            result[artist_name]["albums"].append({"name": release_group_name, "listen_count": listen_count})
+        else:
+            result[artist_name] = {
+                "name": artist_name,
+                "listen_count": listen_count,
+                "albums": [{"name": release_group_name, "listen_count": listen_count}]
+            }
+    sorted_data = sorted(result.values(), key=lambda x: x["listen_count"], reverse=True)
+    return jsonify({"result": sorted_data})
+
```
anshg1214, line 474: Currently, the graph groups albums by artists. However, prioritizing top artists first and inserting albums from their data into their bar is the best approach. Since we have both data and album data, merging them will create the optimal graph. If there’s a listen count without album data, mark it as “other.”

For this:

1. First get the top artists data to ensure we have the correct artist ordering and total listen counts
2. Get release group data to get album information
3. Create a lookup of albums by artist name for quick access
4. For each top artist:
   - Get their total listen count from artist data
   - Get their album data from the lookup
   - Calculate if there are any "missing" listens and add an "Other" category if needed
   - Sort albums by listen count

anshg1214: @MonkeyDo What do you think on this approach?

anshg1214 (changes requested):

listenbrainz/webserver/views/stats_api.py
granth23: With the current approach, this is how the filtering is working

1. Get all the albums
2. Map it to artist with the total listen counts of the artist.
3. Sort it and show all albums by the artist.

Let me know if some changes are required.

MonkeyDo: I agree, this should be a view of the top 25* artists for the selected time period, with each artist's bar broken up into separate albums (+ "other" if that does not sum up to the total artist listen count).
So it starts with the top artists as the base, rather than the top albums.

*I say 25 but this could be another number if it's not enough or too little.
Currently for me it's way too many artists to be useful:
image

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -0,0 +1,143 @@
...
+      return [];
+    }
+    return data.result.map((artist) => ({
+      label: artist.name.split(",")[0],
+      ...artist.albums.reduce(
+        (acc, album) => ({ ...acc, [album.name]: album.listen_count }),
+        {} as Record<string, number>
+      ),
+    })) as ChartDataItem[];
+  };
```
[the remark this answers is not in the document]
anshg1214: I believe it'll be best if you could filter the data in the backend before sending. The Graph is causing the website to crash if I select the select the range to Last Month or above.

anshg1214 (changes requested):
The Graph looks good. Apart from the comments, can we change the color scheme of the graph to something using LB Color theme instead? 

Also, it'll be great if we could also generate a similar graph for sitewide stats as well.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -409,6 +409,81 @@ def get_listening_activity(user_name: str):
...
+        release_group_name = release_group["release_group_name"]
+
+        if artist_name in result:
+            result[artist_name]["listen_count"] += listen_count
+            result[artist_name]["albums"].append({"name": release_group_name, "listen_count": listen_count})
+        else:
+            result[artist_name] = {
+                "name": artist_name,
+                "listen_count": listen_count,
+                "albums": [{"name": release_group_name, "listen_count": listen_count}]
```
anshg1214, line 478: I had deployed this PR on test.LB for some time. This is how my graph looks like:

Pasted Graphic

Although you have sorted the data based on listen_count, this is because we have unlinked listens which can be of same album.

```
{
    "listen_count": 20,
    "name": "Bandish Bandits (Original Series Soundtrack)"
},
{
    "listen_count": 16,
    "name": "Bandish Bandits (Original Series Soundtrack)"
},
```

Although they refer to different data, but the nivo bar graph treats them as one, and only shows one of them causing this issue. So, we can merge the listen count into one, if they don't have MBIDs, which then results into the following graph:

Artist Activity

listenbrainz/webserver/views/stats_api.py
```diff
@@ -409,6 +409,81 @@ def get_listening_activity(user_name: str):
...
+    count = get_non_negative_param("count", default=DEFAULT_ITEMS_PER_GET)
+    stats = db_stats.get(user["id"], "release_groups", stats_range, EntityRecord)
+    if stats is None:
+        raise APINoContent('')
+
+    release_group_list, _ = _process_user_entity(stats, offset, count, entire_range=True)
+    
+    result = {}
+    for release_group in release_group_list:
+        artist_name = release_group["artist_name"].split(",")[0]
```
anshg1214, line 467: Is there any particular reason why we're using the first artist name in the graph?

frontend/js/tests/user/stats/UserArtistActivity.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+const sitewideProps: UserArtistActivityProps = {
+  range: "week",
+};
+
+jest.mock("@nivo/bar", () => ({
+  ...jest.requireActual("@nivo/bar"),
+  ResponsiveBar: ({ data }: any) => (
+    <div data-testid="ResponsiveBar">{JSON.stringify(data)}</div>
+  ),
+}));
```
anshg1214, line 29:
```suggestion
  ResponsiveBar: ({ children }: any) => children({ width: 400, height: 400 }),
```

frontend/js/tests/user/stats/UserArtistActivity.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+      <UserArtistActivity {...props} />,
+      {},
+      {
+        wrapper: reactQueryWrapper,
+      }
+    );
+
+    await waitFor(() => {
+      expect(screen.getByTestId("ResponsiveBar")).toBeInTheDocument();
+    });
```
anshg1214, line 101: In this particular test, you should check if the whole component is being loaded or not.
```suggestion
      expect(screen.getByTestId("user-artist-activity")).toBeInTheDocument();
```

Further more, you can add tests to check if the number of bars are correct or not.

anshg1214 (changes requested):
The graph looks awesome now🔥

I noticed a few improvements which can be made.

listenbrainz/webserver/views/stats_api.py
granth23: Many of the albums have the actors in the movie/song as the other artists which doesn't make sense to be present in the actual graph

MonkeyDo: > Many of the albums have the actors in the movie/song as the other artists which doesn't make sense to be present in the actual graph

This does not necessarily apply to all music styles. In a lot of cases, if not most, you will have featuring artists or collaborative artists that we do want to show.

Random example of the latest release added to MusicBrainz at the time of writing (Magnolia Park, PLVTINUM & Vana): https://musicbrainz.org/recording/fb83b76f-def7-4536-bc3c-2f32c2b78ab0

granth23: Have made the necessary changes

MonkeyDo:
The graph is looking very good!

I think the artist names at the bottom can be tweaked to avoid the overlapping:
image

Maybe using the axisBottom tickRotation, tickPadding or truncateTickAt, or some combination of those?
Ideally keeping as much of it readable, but avoiding this overlapping which doesn't look great with longer names.

It's especially a problem in mobile view with less space available:
image

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -0,0 +1,196 @@
...
+                              {line}
+                            </text>
+                          ))}
+                      </g>
+                    ),
+                  }}
+                  onClick={(barData, event) => {
+                    const albumName = barData.id;
+                    const artistName = barData.indexValue;
+                    const releaseMbid =
```
anshg1214, line 182:
```suggestion
                    const releaseGroupMbid =
```

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -0,0 +1,196 @@
...
+                      albumRedirectMapping[`${artistName}-${albumName}`];
+                    if (releaseMbid) {
+                      window.location.href = `/album/${releaseMbid}`;
+                    }
```
anshg1214, line 186: Since we're using react router, you should use [useNavigate](https://reactrouter.com/6.30.0/hooks/use-navigate#usenavigate).
```suggestion
                      navigate(`/album/${releaseMbid}`);
```

anshg1214 (changes requested):

listenbrainz/webserver/views/stats_api.py
```diff
@@ -483,14 +483,377 @@ def get_artist_activity(user_name: str):
...
-    release_groups_list, _ = _process_user_entity(stats, offset, count, entire_range=True)
+    # user, stats_range = _validate_stats_user_params(user_name)
+    # offset = get_non_negative_param("offset", default=0)
+    # count = get_non_negative_param("count", default=DEFAULT_ITEMS_PER_GET)
+    # stats = db_stats.get(user["id"], "release_groups", stats_range, EntityRecord)
+    # if stats is None:
+    #     raise APINoContent('')
+
+    # release_groups_list, _ = _process_user_entity(stats, offset, count, entire_range=True)
+    temp_data = {
```
MonkeyDo: This should not have been committed. 
It's totally fine to do on your machine, but please check every change that you commit before pressing the button.

anshg1214 (approved):
LG!

Maxr1998:
The new artist activity widget is really cool! But I noticed a small issue with it, it seems that it used the artist name from the recording (listen?), and not the canonical artist name from MusicBrainz to group artists. Thus, I have the same artist twice with different names (one in Latin and one in Hangul).

For what it's worth, my top artists list correctly lists that artist just once.

MonkeyDo:
@Maxr1998 Thanks for reporting the issue!
It's not specifically associated with this new graph, but I think a more general problem with statistics.

Could I ask you to please open a new ticket in https://tickets.metabrainz.org/projects/LB with a screenshot of the stats?

Could you also please check that your top artists for that time range (https://listenbrainz.org/user/Maxr1998/stats/top-artists) whos the same issue, maybe circling or describing the duplicate artist issue?
(This is to ensure that it's a general stats issue rather than linked to the graph)

Maxr1998:
Sure, opened a ticket as LB-1778. It appears to be specific to this new graph, though, since my top artists stats are correct.

PR #3171, opened 2025-02-11
LB-1748: Add playlist cover art on the playlist overview page

MonkeyDo:
With #3149 and the new art API endpoint to get playlist cover art images, we can now show the image on the playlists page.
On hover the title section at the bottom of the card expands and both it and the background image become more opaque.
I lowered the initial opacity because the was a bit overwhelming.

What do we think, is it still too busy?
image

### DIFF
--- a/frontend/css/playlists.less
+++ b/frontend/css/playlists.less
@@ -84,24 +84,6 @@
       width: 100%;
       background-color: @playlistBackgroundColor;
     }
-    @media (hover: hover) {
-      .info {
-        bottom: -60px;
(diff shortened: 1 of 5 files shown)


MonkeyDo:
Neither Ansh nor I were impressed by the look of the page with the cover art mosaics added.
It looks very busy, and makes it look like the page loads very slowly (each cover art in each mosaic loads separately).

Overall not clear if it improves the UX, so closing the PR until we have  a better idea for integrating playlist cover art.

PR #3172, opened 2025-02-12
Fix incremental listener stats

amCap1712:
The latest iteration of incremental stats filters listens based on created timestamp instead of using the entire incremental aggregate, hence the filtering queries used for user stats are no longer valid for entity stats and need to be updated.


mayhem (approved):
LGTM!

PR #3175, opened 2025-02-12
Add disambiguation to track search results

MonkeyDo:
Currently in the track search page, in some cases multiple recordings with the same name will be shown.
Without a disambiguation to look at and compare, this is very confusing.

This PR adds disambiguation (and track length, which was missing) next to the recording title to make it more useful.

Before:
image

After:
image


anshg1214 (approved):
LG!

PR #3176, opened 2025-02-13
Migrate artist map stats to Spark

amCap1712:
Currently artist map stats are generated on the fly in LB webserver and cached in couchdb. This worked well when all the stats were regenerated daily and the dbs were rotated daily as well. However in the new incremental stats regime, stats are only updated when necessary which makes it difficult to figure out when the cached artist map stats are stale. To make it simple, move the artist map stats to spark as well and re-use the artist stats aggregates to generate them ensuring that both are updated at the same time.


mayhem (approved):
Thanks for cranking all these massive changes out!

PR #3177, opened 2025-02-13
LB-1750 Added Show more button to playlist description

Suvid-Singhal:
# Problem

Earlier, the playlist description was not visible if it was very long. 

JIRA Ticket: [LB-1750](https://tickets.metabrainz.org/browse/LB-1750)

# Solution

Fixed it by adding a button to view the full description.

image

image


frontend/css/playlists.less
```diff
@@ -277,3 +278,25 @@
     box-shadow: 0 0 0 2px rgba(53, 48, 112, 0.5);
   }
 }
+
+.wikipedia-extract {
```
MonkeyDo, line 282: I think we should rename this class, it made sense before but now that we use it in the playlist page it should really be renamed.
Obviously not something you added yourself, but since we're in here modifying this specific piece of css let's improve the naming.

I suggest `text-extract` or `text-summary` instead, but open to suggestions if you have a better idea.
You'll need to do a search+replace all and change every instance.

frontend/css/playlists.less
```diff
@@ -1,5 +1,6 @@
 @playlistBackgroundColor: #fafafa;
 @descriptionLines: 3;
```
MonkeyDo, line 2: We already have some CSS regarding cropping the description. Perhaps we can simply reuse the existing code and change the value of `descriptionLines`?

frontend/css/playlists.less
```diff
@@ -277,3 +278,25 @@
     box-shadow: 0 0 0 2px rgba(53, 48, 112, 0.5);
   }
 }
+
+.wikipedia-extract {
+  .playlist-description {
```
MonkeyDo, line 283: Instead of creating a new class that repeats some code, let's refactor the existing `.description` class so that it can be used in the playlist page as well (currently it is only as a child of `#playlists-container` , used on the playlists overview page)

frontend/js/src/playlists/Playlist.tsx
```diff
@@ -445,12 +450,25 @@ export default function PlaylistPage() {
           {playlist.annotation && (
             <div className="wikipedia-extract">
               <div
-                className="content"
+                className={`content ${showMore ? "playlist-description" : ""}`}
```
MonkeyDo, line 453: With my comments above about reusing the existing class, this should be something like 
```suggestion
                className={`description ${showMore ? "expanded" : ""}`}
```

frontend/css/playlists.less
```diff
@@ -277,3 +278,25 @@
...
+    line-height: 1.2em;
+    max-height: @playlistDescriptionLines * 1.2em;
+    text-align: justify;
+    padding-right: 2em;
+    display: -webkit-box;
+    line-clamp: @playlistDescriptionLines;
+    margin-top: 0.5rem;
+    -webkit-line-clamp: @playlistDescriptionLines;
+    -webkit-box-orient: vertical;
+  }
```
MonkeyDo, line 295: For what happens after clicking the show-more button, you probably want a `&.expanded` css definition that removes the line-clamp and overflow styles.
Either that or the other way around, having expanded be the standard style and use `&:not(.expanded)` to set the line-clamping and overflow.

frontend/css/playlists.less
```diff
@@ -277,3 +278,25 @@
...
+
+  .show-more-button {
```
MonkeyDo, line 297: I'm not sure we need custom style here. Instead you should be able to use the classes `btn btn-link pull-right` on the button element like we do for the artist page wikipedia extract

MonkeyDo (changes requested):
Good addition!
Thanks for working on this.
Couple of suggestions below

frontend/css/playlists.less
Suvid-Singhal: I am using almost the same CSS code for cropping the description as used in the class "description" in this file.
I defined a new variable as I needed more lines to show before hiding the description.

I'm sorry I don't understand what do you mean by reusing the code?

Suvid-Singhal:
Hi @MonkeyDo,
I have made the required changes, and everything is working fine.
Do let me know if any other changes or improvements in coding practices are needed 😊

MonkeyDo (changes requested):
I hope you don't mind, I had a few suggestions for improvements but I didn't manage to explain it all in detail in a way that would be easy to follow. I realize my previous comments were already ambiguous and hard to follow...
To avoid more back and forth I implemented it directly.

What I did:
- refactored the .text-summary class to avoid unnecessary nesting
- added a fade effect at the bottom
- hide the "show more" button if there is little text
- flipped the logic of the `showMore` state, the naming of it suggested it should be false at first and true after clicking the button
- looked into merging the styles for .wikipedia-extract and .text-summary , but that requires more work and we'll keep that for another day, another PR.



With that out of the way, I think this is ready to merge!

MonkeyDo:
Sorry, misclick!

MonkeyDo (approved):
Looks good, works good ! :)

PR #3178, opened 2025-02-14
Cache incremental listens and metadata tables in Spark

amCap1712:
Persisting these not so big and commonly used datasets with Spark creates a copy of them in each node's memory and/or disk which speeds up processing of stats.


mayhem (approved):
Looks good!

PR #3179, opened 2025-02-14
LB-1735: Automatically switch search type depending on pasted MB URL

MonkeyDo:
Automatically switches between components to add a recording and add a release if the pasted MusicBrainz URL is of the other entity type.
So for example if you paste a release or release group MBID in the single track search, it will automatically switch to the "Add album" mode, copying the pasted URL over and triggering the search.

The components are so split up (to allow re-usability across the codebase) that this requires a bit of extra fuckery to shoehorn this feature in (requires both switching components and triggering actions on said un-rendered component, requiring delays before triggers etc.)


anshg1214 (approved):
LG!

PR #3180, opened 2025-02-14
Use zstd compression for parquet dumps

amCap1712:
Spark/parquet dumps have been growing a lot in size lately. Having tried multiple failed ways to reduce the size while keeping the existing snapping compression switching to zstd compression which seems to be almost twice as efficient on the test set.


mayhem (approved):
And there is a better alg coming still! https://github.com/kspalaiologos/bzip3 I would love to see how this does on our data.

PR #3183, opened 2025-02-17
feat: Add Recordings page

anshg1214:
This PR adds the recording page

image


anshg1214:
@MonkeyDo Should we move this page to `/track/<recording_mbid>` or keep it on `/recording/<recording_mbid>`?

mayhem:
> @MonkeyDo Should we move this page to `/track/<recording_mbid>` or keep it on `/recording/<recording_mbid>`?

Whichever is decided, please add a redirect for the other.

MonkeyDo:
> @MonkeyDo Should we move this page to `/track/<recording_mbid>` or keep it on `/recording/<recording_mbid>`?

Definitely /track as we do for the other entity types, as mentioned with a redirect also like we do for /release and such

frontend/js/src/track/Track.tsx
```diff
@@ -69,7 +69,7 @@ export default function TrackPage(): JSX.Element {
     RouteQuery(["track", params], location.pathname)
   );
 
-  const { recording, similarRecordings, releaseGroups } =
```
MonkeyDo: I think this would be the type of variable we could argue keeping the name 'recording' makes more sense.
We do want to say 'track' for every user-facing name, but in this case if we get recording metadata from MB it's clearer to keep 'recording', considering MB also has 'tracks' as a different entity type.
But then again, we won't be using MB tracks in this context, so maybe it's fine.

frontend/js/src/track/components/SimilarTracks.tsx
```diff
@@ -1,16 +1,16 @@
...
-import generateTransformedRecordings from "../utils/generateTransformedTracks";
+import SimilarTracksGraph from "./SimilarTracksGraph";
+import generateTransformedTracks from "../utils/generateTransformedTracks";
 
-type SimilarRecordingProps = {
-  onRecordingChange: (recording_mbid: string) => void;
-  recordingGraphNodeInfo: TrackNodeInfo | undefined;
-  similarRecordingsList: TrackNodeInfo[];
+type SimilarTracksProps = {
+  onTrackChange: (track_mbid: string) => void;
```
MonkeyDo: Here I think this is actually dangerous to rename. The more technical the more a developer is likely to interpret this as a musicbrainz track MBID instead of the recording MBID it is.

frontend/js/src/track/Track.tsx
```diff
@@ -0,0 +1,398 @@
...
+    return <div>Loading...</div>;
+  }
+
+  const formattedDateWithHours = new Date(length)
+    .toISOString()
+    .substring(11, 19);
+  // Hide the hours if it's 0
+  const formattedDate = formattedDateWithHours.startsWith("00:")
+    ? formattedDateWithHours.substring(3)
+    : formattedDateWithHours;
```
MonkeyDo: Should probably rename these variables, date > duration

listenbrainz/webserver/views/entity_pages.py
```diff
@@ -338,3 +341,110 @@ def album_entity(release_group_mbid: str):
...
+
+
+@recording_bp.get("/",  defaults={'path': ''})
+@recording_bp.get('/<path:path>/')
+def recording_redirect(path):
+    return render_template("index.html")
+
+
+@track_bp.route("/",  defaults={'path': ''})
+@track_bp.get('/<recording_mbid>/')
```
MonkeyDo: Is it expected to defined both .route and .get?

listenbrainz/webserver/views/entity_pages.py
```diff
@@ -338,3 +341,110 @@ def album_entity(release_group_mbid: str):
...
+                "image": f'https://coverartarchive.org/release-group/{release_group_mbid}/front-500',
+                "image:width": "500",
+                "image:alt": f"Cover art for {recording_name}",
+                "url": f'{current_app.config["SERVER_ROOT_URL"]}/track/{recording_mbid}',
+            }
+
+    return render_template("index.html", og_meta_tags=og_meta_tags)
+
+
+@track_bp.route("/<recording_mbid>/", methods=["POST"])
```
MonkeyDo:
```suggestion
@track_bp.post("/<recording_mbid>/")
```

mbid_mapping/mapping/mb_metadata_cache.py
```diff
@@ -375,6 +383,10 @@ def get_metadata_cache_query(self, with_values=False):
                               FROM musicbrainz.recording r
                               JOIN musicbrainz.artist_credit ac
                                 ON r.artist_credit = ac.id
+                         LEFT JOIN recording_first_release_date rfdr
```
MonkeyDo, line 386: If we have access to this recording_first_release_date, we could send it to the front-end and display it.

MonkeyDo (commented):
I think this will do nicely for a V1 ! Nice work, it's great to see a recording page on the website!

A couple of comments, not sure which you want to address in this PR or at a later date:


- if we have the "first release year" for the recording, we can display it under the duration or something like that (maybe little icons in front? clock for duration, calendar for release date?)
- i am not convinced by the similar tracks being presented as a graph like the similar artists. I think the artist name at the very least is missing to make it a useful feature. Personally, I would make it a grid of listencards like we do on the [top-tracks page](https://listenbrainz.org/user/mr_monkey/stats/top-tracks/?range=week). It would also make the tracks playable without needing to navigate to their respective pages.
- as we discussed, it'll be interesting to add more data from MB to that page, but let's do that as a V2 and get this PR out the door.

listenbrainz/webserver/views/entity_pages.py
anshg1214: It should be .get for both. Good catch

mbid_mapping/mapping/mb_metadata_cache.py
anshg1214: I'll do it in a separate PR because this information will only be populated when the cache has this field. I don't have anyway to test it out.

MonkeyDo (approved):
Deployed again to beta.LB, looking good to me!

I think that's a perfectly fine v1 going out. Haven't had much feedback from the matrix channel, so let's get this out the door :)

MonkeyDo (changes requested):
Actually, I changed my mind sorry XD

There is one crucial thing missing on this page, and I was so focused on the details that I missed it: there is no way to play the track in question !

The addition of a big play button somewhere is absolutely required.
For a quick and easy solution mockup, I copied the button we have on album pages, with a few more css classes: `btn btn-info btn-rounded align-self-baseline fs-3 mt-auto` the last 3 classes are added for sizing and placement.

Image

If you have the patience, we could also add a big play icon when hovering over the cover art, and make it play on click.
Happy for that to be part of v2, though.

PR #3188, opened 2025-02-19
Add a table to record listen history deletion

amCap1712:
It is possible for users' to delete their entire listen history in one command, instead of individually deleting every listen. To handle this spark, record the user_id and the max created value until which listens have been deleted.


mayhem (approved):
:shipit:

PR #3189, opened 2025-02-20
Optimize fresh releases generation

amCap1712:
array_contains is very slow compared to exploding artist_mbids column on both tables and joining on artist_mbid. reduced processing time from 20mins to 5 mins.


mayhem (approved):
less code, faster runtime. WIN!

PR #3193, opened 2025-02-20
Implement listen deletion in the spark cluster

amCap1712:
As the first step to eliminate full dumps from the spark cluster pipeline, implement listen deletion by importing metadata about deleted individual listens and entire user listen histories from LB database. This metadata is stored in HDFS and the listens loaded for stats generation are filtered using the two metadata files to exclude listens that have been deleted from ListenBrainz.

Note that the actual listens are not removed from the Spark, they are just filtered out at stats generation time. Eventually the deleted listens and user listen histories will get cluttered up though and therefore a future PR will implement the logic to periodically remove the deleted listens from HDFS as well.

### DIFF
--- a/listenbrainz/spark/request_manage.py
+++ b/listenbrainz/spark/request_manage.py
@@ -530,6 +530,12 @@ def request_troi_playlists():
     send_request_to_spark_cluster("tags.default")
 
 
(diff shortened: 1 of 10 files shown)


mayhem (approved):
:ship:

PR #3195, opened 2025-02-22
LB-1755: Fix Feed event deletion

anshg1214:
LB-1755


MonkeyDo (approved):
Yes, thanks for fixing that!

I do like adding the expected types to all these react-query functions, it's helped me will a similar hard-to-find issue before.w

PR #3199, opened 2025-02-26
Change dumps sequencing and misc test improvements

amCap1712:
Currently, all dumps use the previous dumps' created timestamp as the start time regardless of whether it is a full dump or an incremental dump. This is problematic if we want to eliminate the use of full dumps and just rely on incremental dumps. For example,

|                    | start               | end (created)       |
|--------------------|---------------------|---------------------|
| Incremental Dump 1 | 2025-01-30 12:00:01 | 2025-01-31 12:00:02 |
| Full Dump 1        | `LISTEN_MINIMUM_DATE` | 2025-02-01 04:00:05 |
| Incremental Dump 2 | 2025-02-01 04:00:05 | 2025-02-02 12:00:02 |

In this case if we were to just use incremental dumps, we would miss the listens created between 2025-01-31 12:00:02 and 2025-02-01 04:00:05. To fix this issue, update the start time of incremental dumps to use the end time of the last incremental dump only. As currently it is not possible to identify from the data_dump table whether a dump is incremental or full, update it to add dump_type column.

Further, to ensure that full dumps don't contain duplicates with incremental dumps add a where clause to full dumps query with a created check in addition to listened_at.

Additionally, while diagnosing test failures made changes to the code to use datetime aware objects everywhere. If dumps related tests fail silent and involve invocation of click CliRunner, pass in catch_exceptions = False to debug.

### DIFF
--- a/listenbrainz/listenstore/dump_listenstore.py
+++ b/listenbrainz/listenstore/dump_listenstore.py
@@ -4,7 +4,7 @@ import subprocess
 import tarfile
 import time
 import uuid
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 
 import pandas as pd
 import psycopg2
@@ -17,7 +17,7 @@ import orjson
 from psycopg2.extras import execute_values
 from sqlalchemy import text
 
-from listenbrainz import DUMP_LICENSE_FILE_PATH, db
+from listenbrainz import DUMP_LICENSE_FILE_PATH
 from listenbrainz.db import DUMP_DEFAULT_THREAD_COUNT
 from listenbrainz.db import timescale
 from listenbrainz.db.user import get_all_usernames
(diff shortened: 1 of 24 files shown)


listenbrainz/listenstore/dump_listenstore.py
```diff
@@ -248,43 +241,37 @@ def dump_listens(self, location, dump_id, start_time=datetime.utcfromtimestamp(0
...
-        archive_path = os.path.join(
-            location, '{filename}.tar.xz'.format(filename=archive_name))
+        archive_path = os.path.join(location, f'{archive_name}.tar.xz')
         with open(archive_path, 'w') as archive:
-
-            xz_command = ['xz', '--compress',
-                           '-T{threads}'.format(threads=threads)]
-            xz = subprocess.Popen(
-                xz_command, stdin=subprocess.PIPE, stdout=archive)
+            xz_command = ['xz', '--compress', f'-T{threads}']
```
mayhem, line 259: Should we move to zstd here as well?

mayhem (approved):
LGTM

listenbrainz/listenstore/dump_listenstore.py
amCap1712: Yup sure.

PR #3205, opened 2025-02-27
feat: Show CB Review form on Entity Page

anshg1214:
This is how the new CB Review form looks like.

image

### DIFF
--- a/frontend/js/src/album/AlbumPage.tsx
+++ b/frontend/js/src/album/AlbumPage.tsx
@@ -13,7 +13,6 @@ import tinycolor from "tinycolor2";
 import { Helmet } from "react-helmet";
 import { useQuery } from "@tanstack/react-query";
 import { Link, useLocation, useParams } from "react-router-dom";
-import NiceModal from "@ebay/nice-modal-react";
 import {
   getRelIconLink,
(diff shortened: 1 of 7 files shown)


frontend/js/src/album/AlbumPage.tsx
```diff
@@ -520,33 +519,19 @@ export default function AlbumPage(): JSX.Element {
...
-                  },
-                  {
-                    type: "artist",
-                    mbid: artist.artists[0].artist_mbid,
-                    name: artist.artists[0].name,
-                  },
-                ],
-              });
+          ) : null}
+          <CBReview
```
MonkeyDo, line 523: I think it would look better if we had reviews to one side and the form to the other side, something like this:
image
Using bootstrap columns I guess.
I had to muck about with the right-hand side reviews margin a bit to remove the top margin so it all aligns.

And for mobile view, I guess something like that? 
image

anshg1214: Cool! This is how it looks right now:

image

And on mobile view: 
image

MonkeyDo (approved):
Yes sir!
Good to go! :rocket:

PR #3212, opened 2025-03-04
Update dumps cron jobs

amCap1712:
1. Remove the full dumps import cron job.
2. Separate dumps lock for all types of dumps.
3. Add cron job to import deleted listens and compact listens in spark.


mayhem (approved):
Now we need to get these cron jobs monitored -- we should talk about the sentry migration soon!

PR #3215, opened 2025-03-06
Migrate from xz to zstd for dumps compression

amCap1712:
Tested on jsonl statistics dumps which are 43G uncompressed, the results are as follows:

|             | xz     | zstd (-10) |
|-------------|--------|------------|
| Compression | 8.3x   | 6.7x       |
| Time Taken  | 87m 3s | 33m 12s    |

xz compresses a little better but zstd is much faster, and I feel the overall tradeoffs are in favour of zstd.


mayhem:
way in favor of zstd!

mayhem (approved):

PR #3216, opened 2025-03-06
Bootstrap 4: Update Styling for Feed Page, Dashboard, Settings and About Page

anshg1214:


frontend/css/playlists.less
```diff
@@ -61,8 +60,10 @@
       bottom: 0;
       border: none;
       padding: 5px;
-      &:extend(.btn-block);
-      &:extend(.btn-info);
```
MonkeyDo, line 65: 100% happy to replace the extends, I think we shouldn't have used them

frontend/css/scaffolding.less
```diff
@@ -0,0 +1,132 @@
...
+body {
+  font-family: Sintony, sans-serif;
+  font-size: 14px;
+  line-height: 1.42857143;
+  color: #46433a;
+  background-color: #fff;
+}
+
+a {
+  color: #353070;
```
MonkeyDo, line 15: Be good to use variables for all the colors in this page, if we have them somewhere as variables.
And if we don;t have a variable but they are copied in other places, let's create a variable for it

frontend/css/theme/bootstrap/utilities.less
```diff
@@ -27,9 +27,6 @@
 .hide {
   display: none !important;
 }
-.show {
-  display: block !important;
-}
```
MonkeyDo, line 32: Is this definitely safe to remove?

frontend/js/src/components/Navbar.tsx
```diff
@@ -35,18 +35,15 @@ function Navbar() {
 
   return (
     <nav role="navigation">
-      <div className="navbar-header">
+      <div className="navbar-light navbar-header">
         <button
           type="button"
-          className="navbar-toggle collapsed"
+          className="navbar-toggler"
```
MonkeyDo, line 41: The toggler styles need to be reviewed as they currently make it pretty small on mobile devices compared to existing website.

Also the toggler button seems to be in fixed position (scroll the page down, it stays on screen) which wasn't the case before

frontend/css/music-services.less
```diff
@@ -1,5 +1,18 @@
...
 
+#user-profile {
+  .card {
+    margin-bottom: 2rem;
+    .card-header {
+      padding: 1rem 1.5rem;
+      .card-title {
+        font-size: 16px;
+        margin: 0;
+      }
```
MonkeyDo, line 11: Would be worth investigating if this needs to be changed for the whole website or only applies to this component

frontend/css/listens-page.less
```diff
@@ -66,6 +66,12 @@
     }
   }
 
+  .dropdown-toggle {
+    &::after {
+      display: none !important;
+    }
+  }
```
MonkeyDo, line 73: Would it make sense to make this a more general style in order to replicate our current design?
For example on the playlist page the caret is now visible:
image


I suppose there are good reasons for BS to want to show carets for all dropdowns, so a decision needs to be made, but in general i would aim to replicate our current look and only afterwards evaluate if we want this change

frontend/js/src/playlists/components/PlaylistMenu.tsx
```diff
@@ -224,114 +224,108 @@ function PlaylistMenu({
   );
   const showAppleMusicExportButton = appleAuth;
   return (
-    <ul
+    <div
```
MonkeyDo, line 227: Ahhhhh, not requiring a specific html element structure anymore?
That does put a smile on my face...

frontend/js/src/playlists/components/PlaylistMenu.tsx
```diff
@@ -224,114 +224,108 @@ function PlaylistMenu({
...
-            >
-              <FontAwesomeIcon icon={faTrashAlt as IconProp} /> Delete
-            </a>
-          </li>
-        </>
-      )}
-      <li role="separator" className="divider" />
-      {showSpotifyExportButton && (
-        <li>
+          <div role="separator" className="dropdown-divider" />
```
MonkeyDo, line 242: The css for those separators needs tweaking, they look invisible at the moment:
image

Border color should be `#e5e5e5` to replicate the existing design

frontend/js/src/user/layout.tsx
```diff
@@ -15,9 +15,13 @@ function NavItem({
 }) {
   return (
     <li
-      className={`${isActive ? "active" : ""} ${isDisabled ? "disabled" : ""}`}
+      className={`nav-item ${isActive ? "active" : ""} ${
+        isDisabled ? "disabled" : ""
+      }`}
     >
-      <NavLink to={url}>{label}</NavLink>
+      <NavLink end className="nav-link" to={url}>
```
MonkeyDo, line 22: Can you explain the addition of `end` here?
Not sure what it does

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
MonkeyDo, line 1: Might want to review the style of the similar users component:
image

frontend/js/src/user-feed/UserFeedLayout.tsx
MonkeyDo, line 1: Not sure where to comment this, so putting it here:

The Feed page could use another pass.
Two things I noticed:

Play all button wraps: 
image


Not sure how but the event hide button now has a .dropdown-item class which jumbles everything up
image

MonkeyDo (changes requested):
Awesome first pass, I might have lots of comments below but it all seems like small details, for the most part.

I am listing here discrepancies I saw that would need to be fixed in the lobes4 package/repo. I wonder if the easiest method would be to add a sass file in LB-server repo with all the overwrites we need, then evaluate which ones we want to move over to lobes4 at the end.

- defaults gutter changed from 15px to 21px
- .btn-link text color should be LB blue, border color should be transparent
- .text-muted class doesn't exist anymore, I propose to add it back as it is used in many many places (`color: #8d8d8d;`)
- same for .help-block class, we might want it back as there are 11 uses
- .btn-info color should be LB blue (maybe can be set more generally to all -info colors?)
- there is a style that overrides links colors which we probably want to remove: `a:visited:not(.btn):not(.nav-link):not(.dropdown-item)`

frontend/css/scaffolding.less
anshg1214: Okay. BTW This is a temporary file which I've added right now. Eventually all these configurations needs to be moved to lobes

frontend/js/src/user/layout.tsx
anshg1214: We only need to set the class as active when we get the exact match of the endpoint : https://reactrouter.com/6.30.0/components/nav-link#end

MonkeyDo: I see, thanks for the explanation.
Was that just missing previously?

anshg1214: Yes, this was missing before, but since we had <li> tag before, the class difference was not visible.

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
MonkeyDo: Maybe it is dependent on screen sizes (I'm testing on 1024*600 for this) or something else, but if I scroll down a bit in the similar users component, here is the original look:
image

And with this PR:
image

frontend/css/theme/bootstrap/utilities.less
MonkeyDo: Hm, let's see if this causes any issues with it missing.
If it does, I'll suggest to reinstate .show and (contrary to everything I said so far) change the side-nav to add important 
 : `display: flex !important;`
 
 Let's see.

frontend/css/donations.less
```diff
@@ -55,7 +55,7 @@
     &:hover {
       scale: 1.1;
     }
-    background: fadeout(white, 30%);
+    background: fadeout(white, 30%) !important;
```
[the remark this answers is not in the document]
anshg1214: So, for all these changes which might arise later also, should I create a new feature branch for Lobes and make changes there?

frontend/js/src/user-feed/UserFeedLayout.tsx
anshg1214: Fixed

frontend/js/src/playlists/components/PlaylistMenu.tsx
anshg1214: Fixed

frontend/js/src/components/Navbar.tsx
anshg1214: Fixed

frontend/css/donations.less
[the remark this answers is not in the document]
anshg1214: Fair enough

MonkeyDo (commented):
Another round of review :)
Looking closer and closer by the minute!

- ~I think the most visible change missing is setting the colors.
I think maybe the blue variable should be set in the scaffolding.less fil, not sure why it is not overwritten by the later import of colors.less...~
NVM I see that is done in #3220 
- The listencard border has changed, I see there is another pesky `!important` rule coming from somewhere.
- The gutter base value seems to have changed too, there is a bit more gutter space between columns in
- You will also need to review all the modals, as I think the layout for the modal titles + close button has changed (could be kept for a separate PR with all the modals changed in one go)
image
- There are a couple of instances of the deprecated glyphicons set still in use in the codebase (3, really) that are not going to work anymore and need to be replaced with the appropriate FA replacement
Like the eye icon here
image
- There is no `btn-default` class anymore (dark grey buttons), but we use it in a bunbch of places so we probably want to recreate it or replace all
- buttons text now wraps, which is probably unwanted (we don't want ot be adding a specific class to every button use)
image

anshg1214:
> You will also need to review all the modals, as I think the layout for the modal titles + close button has changed (could be kept for a separate PR with all the modals changed in one go)

> There is no btn-default class anymore (dark grey buttons), but we use it in a bunbch of places so we probably want to recreate it or replace all

Yes, both these changes have been accommodated in this PR: #3220

MonkeyDo (commented):
Out of my previous feedback, there is one last item still to be fixed:
-The listencard border has changed, I see there is another pesky `!important` rule coming from somewhere.

Apart form this and the small comment below, good to merge and move on to the next PR :)

PR #3218, opened 2025-03-10
LB-1442: Fix sending emails if username contains a special character

fettuccinae:
# Problem
Usernames are added to the "To" header of YIM notification emails, usernames that contains special/non atom characters and arent quoted doesnt comply with [RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.3) and create problems while sending out multiple mails.
   [JIRA ticket](https://tickets.metabrainz.org/browse/LB-1442)

# Solution
If any character of username contains special characters, enquote them.

### DIFF
--- a/listenbrainz/db/year_in_music.py
+++ b/listenbrainz/db/year_in_music.py
@@ -178,6 +178,12 @@ def send_mail(subject, to_name, to_email, content, html, logo, logo_cid):
     current_app.logger.info("Email sent to %s", to_name)
 
 
+def sanitize_username(username):
+    weird_chars = '@,;><:"()'
+    table = str.maketrans('','',weird_chars)
+    return username.translate(table)
+
+
 def notify_yim_users(db_conn, ts_conn, year):
     logo_cid = make_msgid()
     with open("/static/img/year-in-music-24/yim24-header-all-email.png", "rb") as img:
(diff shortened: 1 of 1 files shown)


mayhem:
Why not always quote the string and be done?

fettuccinae:
Yes, always quoting them is faster also there was an oversight on my part when double quotes and backslashes are present, we need to add a backslash to escape them.

mayhem (commented):
This feels a bit overly simplistic. What is the specification of allowed names/characters?

fettuccinae:
> This feels a bit overly simplistic. What is the specification of allowed names/characters?

Alphanums and some of the special characters( "!" / "#" /"$" / "%" /"&" / "'" /"*" / "+" /"-" / "/" / "=" / "?" /"^" / "_" / "`" / "{" /"|" / "}" /"~" ) are allowed without quotations in the header(https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.3).
Remaining special characters needs to be enclosed in quotations, except backslash and double quote.
To include them inside the string, we need to add a backslash and then enclose them in quotations. (https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.4)

> Also note that since quoted-pair
   is allowed in a quoted-string, the quote and backslash characters may
   appear in a quoted-string so long as they appear as a quoted-pair.

mayhem (approved):

PR #3219, opened 2025-03-11
LB-1760

Mshahnawaz1:
## Problem
[LB-1760](https://tickets.metabrainz.org/browse/LB-1760) – The link in settings/music-services/details/ was redirecting to a non-existing page, causing a 404 error.

## Solution
Updated the redirect URL point to the [listenbrainz-server](https://github.com/metabrainz/listenbrainz-server/), correcting the issue.
## Action
Please review and accept this pull request. No additional changes are required.

### DIFF
--- a/frontend/js/src/settings/music-services/details/MusicServices.tsx
+++ b/frontend/js/src/settings/music-services/details/MusicServices.tsx
@@ -355,7 +355,7 @@ export default function MusicServices() {
               premium user -{" "}
               <b>ListenBrainz will never read these pieces of data</b>. Please
               feel free to{" "}
-              <a href="https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/spotify_updater/spotify_read_listens.py">
+              <a href="https://github.com/metabrainz/listenbrainz-server/">
                 inspect our source code
               </a>{" "}
               any time!


anshg1214 (changes requested):
Thanks for fixing this issue. Could you also add the following tags: `target=‘_blank’` and `rel=‘noreferrer’` to make the link open in a new tab?

Mshahnawaz1:
I have redirected it [here](https://github.com/metabrainz/listenbrainz-server/blob/master/listenbrainz/listens_importer/spotify.py). Can it be merged now.

MonkeyDo:
I want to point that the ticket on Jira was already assigned to someone else, and in general that would mean that I would ignore this PR at the benefit of the person who is assigned the ticket.

In the future, please check if the ticket is assigned first. If it is not, then start by assigning it to yourself.
If it is already assigned, try to get in contact with the assigned user to ask if they are still working on it. Generally, if it was assigned recently [see 'Activity' tab] users are likely to still be working on it.

Please do not snipe tickets like that. I know we don't have a ton of tickets marked good-first-bug but it is better to go look for some rather than work on an already assigned ticket.

Mshahnawaz1:
Actually it is assigned to me on jira. My nickname is vedora there.

MonkeyDo:
Haha woopsie!
Sorry about that then @Mshahnawaz1, my bad. Carry on you're doing everything fine  :)

anshg1214 (approved):

kingosticks:
I just fell over this and ended up here when looking to fix it. Should this change be live yet or do you do periodic releases instead?

PR #3221, opened 2025-03-11
LB-1760: Fix the link in the details section to redirect to the correct file.

ahmvdev:
# Problem

The "Inspect our source code" link at `settings/music-services/details/` was redirecting to a non-existent file.
# Solution

Added the correct link now it redirects to spotify.py.

# Action

Review for feedback.


ahmvdev:
just realised this has already been done #3219

PR #3222, opened 2025-03-12
LB-1659: Fix stale closures in BrainzPlayer metadata

MonkeyDo:
It has been reported that the metadata sent by brainzPlayer can contain metadata from the previous track.
This is due to some stale closures in the BrainzPlayer component and the way the provided context updates, and at which point that data is accessed vs. updated.

We previously tried to save separate pieces of data from the BP context into React refs (escaping the regular rendering/update cycle), but I think that just cause more issues like the one reported in the ticket, while making the logic and code harder to understand.

### DIFF
--- a/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
@@ -144,25 +144,10 @@ export default function BrainzPlayer() {
     userPreferences?.brainzplayer?.appleMusicEnabled === false;
 
   // BrainzPlayerContext
-  const {
-    currentListen,
-    currentListenIndex,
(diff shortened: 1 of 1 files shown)


MonkeyDo:
@anshg1214 I didn't really change anything from your commit on a branch, just cherry-picked that commit to a clean new branch.
I tested the player again locally thoroughly and it seems to be behaving as expected, so IMO this is ready to merge.

anshg1214:
@MonkeyDo Cool! BTW, let's merge this PR after #2940 is merged to avoid any merge conflicts on that PR.

MonkeyDo:
Yes, I was thinking the same, I'll resolve any conflicts that will arise.

anshg1214 (approved):

PR #3223, opened 2025-03-12
Fix fallback to release-group cover art

MonkeyDo:
We have some fallback mechanisms in place for cases where there is no release MBID available, but one check is currently too restrictive and prevents the fallback from triggering, for example for pseudo-releases with no cover art.

If the use-submitted clean data with a release MBID and a release-group MBID, if the release does not have cover art it will currently not fall back to the release group cover art because we check if the user-submitted and the auto-mapped release MBIDs are equal.
If there is no mapped release the fallback will not trigger.

### DIFF
--- a/frontend/js/src/utils/utils.tsx
+++ b/frontend/js/src/utils/utils.tsx
@@ -19,6 +19,8 @@ import { getCoverArtCache, setCoverArtCache } from "./coverArtCache";
 const originalFetch = window.fetch;
 const fetchWithRetry = require("fetch-retry")(originalFetch);
 
+let APIServiceInstance = new APIServiceClass(`${window.location.origin}/1`);
+
 const searchForSpotifyTrack = async (
   spotifyToken?: string,
   trackName?: string,
@@ -611,6 +613,7 @@ const getPageProps = async (): Promise<{
     const apiService = new APIServiceClass(
       api_url || `${window.location.origin}/1`
     );
+    APIServiceInstance = apiService;
     globalAppContext = {
       APIService: apiService,
       websocketsUrl: websockets_url,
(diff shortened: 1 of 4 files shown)


frontend/js/src/utils/utils.tsx
```diff
@@ -947,7 +947,9 @@ const getAlbumArtFromListenMetadata = async (
     // release group cover art of the user submitted release mbid next
     const userSubmittedReleaseAlbumArt = await getAlbumArtFromReleaseMBID(
       userSubmittedReleaseMBID,
-      Boolean(caaReleaseMbid) && userSubmittedReleaseMBID !== caaReleaseMbid,
+      (Boolean(caaReleaseMbid) &&
+        userSubmittedReleaseMBID !== caaReleaseMbid) ||
+        true,
```
anshg1214, line 952: This condition will always return true

MonkeyDo: Yeah, not sure what I was thinking.

Looking back at the implementation, I'm not really seeing why this check was here in the first place to decide whether to fall back to RG.

Don't fallback if there is no caaReleaseMbid, or if the caaReleaseMbid is not the same as userSubmittedReleaseMBID.
Inverting the logic, we fallback to RG only if the caaReleaseMbid matches the userSubmittedReleaseMBID.

That makes no sense to me, but maybe I'm missing something?
`getAlbumArtFromReleaseMBID` is only used in two places, this function I'm modifying in this PR and another one that forces a truthy value:
https://github.com/metabrainz/listenbrainz-server/blob/23fece2ef5445c736e96653a92c93b3962841cf2/frontend/js/src/explore/fresh-releases/components/ReleaseCard.tsx#L128

I wonder if this should just be set to `true` here, and also be the default value.
Your thoughts?

MonkeyDo: I pushed another commit where:
1. I allow users to send only a RG MBID but no Release MBID (improbable but possible) and still fetch the cover art for the RG
2. Make the `useReleaseGroupFallback` value the same as the one I posted fro ReleaseCard: use the user-submitted releaseGroupMBID if available and default to `true` otherwise.

anshg1214 (approved):

anshg1214 (commented):
Can we rewrite the function instead to take `releaseMBID` and `releaseGroupMBID` and always use fallback?

MonkeyDo:
OK, thanks for the suggestions, I think this is a good basket of improvements :+1: 
Tested on test.LB, the reported issue with a pseudo-release without cover art is now fixed:

Before:
image

After:
image

PR #3226, opened 2025-03-16
LB-1765 Check if user exists on last.fm while connecting last.fm username

Suvid-Singhal:
# Problem

Success message was shown earlier while connecting to last.fm in settings even if the user did not exist on last.fm

# Solution

Fixed it by checking if the user gives 404 or not on querying its user page.

### DIFF
--- a/listenbrainz/webserver/views/settings.py
+++ b/listenbrainz/webserver/views/settings.py
@@ -252,7 +252,7 @@ def music_services_connect(service_name: str):
         raise APIBadRequest("Missing 'external_user_id' in request.")
     
     
-    lastfm_response = requests.get(f"https://www.last.fm/user/{data['external_user_id']}", timeout=5)
+    lastfm_response = requests.get(f"{current_app.config['LASTFM_API_URL']}?method=user.getinfo&user={data['external_user_id']}&api_key={current_app.config['LASTFM_API_KEY']}&format=json", timeout=5)
     if lastfm_response.status_code == 404:
         raise APINotFound("last.fm user not found.")
 
--- a/listenbrainz/webserver/views/settings.py
+++ b/listenbrainz/webserver/views/settings.py
@@ -6,6 +6,7 @@ from flask import Blueprint, render_template, request, url_for, \
 from flask_login import current_user, login_required
 from werkzeug.exceptions import NotFound, BadRequest
 import requests
+from requests.adapters import HTTPAdapter, Retry
 
 import listenbrainz.db.user as db_user
 import listenbrainz.db.user_setting as db_usersetting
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/settings.py
```diff
@@ -246,8 +247,14 @@ def music_services_connect(service_name: str):
         raise APINotFound("Service %s is invalid." % (service_name,))
 
     data = request.json
+
     if "external_user_id" not in data:
         raise APIBadRequest("Missing 'external_user_id' in request.")
+    
+    
+    lastfm_response = requests.get(f"https://www.last.fm/user/{data['external_user_id']}", timeout=5)
```
amCap1712, line 255: Use `https://www.last.fm/api/show/user.getInfo` instead.

Suvid-Singhal: Done :)

listenbrainz/webserver/views/settings.py
```diff
@@ -246,8 +247,14 @@ def music_services_connect(service_name: str):
         raise APINotFound("Service %s is invalid." % (service_name,))
 
     data = request.json
+
     if "external_user_id" not in data:
         raise APIBadRequest("Missing 'external_user_id' in request.")
+    
+    
+    lastfm_response = requests.get(f"{current_app.config['LASTFM_API_URL']}?method=user.getinfo&user={data['external_user_id']}&api_key={current_app.config['LASTFM_API_KEY']}&format=json", timeout=5)
```
MonkeyDo, line 255: This could be a bit clearer by separating the parameters in an object, like so:
https://github.com/metabrainz/listenbrainz-server/blob/3c7b90c46a3aae6d13f6cd65e2f5b0ba4190ee54/listenbrainz/domain/lastfm.py#L65-L72

listenbrainz/webserver/views/settings.py
```diff
@@ -246,8 +247,14 @@ def music_services_connect(service_name: str):
...
+    if lastfm_response.status_code == 404:
+        raise APINotFound("last.fm user not found.")
```
MonkeyDo, line 257: Could be useful to add the `data['external_user_id']` in this string ("last.FM user XYZ not found")

listenbrainz/webserver/views/settings.py
Suvid-Singhal: Done :)

MonkeyDo (changes requested):
Back-end code looks good :+1:

However, I noticed that the error message (with the new "username XXX not found") is not shown when the error is triggered.
Did you notice that?

I looked into it a bit, and basically this check should be removed:
https://github.com/metabrainz/listenbrainz-server/blob/c6127e1ee302c024c2a762553a35657abe8aa2e4/frontend/js/src/settings/music-services/details/MusicServices.tsx#L216-L221

bodyUsed is a boolean flag that is falsy at first, then gets set to true if the body has been read.
I think I am the one who added this line, looking at git blame, and it looks like I totally misunderstood what it does.

The check should be removed, and an additional `?` added in the next check to ensure we safely access a possible undefined body:
```
      if (response.ok) {
  [...]
      } else {
        const body = await response.json();
        if (body?.error) {
          throw body.error;
        }
        throw response.statusText;
      }
```

amCap1712:
Happy to merge once the frontend shows the proper error.

Suvid-Singhal:
> Did you notice that?

Yes I had, but I noticed that some other API errors were also not showing error messages so I thought it was normal or there was some problem with my setup.
Thanks for suggesting a fix for it!

MonkeyDo:
> Yes I had, but I noticed that some other API errors were also not showing error messages so I thought it was normal or there was some problem with my setup. Thanks for suggesting a fix for it!

Generally it is good to check with someone, because as you see there is always plenty of bugs that go unnoticed, and we all participate in adding more as we add more code :)
It's a natural part of coding on a large codebase, so assume you will encounter bugs, and if you are working on a feature you might as well fix other associated bugs :bug: 

For what it's worth, in this case to determine where the issue was coming from, I first checked my network tab in browser console to see what the server response was looking like. I saw the expected object with an `error` key and the expected error text with complete with username.
So that rules out the back-end entirely, and I focused on the front-end. Since I saw the title of the error in the toast, but not the error message, that narrowed it down to these few lines.

MonkeyDo (approved):
Looking good, thank you for this good improvement (and for fixing a bug I introduced ;) )!

Suvid-Singhal:
Thanks @MonkeyDo for sharing how you found the fix to the bug! This detailed process will really help me fix such issues in future. It's always a nice learning experience with your guidance 😊

PR #3230, opened 2025-03-19
LB-1657 Moved submitting-data page to setting

Mshahnawaz1:
### [LB-1657](https://tickets.metabrainz.org/browse/LB-1657): Submitting-data page need to be moved to setting section

Submitting-data page is currently in about page which is not intuitive for many users.
So moved submitting-data page to setting section.
### Action
No action needed. Please verify this pull request.


MonkeyDo:
Thanks for opening a PR!

I don't fully agree with the ticket as described, I think we should duplicate the access to that page, since the settings pages can only be accessed if logged in.
If a potential new user is looking at the website to get an idea of how it works, I think they should be able to access this page too.

So what you can do is leave the existing code as it was, and only add a link to the page in the settings page sub-menu layout (like you did [here](https://github.com/metabrainz/listenbrainz-server/pull/3230/files#diff-d3df95264b3d28caff23ebf11a4169fe5fa67ff6cdb60c996e972e3076ec43c6R23)) , but instead of pointing to `/add-data` it should point to `to: "../add-data/"` as a relative path.

Hope it makes sense, don't hesitate if you have questions!

Mshahnawaz1:
Thank you for your explanation.
So what iI need to do is [here](https://github.com/metabrainz/listenbrainz-server/pull/3230/files#diff-d3df95264b3d28caff23ebf11a4169fe5fa67ff6cdb60c996e972e3076ec43c6R23) is change to relative path ../add-data and import add-data from ("../../about/add-data/AddData").
Is this correct.

MonkeyDo:
> Thank you for your explanation. So what I need to do is [here](https://github.com/metabrainz/listenbrainz-server/pull/3230/files#diff-d3df95264b3d28caff23ebf11a4169fe5fa67ff6cdb60c996e972e3076ec43c6R23) is change to relative path ../add-data and import add-data from ("../../about/add-data/AddData"). Is this correct.

The only change you need compared to master is adding `{ to: "../add-data/", label: "Submitting data" },` to the src/settings/layout.tsx file. That should be the only required change (adds a link to the add-data page in the settings sub-menu), all the rest can be reverted to be the same as the master branch.

MonkeyDo:
Well, I understand the worry, but as I explained above, all the settings routes require signing in.
I believe users should be able to view the page before they create an account.
So it cannot be only in the settings page.

On the other side, the ticket requests that we should more easily find the page in question for users who would be looking for it in the settings page where we have other tools to connect to your music services.
So it should be added in the settings page too.


I don't see another alternative than duplicating the link to it and having it in both places, unless you have another idea to propose.

Mshahnawaz1:
Currently I don't find an alternative to this, for now we should go with the duplicates.

frontend/js/src/about/layout.tsx
```diff
@@ -12,9 +12,9 @@ const sections: Section[] = [
   { to: "about/", label: "About" },
   { to: "donate/", label: "Donate" },
   { to: "current-status/", label: "Site status" },
-  { to: "add-data/", label: "Submitting data" },
   { to: "data/", label: "Using our data" },
   { to: "terms-of-service/", label: "Terms of service" },
+  { to: "add-data/", label: "Submitting data" },
```
MonkeyDo, line 17: Is there a specific reason for moving this item down in the sub-menu?
If not please revert that change as well.

Aerozol:
Thanks monkey - I didn't think about the whole settings sections being hidden when not logged in!

PR #3232, opened 2025-03-20
LB-1228: Position toast container above music player

MonkeyDo:
The toast messages obstruct the player buttons, especially on mobile sizes where it obstructs all the controls.
Add some CSS to show the toast container above the player.
image


anshg1214 (approved):
LG!

PR #3233, opened 2025-03-21
Fixed Nonetype error!

mAmanullah7:
# Problem
Screenshot 2025-03-21 at 10 48 44 AM

I encountered an issue in the `artist_similarity` function within `listenbrainz/webserver/views/explore.py` at line 36. The specific error was a `TypeError: 'NoneType' object is not subscriptable`, which occurred when I tried to access the first element (`[0]`) of a database query result using `artist_mbid = result.fetchone()[0]`. The root cause was that the `fetchone()` method returned `None` when the database query didn’t find any matching records, likely because the artist name or ID I was searching for didn’t exist in the database, or there might be a logic issue in the SQL query itself. This caused the application to crash since I was attempting to index a `None` value, which isn’t possible.

This issue could affect users trying to explore artist similarity if the queried artist isn’t found, leading to an unhandled error instead of a graceful response.

# Solution

To fix this, I modified the `artist_similarity` function to handle the case where `fetchone()` returns `None`. I introduced a new variable, `result_row`, to store the result of `result.fetchone()`. Then, I added a conditional check: if `result_row` is `None`, I flash a message to the user saying "Artist not found" and redirect them to the explore index page. If `result_row` is not `None`, I proceed to extract the `artist_mbid` from `result_row[0]` and continue with the rest of the function as before. This ensures the application doesn’t crash when the artist isn’t found and provides a better user experience by redirecting with a helpful message.

Here’s the updated code for clarity:

```python
result_row = result.fetchone()
if result_row is None:
    flash("Artist not found")
    return redirect(url_for('explore.index'))
else:
    artist_mbid = result_row[0]
data = {
    "algorithm": "session_based_days_7500_session_300_contribution_5_threshold_10_limit_100_filter_True_skip_30",
    "artist_mbid": artist_mbid
}
return jsonify(data) 
```

I considered other approaches, like adding a default value for `artist_mbid`, but I felt that redirecting with a flash message was more appropriate for the user experience in this context. I also didn’t modify the underlying SQL query since the issue seems to be more about handling the absence of results rather than a query error, but this could be revisited if needed.

# Action

I’d appreciate it if the reviewers could take a look at the conditional logic I added to ensure it aligns with the project’s standards for error handling and user feedback. Additionally, if there are any suggestions for improving the flash message or the redirect behavior, I’d be happy to incorporate them. No other actions are needed beyond merging the change.

### DIFF
--- a/listenbrainz/webserver/views/explore.py
+++ b/listenbrainz/webserver/views/explore.py
@@ -1,4 +1,4 @@
-from flask import Blueprint, render_template, request, jsonify
+from flask import Blueprint, render_template, request, jsonify, flash, redirect, url_for
 from flask_login import current_user
 from sqlalchemy import text
 
@@ -32,8 +32,15 @@ def artist_similarity():
        ORDER BY total_listen_count DESC
           LIMIT 1
      """))
-
-    artist_mbid = result.fetchone()[0]
+    
+    # artist_mbid = result.fetchone()[0]   #getting error if its return None
+    
+    result_row = result.fetchone()
+    if result_row is None:
+        flash("Artist not found")
+        return redirect(url_for('explore.index'))
+    else:
+        artist_mbid = result_row[0]
     data = {
         "algorithm": "session_based_days_7500_session_300_contribution_5_threshold_10_limit_100_filter_True_skip_30",
         "artist_mbid": artist_mbid


MonkeyDo:
> the issue seems to be more about handling the absence of results rather than a query error

This is absolutely right.

However I think the solution of the flash message + redirect is perhaps not what we want.
In all other cases where we don't have data, we simply send back a 404, sometimes with a simple error message, and adapt the front-end to expect such an error if required.

Take for example this non-existing artist: https://listenbrainz.org/artist/37e9d7b2-7779-41b2-b2eb-3685351caad2/
image

Sure, it's pretty ugly and the error page could definitely be improved, but I mainly wanted  to show that this is the usual mechanism we go for.
So the server should indeed ensure  the server code does not raise an exception, and instead return a 404 response if there are no results.

mAmanullah7:
To fix this, I modified the `artist_similarity` function to handle the case where `fetchone()` returns `None`. I introduced a new variable, `result_row`, to store the result of `result.fetchone()`. Then, I added a conditional check: if `result_row` is `None`, I return a 404 response with a JSON error message, following the pattern used elsewhere in the project (e.g., in `entity_pages.py` for non-existing artists). If `result_row` is not `None`, I proceed to extract the `artist_mbid` from `result_row[0]` and continue with the rest of the function as before. This ensures the application doesn’t crash when the artist isn’t found and returns a consistent 404 response, which the front-end can handle as needed.

Here’s the updated code for clarity:

```python
result_row = result.fetchone()
if result_row is None:
    return jsonify({"error": f"Artist not found in the database"}), 404
else:
    artist_mbid = result_row[0]
data = {
    "algorithm": "session_based_days_7500_session_300_contribution_5_threshold_10_limit_100_filter_True_skip_30",
    "artist_mbid": artist_mbid
}
return jsonify(data)
```
I initially considered using a flash message and redirect, but after reviewing the project’s conventions (e.g., in `entity_pages.py` at lines 159–161), I realized that returning a 404 with a JSON error message is the standard approach for handling missing data. This aligns with how ListenBrainz handles similar cases, such as a non-existing artist (e.g., `https://listenbrainz.org/artist/37e9d7b2-7779-41b2-b2eb-3685351caad2/)`, where a `404` is returned. I didn’t modify the underlying SQL query since the issue is more about handling the absence of results rather than a query error, as noted.

MonkeyDo (approved):

mAmanullah7:
Thanks for the review @MonkeyDo 
i just wants to apologize for using ai unintentionally to rewrite my comment in .md format which i should avoided and write  myself! 
Thanks for pointing out the the gsoc starting getting page i've gone through it  & i'll definitely take care from now onwards.

mAmanullah7:
Sure, i just written my comment and converted it to .md format for my own text content thats it!

mAmanullah7:
Sure i understand and again i wants to applogize for use of LLM for formatting my own content.
Yeah i can also feel the difference because of the way i'm writing now and the way i wrote before! its because when i converted .md format it automatically changed my original text to better way and i thought its better so i used this one rather than my own just thinking my english is not much fluent! 

I accept my mistake and i understand that posting original is far better and trustable than posting polished !!
@MonkeyDo from now i'll make sure not to use any kind of LLM to rephrase and correct my format or rephrase my grammatical mistakes!

PR #3235, opened 2025-03-25
LB-1766: Fix Internal Server Error (500) when accessing listening history

fettuccinae:
# Problem

In user.py, Timezone isnt being added to the datetime object (making it offset-naive) which is later sent to timescale_listenstore, where its getting compared with another datetime object with timezone i.e min_user_ts. This is leading to ambigous behaviour for users with less amount of listens.
[Ticket](https://tickets.metabrainz.org/browse/LB-1766)

# Solution

Add UTC timezone to both from_ts and to_ts (similar to how we do it in api.py ) .

### DIFF
--- a/listenbrainz/webserver/views/test/test_user.py
+++ b/listenbrainz/webserver/views/test/test_user.py
@@ -1,7 +1,7 @@
 import json
 import logging
 import time
-from datetime import datetime
+from datetime import datetime, timezone
 from unittest import mock
 
 import orjson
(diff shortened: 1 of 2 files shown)


mayhem:
Did you use AI in any part of this PR?

fettuccinae:
No, I saw that api was working properly for the same query, observed the difference in the way we fetch listens on profile page and in the api call. Logged the from_ts and min_user_ts in timescale_listenstore and then realised the difference.

amCap1712 (approved):
LGTM, Thanks!

PR #3237, opened 2025-03-25
LB-1739: Sitewide stats: return correct 'count' and total_entity_count

MonkeyDo:
Currently the total entity count is returned as 'count' instead of the expected count of results.
Said count is useful to calculate pagination options, and the total entity count also useful for a variety of things.

Adapt these sitewide stats endpoints to return the same format as user stats with count and total_{entity}_count (adapted to each entity name), using the same `count_key` attribute as is used in the `_get_entity_stats` function.
Basically adapting sitewide stats to work the same and return the same data as user stats.

### DIFF
--- a/listenbrainz/testdata/sitewide_top_artists_db.json
+++ b/listenbrainz/testdata/sitewide_top_artists_db.json
@@ -14,5 +14,6 @@
     ],
     "to_ts": 1593043183,
     "stats_range": "all_time",
-    "count": 2
+    "count": 2,
+    "total_artist_count": 123
 }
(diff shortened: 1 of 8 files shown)


MonkeyDo:
Not sure if [this testing utility change](https://github.com/metabrainz/listenbrainz-server/pull/3237/files#diff-de1a985289ca8a5a9cf662700b0b87ed03e6d14fa6a6779b66b6acf4d7081babR250) is correct. I don't understand how the tests were previously testing whether the count was correct, if it wasn't using the count argument passed to `assertSitewideStatEqual`

amCap1712:
I'll try to figure out what's wrong with the tests later today.

amCap1712:
@MonkeyDo the testing utility was working correctly because according to the previous behavior, the count key returns the total entity count and not the actual number of results requested/returned in the current response.

Since the new behavior aligns it with user stats, the changes to the utility are correct as now the actual count passed should be returned.

amCap1712 (approved):
LGTM, Thanks!

PR #3238, opened 2025-03-26
Replace color utilities in AlbumPage and ListeningNow page

MonkeyDo:
This PR is three-fold:

1. fixes an issue with imports for the tinycolor library (`import * as tinycolor from...`  instead of `import tinycolor from...` ) currently breaking the HueSound and MetadataViewer pages
2. Replaces the `getAverageRGBOfImage` utility function with Vibrant library in AlbumPage and MetadataViewer components
3. Change the Typescript configuration `"esModuleInterop": true` (see [TS docs](https://www.typescriptlang.org/tsconfig/#esModuleInterop)) to allow importing ESM modules
  On that topic, this was the issue I encountered before in #3229 and mutilple other times before
  
We had this function (shamelessly copied) to calculate the average color of an image, which produced barely OK results.
In recent work on the mobile BrainzPlayer UI, we introduced [Vibrant](https://vibrant.dev/) which produces much. much better results by picking actual colors from the image rather than averaging the image.

Getting rid of the utility function as it is unused after those changes.

P.S: Please ignore the branch name, I done goofed.


MonkeyDo:
Deployed to test.LB for testing, all the affected pages I tried working perfectly.

Album page even looks better now with one of the fixes in this PR (color gradient actually applied):
image

image

anshg1214 (approved):

PR #3240, opened 2025-03-28
chore: Update Typescript to v5

MonkeyDo:
Nothing breaking for us, just two small changes required to appease small errors.

TS 5.0 announcement blog: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#decorators

**Note: When this is merged, the offscreencanvas types dependency added in #3225 should be removed.**


MonkeyDo:
After merging this, I will start working on removing the LESS compilation and dependencies, and moving everything to SASS instead.

anshg1214 (approved):
LG!

PR #3241, opened 2025-03-28
Disable "Export to Apple Music" option if user not signed in, or into Apple Music

MonkeyDo:
We hide the "Export to Spotify" button in the playlist menu, but the logic to show/hide the Apple Music button was incorrect and the option was always shown, regardless of logged-in status or apple music connection status.

Also made the separator above conditionally rendered if either of the export options are shown to the user instead of having a double separator if neither is shown.


MonkeyDo:
I opted for disabling the link/buttons instead of hiding them, to try and maintain the same UI regardless of log-in and music services status:
Not pictured here is the `cursor: not-allowed` style on hover.

image

PR #3243, opened 2025-04-07
Bootstrap 4 Final Upgrade

anshg1214:


MonkeyDo (commented):
Indeed the preview on test.LB looks quite good already.
Very preliminary review just to note a couple of things before I forget them, feel free to ignore for now.
 
I wanted to note somewhere that we can now (FINALLY!) remove Jquery from the js/lib directory, as well as references to it in listenbrainz/webserver/templates/base.html and index.html

MonkeyDo (commented):
I think all the glyphicons fonts can probably be deleted (from frontend/fonts/), along with any other mentions of glyphicons in the codebase.
Apart from that, i suppose if we are continuing with BS5 straight away I'll do another full pass to review everything

PR #3244, opened 2025-04-07
Upgrade to Bootstrap 5

anshg1214:
See https://getbootstrap.com/docs/5.0/migration/


MonkeyDo:
For posterity:
I ended up having to install and use the react-bootstrap library for the modal components.
The reason for that is that since we inject the modal markup on the fly (NiceModal library creates the modal only when needed), the Bootstrap javascript throws an error because the markup does not exist at the time of initialization.
I tried fixing the issue but I think it's just not made to be used in our specific setup.

Using React-Bootstrap simplifies  a lot the use with NiceModal and it was on the list anyway (see [this note](https://github.com/metabrainz/listenbrainz-server/blob/892bb64a79eb4eb9b5cb7c89f54774279c7ec414/frontend/js/src/user-feed/ThanksModal.tsx#L14-L21) in all our modals...), so why not use it now to fix this problem?

We import only the modal sub-components from RBS so the impact on the JS bundle should be minimal.

MonkeyDo:
Merging this PR now into the bootstrap5 feature branch, where I will work on merging the master branch and resolving merge conflicts and do a final check of everything.

PR #3247, opened 2025-04-08
UserFeed: improve missing "thank you" events preview (Part II)

MonkeyDo:
Following in the steps of #3242 I wanted to add a better system than showing a placeholder for missing thank you events previews.

For this event type, we show a preview of the original event underneath it.
We sometimes do not have the original event in the currently loaded pages of events tough, as they may be older events.

This PR adds an API endpoint and some front-end query cache magic to load these missing events separately when needed.

Before:
image
After:
image

In case the original event was deleted in the meantime, or in case of unforeseen issues:
image

### DIFF
--- a/frontend/js/src/user-feed/UserFeed.tsx
+++ b/frontend/js/src/user-feed/UserFeed.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable jsx-a11y/anchor-is-valid */
 import { IconProp } from "@fortawesome/fontawesome-svg-core";
 import {
   faBell,
@@ -53,7 +52,8 @@ import {
   preciseTimestamp,
 } from "../utils/utils";
 import ThanksModal from "./ThanksModal";
-import { EventType, type FeedFetchParams } from "./types";
+import type { FeedFetchParams } from "./types";
+import { EventType } from "./types";
 import Card from "../components/Card";
 
 enum EventTypeinMessage {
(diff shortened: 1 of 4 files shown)


frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -207,7 +207,45 @@ export default function UserFeedPage() {
...
   const { pages } = data || {}; // safe destructuring of possibly undefined data object
   // Flatten the pages of events from the infite query
-  const events = pages?.map((page) => page.events).flat();
+  const events = pages?.map((page) => page.events).flat() ?? [];
+
+  // Events mentioned in thank you events but not part of current cache, fetched separately as needed
+  const [separatelyLoadedEvents, setSeparatelyLoadedEvents] = React.useState<TimelineEvent<EventMetadata>[]>([])
+
+  React.useEffect(() => {
+    async function fetchOGEvents(missingEventsIDs:number[]) {
```
MonkeyDo, line 216: I could maybe find a better name for this function, no?

MonkeyDo:
@Suvid-Singhal I had to wrestle a bit with react-query to get it to do what I needed it to do, but I arrived at a decent solution. Works like a charm separately loading events that are referenced in a thank-you event, but not currently present on the page.

Suvid-Singhal:
That's really neat @MonkeyDo 
I'm glad that the solution we discussed could be implemented!

MonkeyDo:
> That's really neat @MonkeyDo I'm glad that the solution we discussed could be implemented!

Me too!
My original idea was more complicated than it needed to be, I think.
I was trying to avoid having separate cache items and separate react state for the extra loaded events like what I ended up with:
image
"feed" is where we store the pages of events, but the extra events are stored separately with a different cache key "feed-event" + event_id.

But at least that way the code is pretty readable and can be reasoned with.

I don't think it is a very common use-case to have a timestamp-based infinite query pagination system, and at the same time also need to load single events using their ID.
But oh well, it makes sense in the context of our feed and the thank you events.

MonkeyDo:
I also want to change the icon for the thank you events and increase the size of the icons.
After some use I thought it was not very legible at that small size.
What do we think about this change?

Before:
image
image

After:
image
image

Suvid-Singhal:
Yea it looks much better now
The previous icon reminded me of this:
image
The engine check sign
😂😂😂

I think the first one is the best
The heart pulse looks very similar to like button and comment bubble one doesn't fit "Thank You" imo.
The fire one can work but I prefer the first suggestion.

MonkeyDo:
Thanks for your input, I'm thinking the same

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -295,6 +295,58 @@ def user_feed(user_name: str):
     }})
 
 
+@user_timeline_event_api_bp.get('/user/<user_name>/feed/events/<event_id>')
+@crossdomain
```
anshg1214, line 299: Let's add some more validations to the route.
```suggestion
@user_timeline_event_api_bp.get('/user/<string:user_name>/feed/events/<int:event_id>')
```

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -295,6 +295,58 @@ def user_feed(user_name: str):
     }})
 
 
+@user_timeline_event_api_bp.get('/user/<user_name>/feed/events/<event_id>')
```
anshg1214, line 298:
```suggestion
@user_timeline_event_api_bp.get('/user/<user_name>/feed/event/<event_id>')
```

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -295,6 +295,58 @@ def user_feed(user_name: str):
...
+    user = validate_auth_header()
+    if user_name != user['musicbrainz_id']:
+        raise APIForbidden(
+            "You don't have permissions to view this user's timeline.")
+
+    user_event = db_user_timeline_event.get_user_timeline_event_by_id(
+        db_conn, event_id)
+
+    if not user_event:
+        raise APIBadRequest(f"Event with id {event_id} not found")
```
anshg1214, line 328: This should return 404 and not 400
```suggestion
        raise APINotFound(f"Event with id {event_id} not found")
```

frontend/js/src/user-feed/UserFeed.tsx
```diff
@@ -207,7 +207,45 @@ export default function UserFeedPage() {
...
+      // separately from fetching the feed pages
+      const promises = missingEventsIDs.map( eventId => {
+        return queryClient.ensureQueryData({
+          queryKey: ['feed-event', eventId],
+          queryFn: () => APIService.getFeedEvent(eventId, currentUser.name, currentUser.auth_token as string),
+        });
+      });
+      const resultsArray:TimelineEvent<EventMetadata>[] = [];
+      const promiseResults = await Promise.allSettled(promises);
+      promiseResults.forEach((res,idx) => {
```
anshg1214, line 227: Lookss like the linter is not working on this file. Otherwise there would have been a space after the comma.

anshg1214 (changes requested):
Overall this PR looks good. Just some minor improvements, and we're good to go.

listenbrainz/webserver/views/user_timeline_event_api.py
anshg1214, line 299: Let's add some more validations to the route.

```suggestion

@user_timeline_event_api_bp.get('/user/<string:user_name>/feed/events/<int:event_id>')
@crossdomain
```

anshg1214 (changes requested):

listenbrainz/webserver/views/user_timeline_event_api.py
MonkeyDo: I don't think this is right, as all the other event-related endpoints use the plural form (including ones that affect a single event such as `feed/events/delete` and `feed/events/hide`)

MonkeyDo: Asked on the main chat, general agreement is to keep using the plural form since we use that for all other /events/ endpoints.
Ref: https://chatlogs.metabrainz.org/libera/metabrainz/2025-04-15/?msg=5457090&page=1

PR #3252, opened 2025-04-11
LB-1630: Dashboard for users with sporadic listens loads slowly

fettuccinae:
# Problem

Slow lookup times for users with sporadic listens. [Ticket.](https://tickets.metabrainz.org/browse/LB-1630)

# Solution

1. Changed the maximum number of passes from 10 to 3, fixing the maximum size of search window to 9 months in one pass.
2. Api returns search_start_ts and search_end_ts so that frontend uses these timestamps and doesnt make redundant requests.
3. If listens are less than expected, render a Load More listens button which uses the search_end_ts as max_ts and makes next request.

### DIFF
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -38,6 +38,9 @@ WINDOW_SIZE_MULTIPLIER = 3
 
 LISTEN_COUNT_BUCKET_WIDTH = 2592000
 
+# Set maximum no of passes made. 
+MAX_PASS = 3 # 1 month + 3 months + 9 months
+
 MAX_FUTURE_SECONDS = timedelta(minutes=10)  # max fwd clock skew
 EPOCH = datetime.fromtimestamp(0, timezone.utc)
 
@@ -223,7 +226,7 @@ class TimescaleListenStore:
         min_user_ts, max_user_ts = self.get_timestamps_for_user(user["id"])
 
         if min_user_ts == EPOCH and max_user_ts == EPOCH:
-            return [], min_user_ts, max_user_ts
+            return [], min_user_ts, max_user_ts, EPOCH, EPOCH
 
         if to_ts is None and from_ts is None:
             to_ts = max_user_ts + timedelta(seconds=1)
@@ -301,12 +304,14 @@ class TimescaleListenStore:
 
         t0 = time.monotonic()
 
+        # Search starts from this timestamp.
+        constant_ts = to_ts if not to_dynamic else from_ts 
         passes = 0
         while True:
             passes += 1
 
             # Oh shit valve. I'm keeping it here for the time being. :)
-            if passes == 10:
+            if passes > MAX_PASS:
                 done = True
                 break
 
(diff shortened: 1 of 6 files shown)


fettuccinae:
@amCap1712 I have modified the fetch_listens method as suggested. If we limit the passes itself, we dont need to do anything different for initial load.

fettuccinae:
@amCap1712

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -365,14 +347,30 @@ def fetch_listens(self, user: Dict, from_ts: datetime = None, to_ts: datetime =
...
+                    window_size *= WINDOW_SIZE_MULTIPLIER
+                    to_ts += window_size
+
+                if from_dynamic:
+                    to_ts -= window_size
+                    window_size *= WINDOW_SIZE_MULTIPLIER
+                    from_ts -= window_size
+
+        # Search ends at this timestamp.
+        dynamic_ts = from_ts if from_dynamic else to_ts
```
amCap1712, line 363: Consider the case where the window was extended and a number of listens greater than the limit were found, in that case the while loop above would terminate before exhausting the search timestamps and the caller of this function would skip the timestamps to the end of the window while actually it should start from the the timestamp of the last used listen.

So you should take into consideration the min/max timestamp of the listens returned and not just the window limits.

amCap1712 (changes requested):
I simplified the logic in updating the window sizes but there is a bug that needs to be resolved.

listenbrainz/listenstore/timescale_listenstore.py
amCap1712: Makes sense but this information will be eventually exposed through the API as well so it makes sense to still change it for avoiding confusion with other clients.

fettuccinae: Ahh understood, so , if len(listens) >= limit, set the search_end_ts as last listen's timestamp?

amCap1712: Off the top of my head yes but also take into consideration the order of sorting, from_ts and to_ts. And add tests as well for the same. I'll need to think more on this to see if there are any other edge cases.

amCap1712:
Hi! The code works fine but I haven't merged it yet because I am thinking about the changes in behavior for the API users.

fettuccinae:
> Hi! The code works fine but I haven't merged it yet because I am thinking about the changes in behavior for the API users.

Hey! For the api users, we can keep calling the fetch_listens function till 25 listens are reached or till the search_end_ts has exceeded min_user_ts, or we can just change the api description to "fetching upto 25 listens in 13 months from the timestamp".

MonkeyDo:
@amCap1712 As far as i understand, the website part of this PR should be ready for review?
Let me know if there are more changes you want to implement regarding LB web page loading, will review otherwise.

amCap1712:
@MonkeyDo I discussed this with @mayhem on IRC and we are not entirely happy with this behavior for the API side of things.

PR #3253, opened 2025-04-14
LB-1776: Artist origin map tidy up

granth23:
# Problem
LB-1776: Artist origin map tidy up

# Solution

Big Screens:
image

Narrow Devices:
image

# Action

### DIFF
--- a/frontend/js/src/user/stats/components/UserArtistMap.tsx
+++ b/frontend/js/src/user/stats/components/UserArtistMap.tsx
@@ -92,7 +92,15 @@ export default function UserArtistMap(props: UserArtistMapProps) {
         <div className="col-md-9 col-xs-6">
           <h3 style={{ marginLeft: 20 }}>
             <span className="capitalize-bold">Artist Origins</span>
-            <small>&nbsp;Click on a country to see more details</small>
+            <small>&nbsp;(top 1,000 artists)</small>
+            <p
+              style={{
+                fontSize: 12,
+                color: "#666",
+              }}
+            >
+              Click on a country to see more details
+            </p>
           </h3>
         </div>
         <div className="col-md-2 col-xs-4 text-right" style={{ marginTop: 20 }}>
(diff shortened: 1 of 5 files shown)


frontend/js/src/user/stats/components/UserArtistMap.tsx
```diff
@@ -92,7 +94,16 @@ export default function UserArtistMap(props: UserArtistMapProps) {
         <div className="col-md-9 col-xs-6">
           <h3 style={{ marginLeft: 20 }}>
             <span className="capitalize-bold">Artist Origins</span>
-            <small>&nbsp;Click on a country to see more details</small>
+            {!isMobile ? <small>&nbsp;(top 1,000 artists)</small> : null}
+            <br />
+            <p
```
MonkeyDo, line 99: I read the ticket again, and I see what the issue with the purple color is.
Any `<small>` html element inside a heading element (in this case h3) gets that colour applied.
So I suggest a couple of structure changes to modify that, instead of using inline styles to force a look:

The trick here is using `display: "inline-block"` on the heading element to allow having the `<small>` element on the same line without being a child of the heading
```
        <div className="col-md-9 col-xs-6">
          <h3 className="capitalize-bold" style={{ display: "inline-block" }}>
            Artist Origins
          </h3>
          <small className="hidden-xs">&nbsp;(top 1,000 artists)</small>
        </div>
```
And the text underneath goes in a `<p>` element right under the `row` element with the heading, dropdown and link button:
The reason is so that it has the whole width of the card instead of wrapping.
```
      </div> 
      <p>Click on a country to see more details</p>
      <Loader isLoading={loading}>
```
However, to have a good final look on mobile, there is one more change required:
I don't like those random 20px margins added willy-nilly instead of having a padding in the containing card elements, which is the right way to do this layout.

So in frontend/css/stats.less, add a `padding: 0 20px;` to the `.user-stats-card` definition, and then search in the whole frontend codebase for `style={{ marginLeft: 20 }}` to delete these margins.

The result:
image

MonkeyDo (changes requested):
I suggest lots of small structural changes below, as these headings have been implemented a little bit wrong.

Overall it simplifies the markup quite a bit and removes inline styles.

frontend/js/src/user/stats/components/UserArtistMap.tsx
MonkeyDo: P.S: I think the heading can still be improved with flexbox layout instead of using bootstrap columns, to improve text wrapping, but I think it will be more complex and would suggest keeping that for another day...

PR #3256, opened 2025-04-16
Fix event_type enum serialization

MonkeyDo:
Json serialization of enum string values fails, forcing us to manually serialize the value.
There are a few instances of us doing that in the codebase.
I didn't do the conversion properly in #3247 , but it seemd like a very annoying and easy to forget little obstacle, so I looked into it closer to avoid future copy-paste issues or people like me just not knowing why this needs to be done.

It turns out we can make the enum a subclass of 'str' and that will be enough of a hint for json serialization to workv just fine... Thanks internet stranger! https://stackoverflow.com/a/51976841/4904467

For minimal testing: https://www.online-python.com/xBoJlNdSML

### DIFF
--- a/listenbrainz/db/model/user_timeline_event.py
+++ b/listenbrainz/db/model/user_timeline_event.py
@@ -19,7 +19,7 @@
 from pydantic import BaseModel, NonNegativeInt, constr, conlist
 
 from datetime import datetime
-from enum import Enum
+from enum import StrEnum
 from typing import Union, Optional, List
 
 from data.model.listen import APIListen, TrackMetadata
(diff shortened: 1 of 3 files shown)


listenbrainz/db/model/user_timeline_event.py
```diff
@@ -27,7 +27,7 @@
 from listenbrainz.db.msid_mbid_mapping import MsidMbidModel
 
 
-class UserTimelineEventType(Enum):
+class UserTimelineEventType(str, Enum):
```
amCap1712, line 30: Since we are on Python 3.11 in LB, you can use StrEnum here. https://docs.python.org/3/library/enum.html#enum.StrEnum

amCap1712 (approved):
LGTM, Thanks!

PR #3258, opened 2025-04-18
LB-1777: Artist origin map "play something from this country"

granth23:
# Problem

[LB-1777: Artist origin map "play something from this country"](https://tickets.metabrainz.org/projects/LB/issues/LB-1777?filter=allopenissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC)

# Solution

image

Modified the tooltip to accomodate a `Play` button that plays random songs generated by `LB-Radio` from the country.

# Action


MonkeyDo (changes requested):
Really cool feature !
Nice work.

I do have some requests for changes below, mostly about refactoring and reusing code instead of copying.
The main reason is that if in the future someone changes the radio-playlist fetching code here they might forget to change it in the LBRadio component, or vice-versa.
Refactoring to use a single method in both components prevents that sort of issue.

frontend/js/src/user/stats/components/Choropleth.tsx
```diff
@@ -177,65 +184,212 @@ export default function CustomChoropleth(props: ChoroplethProps) {
...
-            </Link>
-            <br />
           </div>
-        ))}
-      </div>
+
+          {artists?.length > 0 && (
+            <>
+              <hr style={{ margin: "0.5em 0" }} />
+              {artists?.slice(0, 10).map((artist: UserArtistMapArtist) => (
```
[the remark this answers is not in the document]
granth23: Nice add on, have implemented the same

frontend/js/src/user/stats/components/Choropleth.tsx
```diff
@@ -176,66 +180,168 @@ export default function CustomChoropleth(props: ChoroplethProps) {
...
+  );
+
   const customTooltip = useMemo(() => {
-    if (!selectedCountry?.data) {
+    if (!selectedCountry) {
       return null;
     }
 
+    const countryName =
+      selectedCountry.label || (selectedCountry as any).properties?.name;
```
MonkeyDo, line 210: Not a fan of slapping `as any` when there is a  type issue, as it can hide potential problems instead of resolving them.

I do see that the nivo types are not super precise, and this properties.name comes from how we set the label based on our map data.

So we can define a precise type like so, augmented with the structure of [our geodata json](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/user/stats/data/world_countries.json):
```ts
type TFeature = ChoroplethBoundFeature & {
  type: "Feature";
  properties: {
    name: string;
  };
  geometry: {
    type: string;
    coordinates: any;
    id: string;
  };
};
```

Using that type for the `selectedCountry` state, in customTooltip you can revise the country name to be
```ts
const countryName = selectedCountry.properties.name;
```

Two more places where you will use the type: in showTooltipFromEvent two type changes: using GeoMapEventHandler (which is more forgiving when it comes to the type of `feature`), as well as our new type:
```ts
const showTooltipFromEvent: GeoMapEventHandler = useCallback(
    (feature: TFeature, event: React.MouseEvent<HTMLElement>) => {
```

And finally, where we defined the label access function for the ResponsiveChoropleth component:
```ts
label={(feature: TFeature) => feature.properties.name}
```

frontend/js/src/user/stats/components/Choropleth.tsx
```diff
@@ -176,66 +180,168 @@ export default function CustomChoropleth(props: ChoroplethProps) {
...
+                    <br />
+                  </div>
+                ))}
+              </div>
+            </>
+          )}
+        </div>
+      </button>
     );
   }, [selectedCountry, selectedMetric]);
```
MonkeyDo, line 334: I see an eslint warning for this line; you need to add `handlePlayCountryTracks` as a dependency of this hook

MonkeyDo (changes requested):
Looking great so far!

I took some time to see about better types, it wasn't straightforward because the nivo types a not easily extensible, but I got to a decent solution in the end.

Functionally, it's working great!

MonkeyDo (approved):
Thank you!

The feature looks great and works great!

PR #3259, opened 2025-04-18
LB-1784: Hyperlink Artist name and Improve tooltip on the Artist Activity graph

granth23:
# Problem

LB-1784: Hyperlink Artist name and Improve tooltip on the Artist Activity graph

# Solution

image

1. Tooltip changes to just store the Album Name.
2. Name on X-Axis modified to redirect to artist page.

# Action


frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -169,7 +209,13 @@ export default function UserArtistActivity(props: UserArtistActivityProps) {
                     tickPadding: 5,
                     tickRotation: -45,
                     renderTick: (tick) => (
-                      <g transform={`translate(${tick.x},${tick.y})`}>
+                      <g
+                        transform={`translate(${tick.x},${tick.y})`}
+                        onClick={(event) =>
+                          handleArtistLabelClick(tick.value, event)
+                        }
```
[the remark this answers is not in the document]
MonkeyDo: Currently if there is no MBID clicking on the text does nothing, despite it looking like a visited link.
I think the solution above will help with that, returning a Link element or regular text accordingly.

However it will inevitably take a bit of time to replace the SVG text with html elements, rotating and placing it correctly.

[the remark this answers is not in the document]
granth23: I tried implementing the foreignObject thing but wasn't able to do the positioning properly and the multi-line split, would it be possible for you to assist me with the same

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -168,28 +203,59 @@ export default function UserArtistActivity(props: UserArtistActivityProps) {
...
+                      return (
+                        <g transform={`translate(${tick.x},${tick.y})`}>
+                          <foreignObject
+                            x={-width + 10}
+                            y={0}
+                            width={width}
+                            height={height}
+                            style={{ overflow: "visible" }}
+                          >
+                            <div
```
MonkeyDo, line 230: You can have the foreignObject element with a single child element (in this case the `Link` element) without a need for the nested div element.
That does mean that you will need to move the transform styles to the foreignObject.

I also want to propose a change to those transforms to go along with the other changes I suggested.
Basically we can make all the foreignObject elements the same width and height (say 90 by 30, for example) and place them all the same way (`x=-90` and `y=0`), then do all the transforms on the foreignObject element instead.
You might need to play with it a bit, but those values worked pretty well when i tried:
`transform: translate(-10px, 0px) rotate(-45deg)` < positions the foreignObject first then rotates it

If in the future we want to tweak things it'll be easier to reason with transformation on one element rather than two different transforms at a time.


To recap, this is the markup I ended up with:

```html
<g transform="translate(314,0)">
  <foreignObject
    x="-90"
    y="0"
    width="90"
    height="30"
    style="transform: translate(-10px, 0px) rotate(-45deg)">
      <a
        href="/search?search_term=FNORD&amp;search_type=artist"
        class="ellipsis-3-lines">
          Very long text to test ellipsis style class, even longer than needs to be
      </a>
    </foreignObject>
</g>

```

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -166,28 +201,39 @@ export default function UserArtistActivity(props: UserArtistActivityProps) {
...
+                            style={{
+                              overflow: "visible",
+                              transform: "translate(-5px, 5px) rotate(-45deg)",
+                              textAlign: "right",
+                              fontSize: "11px",
+                              lineHeight: "1.25em",
+                            }}
+                          >
+                            <Link to={linkTo} className="ellipsis-3-lines">
+                              {lines.join("\n")}
```
MonkeyDo, line 231: We are now splitting on newlines and reconstituting it afterwards for not reason :)
You'll need to remove the `const lines = ...` line above
```suggestion
                              {tick.value}
```

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -166,28 +201,39 @@ export default function UserArtistActivity(props: UserArtistActivityProps) {
...
+                            x="-90"
+                            y="0"
+                            width="90"
+                            height="30"
+                            style={{
+                              overflow: "visible",
+                              transform: "translate(-5px, 5px) rotate(-45deg)",
+                              textAlign: "right",
+                              fontSize: "11px",
+                              lineHeight: "1.25em",
```
MonkeyDo, line 227: I still think this should have the value of 1em, for some specific screen sizes it prevents overlapping:
image
image

MonkeyDo, line 227: These three text-related css styles should probably be moved to the containing div a few levels up, the one with styles already applied, at line 182:
https://github.com/metabrainz/listenbrainz-server/pull/3259/files#diff-bf40182f960aac5dbf97e7818519330b8d003433247e6c7a748a110c08b3a77aR182

It will only apply to the HTML content inside it (not the SVG text) and prevents having to define it for each of these foreignObjects

MonkeyDo (changes requested):
A few more minor changes, but the result looks great.
After this wave of tweaks I'm happy to merge.

MonkeyDo (approved):
Thank you for these good improvements!

PR #3260, opened 2025-04-23
LB-1768: Show LastFM importer status.

fettuccinae:
# Problem
In lastFm importer there is no other feedback than a "successfully connected" message, leaving users to wonder if they did it right, or if something is broken. [Ticket](https://tickets.metabrainz.org/browse/LB-1768)

# Solution
1. Added 2 columns( "import_status" and "imported_listens") in external_service_oauth table .
2. Added get_import_data function which fetches the value of "import_status" and "imported_listens" and update_import_data which updates the values.
3. Added the update_import_data function to lastfm importer so that it can update it after each loop.
4. Added an endpoint '/music-services/<service_name>/import-status/' which returns the status of import and imported_listens.
5. Changed the '/music-services/<service_name>/connect/' endpoint to return total number of LastFM listens.
6. Added an ImportStatus component which fetches data from '/import-status' and renders it.
7. Changed the ToastMsg while connecting to last.fm to show total number of scrobbles.
(changed lastfm spelling in handleConnectToLastFM)

Import Status


amCap1712 (changes requested):
I think the status/progress column should be a JSONB column in the listens_importer table instead. That way you could have different status for different services as relevant. Also, for the last.fm importer there would be a difference between the initial run and the incremental runs so the JSONB could allow for storing messages accordingly.

fettuccinae:
> I think the status/progress column should be a JSONB column in the listens_importer table instead. That way you could have different status for different services as relevant. Also, for the last.fm importer there would be a difference between the initial run and the incremental runs so the JSONB could allow for storing messages accordingly.

Yes, JSONB column makes sense, I'll make the changes. also, are there more differences between initial run and incremental run except the number of listens to be imported?

MonkeyDo:
ping @amCap1712 , is there anything missing for this PR?
Some live testing perhaps? Happy to help if that's the case.

Asking because we had one more instance of a user asking "how long until...." just a couple of days ago.

amCap1712 (approved):
I renamed the import_info column to status with the intention of merging the error message in it in future. Also, I replace the import progress with last imported listen's time. See commit message for a short explanation.

MonkeyDo:
There is still one small issue that will need to be resolved (not necessarily before merging, but...) :
If I select the import start date of today (no scrobbles to import), it still shows me that "XXXX listens will be imported" using the total count of scrobbles I have, even though there are actually no listens from the selected date onwards.

So we might need to add some logic to calculate the actual number of listens to import (not certain this is easy to do with the LFM API...) or only show the total number is we know the user imports from the beginning (no "Start import from" date selected in the UI)

MonkeyDo (approved):
Nice work @fettuccinae !
Another highly-requested feature in the bag !

fettuccinae:
@MonkeyDo I dont mind the UI improvement at all.
 Most of the heavy lifting was done by you and @amCap1712 , so more nice work to yall! 😁

MonkeyDo:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **ProgrammingError: (psycopg2.errors.UndefinedColumn) column "status" does not exist** `api_v1.latest_import` [View Issue](https://metabrainz-foundation-inc.sentry.io/issues/44116405/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

That was me, I deployed to test.LB but forgot to run the DB update script

PR #3261, opened 2025-05-01
Update python and dependencies

amCap1712:
Update to Python 3.13 and dependencies to latest versions.


mayhem (approved):
Yay, thanks for the update!

PR #3263, opened 2025-05-03
LB-1789: Close button for youtube music player

OSinA04:
# Problem

According to [this ticket](https://tickets.metabrainz.org/projects/LB/issues/LB-1789?filter=allopenissues) we are unable to close the youtube music player window. All we have is the drag button and to close the window we have to reload the page which is quite inconvenient.

# Solution

1. Added a close button to the header of the `YoutubePlayer` component which uses the `onClose` prop passed through `BrainzPlayer`. 
2. Implemented a `handleCloseYoutubePlayer` function which is passed as the onClose prop.
3. Added `.youtube-close-button` css class for styling.

https://github.com/user-attachments/assets/47f0ed3c-378b-468a-86ad-00e8a8eec593

# Action

### DIFF
--- a/frontend/css/brainzplayer.less
+++ b/frontend/css/brainzplayer.less
@@ -114,6 +114,13 @@
             cursor: grabbing;
           }
         }
+
+        .youtube-close-button {
+          height: @youtube-drag-handle-size;
+          display: inline-block;
+          background: fadeout(black, 50%);
+          color: @white;
+        }
       }
     }
   }
--- a/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
@@ -15,6 +15,7 @@ import {
   faArrowsAlt,
   faWindowMaximize,
   faWindowMinimize,
+  faTimes,
 } from "@fortawesome/free-solid-svg-icons";
 import { faYoutube } from "@fortawesome/free-brands-svg-icons";
 import { Link } from "react-router-dom";
@@ -30,6 +31,7 @@ import { dataSourcesInfo } from "../../settings/brainzplayer/BrainzPlayerSetting
 export type YoutubePlayerProps = DataSourceProps & {
   youtubeUser?: YoutubeUser;
   refreshYoutubeToken: () => Promise<string>;
+  onClose?: () => void;
 };
 
 // For some reason Youtube types do not document getVideoData,
@@ -361,7 +363,7 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
   };
 
   render() {
-    const { show } = this.props;
+    const { show, onClose } = this.props;
     const options: Options = {
       playerVars: {
         controls: 0,
@@ -380,6 +382,19 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
     const leftBound =
       document.body.clientWidth - draggableBoundPadding * 2 - 350;
 
+    // Handle close button click
+    const handleClose = () => {
+      if (this.youtubePlayer) {
+        this.youtubePlayer.stopVideo();
+        // Clear playlist
+        this.youtubePlayer.cueVideoById("");
+      }
+      // Hide the player by updating the show prop in the parent component
+      if (this.props.onVisibilityChange) {
+        this.props.onVisibilityChange(false);
+      }
+    };
+
     return (
       <Draggable
         handle=".youtube-drag-handle"
@@ -395,6 +410,13 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
         >
           <button className="btn btn-sm youtube-drag-handle" type="button">
             <FontAwesomeIcon icon={faArrowsAlt} />
+          </button>
+          <button
+            className="btn btn-sm youtube-close-button"
+            type="button"
+            onClick={onClose}
+          >
+            <FontAwesomeIcon icon={faTimes} />
           </button>
           <YouTube
             className="youtube-player"
--- a/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
@@ -855,6 +855,25 @@ export default function BrainzPlayer() {
     });
   };
 
+  const handleCloseYoutubePlayer = () => {
+    // Stop the playback first
+    if (!brainzPlayerContextRef.current.playerPaused) {
+      const dataSource =
+        dataSourceRefs[brainzPlayerContextRef.current.currentDataSourceIndex]
+          ?.current;
+      if (dataSource) {
+        dataSource.togglePlay();
+      }
+    }
+
+    // Reset player state
+    stopPlayerStateTimer();
+    reinitializeWindowTitle();
+
+    // Hide the player
+    dispatch({ isActivated: false });
+  };
+
   const playNextListenFromQueue = (datasourceIndex: number = 0): void => {
     const currentPlayingListenIndex =
       brainzPlayerContextRef.current.currentListenIndex;
@@ -1070,6 +1089,7 @@ export default function BrainzPlayer() {
             handleError={handleError}
             handleWarning={handleWarning}
             handleSuccess={handleSuccess}
+            onClose={handleCloseYoutubePlayer}
           />
         )}
         {userPreferences?.brainzplayer?.soundcloudEnabled !== false && (


frontend/css/brainzplayer.less
```diff
@@ -114,6 +114,13 @@
             cursor: grabbing;
           }
         }
+
+        .youtube-close-button {
+          height: @youtube-drag-handle-size;
+          display: inline-block;
+          background: fadeout(black, 50%);
+          color: @white;
+        }
```
MonkeyDo, line 123: This just seems to duplicate the styles for `.youtube-reduce-button`, which we don't want.
The easiest would be to rename the original class to make it more generic, something like `.youtube-button` or the like.

frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
```diff
@@ -380,6 +382,19 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
...
 
+    // Handle close button click
+    const handleClose = () => {
+      if (this.youtubePlayer) {
+        this.youtubePlayer.stopVideo();
+        // Clear playlist
+        this.youtubePlayer.cueVideoById("");
+      }
+      // Hide the player by updating the show prop in the parent component
+      if (this.props.onVisibilityChange) {
```
MonkeyDo, line 393: See the linting errors here, the `onVisibilityChange` prop is not defined in the  `YoutubePlayerProps` type.

I also don't think you need the `if` check here once you defined the prop in the type.

So this handleClose function should reduce to just calling `this.props.onVisibilityChange(false);`

frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
```diff
@@ -380,6 +382,19 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
     const leftBound =
       document.body.clientWidth - draggableBoundPadding * 2 - 350;
 
+    // Handle close button click
+    const handleClose = () => {
+      if (this.youtubePlayer) {
+        this.youtubePlayer.stopVideo();
+        // Clear playlist
+        this.youtubePlayer.cueVideoById("");
+      }
```
MonkeyDo, line 391: From reading the component, it looks like this code is already taken care of in `componentDidUpdate` when the value of the `show` prop changes to a falsy value:
https://github.com/metabrainz/listenbrainz-server/blob/9c314adb66903f858b7346339c0740a9474bdf8a/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx#L126-L130

So no need to duplicate the code here, as calling `onVisibilityChange` will make the parent component change the `show` prop.

MonkeyDo (changes requested):
Thanks for working on this feature! It's a good improvement that users are going to be happy about :)

OSinA04:
Hi @MonkeyDo, thanks for the review!

Apologies for the delayed response, I was caught up with project submissions and in the middle of moving out of the hostel, so things got a bit hectic. I really appreciate your patience.

I'll go ahead and implement the changes you requested ASAP.

Thanks again!

frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
OSinA04: Thanks for pointing that out, I missed it.

I've added `onVisibilityChange` prop to the `YoutubePlayerProps` type but I've made it optional so it doesn't break stuff - the `if` check is needed.

The `handleClose` function has been reduced to [code](https://github.com/metabrainz/listenbrainz-server/pull/3263/files#diff-7b2db647ed6bef3cf73281269f19be467d4b3f0e571c1ec77f660a0dec132bffR387-R392)

frontend/css/brainzplayer.less
OSinA04: Thanks for the suggestion!

I've renamed @youtube-drag-handle-size to @youtube-button-size to reflect its more general use. Also, I’ve refactored the styles by renaming `.youtube-reduce-button` to `.youtube-button`, and removed the duplicate `.youtube-close-button` class. The close button now uses the `.youtube-button` class.

frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
```diff
@@ -30,6 +31,8 @@ import { dataSourcesInfo } from "../../settings/brainzplayer/BrainzPlayerSetting
 export type YoutubePlayerProps = DataSourceProps & {
   youtubeUser?: YoutubeUser;
   refreshYoutubeToken: () => Promise<string>;
+  onClose?: () => void;
+  onVisibilityChange?: (visible: boolean) => void;
 };
```
MonkeyDo, line 36: For reference, I think one of those was unused.
Now with the other changes I propose we do not need any new prop.
Instead, we will add a local state to the YoutubePlayer component:
```suggestion
};

type YoutubePlayerState = {
  hidePlayer?: boolean;
};
```

To make use of this new `YoutubePlayerState` type, modify the class definition like so:

```tsx
export default class YoutubePlayer
  extends React.Component<YoutubePlayerProps, YoutubePlayerState>
  ```
  
Because this is one of our few remaining old-style class components, you will also need to initialize the state in a class constructor; add this right before the `componentDidUpdate` function:
```tsx
  constructor(props: YoutubePlayerProps) {
    super(props);
    this.state = { hidePlayer: true };
  }
```

And finally, in `componentDidUpdate`, two changes to use the `playerPaused` prop:
use the prop like so: `const { show, volume, playerPaused } = this.props;`
and add this condition block after the existing ones:
```tsx
    if (prevProps.playerPaused && !playerPaused) {
      // Show player if playing music
      this.setState({ hidePlayer: false });
    }
```
Here, we react to changes to the playerPaused prop and show the player if it was previously paused and is now unpaused. This allows us to click the button on the LB player to resume playback *and show the youtubeplayer*

frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
```diff
@@ -380,6 +383,14 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
...
       document.body.clientWidth - draggableBoundPadding * 2 - 350;
 
+    // Handle close button click
+    const handleClose = () => {
+      const { onVisibilityChange } = this.props;
+      if (onVisibilityChange) {
+        onVisibilityChange(false);
+      }
+    };
+
```
MonkeyDo, line 393: Some changes here to use our new local state, and call the internal togglePlay function:

```tsx
    const handleHide = () => {
      this.setState({ hidePlayer: true }, this.togglePlay);
    };

    const isPlayerVisible = show && !hidePlayer;
```
We also create a new `isPlayerVisible` variable that combine the activation state (`show` prop) and our new internal state `hidePlayer` value.
(note that I changed the name of the function to better describe the mechanism. we are hiding rather than closing)

I can't comment on the relevant line below, so doing it here instead: you'll want to use this new isPlayerVisible to apply the "hidden" class instead of using the "show" prop.
While you are in there you can also fix the `data-testid` attribute which doesn't need the hidden class (a mistake from before):
```
        <div
          className={`youtube-wrapper${!isPlayerVisible ? " hidden" : ""}`}
          data-testid="youtube-wrapper"
        >
```

frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
```diff
@@ -396,6 +407,13 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
           <button className="btn btn-sm youtube-drag-handle" type="button">
             <FontAwesomeIcon icon={faArrowsAlt} />
           </button>
+          <button
+            className="btn btn-sm youtube-button"
+            type="button"
+            onClick={onClose}
```
MonkeyDo, line 413: And finally, this will now use `handleHide`:
```suggestion
            onClick={handleHide}
```

MonkeyDo (changes requested):
Nice work on the CSS changes, that's now done.

After some testing of the feature, I realized that if I click the close button and then press play on the LB player, it will skip to the next song instead of continuing where I left it.
That is not quite what I am looking for as a user, and I realize some of my previous suggestions may have participated in that.

So I've tried our options to get this working as a pause button that hides the YT player.

The good news is all the changes we will need in the end are in the YoutubePlayer component, so quite self-contained, and the result works great.

MonkeyDo:
Hello @ojasinha !
Please let me know if you are still interested in working on this, otherwise I'll be happy to make the remaining changes to get it merged. Thanks !

ojasinha:
Hey. I'm sorry, I totally forgot about this. I'll make the changes on the weekend. Thanks for being patient with me!

MonkeyDo (approved):
Thanks you for finishing this off! Good addition, and it's working great now.

I took the liberty to fix a SCSS issue and tweak the CSS a bit.

PR #3264, opened 2025-05-07
Add CRON_SENTRY_DSN to use self-hosted sentry for cron monitors

amCap1712:


mayhem (approved):
Looks good, but I wonder if this PR is enough to move the cron monitoring back to the old setup. Is there anything else that needs doing? (How does it know to look at the DSN section of LOG_SENTRY?

amCap1712:
Discussed in matrix.

PR #3269, opened 2025-05-10
Added a checkbox to make multiple listens or albums keeping the modal open 

RayyanSeliya:
# Problem
[LB-1759](https://tickets.metabrainz.org/browse/LB-1759)  
The problem is when a user wants to add multiple listen in  row (like for each side of a record ) or multiple albums with different timestamps for example ,so the user have to  first submit the listen he is listening to and for adding another listen or album user had to again open the modal and add manually the listen or album he wants to, so this is very frustrating to  open the modal again and again repeatedly, so the user wants an option which help him to keep the modal open so they can add multiple listens with very few clicks

Can see what was earlier :
image

# Solution
I added a checkbox named "Add another" to the Add Listens modal in  `AddListenModal.tsx`  . When a user  tick this add another check box, the modal stays open after they  submit a listen, so they can quickly add more without reopening it each time. If they leave it unchecked, the modal closes as usual. This makes adding multiple listens much faster and provides more control, similar to GitHub’s “Create more” option.

I have created a checkbox as 
``` typeScript
            <div
            className="modal-footer"
            style={{ display: "flex", alignItems: "center", gap: "1rem" }}
             >
            <div style={{ flex: 1 }}>
              <label style={{ userSelect: "none", cursor: "pointer" }}>
                <input
                  type="checkbox"
                  checked={keepModalOpen}
                  onChange={(e) => setKeepModalOpen(e.target.checked)}
                  style={{ marginRight: "0.5em" }}
                />
                Add another
              </label>
            </div>
```
The preview of the solution is in the video below :

https://github.com/user-attachments/assets/1529460f-70a5-4e05-ba18-edd6b3399ed9

### DIFF
--- a/frontend/js/src/user/components/AddAlbumListens.tsx
+++ b/frontend/js/src/user/components/AddAlbumListens.tsx
@@ -5,6 +5,8 @@ import React, {
   useMemo,
   useRef,
   useState,
+  forwardRef,
+  useImperativeHandle,
 } from "react";
 import { toast } from "react-toastify";
 import {
@@ -95,11 +97,10 @@ export function TrackRow({ track, isChecked, onClickCheckbox }: TrackRowProps) {
   );
 }
 
-export default function AddAlbumListens({
-  onPayloadChange,
-  switchMode,
-  initialText,
-}: AddAlbumListensProps) {
+const AddAlbumListens = forwardRef(function AddAlbumListens(
+  { onPayloadChange, switchMode, initialText }: AddAlbumListensProps,
+  ref
+) {
   const { APIService } = useContext(GlobalAppContext);
   const { lookupMBRelease } = APIService;
   const [selectedAlbumMBID, setSelectedAlbumMBID] = useState<string>();
@@ -244,6 +245,15 @@ export default function AddAlbumListens({
     { format: ["minutes", "seconds"] }
   );
 
+  useImperativeHandle(ref, () => ({
+    reset: () => {
+      setSelectedAlbumMBID(undefined);
+      setSelectedAlbum(undefined);
+      setSelectedTracks([]);
+      searchInputRef.current?.triggerSearch("");
+    },
+  }));
+
   return (
     <div>
       <SearchAlbumOrMBID
@@ -252,6 +262,7 @@ export default function AddAlbumListens({
         }}
         switchMode={switchMode}
         ref={searchInputRef}
+        requiredInput={selectedAlbumMBID === undefined}
       />
       <div className="track-info">
         {selectedAlbum && (
@@ -348,4 +359,6 @@ export default function AddAlbumListens({
       </div>
     </div>
   );
-}
+});
+
+export default AddAlbumListens;
--- a/frontend/js/src/user/components/AddAlbumListens.tsx
+++ b/frontend/js/src/user/components/AddAlbumListens.tsx
@@ -108,11 +108,7 @@ const AddAlbumListens = forwardRef(function AddAlbumListens(
   const [selectedTracks, setSelectedTracks] = useState<Array<MBTrackWithAC>>(
     []
   );
-  const searchInputRef = useRef<{
-    focus(): void;
-    triggerSearch(newText: string): void;
-    reset(): void;
-  }>(null);
+  const searchInputRef = useRef<SearchInputImperativeHandle>(null);
 
   const initialTextRef = useRef(initialText);
   React.useEffect(() => {
--- a/frontend/js/src/user/components/AddAlbumListens.tsx
+++ b/frontend/js/src/user/components/AddAlbumListens.tsx
@@ -111,6 +111,7 @@ const AddAlbumListens = forwardRef(function AddAlbumListens(
   const searchInputRef = useRef<{
     focus(): void;
     triggerSearch(newText: string): void;
+    reset(): void;
   }>(null);
 
   const initialTextRef = useRef(initialText);
@@ -250,7 +251,7 @@ const AddAlbumListens = forwardRef(function AddAlbumListens(
       setSelectedAlbumMBID(undefined);
       setSelectedAlbum(undefined);
       setSelectedTracks([]);
-      searchInputRef.current?.triggerSearch("");
+      searchInputRef.current?.reset();
     },
   }));
 
--- a/frontend/js/src/user/components/AddListenModal.tsx
+++ b/frontend/js/src/user/components/AddListenModal.tsx
@@ -101,12 +101,17 @@ export default NiceModal.create(() => {
   const [customTimestamp, setCustomTimestamp] = useState(false);
   const [selectedDate, setSelectedDate] = useState(new Date());
   const [invertOrder, setInvertOrder] = useState(false);
+  const [keepModalOpen, setKeepModalOpen] = useState(false);
   // Used for the automatic switching and search trigger if pasting URL for another entity type
   const [textToSearch, setTextToSearch] = useState<string>();
 
   const closeModal = useCallback(() => {
     modal.hide();
     document?.body?.classList?.remove("modal-open");
+    const backdrop = document?.querySelector(".modal-backdrop");
+    if (backdrop) {
+      backdrop.remove();
+    }
     setTimeout(modal.remove, 200);
   }, [modal]);
 
@@ -186,7 +191,14 @@ export default NiceModal.create(() => {
           />,
           { toastId: "added-listens-success" }
         );
-        closeModal();
+
+        if (!keepModalOpen) {
+          closeModal();
+        } else {
+          // Reset the form state but keep the modal open
+          setSelectedListens([]);
+          setTextToSearch("");
+        }
       } catch (error) {
         handleError(
           error,
@@ -211,6 +223,7 @@ export default NiceModal.create(() => {
     APIService,
     closeModal,
     handleError,
+    keepModalOpen,
   ]);
 
   const switchMode = React.useCallback(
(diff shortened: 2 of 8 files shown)


RayyanSeliya:
@amCap1712  if any changes needed do let me know

frontend/js/src/user/components/AddListenModal.tsx
```diff
@@ -369,19 +382,31 @@ export default NiceModal.create(() => {
...
               type="button"
               className="btn btn-default"
-              data-dismiss="modal"
               onClick={closeModal}
             >
               Close
             </button>
             <button
-              type="submit"
+              type="button"
```
MonkeyDo, line 408: Is this change required? Just out of curiosity.

frontend/js/src/user/components/AddListenModal.tsx
```diff
@@ -186,7 +191,14 @@ export default NiceModal.create(() => {
...
           { toastId: "added-listens-success" }
         );
-        closeModal();
+
+        if (!keepModalOpen) {
+          closeModal();
+        } else {
+          // Reset the form state but keep the modal open
+          setSelectedListens([]);
+          setTextToSearch("");
```
MonkeyDo, line 200: I have the impression that this is not working as expected?

At least in the video in the PR description it shows two selected listens one under the other, when I would expect clicking the 'submit' button to clear the current selection
image


I think this is possibly because some of the logic is in the child components AddSingleListen and AddAlbumListens ?

MonkeyDo (changes requested):
The mechanism I would expect as a user is this: after I checked the "add another" checkbox, when I click on the submit button it should clear the selected listen but keep the modal open.

It seems that this was your intention, but you'll need to dig a bit deeper as there are other interactions with sub-components .
Don't hesitate to ask me questions here that you might have.

RayyanSeliya:
@MonkeyDo  Thankyou for the review ,I'll look onto the suggested changes and queries tomorrow travelling right now 🙂!!

frontend/js/src/user/components/AddListenModal.tsx
RayyanSeliya: @MonkeyDo  Thanks for catching this ! You are right the selected listens are not being fully cleared after submitting, which is why multiple listens are showing up in the modal. I think this is because, even though I’m resetting the `selectedListens` state in the parent, the child components `AddSingleListen` and `AddAlbumListens` might not be resetting their own state or syncing properly.

I’ll dig into how the child components manage their state and make sure everything resets as expected when “Add another” is checked. If you have any suggestions for the best way to reset the child components, I would love to hear them! Otherwise, I will update the PR soon with a fix. Thanks again for your helpful feedback

frontend/js/src/user/components/AddListenModal.tsx
MonkeyDo: I see! The way to ensure the page is not reloaded is to add use the vent's `preventDefault` method.
A random example from the codebase: https://github.com/metabrainz/listenbrainz-server/blob/03d06f4dc2a8055be68d68bc8ecbf0c771658c31/frontend/js/src/components/Navbar.tsx#L26

Generally, keeping the type:submit is important for accessibility (for example for screen readers for visually impaired users).
The literature on accessibility guidelines is quite dry, but if you want more context here it is :) https://www.w3.org/WAI/WCAG21/Techniques/html/H32

MonkeyDo: Nice work :)
This should all work fine, but I suggest a small stylistic change:

For the form submission, you don't need a new handleFormSubmit function, you can directly pass the submitListens function like so:

This was exactly correct, no need to change it:
```
        <form className="modal-content" onSubmit={submitListens}>
```

Then modify the `submitListens` function based on the changes you already made:
```
const submitListens = useCallback(async (e: React.FormEvent<HTMLFormElement>) => {
      e.preventDefault();
      if (auth_token) {
      [...]
```

All I did was, like you, define the expected event type, and add the prevenDefault call right before the rest of the function.
It's essentially the same as what you did, but combining it into submitListens instead of defining a new handleFormSubmit function in-between.

--------------------------------

Regarding the required input, looks like I didn't think everything through ;)
But I still think this is a good accessibility change overall, so as you suggested the second option is the way to go:
"make requiredInput optional and give it a default value"
If you look in the SearchTrackOrMBID component this is how we deal with the autofocus prop:
https://github.com/metabrainz/listenbrainz-server/blob/7f44897ea5ec11fcf12863678f8c1522188cf28e/frontend/js/src/utils/SearchTrackOrMBID.tsx#L43-L55

frontend/js/src/user/components/AddListenModal.tsx
MonkeyDo: Forgot to paste this link about the `key` prop: https://react.dev/reference/react/useState#resetting-state-with-a-key

RayyanSeliya: hey @MonkeyDo I tried the first option to reset the form status by  passing a key prop for child components `AddSingleListen` and `AddAlbumListens` and it is not working !! , so now I will go ahead with the second option !!!

RayyanSeliya: @MonkeyDo  So After proceeding with the second option and doing required changes to  related files this is the conclusion you can see in the video below : 

https://github.com/user-attachments/assets/3151770f-bbd8-450a-b74e-d7c4fa120ceb

So can I go ahead to merge the commit for this pr ?
And thankyou for your efforts you put in this to guide me step by step I really appreciate it !!

MonkeyDo: Yes you can push the commit, I'll do another review.

RayyanSeliya: For sure !!

RayyanSeliya:
@MonkeyDo   I have pushed the new commits  let me know if any changes needed  [commit](https://github.com/metabrainz/listenbrainz-server/pull/3269/commits/eb0aedeb426ebbfaed01b7b613eedb4822bd2308)
I was facing that please fill out this field issue in Add Album  also so i have changed the files  `SearchTrackOrMBID.tsx`  and `SearchAlbumOrMBID.tsx`  accordingly by the second option suggested  requiredInput prop optional and give it a default value

frontend/js/src/user/components/AddAlbumListens.tsx
```diff
@@ -244,6 +245,15 @@ export default function AddAlbumListens({
     { format: ["minutes", "seconds"] }
   );
 
+  useImperativeHandle(ref, () => ({
+    reset: () => {
+      setSelectedAlbumMBID(undefined);
+      setSelectedAlbum(undefined);
+      setSelectedTracks([]);
+      searchInputRef.current?.triggerSearch("");
```
MonkeyDo, line 253: I think this won't do anything as it is, as the triggerSearch imperative handle function calls setInputValue, and the useEffect hook that reacts to changes in inputValue returns if the string value is falsy (including an empty string):
https://github.com/metabrainz/listenbrainz-server/blob/27577fd9eda8be6855f1735d655eafd6810a0b5d/frontend/js/src/utils/SearchAlbumOrMBID.tsx#L158-L162

I think I see in your video demo that the previous search results are still loaded in the dropdown after you submit a listen, and this is why.

However, there is already a reset function defined in there, which we can expose to the parent component with the same imperativehandle hook
https://github.com/metabrainz/listenbrainz-server/blob/27577fd9eda8be6855f1735d655eafd6810a0b5d/frontend/js/src/utils/SearchAlbumOrMBID.tsx#L150-L157

This would mean moving the reset function definition higher up so it is defined before the imperativehandle hook, wrap it in a useCallback hook to avoid unnecessary re-renders (reset will have to be a dependency of the imperativehandle hook), and expose it to the parent.
All in all, like this:
```tsx
const reset = useCallback(() => {
    setInputValue("");
    setSearchResults([]);
    onSelectAlbum();
    setLoading(false);
    searchInputRef?.current?.focus();
  }, [onSelectAlbum]);

  // Allow parents to focus on input and trigger search
  useImperativeHandle(
    inputRefForParent,
    () => {
      return {
        focus() {
          searchInputRef?.current?.focus();
        },
        triggerSearch(newText: string) {
          setInputValue(newText);
        },
        reset,
      };
    },
    [reset]
  );
```

frontend/js/src/user/components/AddSingleListen.tsx
```diff
@@ -47,6 +52,13 @@ export default function AddSingleListen({
     };
   }, [initialText]);
 
+  useImperativeHandle(ref, () => ({
+    reset: () => {
```
MonkeyDo, line 56: Along with my comment above, I think this is where you'll want to call the new function to reset the search component state
(`searchInputRef?.current?.reset()`)

MonkeyDo (changes requested):
I know I made this PR quite a bit more complicated than it needed to be originally, but I believe these are good improvements in flexibility and accessibility. Thanks for sticking around through the rounds of feedback!

RayyanSeliya:
> I know I made this PR quite a bit more complicated than it needed to be originally, but I believe these are good improvements in flexibility and accessibility. Thanks for sticking around through the rounds of feedback!

No worries at all I actually enjoyed working through the feedback and learned a lot in the process! I agree, these changes really do make the codebase more flexible and accessible. Thanks for all your guidance and for taking the time to review everything so thoroughly.

RayyanSeliya:
@MonkeyDo  I’ve updated the search components to expose the reset function through the imperative handle, and I’m now calling `searchInputRef.current?.reset()` from the parent after each submission. This seems to be working well it clears both the input and the search results just like we wanted. see the latest [commit ](https://github.com/metabrainz/listenbrainz-server/pull/3269/commits/e4bca5f15bbe72084412b8746834627b8856a539)

MonkeyDo (approved):
I hope you don't mind but I was pondering how to improve the typescript types situation with the changes you made, and thought I might as well commit the result once I got it working.

I deduplicated the type used in useRef and put it in a types file, then used it with the useRef and useImperativeHandle calls where they were needed.

MonkeyDo:
I deployed this to our test server for in-situ testing, everything works perfectly, nice work!

Merging this now, some users are going to be happy about this :)

RayyanSeliya:
> I deployed this to our test server for in-situ testing, everything works perfectly, nice work!
> 
> Merging this now, some users are going to be happy about this :)

Thank you @MonkeyDo  ! I really enjoyed working on this and appreciate your support throughout. It was a great experience collaborating with you ,looking forward to contributing more!

PR #3274, opened 2025-05-13
Refactor dumps creation, import and management code

amCap1712:
While working on the sample data dumps, I found myself copying a bunch of code over and over. Plus, the dump.py file was getting too big and gnarly. Hence, refactored into a separate module. This PR only rearranges the existing code. A future PR will attempt at removing some of the duplicate code.


mayhem (approved):
As discussed in matrix, approving this one in favor of reading a real PR later. 😬😬

PR #3275, opened 2025-05-13
 Refactor listens similarity data on lines of mlhd

amCap1712:
Listens similarity jobs failed in the past few weeks because of less space available on the spark cluster than usual because of the presence of MLHD dataset. Also, the listens are growing in number as well. To make listen similarity work again, use the same optimizations techniques as MLHD for listens and as a bonus get to deduplicate the sql queries between the two.


mayhem (approved):
so excited!

PR #3277, opened 2025-05-14
LB-1630: Fix dashboard for users with sporadic listens

fettuccinae:
# Problem

    Dashboard for users with sporadic listens loads slowly

# Solution

   limit the search window for initial load to 12 months.

### DIFF
--- a/frontend/js/src/user/Dashboard.tsx
+++ b/frontend/js/src/user/Dashboard.tsx
@@ -681,7 +681,8 @@ export default function Listen() {
               >
                 {listens.map(getListenCard)}
               </div>
-              {listens.length < expectedListensPerPage && (
+              {listens.length < expectedListensPerPage &&
+                nextListenTs === oldestListenTs && (
                 <h5 className="text-center">No more listens to show</h5>
               )}
               <ul className="pager" id="navigation">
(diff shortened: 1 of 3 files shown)


fettuccinae:
> For the front-end parts, we might need to merge in some of the commits from #3252 but I haven't looked at it closely.

I dont think we need to do it, If the user needs to fetch older listens, user can click older button which uses the last track's listened_at_ts as min_ts.

fettuccinae:
@MonkeyDo 
 > Hypothetical scenario: User is on page 1, page is full of recent listens. User navigates to page 2 by clicking "next". Listens past those of page 1 are all much older, so the API returns 0 listens for the limited time window. User is presented with a page with 0 listens, has to click "next" to (hopefully) get a page of results

This change ( getting listens only for a year) is only done for page 1. From page 2 till their last listen, there will be no difference in their dashboard, and on page 1. And we are guaranteed to have atleast one listen because of using latest_listened_ts as upper bound for searching listens.

> Plus we still have the issue that you fixed in the other PR where we currently show "No more listens to show" if we don't have enough listens to fill up the page, rather than checking against the user's oldest listen TS

Yup, i made a commit to fix it.

> I can think of some edge cases where that won't suffice, the simplest being if we just return a few listens, it'll be a verfy empty looking page.
Ideally we would always do our best effort to fill up the expected number of listens per page.

The number of users with less than 25 listens over the range of a year from their latest_listen_ts has to be very less. 
We can always ask them to listen to more music :)

PR #3278, opened 2025-05-15
Refactor dumps code

amCap1712:
Refactor dumps code for better readability and reducing redundancy. Individual commits explain the changes in more detail.


mayhem (approved):
Very good improvement -- I like how all the details of the dump are specified in one location now. Perhaps we can do the playlist improvement at a future point in time.

PR #3285, opened 2025-05-27
LB-1728: Hide lastfm warning when connected

MonkeyDo:
For some reason I thought this was already implemented, but it is not.
The ticket was reopened as it is confusing to have a warning about LFM privacy settings when you are already connected...
The warning is dismissable, but navigating away and back reverts it.
There are some good suggestions to refactor the UX of that page, but in the meantime this should be a good stopgap: when the user is connected, hide the warning.

@Aerozol I know it's not what you asked for, but it took me a full minute to make this PR and fix the confusion issue, at least.
I think we will indeed want to review the connect services page UX, but as we have GSOC projects that will add to it I suggest we wait until after the summer to see if we have more changes to improve those too

### DIFF
--- a/frontend/js/src/settings/music-services/details/MusicServices.tsx
+++ b/frontend/js/src/settings/music-services/details/MusicServices.tsx
@@ -438,29 +438,31 @@ export default function MusicServices() {
               Connect to your Last.FM account to import your entire listening
               history and automatically add your new scrobbles to ListenBrainz.
             </p>
-            <div
-              className="alert alert-warning alert-dismissible fade in"
-              role="alert"
-            >
-              You must first disable the &#34;Hide recent listening
-              information&#34; setting in your Last.fm{" "}
-              <a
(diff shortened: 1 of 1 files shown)


MonkeyDo:
Closed in favor of #3260 which implements the same change, but also adds a preview of the current status of the importer
image

MonkeyDo:
> No problem at all with waiting. That said, let's not leave it on the back burner forever! Considering how many lastfm import peeps we seem to have.

There was ongoing work in progress to add a status for these imports, and serendipitously lucifer finished reviewing that today.
I based the UI of this import status component on the existing one we have for the "Export data" page, for a more unified look and feel, but making it a bit slimmer to fit in better on the "Music services" page.
I think between these two changes (don't show warning when connected + show current status when connected) it should remove any confusion, but I'm open for another review of the whole process to see if we are happy with it, and if the same UX should be applied to other importers in the future.

PR #3289, opened 2025-06-01
LB-889: Integrate music streaming from Funkwhale

mAmanullah7:
## This pr is a part of my GSoC'25 project [Integrate music streaming from Funkwhale & Navidrome](https://community.metabrainz.org/t/integrate-music-streaming-from-funkwhale-navidrome-subsonic/751757)
Allow users to play music from their **Funkwhale** servers as well as **Navidrome** directly in
**BrainzPlayer**, both are self hoisted music streaming platform. **Funkwhale** uses a `OAuth2`
for secure and safe authentication. Currently **Navidrome** using basic subsonic
authentication `username/password + salt`.
LB-889

For **Navidrome** if `OAuth2` available will implement it instead of basic `username/password + salt` authentication

## Funkwhale UI:
Screenshot 2025-07-24 at 2 09 35 PM
Screenshot 2025-07-24 at 2 15 34 PM

### DIFF
--- a/listenbrainz/webserver/views/test/test_settings.py
+++ b/listenbrainz/webserver/views/test/test_settings.py
@@ -252,7 +252,7 @@ class SettingsViewsTestCase(IntegrationTestCase):
         
         # Mock session within the request context
         with self.client.session_transaction() as sess:
-            sess['funkwhale_state'] = test_state
+            sess['state'] = test_state
             sess['funkwhale_host_url'] = 'https://demo.funkwhale.audio'
         
         response = self.client.get(


mAmanullah7:
created `funkwhale_servers` table : to store user credentials(`host_url, tokens, etc`)

`model/funkwhale.py`: sqlalchemy model

`db/funkwhale.py`: server specific token queries

`domain/funkwhale.py`: `OAuth2` flow

`webserver/views/funkwhale.py`: api endpoints

@amCap1712 do let me know if there required any changes!

mAmanullah7:
We've updated funkwhale db schema --> now we are using two tables:
 1. `funkwhale_servers`: to store the `host_url`, `cliend_id`, `client_secret` and `scopes` can be reuse.
 2. `funkwhale_tokens`: to store the `access_token`, `refreshed_token`, and `token_expiry`. 
 
 removed `model/funkwhale.py` as we are not using sqlalchemy model anymore now we are using direct sql.

mAmanullah7:
@amCap1712 everything works fine for me as of now except the refresh token one!! You can have a look and do let me know if any changes required!!

mAmanullah7:
@MonkeyDo @anshg1214 i've tested locally and i'm able to play song from funkwhale servers you can have a look and do let me know if any changes required!!

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -428,10 +429,17 @@ function BrainzPlayerUI(props: React.PropsWithChildren<BrainzPlayerUIProps>) {
...
+                {currentDataSource?.name === "funkwhale" ? (
+                  <FunkwhaleIcon
+                    className="svg-inline--fa funkwhale-icon"
+                    color="#009FE3"
+                  />
+                ) : (
+                  <FontAwesomeIcon
+                    icon={currentDataSource.icon}
+                    color={currentDataSource.iconColor}
+                  />
```
MonkeyDo: I propose to try something I've never tried myself and create a custom icon component that emulates the FontAwesome icons exported from react-fontawesome (using their type definitions, in this case IconDefinition for example).

This would allow you to revert this change that special-cases funkwhale, and instead reference the custom icon in the [`dataSourcesInfo` object](https://github.com/metabrainz/listenbrainz-server/blob/934c99a5bbd6223612ee632572c306cf41c30759/frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx#L24) and in for the static property [FunkwhalePlayer.icon](https://github.com/metabrainz/listenbrainz-server/pull/3289/files#diff-b0f04850f6f6f81c1be4afbbcf4bb620751a848acff73640d48531c29087dcdfR61).


It looks like I'm not the only one to have had that idea, you can try to follow step 1 of this SO answer: https://stackoverflow.com/a/73366672/4904467

For the component's SVG path data, you will need to pass an array of strings, each strign corresponding to the value of the `d` property of each `<path>` svg element, should be this array of strings:
```js
["m 128,157.1 c 17.7,0 32.1,-14.4 32.1,-32.1 0,-0.9 -0.8,-1.7 -1.7,-1.7 h -12.1 c -0.9,0 -1.7,0.8 -1.7,1.7 0,9.1 -7.4,16.6 -16.6,16.6 -9.1,0 -16.6,-7.4 -16.6,-16.6 0,-0.9 -0.8,-1.7 -1.7,-1.7 H 97.6 c -0.9,0 -1.7,0.8 -1.7,1.7 0,17.8 14.4,32.1 32.1,32.1 z",
"m 128,187.4 c 34.3,0 62.3,-28 62.3,-62.3 0,-0.9 -0.8,-1.7 -1.7,-1.7 h -12.1 c -0.9,0 -1.7,0.8 -1.7,1.7 0,25.9 -21,46.9 -46.9,46.9 C 102,172 81,151 81,125.1 c 0,-0.9 -0.8,-1.7 -1.7,-1.7 H 67.4 c -0.9,0 -1.7,0.8 -1.7,1.7 -0.2,34.3 27.8,62.3 62.3,62.3 z",
"m 219,123.4 h -12.1 c -0.9,0 -1.7,0.8 -1.7,1.7 0,42.6 -34.8,77.3 -77.3,77.3 -42.6,0 -77.3,-34.6 -77.3,-77.3 0,-0.9 -0.8,-1.7 -1.7,-1.7 H 37 c -0.9,0 -1.7,0.8 -1.7,1.7 0,51.1 41.6,92.7 92.7,92.7 51.1,0 92.7,-41.6 92.7,-92.7 0,-0.9 -0.8,-1.7 -1.7,-1.7 z",
"m 86.3,83.3 c 6.2,3.2 12.9,3.8 18.9,7.3 3.9,2.3 6.4,4.8 8.8,8.6 3.8,5.7 3.6,12.9 3.6,12.9 l 0.5,7.9 c 0,0 3,7.9 9.7,7.9 7.1,0 9.7,-7.9 9.7,-7.9 l 0.5,-7.9 c 0,0 -0.2,-7.1 3.6,-12.9 2.4,-3.8 4.8,-6.5 8.8,-8.6 6,-3.5 12.7,-4.1 18.9,-7.3 6.2,-3.2 12.2,-7.3 16.3,-13 4.1,-5.7 6,-13.3 3.8,-20 -11.8,-0.6 -25.4,0.8 -35.8,6.4 -14.5,7.7 -23.3,5 -25.9,16.5 h -0.2 c -2.6,-11.6 -11.3,-8.8 -25.9,-16.5 -10.4,-5.6 -24,-7 -35.8,-6.4 -2.3,6.7 -0.3,14.2 3.8,20 4.4,5.8 10.5,9.9 16.7,13 z"]

```

frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx
```diff
@@ -406,10 +470,17 @@ function BrainzPlayerSettings() {
                   <FontAwesomeIcon icon={faGripLines as IconProp} />
                 </span>
                 <span>
-                  <FontAwesomeIcon
-                    icon={item.info.icon}
-                    color={item.info.color}
-                  />
+                  {item.id === "funkwhale" ? (
+                    <FunkwhaleIcon
```
MonkeyDo: Similarly to the comment above, should also make this special case unnecessary if it works.

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -177,6 +225,113 @@ export default function MusicServices() {
...
+  }, []);
+
+  React.useEffect(() => {
+    // Use a small timeout to ensure this runs after any initial renders
+    const timer = setTimeout(() => {
+      setIsInitialLoad(false);
+    }, 100);
+
+    return () => clearTimeout(timer);
+  }, []);
```
MonkeyDo: The state value `isInitialLoad` seems to be unused, so I assume this code should be removed as well as the line creating the isInitialLoad state

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -46,24 +52,62 @@ export default function MusicServices() {
...
+      // For other services or user-initiated actions, proceed normally
+      const fetchUrl = `/settings/music-services/${serviceName}/disconnect/`;
+      let fetchBody;
+      const fetchHeaders: Record<string, string> = {
+        "Content-Type": "application/json",
+      };
+
+      if (serviceName === "funkwhale" && newValue === "disable") {
+        fetchBody = undefined;
+        fetchHeaders.Authorization = `Token ${currentUser?.auth_token}`;
```
MonkeyDo: All these code changes are starting to look very messy.

I think one of two things should happen:
1. refactor custom FW connection code so that the `handleFunkwhaleConnect` handles both connection and disconnection, to keep it separate from existing logic.
2. Make the backend endpoint for funkwhale accept the "disable" action like the other endpoints (ignoring it if need be) so that you don't need to special-case funkwhale here

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -46,24 +52,62 @@ export default function MusicServices() {
...
-          body: JSON.stringify({ action: newValue }),
-          headers: {
-            "Content-Type": "application/json",
-          },
-        }
-      );
+      setUserInteractionStarted(true);
+
+      // For Funkwhale disable action during initial load, don't call backend
+      // This prevents automatic disconnection when component loads with "disable" state
```
MonkeyDo: Despite the comments I don't understand why this extra logic and state value are needed. Can you talk me through it? there might be a cleaner way to implement this.

I might be misunderstanding the issue, but setUserInteractionStarted should not be called on page load, and for FW it seems to only be called for the disconnection action.

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -93,6 +137,10 @@ export default function MusicServices() {
             if (critiquebrainzAuth) critiquebrainzAuth.access_token = undefined;
             break;
           // lastfm and librefm state is now managed in the LFMMusicServicePermissions component
+          case "funkwhale":
+            setFunkwhaleHostUrl("");
+            setConnectedFunkwhaleServers([]);
```
MonkeyDo: Why is this stored in the component state instead of the global context like we do for other music services in the lines just above?

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -177,6 +225,113 @@ export default function MusicServices() {
...
+          title="Funkwhale Connection Error"
+          message={decodeURIComponent(funkwhaleError)}
+        />
+      );
+      // Clear the query parameters from the URL
+      window.history.replaceState(
+        {},
+        document.title,
+        window.location.pathname + window.location.hash
+      );
```
MonkeyDo: This code is repeated in both if/else code blocks, so it can simply be deduplicated and moved to after the if/else block to be run in both cases.

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -177,6 +225,113 @@ export default function MusicServices() {
...
+      if (contentType && contentType.includes("application/json")) {
+        data = await response.json();
+      } else {
+        throw Error("Server returned non-JSON response");
+      }
+
+      if (response.ok) {
+        if (data.url) {
+          // Store host URL in session storage before redirect
+          sessionStorage.setItem("funkwhale_host_url", funkwhaleHostUrl);
```
MonkeyDo: Similarly to comment above, why store this in sessionstorage instead of the global context like we do for other music services?
Of course that would mean storing this data in the database and passing it to the front-end like we do for others; is this where the issue lies?

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -401,6 +556,91 @@ export default function MusicServices() {
...
+            >
+              <strong>Important:</strong> You must be already logged into your
+              Funkwhale server before connecting it to ListenBrainz.
+              <button
+                type="button"
+                className="btn-close"
+                data-bs-dismiss="alert"
+                aria-label="Close"
+              />
+            </div>
```
MonkeyDo: Let's hide the warning when connected (I recently did the same for LastFM)
```suggestion
            {permissions.funkwhale !== "listen" && (
              <div
                className="alert alert-warning alert-dismissible fade show"
                role="alert"
              >
                <strong>Important:</strong> You must be already logged into your
                Funkwhale server before connecting it to ListenBrainz.
                <button
                  type="button"
                  className="btn-close"
                  data-bs-dismiss="alert"
                  aria-label="Close"
                />
              </div>
            )}
```

MonkeyDo (commented):
There are obviously some things I'm not understanding in the FW connection code, but there might be some ways to simplify the code.

frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx
```diff
@@ -78,6 +87,10 @@ function BrainzPlayerSettings() {
     userPreferences?.brainzplayer?.appleMusicEnabled ??
       AppleMusicPlayer.hasPermissions(appleAuth)
   );
+  const [funkwhaleEnabled, setFunkwhaleEnabled] = React.useState(
+    userPreferences?.brainzplayer?.funkwhaleEnabled ??
+      FunkwhalePlayer.hasPermissions(funkwhaleAuth)
+  );
```
MonkeyDo, line 94: Commenting here 'cause I can't comment below.


I have a solution to fix the missing funkwhale service in the list of priorities:
```tsx
const [dataSourcesPriority, setDataSourcesPriority] = React.useState<
    DataSourceKey[]
  >(
    userPreferences?.brainzplayer?.dataSourcesPriority
      ? [
          ...userPreferences?.brainzplayer?.dataSourcesPriority,
          ...defaultDataSourcesPriority.filter(
            (item) =>
              !userPreferences.brainzplayer?.dataSourcesPriority?.includes(item)
          ),
        ]
      : defaultDataSourcesPriority
  );
```

In short? we combine the custom order (which may be incomplete) with the default list of services.
This solution should also be used by BrainzPlayer otherwise the datasource may never be added to the list of datasources. Will comment in BP component.

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -175,6 +226,9 @@ export default function BrainzPlayer() {
     appleMusicEnabled &&
       AppleMusicPlayer.hasPermissions(appleAuth) &&
       "appleMusic",
+    funkwhaleEnabled &&
```
MonkeyDo, line 230: For the issue of new music services not appearing in the list of priorities (and because we use that list of priorities as the base for filtering enabled datasources), you can use this bit of code right after defining `sortedDataSources` on line 242:

```tsx
  // If a new service is added after the user sets their prefered priority list,
  // it will not appear in the list and we need to manually add it otherwise it will be ignored.
  const missingEnabledDataSources = difference(
    enabledDataSources,
    sortedDataSources
  );
  if (missingEnabledDataSources.length) {
    sortedDataSources.push(...missingEnabledDataSources);
  }
```

`difference` utility is imported from lodash


There might be a better solution that does not rely on the priority list but instead uses the enabledDataSources array and reorders it according to the dataSourcesPriority array, that way we would not have to do this manual modification. I'll let you explore that :P

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
mAmanullah7: I just saw you commented this last week really sorry if i missed it when i checked day before i didn't find any!

about the icon i'll try to create a custom one, i'll look into the reference you've provided and will update u!!

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: All this changes will be removed including above & below comment! We will be using simple approach like other services `listen`, `disable` thats it!

all this issues happened after we've updated our db schema, before i was checking `host_url` exist for user and updating connect/disable status! if no `host_url` means its trigger disconnect! as there is only one `funlwhale_servers` table.

Now we are using two tables:
`funkwhale_servers`: to store the host_url, cliend_id, client_secret and scopes can be reuse.
`funkwhale_tokens`: to store the access_token, refreshed_token, and token_expiry

to avoid the multiple oauth app for same host_url we used this approach, like me and you both are using same host_url we dont need to create multiple oauth app and no need to store `client_id` and `client_secret` for every user rather use same oauth app to every user! 

Now i'm totally relying on access_token to check and update the connect/disable everything was going smooth enough but then boom there is an issue in `refresh_token` and then i tried to use other approach and tried to even debug but didn't find any! If you can suggest any approach if `token_expires` and  `refresh_token` failed then how we should handle connect/disable

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: Was about to ask this questions! I tried to use global context like we do but here we have variable `host_url` rather having fixed one like others!! Last time when i tried using global context i was getting some issue in `callback` either `host_url` handling or token exchange!! i was getting some errors i don't remember exactly!

We have to preserve the host_url until user is connected if anything happens in bw the oauth flow with `host_url` simply getting some error like invalid_access token!! i'll try to debug this one!! any suggestions to avoid this errors?

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: as i above mentioned while using global context i was getting some invalid access token! idk whats the actual issue was but when i used session it was working fine for me so i used this temporarily!

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
MonkeyDo: Not your fault at all!
I wrote the comment but did not post it. I was waiting to do a full review instead of just sending this one message, to be considerate of your time.

frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx
mAmanullah7: I also thoght similarly we can dynamically update this and i think its better approach! even a new services will be added or deleted is should be automatically added or removed!

```typescript
const [dataSourcesPriority, setDataSourcesPriority] = React.useState<
    DataSourceKey[]
      >(() => {
        const savedPriority = userPreferences?.brainzplayer?.dataSourcesPriority;
    
        // If no saved preferences, use default
        if (!savedPriority) {
          return defaultDataSourcesPriority;
        }
    
        // Ensure all data sources from defaultDataSourcesPriority are included
        // This handles cases where new data sources (like Funkwhale) were added after user saved preferences
        const migrated = [...savedPriority];
    
        // Add any missing data sources at the end
        defaultDataSourcesPriority.forEach((source) => {
          if (!migrated.includes(source)) {
            migrated.push(source);
          }
          return migrated.filter((source) =>
            defaultDataSourcesPriority.includes(source)
          );
     });
```

smtng like that i was thinking if its worked we can update this else everytime we have to add manually even while impelmenting Navidrome!

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
mAmanullah7: rather including manually in both places we can dynamically add  newly added services in bpSettings let rest same!? any suggestionss?

MonkeyDo: I'm not sure I understand your suggestion. There are two separate issues resulting from the same problem (both based on the value of userPreferences.brainzplayer.dataSourcesPriority stored in the database which will be missing new services added).

- The BrainzplayerSettings changes allow displaying FW in the drag-and-drop list of datasource priorities (it is based on userPreferences?.brainzplayer?.dataSourcesPriority)
- The BrainzPlayer changes allow adding FW as a datasource for playing music

Fixing one does not fix the other, so I do think you do need both changes.


The other option I see are:
- do this change (add missing services in prefs) in [getPageProps](https://github.com/metabrainz/listenbrainz-server/blob/1b946969260ed42c39731d0a5a308b4b7ef86624/frontend/js/src/utils/utils.tsx#L601-L605) instead which is the function that creates the GlobalContext. That way this manipulation is only done in one place.
- instead of doing these manipulation in the front-end, run a database script to add the new service to all the entries in the user_setting table

frontend/js/src/settings/music-services/details/MusicServices.tsx
MonkeyDo: I'm sorry, I didn't understand what the error was or why storing in globalcontext did not work. Can you please explain again with more context (I have very little knowledge into the back-end parts of your project)?

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
mAmanullah7: i'll look into ur suggestions!! i'll try to explain my approach one more time!
both bpSettings and bp use same `dataSourcesPriority` as i can see if i'm not wrong! 
i was thinking  rather manually handling at both places we can simply add newly added services in `defaultDataSourcesPriority` and then also check `userPreferences` if it exist then will simply try to compare both `defaultDataSourcesPriority` and `userPreferences` and if any addition in `defaultDataSourcesPriority` will push this to `userPreferences` in last!  and then we dont need to update anything to bp it will be taken care automatically  , else `defaultDataSourcesPriority` and here we already have added so no need to worry! even if any services deleted can be handled from here itself just we need to care about `defaultDataSourcesPriority` 


```typescript
  const [dataSourcesPriority, setDataSourcesPriority] = React.useState<
    DataSourceKey[]
  >(() => {
    const savedPriority = userPreferences?.brainzplayer?.dataSourcesPriority;

    // If no saved preferences, use default
    if (!savedPriority) {
      return defaultDataSourcesPriority;
    }

    // Ensure all data sources from defaultDataSourcesPriority are included
    // This handles cases where new data sources (like Funkwhale) were added after user saved preferences
    const migrated = [...savedPriority];

    // Add any missing data sources at the end
    defaultDataSourcesPriority.forEach((source) => {
      if (!migrated.includes(source)) {
        migrated.push(source);
      }
    });

    return migrated.filter((source) =>
      defaultDataSourcesPriority.includes(source)
    );
  });
```

in above comment i've returned inside the loop itslef may be you misunderstood! it was just an syntaxerror!


Even your approach looks good to me!

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: About backend rest everything is same except the `host_url` like for spotify we are using `OAUTH_TOKEN_URL = 'https://accounts.spotify.com/api/token'` which is hardcoded so we don't need to take care about session, state, etc every time  one oauth app is enough!
In case of fw its varying and its a user input which we have to use and from there problem starts, like mb have own fw host_url like that!
i even not checked after updating db schema i was using this approach rather global as it was working fine for me!
 
Screenshot 2025-06-09 at 3 13 44 PM 

this is the error i was getting smtng state or session related i'm not sure about this! but now i can try using global context becoz now we have updated our approach! 
if u can suggest any other way to handle?

listenbrainz/domain/funkwhale.py
```diff
@@ -0,0 +1,239 @@
...
+from listenbrainz.webserver import db_conn
+from listenbrainz.domain.external_service import ExternalServiceError, ExternalServiceAPIError, ExternalServiceInvalidGrantError
+
+import sqlalchemy
+
+FUNKWHALE_API_RETRIES = 5
+
+class FunkwhaleService:
+    def __init__(self):
+        self.redirect_url = current_app.config.get('FUNKWHALE_CALLBACK_URL',
```
amCap1712: No need for a fallback here, because this setting will be required for using funkwhale.

listenbrainz/webserver/views/funkwhale.py
```diff
@@ -0,0 +1,366 @@
...
+from listenbrainz.domain.funkwhale import FunkwhaleService
+from listenbrainz.webserver.decorators import crossdomain
+from listenbrainz.webserver.views.api_tools import validate_auth_header
+from listenbrainz.webserver.errors import APIBadRequest, APINotFound, APIInternalServerError, APIUnauthorized, APIForbidden, APIServiceUnavailable
+from listenbrainz.domain.external_service import ExternalServiceError, ExternalServiceAPIError, ExternalServiceInvalidGrantError
+from listenbrainz.db import funkwhale as db_funkwhale
+
+funkwhale_api_bp = Blueprint('funkwhale_api_v1', __name__)
+
+def validate_funkwhale_url(url: str) -> str:
```
amCap1712: Don't implement these endpoints separately but integrate them in existing external service connections. For callback, the endpoint already exists so the callback code here should be merged into that. Note that the callback URL will change (profile vs settings at start of the url).

listenbrainz/db/funkwhale.py
```diff
@@ -0,0 +1,104 @@
...
+        if existing_server.get('client_id') and existing_server.get('client_secret'):
+            return existing_server['id']
+    
+    # If no server exists or it has missing credentials, create/update it
+    result = db_conn.execute(sqlalchemy.text("""
+        INSERT INTO funkwhale_servers (host_url, client_id, client_secret, scopes)
+        VALUES (:host_url, :client_id, :client_secret, :scopes)
+        ON CONFLICT (host_url) DO UPDATE SET
+            client_id = CASE 
+                WHEN funkwhale_servers.client_id IS NULL OR funkwhale_servers.client_id = ''
```
amCap1712: `funkwhale_servers.client_id` and `funkwhale_servers.client_secret` should be made not-nullable and the queries will also become simpler as a result.

listenbrainz/domain/funkwhale.py
```diff
@@ -0,0 +1,239 @@
...
+        return user
+
+    def add_new_user(self, user_id: int, host_url: str, token: dict, client_id: str, client_secret: str, scopes: str) -> bool:
+        access_token = token['access_token']
+        refresh_token = token['refresh_token']
+        expires_at = int(time.time()) + token['expires_in']
+        from datetime import timezone
+        token_expiry_datetime = datetime.fromtimestamp(expires_at, tz=timezone.utc)
+        # Get user details from Funkwhale API
+        headers = {'Authorization': f'Bearer {access_token}'}
```
amCap1712: Is this response used anywhere? The token data can be assumed to be valid otherwise.

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
MonkeyDo: So I think we are talking about the same thing, but I disagree that a single change in the BrainzPlayerSettings component will solve the situation for the BrainzPlayer component as well.
The reason is that if I don't visit the BPSettings page, this code is never run.

So to be clearer here is my scenario:
1. I previously had changed my preferred order of music sources in [BPSettings](https://listenbrainz.org/settings/brainzplayer/), and this list is store in the database
2. I open LB, then go to the [MusicServices page](https://listenbrainz.org/settings/music-services/details/) and connect to my Funkwhale instance.
4. I go back to my dashboard. At that point I have FunkWhale in [`enabledDataSources`](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx#L173), but it is missing in [`dataSourcesPriority`](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx#L184) since we load the userPreferences from the database and I have not changed my BPSettings
5. I click play on a listencard, nothing happens (or to be clear, it does not play with funkwhale). The reason here is that `sortedDataSources` will not contain funkwhale because we take the `dataSourcesPriority` and filter that

From my understanding, making the change only in BPSettings is not enough to make BrainzPlayer work with a new music service.

Now my proposed solution:
We can go to the root of all this data and change the value of userPreferences.brainzplayer.dataSourcesPriority in [getPageProps](https://github.com/metabrainz/listenbrainz-server/blob/1b946969260ed42c39731d0a5a308b4b7ef86624/frontend/js/src/utils/utils.tsx#L601-L605) which is the function that builds the GlobalContext values.
That way, since both the BrainzPlayerSettings and the BrainzPlayer components use that `dataSourcesPriority` value, it should solve the issue for both.

listenbrainz/domain/funkwhale.py
mAmanullah7: not used anywhere just used for debugging will remove!

listenbrainz/webserver/views/funkwhale.py
mAmanullah7: i'll try but it wont be any problem? we are not using the external services rather using funkwhale we have to take care of host_url every time its not hardcoded its user input? if we use every endpoint it will be complex to handle i think and even we are aiming to have multiple `host_url`! even one-to-one host_url it will be difficult to handle host_url dynamically and there is chances having trouble in exchanging token etc while managing host_url dynamically?

if you can give me a more clear approach i'll proceed with it!? 
like u wants me to not create endpoints in different fw.py rather in settings.py where other endpoints?
DO u wanna merge all endpoints or specifically like callback etc?

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: I've tested using global context now it works in both `creating oauth app & connecting `and `connecting a existing oauth app` without any state related or session related error! If any errors occurs again will see and now will be sticking using global context!

Will push changes by tonight!

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
mAmanullah7: Got it now! i understand now what if a user never visit BPsettings page, then my approach fails! 
I'll try to implement your approach once i'll let you know if i need any help!

mAmanullah7: [getPageProps](https://github.com/metabrainz/listenbrainz-server/blob/1b946969260ed42c39731d0a5a308b4b7ef86624/frontend/js/src/utils/utils.tsx#L601-L605) this can alone fix both the problem, we only need to update the `getPageProps` and `BPsettings`.?

mAmanullah7: i'm thinking migration should run for all users on every page load via `getPageProps` if wont be any problem? rather depending on user! Even we do we need to make changes in `BP` as suggested by you along with `BPsettings`.

```typescript
  // If a new service is added after the user sets their prefered priority list,
  // it will not appear in the list and we need to manually add it otherwise it will be ignored.
  const missingEnabledDataSources = difference(
    enabledDataSources,
    sortedDataSources
  );
  if (missingEnabledDataSources.length) {
    sortedDataSources.push(...missingEnabledDataSources);
  }

```
any suggestions?

MonkeyDo: I took a bit more time to see if I could come up with a very simple drop-in solution, and I think I have something.
Sorry for changing solutions at the last minute but I think this will just be easier.
RayyanSeliya is having the same issue for the InternetArchive player, and they can add the same simple change where needed.
It should not really affect anything in terms of performance considering we are comparing two tiny arrays.

This is a simple one-line change for this line:
https://github.com/metabrainz/listenbrainz-server/blob/934287c72f3e4b7eaade9d14685b66fec1f3bf80/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx#L184-L186

```ts
  // Combine saved priority list and default list to add any new music service at the end
  // then filter out disabled datasources (new ones will be enabled by default)
  const sortedDataSources = union(
    dataSourcesPriority,
    defaultDataSourcesPriority
  ).filter((key) => enabledDataSources.includes(key));
 ```
 
 `union` is imported from the lodash library and combines two arrays without duplicates and keeping the order of the first one, just what we need.
 
You should also be able to use `union` similarly when setting the `dataSourcesPriority` state value in BrainzPlayerSettings:
https://github.com/metabrainz/listenbrainz-server/blob/934287c72f3e4b7eaade9d14685b66fec1f3bf80/frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx#L85-L90

```tsx
  // Combine saved priorirty list and default list to add any new music service at the end
  const [dataSourcesPriority, setDataSourcesPriority] = React.useState<
    DataSourceKey[]
  >(
    union(
      userPreferences?.brainzplayer?.dataSourcesPriority ?? [],
      defaultDataSourcesPriority
    )
  );
```

mAmanullah7: Looks good to me! i'll make changes and push this along with fwicon!

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
MonkeyDo: For posterity, I re-created the icon definition here: https://codesandbox.io/p/sandbox/ywt3jy

frontend/css/sass/brainzplayer.scss
```diff
@@ -703,3 +703,28 @@ $alert-thumbnail-size: 60px;
...
+.fa-funkwhale {
+  width: 1.2em !important;
+  height: 1.2em !important;
+  
+  // Match the vertical alignment used by other music service icons
+  vertical-align: middle !important;
+  
+  // Default duotone colors for Funkwhale branding
+  --fa-primary-color: #009fe3; // Blue arcs
+  --fa-secondary-color: #3c3c3b; // Dark bird
```
MonkeyDo: Bird? XD

That's a whale fluke, or tail

frontend/js/src/common/brainzplayer/BrainzPlayerUI.tsx
```diff
@@ -39,6 +39,7 @@ import { FeedbackValue } from "./utils";
 import VolumeControlButton from "./VolumeControlButton";
 import { COLOR_LB_BLUE, COLOR_LB_ORANGE } from "../../utils/constants";
 import { DataSourceType } from "./BrainzPlayer";
+import faFunkwhale from "../icons/faFunkwhale";
```
MonkeyDo: Unused import

mAmanullah7:
Screenshot 2025-07-24 at 2 09 35 PM
Screenshot 2025-07-24 at 2 15 34 PM

@amCap1712 Everything working fine for me you can have a look!

frontend/js/src/common/brainzplayer/FunkwhalePlayer.tsx
```diff
@@ -0,0 +1,539 @@
...
+
+      return await response.json();
+    } catch (error) {
+      return null;
+    }
+  };
+
+  getFunkwhaleInstanceURL = (): string => {
+    const { funkwhaleAuth: funkwhaleUser = undefined } = this.context;
+    return funkwhaleUser?.instance_url || "https://demo.funkwhale.audio";
```
MonkeyDo: We might not want to add the default demo in production?

listenbrainz/domain/funkwhale.py
```diff
@@ -0,0 +1,171 @@
...
+            "host_url": host_url,
+            "client_id": server["client_id"],
+            "client_secret": server["client_secret"],
+            "access_token": token["access_token"],
+            "refresh_token": token["refresh_token"],
+            "token_expiry": token["token_expiry"],
+            "funkwhale_server_id": server["id"]
+        }
+        if refresh and self.user_oauth_token_has_expired(user):
+            user = self.refresh_access_token(user_id, server, user["refresh_token"])
```
MonkeyDo: This was causing an uncaught error for me causing 500s:
Had to resolve it before I was able to open any page.
Should it be caught here?
```suggestion
            try:
                user = self.refresh_access_token(user_id, server, user["refresh_token"])
            except:
                return None
```

frontend/js/src/common/brainzplayer/FunkwhalePlayer.tsx
```diff
@@ -0,0 +1,539 @@
...
+            src: artworkUrl,
+            sizes: "500x500",
+            type: "image/jpeg",
+          },
+        ]
+      : [];
+
+    onTrackInfoChange(
+      currentTrack.title,
+      currentTrack.listen_url || "",
```
MonkeyDo: I would return currentTrack.fid which is the federated link to the page. When I tried the BrainzPlayer menu option "Open in funkwhale" it tried to send me to an API endpoint.

Example `fid` : https://funkwhale-dev.metabrainz.org/federation/music/tracks/d82a81cf-e082-4638-8f02-cfda4c3cf7db

frontend/js/src/common/brainzplayer/FunkwhalePlayer.tsx
```diff
@@ -0,0 +1,539 @@
...
+      currentTrack.artist?.name || "",
```
MonkeyDo: This is not the structure I see in the API response (see below) and there was no artist names in the brainzplayer UI

Image

frontend/js/src/common/brainzplayer/FunkwhalePlayer.tsx
```diff
@@ -0,0 +1,539 @@
...
+    try {
+      const track = await searchForFunkwhaleTrack(
+        this.accessToken,
+        this.getFunkwhaleInstanceURL(),
+        trackName,
+        artistName
+      );
+
+      if (track && track.listen_url) {
+        if (track.is_playable === false) {
```
MonkeyDo: To avoid the empty block, maybe reverse this condition?

```suggestion
        if (track.is_playable !== false) {
        // the bit of code below
```
You'll also be able to remove the other is_playable check below

frontend/js/src/common/brainzplayer/FunkwhalePlayer.tsx
mAmanullah7: Yup! better to add error handling rather fall back to demo url

```typescript
getFunkwhaleInstanceURL = (): string => {
  const { funkwhaleAuth: funkwhaleUser = undefined } = this.context;
  if (!funkwhaleUser?.instance_url) {
    throw new Error(
      "No Funkwhale instance URL available - user not connected"
    );
  }
  return funkwhaleUser.instance_url;
};
```

listenbrainz/domain/funkwhale.py
mAmanullah7: Couldn't able to reproduce the error! both should work fine but think adding error handling would be better option!

@MonkeyDo if you haven't updated the `funkwhale_tables` i would recommend you to update it! 

`./develop.sh psql`

`funkwhale_servers`;
 ```
BEGIN;

DROP TABLE IF EXISTS funkwhale_servers CASCADE;

CREATE TABLE funkwhale_servers (
    id                  INTEGER GENERATED ALWAYS AS IDENTITY,
    host_url            TEXT NOT NULL UNIQUE,
    client_id           TEXT NOT NULL,
    client_secret       TEXT NOT NULL,
    scopes              TEXT NOT NULL,
    created             TIMESTAMP WITH TIME ZONE DEFAULT NOW()
);

ALTER TABLE funkwhale_servers ADD CONSTRAINT funkwhale_servers_id_pkey PRIMARY KEY (id);

CREATE INDEX host_url_ndx_funkwhale_servers ON funkwhale_servers (host_url);

COMMIT;
```
then `funlwhale_tokens`;

```
BEGIN;

DROP TABLE IF EXISTS funkwhale_tokens;

CREATE TABLE funkwhale_tokens (
    id                  INTEGER GENERATED ALWAYS AS IDENTITY,
    user_id             INTEGER NOT NULL,
    funkwhale_server_id INTEGER NOT NULL,
    access_token        TEXT NOT NULL,
    refresh_token       TEXT NOT NULL,
    token_expiry        TIMESTAMP WITH TIME ZONE NOT NULL
);

ALTER TABLE funkwhale_tokens ADD CONSTRAINT funkwhale_tokens_id_pkey PRIMARY KEY (id);

ALTER TABLE funkwhale_tokens
    ADD CONSTRAINT funkwhale_tokens_user_id_foreign_key
    FOREIGN KEY (user_id)
    REFERENCES "user" (id)
    ON DELETE CASCADE;

ALTER TABLE funkwhale_tokens
    ADD CONSTRAINT funkwhale_tokens_server_id_foreign_key
    FOREIGN KEY (funkwhale_server_id)
    REFERENCES funkwhale_servers (id)
    ON DELETE CASCADE;

CREATE INDEX user_id_ndx_funkwhale_tokens ON funkwhale_tokens (user_id);
CREATE INDEX server_id_ndx_funkwhale_tokens ON funkwhale_tokens (funkwhale_server_id);
CREATE UNIQUE INDEX unique_user_server_funkwhale_tokens ON funkwhale_tokens (user_id, funkwhale_server_id);

COMMIT; 
```
and then again you have to reconnect to funkwhale servers;

File: frontend/js/src/common/brainzplayer/FunkwhalePlayer.tsx [hunk 23]
```diff
@@ -0,0 +1,559 @@
...
+        musicService.toLowerCase().includes("funkwhale")) ||
+      Boolean(FunkwhalePlayer.getURLFromListen(listen))
+    );
+  }
+
+  static getURLFromListen = (
+    listen: Listen | JSPFTrack
+  ): string | undefined => {
+    const originURL = _get(listen, "track_metadata.additional_info.origin_url");
+    if (originURL && /funkwhale/.test(originURL)) {
```
amCap1712: The Funkwhale pod url might not always contain `funkwhale` in the URL.

mAmanullah7: pushed the latest changes you can have a look!
`https://open.audio/@amanxllah7` i even verified this it worked for me!

frontend/js/src/common/brainzplayer/FunkwhalePlayer.tsx
MonkeyDo: I have to admit I am curious as to where you got that data structure from.

mAmanullah7: `currentTrack.artist?.name`  its same from very initial part when i integrated fw in BP! i even didn't notice until you mentioned! Initially i've taken reference from existing services and then refactored! i think in flow or in misunderstanding i've written this! give some moments i'll once check may be while exploring api and responses and all i've messed up!

mAmanullah7: yeah i  think i may have used this in flow or in confusion cant exactly say as this similar to navidrome/subsonic rather than fw! you can see [http://your-server/rest/search3](https://www.subsonic.org/pages/api.jsp)
Screenshot 2025-08-18 at 9 14 42 PM

mAmanullah7: @MonkeyDo Thanks for pointing this one! This is not the only one which is using a different data structures as compare to the above you mentioned, after debugging for long time finally checked the previous commits and then noticed this inconsistency started after merging two different endpoints to a single endpoint after discussing with lucifer! We were using two different endpoints `/api/v1/search` and `/api/v1/tracks` so the above mentioned ds belongs to the `/api/v1/search` : https://docs.funkwhale.audio/swagger/#/search/get_search_results and even funkwhale supports subsonic api as well: https://docs.funkwhale.audio/developer/api/subsonic.html but we are not using! Now we use single endpoint `/api/v1/tracks` and then try to filter out tracks on the basis of artist name if have more than one tracks!

frontend/js/tests/common/brainzplayer/FunkwhalePlayer.test.tsx
```diff
@@ -0,0 +1,293 @@
...
+          track_metadata: {
+            additional_info: {
+              funkwhale_id: "12345",
+            },
+            artist_name: "Test Artist",
+            track_name: "Test Track",
+          },
+        };
+        expect(FunkwhalePlayer.getURLFromListen(listen)).toBe("12345");
+      });
```
MonkeyDo: I think this duplicates the test at line 151 https://github.com/metabrainz/listenbrainz-server/pull/3289/files#diff-4f3859a0dd7a038efe728caa46db0fca0448820f56aa40e2065018a88d811f0aR151-R163

frontend/js/tests/common/brainzplayer/FunkwhalePlayer.test.tsx
```diff
@@ -0,0 +1,293 @@
...
+    it("should render successfully", () => {
+      const { container } = render(
+        <GlobalAppContext.Provider value={defaultContext}>
+          <FunkwhalePlayer {...defaultProps} />
+        </GlobalAppContext.Provider>
+      );
+
+      expect(container).toBeTruthy();
+      // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+      const audioElement = container.querySelector("audio");
```
MonkeyDo: The correct way to test this (and not have to disable linting rules that try to force the correct way ;) ) is to add a testid to the audio element:
`<audio data-testid="funkwhale-audio" ref={this.audioRef} crossOrigin="anonymous" preload="metadata">`
then you can use this assertion in the test:
 `expect(screen.getByTestId("funkwhale-audio"")).toBeInTheDocument();`

Don't forget to remove the eslint-disable comments

frontend/js/tests/common/brainzplayer/FunkwhalePlayer.test.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+    it("should hide the player when show prop is false", () => {
+      const { container } = render(
+        <GlobalAppContext.Provider value={defaultContext}>
+          <FunkwhalePlayer {...defaultProps} show={false} />
+        </GlobalAppContext.Provider>
+      );
+
+      // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+      const playerContainer = container.querySelector(".funkwhale-player");
```
MonkeyDo: Same as above: add `data-testid="funkwhale-player"` in the component and test with `getByTestId("funkwhale-player")`

MonkeyDo (commented):
OK, everything looks quite good now, good work on adding tests.
If you wanted to implement further functional tests, a good source of inspiration is the SpotifyPlayer tests: https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/tests/common/brainzplayer/SpotifyPlayer.test.tsx#L213-L238
That being said, I think the current tests are probably sufficient.

Overall I think this is ready for being merged, and more thoroughly tested by the wider community. Nice job!

It would be great if once merged and deployed to test/beta (to be coordinated with lucifer or I) you could write a message in the LB matrix channel and/or in the community forums to call for beta testers who use funkwhale. Always the best way to find the bugs that slipped through :)

mAmanullah7:
> OK, everything looks quite good now, good work on adding tests. If you wanted to implement further functional tests, a good source of inspiration is the SpotifyPlayer tests: https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/tests/common/brainzplayer/SpotifyPlayer.test.tsx#L213-L238 That being said, I think the current tests are probably sufficient.
> 
> Overall I think this is ready for being merged, and more thoroughly tested by the wider community. Nice job!
> 
> It would be great if once merged and deployed to test/beta (to be coordinated with lucifer or I) you could write a message in the LB matrix channel and/or in the community forums to call for beta testers who use funkwhale. Always the best way to find the bugs that slipped through :)

about additional test surely i'll create a separate pr but before writing additional test i'll work on `navidrome` integration #3331 and also i've to write basic backend and frontend test!

Sure posting on community forums as well in matrix channel both would be great once this merged i'll do that!

frontend/js/tests/common/brainzplayer/FunkwhalePlayer.test.tsx
mAmanullah7: Thanks!

PR #3290, opened 2025-06-02
update readthedocs config

amCap1712:
doc builds are failing because of missing imports on older python versions.


MonkeyDo:
We broke something in #3288 , hence the failing test.
Fixing that now, sorry !

PR #3292, opened 2025-06-03
Use cosine similarity instead of pearson coefficient for similar users

amCap1712:
Calculating the pearson coefficient using the inbuilt spark utilities requires building a dense feature matrix/dataframe which often leads to a Out of Memory error. The `columnSimilarities` on `RowMatrix` should avoid this and work well enough for our use case.


MonkeyDo:
Happy to help test/compare the current similarities I have with users with the new calculation method

mayhem (approved):

PR #2965, opened 2024-08-22
SoundCloud Playlist Export & Import functionalities

rimma-kubanova:
Implemented export and import features for SoundCloud

IMPORT:
* Added modal for SoundCloudPlaylistImport 
* Added endpoints for SoundCloud Playlist import and playlist tracks import

EXPORT:
* Added troi call for SoundCloud export
* Added endpoint for SoundCloud export on back
* Added button for SoundCloud export on front

PR #3296, opened 2025-06-05
Fix Bootstrap5 migration issues

MonkeyDo:
After deploying #3272 we (devs + users) found a few forgotten or missed style issues.

- anchor tags were the wrong color when used in combination with .btn and .dropdown-item class (removed custom BS3 css)
- the accordion styles in the playing now page were broken. I replaced them with React-Bootstrap Accordion component while at it.
- Also improved one other accordion component we have, and got rid of BS3 accordion CSS
- Fixed Soundcloud modal and import buttons (PR #2965 was merged after the BS5 migration PR, needed to be adapted too)
- Fixed some pre-existing overflow issues (dashboard pagination, playlists page top nav), now easier with better utility classes with BS5
- Fixed overflow/read more button on entity pages wikipedia extract
- Fixed playlist edit modal NamePill component

This PR fixes all leftover issues I could find, and removes a bit of custom bootstrap 3 CSS in preparation of removing all of it.
This is a good first step towards that goal, and is relevant to fixing the styling issues mentioned


MonkeyDo:
Deployed to test.LB

frontend/js/src/common/Accordion.tsx
```diff
@@ -5,32 +5,35 @@ import { faChevronCircleRight } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { uniqueId } from "lodash";
 import * as React from "react";
+import { Card, CardBody, Collapse, CardHeader } from "react-bootstrap";
```
anshg1214, line 8: Good that we're starting to refactor components by using react-bootstrap

frontend/css/scaffolding.less
```diff
@@ -11,16 +11,16 @@ body {
...
+// a {
+//   color: #353070;
+//   text-decoration: none;
+
+//   &:hover,
+//   &:focus {
+//     color: #49429b;
+//     text-decoration: none;
+//   }
+// }
```
anshg1214: We can probably remove this now?

anshg1214 (approved):
LG!

frontend/js/src/common/Accordion.tsx
MonkeyDo: Yes, especially I thought since it allows us to get rid of some BS3 less files in the process, less moving over of styles to do

PR #3297, opened 2025-06-08
[Feature] Create Main importer for listening history files

Suvid-Singhal:
## New Feature
Created the main importer endpoint for uploading the listening history files
Also created a background task for the import process

### DIFF
--- a/listenbrainz/webserver/views/api.py
+++ b/listenbrainz/webserver/views/api.py
@@ -23,9 +23,9 @@ from listenbrainz.webserver.errors import APIBadRequest, APIInternalServerError,
 from listenbrainz.webserver.models import SubmitListenUserMetadata
 from listenbrainz.webserver.utils import REJECT_LISTENS_WITHOUT_EMAIL_ERROR, REJECT_LISTENS_FROM_PAUSED_USER_ERROR
 from listenbrainz.webserver.views.api_tools import insert_payload, log_raise_400, validate_listen, \
-    is_valid_uuid, MAX_LISTEN_PAYLOAD_SIZE, MAX_LISTENS_PER_REQUEST, MAX_LISTEN_SIZE, LISTEN_TYPE_SINGLE, \
-    LISTEN_TYPE_IMPORT, _validate_get_endpoint_params, LISTEN_TYPE_PLAYING_NOW, validate_auth_header, \
-    get_non_negative_param, _parse_int_arg
+    is_valid_uuid, MAX_LISTEN_PAYLOAD_SIZE, MAX_LISTENS_PER_REQUEST, MAX_LISTEN_SIZE, MAX_FILE_UPLOAD_SIZE, \
+    LISTEN_TYPE_SINGLE, LISTEN_TYPE_IMPORT, _validate_get_endpoint_params, LISTEN_TYPE_PLAYING_NOW, validate_auth_header, \
+    get_non_negative_param, _parse_int_arg, upload_listening_history_files
 
 api_bp = Blueprint('api_v1', __name__)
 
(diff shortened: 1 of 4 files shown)


listenbrainz/webserver/views/import.py
```diff
@@ -0,0 +1,121 @@
...
+        fetch_oauth_query = text("""
+            SELECT id FROM external_service_oauth
+            WHERE user_id = :user_id AND service = :service
+        """)
+        result = db_conn.execute(fetch_oauth_query, {
+            "user_id": current_user.id,
+            "service": service
+        }).first()
+
+        if result:
```
amCap1712: These should be created by the importer when it has started to or processed the import I feel. Until then just track everything needed in the import tasks's metadata.

listenbrainz/webserver/views/import.py
```diff
@@ -0,0 +1,121 @@
...
+        current_app.logger.error('Error while importing user data: %s', current_user.musicbrainz_id, exc_info=True)
+        raise APIInternalServerError(f'Error while importing user data {current_user.musicbrainz_id}, please try again later.')
+
+@import_bp.get("/list/")
+@api_login_required
+@web_listenstore_needed
+def list_export_tasks():
+    """ Retrieve the all import tasks for the current user """
+    result = db_conn.execute(
+        text("SELECT * FROM listens_importer WHERE user_id = :user_id ORDER BY last_updated DESC"),
```
amCap1712: This should read from background tasks table with a filter instead of listens importer.

amCap1712: Same below.

listenbrainz/webserver/views/api.py
```diff
@@ -877,3 +877,29 @@ def get_artist_radio_recordings(seed_artist_mbid):
...
+    file = request.files.get("file")
+    file_type = request.form.get("type")
+
+    if not file or not file_type:
+        APIBadRequest("Missing file or type")
+
+    if file.content_length and file.content_length > MAX_FILE_UPLOAD_SIZE:
+        APIBadRequest("File size exceeds 100MB")
+    
+    file_path = upload_listening_history_files(file, file.filename)
```
amCap1712, line 900: You need to store this file_path in the import task.

listenbrainz/webserver/views/import.py
```diff
@@ -0,0 +1,121 @@
...
+from listenbrainz.webserver.login import api_login_required
+
+import_bp = Blueprint("import", __name__)
+
+
+@import_bp.post("/")
+@api_login_required
+@web_listenstore_needed
+def create_import_task(file_path, service):
+    """ Add a request to import the listening history data in background. """
```
amCap1712: The way this should work is that the frontend calls an API to upload a file, after the file is uploaded an import task should be created with the file path stored in its metadata.

listenbrainz/webserver/views/import_listens.py
```diff
@@ -0,0 +1,118 @@
...
+    } for row in rows])
+
+
+@import_bp.get("/<import_id>/")
+@api_login_required
+@web_listenstore_needed
+def get_import_task(import_id):
+    """ Retrieve the requested import's data if it belongs to the specified user """
+    result = db_conn.execute(
+        text("SELECT * FROM listens_importer WHERE user_id = :user_id AND id = :import_id"),
```
amCap1712, line 107: This should use background_tasks table too. Use background_tasks' id where ever needed unless needed otherwise.

listenbrainz/webserver/views/import_listens.py
```diff
@@ -0,0 +1,118 @@
...
+        raise APIBadRequest(message="Data import already requested.")
+
+    except DatabaseError:
+        current_app.logger.error('Error while importing user data: %s', current_user.musicbrainz_id, exc_info=True)
+        raise APIInternalServerError(f'Error while importing user data {current_user.musicbrainz_id}, please try again later.')
+
+@import_bp.get("/list/")
+@api_login_required
+@web_listenstore_needed
+def list_export_tasks():
```
amCap1712, line 86: `list_import_tasks`

listenbrainz/webserver/views/import_listens.py
```diff
@@ -0,0 +1,118 @@
...
+        """)
+        result = db_conn.execute(fetch_oauth_query, {
+            "user_id": current_user.id,
+            "service": service
+        }).first()
+
+        if result:
+            external_oauth_id = result.id
+        else:
+            insert_oauth_query = text("""
```
amCap1712, line 33: Like I mentioned in a previous comment, the `external_service_oauth` field and the `listens_importer` field should not be modified here. This should be modified by the background task handler that will process the zip file. You should only modify/create the background tasks table here.

Suvid-Singhal:
Closing this PR due to change in implementation plans

PR #3298, opened 2025-06-09
feat: Improve React Router Lazy Loading and improve loading times

anshg1214:
As a part of improving the load times of LB, in this PR:
1. Refactors the routes to follow the recommended system by react router 7: https://remix.run/blog/faster-lazy-loading
2. Refactors the UserSocialNetwork Component to have loaders in place, and fetch data parallelly to improve load times


frontend/js/src/user/components/follow/UserSocialNetwork.tsx
```diff
@@ -8,254 +8,174 @@ import SimilarUsersModal from "./SimilarUsersModal";
...
+    const {
+      getFollowersOfUser,
+      getFollowingForUser,
+      getSimilarUsersForUser,
+      getSimilarityBetweenUsers,
+      getUserEntity,
+    } = APIService;
+
+    // Get followers
+    getFollowersOfUser(profileUser.name)
```
MonkeyDo, line 48: I think eventually it would be nice to use react-query for these API calls to aggressively cache that data

MonkeyDo (commented):
Was the transformation of UserSocialNetwork from class to functional component required for the lazy loading changes?
It looks like it's bundled in this PR but not clear to me how it's connected, would be worth noting why in the description.

Is it a conscious choice to only modify the 'lazy' syntax in this one routes file only?
If so it would be good to add a note in the PR description for that, so for future reference we know why we have two different styles.

If it wasn't on purpose, probably best to modify the others too to follow the same style, for future copy-paste.

MonkeyDo:
Other than the comment above, current PR works fine (I tested on test.LB), the conversion from class to func component and the change of tests from enzyme to testing-library is a good continuation, thanks !

frontend/js/src/user/components/follow/UserSocialNetwork.tsx
anshg1214: Yes, eventually we'll move this to react query, but that's for another PR.

anshg1214:
@MonkeyDo, this PR was intended to address a broader aspect of refactoring both the frontend and backend code to improve load times. I worked on a few improvements, but as discussed, they weren’t significant enough and required a lot of engineering effort. Therefore, I’ve decided to focus on improving the router and the component that was loading too slowly.

frontend/js/src/settings/routes/index.tsx
```diff
@@ -7,104 +7,104 @@ const getSettingsRoutes = (): RouteObject[] => {
...
-            return { Component: ResetToken.default };
+          lazy: {
+            Component: async () => {
+              return (await import("../resettoken/ResetToken")).default;
+            },
           },
         },
         {
           path: "music-services/details/",
           loader: RouteLoader,
```
MonkeyDo: All the loaders in this file also need to be refactored I suppose?

frontend/js/src/about/routes/index.tsx
```diff
@@ -12,51 +12,54 @@ const getAboutRoutes = (): RouteObject[] => {
...
-            return { Component: AddData.default };
+          lazy: {
+            Component: async () => {
+              return (await import("../add-data/AddData")).default;
+            },
           },
         },
         {
           path: "current-status/",
           loader: RouteQueryLoader("current-status"),
```
MonkeyDo: This loader needs to be refactored too

anshg1214: Fixed

frontend/js/src/settings/routes/index.tsx
anshg1214: Thanks for pointing out. I've refactored this file too

PR #3301, opened 2025-06-12
Add Libre.fm backend importer

amCap1712:
Similar to last.fm backend importer. Most of the code is reused between the two with a couple of small differences in API error behavior.


mayhem (approved):
Great, lets get this deployed!

amCap1712:
The frontend tests are passing locally.

MonkeyDo (commented):
- I refactored a couple more bits of code where I thought we didn't need to pass down functions, considering the LFM permissions management is now done entirely in the new components.
- I renamed the new component to avoid possible confusion.
- I made the /settings/import/ route a redirect instead of completely removing it, as it was historically our main entrypoint for importing history and we might have third-parties linking to that page specifically.

MonkeyDo:
After some more testing in beta to make sure I didn't break anything, I am satisfied with the new feature.
Thanks lucifer for working on this!

> The frontend tests are passing locally.

I am looking into this test file as we speak, refactoring it to use testing-library instead of enzyme which should resolve these phantom failures (it has in the past for other modals testing files).

MonkeyDo:
I reported the suspicious activity of this user approving  PRs after merging, looks like botting.

PR #3304, opened 2025-06-18
Rewrite tests for PersonalRecommendationsModal

MonkeyDo:
We were seeing some random-looking test failures in #3301 caused by this PersonalRecommendationsModal test file.
This is an issue we have seen in the past related to NiceModal and the Enzyme test harness.
We have sloooowly been rewriting test files to use react-testing-library instead of the abandoned enzyme, here we are migrating one more test file while slightly improving the tests.


anshg1214 (approved):
LG!

PR #3306, opened 2025-06-19
Fix "Invalid video id" in YoutubePlayer

amCap1712:
While using a debugger to trace the issue, I found the following line inside the Youtube's iFramePlayer code triggered the issue: 
```javascript
if((a=String(Y(a,"videoId")))&&(a.length!==11||!a.match(/^[a-zA-Z0-9\-_]+$/)))throw Error("Invalid video id");
```
I am guessing that code is basically stringifying the input videoId, `String(null)` evaluates to `"null"` and `String(undefined)` to `"undefined"` which leads to the error.

Empty string avoids the issue, hence using that as default videoId.

### DIFF
--- a/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx
@@ -402,6 +402,7 @@ export default class YoutubePlayer extends React.Component<YoutubePlayerProps>
             onError={this.onError}
             onStateChange={this.handlePlayerStateChanged}
             onReady={this.onReady}
+            videoId=""
           />
         </div>
       </Draggable>


mwiencek:
Another useful thing to have would be a closer [error boundary](https://legacy.reactjs.org/docs/error-boundaries.html) so that one component doesn't bring down the entire page. 😅 (I couldn't access the site at all yesterday or today.)

MonkeyDo:
> Another useful thing to have would be a closer [error boundary](https://legacy.reactjs.org/docs/error-boundaries.html) so that one component doesn't bring down the entire page. 😅 (I couldn't access the site at all yesterday or today.)

I was not aware this broke the whole page!
I just saw an error in the console in my case.
Thanks, will review our error boundaries for the music player

PR #3307, opened 2025-06-19
Reinstate /settings/import page

MonkeyDo:
This page was removed in [#3301](https://github.com/metabrainz/listenbrainz-server/issues/3301) as we moved the last feature out of that page.
However we have an ongoing GSOC project for importing backup files, which will need a place to live. This is the page,.

Since the removal has not been deployed to prod yet, rather than re-adding it at a later date when we have the feature, we are opting for re-adding it now with a "coming soon" placeholder so the page never officially went away for users.

I also modified the page content to make it a bit clearer and clearly point to other relevant pages.

image

### DIFF
--- a/frontend/js/src/about/add-data/AddData.tsx
+++ b/frontend/js/src/about/add-data/AddData.tsx
@@ -383,6 +383,13 @@ export default function AddData() {
             </Link>
           </em>
         </li>
+        <li>
+          <em>
+            <Link to="/settings/import/">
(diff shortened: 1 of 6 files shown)


MonkeyDo:
Failing front-end test is related to #3304

MonkeyDo:
Merging without review: simple UI changes that don't add anything new, failing tests are unrelated.

PR #3308, opened 2025-06-21
User genre activity

granth23:
# [WIP] Genre Statistics Visualization

This PR introduces the initial implementation of the **Genre Stats** feature, part of my GSoC 2025 project: [Development of Advanced User Statistics Visualizations](https://community.metabrainz.org/t/gsoc-2025-development-of-advanced-user-statistics-visualizations/750040).

The Genre Stats visualization aims to give users insights into their listening habits based on **music genres**. This feature is designed to show which genres a user listens to most, along with time-based filtering and engaging visualizations.

## Current Status

* **Backend**:

  * Core logic for aggregating genre-based listen stats has been implemented.
  * Fetches user listen data and maps it to genre information (via MusicBrainz tags).

* **Frontend**:

  * A working mockup is implemented using **hardcoded data** to prototype the UI and charts.
  * Final integration with the backend API is **pending**, to be completed in the next stage.

## Tech Details

* Data is aggregated by genre and grouped over defined time ranges (e.g., last 7 days, month, all time).
* Visual components (charts) are implemented using **Nivo**.
* The API is structured to return genre stats with listen count and/or duration as metrics.

## Work in Progress

* Finalize and test backend API responses.
* Replace mock frontend data with dynamic API calls.
* Add error handling for users with limited or no genre-tagged listens.
* Polish visual layout and responsiveness.


frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
```diff
@@ -0,0 +1,271 @@
...
+    listen_count: number;
+  }>;
+};
+
+export type UserGenreDayActivityResponse = {
+  result: GenreTimeframeData[];
+};
+
+// Define a color mapping for genres
+const genreColors: Record<string, string> = {
```
MonkeyDo: I know from experience colors are going to be annoying to work with in nivo, so I thought I'd save you some time...

Nivo uses D3 color schemes in the background.

In this case for this specific graph, I would recommend one of the cyclical color schemes (probably interpolateRainbow I'd say):
https://d3js.org/d3-scale-chromatic/cyclical

Which you should be able to use with something like this:

```
import { cyclicalColorInterpolators } from "@nivo/colors";
import { scaleSequential } from "d3-scale";`
```
Generating a scale from the color scheme and our data.
You'll want to map the data (the 'domain' here) between 0 and 1 (here I'm mapping using 0 and the max value, we'll want to use the hour instead):
```
const colorScale = scaleSequential(cyclicalColorInterpolators.rainbow).domain([
  0,
  Math.max(...data.map((d) => d.value)),
]);
```
And then using it in the graph component options:
`colors={(d) => colorScale(d.value)}`

frontend/js/src/utils/APIService.ts
```diff
@@ -602,6 +602,24 @@ export default class APIService {
     return response.json();
   };
 
+  getUserGenreDayActivity = async (
+    userName?: string,
```
anshg1214: why is `username` optional here? Currently we don't have support for sitewide stat, so we should make it compulsory

listenbrainz/webserver/views/stats_api.py
```diff
@@ -523,7 +524,19 @@ def get_artist_activity(user_name: str):
     release_groups_list, _ = _process_user_entity(stats, offset, count, entire_range=True)
     result = _get_artist_activity(release_groups_list)
     return jsonify({"result": result})
+
+@stats_api_bp.get("/user/<user_name>/genre-activity")
+@crossdomain
+@ratelimit()
+def get_genre_activity(user_name: str):
+    user, stats_range = _validate_stats_user_params(user_name)
```
anshg1214: Add docstrings and add sample json data.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -523,7 +524,19 @@ def get_artist_activity(user_name: str):
...
+    stats = db_stats.get(1, "genre_activity", stats_range, GenreActivityRecord)
```
anshg1214: You need to pass user information, and not the user_id 1.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -523,7 +524,19 @@ def get_artist_activity(user_name: str):
     release_groups_list, _ = _process_user_entity(stats, offset, count, entire_range=True)
     result = _get_artist_activity(release_groups_list)
     return jsonify({"result": result})
+
+@stats_api_bp.get("/user/<user_name>/genre-activity")
+@crossdomain
+@ratelimit()
+def get_genre_activity(user_name: str):
```
anshg1214, line 532: Do you also need offset and count as a parameter in the request like how other stats work?

listenbrainz/webserver/views/stats_api.py
```diff
@@ -523,7 +524,19 @@ def get_artist_activity(user_name: str):
...
+    user, stats_range = _validate_stats_user_params(user_name)
+    stats = db_stats.get(1, "genre_activity", stats_range, GenreActivityRecord)
+    if stats is None:
+        raise APINoContent('')
     
+    genre_activity = [x.dict() for x in stats.data.__root__]
+    return jsonify({"result": genre_activity})
+
```
anshg1214: Extra white space
```suggestion

```

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
```diff
@@ -0,0 +1,286 @@
...
+        0
+      );
+      const baseHour =
+        TIME_PERIODS.find((p) => p.timeRange === timeframe.timeRange)?.hour ??
+        0;
+
+      return timeframe.genres.map((genre, index) => {
+        const hourVariation = baseHour + ((index * 0.5) % 6);
+        const value =
+          genre.listen_count === 0 ? 100 : (genre.listen_count / total) * 100;
```
anshg1214: Can you add some comments here explaining what we're trying to achieve here?

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
```diff
@@ -0,0 +1,286 @@
...
+    const groupedData = groupDataByTimePeriod(rawData.result, timezoneOffset);
+
+    return groupedData.flatMap((timeframe) => {
+      const total = timeframe.genres.reduce(
+        (acc, genre) => acc + genre.listen_count,
+        0
+      );
+      const baseHour =
+        TIME_PERIODS.find((p) => p.timeRange === timeframe.timeRange)?.hour ??
+        0;
```
anshg1214: Again - Can you add some comments here?

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
```diff
@@ -0,0 +1,286 @@
...
+    },
+    {
+      position: { bottom: "5px", left: "50%", transform: "translateX(-50%)" },
+      label: "12PM",
+    },
+    {
+      position: { top: "50%", left: "25%", transform: "translateY(-50%)" },
+      label: "6PM",
+    },
+  ];
```
anshg1214: Since this is constant, you can move it outside the component.

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
```diff
@@ -0,0 +1,286 @@
...
+      style={{
+        background: "white",
+        padding: "12px",
+        border: "1px solid #ccc",
+        borderRadius: "4px",
+        boxShadow: "0 2px 4px rgba(0,0,0,0.2)",
+        textAlign: "center",
+        opacity: 0.9,
+      }}
+    >
```
anshg1214: We usually avoid inline css, so you can define a class and move the css there.

anshg1214 (changes requested):
The PR Looks good on a high level. We can have more idea once this PR is deployed on test.LB and we can test it on live data.

There are a couple of changes required right now. Apart from them, you should use mocked data to write frontend tests for the stat.

anshg1214:
For the spark changes, I believe @amCap1712 will be the best person to have a look.

listenbrainz/webserver/views/stats_api.py
granth23: No, in my opinion we don't need to use any subset of the entire stats generated, could you please let me know if I am going wrong with something?

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
granth23: Done

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
granth23: Added style to stats.less

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
granth23: Implemented the same, thanks for the recommendation

listenbrainz/webserver/views/stats_api.py
anshg1214: Hmm, correct, we don't need this subset. So, we can probably ignore it.

frontend/css/stats.less
```diff
@@ -112,3 +112,13 @@
...
+
+.custom-tooltip-genre-stats {
+  background: white;
+  padding: 12px;
+  border: 1px solid #ccc;
+  border-radius: 4px;
+  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.2);
+  text-align: center;
+  opacity: 0.9;
+}
```
MonkeyDo: We just last week merge some changes to our CSS preprocessing, moving from Less to Sass.

Thankfully the solution is simple to resolve the merge conflict: move thes changes to the new [stats.scss file](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/css/sass/stats.scss) and remove the changes from this file (will be deleted after the merge)

anshg1214:
@granth23 I've deployed this PR to [test.lb](https://test.listenbrainz.org/user/anshgoyal31/stats/?range=this_week).

You need to improve on how the graph looks on smaller screen. The Time markers - 12 AM and 12 PM need to be closer to the graph, and 6 AM, 6 PM should not be on the graph. 
image

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
```diff
@@ -164,6 +147,19 @@ export default function UserGenreDayActivity({
...
+  React.useEffect(() => {
+    const checkMobile = () => {
+      setIsMobile(window.innerWidth < 768);
+    };
+
+    checkMobile();
+    window.addEventListener("resize", checkMobile);
+    return () => window.removeEventListener("resize", checkMobile);
+  }, []);
+
```
MonkeyDo: We have a hook already implemented for this: `useMediaQuery`
See for example:
https://github.com/metabrainz/listenbrainz-server/blob/dd7afbfa4cb915eb9eacda9154f0bf396ad5f66e/frontend/js/src/user/charts/UserEntityChart.tsx#L135

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
```diff
@@ -240,6 +236,101 @@ export default function UserGenreDayActivity({
     });
   }, [colorScale, rawData, timezoneOffset]);
 
+  // Responsive time markers - all four markers with mobile-optimized positioning
+  const getTimeMarkersConfig = () => {
+    if (isMobile) {
+      return [
+        {
+          position: {
+            top: "15px",
```
MonkeyDo: While playing with the markers, I came up with something that seems to work for all sizes, with the markers inside the circle.
I'm not sure it's better than the current implementation as it changes the overall look, but seems to be more flexible as the placement does not change depending on the parent container size:
image

`top: "calc(50% - 80px)"`  <- this calculated position works the same for every marker, but applied to the correct attribute ('top' for the top one, 'bottom' for the bottom one, etc.)
For the smaller screen sizes the 80px needs to be 40px (the same value as the right and left margins defined in getChartConfig)

MonkeyDo (commented):
Overall looking quite good!

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
granth23: fixed

frontend/js/src/user/stats/components/UserGenreDayActivity.tsx
granth23: thanks for the suggestion, looks great

listenbrainz_spark/stats/user/tests/__init__.py
```diff
@@ -36,6 +37,14 @@ def upload_metadata_cache(cls) -> None:
             RECORDING_ARTIST_DATAFRAME,
             cls.path_to_data_file("recording_artist.parquet")
         )
+        upload_to_HDFS(
```
amCap1712, line 40: Please add tests on lines of existing spark stats tests, making use of these test data files.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -523,7 +524,20 @@ def get_artist_activity(user_name: str):
...
     result = _get_artist_activity(release_groups_list)
     return jsonify({"result": result})
+
+
+@stats_api_bp.get("/user/<user_name>/genre-activity")
+@crossdomain
+@ratelimit()
+def get_genre_activity(user_name: str):
+    user, stats_range = _validate_stats_user_params(user_name)
+    stats = db_stats.get(user['id'], "genre_activity", stats_range, GenreActivityRecord)
```
amCap1712, line 585: Also, add tests for the API. Look at existing stats API tests for inspiration.

listenbrainz_spark/stats/user/tests/__init__.py
granth23: added

listenbrainz/webserver/views/stats_api.py
granth23: added

anshg1214 (approved):
LG!

PR #3310, opened 2025-06-22
Fix typo in footer text

mmaismma:
# Problem

There is a typo in the footer text. An 'is' was mistyped twice near the end. Highlighted below.

> You can use ListenBrainz to track your listening habits, discover new music with personalized recommendations, and share your musical taste with others using our visualizations. User listen data and text is **_is_** made public under the [Creative Commons Zero (CC0)](https://creativecommons.org/public-domain/cc0/) license.

# Solution

Just removed the extra "is".

# Action

None needed.


MonkeyDo (approved):
Thanks for spotting that! :)

PR #3311, opened 2025-06-23
Add "Benben" player to add data

Aerozol:
Added "Benben" player
+
Standardised anything like "audio player" to "music player" (which was picked simply because it was more common)


MonkeyDo (approved):
Thank you!

PR #3313, opened 2025-06-25
Refactor LESS styles to use SASS instead

MonkeyDo:
After moving from Bootstrap v3 to v5 (#3243 and #3244) we are left with two CSS preprocessors: less we had previously and Sass introduced because we wanted to customize Bootstrap5.

This PR refactors Less code into Sass, and removes the LESS build dependencies.

Fairly involved as we had a lot of small custom css details (and I am sure we will find more missing)


anshg1214:
This is a significant refactor, so I’ll start reviewing the PR file by file.

MonkeyDo:
> This is a significant refactor, so I’ll start reviewing the PR file by file.

Good luck!
:saluting_face: 

I deployed the PR to test.LB to make it easier to go page by page

frontend/css/sass/playlists.scss
```diff
@@ -1,6 +1,6 @@
-@playlistBackgroundColor: #fafafa;
-@descriptionLines: 3;
-@longDescriptionLines: 8;
+$playlist-background-color: $body-tertiary-bg;
```
anshg1214, line 1: Have we intentionally changed this color from `#fafafa` to `#f9f6f5`?

anshg1214:
I tested every component and compared with the production version.  It looks good overall, just two main things which needs to be fixed:

- `.card` Border Color should be #eee
- `.capitalize-bold` class defination is missing

anshg1214 (changes requested):
I tested every component and compared with the production version.  It looks good overall, just two main things which needs to be fixed:

- `.card` Border Color should be #eee
- `.capitalize-bold` class defination is missing

Also, I noticed that the background colours of sidebar and playlist card are different. Have they been changed intensionally?

frontend/css/sass/playlists.scss
MonkeyDo: Where I was able to do it without (much) noticeable changes, I've tried to reduce the number of different arbitrary colors we use, and use variables instead, in an effort to harmonize the look of various components.
However in this case it is noticeable. I think ideally we would either use this new color, or change the $body-tertiary-bg to use #fafafa instead.

I think changing $body-tertiary-bg to #fafafa makes sense as it doesn't seem used in a lot of places, and that way we don't introduce this change

anshg1214 (approved):
LG!

PR #3314, opened 2025-06-28
Artist Evolution Visualization

granth23:
# Artist Evolution Visualization

This PR finalizes the implementation of the **Artist Evolution** feature, developed as part of my GSoC 2025 project: [Development of Advanced User Statistics Visualizations](https://community.metabrainz.org/t/gsoc-2025-development-of-advanced-user-statistics-visualizations/750040).

The Artist Evolution visualization allows users to track how their listening preferences for different **artists** change over a defined time range. It provides a stacked stream graph that illustrates artist-wise listen counts distributed across days of the week or other selected durations.

### Backend

- Fully implemented data aggregation logic to compute listen counts per artist over time.
- Supports flexible time ranges (e.g., past 7 days) with daily granularity.
- Groups listens by user and date, and maps them to artist identities using MusicBrainz data.

### Frontend

- Integrated with the backend API to fetch real-time user data.
- Utilizes `@nivo/stream` to render a responsive and interactive stream graph.
- Key features include:
  - Color-coded layers per artist.
  - Hover tooltips to show listen counts by artist per day.
  - Scroll and responsive layout handling for varied screen sizes.
  - Artist labels and legends for clarity in dense visualizations.

## Completed Work

- Backend and frontend fully integrated.
- Time-range filtering and day-wise grouping implemented.
- Tooltips and accessibility improvements added.
- Manual verification conducted for accuracy and UI responsiveness.
- Visual layout tuned for handling users with many distinct artists.

## Current Limitation

Spark-based test coverage has not been executed due to missing relevant listen data in the test suite. This will be addressed once genre-tagged and artist-resolved listens are added to the test environment.

## Example Preview

image


frontend/js/src/utils/types.d.ts
```diff
@@ -305,6 +305,13 @@ declare type UserArtistActivityResponse = {
   }>;
 };
 
+export interface UserArtistEvolutionActivityResponse {
```
MonkeyDo: This is the small change that is making typescript freak out and all the tests fail :P

```suggestion
declare interface UserArtistEvolutionActivityResponse {
```

granth23: Thank you so much for this, was super confused why this was happening

amCap1712:
Unable to build image and deploy to spark because of frontend tslint errors.

MonkeyDo:
See https://github.com/metabrainz/listenbrainz-server/actions/runs/16147025214

> [test: frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx#L208](https://github.com/metabrainz/listenbrainz-server/pull/3314/files#annotation_36467091231)
Argument of type 'number | undefined' is not assignable to parameter of type 'number'.

anshg1214:
@granth23 I wanted to review this PR and deploy to test.lb. Can you fix the merge conflicts on this PR?

granth23:
> @granth23 I wanted to review this PR and deploy to test.lb. Can you fix the merge conflicts on this PR?

Yes will wind it up today itself

anshg1214:
@granth23 Can you add mock API Responses in the description? It'll be easier to test the PR

granth23:
Hey @anshg1214 thanks for resolving the error.
Regarding the mock API response, I am facing some spark issues right now trying to get them resolved, for the time being could you use the attached JSON file as the mock response.
[mockup.json](https://github.com/user-attachments/files/21813753/mockup.json)

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+  rawData.forEach((timeUnit) => {
+    if (timeUnit && typeof timeUnit === "object") {
+      Object.keys(timeUnit).forEach((key) => {
+        // Skip the 'id' field - it's metadata, not an artist
+        if (key !== "id") {
+          allArtistNames.add(key);
+        }
+      });
+    }
+  });
```
MonkeyDo: I don't think this is used anywhere. Is there a reason for keeping it?

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+        // Skip the 'id' field and only process actual artist data
+        if (artist !== "id" && typeof count === "number") {
+          if (!artistTotals[artist]) {
+            artistTotals[artist] = 0;
+          }
+          artistTotals[artist] += count;
+        }
+      });
+    }
+  });
```
MonkeyDo: Suggestion for a slightly more concise version using [lodash's `toPair`](https://lodash.com/docs/4.17.15#toPairs) which will return an empty array when not called with an object (undefined, bull, other object types...)

```suggestion
  const artistTotals = rawData
    .flatMap(toPairs)
    .reduce((totals, [artist, count]) => {
      // Skip the 'id' field and only process actual artist data
      if (artist !== "id" && typeof count === "number") {
        totals[artist] = (totals[artist] || 0) + count;
      }
      return totals;
    }, {} as Record<string, number>);
```

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+
+  // Get top 5 artists by total listens
+  const topArtists = Object.entries(artistTotals)
+    .sort(([, a], [, b]) => b - a)
+    .slice(0, 5)
```
MonkeyDo: Does this mean we only ever show the top 5 artists in this chart?

I think we could show more if we have the data, maybe up to 10?
Would have to see what the graphs look like in that case, but seems like a shame to show only top5 IMO

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+      "Tuesday",
+      "Wednesday",
+      "Thursday",
+      "Friday",
+      "Saturday",
+      "Sunday",
+    ];
+  }
+  if (timeRange.includes("month")) {
+    return Array.from({ length: 30 }, (_, i) => (i + 1).toString());
```
MonkeyDo: This assumes that all months are 30 days?

You can use this instead from the [date-fns library](https://date-fns.org/v4.1.0/docs/getDaysInMonth):
`getDaysInMonth(new Date())`

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+      "August",
+      "September",
+      "October",
+      "November",
+      "December",
+    ];
+  }
+  if (timeRange.includes("all_time")) {
+    const currentYear = new Date().getFullYear();
+    const safeOffsetYear = offsetYear || 2020;
```
MonkeyDo: Can you explain what the safeOffset is, and why this magic number of the year 2020?
How well will that hardcoded value work in 10 years for example?

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+    const yearRange = currentYear - safeOffsetYear + 1;
+    return Array.from({ length: yearRange }, (_, i) =>
```
MonkeyDo: In my opinion this is not very readable, so we might as well use utility functions from date-fns (eachYearOfInterval, format and parse imported from there).
Something like this should work:
```ts
const now = new Date();
return eachYearOfInterval({
  start: parse(safeOffsetYear + 1,"yyyy", now)),
  end: now
}).map(d => format(d,"yyyy"))
```

Not sure if the `+ 1` added to the offset is needed in this case, because I don't really understand what it represents :P
With the +1, the array of years will *not* include the offset year itself:

image


Not sure if it is because of the mock data I am using, but if I am on the "all time" range I see data for the bright future of 2026, which could be related to that +1? :
image

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+    );
+  }
+  return ["Period 1", "Period 2", "Period 3", "Period 4", "Period 5"];
+};
+
+// Custom tooltip function moved outside of render
+const renderCustomTooltip = (tooltipProps: any, orderedTimeUnits: string[]) => {
+  const { slice } = tooltipProps;
+
+  if (!slice || typeof slice.index === "undefined") {
```
MonkeyDo: How do we end up with the string "undefined"?
Sounds like maybe some transforation step calls toString on undefined values? Or is it an artifact of nivo?

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+        padding: "9px 12px",
+        border: "1px solid #ccc",
+        borderRadius: "4px",
+        fontSize: "12px",
+        boxShadow: "0 2px 4px rgba(0,0,0,0.1)",
+        maxWidth: "200px",
+      }}
+    >
+      <div style={{ marginBottom: "4px", fontWeight: "bold" }}>
+        {orderedTimeUnits[slice.index] || `Time Unit ${slice.index + 1}`}
```
MonkeyDo: I think this is a more human friendly text, like you did in `getOrderedTimeUnits` fallback 
```suggestion
        {orderedTimeUnits[slice.index] || `Period ${slice.index + 1}`}
```

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+          ))}
+    </div>
+  );
+};
+
+// Helper function for rendering different legend texts
+const getLegendText = (timeRange: UserStatsAPIRange) => {
+  if (timeRange === "week") return "Days of Week";
+  if (timeRange === "month") return "Days of Month";
+  if (timeRange === "year") return "Months";
```
MonkeyDo:
```suggestion
  if (timeRange === "week") return "Days of the week";
  if (timeRange === "month") return "Days of the month";
  if (timeRange === "year") return "Months";
```

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,452 @@
...
+  React.useEffect(() => {
+    const checkMobile = () => {
+      setIsMobile(window.innerWidth <= 768);
+    };
+
+    checkMobile();
+    window.addEventListener("resize", checkMobile);
+
+    return () => window.removeEventListener("resize", checkMobile);
+  }, [setIsMobile]);
```
MonkeyDo: We have a hook you can use for this which results in a one-liner and removes the state and this effect:
`const isMobile = useMediaQuery({ maxWidth: 767 });`

https://github.com/metabrainz/listenbrainz-server/blob/e94ac084f0d3381ae1ab9511586b90012cdc575a/frontend/js/src/explore/fresh-releases/utils.tsx#L32

frontend/js/src/utils/APIService.ts
```diff
@@ -602,6 +602,30 @@ export default class APIService {
     return response.json();
   };
 
+  getUserArtistEvolutionActivity = async (
+    userName?: string,
+    range: UserStatsAPIRange = "all_time"
+  ): Promise<UserArtistEvolutionActivityResponse> => {
+    let url;
+    if (userName) {
+      url = `${this.APIBaseURI}/stats/user/${userName}/artist-evolution-activity`;
```
MonkeyDo: I just changed this in the whole codebase in #3348 :
```suggestion
      url = `${this.APIBaseURI}/stats/user/${encodeURIComponent(userName)}/artist-evolution-activity`;
```

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+      retry: false,
+      staleTime: 0,
+      gcTime: 0, // Changed from cacheTime to gcTime (React Query v4+)
+    },
+  },
+});
+
+const reactQueryWrapper = ({ children }: any) => (
+  <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+);
```
MonkeyDo: We have a fixture for this available:
`import { ReactQueryWrapper } from "../../test-react-query";`

If you find that you need those cancel calls in `beforeEach` and `afterEach`, you can import queryClient from the same file.

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+        {},
+        { wrapper: reactQueryWrapper }
+      );
+    });
+  
+    await waitFor(() => {
+      expect(screen.getByText("No artist evolution data available for this time period")).toBeInTheDocument();
+    }, { timeout: 8000 });
+  }, 15000);
+});
```
MonkeyDo, line 117: Would be nice to see a couple of tests for scenarios where the component renders correctly, i.e for example testing that the "Mock Stream Chart" is rendered

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: This was being used earlier in the side legend, will remove it

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Cool, implemented the same
Thanks for the suggestion

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Do you think we could give like a slider of how many to show?

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Okay

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: The safeOffsetYear is acting as a fallback starting year in case the API response doesn’t provide a valid offsetYear. The reason 2020 was used as a default is that LB only started collecting data around that time. That said, in practice, the API usually provides the correct offsetYear, so the hardcoded default is rarely used.

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
granth23: Okay, thanks for the suggestion

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Done

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Done

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Done

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Yes it might read awkwardly, switched to a stricter and tighter guard.

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: For now I’ve removed the inline magic number and introduced an EARLIEST_DATA_YEAR constant, so the intent is explicit and maintainable. In practice, we still prefer the API’s offsetYear, and this fallback is rarely used, but now it’s clear and easier to adjust in the future.

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
granth23: Thanks! I added positive‑path tests that mock @nivo/stream and verify the chart renders. The tests cover both user and sitewide variants and assert via a data-testid="artist-evolution-stream" hook (added to the component) as well as the mocked “Mock Stream Chart”. I also included a failure-path test that surfaces the API error message.

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Refactored getAllTimeYearLabels to use eachYearOfInterval directly—simpler, no +1, and capped at current year to avoid future labels. Fixes the off-by-one; can add a test to assert last label = current year if needed.

frontend/js/src/utils/APIService.ts
granth23: Done

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
MonkeyDo: > Do you think we could give like a slider of how many to show?

Yes, that would work. Either that or a dropdown with a few options (5,10,15,20? something like that, generated automatically considering the number of artists available in the calculated stats)

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
MonkeyDo: > For now I’ve removed the inline magic number and introduced an EARLIEST_DATA_YEAR constant, so the intent is explicit and maintainable. In practice, we still prefer the API’s offsetYear, and this fallback is rarely used, but now it’s clear and easier to adjust in the future.

Thanks for the clarification, and indeed that constant is much more explicit.
I'm not sure about 2020 as a starting point though, "LB only started collecting data around that time" is not correct, not sure if you meant something else but LB started collecting user data in 2015, the beginning of the project.

But considering I can import my historical data from lastfm or spotify, you can expect that for the "all time" range we would be able to show older dates. I would recommend 2002, the year LastFM got its start. We are much less likely to have data from before that date.
Not sure if it makes sense in the context of this default offset. If it's too old than perhaps as default the LB start year (2015) makes more sense.

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Cool will implement the same based on the current SQL query

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Yes, I’ve removed the hardcoded usage entirely by introducing the EARLIEST_DATA_YEAR constant, which is derived from the user’s own available data. The only remaining references to 2020 are in fallback cases where a user has no data for the stat, so they don’t affect actual results.

frontend/js/tests/user/stats/UserArtistActivity.test.tsx
```diff
@@ -1,112 +1,104 @@
+// frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
```
MonkeyDo: ❓ Why this comment?
Between the odd comment and the fact that this is the wrong file, it makes me think you're using an LLM to write your tests instead of taking the time to focus on them yourself, to be honest.

frontend/js/tests/user/stats/UserArtistActivity.test.tsx
MonkeyDo, line 1: Why are the tests in this file being replaced?
They were existing tests for the  UserArtistActivity component, why did you repurpose the test file for UserArtistEvolutionActivity ?

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: { retry: false, staleTime: 0, gcTime: 0 },
+  },
+});
+
+const reactQueryWrapper = ({ children }: any) => (
+  <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+);
```
MonkeyDo: You can import both of these from a utility file: 
https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/tests/test-react-query.tsx

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+    const handlers = [
+      // API now returns RAW rows; tests use that fixture
+      http.get("/1/stats/user/foobar/artist-evolution-activity", async ({ request }) => {
+        const url = new URL(request.url);
+        const range = url.searchParams.get("range");
+
+        switch (range) {
+          case "week":
+            return HttpResponse.json(userArtistEvolutionActivityResponse);
+          case "empty":
```
MonkeyDo: Where is this from?
When can the range be equal to the string "empty"?

Also, why redefine this scenario in the test at line 113 instead of using one of the ranges  ("year" I suppose) like you do for "month" below?

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+      );
+    });
+
+    await waitFor(() => {
+      expect(screen.getByTestId("artist-evolution")).toBeInTheDocument();
+    });
+  });
+
+  it("displays error message when API call fails", async () => {
+    await act(async () => {
```
MonkeyDo, line 70: You don't need to wrap these with act at all.

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+  it("renders correctly", async () => {
+    await act(async () => {
+      renderWithProviders(
+        <ArtistEvolutionActivityStreamGraph {...userProps} />, {}, { wrapper: reactQueryWrapper }
+      );
+    });
+
+    await waitFor(() => {
+      expect(screen.getByTestId("artist-evolution")).toBeInTheDocument();
+    });
```
MonkeyDo: For reference, I use this cheatsheet all the time. The grid at the top is useful to determine which type ofquery to use when you need to await or not:
https://testing-library.com/docs/react-testing-library/cheatsheet/#queries

```suggestion
    await screen.findByTestId("artist-evolution");
```

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+    await waitFor(
+      () => {
+        // Error icon from FontAwesome when hasError=true
+        const errorIcon = screen.getByRole("img", { hidden: true });
+        expect(errorIcon).toHaveAttribute("data-icon", "circle-exclamation");
+      },
+      { timeout: 8000 }
+    );
+  }, 15000);
```
MonkeyDo: You don't need this timeout, nor the 15 second timing for the test.
Same for the test below.

MonkeyDo (changes requested):
Don't know what happened with the test files, but you need to revert the changes made to the wrong test files and integrate the changes you made into the correct test file.

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,550 @@
...
+      <div
+        className={`d-flex align-items-center justify-content-between ${
+          isMobile ? "mb-3" : ""
+        } flex-wrap mt-3`}
+      >
+        <h3 className="capitalize-bold m-0">Artist Evolution</h3>
+        <div className="d-flex align-items-center flex-shrink-0">
+          <span className="me-2">Top artists</span>
+          <div className="me-2" style={{ width: isMobile ? 140 : 220 }}>
+            <Slider
```
MonkeyDo: If I'm honest, I find the slider big and clunky, and I'm not convinced it's better than a simple dropdown with a couple of options.
What do you think about this?

Image

Or alternatively, with a number input?

Image

Also, instead of using the me-2 classes for margin, you can use the css class `gap-2` on the flex parent element to apply a gap between each child

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -297,26 +376,28 @@ export default function ArtistEvolutionActivityStreamGraph(
...
 
-  const orderedTimeUnits = getOrderedTimeUnits(
-    range,
-    rawData.payload.offset_year,
-    (rawData as any).payload.from_ts,
-    (rawData as any).payload.to_ts
+  const artistHref = React.useCallback(
+    (name: string) => {
+      const mbid = artistMap[name];
+      return mbid ? `/artist_${mbid}` : "#";
```
MonkeyDo: This is incorrect, the link should be `/artist/${mbid}`

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -446,7 +527,12 @@ export default function ArtistEvolutionActivityStreamGraph(
                           style: { itemTextColor: "#000000" },
                         },
                       ],
-                    },
+                      onClick: (datum: any) => {
+                        const name = datum.label;
+                        const mbid = artistMap[name];
+                        if (mbid) window.location.href = `/artist_${mbid}`;
```
MonkeyDo: Same here, should be  `/artist/${mbid}`.

You also probably want to use `navigate` from the react-router package rather than navigating like this, so as not to break the SPA navigation.
Example of usage here:
https://github.com/metabrainz/listenbrainz-server/blob/8ed93eb9f8c12cb465e810de71b7ccf2b9e007fb/frontend/js/src/artist/ArtistPage.tsx#L270-L273

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,550 @@
...
+                      },
+                    },
+                    grid: {
+                      line: {
+                        stroke: "#dddddd",
+                        strokeWidth: 1,
+                      },
+                    },
+                  }}
+                  legends={[
```
MonkeyDo, line 436: I'm not a big fan of the legend preventing the graph from taking the available horizontal space on the page.
It looks better to me on smaller screen sizes when the legend is underneath.

Another option would be to anchor it "top-left" with a smaller translateX value (say 20 for example), and hope that the graph lines won't cover it entirely depending on the data?
Image

frontend/js/tests/user/stats/UserArtistActivity.test.tsx
granth23: So my frontend tests usually run on my Windows setup, while the full LB stack is on the Linux (wolf) setup. To move stuff over I just dumped all my edits into one scratch file with filenames written on top. That random filename comment you saw was just left over from that, and I also ended up pasting changes into the wrong test file by mistake. I’ll fix it up and move things where they actually belong.

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
granth23: Fixed

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
granth23: That slipped in from an early stub. HAve removed the same.

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
granth23: FIxed

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
granth23: Fixed

frontend/js/tests/user/stats/UserArtistEvolutionActivity.test.tsx
granth23: Removed

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: FIxed

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Fixed

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: Used the artist map dropdown as reference and have implemented the dropdown for 5/10/15

frontend/js/tests/user/stats/UserArtistActivity.test.tsx
granth23: I messed up while copying pasting from the Windows env to the Linux env, have restored it now

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
granth23: image
This could be one issue, giving the entire legend a background overlay isn't possible Nivo doesn't support that else we will have to make a separate div for the legend, we can give it a row by row overlay though

MonkeyDo: Yes, I thought that could end up looking bad.
I think with the row-by-row overlay it would probably look fine, but I've never tried.

I think it'll be easier than to recreate a legend component...

listenbrainz/webserver/views/stats_api.py
```diff
@@ -590,6 +591,103 @@ def get_genre_activity(user_name: str):
     return jsonify({"result": genre_activity})
 
 
+def _transform_artist_evolution_activity_data(raw_data, stats_range):
```
amCap1712: The user stats and sitewide stats API should return similar structured data so not sure why this function is being used in sitewide stats but not user stats.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -590,6 +591,103 @@ def get_genre_activity(user_name: str):
...
+        
+        result.append(result_item)
+    
+    return result, offset_year
+
+
+@stats_api_bp.get("/user/<user_name>/artist-evolution-activity")
+@crossdomain
+@ratelimit()
+def get_artist_evolution_activity(user_name: str):
```
amCap1712, line 659: Missing documentation.

listenbrainz/webserver/views/stats_api.py
```diff
@@ -1335,6 +1433,34 @@ def get_sitewide_artist_activity():
     return jsonify({"result": result})
 
 
+@stats_api_bp.get("/sitewide/artist-evolution-activity")
+@crossdomain
+@ratelimit()
+def get_sitewide_artist_evolution_activity():
```
amCap1712, line 1528: Missing documentation.

granth23: done

listenbrainz/webserver/views/stats_api.py
granth23: done

listenbrainz/webserver/views/stats_api.py
granth23: fixed, the user stats api is the correct one so fixed sitewide accordingly

MonkeyDo (changes requested):
The previous review comment about the legend has not been addressed, so we're basically at the same point.

Other changes look good.

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -0,0 +1,569 @@
...
+    hasError = false,
+    errorMessage = "",
+  } = loaderData || {};
+
+  const isMobile = useMediaQuery("(max-width: 767px)");
+
+  const [topN, setTopN] = React.useState<number>(10);
+  const onTopNChange: React.ChangeEventHandler<HTMLSelectElement> = (e) => {
+    setTopN(parseInt(e.target.value, 10));
+  };
```
MonkeyDo, line 294: This was supposedly introduced to implement the topN select, but the implementation was changed to use a dropdown instead. 
I assume it was changed to follow the other dropdowns on the page, but I think the select is about perfect for this job, and has better accessibility support. (we should probably change the other dropdown to a select, to be honest.)

And so concise after a small refactor!
23 lines reduced to 4:
```tsx
<select className="form-select" onChange={onTopNChange}>
  {[5, 10, 15].map((num) => {
    return <option value={num}>{num}</option>;
  })}
</select>
```

Image

If you still want to use a dropdown instead, you can remove this unused function

granth23:
> The previous review comment about the legend has not been addressed, so we're basically at the same point.
> 
> Other changes look good.

Actually was waiting for Era to be merged so, can merge conflicts as well, I have done the implementation of same on local will make the changes and wind it up

frontend/js/src/utils/types.d.ts
```diff
@@ -322,15 +322,13 @@ declare type UserEraActivityResponse = {
 
 declare type UserArtistEvolutionActivityResponse = {
   payload: {
-    artist_evolution_activity: Array<{
-      [artistName: string]: number | string;
-    }>;
-    offset_year?: number;
+    user_id: string;
+    artist_evolution_activity: RawRow[];
```
MonkeyDo: Please also move the RawRow type to this file, and rename it to something less generic. Maybe "RawUserArtistEvolutionRow" or something like that?

granth23: done

MonkeyDo (approved):
:+1:

listenbrainz/webserver/views/stats_api.py
amCap1712: In that case, this function is not used anywhere AFAICS and should be removed.

granth23: Done

data/model/user_entity.py
```diff
@@ -5,13 +5,14 @@
...
-from data.model.user_genre_activity import GenreActivityRecord
 from data.model.user_recording_stat import RecordingRecord
 from data.model.user_release_group_stat import ReleaseGroupRecord
 from data.model.user_release_stat import ReleaseRecord
+from data.model.user_artist_evolution_activity import ArtistEvolutionActivityRecord
+from data.model.user_genre_activity import GenreActivityRecord
 
 EntityRecord = Union[
-    RecordingRecord, ReleaseGroupRecord, ReleaseRecord, ArtistRecord, GenreActivityRecord, EraActivityRecord
+    RecordingRecord, ReleaseGroupRecord, ReleaseRecord, ArtistRecord, ArtistEvolutionActivityRecord, GenreActivityRecord, EraActivityRecord, ArtistEvolutionActivityRecord
```
amCap1712: Remove duplicate
```suggestion
    RecordingRecord, ReleaseGroupRecord, ReleaseRecord, ArtistRecord, GenreActivityRecord, EraActivityRecord, ArtistEvolutionActivityRecord
```

PR #3316, opened 2025-07-02
Integrating Internet Archive recordings into Brainzplayer 

RayyanSeliya:
## Summary

This PR adds support for Internet Archive as a new data source in BrainzPlayer. Users can now play tracks from the Internet Archive, see the IA icon, and set its priority alongside other music services in the player settings.

## Key Changes

- Added Internet Archive player component with hardcoded track (for now).
- Integrated Internet Archive into BrainzPlayer and settings UI.
- Added Internet Archive icon for better visual distinction not an official one will add it later .
- Users can set the priority/order of Internet Archive in the player settings.

## Demo

I have attached a video of the player see below 

https://github.com/user-attachments/assets/536ae6e2-bfea-4732-a4e5-b1f3f41a6155

### DIFF
--- a/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
@@ -1,5 +1,5 @@
 import { IconProp } from "@fortawesome/fontawesome-svg-core";
-import { faPlayCircle } from "@fortawesome/free-solid-svg-icons";
+import { faPlayCircle, faArchive } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
   has as _has,
@@ -30,7 +30,9 @@ import SoundcloudPlayer from "./SoundcloudPlayer";
 import SpotifyPlayer from "./SpotifyPlayer";
 import YoutubePlayer from "./YoutubePlayer";
 import AppleMusicPlayer from "./AppleMusicPlayer";
+import InternetArchivePlayer from "./InternetArchivePlayer";
 import {
+  dataSourcesInfo,
   DataSourceKey,
   defaultDataSourcesPriority,
 } from "../../settings/brainzplayer/BrainzPlayerSettings";
@@ -55,7 +57,8 @@ export type DataSourceTypes =
   | SpotifyPlayer
   | YoutubePlayer
   | SoundcloudPlayer
-  | AppleMusicPlayer;
+  | AppleMusicPlayer
+  | InternetArchivePlayer;
 
 export type DataSourceProps = {
   show: boolean;
@@ -179,17 +182,21 @@ export default function BrainzPlayer() {
       SoundcloudPlayer.hasPermissions(soundcloudAuth) &&
       "soundcloud",
     youtubeEnabled && "youtube",
-  ].filter(Boolean) as Array<DataSourceKey>;
+    true && "internetArchive",
+  ].filter(Boolean) as Array<DataSourceKey | "internetArchive">;
 
-  const sortedDataSources = dataSourcesPriority.filter((key) =>
-    enabledDataSources.includes(key)
-  );
+  const sortedDataSources = dataSourcesPriority
+    .filter((key) => enabledDataSources.includes(key))
+    .concat(
+      enabledDataSources.includes("internetArchive") ? ["internetArchive"] : []
+    );
 
   // Refs
   const spotifyPlayerRef = React.useRef<SpotifyPlayer>(null);
   const youtubePlayerRef = React.useRef<YoutubePlayer>(null);
   const soundcloudPlayerRef = React.useRef<SoundcloudPlayer>(null);
   const appleMusicPlayerRef = React.useRef<AppleMusicPlayer>(null);
+  const internetArchivePlayerRef = React.useRef<InternetArchivePlayer>(null);
   const dataSourceRefs: Array<React.RefObject<
     DataSourceTypes
   >> = React.useMemo(() => {
@@ -207,6 +214,9 @@ export default function BrainzPlayer() {
           break;
         case "appleMusic":
           dataSources.push(appleMusicPlayerRef);
+          break;
+        case "internetArchive":
+          dataSources.push(internetArchivePlayerRef);
           break;
         default:
         // do nothing
@@ -1119,6 +1129,28 @@ export default function BrainzPlayer() {
             handleSuccess={handleSuccess}
           />
         )}
+
+        {/* Add this line for Internet Archive */}
+        <InternetArchivePlayer
+          show={
+            brainzPlayerContextRef.current.isActivated &&
+            dataSourceRefs[
+              brainzPlayerContextRef.current.currentDataSourceIndex
+            ]?.current instanceof InternetArchivePlayer
+          }
+          ref={internetArchivePlayerRef}
+          playerPaused={brainzPlayerContextRef.current.playerPaused}
+          onPlayerPausedChange={playerPauseChange}
+          onProgressChange={progressChange}
+          onDurationChange={durationChange}
+          onTrackInfoChange={throttledTrackInfoChange}
+          onTrackEnd={playNextTrack}
+          onTrackNotFound={failedToPlayTrack}
+          handleError={handleError}
+          handleWarning={handleWarning}
+          handleSuccess={handleSuccess}
+          onInvalidateDataSource={invalidateDataSource}
+        />
       </BrainzPlayerUI>
     </div>
   );
--- a/frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
@@ -1,0 +1,126 @@
+import React from "react";
+import { faArchive } from "@fortawesome/free-solid-svg-icons";
+import { DataSourceProps, DataSourceType } from "./BrainzPlayer";
+
+// Hardcoded track for now
+const HARDCODED_TRACK = {
+  track_id: "https://archive.org/details/00TtuloInttrprete66",
+  name:
+    "Los Norteños / Cuando Canta La Lluvia - Perez Prado y Hermanas Montoya (Very Rare Recordings)",
+  artist: ["Pérez Prado y Orquesta con Hermanas Montoya"],
+  album: "RCA Victor #70-9428",
+  stream_urls: [
+    "https://archive.org/download/00TtuloInttrprete66/Cuando%20Canta%20La%20Lluvia.m4a",
+    "https://archive.org/download/00TtuloInttrprete66/Cuando%20Canta%20La%20Lluvia.mp3",
+  ],
+  artwork_url:
+    "https://archive.org/download/00TtuloInttrprete66/Cuando%20Canta%20La%20Lluvia.png",
+};
+
+export default class InternetArchivePlayer
+  extends React.Component<DataSourceProps>
+  implements DataSourceType {
+  public name = "internetArchive";
+  public domainName = "archive.org";
+  public icon = faArchive;
+  public iconColor = "#6c757d";
+  audioRef: React.RefObject<HTMLAudioElement>;
+
+  constructor(props: DataSourceProps) {
+    super(props);
+    this.audioRef = React.createRef();
+  }
+
+  playListen = () => {
+    const {
+      onPlayerPausedChange,
+      onTrackInfoChange,
+      onDurationChange,
+    } = this.props;
+    if (this.audioRef.current) {
+      this.audioRef.current.currentTime = 0;
+      this.audioRef.current.play();
+      onPlayerPausedChange(false);
+      onTrackInfoChange(
+        HARDCODED_TRACK.name,
+        HARDCODED_TRACK.track_id,
+        HARDCODED_TRACK.artist.join(", "),
+        HARDCODED_TRACK.album,
+        [{ src: HARDCODED_TRACK.artwork_url }]
+      );
+      onDurationChange(this.audioRef.current.duration * 1000 || 0);
+    }
+  };
+
+  togglePlay = () => {
+    const { playerPaused, onPlayerPausedChange } = this.props;
+    if (!this.audioRef.current) return;
+    if (playerPaused) {
+      this.audioRef.current.play();
+      onPlayerPausedChange(false);
+    } else {
+      this.audioRef.current.pause();
+      onPlayerPausedChange(true);
+    }
+  };
+
+  seekToPositionMs = (ms: number) => {
+    const { onProgressChange } = this.props;
+    if (this.audioRef.current) {
+      this.audioRef.current.currentTime = ms / 1000;
+      onProgressChange(ms);
+    }
+  };
+
+  canSearchAndPlayTracks = () => false; // For now, no search
+
+  datasourceRecordsListens = () => false;
+
+  handleAudioEnded = () => {
+    const { onTrackEnd } = this.props;
+    onTrackEnd();
+  };
+
+  handleTimeUpdate = () => {
+    const { onProgressChange } = this.props;
+    if (this.audioRef.current) {
+      onProgressChange(this.audioRef.current.currentTime * 1000);
+    }
+  };
+
+  handleLoadedMetadata = () => {
+    const { onDurationChange } = this.props;
+    if (this.audioRef.current) {
+      onDurationChange(this.audioRef.current.duration * 1000);
+    }
+  };
+
+  render() {
+    const { show, playerPaused } = this.props;
+    if (!show) return null;
+    return (
+      <div className="internet-archive-player">
+        <img
+          src={HARDCODED_TRACK.artwork_url}
+          alt={HARDCODED_TRACK.name}
+          width={60}
+          style={{ marginRight: 10 }}
+        />
+        <audio
+          ref={this.audioRef}
+          src={HARDCODED_TRACK.stream_urls[0]}
+          onEnded={this.handleAudioEnded}
+          onTimeUpdate={this.handleTimeUpdate}
+          onLoadedMetadata={this.handleLoadedMetadata}
+          autoPlay={!playerPaused}
+        >
+          <track kind="captions" />
+        </audio>
+        <div>
+          <strong>{HARDCODED_TRACK.artist.join(", ")}</strong> –{" "}
+          {HARDCODED_TRACK.name}
+        </div>
+      </div>
+    );
+  }
+}
--- a/frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx
+++ b/frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx
@@ -5,13 +5,13 @@ import {
   faSoundcloud,
   faYoutube,
 } from "@fortawesome/free-brands-svg-icons";
+import { faArchive, faGripLines } from "@fortawesome/free-solid-svg-icons";
 import { Helmet } from "react-helmet";
 import { Link } from "react-router";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { toast } from "react-toastify";
 import ReactTooltip from "react-tooltip";
 import { ReactSortable } from "react-sortablejs";
-import { faGripLines } from "@fortawesome/free-solid-svg-icons";
 import { IconDefinition, IconProp } from "@fortawesome/fontawesome-svg-core";
 import Switch from "../../components/Switch";
 import GlobalAppContext from "../../utils/GlobalAppContext";
@@ -42,6 +42,11 @@ export const dataSourcesInfo = {
     icon: faApple,
     color: "#000000",
   },
+  internetArchive: {
+    name: "Internet Archive",
+    icon: faArchive,
+    color: "#6c757d",
+  },
 } as const;
 
 export type DataSourceKey = keyof typeof dataSourcesInfo;
@@ -52,6 +57,7 @@ export const defaultDataSourcesPriority = [
   "appleMusic",
   "soundcloud",
   "youtube",
+  "internetArchive",
 ] as DataSourceKey[];
 
 function BrainzPlayerSettings() {
@@ -165,6 +171,8 @@ function BrainzPlayerSettings() {
     currentUser?.auth_token,
     userPreferences,
   ]);
+
+  const internetArchivePlayerRef = React.useRef<any>(null);
 
   return (
     <>
--- a/frontend/js/src/utils/types.d.ts
+++ b/frontend/js/src/utils/types.d.ts
@@ -695,7 +695,7 @@ declare type BrainzPlayerSettings = {
   appleMusicEnabled?: boolean;
   brainzplayerEnabled?: boolean;
   dataSourcesPriority?: Array<
-    "spotify" | "youtube" | "soundcloud" | "appleMusic"
+    "spotify" | "youtube" | "soundcloud" | "appleMusic" | "internetArchive"
   >;
 };
 


frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -179,8 +184,10 @@ export default function BrainzPlayer() {
       SoundcloudPlayer.hasPermissions(soundcloudAuth) &&
       "soundcloud",
     youtubeEnabled && "youtube",
-  ].filter(Boolean) as Array<DataSourceKey>;
+    internetArchiveEnabled && "internetArchive",
+  ].filter(Boolean) as DataSourceKey[];
 
+  // Use the enabled sources to filter the priority list
```
MonkeyDo, line 248: So the issue you are having with the player is right here. We ran into the same issue for the Navidrome player GSOC project and I will direct you there in a minute for code snippets.

Basically, this comment is missing one detail: the priority list is pulled directly from the database, from user preferences JSONB column.
What this means is that your new service "internetArchive" does not appear in that list for any user, and because we use the incomplete `dataSourcesPriority` as a base to filter some out, "internetArchive" never makes it in the list of available data sources.

After discussing with m.Amanullah7, there are two main issues with new services added to BP:
1. not added to the brainzplayer datasources, preventing playback and
2. the service does not appear in the [BrainzPlayerSettings page](https://listenbrainz.org/settings/brainzplayer) to order the priority


For both issues, we can fix the source of the issue by patching the [getPageProps](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/utils/utils.tsx#L601-L605) function which takes user preferences sent by the server and makes them available in the global context.
We can ensure all missing values from `defaultDataSourcesPriority` (including your new service) are added at the end of the `dataSourcesPriority` array.

I will now refer you to the discussion thread on the other PR where I proposed some code snippets and where we discussed the issue and solution.
I think you should find enough info and code there to solve the issue, but if you are still unclear do get in touch with me.
https://github.com/metabrainz/listenbrainz-server/pull/3289#discussion_r2190273668

MonkeyDo: OK, I think I now have a simpler one-liner solution:
https://github.com/metabrainz/listenbrainz-server/pull/3289#discussion_r2204922536
One change in BrainzPlayer so the internet archive datasource is not filtered out, and one change in BrainzPlayerSettings so that Internet Archive appears in the priority list at the bottom of that page

frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx
```diff
@@ -159,13 +170,16 @@ function BrainzPlayerSettings() {
...
+    internetArchiveEnabled,
     brainzplayerEnabled,
     dataSourcesPriority,
     APIService,
     currentUser?.auth_token,
     userPreferences,
   ]);
 
+  const internetArchivePlayerRef = React.useRef<any>(null);
+
```
MonkeyDo: This looks pasted in the wrong component?

RayyanSeliya: Sure will  remove

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
RayyanSeliya: > OK, I think I now have a simpler one-liner solution: [#3289 (comment)](https://github.com/metabrainz/listenbrainz-server/pull/3289#discussion_r2204922536) One change in BrainzPlayer so the internet archive datasource is not filtered out, and one change in BrainzPlayerSettings so that Internet Archive appears in the priority list at the bottom of that page

hey @MonkeyDo thx for your suggestion and having a look to this issue I’ve implemented all the suggested changes for BrainzPlayer and BrainzPlayerSettings, including the union fix for the data source priority. I also updated the InternetArchivePlayer to use `componentDidUpdate` for triggering playback after setting the track, following React best practices.
When I hardcode a track in `componentDidMount`, playback works perfectly, so the player and audio logic are fine. However, when I use the API-driven approach (fetching a track from `/1/internet_archive/search`), the player doesn’t play, even though the API returns valid data and the state updates.
I suspect the issue might be with how or when playback is triggered from the UI, or possibly with how the listen object is passed down, rather than with the player component itself.
Is there a recommended pattern I should follow to align InternetArchivePlayer’s playback flow with the other player components? For example, should I implement a static method like` isListenFromThisService`, or is there a `utility`/`helper` I should use to ensure playback is triggered in the same way as the other sources? 
I am pretty sure that  i am missing some part of the code and  waiting for your guidance to ace this ! thx

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -35,15 +35,6 @@ export default class InternetArchivePlayer
     };
   }
 
-
-  componentDidUpdate(prevProps: DataSourceProps, prevState: State) {
```
MonkeyDo: I think one of the main issue with playback was the choice to use this lifecycle method to actually play the track.
I do see that we need to set the currentTrack in state because we use some elements of it for rendering (artwork_url, name, artists, etc.), however we can separate this rendering from the audio playing, which we will control programmatically with the `audioRef`.

I understand the fear that the state would not be updated in time to play the media, but when using `this.setState` you can pass a callback function as the second parameter, and this function is only called once the state is actually updated. This is what I do on line 84, calling this.playCurrentTrack as a callback after setting the currentTrack:
```ts
        this.setState(
          { currentTrack: data.results[0], loading: false },
          this.playCurrentTrack
        );
```

I suspect that the issues you were having might have been caused by the audio element being re-rendered because of `src={currentTrack.stream_urls[0]}`, which might cause concurrency issues trying to replace the media and play at the same time.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -86,7 +77,12 @@ export default class InternetArchivePlayer
       const data = await response.json();
 
       if (data.results && data.results.length > 0) {
-        this.setState({ currentTrack: data.results[0], loading: false });
+        // TODO: It might make sense to sanity-check the results to see if the first one is indeed the best match
+        // We do something like this in the AppleMusicPlayer with the fuzzysort library
+        this.setState(
+          { currentTrack: data.results[0], loading: false },
+          this.playCurrentTrack
```
MonkeyDo: Call this function as a callback *after* the state has been set.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -102,16 +98,21 @@ export default class InternetArchivePlayer
...
-  playCurrentTrack = () => {
+  playCurrentTrack = async () => {
     const {
       onPlayerPausedChange,
       onTrackInfoChange,
       onDurationChange,
     } = this.props;
     const { currentTrack } = this.state;
     if (this.audioRef.current && currentTrack) {
+      this.audioRef.current.src = currentTrack.stream_urls[0];
```
MonkeyDo: Instead of relying on the `src` attribute set during rendering of the element, we can set it programmatically here.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -102,16 +98,21 @@ export default class InternetArchivePlayer
...
       this.audioRef.current.currentTime = 0;
```
MonkeyDo: This line can be removed: once you call play(), it will be automatically reset to 0.
Currently this could potentially be causing some issues as I think it tries to play whichever is the currentTrack from position 0, and depending on when everything is called it could be the previous track.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -102,16 +98,21 @@ export default class InternetArchivePlayer
...
-      this.audioRef.current.play();
+      try {
+        await this.audioRef.current.play();
```
MonkeyDo, line 150: I `await` this call becaus eit is an asynchronous function.

*ALWAYS* wrap your async function calls in a try-catch block, or if using the promise syntax instead of async-await, use .catch

Otherwise, any error thrown will be silently ignored.

Either 
```
try {
        await this.audioRef.current.play();
catch(error) { .....
```

OR

```
this.audioRef.current.play().catch(error=>{ .... })
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -102,16 +98,21 @@ export default class InternetArchivePlayer
...
+      } catch (error) {
+        console.error("InternetArchive playback error:", error);
```
MonkeyDo: You'll want to show something to the user so they have some feedback on what happened. See the `handleError` prop and how it is used in other players.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -126,15 +127,19 @@ export default class InternetArchivePlayer
     }
   };
 
-  togglePlay = () => {
+  togglePlay = async () => {
```
MonkeyDo, line 169: Same as above, making this function async so I can `await` the play and pause function calls, only so that I can wrap them in try-catch to catch any errors.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -126,15 +127,19 @@ export default class InternetArchivePlayer
...
+    try {
+      if (playerPaused) {
+        await this.audioRef.current.play();
+        onPlayerPausedChange(false);
+      } else {
+        await this.audioRef.current.pause();
+        onPlayerPausedChange(true);
+      }
+    } catch (error) {
+      console.error("InternetArchive playback error:", error);
```
MonkeyDo: Same as above, use `handleError` so there is visual feedback on the page that something went wrong.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -170,11 +175,11 @@ export default class InternetArchivePlayer
             )}
             <audio
               ref={this.audioRef}
-              src={currentTrack.stream_urls[0]}
```
MonkeyDo: I think this src and/or the autoPlay props was probably what was causing the player not to work properly, but I am not certain.

In any case, we want autoplay to be true, otherwise we cannot call play() without a user interaction.
Similarly, we don't need to re-render the audio element when the track changes, instead we can set it programmatically (line 109).

I also added controls={false} to explicitly disabled the default browser media controls, just in case.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -170,11 +175,11 @@ export default class InternetArchivePlayer
...
               onEnded={this.handleAudioEnded}
               onTimeUpdate={this.handleTimeUpdate}
               onLoadedMetadata={this.handleLoadedMetadata}
```
MonkeyDo: Would it make sense to use `onDurationChange` instead, and renaming `handleLoadedMetadata`, considering the only thing it does is duration related?
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/durationchange_event

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
MonkeyDo: For reference (I left comments below), I think the issue was related to the rendering lifecycle (using componentDidUpdate) and a combo of setting the `src` and `autoplay` props on the audio element.
Not exactly sure which one was the culprit, though.

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+  datasourceRecordsListens = () => false;
+
+  render() {
+    const { show, playerPaused } = this.props;
+    const { loading, currentTrack } = this.state;
+    if (!show) return null;
+
+    return (
+      <div className="internet-archive-player">
+        {loading && <div>Searching Internet Archive...</div>}
```
MonkeyDo: This is not required, and doesn't look very good currently.
We don't do this for other music players, so i would say just drop it.
I will leave other comments in the code to remove the whole "loading" state, which is not required anymore.

```suggestion
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+              controls={false}
+            >
+              <track kind="captions" />
+            </audio>
+            <div>
+              <strong>{currentTrack.artist.join(", ")}</strong> –{" "}
+              {currentTrack.name}
+            </div>
+          </>
+        )}
```
MonkeyDo: Some more changes:

- Don't render the audio element conditionally, it is not required and may be the cause of the issues you are having?
- The track name rendering here is not necessarry, this is not how we display the text on the player (we send it using the onTrackInfoChange prop instead)
- ~~I don't know what that `<track kind="captions" />` element is doing there, looks like copy-pasted boilerplate or LLM gaff~~
- the marginRight style on the image is not necessary

```suggestion
        {currentTrack?.artwork_url && (
          <img
            src={currentTrack.artwork_url}
            alt={currentTrack.name}
            width={60}
          />
        )}
        <audio
          ref={this.audioRef}
          onEnded={this.handleAudioEnded}
          onTimeUpdate={this.handleTimeUpdate}
          onDurationChange={this.handleLoadedMetadata}
          autoPlay
          controls={false}
        />
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+        this.setState(
+          { currentTrack: data.results[0], loading: false },
+          this.playCurrentTrack
+        );
+      } else {
+        this.setState({ loading: false, currentTrack: null });
+        onTrackNotFound();
+      }
+    } catch (err) {
+      this.setState({ loading: false, currentTrack: null });
```
MonkeyDo:
```suggestion
      this.setState({ currentTrack: null });
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+      if (data.results && data.results.length > 0) {
+        // TODO: It might make sense to sanity-check the results to see if the first one is indeed the best match
+        // We do something like this in the AppleMusicPlayer with the fuzzysort library
+        this.setState(
+          { currentTrack: data.results[0], loading: false },
+          this.playCurrentTrack
+        );
+      } else {
+        this.setState({ loading: false, currentTrack: null });
```
MonkeyDo:
```suggestion
        this.setState({ currentTrack: null });
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+      const response = await fetch(
+        `/1/internet_archive/search?${params.toString()}`
+      );
+      const data = await response.json();
+
+      if (data.results && data.results.length > 0) {
+        // TODO: It might make sense to sanity-check the results to see if the first one is indeed the best match
+        // We do something like this in the AppleMusicPlayer with the fuzzysort library
+        this.setState(
+          { currentTrack: data.results[0], loading: false },
```
MonkeyDo:
```suggestion
          { currentTrack: data.results[0] },
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+    const { onTrackNotFound, handleError } = this.props;
+    const trackName = getTrackName(listen);
+    const artistName = getArtistName(listen);
+
+    if (!trackName && !artistName) {
+      onTrackNotFound();
+      return;
+    }
+
+    this.setState({ loading: true, currentTrack: null });
```
MonkeyDo: This can be removed entirely, now we don't need the loading state and the currentTrack is going to be set at the end of this function regardless of outcome.
```suggestion
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+  public domainName = "archive.org";
+  public icon = faArchive;
+  public iconColor = "#6c757d";
+  audioRef: React.RefObject<HTMLAudioElement>;
+
+  constructor(props: DataSourceProps) {
+    super(props);
+    this.audioRef = React.createRef();
+    this.state = {
+      loading: false,
```
MonkeyDo:
```suggestion
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+  track_id: string;
+  name: string;
+  artist: string[];
+  album?: string;
+  stream_urls: string[];
+  artwork_url?: string;
+};
+
+type State = {
+  loading: boolean;
```
MonkeyDo:
```suggestion
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,197 @@
...
+    }
+  };
+
+  searchAndPlayTrack = async (listen: any) => {
+    const { onTrackNotFound, handleError } = this.props;
+    const trackName = getTrackName(listen);
+    const artistName = getArtistName(listen);
+
+    if (!trackName && !artistName) {
+      onTrackNotFound();
```
MonkeyDo, line 100: Let's warn the user like we do in other music players:

https://github.com/metabrainz/listenbrainz-server/blob/a6f6f8e092c5b7ee7ce496db901d7ec577acba7e/frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx#L190-L194

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
MonkeyDo: Regarding the `<track>` element, my bad! I just saw that there is an ESLint warning if it is not provided.
That being said, I think we should remove it and silence the warning, because the empty track element could be confusing for visually impaired users.
I think considering we don't have caption text available, silencing the error is better than providing an empty track element and won't affect the user.

You can add this comment line above the `<audio ...` element:
`{/* eslint-disable-next-line jsx-a11y/media-has-caption */}`

listenbrainz/webserver/views/internet_archive_api.py
```diff
@@ -0,0 +1,15 @@
+from flask import Blueprint, request, jsonify
+from listenbrainz.webserver import ts_conn
+from listenbrainz.metadata_cache.internetarchive.store import search_ia_tracks
+
+internet_archive_api_bp = Blueprint('internet_archive_api', __name__)
+
+@internet_archive_api_bp.route('/internet_archive/search', methods=['GET'])
```
MonkeyDo: The prefix should be set when registering the blueprint in init.py, will leave comment there.
```suggestion
@internet_archive_api_bp.get('/search')
```

listenbrainz/webserver/__init__.py
```diff
@@ -435,3 +435,6 @@ def _register_blueprints(app):
 
     from listenbrainz.webserver.views.atom import atom_bp
     app.register_blueprint(atom_bp, url_prefix='/syndication-feed')
+
+    from listenbrainz.webserver.views.internet_archive_api import internet_archive_api_bp
+    app.register_blueprint(internet_archive_api_bp, url_prefix='/1')
```
MonkeyDo:
```suggestion
    app.register_blueprint(internet_archive_api_bp, url_prefix=API_PREFIX+"/internet_archive")
```

MonkeyDo (changes requested):
One more !

MonkeyDo (commented):
Apart from the remaining failing test, the UI part looks done to me.

frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx
```diff
@@ -381,6 +398,37 @@ function BrainzPlayerSettings() {
...
+            value="internetArchive"
+            checked={internetArchiveEnabled}
+            onChange={() => setInternetArchiveEnabled(!internetArchiveEnabled)}
+            switchLabel={
+              <span
+                className={`text-brand ${
+                  !internetArchiveEnabled ? "text-muted" : ""
+                }`}
+              >
+                <span className={internetArchiveEnabled ? "text-success" : ""}>
```
MonkeyDo: I think this is copied from other elements on the page probably, but the addition of this class does nothing.

```suggestion
                <span>
```

frontend/css/sass/brainzplayer.scss
```diff
@@ -701,3 +701,28 @@ $alert-thumbnail-size: 60px;
...
 }
+
+// Custom styling for Internet Archive icon
+svg.fa-internetarchive,
+.fa-internetarchive {
+  width: 1.2em !important;
+  height: 1.2em !important;
+
+  // Match the vertical alignment used by other music service icons
+  vertical-align: middle !important;
```
MonkeyDo: I don't think this styling should be applied everywhere equally.
There are three places the icon appears as far as I can tell:
- BrainzPlayer settings switch
- BrainzPlayer settings priority order
- BrainzPlayerUI

In both of the first two cases, I think the original width and height as well as the vertical alignment are closer to the other services icons.

In the BrainzPlayerUI, I can see that with default sizes the icon appears pretty big, but it seems to be the same dimensions as other icons.

I think we probably don't need these styles?

frontend/css/sass/brainzplayer.scss
```diff
@@ -701,3 +701,28 @@ $alert-thumbnail-size: 60px;
...
+}
+
+// When in a muted context, use currentColor to match other services
+.text-muted svg.fa-internetarchive,
+.text-muted .fa-internetarchive {
+  --fa-primary-color: currentColor !important;
+  --fa-secondary-color: currentColor !important;
+  --fa-primary-opacity: 1 !important;
+  --fa-secondary-opacity: 1 !important;
+}
```
MonkeyDo: I'm not sure what this is supposed to do. I tried disabling the styles and couldn't detect a change. Am I missing something?

The color is already defined in the [InternetArchivePlayer](https://github.com/metabrainz/listenbrainz-server/pull/3316/files#diff-93dcae7adff33e88ea7f019c1bb28a4f5a0ae9662d004c570fd167364f8f7508R25) and [BrainzPlayerSettings](https://github.com/metabrainz/listenbrainz-server/pull/3316/files#diff-b28ff595a4a059f4800f58a2f6e4f871c8458295ecb32c05800c0a88e83625c1R50) components, which applies it to the SVG element so changing the styles as above does not apply any change.
Same with `currentColor` which will equal the color defined in the SVG.

If I am missing something, please tell me. Otherwise I think all the style changes can be removed

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -1,5 +1,5 @@
 import { IconProp } from "@fortawesome/fontawesome-svg-core";
-import { faPlayCircle } from "@fortawesome/free-solid-svg-icons";
+import { faPlayCircle, faArchive } from "@fortawesome/free-solid-svg-icons";
```
MonkeyDo: Not in use anymore.
```suggestion
import { faPlayCircle } from "@fortawesome/free-solid-svg-icons";
```

frontend/js/src/common/brainzplayer/InternetArchivePlayer.tsx
```diff
@@ -0,0 +1,199 @@
...
+    const { currentTrack } = this.state;
+    if (!show) return null;
+
+    return (
+      <div className="internet-archive-player">
+        {currentTrack?.artwork_url && (
+          <img
+            src={currentTrack.artwork_url}
+            alt={currentTrack.name}
+            width={60}
```
MonkeyDo: This should already be taken care of by existing CSS without the need to hardcode the size here.

See https://github.com/metabrainz/listenbrainz-server/blob/e18111661afaf18834dfce4e138e380e674564cd/frontend/css/sass/brainzplayer.scss#L58-L72

frontend/css/sass/brainzplayer.scss
RayyanSeliya: @MonkeyDo Thanks for checking this out! I actually tested this and the CSS is needed for the icon to work properly.
The Internet Archive icon uses FontAwesome's duotone system (like Funkwhale does), so it needs those CSS custom properties to control the colors. Without the CSS, the icon stays the same color whether it's enabled or disabled.

With CSS: Icon brightens when enabled, fades when disabled
Without CSS: Icon stays the same color no matter what

I followed the exact same pattern as Funkwhale (lines 703-728 in the same file) which has been working fine. The React `color` prop sets the initial color, but the CSS handles the state changes when the `.text-muted` class gets applied.

frontend/css/sass/brainzplayer.scss
RayyanSeliya: you are absolutely right  @MonkeyDo  will remove those sizing styles and just keep the color-related CSS that handles the enabled/disabled state changes.

frontend/css/sass/brainzplayer.scss
MonkeyDo: Well, I tested again as well, and removing the fa-internetarchive css class does not change anything, neither when IA player is enabled nor disabled.
Not exactly sure how we arrive at different conclusions. Where did you see that without the CSS the color didn't change?
I tested by commenting out the custom CSS, and the icon color does change when enabling/disabling IA.
Are we looking at the same page?

image
image


image
image

frontend/js/src/utils/types.d.ts
```diff
@@ -704,10 +704,17 @@ declare type BrainzPlayerSettings = {
   spotifyEnabled?: boolean;
   soundcloudEnabled?: boolean;
   appleMusicEnabled?: boolean;
+  internetArchiveEnabled?: boolean;
   funkwhaleEnabled?: boolean;
+  internetArchiveEnabled?: boolean;
```
MonkeyDo: One more duplicate

RayyanSeliya: thx @MonkeyDo   for pointing out i just removed it !

frontend/css/sass/brainzplayer.scss
RayyanSeliya: yeah @MonkeyDo  my bad i have just removed the unnecessary css i also tried it was not changing anything ! the color was changing without the css too !

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+
+  afterAll(() => {
+    fetchMock.disableMocks();
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    fetchMock.resetMocks();
+    mockPlay.mockClear();
+    mockPause.mockClear();
```
MonkeyDo: Not very important, but the jest.clearAllMocks call above means you don't have to call it on each mock, so you can remove these lines.

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+          track_metadata: {
+            additional_info: {
+              origin_url: "https://archive.org/download/test-track/test.mp3",
+            },
+            artist_name: "Test Artist",
+            track_name: "Test Track",
+          },
+        };
+        expect(InternetArchivePlayer.isListenFromThisService(listen)).toBe(true);
+      });
```
MonkeyDo: Those two tests seem to be testing the same thing, namely `originURL.includes('archive.org')`

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+        },
+      };
+
+      await act(async () => {
+        playerRef.current?.playListen(listen);
+      });
+
+      await waitFor(() => {
+        expect(defaultProps.onTrackNotFound).toHaveBeenCalled();
+      });
```
MonkeyDo, line 292: Might be good to ensure it's not trying to play anything, considering you make the same assertion in the test above.
Generally, if I test a mechanism that should happen in scenario A, I make sure that the same mechanism does not happen in scenario B for completion.

You could also test that `mockPlay` was not called, for completion.

```suggestion
        expect(defaultProps.onPlayerPausedChange).not.toHaveBeenCalled();
      });
```

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+
+      await act(async () => {
+        playerRef.current?.playListen(listen);
+      });
+
+      expect(defaultProps.handleWarning).toHaveBeenCalledWith(
+        "We are missing a track title, artist or album name to search on Internet Archive",
+        "Not enough info to search on Internet Archive"
+      );
+      expect(defaultProps.onTrackNotFound).toHaveBeenCalled();
```
MonkeyDo, line 351: Let's be completionists and add the assertion here as well that the play function was not called.

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+    it("should seek to position correctly", () => {
+      const playerRef = React.createRef<InternetArchivePlayer>();
+      render(
+        <GlobalAppContext.Provider value={defaultContext}>
+          <InternetArchivePlayer {...defaultProps} ref={playerRef} />
+        </GlobalAppContext.Provider>
+      );
+
+      act(() => {
+        playerRef.current?.seekToPositionMs(30000); // 30 seconds
```
MonkeyDo: It would make sense to test the internal audioElement's currentTime before and after calling seekToPositionMs, since we set it in this function.

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+
+    it("should seek to position correctly", () => {
+      const playerRef = React.createRef<InternetArchivePlayer>();
+      render(
+        <GlobalAppContext.Provider value={defaultContext}>
+          <InternetArchivePlayer {...defaultProps} ref={playerRef} />
+        </GlobalAppContext.Provider>
+      );
+
+      act(() => {
```
MonkeyDo: You always need to await act calls. Not sure whether they are needed here and the three other instances where they are not awaited?

EDIT: tried removing those 4 non-awaited act calls and no changes, no warnings, so they are indeed not necessary.

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+
+      const audioElement = screen.getByTestId("internet-archive-player").querySelector("audio");
+      
+      // Simulate the timeupdate event
+      act(() => {
+        fireEvent.timeUpdate(audioElement!);
+      });
+
+      // Should call onProgressChange with current time in milliseconds
+      expect(defaultProps.onProgressChange).toHaveBeenCalled();
```
MonkeyDo: Why not test toHaveBeenCalledWith, considering that is what you describe in the comment just above it?

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+          track_name: "Test Track",
+        },
+      };
+
+      await act(async () => {
+        playerRef.current?.playListen(listen);
+      });
+
+      await waitFor(() => {
+        const artwork = screen.getByAltText("Test Track");
```
MonkeyDo: I think this would be clearer:

```suggestion
        const artwork = screen.getByRole("img");
```

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+          track_name: "Test Track",
+        },
+      };
+
+      await act(async () => {
+        playerRef.current?.playListen(listen);
+      });
+
+      await waitFor(() => {
+        expect(screen.queryByAltText("Test Track")).not.toBeInTheDocument();
```
MonkeyDo: Same change as above:

```suggestion
        expect(screen.queryByRole("img")).not.toBeInTheDocument();
```

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+          "Los Norteños / Cuando Canta La Lluvia",
+          "https://archive.org/details/00TtuloInttrprete66",
+          "Pérez Prado y Orquesta con Hermanas Montoya",
+          "RCA Victor #70-9428",
+          [{ src: "https://archive.org/download/00TtuloInttrprete66/Cuando Canta La Lluvia.png" }]
+        );
+      });
+
+      expect(defaultProps.onPlayerPausedChange).toHaveBeenCalledWith(false);
+    });
```
MonkeyDo, line 264: This  is the one perfect occasion to use the otherwise unused mockPlay you set up at the beginning.
It makes me question why you set it up in the first place if you were not planning on using it... Was it AI work?

You can also check the src attribute is set correctly:

```suggestion
      expect(mockPlay).toHaveBeenCalled();
      expect(audioElement).toHaveAttribute(
        "src",
        "https://archive.org/download/00TtuloInttrprete66/Cuando Canta La Lluvia.m4a"
      );
    });
```

Probably also a good idea to check that everything else that happens in that playCurrentTrack function happens correctly: calling onTrackInfoChange and onDurationChange

frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -0,0 +1,564 @@
...
+
+Object.defineProperty(window.HTMLMediaElement.prototype, 'play', {
+  writable: true,
+  value: mockPlay,
+});
+
+Object.defineProperty(window.HTMLMediaElement.prototype, 'pause', {
+  writable: true,
+  value: mockPause,
+});
```
MonkeyDo, line 23: I do have to say, the fact that you set those mocks to then not use them in your tests is a lost opportunity.

What was the reason for setting up those mocks in the first place?

RayyanSeliya: you are totally right  @MonkeyDo  , I was getting those annoying JSDOM errors initially about `play` not being implemented , so i set up the mocks to fix that , and then i got distracted in testing all the other functionality and forgot to actually verify the `play` was happening . its my mistake coz new to testing though ! srry about that ! , now i have added the `expect(mockPlay).toHaveBeenCalled()` assertion now along with checking the `src` attribute . good catch btw ! and also today i have the presentation for the final evaluation of this project to the  C4GT organisation  so was a bit hurry caused this mistake !

MonkeyDo (approved):
Front-end tests look good!

PR #3317, opened 2025-07-04
feat. Create the frontend for the listens importer

Suvid-Singhal:
This PR creates the frontend for the listens importer

It is still work in progress

Created the barebones UI as of now

### DIFF
--- a/frontend/js/src/settings/import/ImportListens.tsx
+++ b/frontend/js/src/settings/import/ImportListens.tsx
@@ -10,6 +10,8 @@ type ImportListensLoaderData = {
   user_has_email: boolean;
 };
 
+function handleListensSubmit() {}
+
 export default function ImportListens() {
   const data = useLoaderData() as ImportListensLoaderData;
   const { user_has_email: userHasEmail } = data;
@@ -73,14 +75,76 @@ export default function ImportListens() {
         avoid duplicates, be sure to set the appropriate limit date and time.
       </p>
 
-      <h3>
-        Coming soon
-        <FontAwesomeIcon icon={faPersonDigging} size="sm" className="ms-2" />
-      </h3>
+      <h3 className="card-title">Import from Listening History Files</h3>
+      <br />
       <p>
-        We are currently working on this feature as a matter of high priority,
-        please stay tuned.
+        Migrate your listens from different streaming services to Listenbrainz!
       </p>
+      <div className="card">
+        <div className="card-body">
+          <form onSubmit={handleListensSubmit}>
+            <div
+              className="flex flex-wrap justify-content-center"
+              style={{ gap: "1em" }}
+            >
+              <div>
+                <label className="form-label" htmlFor="datetime">
+                  Choose a File:
+                </label>
+                <input
+                  type="file"
+                  className="form-control"
+                  name="filename"
+                  required
+                />
+              </div>
+
+              <div>
+                <label className="form-label" htmlFor="datetime">
+                  Select Service:
+                </label>
+                <select className="form-select w-100">
+                  <option value="spotify">Spotify</option>
+                  <option value="listenbrainz">Listenbrainz</option>
+                  <option value="applemusic">Apple Music</option>
+                </select>
+              </div>
+
+              <div>
+                <label className="form-label" htmlFor="start-datetime">
+                  Start import from (optional):
+                </label>
+                <input
+                  type="date"
+                  className="form-control"
+                  max={new Date().toISOString()}
+                  name="ImportStartDatetime"
+                  title="Date and time to start import at"
+                />
+              </div>
+
+              <div>
+                <label className="form-label" htmlFor="end-datetime">
+                  End date for import (optional):
+                </label>
+                <input
+                  type="date"
+                  className="form-control"
+                  max={new Date().toISOString()}
+                  name="ImportStartDatetime"
+                  title="Date and time to end import at"
+                />
+              </div>
+
+              <div style={{ flex: 0, alignSelf: "end" }}>
+                <button type="submit" className="btn btn-success">
+                  Import Listens
+                </button>
+              </div>
+            </div>
+          </form>
+        </div>
+      </div>
     </>
   );
 }
--- a/admin/sql/create_foreign_keys.sql
+++ b/admin/sql/create_foreign_keys.sql
@@ -151,4 +151,10 @@ ALTER TABLE user_data_export
     REFERENCES "user" (id)
     ON DELETE CASCADE;
 
+ALTER TABLE user_data_import
+    ADD CONSTRAINT user_data_import_user_id_foreign_key
+    FOREIGN KEY (user_id)
+    REFERENCES "user" (id)
+    ON DELETE CASCADE;
+
 COMMIT;
--- a/admin/sql/create_indexes.sql
+++ b/admin/sql/create_indexes.sql
@@ -64,6 +64,6 @@ CREATE UNIQUE INDEX user_data_export_deduplicate_waiting_idx ON user_data_export
 
 CREATE INDEX user_data_import_user_id_idx ON user_data_import (user_id);
 
-CREATE UNIQUE INDEX user_data_import_deduplicate_waiting_idx ON user_data_import (user_id, type) WHERE status = 'waiting' OR status = 'in_progress';
+CREATE UNIQUE INDEX user_data_import_deduplicate_waiting_idx ON user_data_import (user_id, service) WHERE status = 'waiting' OR status = 'in_progress';
 
 COMMIT;
--- a/admin/sql/create_indexes.sql
+++ b/admin/sql/create_indexes.sql
@@ -62,4 +62,8 @@ CREATE INDEX user_data_export_user_id_idx ON user_data_export (user_id);
 
 CREATE UNIQUE INDEX user_data_export_deduplicate_waiting_idx ON user_data_export (user_id, type) WHERE status = 'waiting' OR status = 'in_progress';
 
+CREATE INDEX user_data_import_user_id_idx ON user_data_import (user_id);
+
+CREATE UNIQUE INDEX user_data_import_deduplicate_waiting_idx ON user_data_import (user_id, type) WHERE status = 'waiting' OR status = 'in_progress';
+
 COMMIT;
--- a/admin/sql/create_primary_keys.sql
+++ b/admin/sql/create_primary_keys.sql
@@ -34,4 +34,6 @@ ALTER TABLE background_tasks ADD CONSTRAINT background_tasks_id_pkey PRIMARY KEY
 
 ALTER TABLE user_data_export ADD CONSTRAINT user_data_export_id_pkey PRIMARY KEY (id);
 
+ALTER TABLE user_data_import ADD CONSTRAINT user_data_import_id_pkey PRIMARY KEY (id);
+
 COMMIT;
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -248,9 +248,8 @@ CREATE TABLE user_data_import (
     user_id             INTEGER NOT NULL,
     service             user_data_import_service_type NOT NULL,
     metadata            JSONB,
-    uploaded_filename   TEXT,
-    file_path           TEXT,
-    from_date           TIMESTAMPTZ NOT NULL DEFAULT '1900-01-01 00:00:00+00',
+    file_path           TEXT NOT NULL,
+    from_date           TIMESTAMPTZ NOT NULL DEFAULT 'epoch',
     to_date             TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     created             TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -243,6 +243,17 @@ CREATE TABLE user_data_export (
     created             TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
 
+CREATE TABLE user_data_import (
+    id                  INTEGER GENERATED ALWAYS AS IDENTITY,
+    user_id             INTEGER NOT NULL,
+    service             user_data_import_service_type NOT NULL,
+    status              user_data_import_status_type NOT NULL,
+    metadata            JSONB,
+    uploaded_filename   TEXT,
+    file_path           TEXT,
+    created             TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
 -- The following line is now executed by the init-db action from manage.py. If you create a DB without the init-db function
 -- you will need to execute the following GRANT in order to complete your DB setup.
 --GRANT ALL PRIVILEGES ON ALL TABLES IN SCHEMA public TO listenbrainz;
--- a/admin/sql/create_tables.sql
+++ b/admin/sql/create_tables.sql
@@ -247,10 +247,11 @@ CREATE TABLE user_data_import (
     id                  INTEGER GENERATED ALWAYS AS IDENTITY,
     user_id             INTEGER NOT NULL,
     service             user_data_import_service_type NOT NULL,
-    status              user_data_import_status_type NOT NULL,
     metadata            JSONB,
     uploaded_filename   TEXT,
     file_path           TEXT,
+    from_date           TIMESTAMPTZ NOT NULL DEFAULT '1900-01-01 00:00:00+00',
+    to_date             TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     created             TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
 
--- a/admin/sql/create_types.sql
+++ b/admin/sql/create_types.sql
@@ -19,10 +19,14 @@ CREATE TYPE user_stats_type AS ENUM('artists', 'releases', 'recordings', 'daily_
 
 CREATE TYPE do_not_recommend_entity_type AS ENUM ('artist', 'release', 'release_group', 'recording');
 
-CREATE TYPE background_tasks_type AS ENUM ('delete_listens', 'delete_user', 'export_all_user_data');
+CREATE TYPE background_tasks_type AS ENUM ('delete_listens', 'delete_user', 'export_all_user_data', 'import_listens');
 
 CREATE TYPE user_data_export_status_type AS ENUM ('in_progress', 'waiting', 'completed', 'failed');
 
+CREATE TYPE user_data_import_status_type AS ENUM ('in_progress', 'waiting', 'completed', 'failed');
+
 CREATE TYPE user_data_export_type_type AS ENUM ('export_all_user_data');
+
+CREATE TYPE user_data_import_service_type AS ENUM ('spotify', 'applemusic', 'listenbrainz');
 
 CREATE TYPE data_dump_type_type AS ENUM ('incremental', 'full');
--- a/admin/sql/create_types.sql
+++ b/admin/sql/create_types.sql
@@ -23,8 +23,6 @@ CREATE TYPE background_tasks_type AS ENUM ('delete_listens', 'delete_user', 'exp
 
 CREATE TYPE user_data_export_status_type AS ENUM ('in_progress', 'waiting', 'completed', 'failed');
 
-CREATE TYPE user_data_import_status_type AS ENUM ('in_progress', 'waiting', 'completed', 'failed');
-
 CREATE TYPE user_data_export_type_type AS ENUM ('export_all_user_data');
 
 CREATE TYPE user_data_import_service_type AS ENUM ('spotify', 'applemusic', 'listenbrainz');
--- a/admin/sql/updates/2025-06-25-add-user-data-import.sql
+++ b/admin/sql/updates/2025-06-25-add-user-data-import.sql
@@ -1,6 +1,5 @@
 ALTER TYPE background_tasks_type ADD VALUE 'import_listens';
 
-CREATE TYPE user_data_import_status_type AS ENUM ('in_progress', 'waiting', 'completed', 'failed');
 CREATE TYPE user_data_import_service_type AS ENUM ('spotify', 'applemusic', 'listenbrainz');
 
 CREATE TABLE user_data_import (
@@ -8,9 +7,8 @@ CREATE TABLE user_data_import (
     user_id             INTEGER NOT NULL,
     service             user_data_import_service_type NOT NULL,
     metadata            JSONB,
-    uploaded_filename   TEXT,
-    file_path           TEXT,
-    from_date           TIMESTAMPTZ NOT NULL DEFAULT '1900-01-01 00:00:00+00',
+    file_path           TEXT NOT NULL,
+    from_date           TIMESTAMPTZ NOT NULL DEFAULT 'epoch',
     to_date             TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     created             TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
@@ -24,6 +22,6 @@ ALTER TABLE user_data_import
     ON DELETE CASCADE;
 
 CREATE INDEX user_data_import_user_id_idx ON user_data_import (user_id);
-CREATE UNIQUE INDEX user_data_import_deduplicate_waiting_idx ON user_data_import (user_id, service) WHERE status = 'waiting' OR status = 'in_progress';
+CREATE UNIQUE INDEX user_data_import_deduplicate_waiting_idx ON user_data_import (user_id, service) WHERE metadata->>'status' IN ('waiting', 'in_progress');
 
 COMMIT;
--- a/admin/sql/updates/2025-06-25-add-user-data-import.sql
+++ b/admin/sql/updates/2025-06-25-add-user-data-import.sql
@@ -3,10 +3,6 @@ ALTER TYPE background_tasks_type ADD VALUE 'import_listens';
 CREATE TYPE user_data_import_status_type AS ENUM ('in_progress', 'waiting', 'completed', 'failed');
 CREATE TYPE user_data_import_service_type AS ENUM ('spotify', 'applemusic', 'listenbrainz');
 
-DROP INDEX background_tasks_user_id_task_type_idx;
-CREATE UNIQUE INDEX background_tasks_user_id_task_type_uniq_idx ON background_tasks (user_id, task);
-
-
 CREATE TABLE user_data_import (
     id                  INTEGER GENERATED ALWAYS AS IDENTITY,
     user_id             INTEGER NOT NULL,
@@ -27,6 +23,6 @@ ALTER TABLE user_data_import
     ON DELETE CASCADE;
 
 CREATE INDEX user_data_import_user_id_idx ON user_data_import (user_id);
-CREATE UNIQUE INDEX user_data_import_deduplicate_waiting_idx ON user_data_import (user_id, type) WHERE status = 'waiting' OR status = 'in_progress';
+CREATE UNIQUE INDEX user_data_import_deduplicate_waiting_idx ON user_data_import (user_id, service) WHERE status = 'waiting' OR status = 'in_progress';
 
 COMMIT;
--- a/admin/sql/updates/2025-06-25-add-user-data-import.sql
+++ b/admin/sql/updates/2025-06-25-add-user-data-import.sql
@@ -1,0 +1,32 @@
+ALTER TYPE background_tasks_type ADD VALUE 'import_listens';
+
+CREATE TYPE user_data_import_status_type AS ENUM ('in_progress', 'waiting', 'completed', 'failed');
+CREATE TYPE user_data_import_service_type AS ENUM ('spotify', 'applemusic', 'listenbrainz');
+
+DROP INDEX background_tasks_user_id_task_type_idx;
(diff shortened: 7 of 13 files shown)


MonkeyDo:
Overall the UI looks OK to me.
I think it could probably be improved, but it's hard to know how without adding more importers.
One thing I had in mind is hiding the optional inputs behind an accordion of sorts, but I don't think it's crucial in this case.

For the mobile UI, ideally we would have two fields on each line and the button underneath, I think that would look nicer at those screen sizes.


I've also mentioned that there will eventually be an ImportStatus component in there. Might be worth doing a mockup to see how it looks.
Same thing with the list of past imports that will be there.

Suvid-Singhal:
Hi @MonkeyDo,
I have tried to integrate the submitting listens to backend part here in the frontend but I'm facing some issues here. It would be really great if you could help me out here whenever you're free 😅
WillI will also try to integrate the import status after posting to backend part is sorted

Thanks

frontend/js/src/utils/APIService.ts
```diff
@@ -2072,7 +2072,7 @@ export default class APIService {
   }
 
   importListens = async (userToken: string, formData: FormData) => {
-    const url = `${this.APIBaseURI}/import`;
+    const url = `http://localhost:8100/import/`;
```
MonkeyDo: This looks like maybe it should not have been committed...

Suvid-Singhal: Yea I think I should make a variable for this and change it in testing maybe?
Maybe `LBBaseURL`?
What do you think?

MonkeyDo: It already exists in the config file and is already passed to the front-end.
If you need to hit your local server instead of the production api, you need set it there:

https://github.com/metabrainz/listenbrainz-server/blob/a6f6f8e092c5b7ee7ce496db901d7ec577acba7e/listenbrainz/config.py.sample#L141-L144

MonkeyDo: Ah, but is this because you are hitting an endpoint that is not an API endpoint (i.e. without the leading `/1/` prefix) ?
If so then I don't think this is the correct place to put this function, considering this class is specifically about API endpoints.
Was there a reason for making this import endpoint not as an API endpoint? That looks like the mistake to me, but maybe I am missing previous discussions and decisions.
If the solution is to make the endpoint an API endpoint, change the init.py line to `app.register_blueprint(import_bp, url_prefix=API_PREFIX+'/import')` (only added `API_PREFIX+`)

If the endpoint stays as-is, I would remove this wrapper function for APIService and handle everything directly in ImportListens.tsx

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -3,21 +3,329 @@ import * as React from "react";
...
+          <dt className="col-4">Progress</dt>
+          <dd className="col-8">{im.metadata.progress}</dd>
+          <dt className="col-4">Requested on</dt>
+          <dd className="col-8">{format(im.created, "PPp")}</dd>
+          <dt className="col-4">Import #</dt>
+          <dd className="col-8">{im.import_id}</dd>
+          <dt className="col-4">File name</dt>
+          <dd className="col-8">{im.metadata.filename}</dd>
+          <dt className="col-4">Start date</dt>
+          <dd className="col-8">{format(im.from_date, "PPP")}</dd>
```
MonkeyDo: I suppose you didn't test this without a date entered, as it breaks the page completely right after submitting a file with no dates.
After refreshing the page, a date does exist (0 // unix epoch) and renders fine.

Image

Please be more thorough in testing your features

MonkeyDo: If the from_date is equal to zero, it might be better to show a dash "-" or slash "/" instead of "January 1st, 1970"

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -3,21 +3,329 @@ import * as React from "react";
...
+      <details>
+        <summary>
+          <FontAwesomeIcon
+            icon={faChevronCircleRight}
+            size="sm"
+            className="summary-indicator"
+          />
+          Details
+        </summary>
+        <dl className="row">
```
MonkeyDo: The service is missing in the details shown here, and would be useful to differentiate imports.

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -3,21 +3,329 @@ import * as React from "react";
...
+
+const API_PREFIX = "/1";
+
+function renderImport(
+  im: Import,
+  cancelImport: (event: React.SyntheticEvent, importToCancelId: number) => void,
+  fetchImport: (importId: number) => Promise<any>
+) {
+  const extraInfo = (
+    <p>
```
MonkeyDo: This is causing a validation error in the console as it is not valid HTML apparently:

> `<summary> cannot appear as a descendant of <p>`
> `<details> cannot appear as a descendant of <p>`

```suggestion
    <div>
```

frontend/js/src/settings/import/ImportListens.tsx
Suvid-Singhal: Yes, I missed testing that. My bad :(
Implemented the suggestions now!

MonkeyDo:
To fix the tests, you'll need a couple of things:

1. Fix an unrelated issue with our setup with JSDom and msw (`TypeError: RequestInit: Expected signal ("AbortSignal {}") to be an instance of AbortSignal.`)
```
import { enableFetchMocks } from 'jest-fetch-mock'
[...]
// fix for issue with JSDom fetch and msw
enableFetchMocks();
```

2. Fix the server/API mock paths and add missing server mock, + create the memoryRouter after setting up msw server:
```
beforeAll(async () => {
    server = setupServer(
      http.post("/settings/import/", () => {
        return HttpResponse.json({ "user_has_email":true });
      }),
      http.get("/1/import-listens/list/", () => {
        return HttpResponse.json(mockImports);
      }),
      http.post("/1/import-listens/cancel/1/", ({ params }) => {
        return HttpResponse.json({ success: true, id: params.id });
      }),
      http.post("/1/import-listens/1/", ({ params }) => {
        return HttpResponse.json(mockImports[0]);
      })
    );
    server.listen();
    // Create the router *after* MSW mock server is set up
    // See https://github.com/mswjs/msw/issues/1653#issuecomment-1781867559
    router = createMemoryRouter(routes, {
      initialEntries: ["/settings/import/"],
    });
  });
```
3. Fix rendering setup:
The `false` at the end instructs renderWithProviders that we are providing our own router.
You'll be able to reuse the router you created in beforeAll.
```
renderWithProviders(
      <RouterProvider router={router} />,
      {},
      {
        wrapper: ReactQueryWrapper,
      },
      false
    );
```


With those changes, tests are running as expected, although you still have some work to do to make them pass

MonkeyDo:
I will also note that another way we test these things usually is to create utility functions in the APIService component (i.e. instead of calling the API in ImportListens). This allows us to provision a custom APIService when testing, where we can mock the functions we need. That means we don't actually need to mock the server response since it is never actually called.
Example of that here:
https://github.com/metabrainz/listenbrainz-server/blob/fc3fb5761ad24a0841a3abe7b58b0755f1c776d2/frontend/js/tests/user/stats/UserEntityChart.test.tsx#L72-L81
https://github.com/metabrainz/listenbrainz-server/blob/fc3fb5761ad24a0841a3abe7b58b0755f1c776d2/frontend/js/tests/user/stats/UserEntityChart.test.tsx#L165
https://github.com/metabrainz/listenbrainz-server/blob/fc3fb5761ad24a0841a3abe7b58b0755f1c776d2/frontend/js/tests/user/stats/UserEntityChart.test.tsx#L179

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -69,13 +69,23 @@ function renderImport(
...
           <dt className="col-4">File name</dt>
           <dd className="col-8">{im.metadata.filename}</dd>
+          <dt className="col-4">Service</dt>
+          <dd className="col-8">
+            {Services[(im.service as unknown) as keyof typeof Services]}
+          </dd>
           <dt className="col-4">Start date</dt>
-          <dd className="col-8">{format(im.from_date, "PPP")}</dd>
+          <dd className="col-8">
+            {im.from_date && new Date(im.from_date).getTime() !== 0
```
MonkeyDo: You could use the `isValid` utility function from date-fns, since we're already using the `format` function:

```suggestion
            {isValid(im.from_date)
```

Same for the other date below

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -118,6 +128,7 @@ function renderImport(
         <button
           type="button"
           className="btn btn-sm btn-transparent"
+          aria-label="Refresh"
```
MonkeyDo, line 130: 👏

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+      
+    });
+    await waitFor(() => {
+      expect(screen.getByText(/choose a File/i)).toBeInTheDocument();
+      
+    });
+    await waitFor(() => {
+      expect(screen.getByRole("button", { name: /import listens/i })).toBeInTheDocument();
+      
+    });
```
MonkeyDo: You should be able to put all these assertions inside the same waitFor block rather than each in a separate waitFor.
Same goes for the other tests.

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -78,19 +78,44 @@ describe("ImportListensPage", () => {
...
+      expect(screen.getByText(/select Service/i)).toBeInTheDocument();
+      
+    });
+    await waitFor(() => {
+      expect(screen.getByText(/choose a File/i)).toBeInTheDocument();
       
     });
     await waitFor(() => {
-      screen.debug(undefined, Infinity);
+      expect(screen.getByRole("button", { name: /import listens/i })).toBeInTheDocument();
```
MonkeyDo: Here we can also check that the button is disabled at first:
```
      const importButton = screen.getByRole("button", {
        name: /import listens/i,
      });
      expect(importButton).toBeInTheDocument();
      expect(importButton).toBeDisabled();
```

This then allows you to add another test to check that when you fill up the required inputs, the button becomes enabled (`expect(importButton).not.toBeDisabled();`).
To mock user interactions you should use user-event: https://testing-library.com/docs/user-event/intro
See an example of that here: https://github.com/metabrainz/listenbrainz-server/blob/fc396e075e478af7bb79f02ca3d8a00884a21dcc/frontend/js/tests/cb-review/CBReviewModal.test.tsx#L137-L148
For file upload specifically: https://testing-library.com/docs/user-event/utility#upload

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+      {},
+      { wrapper: ReactQueryWrapper },
+      false
+    );
+
+    await waitFor(() => {
+      expect(screen.getByText("spotify.zip")).toBeInTheDocument();
+    });
+    const cancelBtn = screen.getByRole("button", { name: "Cancel import" });
+    await user.click(cancelBtn);
```
MonkeyDo, line 224: The setup is good, but there are no test assertions !
You should ensure that the server got hit with an API call to `POST /1/import-listens/cancel/1` , which if you look at the console when running the tests is not currently defined in the mock server.

The best way to do this with MSW is to have the mock server return a success response or a failure response, then check that the front-end shows the toast message you expect for that scenario.
That way you are not testing the internals of the component but rather the user interaction: locate button, click button, see success message.

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+      {},
+      { wrapper: ReactQueryWrapper },
+      false
+    );
+
+    await waitFor(() => {
+      expect(screen.getByText("spotify.zip")).toBeInTheDocument();
+    });
+    const refreshBtn = screen.getByRole("button", { name: "Refresh" });
+    await user.click(refreshBtn);
```
MonkeyDo: Same as above, there is no assertion here, it's a phantom test.
I think both of these last tests should have both a failure and a success tests (so 2 tests for each) that test the toast message shown on screen.

You can override the msw handlers with `server.use(...)` at the beginning of the test: https://mswjs.io/docs/best-practices/network-behavior-overrides

MonkeyDo (commented):
Nice work on getting those tests set up!
I think you could go a bit further with the testing but it's definitely on the right tracks.

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  it("enables the import button after a file is uploaded", async () => {
+    renderWithProviders(<RouterProvider router={router} />, {}, { wrapper: ReactQueryWrapper }, false);
+    const importButton = screen.getByRole("button", { name: /import listens/i });
+    expect(importButton).toBeDisabled();
+
+    const file = new File(['{ "foo": "bar" }'], 'test.json', { type: 'application/json' });
+    const input = screen.getByLabelText(/choose a file/i);
+
+    fireEvent.change(input, { target: { files: [file] } });
```
MonkeyDo, line 127:
```suggestion
    await user.upload(input, file);
```

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    renderWithProviders(<RouterProvider router={router} />, {}, { wrapper: ReactQueryWrapper }, false);
+    await waitFor(() => {
+      expect(screen.getByText("spotify.zip")).toBeInTheDocument();
+      expect(screen.getByText("August 11th, 2025")).toBeInTheDocument();
+      expect(screen.getByText("-")).toBeInTheDocument();
+    });
+  });
+
+it("calls import endpoint when import listens clicked", async () => {
+  let submitAPICalled = false;
```
MonkeyDo: Should redefine this here as it is already defined in the scope above.

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  server.use(
+    http.post("/settings/import/", (req) => {
```
MonkeyDo: This is the main issue you were facing: the import submissions do not hit this endpoint.
You should be hitting the API endpoint instead.

It's good to have your component code open side-by-side with you test file so you can read the code and follow what should happen at what time.
It should help to avoid these mistakes.


```suggestion
    http.post("/1/import-listens/", (req) => {
```

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+      submitAPICalled = true;
+      return HttpResponse.json({ user_has_email: true });
```
MonkeyDo: Shouldn't return this content.

While testing locally I created a new mock import:
```ts
const newImport = {
  import_id: 3,
  service: "spotify",
  created: "2025-08-12T10:00:00Z",
  metadata: {
    filename: "spotify_test.zip",
    progress: "Your data import will start soon.",
    status: "waiting",
  },
  from_date: "1970-01-01T00:00:00Z",
  to_date: "2025-08-11T00:00:00Z",
};
```

Then use it like so:

```suggestion
      return HttpResponse.json(newImport);
```

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    false
+  );
+
+  const input = screen.getByLabelText(/choose a file/i);
+
+  const file = new File(["spotify_test"], "spotify_test.zip", { type: "application/zip" });
+  fireEvent.change(input, { target: { files: [file] } });
+
+  const importButtonAfter = await screen.findByRole("button", { name: /import listens/i });
+  await waitFor(() => expect(importButtonAfter).not.toBeDisabled());
```
MonkeyDo: I think you might run into issues here:
check how you defined `hasAnImportInProgress`  and the status set in the mockImports.
I think you will find that because one of the mock imports is not of status "completed", this button won't be enabled in any of the tests.

MonkeyDo: Shouldn't need the waitFor here because we `await` the user interaction above (await user.upload(... )

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    <RouterProvider router={router} />,
+    {},
+    { wrapper: ReactQueryWrapper },
+    false
+  );
+
+  const input = screen.getByLabelText(/choose a file/i);
+
+  const file = new File(["spotify_test"], "spotify_test.zip", { type: "application/zip" });
+  fireEvent.change(input, { target: { files: [file] } });
```
MonkeyDo:
```suggestion
  await user.upload(input, file);
```

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const importButtonAfter = await screen.findByRole("button", { name: /import listens/i });
+  await waitFor(() => expect(importButtonAfter).not.toBeDisabled());
+
+  fireEvent.click(importButtonAfter);
```
MonkeyDo: I ran into an odd issue with FormData. I hade to add this at the top of the page to mock it:
```
/* 
  Mocking FormData to resolve an error with msw/jest-fixed-jsdom
  See https://github.com/mswjs/jest-fixed-jsdom/issues/32
*/
Object.defineProperty(window, "FormData", {
  writable: true,
  value: jest.fn().mockImplementation(() => ({
    append: jest.fn(),
  })),
});
```

Somehow clicking the submit button did not actually submit the form, I think related to the issue above.
I opted for this method to get the submission working:

```suggestion
        // Using fireEvent.submit instead of userEvent.click
    // because the click event does not submit the form
    fireEvent.submit(submitButton);
```

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  await waitFor(() => expect(submitAPICalled).toBe(true));
```
MonkeyDo: Also test for the presence of the new import, after receiving a response from the API it is added to the page:

`expect(screen.getByText("spotify_test.zip")).toBeInTheDocument();`

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+    }));
+    renderWithProviders(<RouterProvider router={router} />, {}, { wrapper: ReactQueryWrapper }, false);
+
+    await waitFor(() => {
+      expect(screen.getByText("spotify.zip")).toBeInTheDocument();
+    });
+
+    const cancelBtn = screen.getByRole("button", { name: "Cancel import" });
+    await user.click(cancelBtn);
+    await waitFor(() => expect(cancelAPICalled).toBe(true));
```
MonkeyDo: Please also test that the filename for the cancelled import does not appear in the page anymore

`expect(screen.queryByText("spotify.zip")).not.toBeInTheDocument();`

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+  });
+
+  it("calls refresh endpoint when refresh clicked", async () => {
+    server.use(
+      http.get("/1/import-listens/1/", (req) => {
+        refreshAPICalled = true;
+        return HttpResponse.json(mockImports[0]);
```
MonkeyDo: Here, to properly test the refreshing functionality, you would return an updated object, for example with an updated filename (won't happen in practice, but fine to test the mechanism):

```suggestion
        return HttpResponse.json(
          merge(mockImports[0], {
            metadata: { filename: "spotify-updated.zip" },
          })
        );
```

Test this before clicking the refresh button:
```
await waitFor(() => {
      expect(screen.getByText("spotify.zip")).toBeInTheDocument();
    });
    expect(screen.queryByText("spotify-updated.zip")).not.toBeInTheDocument();
```

And then after clicking the refresh button and checking refreshAPICalled:
```
    expect(screen.getByText("spotify-updated.zip")).toBeInTheDocument();
    expect(screen.queryByText("spotify.zip")).not.toBeInTheDocument();

```

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+        return HttpResponse.json({ success: true, id: 1 });
+      }),
+      http.get("/1/import-listens/1/", (req) => {
+        return HttpResponse.json(mockImports[0]);
+      })
+    );
+
+    server.events.on("request:start", ({ request }) => {
+      console.log("MSW captured:", request.method, request.url);
+    });
```
MonkeyDo: When you are done and all tests pass, please remove the logging.

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -148,6 +160,7 @@ export default function ImportListens() {
 
   const [loading, setLoading] = React.useState(false);
   const [imports, setImports] = React.useState<Array<Import>>([]);
+  const [fileSelected, setFileSelected] = React.useState<File | null>(null);
```
MonkeyDo: I am doubting why you need to store the entire file in memory instead of a simple boolean, if this is only used for disabling the submit button.

MonkeyDo (changes requested):
I had to struggle on the tests too.

Some of it was pretty simple mistakes, but some of it was weird bugs with the interraction between the testing framework and the server mocking library.

frontend/js/src/settings/import/ImportListens.tsx
MonkeyDo, line 1: The unused `formDate.get` lines in createImport should be removed

 https://github.com/metabrainz/listenbrainz-server/pull/3317/files#diff-0b2493481e063c2fe80c27fc4faa0ea2d16746c35835693f2e522f5e970ea74eR256-R259

MonkeyDo, line 1: On line 254 you can use `const form = event.currentTarget;` and drop the forced type `as HTMLFormElement`

frontend/js/tests/user/import-listens/ImportListens.test.tsx
Suvid-Singhal: I tried doing this earlier as well but it just doesn't work or enable the button. `fireEvent` seems to do the trick and all the tests pass so I just kept it.

frontend/js/tests/user/import-listens/ImportListens.test.tsx
Suvid-Singhal: Thanks a lot for this!
I didn't even think of this (and still kinda don't know what this does properly 😭)

frontend/js/tests/user/import-listens/ImportListens.test.tsx
MonkeyDo: Sorry, typo!

It should be "You **shouldn't** redefine this here as it is already defined in the scope above."

frontend/js/tests/user/import-listens/ImportListens.test.tsx
MonkeyDo: I dug into it a little bit more, the reason the test was failing when you used user.upload is because the button is not in fact disabled.
When writing tests like these, always assume first that if the test is failing, your code is not doing what you expect it to.

I don't know why the test assertion passes when you use fireevent, but I would stay away from that from here onwards. As I pointed out in another comment, your logic for `hasAnImportInProgress` combined with your mock imports results in the button never being enabled in the tests.
So the tests should be failing!

This is because `hasAnImportInProgress` checks if any of the imports is not "complete", rather than checking if one is in progress.
The first mock import has status "waiting", so the button will not be enabled.


So a couple of things:
- use the more standard `await user.upload(...` as suggested above
- review the `hasAnImportInProgress` logic to make sure you understand what it does
- if the `hasAnImportInProgress` logic is sound for the feature, change the mock import's status

Suvid-Singhal: Yes, I had tried changing the import status of all the imports to completed previously in server.use(), but it was also altering with other tests below it. So should I move this test to the end?

I had tried changing the /list/ endpoint by returning only completed imports

frontend/js/tests/user/import-listens/ImportListens.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+  const input = screen.getByLabelText(/choose a file/i);
+
+  const file = new File(["spotify_test"], "spotify_test.zip", { type: "application/zip" });
+  fireEvent.change(input, { target: { files: [file] } });
+
+  const importButtonAfter = await screen.findByRole("button", { name: /import listens/i });
+  await waitFor(() => expect(importButtonAfter).not.toBeDisabled());
+
+  fireEvent.submit(importButtonAfter);
```
MonkeyDo, line 192: After some more digging, I finally found what this issue is and why we could not use `await user.click(importButtonAfter);` to trigger the import successfully.

There is an issue with the user-event library when it comes to file inputs with the `required` attribute.
See https://github.com/testing-library/user-event/issues/1293 and https://github.com/testing-library/user-event/issues/1133

Using instead `fireEvent.submit` to submit the form circumvents this user-event issue and forces the submission event.
Not sure I like that, as it could obfuscate a real issue in the component, but in the meantime we don't have a lot of options.

Might be a good idea to leave a comment explaining why we use this, with a link to the relevant issues above.

Suvid-Singhal: Yes I was also skeptical that why are we "submitting" the form and not clicking the button and simulating user action to submit it. But anyways, it seems to be the best alternative for now...

frontend/js/tests/user/import-listens/ImportListens.test.tsx
Suvid-Singhal: Thanks for the prompt reply on chat! I was able to solve the issues :)

Bqleine:
Hi, just chiming in as an external looker interested in the feature!

Since the UI is already out in production, I’ll give my thoughts and you can do what you want with them.

- It’s not clear to me what file I should upload, I got a zip with many files from Spotify. Do I have to add them one-by-one or can I add the whole zip? Or ideally could I select multiple files since I don’t have the zip anymore.
- Will the “Your Year in Music” playlists get updated with my new listens?

In any case, thanks a lot for working on this! Its going to be awesome seeing my stats with my listens of all time.

P.S. The import is throwing 410 Gone errors on listenbrainz.org, I assume it wasn’t meant to be there since the pull request is not merged yet?

Suvid-Singhal:
Hi @Bqleine,

- You should upload the entire Spotify listening history zip file that you got from Spotify. No need to extract it or separate the .json files. The importer will handle it all.
- I am not sure about this. Perhaps @amCap1712 can help you in this.

The feature is not rolled out yet but stay tuned for the next release where it would be deployed successfully :)
It's so good to see that people are interested in this feature and would be using it ❣️

MonkeyDo:
Regarding Year in Music, I'm afraid we have not currently planned on re-calculating the data for those pages.

It might eventually possible as we are slowly putting in place an on-demand type of system to request statistics calculations for older data.

Suvid-Singhal:
Hi @Bqleine,
Thank you for raising the concerns regarding what type of files are accepted by the importer. We recognised that it might be an issue for many other people as well. So we have added an info box to clarify the same! 😃

MonkeyDo:
@Suvid-Singhal I fixed a major issue with the overall mechanism: instead of calling the API endpoints using the APIBaseURI provided by the APIService, we were hitting `/1/.....` which while it works fine on local development, will completely break once in production where the API endpoints are on another domain
So instead of hitting `api.listenbrainz.org/1/import-listens/...` they would hit `listenbrainz.org/1/import-listens/...` which don't exist and would result in errors (not found at best, method not allowed otherwise).

I also took the liberty of moving the service select as the first step in the form, which made more sense to me from a user's perspective: first select the service you want to import from, and only then select the file to import. Small change but I thought it made sense.
Might also make sense in the future if we want to dynamically change the allowed file types depending on the selected service.

Bqleine:
Hi, just tried it out. It worked great!

Thanks a lot to you and everyone involved for the awesome feature.

MonkeyDo:
>  Might also make sense in the future if we want to dynamically change the allowed file types depending on the selected service.

@Suvid-Singhal I thought you might be interested in this, the use-case came up much sooner than I expected with the addition of the Libre.fm importer which accepts .csv files.

This is for reference only, I thought you might be interested in seeing how I implemented this.
I very much enjoyed it myself, always fun to play with typescript enums and end up with an elegant solution:
https://github.com/metabrainz/listenbrainz-server/pull/3368/files/ef696ae903d134afb6750c168d28853c0efdabc1..55fa8ede551c1484c3553086f43d939f2f7ec2b8

Suvid-Singhal:
Wow that really does seem to be an elegant solution. It's always fun to learn new things from you @MonkeyDo 😀

PR #3318, opened 2025-07-11
Refactor remaining Enzyme tests to React Testing Library

MonkeyDo:
Continuing the work from #2552 , #2577 and #2638 and FINALLY getting rid of Enzyme for good!
Just a month shy of the second anniversary of the start of this migration...

I recommend reviewing commit by commit for mental sanity.


frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx
```diff
@@ -579,6 +579,7 @@ export default class SpotifyPlayer
         durationMs: duration,
         currentSpotifyTrack: current_track ?? undefined,
       });
+      onDurationChange(duration);
```
MonkeyDo, line 582: Not sure this was supposed to be commited or if mistake, will check

MonkeyDo: Yes, this is expected, and was missing from the implementation, most likely causing a discrete bug.
We were setting the new durationMs in the component state but not passing that info back up to the parent BP UI because we return early in this condition.
The previous test suite tested internal state so didn't catch that mistake. To keep the test similar I needed to fix that bug.

anshg1214 (approved):

PR #3320, opened 2025-07-15
LB-1790: Render album link for album reviews in user feed

MonkeyDo:
Currently album reviews from CritiqueBrainz are not displayed properly because we shoehorned it into a component made for displaying recording info.
The issue points out the link was wrong and pointed to a recording page instead of album page.

Made a couple of modifications to the ListenCard to fallback to showing album info (and correct links) when there is no recording info.

Added a `getAlbumLink` utility to link to release group, or release if the rg is not available.

Also fixes a small CSS issue with review content in the feed.
Note that there is still a markdown rendering/parsing issue

Before:
image
With incorrect link to the non-existing "musicbrainz.org/recording/ad5bc068-a0a4-4c8e-b0cb-c69d806cd052"

After:
image
With correct link to "/album/ad5bc068-a0a4-4c8e-b0cb-c69d806cd052"

### DIFF
--- a/frontend/js/src/utils/utils.tsx
+++ b/frontend/js/src/utils/utils.tsx
@@ -325,7 +325,7 @@ const getAlbumLink = (listen: Listen): JSX.Element | string => {
     return <Link to={`/album/${releaseGroupMBID}`}>{releaseName}</Link>;
   }
   if (releaseMBID) {
-    return <Link to={`/release/${releaseGroupMBID}`}>{releaseName}</Link>;
+    return <Link to={`/release/${releaseMBID}`}>{releaseName}</Link>;
   }
   return releaseName;
 };


frontend/js/src/utils/utils.tsx
```diff
@@ -315,6 +316,20 @@ const getTrackLink = (listen: Listen): JSX.Element | string => {
...
+const getAlbumLink = (listen: Listen): JSX.Element | string => {
+  const releaseName = getReleaseName(listen);
+  const releaseGroupMBID = getReleaseGroupMBID(listen);
+  const releaseMBID = getReleaseMBID(listen);
+
+  if (releaseGroupMBID) {
+    return <Link to={`/album/${releaseGroupMBID}`}>{releaseName}</Link>;
+  }
+  if (releaseMBID) {
+    return <Link to={`/release/${releaseGroupMBID}`}>{releaseName}</Link>;
```
anshg1214: This should be releaseMBID right?
```suggestion
    return <Link to={`/release/${releaseMBID}`}>{releaseName}</Link>;
```

MonkeyDo: Yes, absolutely! Thanks for catching my bad copypasta

anshg1214 (approved):
LG!

PR #3322, opened 2025-07-18
fix: Errors on Entity page

anshg1214:
Before:
image

After:
image


MonkeyDo (approved):
Makes sense! 

EDIT: Although some server tests look like they need to be amended to add the `path` argument manually

PR #3327, opened 2025-07-22
feat: Migrate BrainzPlayer context to Jotai

anshg1214:


anshg1214:
This PR is currently deployed on https://test.listenbrainz.org/

MonkeyDo:
Found some bugs with the queue :fearful: 
I tried with playing music from spotify and from youtube, just in case it was related to one music service, but I get the same results for both.

1: Every time I tried, the first song I try to play does not play
2: open dashboard, open BP queue, click "move to queue" button, start playing a song, then click "clear queue" button (you need to be playing a song to have that button). Now clicking on the play icon on listencards on the page, the value of `listen` in `searchAndPlayTrack` is undefined so this code block is triggered: https://github.com/metabrainz/listenbrainz-server/blob/40dfa8484a5dfeccac57be8e81e1f3756987bb84/frontend/js/src/common/brainzplayer/SpotifyPlayer.tsx#L189-L196
3: not sure how to replicate, but sometimes when doing the above, after clearing the queue clicking on the play button on listencards adds the listen to the queue instead of playing it
4: when in the situation described in 3., despite there being some items in the queue, finishing a song does not start playing the next song, kind of as if there was no queue. and as 3. describes clickign play icon on cards does not play either.

anshg1214:
~@MonkeyDo I've fixed the first issue, but I'm not able to recreate the other three issues.~

EDIT: All the issues were resolved.

frontend/js/src/common/listens/ListenCard.tsx
```diff
@@ -103,8 +104,9 @@ export type ListenCardState = {
 
 type ListenCardPropsWithDispatch = ListenCardProps & {
```
MonkeyDo: Might make sense to rename this, now that we don't use anything named dispatch anymore?
```suggestion
type ListenCardPropsWithActions = ListenCardProps & {
```

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
MonkeyDo, line 1: Not sure exactly what the issue is, but the search fallback mechanism does not seem to work.
If not track is found, nothing happens (i.e. it does not try to search using the next datasource)
It does highlight the new track, suggesting that something is happening, but BrainzPlayer does not start playing the track.

For posterity, I was using this artist page to test: https://test.listenbrainz.org/artist/1ba797b6-b445-4c04-b6b3-458ba9730ad9/
Playing "Enter the mirror" works fine, but playing "The Night Collectors" does not. Looking into network requests, there is no spotify match for that song, and no other network request is made, nothing else happens.

Image

anshg1214: Fixed

MonkeyDo:
OK, I ended up refactoring a bunch of stuff in order to nuke this `show` prop we were using (as I suggested when we met in pereson), which let's face it was pretty hacky.
It was causing some issues with update lifecycle (not resolved by 996d8b5) and probably causing some unnecessary rerenders.

I opted to use a new atom to store the currently selected datasource's name, and then in each player compare that with the datasource's name to determine if it's currently selected. Happy to discuss other options if we need.

I did all this on a separate branch as I wasn't exactly sure where I was going and didn't want to dirty your branch.
https://github.com/metabrainz/listenbrainz-server/tree/refs/heads/monkey/jotai-brainzplayer-testing

Currently up on test for the time being

anshg1214:
I saw the changes, and they look good. So we can merge your branch with this branch.

MonkeyDo:
I took a moment to try out the mechanism I was talking about during the summit:

If I try to play a listen from then middle of the page,p clear the ambient queue above that listen so playback continues from the next one in the page rather than the top.
It's one commit in the branch `monkey/brainzplayer-clear-ambient-queue`

I did notice one unfortunate side-effect which might be a blocker for that idea until we implement something more complex:

Say I have 25 listens on the page as my ambient queue. I click play on listen # 10, the ambient queue now contains listens 11-25.
If I then click play on listen # 5, the ambient queue still only contains listens 11-25, with no sign of listens 6-10 anymore. 
Playback will continue from listen 11.

PR #3328, opened 2025-07-24
LB-1805: integrate @zumer/snapdom for image downloading and remove html2canvas

anshg1214:
LB-1805


anshg1214:
The way YIM handles image different is completely different, and that's not using html2canvas right now. We're using `file-saver` right now for those. I think we can refactor them too, but in a separate PR.

PR #3329, opened 2025-07-24
LB-1808: Fix scrollbar glitch

anshg1214:
LB-1808

### DIFF
--- a/frontend/js/src/tags/TagsComponent.tsx
+++ b/frontend/js/src/tags/TagsComponent.tsx
@@ -311,15 +311,11 @@ export default function AddTagSelect(props: {
             flexWrap: "nowrap",
             overflowX: "auto",
             paddingRight: "3.5em",
+            scrollbarWidth: "none",
+            "-ms-overflow-style": "none",
             "::-webkit-scrollbar": {
-              height: "5px",
-              backgroundColor: "#f5f5f5",
-            },
-            "::-webkit-scrollbar-track": {
-              backgroundColor: "#f5f5f5",
-            },
-            ":hover::-webkit-scrollbar-thumb": {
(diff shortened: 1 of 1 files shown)


MonkeyDo (commented):
I'm not sure I love the wholesale removal of the scrollbar.
On chrome-based browsers the scrollbar container does not seem overlap the content:

Image

Firefox does not at first, but when hovering it the height of the scrollbar extends:

Image
Hovering:
Image

So I think it might be a question of refining the styles of the scrollbar so that it works a bit better cross-browser. A starting point would be to replicate the issue from the ticket by trying on the same browser the user has.
It might be worth reviewing the scrollbar style we have in frontend/css/sass/scroll-container.scss to see if it improves the situation at all, but I don't have a lot of hope for that.


Another possibility would be to add arrows to the right and left for scrolling, like we do in a couple of places, which uses some javascript to automatically show or hide the arrows (see HorizontalScrollContainer component).

frontend/js/src/tags/TagsComponent.tsx
```diff
@@ -311,15 +311,11 @@ export default function AddTagSelect(props: {
             flexWrap: "nowrap",
             overflowX: "auto",
             paddingRight: "3.5em",
+            scrollbarWidth: "none",
```
MonkeyDo: I think the addition of `scrollbarWidth: "thin"` is all we would need to resolve the issue for Firefox (reverting the webkit-specific styles)

MonkeyDo: I confirmed this seems to resolve the issue
```suggestion
            scrollbarWidth: "thin",
```

PR #3330, opened 2025-07-25
LB-1809: add automatic LB /track URL recognition

MonkeyDo:
In the same way we automatically recognize MusicBrainz recording URLs in the "Add listens" and "Link listen" modals, adding the new listenbrainz.org/track/$MBID route.
Also renamed existing regexps to avoid confusion.

image


anshg1214 (approved):
LG!

PR #3331, opened 2025-07-26
Integrate music streaming from Navidrome 

mAmanullah7:
## This pr is a part of my GSoC'25 project [Integrate music streaming from Funkwhale & Navidrome](https://community.metabrainz.org/t/integrate-music-streaming-from-funkwhale-navidrome-subsonic/751757)

This is a continuation of #3289 
allow users to play music from their Navidrome servers  directly in BrainzPlayer

Will be implementing basic `username/password + salt` authentication

Using two tables similar to funkwhale
* `navidrome_servers` -> to store host_url
* `navidrome_tokens` -> to store user credentials

## Navidrome UI:
Screenshot 2025-08-24 at 12 41 49 AM
Screenshot 2025-08-24 at 12 42 06 AM

### DIFF
--- a/listenbrainz/domain/tests/test_navidrome.py
+++ b/listenbrainz/domain/tests/test_navidrome.py
@@ -63,7 +63,7 @@ class NavidromeServiceTestCase(IntegrationTestCase):
             service = NavidromeService()
             with self.assertRaises(ExternalServiceError) as context:
                 service.authenticate(
-                    "https://demo.navidrome.org/rest/ping",
+                    "https://demo.navidrome.org",
                     "testuser",
                     "wrongpassword"
                 )
@@ -164,3 +164,35 @@ class NavidromeServiceTestCase(IntegrationTestCase):
 
             service.remove_user(self.user["id"])
             self.assertIsNone(service.get_user(self.user["id"]))
+
+    @requests_mock.Mocker()
+    def test_url_normalization_trailing_slash(self, mock_requests):
+        """Test that URLs with trailing slashes are properly normalized"""
+        mock_requests.get(
+            "https://demo.navidrome.org/rest/ping",
+            status_code=200,
+            json={
+                "subsonic-response": {
+                    "status": "ok",
+                    "version": "1.16.1",
+                    "type": "navidrome",
+                    "serverVersion": "0.49.3"
+                }
+            }
+        )
+        with self.app.app_context():
+            service = NavidromeService()
+            
+            service.connect_user(
+                self.user["id"],
+                "https://demo.navidrome.org/",  # Note the trailing slash
+                "testuser",
+                "testpassword"
+            )
+            
+            result = service.get_user(self.user["id"])
+            self.assertIsNotNone(result)
+            self.assertEqual(result["instance_url"], "https://demo.navidrome.org")  # No trailing slash
+            
+            db_result = db_navidrome.get_user_token(self.db_conn, user_id=self.user["id"])
+            self.assertEqual(db_result["host_url"], "https://demo.navidrome.org")  # No trailing slash


MonkeyDo:
@mAmanullah7 No need to merge master in, it is fine if it lags behind provided it will merge cleanly.

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
```diff
@@ -0,0 +1,429 @@
...
+      listen,
+      "track_metadata.additional_info.music_service"
+    );
+    return (
+      isString(musicService) && musicService.toLowerCase().includes("navidrome")
+    );
+  }
+
+  public name = "navidrome";
+  public domainName = "navidrome";
```
MonkeyDo: I this this should be removed. It is used for constructing the music_service domain name when submitting listens from the LB website.
It will default to trying to resolve it from the URL of the track being played:
https://github.com/metabrainz/listenbrainz-server/blob/437be26e276f5bc91037f843c19d0aa83a758367/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx#L421-L433

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
```diff
@@ -0,0 +1,429 @@
...
+    audioElement.removeEventListener("pause", this.onPause);
+    audioElement.removeEventListener("ended", this.onTrackEnd);
+    audioElement.removeEventListener("error", this.onError);
+    audioElement.removeEventListener("canplay", this.onCanPlay);
+  };
+
+  onLoadedMetadata = (): void => {
+    const { onDurationChange } = this.props;
+    const audioElement = this.audioRef.current;
+    if (audioElement) {
```
MonkeyDo: All of these event listeners take in one `event` argument.
The event.target is the audio element, which you can directly target instead of having this check of audioRef, knowing that you will always have an existing event.target
This is true for all the listeners added.

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
```diff
@@ -0,0 +1,429 @@
...
+      return null;
+    }
+
+    return {
+      u: navidromeUser.username,
+      t: navidromeUser.md5_auth_token,
+      s: navidromeUser.salt,
+      v: "1.16.1",
+      c: "listenbrainz",
+      f: "json",
```
MonkeyDo, line 275: Might be good to add some comments to document what these related to, for future developers.

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
```diff
@@ -0,0 +1,429 @@
...
+        "Error searching on Navidrome"
+      );
+    }
+  };
+
+  handleAuthenticationError = (): void => {
+    const { onInvalidateDataSource } = this.props;
+
+    onInvalidateDataSource(
+      this as any,
```
MonkeyDo: Why `any` here? Is it required for compiling properly?
If forcing the type, it should be `as DataSourceTypes` which is what is defined for this function:
https://github.com/metabrainz/listenbrainz-server/blob/437be26e276f5bc91037f843c19d0aa83a758367/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx#L85-L88

Forcing generic types like this is a good way to hide a potential issue, and should be used sparingly, only when there is no other choice.

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
```diff
@@ -0,0 +1,429 @@
...
+    const audioElement = this.audioRef.current;
+    if (audioElement) {
+      const safeVolume = Number.isFinite(volume) ? volume : 100;
+      audioElement.volume = safeVolume / 100;
+    }
+  };
+
+  pauseAudio = (): void => {
+    const audioElement = this.audioRef.current;
+    if (audioElement && !audioElement.paused) {
```
MonkeyDo: Reading the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/pause):
>  if the media is already in a paused state this method will have no effect.

```suggestion
    if (audioElement) {
```

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
```diff
@@ -0,0 +1,429 @@
...
+    this.updateVolume();
+  }
+
+  componentDidUpdate(prevProps: DataSourceProps) {
+    const { show, volume } = this.props;
+    if (prevProps.show !== show) {
+      if (show) {
+        this.setupAudioListeners();
+      } else {
+        this.pauseAudio();
```
MonkeyDo: Since this function is not used anywhere else, is it needed?
I think you could just call directly:

```suggestion
        this.audioRef.current?.pause()
```

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -257,10 +283,135 @@ export default function MusicServices() {
...
+      const username = formData.get("navidromeUsername") as string;
+      const password = formData.get("navidromePassword") as string;
+
+      if (!hostUrl || !username || !password) {
+        throw Error(
+          "Navidrome server URL, username, and password are required"
+        );
+      }
+
+      if (!(hostUrl.startsWith("http://") || hostUrl.startsWith("https://"))) {
```
MonkeyDo: I think this is a bit more explicit and has less potential to go wrong:
```suggestion
      if (!hostUrl.startsWith("http://") && !hostUrl.startsWith("https://")) {
```

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -257,10 +283,135 @@ export default function MusicServices() {
...
+        throw Error("Navidrome server URL must start with http:// or https://");
+      }
+
+      // If we're in edit mode and already connected, disconnect first to avoid duplicates
+      if (navidromeIsEditing && permissions.navidrome === "listen") {
```
MonkeyDo: I wonder if you need a separate navidromeIsEditing state here.

Are we not able to determine this from the available data?
Wouldn't the `permissions.navidrome === "listen"` check here be sufficient to justify disconnecting?

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -257,10 +283,135 @@ export default function MusicServices() {
...
+      setNavidromeIsEditing(true);
+    }
+  };
+
+  React.useEffect(() => {
+    setNavidromeEditValues({
+      hostUrl: navidromeAuth?.instance_url || "",
+      username: navidromeAuth?.username || "",
+    });
+  }, [navidromeAuth?.instance_url, navidromeAuth?.username]);
```
MonkeyDo: The only place I saw where you set the instanceUrl and username are at line 353-355, and you already call setNavidromeEditValues right after that.
Is this effect hook necessary?

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -257,10 +283,135 @@ export default function MusicServices() {
...
+
   React.useEffect(() => {
     const params = new URLSearchParams(window.location.search);
     const funkwhaleError = params.get("error");
     const funkwhaleSuccess = params.get("success");
+    const navidromeError = params.get("navidrome_error");
+    const navidromeSuccess = params.get("navidrome_success");
```
MonkeyDo, line 412: Could this mechanism be made more generic so that it could work fro Funkwhale and Navidrome (and possible future ones) without needing to duplicate all this?

frontend/js/src/settings/music-services/details/MusicServices.tsx
```diff
@@ -600,6 +773,173 @@ export default function MusicServices() {
...
+                      type="radio"
+                      id="navidrome_connected"
+                      name="navidrome"
+                      value="listen"
+                      checked
+                    />
+                    <label htmlFor="navidrome_connected">
+                      <div className="title">Connected to Navidrome</div>
+                      <div className="details">
+                        Your Navidrome server is connected and ready to play
```
MonkeyDo: This doesn't offer much more information and repeats the title. Could be useful to show the URL of the instance, maybe username, so that it is more useful.

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
MonkeyDo: Just saw the comments in the type file, the comment could just point to the type file if you want.

mAmanullah7:
> Was not able to test playback as of yet. Do we have a development Navidrome instance for testing?

Don't have idea @amCap1712 would be right person to answer this i've setup navidrome locally and tested. Cant able to run locally just pulled the latest changes and got a big merge conflict i'll fix this first little busy rn will update you soon once i'm done. Thanks :)

aereaux:
If it would help with testing I can test using my navidrome if there's a listenbrainz-server instance I can test with, or I can provide access to my navidrome temporarily.

MonkeyDo:
> just pulled the latest changes and got a big merge conflict i'll fix this first little busy rn will update you soon once i'm done.

@mAmanullah7 after a force-push (like lucifer did) you need to drop your local changes and reset it > https://stackoverflow.com/questions/14787776/how-to-pull-after-a-forced-git-push

@amCap1712 you might want to check the latest merge commit against what you force-pushed before

amCap1712:
@mAmanullah7 you can use [https://demo.navidrome.org](https://demo.navidrome.org) for testing.

MonkeyDo:
> If it would help with testing I can test using my navidrome if there's a listenbrainz-server instance I can test with, or I can provide access to my navidrome temporarily.

Thanks for the offer @aereaux I have deployed this PR to test.listenbrainz.org  if you would like to test it.

As for local testing, I will also set up navidrome locally for easier testing.

aereaux:
> Thanks for the offer @aereaux I have deployed this PR to test.listenbrainz.org if you would like to test it.

Thanks!  I'm unable to connect the service, though:

```
Error: Unable to connect to server: 404 Client Error: Not Found for url: https://server2.jmad.org/rest/ping?u=<username>&t=<token>&s=<s>&v=1.16.0&c=ListenBrainz&f=json
```

My navidrome is at `https://server2.jmad.org/music/` not `https://server2.jmad.org/`, even though I entered `https://server2.jmad.org/music/` in the "Navidrome Server URL" field.  It looks like it's just using the host or eating some of the URL?

mAmanullah7:
> @mAmanullah7 you can use https://demo.navidrome.org for testing.

Thanks @amCap1712, for testing i've already set up navidrome locally and using same for testing, last time when i tested everything was fine i'll first solve these conflicts locally and then i'll retest.


> My navidrome is at `https://server2.jmad.org/music/` not `https://server2.jmad.org/`, even though I entered `https://server2.jmad.org/music/` in the "Navidrome Server URL" field. It looks like it's just using the host or eating some of the URL?

@aereaux i've fixed it should work fine! You can now test it locally if you wants. Thanks :) for reporting the issue.

> @mAmanullah7 after a force-push (like lucifer did) you need to drop your local changes and reset it > https://stackoverflow.com/questions/14787776/how-to-pull-after-a-forced-git-push

Thanks for this @MonkeyDo really need this!

sbruder:
I just tested this on the ListenBrainz test instance, and it worked quite well. One thing that I noticed is that it matches the listen to the track on Navidrome by searching for the recording name and artist name. I think this can lead to issues in two cases:

- There are multiple recordings by the same artist with the same title. While it seems unlikely, I have multiple instances of this in my library, like C418’s album [life changing moments seem minor in pictures](https://musicbrainz.org/release/d52c016f-7a08-4a78-abc9-1331c5e3666b), which includes 7 tracks with the name “Long songs are for people who don't have time to make them short”, all being quite different.
- There are multiple releases in a release group with different track names for the same recording. This is quite common with transliterated pseudo-releases. It can also happen outside of that, like in the case of [\[unknown\]](https://musicbrainz.org/recording/05248c45-a89f-4dbb-988f-37eba5d1b3a4) on Radiohead’s Kid A and KID A MNESIA (only counting official releases), which is also known as “\[untitled\]” or “Untitled”. Therefore, the ListenBrainz backend wasn’t able to match the MusicBrainz recording (“\[unknown\]”) to a song on navidrome (“\[untitled\]”).

In my tests, navidrome always returned the correct recording (though not necessarily from the right release) when searching directly for the MusicBrainz recording ID. This leaves the possibility of incorrect results when multiple versions of the same recording are present in the Navidrome library, though it should still be more accurate and that problem could at least somewhat be mitigated by matching the album name to the release name (as Navidrome does not return the release MBID on the `search3` endpoint).

This, however, would require the Navidrome library to be tagged with MusicBrainz IDs as the search would otherwise fail. One possibility to work around this is to fall back to the current approach when the MBID search returns no results.

Would my described approach be possible to implement for this feature? I think it would make this feature more robust.

In any case, thank you all for working on making streaming from self-owned sources possible!

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
```diff
@@ -0,0 +1,429 @@
...
+          return;
+        }
+      }
+
+      handleWarning(
+        `"${trackName}" by ${artistName} is not available on your Navidrome server`,
+        "Track not available on Navidrome"
+      );
+      onTrackNotFound();
+    } catch (errorObject) {
```
MonkeyDo, line 346: I think you're missing a call to `onTrackNotFound` in this catch block or somewhere else, but if the search mechanism fails it completely stalls the player for me.

I am testing with the demo.navidrome.org server and the response in not JSON in one case (just for some context).
Regardless of the reason, the thrown error results in the toast error I see in this catch block, but nothing else happens.
It should be trying to play the track with the next datasource by calling onTrackNotFound.

frontend/js/src/utils/utils.tsx
```diff
@@ -326,6 +326,59 @@ const searchForFunkwhaleTrack = async (
...
+    const response = await fetch(searchUrl);
+
+    if (!response.ok) {
+      const errorBody = await response.json().catch(() => ({}));
+      const error = new Error(errorBody.detail || response.statusText);
+      (error as any).status = response.status;
+      throw error;
+    }
+
+    const data = await response.json();
```
MonkeyDo: Might be good to check the response headers to see if we expect a json response.

Testing with the demo.navidrome.org server, I get a 302 redirection which results in an HTML page. Calling response.json() on that means you get a less than useful parsing error:

Image

frontend/js/src/utils/utils.tsx
```diff
@@ -326,6 +326,59 @@ const searchForFunkwhaleTrack = async (
...
+    );
+  }
+
+  const query = `${trackName || ""} ${artistName || ""}`.trim();
+  if (!query) {
+    throw new Error("No search terms provided");
+  }
+
+  try {
+    const searchUrl = `${instanceURL}/rest/search3?query=${encodeURIComponent(
```
MonkeyDo: I think the reason my calls to the navidrome demo server are failing might be because there is  a double slash in the URL of the request I see:

`https://demo.navidrome.org//rest/search3?query=Prayer%20for%20Rain.....`

You can ensure in getNavidromeInstanceURL that the returned string contains/does not contains a trailing slash, and modify the rest of the code accordingly.

Since your reflex (and mine) is to write a / in these template strings like `${instanceURL}/rest/...`, maybe ensuring there is no trailing slash would be better.

Something using `instanceURL.endsWith("/")` and `instanceURL.slice(0, -1)` to remove the last character.


See also in APIService for reference:
https://github.com/metabrainz/listenbrainz-server/blob/e7cc76b5680f64d7a589a90cd502b26b732c9cc7/frontend/js/src/utils/APIService.ts#L22-L24

MonkeyDo (commented):
Some more comments while testing locally.

frontend/js/src/utils/utils.tsx
mAmanullah7: @MonkeyDo just wanna confirm is this error still exist its same issue reported by @aereaux i've fixed that yesterday itself and pushed the changes 
https://github.com/metabrainz/listenbrainz-server/pull/3331/commits/ccdee6679e94d9f32e48fcee7c20e5bd43ba0df2 i've fi think this should resolve the issue.

mAmanullah7: Screenshot 2025-09-09 at 11 35 32 PM
i've tested can you once test again after pulling latest changes. i think we should let backend handle the validation of `host_url` rather explicitly checking same thing in frontend side!. testted using same demo server: https://demo.navidrome.org/

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: yeah `permissions.navidrome === "listen"` is enough. the thing is that was trying random approach to avoid duplication but decided two way approach first disconnect then connect! will remove this now no use even it can be useful if inputs are not in readOnly.

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
mAmanullah7: i'll add docs for reference.

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
mAmanullah7:
```typescript
Found 1 error in 8613 ms.
static_builder-1    | ERROR in ./frontend/js/src/common/brainzplayer/BrainzPlayer.tsx:436:50
static_builder-1    | TS2339: Property 'domainName' does not exist on type 'DataSourceTypes'.
static_builder-1    |   Property 'domainName' does not exist on type 'NavidromePlayer'.
static_builder-1    |     434 |
static_builder-1    |     435 |     const musicServiceName = dataSource.current?.name;
static_builder-1    |   > 436 |     let musicServiceDomain = dataSource.current?.domainName;
static_builder-1    |         |                                                  ^^^^^^^^^^
static_builder-1    |     437 |     // Best effort try?
static_builder-1    |     438 |     if (!musicServiceDomain && brainzPlayerContextRef.current.currentTrackURL) {
static_builder-1    |     439 |       try {
```
this wont be any problem? added to avoid this error.?

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
mAmanullah7: when i used to switch between Navidrome and Funkwhale branche, there's DataSourceTypes mismatch that was causing the page to break, so to avoid that.

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: yes i can  see its duplicating what's already being done, better to remove!

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: Yes we can simplify rather hardcoding we can make it more generic! if you don't mind  can we use this as of now i'll crate a pr soon just little busy in college stuff.!

frontend/js/src/settings/music-services/details/MusicServices.tsx
mAmanullah7: if we display url or username wont look nice as already both url and username can be seen in user input field. Displaying same thing wont look nice rather i'll simplify similar to funkwhale and remove extra lines of code!

mAmanullah7:
@sbruder Thank you so much for the thorough testing and detailed feedback! I really appreciate you taking the time to test and identifying these potential matching issues. I'll surely look into your proposed solution, i think since this is our first time integrating Navidrome into LB, i wanted to start with the basics. We can implement this enhancement in future after some feedback from community.

aereaux:
One question I have is if this submits the completed listens back to navidrome or directly to listenbrainz, or neither?  Or is that going to be a further improvement?

mAmanullah7:
> One question I have is if this submits the completed listens back to navidrome or directly to listenbrainz, or neither? Or is that going to be a further improvement?

@aereaux it does submit listen to listenbrainz, but not to navidrome. Noted for future improvement

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
MonkeyDo: The comment above was not addressed

frontend/js/src/utils/utils.tsx
MonkeyDo: This comment was not addressed.
Here's a good example of how to check the response type:
https://github.com/metabrainz/listenbrainz-server/blob/b64df9a78807609c57f8d7d269550848016b7c9e/frontend/js/src/settings/music-services/details/MusicServices.tsx#L226-L231

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
MonkeyDo, line 1: We are missing one change without which the player is never activated.
How did you manage to test the new player source without that change?

On line 203 or thereabouts, you need to add 
`userPreferences?.brainzplayer?.navidromeEnabled === false &&` like we have for the other sources.
We destructure userPreferences.brainzplayer later on, on line 238, so it's easy to miss, although as I stated above I am curious how you got it working for testing.
I will refactor this bit of code after merging this PR to make it harder to forget this in the future.

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
```diff
@@ -0,0 +1,436 @@
...
+    const musicService = _get(
+      listen,
+      "track_metadata.additional_info.music_service"
+    );
+    return (
+      isString(musicService) && musicService.toLowerCase().includes("navidrome")
+    );
+  }
+
+  public name = "navidrome";
```
MonkeyDo, line 44: My bad! I previously asked you to remove `domainName`, but it is causing an error as it is expected by other parts of the code.

However for our use case we can set it to `false` rather than the string "navidrome.org", which wouldn't make sense in this context as it is used to construct listen metadata.


```suggestion
  public name = "navidrome";
  public domainName = false;
```

frontend/js/src/utils/utils.tsx
MonkeyDo: The double slash endpoint is still an issue.
I tested connecting to https://demo.navidrome.org (no double slash) and https://demo.navidrome.org/ (with trailing slash, causes API call to `https://demo.navidrome.org//rest/search3?query=` with double slash) since both are accepted as-is by the server.

I'm not against sanitizing on the sever side but then that needs to be implemented.

I still think it is fine to add the code I suggested above to `getNavidromeInstanceURL` since it's mainly the front-end who cares about this trailing slash

frontend/js/src/common/brainzplayer/NavidromePlayer.tsx
mAmanullah7: my bad it was on my list thought i'll update this once i got confirmation about the trailing slash issue :)

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
mAmanullah7: as far i remembered i've included as you mentioned above but i think at the time of merge conflicts either with internet archive or with others  by mistake it was removed, lot of duplicates and others things i've fixed later on even after that resolving merge conflicts!

frontend/js/src/utils/utils.tsx
mAmanullah7: Initially i thought once the trailing slash issue will be resolved then we don't need this validation as it will always return json but realised we never know any kind of unexpected things can be happen!

frontend/js/src/utils/utils.tsx
mAmanullah7: Yeah i've fixed this issue now added validation both backend and frontend side, now we are saving `host_url` after validating and removing extra trailing slashes, i thought like if we don't have any trailing slashes then we don't need this frontend validation but as you said that it gonna fail if user not refresh their page and i tried to dry run and yeah that was a good one which i didn't noticed thanks :)

MonkeyDo (approved):
I took the liberty to remove duplicate URL normalization in  `searchForNavidromeTrack`.
The normalization is already done by `getNavidromeInstanceURL` when you call the function.
https://github.com/metabrainz/listenbrainz-server/blob/d3ef052ec2ea33d245ab763bea6eab76d4703601/frontend/js/src/common/brainzplayer/NavidromePlayer.tsx#L323-L329


With all that done, I was able to successfully play a track from the Navidrome demo instance! 🎉 Nice work !

I will note that the search needs some more improvements*, but that we can merge and deploy this PR.

* We need to play with search terms to improve our chances:
I looked up a track on the demo instance, then added it as a listen to my local LB profile so I could play it.
Hitting the search endpoint, it returned no results which was surprising, until I realized it's because the featureing artist is in the track name on navidrome but in the artist name on LB:
- searching for ["Vision Fiberoptics The Polish Ambassador ft. Sean Haefeli"](https://demo.navidrome.org/rest/search3?query=Vision%20Fiberoptics%20The%20Polish%20Ambassador%20ft.%20Sean%20Haefeli&songCount=1&u=demo&t=b6054fd4dfe6b9888da01e789bbef132&s=1759748056224&v=1.16.1&c=listenbrainz&f=json) returns no results
- searching without the artist name ["Vision Fiberoptics"](https://demo.navidrome.org/rest/search3?query=Vision%20Fiberoptics&songCount=1&u=demo&t=b6054fd4dfe6b9888da01e789bbef132&s=1759748056224&v=1.16.1&c=listenbrainz&f=json) does return the track I am looking for, but could also return a track by another artist in some circumstances
- Searching with artist name but without the featuring artist ["Vision Fiberoptics The Polish Ambassador"](https://demo.navidrome.org/rest/search3?query=Vision%20Fiberoptics%20The%20Polish%20Ambassador&songCount=1&u=demo&t=b6054fd4dfe6b9888da01e789bbef132&s=1759748056224&v=1.16.1&c=listenbrainz&f=json) is the best of both worlds, but requires making assumptions in the front-end and modifying search terms in complicated ways.

I don't see a way to separate track name and artist name in the subsonic API, so not sure what the path forwards is...

mAmanullah7:
> I took the liberty to remove duplicate URL normalization in `searchForNavidromeTrack` (pushed in commit [fb135cc](https://github.com/metabrainz/listenbrainz-server/commit/fb135cca5e075795948414b4fa4d076006b9ccec)). The normalization is already done by `getNavidromeInstanceURL` when you call the function.
> 
> https://github.com/metabrainz/listenbrainz-server/blob/d3ef052ec2ea33d245ab763bea6eab76d4703601/frontend/js/src/common/brainzplayer/NavidromePlayer.tsx#L323-L329
> 
> With all that done, I was able to successfully play a track from the Navidrome demo instance! 🎉 Nice work !
> 
> I will note that the search needs some more improvements, but that we can merge and deploy this PR.
> 
> * We need to play with search terms to improve our chances:
>   I looked up a track on the demo instance, then added it as a listen to my local LB profile so I could play it.
>   Hitting the search endpoint, it returned no results which was surprising, until I realized it's because the featureing artist is in the track name on navidrome but in the artist name on LB:
> * searching for ["Vision Fiberoptics The Polish Ambassador ft. Sean Haefeli"](https://demo.navidrome.org/rest/search3?query=Vision%20Fiberoptics%20The%20Polish%20Ambassador%20ft.%20Sean%20Haefeli&songCount=1&u=demo&t=b6054fd4dfe6b9888da01e789bbef132&s=1759748056224&v=1.16.1&c=listenbrainz&f=json) returns no results
> * searching without the artist name ["Vision Fiberoptics"](https://demo.navidrome.org/rest/search3?query=Vision%20Fiberoptics&songCount=1&u=demo&t=b6054fd4dfe6b9888da01e789bbef132&s=1759748056224&v=1.16.1&c=listenbrainz&f=json) does return the track I am looking for, but could also return a track by another artist in some circumstances
> * Searching with artist name but without the featuring artist ["Vision Fiberoptics The Polish Ambassador"](https://demo.navidrome.org/rest/search3?query=Vision%20Fiberoptics%20The%20Polish%20Ambassador&songCount=1&u=demo&t=b6054fd4dfe6b9888da01e789bbef132&s=1759748056224&v=1.16.1&c=listenbrainz&f=json) is the best of both worlds, but requires making assumptions in the front-end and modifying search terms in complicated ways.
> 
> I don't see a way to separate track name and artist name in the subsonic API, so not sure what the path forwards is...

I'll look into this if i find ant possible way i'll update this!

PR #3333, opened 2025-08-01
Add notification_sender.py to use notification/send endpoint

fettuccinae:
Integration of MetaBrainz `notification/send `endpoint into ListenBrainz

Changes:
Added `send_multiple_notifications`function:
Sends a list of notification dictionaries to the `metabrainz.org/notification/send `endpoint.

Added `fetch_token `function:
Retrieves an access token from Redis cache if available.
If not found or expired, it uses the client credentials grant flow to request a new token.

### DIFF
--- a/listenbrainz/domain/notification_sender.py
+++ b/listenbrainz/domain/notification_sender.py
@@ -1,0 +1,99 @@
+from requests.auth import HTTPBasicAuth
+from oauthlib.oauth2 import BackendApplicationClient
+from requests_oauthlib import OAuth2Session
+from brainzutils import cache
+from flask import current_app
+import requests
+
+
+TOKEN_CACHE_KEY = "notification_access_token"
+# METABRAINZ_SEND_NOTIFICATIONS_URL = "https://metabrainz.org/notification/send"
+METABRAINZ_SEND_NOTIFICATIONS_URL = "http://metabrainz-web-1:8000/notification/send"
+
+
+def send_notification(
+    subject: str,
+    body: str,
+    user_id: int,
+    user_email: str,
+    from_addr: str,
+    project: str = "listenbrainz",
+    send_email: bool = True,
+    important: bool = True,
+    expire_age: int = 7,
+):
+    """Function to send a single notification.
+    This function prepares a single notification and uses `send_multiple_notifications` to send it.
+
+    Args:
+        ``subject``
+        ``body``
+        ``userid``
+        ``user_email``
+        ``from_addr``
+        ``project``
+        ``send_email``
+        ``important``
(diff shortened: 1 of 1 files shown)


fettuccinae:
PR ready for review -- @mayhem @amCap1712

mayhem:
I've created a new empty branch called "metabrainz-notifications", same as meb.org. Can you please rebase this PR to target that branch instead?

fettuccinae:
> I've created a new empty branch called "metabrainz-notifications", same as meb.org. Can you please rebase this PR to target that branch instead?


Done.

listenbrainz/domain/notification_sender.py
```diff
@@ -0,0 +1,99 @@
...
+
+    token = _fetch_token()
+    tokenized_url = METABRAINZ_SEND_NOTIFICATIONS_URL + f"?token={token}"
+
+    response = requests.post(url=tokenized_url, json=notifications)
+    response.raise_for_status()
+
+
+def _fetch_token() -> str:
+    """Helper function to fetch token from redis cache, If no token is found or it's expired, a new token is requested."""
```
mayhem: Token for what? We should mention that we're fetching OAuth2 tokens.

mayhem (approved):
Looks ok to me, but lets wait for @amCap1712 to also have a look.

listenbrainz/domain/notification_sender.py
fettuccinae: Ahh yes, updated the docstring.

listenbrainz/domain/notification_sender.py
```diff
@@ -0,0 +1,98 @@
...
+    Args:
+        ``notifications``: A list of notification dictionaries to be sent.
+
+    Raises:
+        A HTTPError if there's a failure.
+
+    """
+
+    token = _fetch_token()
+    tokenized_url = METABRAINZ_NOTIFICATIONS_SEND_URL + f"?token={token}"
```
amCap1712: Why not pass the token in headers?

fettuccinae: https://github.com/metabrainz/metabrainz.org/blob/16057b95f7892f621aa8ba7bc6dec5c3eb2408bb/metabrainz/decorators.py#L80

The decorator was set-up to send the token through url instead of authorization header for simplicity

amCap1712: Sorry I missed this in that PR, please update to use the Authorization header only in both MeB and here.

fettuccinae: Done.

PR #3341, opened 2025-08-13
Replace send_mail instances with send_notification

fettuccinae:
Changed send_mail instances to use send_notifications.

### DIFF
--- a/listenbrainz/background/export.py
+++ b/listenbrainz/background/export.py
@@ -6,7 +6,7 @@ from datetime import datetime, date, time, timedelta, timezone
 from pathlib import Path
 
 import orjson
-from brainzutils.mail import send_mail
+from listenbrainz.domain.notification_sender import send_notification
 from dateutil.relativedelta import relativedelta
 from flask import current_app, render_template
 from sqlalchemy import text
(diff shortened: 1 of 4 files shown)


fettuccinae:
@mayhem-- PR ready for review.

listenbrainz/listens_importer/tests/test_spotify_read_listens.py
```diff
@@ -100,16 +100,16 @@ def test_parse_play_to_listen_many_artists(self):
 
         self.assertDictEqual(listen, expected_listen)
 
-    @patch('listenbrainz.listens_importer.base.send_mail')
-    def test_notify_user(self, mock_send_mail):
+    @patch('listenbrainz.listens_importer.base.send_notification')
```
mayhem, line 103: I think all the notification related tests should be in their own file.

fettuccinae: I think test is here because its for the `notify_error `function that runs when the importer fails. I didnt originally write the test, just made a change to use `send_notifications` instead of `send_mail`. Should I remove it?

mayhem:
if it was already there and if its works right now, then great. lets keep it. Ready for merge then?

fettuccinae:
>Ready for merge then?

Yes, please.

PR #3343, opened 2025-08-14
Use AlgorithmEnum in artist similarity labs API

MonkeyDo:
While pointing out the existence of the similarity endpoints to a user, I realized the recording similarity page can be used without much prior knowledge because the algorithm input is a dropdown with the available options.

The artist similarity page on the other hand does not have that, requiring that you know the magic string that will make this  incantation work.

Porting the logic over from recording to artist similarity, although I don't exactly know how to test this.


MonkeyDo:
@amCap1712 I have no idea how to test this, asking for your help :)

As far as I could tell we don't have our test setup available to access labs API running on test, and I don;t have the required data locally to test it that way either.

PR #3344, opened 2025-08-15
Add frontend for digest settings

fettuccinae:
Added a Notification Settings page according to the feedback by @MonkeyDo in #3345. 
EDIT: UI mockups

Screenshot 2025-08-29 173917
Screenshot 2025-08-29 173932

### DIFF
--- a/frontend/js/src/settings/layout.tsx
+++ b/frontend/js/src/settings/layout.tsx
@@ -29,7 +29,7 @@ const sections: Section[] = [
       { to: "./", label: "User settings" },
       {
         to: "notifications/",
-        label: "Notification settings",
+        label: "Notifications",
       },
       { to: "select_timezone/", label: "Timezone" },
       { to: "troi/", label: "Playlist preferences" },
--- a/frontend/js/src/settings/layout.tsx
+++ b/frontend/js/src/settings/layout.tsx
@@ -27,6 +27,10 @@ const sections: Section[] = [
     title: "Account",
     links: [
       { to: "./", label: "User settings" },
+      {
+        to: "notification-settings/",
+        label: "Notification settings",
+      },
       { to: "select_timezone/", label: "Timezone" },
       { to: "troi/", label: "Playlist preferences" },
       { to: "export/", label: "Export data" },
--- a/frontend/js/src/settings/layout.tsx
+++ b/frontend/js/src/settings/layout.tsx
@@ -28,7 +28,7 @@ const sections: Section[] = [
     links: [
       { to: "./", label: "User settings" },
       {
-        to: "notification/",
+        to: "notifications/",
         label: "Notification settings",
       },
       { to: "select_timezone/", label: "Timezone" },
--- a/frontend/js/src/settings/layout.tsx
+++ b/frontend/js/src/settings/layout.tsx
@@ -28,7 +28,7 @@ const sections: Section[] = [
     links: [
       { to: "./", label: "User settings" },
       {
-        to: "notification-settings/",
+        to: "notification/",
         label: "Notification settings",
       },
       { to: "select_timezone/", label: "Timezone" },
--- a/frontend/js/src/settings/notification-settings/NotificationSettings.tsx
+++ b/frontend/js/src/settings/notification-settings/NotificationSettings.tsx
@@ -7,7 +7,7 @@ import Switch from "../../components/Switch";
 export default function NotificationSettings() {
   const [notificationsEnabled, setNotificationsEnabled] = React.useState(false);
   const [digestEnabled, setDigestEnabled] = React.useState(false);
-  const [digestAge, setDigestAge] = React.useState<number | null>(null);
+  const [digestAge, setDigestAge] = React.useState<number>(7);
   const [initialDigestEnabled, setInitialDigestEnabled] = React.useState(false);
   const [initialDigestAge, setInitialDigestAge] = React.useState<number | null>(
     null
@@ -131,11 +131,10 @@ export default function NotificationSettings() {
                 id="digest-age"
                 min={1}
                 max={100}
-                value={digestAge ?? ""}
+                defaultValue={digestAge}
                 style={{ maxWidth: "4em" }}
                 onChange={(e) => {
-                  const inputValue = e.target.value;
-                  setDigestAge(inputValue === "" ? null : Number(inputValue));
+                  setDigestAge(Number(e.target.value));
                 }}
               />{" "}
               days
@@ -146,7 +145,7 @@ export default function NotificationSettings() {
           <button
             className="btn btn-lg btn-info"
             type="submit"
-            disabled={saving || hasNoChanges}
+            disabled={saving || hasNoChanges || !digestAge}
           >
             Save notification settings
           </button>
--- a/frontend/js/src/settings/notification-settings/NotificationSettings.tsx
+++ b/frontend/js/src/settings/notification-settings/NotificationSettings.tsx
@@ -1,43 +1,25 @@
 import * as React from "react";
 import { toast } from "react-toastify";
 import { Helmet } from "react-helmet";
+import { useLoaderData } from "react-router";
 
 import Switch from "../../components/Switch";
 
-export default function NotificationSettings() {
-  const [notificationsEnabled, setNotificationsEnabled] = React.useState(false);
-  const [digestEnabled, setDigestEnabled] = React.useState(false);
-  const [digestAge, setDigestAge] = React.useState<number>(7);
-  const [initialDigestEnabled, setInitialDigestEnabled] = React.useState(false);
-  const [initialDigestAge, setInitialDigestAge] = React.useState<number | null>(
-    null
-  );
-  const [loading, setLoading] = React.useState(true);
-  const [saving, setSaving] = React.useState(false);
+type NotificationPreferenceLoaderData = {
+  digest: boolean;
+  digest_age: number;
+};
 
-  React.useEffect(() => {
-    async function fetchDigestSettings() {
-      try {
-        const response = await fetch("/settings/digest-setting/");
-        if (!response.ok) {
-          throw new Error(`${response.status} HTTP response.`);
-        }
-        const data = await response.json();
-        setDigestEnabled(data.digest);
-        setDigestAge(data.digest_age);
-        setInitialDigestEnabled(data.digest);
-        setInitialDigestAge(data.digest_age);
-        setNotificationsEnabled(data.digest);
-      } catch (error) {
-        // eslint-disable-next-line no-console
-        console.error("Could not fetch notification settings.", error);
-        toast.error("Failed to load notification settings.");
-      } finally {
-        setLoading(false);
-      }
-    }
-    fetchDigestSettings();
-  }, []);
+export default function NotificationSettings() {
+  const loaderData = useLoaderData() as NotificationPreferenceLoaderData;
+
+  const [digestEnabled, setDigestEnabled] = React.useState(loaderData.digest);
+  const [digestAge, setDigestAge] = React.useState(loaderData.digest_age);
+  // change this when notifications API is added.
+  const [notificationsEnabled, setNotificationsEnabled] = React.useState(
+    loaderData.digest
+  );
+  const [saving, setSaving] = React.useState(false);
 
   const updateDigestSettings = async (e: React.FormEvent) => {
     e.preventDefault();
@@ -53,25 +35,28 @@ export default function NotificationSettings() {
           digest_age: digestAge,
         }),
       });
+      const responseData = await response.json();
       if (!response.ok) {
-        throw new Error(`${response.status} HTTP response.`);
+        throw new Error(responseData.error);
       }
       toast.success("Notification settings saved successfully");
     } catch (error) {
-      // eslint-disable-next-line no-console
-      console.error("Could not update notification settings.", error);
-      toast.error("Failed to save notification settings. Please try again.");
+      toast.error(`Failed to save notification settings.${error}`);
     } finally {
       setSaving(false);
     }
   };
 
   const hasNoChanges =
-    digestEnabled === initialDigestEnabled && digestAge === initialDigestAge;
+    // change this when notifications API is added.
+    notificationsEnabled === loaderData.digest &&
+    digestEnabled === loaderData.digest &&
+    digestAge === loaderData.digest_age;
 
-  if (loading) {
-    return <div>Loading notification settings...</div>;
-  }
+  const isDigestAgeValid =
+    !notificationsEnabled ||
+    !digestEnabled ||
+    (notificationsEnabled && digestEnabled && digestAge);
 
   return (
     <>
@@ -131,10 +116,10 @@ export default function NotificationSettings() {
           />
           {digestEnabled && (
             <div className="mb-4">
-              Send digest emails every{" "}
+              Send digest emails every
               <input
                 type="number"
-                className="d-inline form-control"
+                className="d-inline form-control ms-2 me-2"
                 id="digest-age"
                 min={1}
                 max={100}
@@ -143,7 +128,7 @@ export default function NotificationSettings() {
                 onChange={(e) => {
                   setDigestAge(Number(e.target.value));
                 }}
-              />{" "}
+              />
               days
             </div>
           )}
@@ -152,7 +137,7 @@ export default function NotificationSettings() {
           <button
             className="btn btn-lg btn-info"
             type="submit"
-            disabled={saving || hasNoChanges || !digestAge}
+            disabled={saving || hasNoChanges || !isDigestAgeValid}
           >
             Save notification settings
           </button>
--- a/frontend/js/src/settings/notification-settings/NotificationSettings.tsx
+++ b/frontend/js/src/settings/notification-settings/NotificationSettings.tsx
@@ -1,0 +1,157 @@
+import * as React from "react";
+import { toast } from "react-toastify";
+import { Helmet } from "react-helmet";
+
+import Switch from "../../components/Switch";
+
+export default function NotificationSettings() {
+  const [notificationsEnabled, setNotificationsEnabled] = React.useState(false);
+  const [digestEnabled, setDigestEnabled] = React.useState(false);
+  const [digestAge, setDigestAge] = React.useState<number | null>(null);
+  const [initialDigestEnabled, setInitialDigestEnabled] = React.useState(false);
+  const [initialDigestAge, setInitialDigestAge] = React.useState<number | null>(
+    null
+  );
+  const [loading, setLoading] = React.useState(true);
+  const [saving, setSaving] = React.useState(false);
+
+  React.useEffect(() => {
+    async function fetchDigestSettings() {
+      try {
+        const response = await fetch("/settings/digest-setting/");
+        if (!response.ok) {
+          throw new Error(`${response.status} HTTP response.`);
+        }
+        const data = await response.json();
+        setDigestEnabled(data.digest);
+        setDigestAge(data.digest_age);
+        setInitialDigestEnabled(data.digest);
+        setInitialDigestAge(data.digest_age);
+        setNotificationsEnabled(data.digest);
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error("Could not fetch notification settings.", error);
+        toast.error("Failed to load notification settings.");
+      } finally {
+        setLoading(false);
+      }
+    }
+    fetchDigestSettings();
+  }, []);
+
+  const updateDigestSettings = async (e: React.FormEvent) => {
+    e.preventDefault();
+    setSaving(true);
+    try {
+      const response = await fetch("/settings/digest-setting/", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify({
+          digest: digestEnabled && notificationsEnabled,
+          digest_age: digestAge,
+        }),
+      });
+      if (!response.ok) {
+        throw new Error(`${response.status} HTTP response.`);
+      }
+      toast.success("Notification settings saved successfully");
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error("Could not update digest settings.", error);
+      toast.error("Failed to save digest settings. Please try again.");
+    } finally {
+      setSaving(false);
+    }
+  };
+
+  const hasNoChanges =
+    digestEnabled === initialDigestEnabled && digestAge === initialDigestAge;
+
+  if (loading) {
+    return <div>Loading notification settings...</div>;
+  }
+
+  return (
+    <>
+      <Helmet>
+        <title>Notification Settings</title>
+      </Helmet>
+      <h2 className="page-title">Notification settings</h2>
+      <form onSubmit={updateDigestSettings}>
+        <Switch
+          id="enable-notifications"
+          checked={notificationsEnabled}
+          onChange={(e) => setNotificationsEnabled(!notificationsEnabled)}
+          value="notifications"
+          switchLabel={
+            <span
+              className={`text-brand ${
+                !notificationsEnabled ? "text-muted" : ""
+              }`}
+            >
+              <span>Enable e-mail notifications</span>
+            </span>
+          }
+        />
+        <details open={notificationsEnabled}>
+          <summary
+            className="mt-4"
+            onClick={(e) => {
+              e.preventDefault();
+            }}
+          >
+            {!notificationsEnabled && (
+              <p className="alert alert-info mb-0 d-inline-block">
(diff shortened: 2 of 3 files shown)


fettuccinae:
Hey @MonkeyDo , Could you take a look this PR ?

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
```diff
@@ -0,0 +1,156 @@
...
+              }`}
+            >
+              <span>Enable e-mail notifications</span>
+            </span>
+          }
+        />
+        <details open={notificationsEnabled}>
+          <summary
+            className="mt-4"
+            onClick={(e) => {
```
fettuccinae, line 93: I added this because clicking on the summary would toggle the details's state. (Same thing happens in BrainzPlayer settings.)

fettuccinae:
@MonkeyDo

frontend/js/src/settings/routes/index.tsx
```diff
@@ -122,6 +122,14 @@ const getSettingsRoutes = (): RouteObject[] => {
           // Keep the /settings/import/ route for LastFM/LibreFM historical links,
           // and redirect to the music services page that replace those manual importers
           element: <Navigate to="../music-services/details/" replace />,
+          },
+        {
+          path: "notification-settings/",
```
MonkeyDo: To avoid repeating in "/settings/notification-settings/":
```suggestion
          path: "notifications/",
```

frontend/js/src/settings/layout.tsx
```diff
@@ -27,6 +27,10 @@ const sections: Section[] = [
     title: "Account",
     links: [
       { to: "./", label: "User settings" },
+      {
+        to: "notification-settings/",
```
MonkeyDo: Along with the change above: 
```suggestion
        to: "notifications/",
```

fettuccinae:
> Is this ready for review?

Yes @MonkeyDo , its ready for review. Also about [mayhem's comment on the previous PR](https://github.com/metabrainz/listenbrainz-server/pull/3345#pullrequestreview-3135902457), where do you think we should show this information?

MonkeyDo:
> Yes @MonkeyDo , its ready for review. Also about [mayhem's comment on the previous PR](https://github.com/metabrainz/listenbrainz-server/pull/3345#pullrequestreview-3135902457), where do you think we should show this information?

I rephrased it a bit and suggest the simplest solution of them all: an HTML paragraph (`<p>`) with the text:

```html
We will always email you right away about <b>important account issues</b> (like problems with connected services).<br/>
For everything else, you can choose to turn off notifications or receive them in a periodic digest.
```

image

And keeping the alert-info to catch user's eyes if they disable email notifications:
image

fettuccinae:
@MonkeyDo, Done.
I have binded notificationEnabled and digestEnabled booleans, because there is no endpoint for us to set notification options rn, and if a user have their digest disabled, we only send important notifications(for now).

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const [initialDigestAge, setInitialDigestAge] = React.useState<number | null>(
+    null
+  );
+  const [loading, setLoading] = React.useState(true);
+  const [saving, setSaving] = React.useState(false);
+
+  React.useEffect(() => {
+    async function fetchDigestSettings() {
+      try {
+        const response = await fetch("/settings/digest-setting/");
```
MonkeyDo: While this is correct and works, how we usually do this initial data fetching is with a react-router RouteLoader utility that automatically hits a POST endpoint on the server of the same name as the page. See for example:
https://github.com/metabrainz/listenbrainz-server/blob/c882a9594398d52f6b77b7b5ca64eaaeb0e6a3cc/frontend/js/src/settings/routes/index.tsx#L34-L42

And that way, you can use the initial data with a hook, and as a bonus you won't need to store the 'initial...' values in the state as they will be present in the loaderData for you to compare with:
https://github.com/metabrainz/listenbrainz-server/blob/c882a9594398d52f6b77b7b5ca64eaaeb0e6a3cc/frontend/js/src/settings/music-services/details/MusicServices.tsx#L44-L50

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
```diff
@@ -0,0 +1,163 @@
...
+          digest_age: digestAge,
+        }),
+      });
+      if (!response.ok) {
+        throw new Error(`${response.status} HTTP response.`);
+      }
+      toast.success("Notification settings saved successfully");
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error("Could not update notification settings.", error);
```
MonkeyDo: Probably don't want to log this to console, and instead show the error to the user in the toast message

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
```diff
@@ -0,0 +1,163 @@
...
+      setSaving(false);
+    }
+  };
+
+  const hasNoChanges =
+    digestEnabled === initialDigestEnabled && digestAge === initialDigestAge;
+
+  if (loading) {
+    return <div>Loading notification settings...</div>;
+  }
```
MonkeyDo: If using the react-router RouteLoader, this won't be necessary anymore.

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  return (
+    <>
+      <Helmet>
+        <title>Notification Settings</title>
+      </Helmet>
+      <h2 className="page-title">Notification settings</h2>
```
MonkeyDo, line 66: What is this css class from?
```suggestion
      <h2>Notification settings</h2>
```

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
```diff
@@ -0,0 +1,163 @@
...
+          <Switch
+            id="enable-digest"
+            value="digest"
+            checked={digestEnabled}
+            onChange={(e) => setDigestEnabled(!digestEnabled)}
+            switchLabel={
+              <span className={`h4 ${!digestEnabled ? "text-muted" : ""}`}>
+                <span>Enable digest</span>
+              </span>
+            }
```
MonkeyDo, line 115: You could use utility CSS classes (`ms-2` and `me-2` for example, numbers to be chosen appropriately) to add a margin before and after instead of using spaces.

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
```diff
@@ -0,0 +1,163 @@
...
+                min={1}
+                max={100}
+                defaultValue={digestAge}
+                style={{ maxWidth: "4em" }}
+                onChange={(e) => {
+                  setDigestAge(Number(e.target.value));
+                }}
+              />{" "}
+              days
+            </div>
```
MonkeyDo, line 133: That last condition might need to check that email and digests are enabled and not empty, rather than just not empty. Otherwise, I have a scenario for you:

1. enable everything
2. remove number from input
3. save button becomes disabled
4. disable emails or email digests
5. save button is still disabled despite the number thing being not-applicable

Maybe needs an `isValid` const which calculates the required scenarios?
1. email disabled
2. email enabled and digest disabled
3. email and digest both enabled and digestAge is a valid number

frontend/js/src/settings/routes/index.tsx
```diff
@@ -122,6 +122,14 @@ const getSettingsRoutes = (): RouteObject[] => {
           // Keep the /settings/import/ route for LastFM/LibreFM historical links,
           // and redirect to the music services page that replace those manual importers
           element: <Navigate to="../music-services/details/" replace />,
+          },
+        {
+          path: "notification/",
+          lazy: {
+            Component: async () => {
+              return (await import("../notification-settings/NotificationSettings")).default;
+            },
```
MonkeyDo: This is where the react-query loader would be added.

MonkeyDo (changes requested):
The UI is looking good! A couple of changes for some edge cases, and some compliance to the codebase style that will simplify a couple of things.

frontend/js/src/settings/layout.tsx
MonkeyDo: @fettuccinae It should be "notifications" with an 's' ;)

fettuccinae: Oops! fixed it now.

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
fettuccinae: ah, good catch!

Change I made: 
```
  const isDigestAgeValid =
    !notificationsEnabled ||
    !digestEnabled ||
    (notificationsEnabled && digestEnabled && digestAge);
```
We check if digestAge is valid only when both email and digest are enabled, else we dont check if its a valid number.

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
fettuccinae: https://github.com/metabrainz/listenbrainz-server/blob/1235c47ed65db9d4ba93fd06f1b0989bb5213af6/frontend/js/src/settings/brainzplayer/BrainzPlayerSettings.tsx#L194

I thought it was a custom defined css class.

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
fettuccinae: https://github.com/fettuccinae/listenbrainz-server/blob/ec7ad2fd81291e687cb0c2543aa6287f9659cc32/frontend/js/src/settings/notification-settings/NotificationSettings.tsx#L63

Should we show the complete error in a toast message or show the abstracted version of error?

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
MonkeyDo: My bad, sorry! Just saw it is used everywhere, but not defined.

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
MonkeyDo: IMO I think it would be useful to have more details than just 'something went wrong' so users can report issues they encounter. So `error.toString()` might give you a good text version, but you might need to try to trigger an error and see the result.


EDIT: Just seeing that you throw the error above with the statuscode. I think that would be a good start.
Might be good to try and get an errormessage from the API reponse, but not sure what shape that will take: if the API returns a json object with the message in it, you'll need to `await response.json()`. Otherwise probably `await response.text()`

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
MonkeyDo: Adding the server side of this example for reference:
https://github.com/metabrainz/listenbrainz-server/blob/1235c47ed65db9d4ba93fd06f1b0989bb5213af6/listenbrainz/webserver/views/settings.py#L192-L254

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
fettuccinae: https://github.com/metabrainz/metabrainz.org/blob/16057b95f7892f621aa8ba7bc6dec5c3eb2408bb/metabrainz/errors.py#L33-L37

API responds with a json object. I'll throw the response.error instead

frontend/js/src/settings/layout.tsx
```diff
@@ -27,6 +27,10 @@ const sections: Section[] = [
     title: "Account",
     links: [
       { to: "./", label: "User settings" },
+      {
+        to: "notifications/",
+        label: "Notification settings",
```
MonkeyDo:
```suggestion
        label: "Notifications",
```

fettuccinae:
> 1. There is front-end code in #3345 that needs to be removed, as it will be replaced by the code in this PR

  I'll remove front end bits from #3345. 

> 2. The back-end needs a new POST endpoint that will return the user's notifications settings saved in the database, at POST `/settings/notifications/` (what `fetchDigestSettings` used to do). Currently the queryloader automatically tries to hit that endpoint, which results in an error:

 Yes, I was waiting for this PR to be approved so I could update the endpoint names accordingly.

frontend/js/src/settings/notification-settings/NotificationSettings.tsx
fettuccinae: No problem!

MonkeyDo:
> Yes, I was waiting for this PR to be approved so I could update the endpoint names accordingly.

OK, thanks for the update!
For what it's worth, you will need two POST endpoints. The one that already exists in 3345 for updating the notification settings is not the same as the one you need for the routeloader

EDIT: I see you are already on that :+1:

fettuccinae:
@MonkeyDo Done.
Small question: why do we use POST to fetch data instead of GET in route loader?  My understanding was that to use GET to read and POST to write .

MonkeyDo:
> @MonkeyDo Done. Small question: why do we use POST to fetch data instead of GET in route loader? My understanding was that to use GET to read and POST to write .

Yes indeed you are right, and we know it looks wrong.
However at the moment we haven't found a good way to set up our single-page application for use with flask other than:
1. hit a GET endpoint to render and send the HTML page template and JS bundles
2. react gets activated on the page, front-end routing takes over 
3. hit a POST endpoint of the same name as the current route to fetch data for that component

And example of that here:

https://github.com/metabrainz/listenbrainz-server/blob/a1fe029d454a8c4e56f102ac57cbd5968cff7d2f/listenbrainz/webserver/views/donors.py#L15-L22

MonkeyDo:
I am hitting an error when I try to load the page, something to do with Oauth grant type.
Would appreciate some help figuring out what I have that's not correctly set up, or alternatively if you are hitting the same issue.

```
172.20.0.1 - - [30/Sep/2025 12:38:27] "POST /settings/notifications/ HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 1536, in __call__
    return self.wsgi_app(environ, start_response)
           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 1514, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 1511, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 919, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.13/site-packages/flask_debugtoolbar/__init__.py", line 222, in dispatch_request
    return view_func(**view_args)
  File "/code/listenbrainz/listenbrainz/webserver/login/__init__.py", line 77, in decorated
    return f(*args, **kwargs)
  File "/code/listenbrainz/listenbrainz/webserver/views/settings.py", line 383, in get_digest_setting
    data = get_digest_preference(user["musicbrainz_row_id"])
  File "/code/listenbrainz/listenbrainz/domain/metabrainz_notifications.py", line 94, in get_digest_preference
    token = _fetch_token()
  File "/code/listenbrainz/listenbrainz/domain/metabrainz_notifications.py", line 144, in _fetch_token
    token = oauth.fetch_token(token_url=token_url, auth=HTTPBasicAuth(client_id, client_secret))
  File "/usr/local/lib/python3.13/site-packages/requests_oauthlib/oauth2_session.py", line 406, in fetch_token
    self._client.parse_request_body_response(r.text, scope=self.scope)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/clients/base.py", line 426, in parse_request_body_response
    self.token = parse_token_response(body, scope=scope)
                 ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 444, in parse_token_response
    validate_token_parameters(params)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 451, in validate_token_parameters
    raise_from_error(params.get('error'), params)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/errors.py", line 398, in raise_from_error
    raise cls(**kwargs)
    
oauthlib.oauth2.rfc6749.errors.UnsupportedGrantTypeError: (unsupported_grant_type) 
```

fettuccinae:
> I am hitting an error when I try to load the page, something to do with Oauth grant type. Would appreciate some help figuring out what I have that's not correctly set up, or alternatively if you are hitting the same issue.
> 
> ```
> 172.20.0.1 - - [30/Sep/2025 12:38:27] "POST /settings/notifications/ HTTP/1.1" 500 -
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 1536, in __call__
>     return self.wsgi_app(environ, start_response)
>            ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 1514, in wsgi_app
>     response = self.handle_exception(e)
>   File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 1511, in wsgi_app
>     response = self.full_dispatch_request()
>   File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 919, in full_dispatch_request
>     rv = self.handle_user_exception(e)
>   File "/usr/local/lib/python3.13/site-packages/flask/app.py", line 917, in full_dispatch_request
>     rv = self.dispatch_request()
>   File "/usr/local/lib/python3.13/site-packages/flask_debugtoolbar/__init__.py", line 222, in dispatch_request
>     return view_func(**view_args)
>   File "/code/listenbrainz/listenbrainz/webserver/login/__init__.py", line 77, in decorated
>     return f(*args, **kwargs)
>   File "/code/listenbrainz/listenbrainz/webserver/views/settings.py", line 383, in get_digest_setting
>     data = get_digest_preference(user["musicbrainz_row_id"])
>   File "/code/listenbrainz/listenbrainz/domain/metabrainz_notifications.py", line 94, in get_digest_preference
>     token = _fetch_token()
>   File "/code/listenbrainz/listenbrainz/domain/metabrainz_notifications.py", line 144, in _fetch_token
>     token = oauth.fetch_token(token_url=token_url, auth=HTTPBasicAuth(client_id, client_secret))
>   File "/usr/local/lib/python3.13/site-packages/requests_oauthlib/oauth2_session.py", line 406, in fetch_token
>     self._client.parse_request_body_response(r.text, scope=self.scope)
>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/local/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/clients/base.py", line 426, in parse_request_body_response
>     self.token = parse_token_response(body, scope=scope)
>                  ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
>   File "/usr/local/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 444, in parse_token_response
>     validate_token_parameters(params)
>     ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
>   File "/usr/local/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 451, in validate_token_parameters
>     raise_from_error(params.get('error'), params)
>     ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/local/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/errors.py", line 398, in raise_from_error
>     raise cls(**kwargs)
>     
> oauthlib.oauth2.rfc6749.errors.UnsupportedGrantTypeError: (unsupported_grant_type) 
> ```

The `get_digest_preference `function [here](https://github.com/metabrainz/listenbrainz-server/blob/d80f0d294fba02cb5853c9ebaca1f42a8b950954/listenbrainz/webserver/views/settings.py#L383), is trying to generate an access token from MeB.org but the metabrainz-notification branch havent been merged there yet and hence the error. Can you please clone the metabrainz-notification branch locally and try running both MeB and LB containers on the same docker network or you can just hard code the data returned by get_digest_preference if thats alright.

fettuccinae:
> > @MonkeyDo Done. Small question: why do we use POST to fetch data instead of GET in route loader? My understanding was that to use GET to read and POST to write .
> 
> Yes indeed you are right, and we know it looks wrong. However at the moment we haven't found a good way to set up our single-page application for use with flask other than:
> 
>     1. hit a GET endpoint to render and send the HTML page template and JS bundles
> 
>     2. react gets activated on the page, front-end routing takes over
> 
>     3. hit a POST endpoint of the same name as the current route to fetch data for that component
> 
> 
> And example of that here:
> 
> https://github.com/metabrainz/listenbrainz-server/blob/a1fe029d454a8c4e56f102ac57cbd5968cff7d2f/listenbrainz/webserver/views/donors.py#L15-L22

Makes sense. But could we just not make another GET endpoint with sole purpose of fetching

> Yes indeed you are right, and we know it looks wrong. However at the moment we haven't found a good way to set up our single-page application for use with flask other than:
> 
>     1. hit a GET endpoint to render and send the HTML page template and JS bundles
> 
>     2. react gets activated on the page, front-end routing takes over
> 
>     3. hit a POST endpoint of the same name as the current route to fetch data for that component
> 
> 
> And example of that here:
> 
> https://github.com/metabrainz/listenbrainz-server/blob/a1fe029d454a8c4e56f102ac57cbd5968cff7d2f/listenbrainz/webserver/views/donors.py#L15-L22

Makes sense!

fettuccinae:
> > @MonkeyDo Done. Small question: why do we use POST to fetch data instead of GET in route loader? My understanding was that to use GET to read and POST to write .
> 
> Yes indeed you are right, and we know it looks wrong. However at the moment we haven't found a good way to set up our single-page application for use with flask other than:
> 
>     1. hit a GET endpoint to render and send the HTML page template and JS bundles
> 
>     2. react gets activated on the page, front-end routing takes over
> 
>     3. hit a POST endpoint of the same name as the current route to fetch data for that component
> 
> 
> And example of that here:
> 
> https://github.com/metabrainz/listenbrainz-server/blob/a1fe029d454a8c4e56f102ac57cbd5968cff7d2f/listenbrainz/webserver/views/donors.py#L15-L22

Makes sense!

MonkeyDo:
> Makes sense. But could we just not make another GET endpoint with sole purpose of fetching

It would be possible, but then we would have to have a separate endpoint and hardcode that endpoint for each route, which we wanted to avoid.
Another option would be to use the same suffix for every route, say for example "-data": if you hit GET /user/$username/ to get the HTML page, the RouteLoader would hit GET /user/$username-data/ 
Another option yet, and probably the way to go, would be to use a custom request header to signify we want data for the route, not the HTML page.

MonkeyDo:
> The `get_digest_preference `function [here](https://github.com/metabrainz/listenbrainz-server/blob/d80f0d294fba02cb5853c9ebaca1f42a8b950954/listenbrainz/webserver/views/settings.py#L383), is trying to generate an access token from MeB.org but the metabrainz-notification branch havent been merged there yet and hence the error. Can you please clone the metabrainz-notification branch locally and try running both MeB and LB containers on the same docker network or you can just hard code the data returned by get_digest_preference if thats alright.

Thanks, I hardcoded the return value, I'm only looking at checking the UI one last time

fettuccinae:
> > Makes sense. But could we just not make another GET endpoint with sole purpose of fetching
> 
> It would be possible, but then we would have to have a separate endpoint and hardcode that endpoint for each route, which we wanted to avoid. Another option would be to use the same suffix for every route, say for example "-data": if you hit GET /user/$username/ to get the HTML page, the RouteLoader would hit GET /user/$username-data/ Another option yet, and probably the way to go, would be to use a custom request header to signify we want data for the route, not the HTML page.

Ohh, got it!  Thanks for the detailed explanation.

PR #3348, opened 2025-08-21
Encode usernames in all URIs

MonkeyDo:
Reported in support email, a user can't access their pages unless they encode their username, which contains a question mark. This question mark breaks all constructed URIs as the question mark is not encoded and is considered as the start of query string arguments (i.e the end of the URI).

Didn't know the usernames on MusicBrainz were so permissive, so LB was not set up to encode special characters in all URIs.

It's a single change, but in  a lot of places...


frontend/js/src/playlists/Playlist.tsx
```diff
@@ -426,7 +426,11 @@ export default function PlaylistPage() {
                     With the help of:&ensp;
                     {customFields.collaborators.map((collaborator, index) => (
                       <React.Fragment key={collaborator}>
-                        <Link to={sanitizeUrl(`/user/${collaborator}/`)}>
+                        <Link
+                          to={sanitizeUrl(
+                            `/user/${encodeURIComponent(collaborator)}/`
+                          )}
+                        >
                           {collaborator}
```
anshg1214, line 438: I'm not sure if we need `sanitizeUrl` here?

frontend/js/src/user/year-in-music/2022/YearInMusic2022.tsx
```diff
@@ -278,7 +278,11 @@ export default class YearInMusic extends React.Component<
         <div className="center-p">
           <object
             style={{ maxWidth: "100%" }}
-            data={`${APIService.APIBaseURI}/art/year-in-music/2022/${user.name}?image=${coverArtKey}`}
+            data={`${
+              APIService.APIBaseURI
+            }/art/year-in-music/2022/${encodeURIComponent(
+              user.name
+            )}?image=${coverArtKey}`}
```
anshg1214, line 285: Instead of calling this function at multiple places, can we just encode the username once?

frontend/js/src/user/year-in-music/2023/YearInMusic2023.tsx
```diff
@@ -303,7 +303,11 @@ export default class YearInMusic extends React.Component<
         <div className="center-p heading">
           <object
             className="img-header"
-            data={`${APIService.APIBaseURI}/art/year-in-music/2023/${user.name}?image=${coverArtKey}&branding=False`}
+            data={`${
+              APIService.APIBaseURI
+            }/art/year-in-music/2023/${encodeURIComponent(
+              user.name
+            )}?image=${coverArtKey}&branding=False`}
```
anshg1214: Same here

frontend/js/src/playlists/Playlist.tsx
MonkeyDo: I think collaborator is a username, so probably need it if we want to navigate to their dashboard.

EDIT: Just tested with adding the user who reported the issue, and indeed the link to their profile is malformed if not encoded

PR #3351, opened 2025-08-28
Funkwhale player not submitting listens to LB

mAmanullah7:
`datasourceRecordsListens()` method return `true` which causing  BP to skip listen submission!

`datasourceRecordsListens()`  should return `false`

this will probably fix this issue!
  
 ```typescript
 datasourceRecordsListens = (): boolean => {
    return false; // record listens
  };
 ```


mAmanullah7:
@MonkeyDo i've fixed this reported on mb forum! you can have look! and tests are passing locally!

MonkeyDo (approved):
Yep, didn't realize while reviewing the PR...

Maybe this is a good occasion to rename this function so that it's purpose is clearer when reading without context. I'll think about that.

Thanks for the quick fix!

PR #3352, opened 2025-08-28
Interpolate SASS variables when used as CSS variables

MonkeyDo:
CSS custom properties are evaluated at runtime, which means our compiled CSS files end up with malformed CSS custom properties when using a SASS variable (which is evaluated at compile-time).
We need to interpolate the SASS variables by wrapping them with `#{ }`
See https://sass-lang.com/documentation/style-rules/declarations/#custom-properties

Ran into this issue when looking at the donations page, one of the flairs (sliced) was making the username completely transparent as a result:
image
Looking into it, this was the content of the *compiled* CSS file (not the $variables that should have been compiled from SASS into their value). 
Another issue was the incorrect compound selector that resulted in the class `.flaira` instead of the `a.flair` we wanted to achieve.
image

In this case, they need to be explicitly interpolated otherwise they will be ignored at compile time, and will be invalid at runtime.

Result after compilation:
image

### DIFF
--- a/frontend/css/sass/flairs.scss
+++ b/frontend/css/sass/flairs.scss
@@ -17,7 +17,9 @@
     overwrite styles for hover, visited, hover+visited
   */
   --text-color: #{$body-color};
-  &a {
+  // & char must appear at beginning of selector, so can't do `&a` which compiles to .flaira instead of a.flair
+  // Using at-root to escape nesting, then interpolating ampersand (#{&}) to end up with the correct structure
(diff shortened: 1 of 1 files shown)


MonkeyDo:
While I am here, I will make a note of some css file imports that are currently broken.
The files are required at runtime (and are missing) instead of being bundled witht eh rest of the css at compile time.

anshg1214 (approved):

PR #3357, opened 2025-09-02
Importer: Filter out invalid and skipped listens

Suvid-Singhal:
# Problem

The skipped and invalid listens were also imported earlier by the importer. This PR fixes the same.

# Solution

Added a few filters to filter invalid listens in the code.


listenbrainz/background/listens_importer/spotify.py
```diff
@@ -30,6 +30,11 @@ def _parse_listen_batch(self, batch: list[dict[str, Any]]) -> list[dict[str, Any
         items = []
         for item in batch:
             try:
+                if item["skipped"] or item["incognito_mode"]:
+                    continue
+                if item["ms_played"] < 30000:
+                    if item["reason_end"] in ["fwdbtn", "clickrow", "playbtn", "remote"]:
```
MonkeyDo: Found a couple more reasons that I saw related to "skipped" tracks:

```suggestion
                    if item["reason_end"] in ["fwdbtn", "backbtn", "clickrow", "clickside", "playbtn", "remote", "logout", "popup", "trackerror", "unexpected-exit", "unexpected-exit-while-paused"]:
```

I will note here I also saw a few more reason_end cases which could be of interest, but requires more analysis.
Here are all the ones I found (note the first one is empty string):
```
[
        "",
        "backbtn",
        "clickrow",
        "clickside",
        "endplay",
        "fwdbtn",
        "logout",
        "playbtn",
        "popup",
        "remote",
        "trackdone",
        "trackerror",
        "unexpected-exit",
        "unexpected-exit-while-paused",
        "unknown",
        "uriopen"
    ]
```

MonkeyDo:
Sorry about the force-push, forgot to anonymize  some data :see_no_evil:

kellnerd:
All `"reason_end": null` listens which I've found were recorded on Android, so I didn't fully censor that field just in case:

```json
   {
        "ts": "2017-12-04T20:50:34Z",
        "username": "MyUsername",
        "platform": "Android",
        "ms_played": 24437,
        "conn_country": "DE",
        "ip_addr_decrypted": null,
        "user_agent_decrypted": null,
        "master_metadata_track_name": "I Knew You Were Trouble",
        "master_metadata_album_artist_name": "Taylor Swift",
        "master_metadata_album_album_name": "Red",
        "spotify_track_uri": "spotify:track:6RjKVXi31fui9m4dbiGRrL",
        "episode_name": null,
        "episode_show_name": null,
        "spotify_episode_uri": null,
        "reason_start": "clickrow",
        "reason_end": null,
        "shuffle": false,
        "skipped": null,
        "offline": false,
        "offline_timestamp": 1512420608296,
        "incognito_mode": false
    },
```

kellnerd (commented):
The big advantage of LB server (over external importer tools) is that we have access to the track duration from the Spotify metadata cache (in the processing step below), which is missing from the Spotify export files.
This means we are actually able to compare `ms_played` against `duration_ms` and check the 50% played (or 4 minutes) condition.
If we implement that, we can relax the current filtering conditions a bit to only skip listens where it makes absolutely no sense to query the metadata cache for the total duration.

listenbrainz/background/listens_importer/spotify.py
```diff
@@ -30,6 +35,12 @@ def _parse_listen_batch(self, batch: list[dict[str, Any]]) -> list[dict[str, Any
         items = []
         for item in batch:
             try:
+                if (
+                    item["skipped"] or item["incognito_mode"] or
+                    (item["ms_played"] < 30000 and item["reason_end"] in SKIP_REASONS)
```
phw: I'm not sure I agree with excluding anything skipped. This isn't what the Spotify live integration does, and is also not what any other scrobble app would do. If the song was listened to long enough it would count as a listen.

Same for excluding everything under 30 seconds.

kellnerd: I agree. Out of my 69 listens with `skipped: true` only a single one would qualify for the 50% threshold, but they exist. I can provide a sample if needed.

MonkeyDo: Yes, fair point, thanks for catching that!

MonkeyDo: > Same for excluding everything under 30 seconds. 

WRT the spotify importer, I think following their own 30sec threshold is probably better than the (50% or 4min) threshold, at least until we can customize the filtering further.
As far as I know, a spotify listen is only added to your spotify history (the one we query with the live integration) after 30 second playback, not depending on percentage of track played.

MonkeyDo (approved):
Thanks for the help from everybody!
Group fix means better fix :)

PR #3360, opened 2025-09-03
Fix API call in InternetArchivePlayer

MonkeyDo:
Currently the Internet Archive player is calling `listenbrainz.org/1/...` which results in an HTML error page (410 gone).
Use the APIBaseURI instead, otherwise this only works in local development.

Also some automated formatting.

### DIFF
--- a/frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
+++ b/frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
@@ -1,5 +1,11 @@
 import * as React from "react";
-import { render, screen, act, waitFor, fireEvent } from "@testing-library/react";
+import {
+  render,
+  screen,
+  act,
+  waitFor,
+  fireEvent,
+} from "@testing-library/react";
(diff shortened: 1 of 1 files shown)


frontend/js/tests/common/brainzplayer/InternetArchivePlayer.test.tsx
```diff
@@ -238,7 +253,7 @@ describe("InternetArchivePlayer", () => {
 
       await waitFor(() => {
         expect(fetchMock).toHaveBeenCalledWith(
-          "/1/internet_archive/search?track=Los+Norte%C3%B1os&artist=P%C3%A9rez+Prado+y+Orquesta+con+Hermanas+Montoya"
+          "foo/1/internet_archive/search?track=Los+Norte%C3%B1os&artist=P%C3%A9rez+Prado+y+Orquesta+con+Hermanas+Montoya"
```
MonkeyDo, line 256: This is the only real change in this test file, the rest are automatic formatting changes.

PR #3361, opened 2025-09-03
Update AddData.tsx with MPV plugin

madeindjs:
Add a link to the MPV player, and my own ListenBrainz plugin.

### DIFF
--- a/frontend/js/src/about/add-data/AddData.tsx
+++ b/frontend/js/src/about/add-data/AddData.tsx
@@ -91,7 +91,7 @@ export default function AddData() {
             <a href="https://mpv.io/">MPV</a>
           </em>
           , cross-platform multimedia player:{" "}
-          <a href="https://github.com/madeindjs/dotfiles/blob/62aa5fe49a23188f74c07454afff8a7866357d88/mpv/.config/mpv/scripts/listenbrainz.js">
+          <a href="https://gist.github.com/madeindjs/f33225cf4d8fdc9f61e0fe3ebe2dcce7">
             <code>MPV Listenbrainz plugin</code>
           </a>
         </li>
--- a/frontend/js/src/about/add-data/AddData.tsx
+++ b/frontend/js/src/about/add-data/AddData.tsx
@@ -88,6 +88,15 @@ export default function AddData() {
         </li>
         <li>
           <em>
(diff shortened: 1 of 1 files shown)


frontend/js/src/about/add-data/AddData.tsx
```diff
@@ -86,6 +86,15 @@ export default function AddData() {
             <code>wylt</code>
           </a>
         </li>
+        <li>
+          <em>
+            <a href="https://mpv.io/">MPV</a>
+          </em>
+          , cross-platform multimedia player:{" "}
+          <a href="https://github.com/madeindjs/dotfiles/blob/62aa5fe49a23188f74c07454afff8a7866357d88/mpv/.config/mpv/scripts/listenbrainz.js">
```
MonkeyDo: I am a little bit worried that since this is a permalink to the current version, any improvements in the future would either mean we need to update the link, or that users will be using the old version.

On the other hand, permalinks are an assurance that you will find said file in this exact spot even if it is moved in the future.

Personally, I would rather have a link to the main branch, and updating the link if and when the file gets moved.

Another option would be to have a wiki page with some installation/configuration instructions, and have a link to the JS file on that wiki page itself that way you can control the link on your side.

Happy to hear your thoughts :)

madeindjs: Good point!

I also expose this script on [this GitHub Gist](https://gist.github.com/madeindjs/f33225cf4d8fdc9f61e0fe3ebe2dcce7). I guess it offers the best of both worlds: a unique URL and the latest version. 

```suggestion
          <a href="https://gist.github.com/madeindjs/f33225cf4d8fdc9f61e0fe3ebe2dcce7">
```

What do you think?

MonkeyDo: Yes, I think that might be the best bet in these circumstances, thanks!

If you have a couple more minutes, I think adding some comments at the top with some installation instructions could be handy for users (even if just pointing to the relevant MPV documentation).
I don't use MPV myself but I was surprised to see that you could make plugins in JS :)

MonkeyDo (approved):
Thanks for making another cool plugin, and for taking the time to come add it here :)

PR #3363, opened 2025-09-04
LB-1671:  Created for you: Add a note for recent imports

MonkeyDo:
We have been getting the same question over and over in the Matrix channel, so clearly it is not clear enough to newly arrived users, and needs a bit more clarity.
The existing modal "How and when are statistics calculated?" isn't particularly clear at a glance either, so it doesn't help much in this respect.

Before:
image

After:
image

### DIFF
--- a/frontend/js/src/user/recommendations/RecommendationsPage.tsx
+++ b/frontend/js/src/user/recommendations/RecommendationsPage.tsx
@@ -347,6 +347,19 @@ export default function RecommendationsPage() {
             Oh no. Either something’s gone wrong, or you need to submit more
             listens before we can prepare delicious fresh produce just for you.
           </p>
+          <p>
+            If you recently imported your listens, it may take up to a week.
+            <br />
+            We generate weekly playlists on monday mornings (
+            <a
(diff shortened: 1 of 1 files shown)


MonkeyDo:
@Aerozol If you have a minute to review the image above, do you think it will be sufficient as a stopgap measure until we have a way to generate this data for users after an import?

Aerozol:
Heya, it looks fine, though maybe it would be a good idea to cut the text out of the image completely?

And if "more details" goes to the same link as "(i) how and when..." I would just have the one.

MonkeyDo:
The link is directly to the [docs page about update intervals](https://listenbrainz.readthedocs.io/en/latest/general/data-update-intervals.html), which is kind of buried in the modal that opens when you click the "How and when..." link (obviously it was not fulfilling the need for clarity on that topic since users keep asking).

> maybe it would be a good idea to cut the text out of the image completely?

Not opposed to that, I guess we will generally want to rephrase the whole thing to include the various scenarios.
I will merge this PR to deploy it, and we can reassess if/how the text in the image should be changed (probably removed from the image so we can edit the text more easily).

PR #3367, opened 2025-09-04
Improve Artist Evolution graph interpolation

granth23:
# Problem

Improved the Artist Evolution Statistics frontend visual

image

# Solution

Fixed the ordering and offsetType

image

# Action

Please review the changes to ensure they align with the project's style guide.

### DIFF
--- a/frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
+++ b/frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
@@ -158,6 +158,7 @@ const transformArtistEvolutionActivityData = (
   const topArtists = Object.entries(artistTotals)
     .sort(([, a], [, b]) => b - a)
     .slice(0, topN)
+    .reverse()
     .map(([name]) => name);
 
   let orderedTimeUnits: string[] = [];
@@ -406,7 +407,8 @@ export default function ArtistEvolutionActivityStreamGraph(
                   enableGridX
                   enableGridY
                   offsetType="none"
-                  order="none"
+                  order="ascending"
+                  curve="basis"
                   colors={{ scheme: "nivo" }}
                   fillOpacity={0.85}
                   borderColor={{ theme: "background" }}


frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -405,7 +405,8 @@ export default function ArtistEvolutionActivityStreamGraph(
                   axisLeft={{ tickSize: 5, tickPadding: 5, tickRotation: 0 }}
                   enableGridX
                   enableGridY
-                  offsetType="diverging"
+                  offsetType="none"
+                  order="none"
```
MonkeyDo: A user's comment made me think that using "ascending" order would make more sense, ending up with the artist # 1 at the top. 

```suggestion
                  order="ascending"
```

This will also require a change at line 161 where you defined the `topArtists` array:  the addition of `.reverse()` so that the legend also follows the same logic (# 1 artist at the top)

```
const topArtists = Object.entries(artistTotals)
    .sort(([, a], [, b]) => b - a)
    .slice(0, topN)
    .reverse()
    .map(([name]) => name);
```

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -405,7 +405,8 @@ export default function ArtistEvolutionActivityStreamGraph(
...
                   colors={{ scheme: "nivo" }}
```
MonkeyDo, line 412: I would also suggest using this curve interpolation, after trying the different options:

```suggestion
                  curve="basis"
                  colors={{ scheme: "nivo" }}
```

granth23: fixed

Aerozol:
Looks less cool, but definitely more useful!

I can open another ticket for this, but what would be particularly useful is being able to hover over a colour and see what artist it is. I see a mouseover which says 'no data available' (when I test) that I suspect should already be doing this?

PR #3368, opened 2025-09-04
Add libre.fm archive importer

amCap1712:
Implement importer for libre.fm export files. See the test data file in the PR for example format.

### DIFF
--- a/listenbrainz/background/listens_importer/base.py
+++ b/listenbrainz/background/listens_importer/base.py
@@ -33,7 +33,15 @@ class BaseListensImporter(ABC):
         self.importer_name = IMPORTER_NAME
 
     def import_listens(self, user_id: int, import_task: dict[str, Any]) -> None:
-        """Main entry point for importing listens."""
+        """Main entry point for importing listens.
+
+        Retrieves import task from database, processes the import file, parses the items into
+        listens, validates them and submits them to the database.
+
+        Args:
+            user_id: the user whose listens are to be submitted
+            import_task: the import task dict
+        """
         user = db_user.get(self.db_conn, user_id)
         if user is None:
             current_app.logger.error("User with id: %s does not exist, skipping import.", user_id)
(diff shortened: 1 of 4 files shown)


frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -436,7 +443,7 @@ export default function ImportListens() {
                   id="file-upload"
                   className="form-control"
                   name="file"
-                  accept=".zip"
+                  accept=".zip,.csv"
```
MonkeyDo: During development of the spotify/LB importer, i had suggested that we could make this field dynamic depending on the selected service, in an effort to guide users a bit more.
Do you think that makes sense?

Would probably mean defining a dict above with file types for each service, as well as a bunch of other changes since currently we don't store the selected service in state (using FormData instead of state values).

Happy to do it for you if you would like.

listenbrainz/testdata/librefm.csv
MonkeyDo, line 1: I don't have an export myself (and actually don;'t think I have any listens in my librefm account), but is this really all the information available for each track?
We can't do much more than what I see in the code  I guess.

MonkeyDo (commented):
Front-end parts look good.
I also glanced at the python code but will leave proper reviewing for mayhem.

frontend/js/src/settings/import/ImportListens.tsx
MonkeyDo: Well... I always love playing with typescript enums and it satisfies my craving for organization and terseness, so I went ahead and implemented it, I hope you don't mind...

amCap1712: No worries, thanks!

listenbrainz/testdata/librefm.csv
amCap1712: Yes, we've been in touch with the libre.fm founder/maintainer over support for a while now and this is all the information that is available.

PR #3371, opened 2025-09-09
Add robots.txt

mayhem:
As discussed in IRC, we've needed a robots.txt for the companies that still respect them.

Only the robots.txt stuff is new, the other changes are just formatting cleanup.

### DIFF
--- a/listenbrainz/webserver/views/index.py
+++ b/listenbrainz/webserver/views/index.py
@@ -58,7 +58,7 @@ Allow: /explore/art-creator/
 """
 
 
-@index_bp.get("/robots.txt")
+@index_bp.get("/robots.txt/")
 def robots_txt():
     return Response(ROBOTS_TXT_CONTENT, mimetype='text/plain')
 


listenbrainz/webserver/views/index.py
```diff
@@ -37,6 +37,31 @@
...
+Disallow: /artist/
+Disallow: /album/
+Disallow: /release/
+Disallow: /release-group/
+Disallow: /recording/
+Disallow: /track/
+Allow: /explore/fresh-releases/
+Allow: /explore/huesound/
+Allow: /explore/cover-art-collage/
+Allow: /explore/art-creator/
```
amCap1712, line 57: Note that this does not forbid LB radio URLs.

amCap1712:
I think it would makes sense to add another robots.txt for api.listenbrainz.org, which forbids indexing all urls. The flask endpoint would need to detect whether its running under the `api` subdomain or not for it work properly. Also, the gateways will need to be updated to serve `api.listenbrainz.org/robots.txt` as currently all urls not starting with `/1` are blocked on api subdomains.

No need to block the PR on these changes though. We can do it later.

listenbrainz/webserver/views/index.py
mayhem: But there is:

Disallow: /explore/

Which disallows all explore including lb-radio. The Allow statements allow specific exceptions...

amCap1712: Sorry, I meant the LB radio api endpoint.

PR #3372, opened 2025-09-10
Refactor listencard

MonkeyDo:
While investigating LB-1643, I realized our stopgap solution of using a functional component wrapper (to be able to use react hooks) on top of a class component means that after manually mapping a listen the cover art does not update.
The cover art cache is set up in the wrapper, and no information is brought back to the wrapper from the class component after matching.

So it is finally time to rewrite the ListenCard into a functional component.

For LB-1643, I also added an extra step of fetching from the LB metadata cache API so that for the user's visual feedback the listen appears completely matched (instead for example of missing artist link).
Perhaps we could instead modify the recording-search API response to return artist MBID(s)?

### DIFF
--- a/frontend/js/src/common/listens/ListenCard.tsx
+++ b/frontend/js/src/common/listens/ListenCard.tsx
@@ -347,29 +347,31 @@ export default function ListenCard(props: ListenCardProps) {
       <div className="main-content">
         {beforeThumbnailContent}
         {thumbnail}
-        {listenDetails || (
-          <div className="listen-details">
-            <div className="title-duration">
-              <div
-                title={trackName ?? releaseName}
-                className={compact ? "ellipsis" : "ellipsis-2-lines"}
-              >
-                {trackName
-                  ? getTrackLink(displayListen)
-                  : getAlbumLink(displayListen)}
-              </div>
-              {trackDurationMs && (
-                <div className="small text-muted" title="Duration">
-                  {isNumber(trackDurationMs) &&
-                    millisecondsToStr(trackDurationMs)}
+        <div className="listen-details">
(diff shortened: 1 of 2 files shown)


frontend/js/src/common/listens/ListenCard.tsx
```diff
@@ -96,706 +94,615 @@ export type ListenCardProps = {
...
+        compact ? " compact" : ""
+      }${additionalContent ? " has-additional-content" : " "} ${
+        className || ""
+      }`}
+      data-testid="listen"
+    >
+      <div className="main-content">
+        {beforeThumbnailContent}
+        {thumbnail}
+        {listenDetails || (
```
anshg1214: You need to revert this change and add a parent div with class `listen-details`.

The top-album and top-track stats break because of this.

anshg1214: image

frontend/js/src/common/listens/ListenCard.tsx
```diff
@@ -96,706 +94,615 @@ export type ListenCardProps = {
...
+  }, [dispatch, displayListen]);
+
+  const openMBIDMappingModal = React.useCallback(async () => {
+    try {
+      const linkedTrackMetadata: TrackMetadata = await NiceModal.show(
+        MBIDMappingModal,
+        {
+          listenToMap: displayListen,
+        }
       );
```
anshg1214, line 244: We replace this modal by bootstrap modal in a followup PR

anshg1214 (changes requested):
Just a small change. Otherwise looks good

PR #3373, opened 2025-09-12
LB-1829: Reorganize stats page and show cover art grid

MonkeyDo:
With more stats added this month, I think it is a good time to implement some improvements I have been thinking about, namely adding a cover art grid to show the top albums, and adding links(buttons) to the art creator page, preloading a specific template adapted to the entity type.

image

### DIFF
--- a/docs/developers/devel-env.rst
+++ b/docs/developers/devel-env.rst
@@ -118,7 +118,7 @@ Update the ``LASTFM_API_KEY`` field with your Last.fm API key.
 You also need to update the ``API_URL`` field value to ``http://localhost:8100``.
 
 To use the Spotify importer you need to register an application on the
-`Spotify Developer Dashboard`_. Use ``http://localhost:8100/settings/music-services/spotify/callback/``
+`Spotify Developer Dashboard`_. Use ``http://127.0.0.1:8100/settings/music-services/spotify/callback/``
 as the callback URL.
 
 After that, fill out the Spotify client ID and client secret in the following
--- a/frontend/css/sass/playlists.scss
+++ b/frontend/css/sass/playlists.scss
@@ -9,6 +9,10 @@ $long-description-lines: 8;
   gap: 1rem;
 }
 
+.playlist-info .description {
+  white-space: pre-wrap;
+}
+
 #playlists-container {
   display: flex;
   flex-wrap: wrap;
--- a/frontend/js/src/playlists/Playlist.tsx
+++ b/frontend/js/src/playlists/Playlist.tsx
@@ -470,7 +470,7 @@ export default function PlaylistPage() {
             )}
           </div>
           {playlist.annotation && (
-            <div>
+            <div className="description">
               <div
                 className={`${isLongDescription ? "text-summary long" : ""} ${
                   showMore ? "expanded" : ""
--- a/frontend/js/src/user/stats/components/UserListeningActivity.tsx
+++ b/frontend/js/src/user/stats/components/UserListeningActivity.tsx
@@ -346,7 +346,7 @@ export default function UserListeningActivity(
   ): UserListeningActivityData => {
     const { dateFormat } = rangeMap.all_time;
     let totalListensForAllTime = 0;
-    let totalYears = 0;
+    let totalActiveYears = 0;
     const allTimeData = [];
     const currYear = new Date().getFullYear();
     let encounteredNonEmptyYear: boolean = false;
@@ -355,9 +355,9 @@ export default function UserListeningActivity(
       const yearData = unprocessedData.payload.listening_activity.filter(
         (year) => year.time_range === String(i)
       )[0];
-      totalYears += 1;
 
       if (yearData) {
+        totalActiveYears += 1;
         if (encounteredNonEmptyYear === false) {
           if (yearData.listen_count > 0) {
             encounteredNonEmptyYear = true;
@@ -388,7 +388,7 @@ export default function UserListeningActivity(
 
     setAvgListens(
       totalListensForAllTime > 0
-        ? Math.ceil(totalListensForAllTime / totalYears)
+        ? Math.ceil(totalListensForAllTime / totalActiveYears)
         : 0
     );
     setTotalListens(totalListensForAllTime);
--- a/frontend/js/tests/user/stats/UserListeningActivity.test.tsx
+++ b/frontend/js/tests/user/stats/UserListeningActivity.test.tsx
@@ -213,7 +213,7 @@ describe.each([
 
     // Expect the total listens and average listens per year to be displayed
     expect(screen.getByText("3,845")).toBeInTheDocument();
-    expect(screen.getByText("161")).toBeInTheDocument();
+    expect(screen.getByText("1,282")).toBeInTheDocument();
   });
 
   it("displays error message when API call fails", async () => {
--- a/listenbrainz/background/delete.py
+++ b/listenbrainz/background/delete.py
@@ -6,6 +6,7 @@ from data.model.external_service import ExternalServiceType
 from listenbrainz.listenstore.timescale_listenstore import REDIS_USER_LISTEN_COUNT
 from listenbrainz.webserver import timescale_connection
 from listenbrainz.db import user as db_user, listens_importer
+from listenbrainz.webserver.listens_cache import invalidate_user_listen_caches
 
 
 def delete_user(db_conn, user_id: int, created: datetime):
@@ -30,5 +31,6 @@ def delete_listens_history(db_conn, user_id: int, created: datetime):
     """
     timescale_connection._ts.delete(user_id, created)
     cache.delete(REDIS_USER_LISTEN_COUNT + str(user_id))
+    invalidate_user_listen_caches(user_id)
     listens_importer.update_latest_listened_at(db_conn, user_id, ExternalServiceType.LASTFM, 0)
     db_conn.commit()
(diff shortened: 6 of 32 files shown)


MonkeyDo:
@Aerozol kindly asking for your input on this, it's been a long time we've been talking about adding some cool cover art in the stats page, so here it is finally :)

Aerozol:
Very cool @MonkeyDo! I just recently made [a related ticket :0](https://tickets.metabrainz.org/browse/LB-1829)

I looove that this shows the album art grid, but imo it's a a regression in terms of being able to easily see my stats at a glance. I really like having the three 'top X' columns right next to each other. Having top artists and top tracks in such different configurations bugs me.

My suggestion would be to keep the Top X columns as-is, and instead make "Listening Activity" 2/3rds, and put the art grid next to it as 1/3rd.

MonkeyDo:
> My suggestion would be to keep the Top X columns as-is, and instead make "Listening Activity" 2/3rds, and put the art grid next to it as 1/3rd.

Summin' like this? 
image


Collapses in mobile sizes:
image

Aerozol:
[Yeah boi](https://www.youtube.com/watch?v=fvtQYsckLxk)

MonkeyDo:
> [Yeah boi](https://www.youtube.com/watch?v=fvtQYsckLxk)

OK, I also implemented the button you designed (minus slightly different icon...):
image

I think we will eventually want to add the same button for other top entities, but pointing to a specific template (some are artist-based, some album-based), which we can't currently do.
In the meantime, here it is introduced

Aerozol:
Looking great!! Because I always have to have something to say:

- I wonder if the second 'visualise and share' could just be the icon minus the text (cause it's repeating the first one). Maybe fine for now, but if we end up with heaps of those buttons then would be good
- Even if it's not really needed giving the cover art collage a title could balance out the whitespace at the top/balance it with the elements next to it

Hope I'm not bikeshedding too much!

MonkeyDo:
> Looking great!! Because I always have to have something to say:
> 
> * I wonder if the second 'visualise and share' could just be the icon minus the text (cause it's repeating the first one). Maybe fine for now, but if we end up with heaps of those buttons then would be good
> * Even if it's not really needed giving the cover art collage a title could balance out the whitespace at the top/balance it with the elements next to it
> 
> Hope I'm not bikeshedding too much!

Noooo! The feedback helps a lot! :smile: 
We might as well do it right.

I made the changes you described, and along with taking care of LB-1828 it makes the three top entity cards nice and clean.

I also took the time to make the changes to allow us linking to the art creator page with a specific time range and layout, and added the button on the top artists card (links to designer-top-10) and on the top albums (links to the grid template). That's a lot more useful and less confusing.

Pretty obviously I didn't put a button on the top tracks, because we don't have a template in the art creator that fits tracks... Something for us to work on adding, eh?

image

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -156,10 +158,27 @@ export default function ArtCreator() {
...
+      const rangeParam = params.get("range") as keyof typeof TimeRangeOptions;
+      if (rangeParam in TimeRangeOptions) {
+        setTimeRange(rangeParam);
+      }
+    }
+  }, []);
+
   const updateStyleCallback = useCallback(
     (event: React.ChangeEvent<HTMLSelectElement>) =>
-      setStyle(TemplateEnum[event.target.value as TemplateNameEnum]),
```
MonkeyDo, line 161: I realized while making those changes that we broke something at some point. Selecting fromt he gallery at the top sets the default grid layout and number of albums, but doing so from the dropdown select did not do the same

frontend/js/src/user/stats/components/UserArtistActivity.tsx
```diff
@@ -190,7 +192,12 @@ export default function UserArtistActivity(props: UserArtistActivityProps) {
                     )
                   )}
                   indexBy="label"
-                  margin={{ top: 20, right: 80, bottom: 100, left: 120 }}
+                  margin={{
+                    top: 20,
+                    right: isMobile ? 0 : 80,
+                    bottom: 100,
+                    left: isMobile ? 20 : 120,
+                  }}
```
MonkeyDo, line 200: This was not meant to make it into this PR, but I don't want tot open a new PR just for these few lines...
Sorry not sorry.

frontend/js/src/user/stats/UserReports.tsx
```diff
@@ -145,6 +183,15 @@ export default function UserReports() {
               entity="artist"
               user={user}
               terminology="artist"
+              extraButtons={[
+                <Link
+                  to={`/explore/art-creator/?range=${range}&style=designer-top-10`}
```
anshg1214: We should show the top artist artwork here.
```suggestion
                  to={`/explore/art-creator/?range=${range}&style=designer-top-5`}
```

frontend/js/src/user/stats/UserReports.tsx
```diff
@@ -134,7 +146,33 @@ export default function UserReports() {
...
+              />
+              <Link
+                to={`/explore/art-creator/?range=${range}&style=grid-stats-alt`}
+                className="mb-4 btn btn-info btn-small-rounding"
+              >
+                Visualize & share{" "}
+                <FontAwesomeIcon icon={faBorderAll} size="lg" />
+              </Link>
+            </Card>
+          </div>
```
anshg1214: The logic of displaying these stats is incorrect here.
1. If the user is logged out, and visits the stat's page, it shows an empty div cuz `encodedUserOrCurrentUserName` would be null.

2. if the user is logged in and on their own stats, when they go to the global stats, it still shows their own stat, which is not a good solution imo. It feels like something is broken.

3. If the user is on another user's profile, and you click the global. it should display global stats.

I think the best solution here is to either generate the global stats, or hide it completely

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -156,10 +158,27 @@ export default function ArtCreator() {
     },
     [setStyle]
   );
+
+  React.useEffect(() => {
+    // On load, check URL params for custom style and range
+    if (params.get("style")) {
```
anshg1214: If we're setting the initial settings from the search params, we should also update the search param when user switches to a different template style.

MonkeyDo: Fair, I just took the shortcut, but it's nicer now that it's properly implemented, with username as well so we can share URLs easier.

frontend/js/src/user/stats/UserReports.tsx
MonkeyDo: Thanks for spotting all this, I completely missed it  :P

Opted for hiding the album grid card when looking at global stats, because we don't have those set up for sitewide stats.

Also fixed the links to the ArtCreator page so they send you there with the right username.

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -139,27 +143,41 @@ export default function ArtCreator() {
...
+    }
+    if (searchParams.has("range")) {
+      const rangeParam = searchParams.get(
+        "range"
+      ) as keyof typeof TimeRangeOptions;
+      if (rangeParam in TimeRangeOptions) {
+        setTimeRange(rangeParam);
+      }
+    }
+  }, []);
```
anshg1214, line 192: Why do we need states now? We're having two sources of data.
1. Search Params
2. States

We can only use the search params as the source of truth everywhere.

Aerozol:
This just reminded me of something that's always bugged me and to finally make a ticket: https://tickets.metabrainz.org/browse/LB-1854

I am just looking at screenshots so I'm not sure if it is already applied here - if the user has listened to less than 11 (based on the screenshot) releases then it should default to a different grid type, or it will look broken.

If you are keen to merge then it could be left for later.

MonkeyDo:
> This just reminded me of something that's always bugged me and to finally make a ticket: https://tickets.metabrainz.org/browse/LB-1854
> 
> I am just looking at screenshots so I'm not sure if it is already applied here - if the user has listened to less than 11 (based on the screenshot) releases then it should default to a different grid type, or it will look broken.
> 
> If you are keen to merge then it could be left for later.

Yeah, that's a good ticket, we have the beginnings of a solution in place for playlist cover art, we limit the layout options based on the number of available cover art images:
image
image


Definitely for another PR, but it'll be interested to make an automatic system to decide on a layout.

Aerozol:
Cool - so to clarify, in the current implementation, if I have listened to less than 11 albums (based on the screenshot), will the collage have gaps?

(just wondering, don't let any of this hold up merging)

MonkeyDo:
> Cool - so to clarify, in the current implementation, if I have listened to less than 11 albums (based on the screenshot), will the collage have gaps?
> 
> (just wondering, don't let any of this hold up merging)

Yes, with fewer than 11 albums there will be gaps.

frontend/js/src/explore/art-creator/ArtCreator.tsx
MonkeyDo: I did those changes in cd8784b13 , although I will say IMO that's a lot of boilerplate needed compared to the previous implementation, for the same functionality (validate values on load, replace if invalid, update when changed).

Which option would you prefer?

EDIT: Looks like this is on other people's minds too: https://github.com/remix-run/react-router/discussions/13713

EDIT 2: Condensed the boilerplate a bit in 5eb8616

MonkeyDo:
> The PR Looks good. One thing, I noticed that sometimes the size of album collage is not correct.
> 
Thanks, I'll merge and deploy to beta, then see if I still get the issue.
Which browser and version was it on? Was it during local dev or deployed on test? (if the former, ignore the issue, some [changes to SVG templates](https://github.com/metabrainz/listenbrainz-server/pull/3373/files#diff-94269ada44a0eaf92eeca4d90be10bafd0e94dbe8622ef2b0fe297522b420d91R14) in this PR are missing in prod for sizing to work properly, could simply have been pointed to prod API)

PR #3377, opened 2025-10-07
Improved Navidrome Search for Tracks 

mAmanullah7:
[problem](https://github.com/metabrainz/listenbrainz-server/pull/3331#pullrequestreview-3303907204) mentioned by @MonkeyDo 

i've improved the searching of track as mentioned above and also in community chat! 
image
image

Now the The `searchForNavidromeTrack()` function now tries multiple approaches:

- Track + main artist (featuring artists removed)
- Track + full artist name
- Track name only 
- Artist only (fallback)

Screenshot 2025-10-07 at 11 29 41 PM
Screenshot 2025-10-07 at 11 32 15 PM

### DIFF
--- a/frontend/js/src/utils/utils.tsx
+++ b/frontend/js/src/utils/utils.tsx
@@ -357,7 +357,7 @@ const performNavidromeSearch = async (
 ): Promise<NavidromeTrack | null> => {
   const searchUrl = `${instanceURL}/rest/search3?query=${encodeURIComponent(
     query
-  )}&songCount=5&${authParams}`;
+  )}&songCount=1&${authParams}`;
 
   const response = await fetch(searchUrl);
 


mAmanullah7:
@MonkeyDo as you mentioned i tried this and its works you can have a look now, needed one suggestions like only track or artist name is not a good way i've just added like this can fail when we have multiple tracks with same name but if it does it will be taken care by first two search methods, should we keep this or we should remove if first two fails then let other priority take care of that! or should we keep so at least it should play from navidrome!

frontend/js/src/utils/utils.tsx
```diff
@@ -338,40 +393,63 @@ const searchForNavidromeTrack = async (
...
+      if (result) {
+        return result;
+      }
     }
-    const searchResult = data["subsonic-response"]?.searchResult3;
 
-    if (searchResult?.song && searchResult.song.length > 0) {
-      return searchResult.song[0];
+    // Artist name only
+    if (artistName && !trackName) {
```
MonkeyDo: In this particular case, I would not perform a search.
If we only have the artist name, we can't find that actual track, so I don't think this makes sense, as you mentioned in your comment.

frontend/js/src/utils/utils.tsx
```diff
@@ -338,40 +393,63 @@ const searchForNavidromeTrack = async (
...
 
-    let data;
-    const contentType = response.headers.get("content-type");
-    if (contentType && contentType.includes("application/json")) {
-      data = await response.json();
-    } else {
-      throw new Error("Server returned non-JSON response");
+    // Try track name only as last resort
+    // Only if we don't have artist name, to avoid matching wrong tracks
+    if (trackName && !artistName) {
```
MonkeyDo: I suppose we can see how well this works, but I don't have super high hopes for it to work well. Maybe this should be asked to the community, so that users can test themselves if they get good results with only trackname

frontend/js/src/utils/utils.tsx
```diff
@@ -326,6 +326,61 @@ const searchForFunkwhaleTrack = async (
...
+ * Remove featuring artists from artist name to improve search matching.
+ * Handles various formats: "feat.", "ft.", "featuring", "feat", "ft"
+ * Also removes content in parentheses at the end of artist names.
+ */
+const removeFeaturingArtists = (artistName: string): string => {
+  if (!artistName) return artistName;
+
+  let cleaned = artistName.replace(/\s+(?:feat\.?|ft\.?|featuring)\s+.+$/i, "");
+  cleaned = cleaned.replace(
+    /\s*\([^)]*(?:feat\.?|ft\.?|featuring)[^)]*\)\s*$/i,
```
MonkeyDo: This is quite a complex thing to do.
I would suggest basing your mechanism on the existing MusicBrainz utility which is much more battle-tested:
https://github.com/metabrainz/musicbrainz-server/blob/master/root/static/scripts/edit/utility/guessFeat.js

You will need to simplify it to fit this use-case, but at least the regexps should be more precise.

frontend/js/src/utils/utils.tsx
mAmanullah7: yes i'll remove this its worst we can do! like some music streaming platform use this for searching songs but those  platforms only have one source to play songs lol they don't have any options rather brute force to somehow find tracks thats y lb is here :) its better to switch to different datasource rather brute forcing to particular services.

frontend/js/src/utils/utils.tsx
mAmanullah7: it won't work it's failing(so i changes priority)  when same track name like initially this came to my mind similar to funkwhale how we do but later realised that fw dont allow to include artist name but navidrome search3 does then its waste to use only track name and then again filter to check artistname to match tracks, and even i think that it wont trigger as it will be taken care by

- Track + main artist (featuring artists removed)
- Track + full artist name

so its better to remove this also.

frontend/js/src/utils/utils.tsx
mAmanullah7: Sure i'll look into it!

mAmanullah7:
> OK, that's looking like a great improvement now, and I think it strike the right balance.
> 
> I would suggest one final change (that I could have thought about earlier...): doing the normal full search first, and falling back on the cleaned version, would mean that in most cases the regex is not going to be run (assuming most cases don't have a feat.) saving some resources.

Good one! Thanks :) this one gonna remove unnecessary complexity for each request!

frontend/js/src/utils/utils.tsx
```diff
@@ -326,6 +326,68 @@ const searchForFunkwhaleTrack = async (
...
+};
+
+const performNavidromeSearch = async (
+  instanceURL: string,
+  authParams: string,
+  query: string
+): Promise<NavidromeTrack | null> => {
+  const searchUrl = `${instanceURL}/rest/search3?query=${encodeURIComponent(
+    query
+  )}&songCount=5&${authParams}`;
```
MonkeyDo: Didn't notice this was changed, but I think we only need the one result?
```suggestion
  )}&songCount=1&${authParams}`;
```

MonkeyDo (approved):
Looking good! I will merge this and deploy to beta for further testing.

P.S: Thanks for working on this!
I know your GSOC project is long finished, and I appreciate the time you are putting to work with users and with me to improve the functionality further.

frontend/js/src/utils/utils.tsx
mAmanullah7: my bad forgot to change this one initially was checking what are the options we can do like similar to funkwhale.

PR #3379, opened 2025-10-08
Import XSPF playlist files

MonkeyDo:
A user on Mastodon was looking for a way to export their playlists, and someone recommended Soundiiz.
Since XSPF is a playlist interchange format (that we support exporting to), we should have a way to support importing XSPF playlists.

I refactored the existing playlist file import modal to support XSPF, parsing the file client-side with the [xspf-js library](https://github.com/hawyar/xspf-js) before sending a JSPF document to the server.

Unfortunately, currently upon playlist creation, every track's identifier field **must be a MusicBrainz recording URL**, making is less useful for the use-case described above (we can still import our own exported playlist xspf file).
We should think about allowing importing JSPF with identifiers other than MB recordings, especially if it is an identifier we can potentially use for metadata, for example a Spotify URL (https://soundiiz.com/tutorial/export-spotify-to-xspf).

Also added a checkbox to make the playlist private, as we need to set that to a value before sending the JSPF to the server, for some reason.
image

### DIFF
--- a/frontend/js/src/user/playlists/components/ImportJSPFPlaylistModal.tsx
+++ b/frontend/js/src/user/playlists/components/ImportJSPFPlaylistModal.tsx
@@ -1,19 +1,29 @@
 import * as React from "react";
 import NiceModal, { useModal, bootstrapDialog } from "@ebay/nice-modal-react";
-import { Modal } from "react-bootstrap";
+import { Button, Form, Modal } from "react-bootstrap";
 import { toast } from "react-toastify";
 
 import { Link } from "react-router";
+import { set } from "lodash";
 import GlobalAppContext from "../../../utils/GlobalAppContext";
 import { ToastMsg } from "../../../notifications/Notifications";
+import { MUSICBRAINZ_JSPF_PLAYLIST_EXTENSION } from "../../../playlists/utils";
+
+const acceptExtensions = ["jspf", "json", "xspf"];
+const acceptExtensionsWithDot = acceptExtensions
+  .map((ext) => `.${ext}`)
+  .join(", ");
 
 export default NiceModal.create(() => {
   const modal = useModal();
 
   const { currentUser, APIService } = React.useContext(GlobalAppContext);
 
-  const [fileError, setFileError] = React.useState<string | null>(null);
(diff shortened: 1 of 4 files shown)


mayhem:
You're doing it 'rong. Use the troi resolve playlist feature for this exact purpose.

MonkeyDo:
> You're doing it 'rong. Use the troi resolve playlist feature for this exact purpose.

OK, that was my hunch, but I need more guidance than that as I haven't used troi.

One issueI am facing is that sending a playlist exported from another service to the create_playlist endpoint [returns an error](https://github.com/metabrainz/listenbrainz-server/blob/866c872e1bf735b8679884b6c070e952080cea0a/listenbrainz/webserver/views/playlist_api.py#L103-L112) as the validation function expects MB recording URLs as each track's identifier.

So my options as far as I understand are: 
1. modify the validate_playlist function to allow for non-MB identifiers (maybe with a flag sent in the create_playlist endpoint request?), then use troi to resolve the tracks to MB recordings as part of the playlist creation process.
2. Create another endpoint that will resolve JSPF tracks to MB recordings, call that from the front-end then once resolved send the playlist to the create_playlist endpoint.

I think #1 makes more sense to me.

In any case, I would still appreciate some guidance in how to use troi.
In some cases, the existing identifiers from the exported playlists will contain URIs to music services that LB supports, so we could potentially use existing methods to resolve recordings from, say, [spotify URLs in the playlist](https://soundiiz.com/data/fileExamples/playlistExport.xspf).
I see for example I could pair Spotipy's `tracks` method to get the tracks from Spotify, then pass those to the `mbid_mapping_tracks` function.
Should these new functions (parse identifiers to match a music service, get tracks from music service) be part of a new patch? Or just functions exported from troi?

mayhem:
I think we should have a larger planning discussion on this feature. I *really* hope to return to the office next week -- can we do it then?

MonkeyDo:
> I think we should have a larger planning discussion on this feature. I _really_ hope to return to the office next week -- can we do it then?

Absolutely!
I agree this will be easier in person, there is so much I don't know about troi that being able to go through it all with you will be necessary.
No problem with waiting until you are feeling better!

PR #3395, opened 2025-11-04
Add webhook receiver for MeB OAuth

amCap1712:


amCap1712:
todo:

1. remove /login page only showing sign in with musicbrainz on LB side.
2. update sign in link in navbar.

frontend/js/src/home/Homepage.tsx
```diff
@@ -66,18 +67,24 @@ function HomePage() {
...
-      className="create-account-button"
-      href={`https://musicbrainz.org/register?returnto=${window.document.location.href}`}
-    >
+    <a className="create-account-button" href={buildAuthUrl("register")}>
       Create Account
     </a>
   );
   const loginButton = (
-    <Link className="login-button" to="/login/">
+    <a className="login-button" href={buildAuthUrl("login")}>
```
MonkeyDo, line 76: Making a note that #3398 also added a create account link in the navbar.

frontend/js/src/utils/ProtectedRoutes.tsx
```diff
@@ -1,19 +1,21 @@
...
-  const urlEncodedPathname = encodeURIComponent(pathname);
 
-  return currentUser?.name ? (
-    <Outlet />
-  ) : (
-    <Navigate to={`/login/?next=${urlEncodedPathname}`} replace />
-  );
+  if (!currentUser?.name) {
+    // Redirect to Flask login route (not React Router)
+    window.location.href = buildAuthUrl("login", pathname);
```
MonkeyDo, line 14: Does the path still need to be URLEncoded? If memory serves this was added to support exotic usernames

MonkeyDo (commented):
I've only properly looked at the front-end parts, but it looks fine at a glance.

How can I test this thoroughly? I'd like to click every login/logout link we have on the site...

frontend/js/src/login/Login.tsx
```diff
@@ -1,73 +0,0 @@
...
-          included in these data dumps.
-        </p>
-        <p>
-          Furthermore, you grant the MetaBrainz Foundation permission to process
-          your listening history and include it in new open source tools such as
-          recommendation engines that the ListenBrainz project is building. For
-          details on processing your listening history, please see our{" "}
-          <a href="https://metabrainz.org/gdpr">GDPR compliance statement</a>.
-        </p>
-        <p>
```
MonkeyDo, line 51: I feel like we should not entirely lose this, and should show this warning somewhere so users are informed. Not sure where would be appropriate though.
Maybe on the MEB website, if we know they are coming from LB, or if we have a generic message saying "your data in MetaBrainz projects will be publicly available"

listenbrainz/config.py.sample
MonkeyDo, line 103: I suppose this should be changed to meb.org once we are ready to deploy?

listenbrainz/webserver/views/import_listens.py
MonkeyDo: I think we don't need this `mb_engine` guard anymore (happens in a few files, next to `REJECT_LISTENS_WITHOUT_USER_EMAIL`), considering we don't use the MB database anymore for user account related stuff.

I think this might be causing the integration test failures related to rejection-without-email.

Also noting that we removed the mock previously in place to mock mb_engine [here](https://github.com/metabrainz/listenbrainz-server/blob/d42ecff4a7861119bc16811125d64cd7cc8116ef/listenbrainz/tests/integration/test_listens_importer.py#L120), which is likley why that test fails

Changed that in commit d9f3dce0cdad21421cf32975a7c25bcaefb713d8

PR #3469, opened 2025-12-23
LB-1884: Implement Navidrome 'Starred' tracks import

Harizz076:
# Problem

Right now you can scrobble from Navidrome just fine, but you have to manually keep your favorites in sync which is pretty annoying. This PR adds support for importing starred tracks from Navidrome into ListenBrainz. Solves LB-1884.

# Solution

Following the feedback from the initial review, I scrapped the old approach and completely rewrote the logic to properly integrate with the existing NavidromeService class so the connection and token handling flow isn't messed up anymore. 

Instead of dealing with XML, I switched the API call to request JSON from the Subsonic getStarred endpoint. Navidrome only gives us basic text metadata for tracks, so I hooked it up to the MBIDMapper to look up the correct MusicBrainz recording MBIDs before inserting anything into the feedback table. 

**Regarding Local Instances:** 

Since Navidrome URLs vary depending on where a user is hosting it, the import route now pulls the specific instance URL and credentials directly from the database for that user instead of relying on the frontend payload. 

I kept the import logic on the backend to avoid frontend CORS headaches. However, because this relies on the ListenBrainz backend *pulling* data inward to the user's home network, **this PR currently requires the user's Navidrome instance to be publicly accessible over the internet** (via port forwarding, reverse proxy, or a tunnel like pinggy or ngrok). 

Ultimately, for users with strictly local LAN IPs (like 192.168.x.x), the ideal solution would be for Navidrome to natively "push" feedback (starred tracks) to us in the background, exactly like they currently do for scrobbles. We shouldn't necessarily have to pull it from them. That being said, having this explicit import feature available here alongside Last.fm is still really nice as a double-feature for users who do have their servers exposed for the ListenBrainz web player!

image
image
image

I've also updated the frontend to expose this import option alongside Last.fm, added some proper error handling so connection timeouts return a 400 instead of 500s, and threw in some integration tests covering the happy path, missing credentials, and server errors. I tested everything locally against a fresh Navidrome container and it's mapping and importing cleanly. 

# AI usage

* [ ] I did not use any AI

> Used an AI assistant to help me rewrite the service integration after the initial review, and specifically to help write the integration tests. I made sure to manually test and understand all the changes.

#### If you did use AI:

# Action

Let me know if the MBID match thresholds look okay.


amCap1712 (changes requested):
Did you try out this yourself before opening the PR? The PR removes the code to connect Navidrome accounts which is obviously wrong. 

This seems like an AI generated PR, and one where you don't understand the code you have submitted. Please update this PR and explain yourself. Otherwise, I'll close it and won't review any future PRs you submit. Feel free to reach out in our Matrix channel if you have questions.

Harizz076:
I did use an AI assistant to help draft the new import logic. I was too eager to get the feature working and committed the changes without reviewing the diff closely enough, which caused the existing  NavidromeService class in navidrone.py   to be overwritten. I relied on a standalone mock script to verify the import logic, but I failed to verify the integration with the existing service or test the connection flow before opening the PR. I've definitely learned the hard way that I need to be much more careful and fully understand the code I'm submitting.

I have now manually restored the original  NavidromeService  implementation and integrated the new function correctly. I am currently in the process of manually testing the changes to ensure everything works as expected.

If this is still not satisfactory, I am happy to close this PR and start over from scratch. Going forward, I will avoid using AI tools for contributions, or if I do use them for assistance, I will explicitly disclose it in the PR description.

I sincerely apologize for wasting your time and for the confusion caused. Thanks again for your feedback. If any queries I will definitely reach out to you in the Matrix chanel.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -315,4 +315,18 @@ def import_feedback():
         counts = lastfm.import_feedback(user["id"], data["user_name"])
         return jsonify(counts)
 
+    if data["service"] == "navidrome":
+        if not all(k in data for k in ("navidrome_url", "auth_token", "salt", "user_name")):
+             raise APIBadRequest("Missing required Navidrome credentials (navidrome_url, auth_token, salt, user_name)")
+        
+        from listenbrainz.domain import navidrome
+        counts = navidrome.import_starred_tracks(
```
amCap1712: We'll need to think more about how to implement this. As in the current form, it won't support importing from local instances that are not exposed on public internet. We could implement it on the frontend but has its own issues and is more prone to breaking. 

Will need to discuss with others when they are back from holidays.

listenbrainz/domain/navidrome.py
```diff
@@ -135,4 +139,95 @@ def remove_user(self, user_id: int):
...
+    mapper = MBIDMapper()
+    
+    # Subsonic API parameters
+    params = {
+        'u': username,
+        't': auth_token,
+        's': salt,
+        'v': '1.16.1',
+        'c': 'ListenBrainz',
+        'f': 'xml'
```
amCap1712: Subsonic API supports json and should be used instead of xml.

listenbrainz/db/feedback.py
```diff
@@ -295,3 +298,25 @@ def get_feedback_for_multiple_recordings_for_user(db_conn, user_id: int, user_na
     query = query_base + query_remaining
     result = db_conn.execute(text(query), params)
     return [Feedback(user_id=user_id, user_name=user_name, **row) for row in result.mappings()]
+
+
+def bulk_insert_loved_tracks(db_conn, user_id: int, feedback: list[tuple[int, str]], column: str = 'recording_mbid'):
```
amCap1712, line 303: Instead of duplicating the queries, let's try to refactor the existing Lfm ones and reuse them for both.

listenbrainz/webserver/views/feedback_api.py
```diff
@@ -315,4 +315,18 @@ def import_feedback():
...
+            user["id"], 
+            data["navidrome_url"], 
+            data["auth_token"],
```
amCap1712: All of these values should be read from the database tables and not the request.

amCap1712 (commented):
Haven't tested locally yet but general comments. We'll need to discuss how/if to support local instances. Best to wait for now as that will also determine how we expose this to the user in the UI (on the import page or somewhere else?).

listenbrainz/webserver/views/feedback_api.py
MonkeyDo: For context, as far as I understand, currently you cannot connect to a local navidrome instance for anything anyway, as I think HTTPS is required for connection.
I get this error when I try:

>Unable to connect to server: HTTPConnectionPool(host='localhost', port=4533): Max retries exceeded with url: /rest/ping?u=monkey&t=fef21cac27dd52fb87117af05c8766aa&s=1768310588608&v=1.16.0&c=ListenBrainz&f=json (Caused by NewConnectionError("HTTPConnection(host='localhost', port=4533): Failed to establish a new connection: [Errno 111] Connection refused"))

amCap1712: How do you get that error? The error looks like a Python stacktrace to me. If the code is running inside docker then its expected, it won't be able to connect to a local navidrome instance.

The only way it could connect to a local instance is if we implement in it on the frontend.

MonkeyDo: I guess I was weighing in on "in the current form, it won't support importing from local instances that are not exposed on public internet"
As far as I understand there are currently no interactions with locally run (non internet exposed) servers, so it's in line with our current support, no?

mAmanullah7: Hey @MonkeyDo if you're using navidrome locally then`localhost` wont work afaik but navidrome exposes a `Network` address try using that instead it will work! 

```bash
4:12:01 PM JS.1 |    ➜  Local:   http://localhost:4533/
4:12:01 PM JS.1 |    ➜  Network: http://172.20.10.8:4533/
```

Harizz076:
@amCap1712: I have updated the PR description and made some changes, can you please review it? We can discuss more about this feature but I think ultimately, we need navidrome devs to implement this feature on their side if we want to support importing from local instances that are not exposed on public internet. 

There are ways we can do it on the frontend but i don't know if it will be reliable enough. That being said, I am open to trying out different approaches and reporting them here if you still want me to try them. Thanks!

arsinclair:
FYI, there's also a PR on Navidrome's side to add a bidirectional sync for ratings and loved tracks, but it seems it got a little stuck: https://github.com/navidrome/navidrome/pull/2058

Perhaps everyone will be interested to see if some motion is possible there.

listenbrainz/domain/navidrome.py
```diff
@@ -141,4 +144,95 @@ def remove_user(self, user_id: int):
...
+    total_found = len(songs)
+    total_mapped = 0
+    total_imported = 0
+    feedback_to_insert = []
+    
+    for song in songs:
+        artist = song.get("artist")
+        title = song.get("title")
+        album = song.get("album")
+        starred_date = song.get("starred")
```
RustyNova016, line 204: For info: Navidrome returns the musicbrainz id of tracks that are properly tagged with picard. If the track got the MBID field, you should use it instead of matching naively

Harizz076: Thank you for letting me know about this :)

PR #3498, opened 2026-01-12
feat: Added YouTube listens importer

nthsneha:
# This PR introduces an SQL schema update, don't forget to update the database and the background tasks container

This pull request adds support for importing listening history from YouTube Music, allowing users to upload their YouTube watch history exports and automatically filter music tracks from regular videos. The changes encompass backend, frontend and database layers to enable users to upload and process YouTube JSON exports, and include integration tests.

Jira Ticket: https://tickets.metabrainz.org/browse/LB-1044

### Backend: YouTube Music import support

- Added a new `YouTubeListensImporter` class in `listenbrainz/background/listens_importer/youtube.py` to parse YouTube watch history JSON exports and convert YouTube Music items to ListenBrainz listens. 

### Frontend: User interface updates

- Added YouTube as an option in the import UI (`frontend/js/src/settings/import/ImportListens.tsx`), including accepted file types and updated user instructions. 

### Database: Service type

- Extended the `user_data_import_service_type` enum to include `youtube` with migration script. 

### Testing and sample data

- Added an integration test to verify YouTube imports and a sample YouTube watch history JSON for test coverage. The sample file contains both regular YouTube videos (filtered out) and YouTube Music tracks (imported).

### Actions

I have gone through [CONTRIBUTING.md](https://github.com/metabrainz/listenbrainz-server/blob/master/.github/CONTRIBUTING.md) and tried my best to deliver accordingly. This is my first PR and I might have made some unnoticed slips here and there. Please do let me know if that's the case.


nthsneha:
@MonkeyDo Please take a look and let me know about the changes I should make.

listenbrainz/background/listens_importer/youtube.py
```diff
@@ -0,0 +1,115 @@
...
+                if not title:
+                    continue
+
+                channel_name = ""
+                subtitles = item.get("subtitles", [])
+                if subtitles and isinstance(subtitles, list):
+                    channel_name = subtitles[0].get("name", "")
+
+                track_name = title
+                artist_name = channel_name if channel_name else "Unknown Artist"
```
shirsakm: I think it would be better to skip the listen if the artist name is not specified. I believe that is how it's done in every other importer as well, either outright, or by the artist name being set to a null value, which gets dropped during validation.

The [relevant documentation](https://listenbrainz.readthedocs.io/en/latest/users/json.html#payload-json-details) also mentions that a minimal payload must include `artist_name` and while this does technically provide one, it kind of goes against the spirit of what it means.

listenbrainz/background/listens_importer/youtube.py
```diff
@@ -0,0 +1,115 @@
...
+
+                track_metadata: dict[str, Any] = {
+                    "artist_name": artist_name,
+                    "track_name": track_name,
+                }
+
+                additional_info: dict[str, Any] = {
+                    "submission_client": self.importer_name,
+                    "music_service": "youtube.com",
```
shirsakm: This value should probably be set to `music.youtube.com` no? I mean `youtube.com` is fine, but the importer does skip regular YouTube videos, and attempts to parse only if the header specifies YouTube Music.

[Here is the relevant documentation](https://listenbrainz.readthedocs.io/en/latest/users/json.html#id2).

listenbrainz/testdata/youtubemusic.json
shirsakm, line 1: Consider adding or modifying a test case where the `titleUrl` field is not in the `/watch?v=` format, given that you have added a whole function to handle a whole bunch of different formats. 

To be fair, I have never seen it in any other format in my own watch history files, but since you are handling a lot of formats, it would probably be a good idea to test them.

nthsneha:
@shirsakm 
Made the required changes. Also, for the format handling function, I believe it is a better idea to keep only one format so removed the helper function and refactored the code. Do take a look and let me know!

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,108 @@
...
+                video_url = item.get("titleUrl", "")
+                if video_url:
+                    additional_info["origin_url"] = video_url
+                    match = re.search(r'watch\?v=([^&]+)', video_url)
+                    if match:
+                        video_id=match.group(1)
+                    else: 
+                        continue
+                    if video_id:
+                        additional_info["youtube_id"] = video_id
```
fettuccinae: Do we need `youtube_id` if it already exists in `origin_url` ?

listenbrainz/background/listens_importer/youtube.py
```diff
@@ -0,0 +1,108 @@
...
+        youtube_music_items: list[dict[str, Any]] = []
+        for item in data:
+            # Only process YouTube Music items
+            if item.get("header") != "YouTube Music":
+                continue
+
+            try:
+                time_str = item.get("time", "")
+                timestamp = datetime.fromisoformat(time_str.replace('Z', '+00:00'))
+            except (KeyError, TypeError, ValueError):
```
fettuccinae: A small nitpick, I dont think we can get KeyError when accessing dict using `get()`.

frontend/js/src/settings/import/ImportListens.tsx
```diff
@@ -36,6 +36,7 @@ enum Services {
   librefm = "Libre.fm",
   panoscrobbler = "PanoScrobbler",
   maloja = "Maloja",
+  youtube = "Youtube",
```
fettuccinae: I think we should use `youtubemusic` instead of  `youtube` (Here and elsewhere) to avoid ambiguity for users who might  use youtube to listen to music (as we later discard non youtube music items ).

listenbrainz/background/listens_importer/youtubemusic.py
shirsakm: I am pretty sure there is no existing field known as `youtube_id`, it's probably best to only set `origin_url` as the full URL. But I assume `spotify_id` was added for a reason, this might be an acceptable addition as well then.

listenbrainz/background/listens_importer/youtube.py
```diff
@@ -0,0 +1,108 @@
...
+                }
+
+                video_url = item.get("titleUrl", "")
+                if video_url:
+                    additional_info["origin_url"] = video_url
+                    match = re.search(r'watch\?v=([^&]+)', video_url)
+                    if match:
+                        video_id=match.group(1)
+                    else: 
+                        continue
```
shirsakm: I might be mistaken here, but wouldn't this `continue` skip any entry where it can't find a proper URL? This should probably not be the case, as that isn't a mandatory field.

shirsakm:
@nthsneha I am not a maintainer, just someone with a lot of free time and some experience working on similar PRs. But with that warning out of the way, looks good to me! Good job. Just gotta wait now for Monkey or Lucifer to review it and give it the go ahead.

P.S. I am perpetually anxious this will get reviewed and the whole PR will be wrong and I will be called a moron for saying it looks good to me. XD

fettuccinae:
I’ve tested this PR locally and it looks good to me.

A quick question: should we sanitise `artist_name` and `title`? My data contains artist names like `TheWeekndVEVO` and `Drake - Topic`, which could be sanitised to `TheWeeknd` and `Drake`. Similarly, some titles are like `J. Cole - MIDDLE CHILD`, which could be converted to `MIDDLE CHILD`.

MonkeyDo:
> I’ve tested this PR locally and it looks good to me.
> 
> A quick question: should we sanitise `artist_name` and `title`? My data contains artist names like `TheWeekndVEVO` and `Drake - Topic`, which could be sanitised to `TheWeeknd` and `Drake`. Similarly, some titles are like `J. Cole - MIDDLE CHILD`, which could be converted to `MIDDLE CHILD`.

That sound like a job for the mapper, actually, which will be done automatically further down in the pipeline.

admin/sql/create_types.sql
MonkeyDo, line 1: You also need to create a separate schema update file that adds the new service type. See for reference:
https://github.com/metabrainz/listenbrainz-server/blob/master/admin/sql/updates/2025-11-12-add-panoscrobbler-data-import-type.sql

frontend/js/src/settings/import/ImportListens.tsx
MonkeyDo: I think that makes sense for how it's presented to the user ("Youtube Music" string).
Should the key name also be cahnged, to be something like this?

```suggestion
  youtubemusic = "Youtube Music",
```

listenbrainz/background/listens_importer/youtube.py
```diff
@@ -0,0 +1,108 @@
...
+from datetime import datetime, timezone
+from typing import Any, Iterator
+
+from flask import current_app
+from more_itertools import chunked
+
+from listenbrainz.background.listens_importer.base import BaseListensImporter
+
+
+class YouTubeListensImporter(BaseListensImporter):
```
MonkeyDo: Similarly to previous comment by fettuccinae, both the class name and the file name should specifiy Youtube *Music* just in case we implement something else for regular youtube in the future

listenbrainz/background/listens_importer/youtube.py
```diff
@@ -0,0 +1,108 @@
...
+
+                additional_info: dict[str, Any] = {
+                    "submission_client": self.importer_name,
+                    "music_service": "music.youtube.com",
+                }
+
+                video_url = item.get("titleUrl", "")
+                if video_url:
+                    additional_info["origin_url"] = video_url
+                    match = re.search(r'watch\?v=([^&]+)', video_url)
```
MonkeyDo: I think there might be other formats used than `watch?v=`, this might need more research.
A quick internet search for example yielded this:
https://gist.github.com/rodrigoborgesdeoliveira/987683cfbfcc8d800192da1e73adc486

There might be some more advanced regexps out there already.

listenbrainz/background/listens_importer/youtube.py
MonkeyDo: Agreed, especially since at that point we might already have a `video_url` which is probably sufficient in most cases.

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,108 @@
...
+                channel_name = ""
+                subtitles = item.get("subtitles", [])
+                if subtitles and isinstance(subtitles, list):
+                    channel_name = subtitles[0].get("name", "")
+
+                if not channel_name:
+                    continue
+
+                track_name = title
+                artist_name = channel_name
```
MonkeyDo: Can we do the same simple processing you did with the leading "Watched " in the title, and remove potential trailing " - Topic" in the channel name?

MonkeyDo (changes requested):
Thanks for opening a PR!
It's looking pretty good, with a few more comments, apart from what has been reporter by @shirsakm and @fettuccinae (thanks for the help reviewing!).

I am trying to get my hands on a YTMusic history export so I can test this locally (I have disabled it for my account...)

nthsneha:
@MonkeyDo I've made the requested changes. Please do check and let me know!

listenbrainz/background/listens_importer/youtube.py
shirsakm: I don't have any documentation on this, but I do have anecdotal evidence in the form of my Takeout with 10k entries. There doesn't seem to be anything other than the `/watch?v=` format either from `youtube.com` or from `music.youtube.com`.

But handling more formats is not really a bad thing, I think it would be fine to include this regex check provided in [this comment](https://gist.github.com/rodrigoborgesdeoliveira/987683cfbfcc8d800192da1e73adc486?permalink_comment_id=5567955#gistcomment-5567955) of the Gist you shared. I have very little experience with regex, so I don't know if it's efficient or not and if that matters, but that's just my two cents.

listenbrainz/tests/integration/test_listens_importer.py
```diff
@@ -858,6 +858,15 @@ def test_import_youtube(self):
         self.assert200(response)
         import_id = response.json["import_id"]
 
+        # Manually process the background task since the background processor isn't running in tests
+        from listenbrainz.background.background_tasks import BackgroundTasks, get_task, remove_task
+        processor = BackgroundTasks()
```
shirsakm: This seems like debug code to me that you wrote to get tests passing. This definitely shouldn't be included in the final PR.

I have never delved deep enough to have experience with these modules, but I assume it will be setup already as all other tests work the same.

shirsakm: I can confirm, the tests run just fine without the extra code.

listenbrainz/tests/integration/test_listens_importer.py
```diff
@@ -844,6 +844,57 @@ def test_import_maloja_empty(self):
...
+
+        url = self.custom_url_for("api_v1.get_listens", user_name=self.user["musicbrainz_id"])
+        response = self.wait_for_query_to_have_items(url, num_items=1, attempts=20)
+        listens = response.json["payload"]["listens"]
+        self.assertEqual(len(listens), 1)
+
+        listen = listens[0]
+        self.assertEqual(listen["listened_at"], 1639818816)
+        track_metadata = listen["track_metadata"]
+        self.assertEqual(track_metadata["artist_name"], "Rage Against the Machine - Topic")
```
shirsakm: This needs to be updated to `"Rage Against the Machine"`.

listenbrainz/tests/integration/test_listens_importer.py
```diff
@@ -844,6 +844,57 @@ def test_import_maloja_empty(self):
...
+        self.assertEqual(track_metadata["track_name"], "Killing In the Name")
+        additional_info = track_metadata["additional_info"]
+        self.assertEqual(additional_info["submission_client"], "YouTube Music History Importer")
+        self.assertEqual(additional_info["music_service"], "youtube.com")
```
shirsakm: This needs to be updated to `"music.youtube.com"`.

listenbrainz/background/listens_importer/youtubemusic.py
MonkeyDo: We do support it on the website, if it is there, so why not?

https://github.com/metabrainz/listenbrainz-server/blob/df486fcd8de17f488c577c51a340e1a447cdf914/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx#L54

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,119 @@
...
+                if not title:
+                    continue
+
+                channel_name = ""
+                subtitles = item.get("subtitles", [])
+                if subtitles and isinstance(subtitles, list):
+                    channel_name = subtitles[0].get("name", "")
+
+                if not channel_name:
+                    continue
```
MonkeyDo: Interestingly, looking at someone's history export file, there are cases where there is no information provided apart from the video URL. Here is such an example:
```json
{
  "header": "YouTube Music",
  "title": "Watched https://music.youtube.com/watch?v=qQ0zxuWFxrY",
  "titleUrl": "https://music.youtube.com/watch?v=qQ0zxuWFxrY",
  "time": "2025-06-03T06:03:45.443Z",
  "products": ["YouTube"],
  "activityControls": ["YouTube watch history"]
}
```

In such cases, we could potentially do API calls to get information back from Youtube's API, sending multiple video ids in a batch. How would you suggest we handle these cases and batch the lookups?

The API we would use:
https://developers.google.com/youtube/v3/docs/videos/list?apix_params=%7B%22part%22%3A%5B%22contentDetails%2CrecordingDetails%2Cplayer%2Csnippet%22%5D%2C%22id%22%3A%5B%22qQ0zxuWFxrY%22%5D%7D#parameters

See the response, where we can find a title and channelTitle, and even a duration:
```json
{
  "kind": "youtube#videoListResponse",
  "etag": "WcmtbaSiSpgpDoLIRubkjZHU83w",
  "items": [
    {
      "kind": "youtube#video",
      "etag": "W0DPshjjMCbqhIbK_72f7-4_2D0",
      "id": "qQ0zxuWFxrY",
      "snippet": {
        "publishedAt": "2019-10-03T10:08:31Z",
        "channelId": "UCApYZt0aLG30DTt0g4yKLmw",
        "title": "The Reason",
        "description": "Provided to YouTube by Universal Music Group\n\nThe Reason · Hoobastank\n\nThe Reason\n\n℗ 2003 UMG Recordings, Inc.\n\nReleased on: 2018-01-01\n\nProducer: Howard Benson\nRecording Engineer, Digital Editing Engineer: Mike Plotnikoff\nRecording Engineer: Eric Miller\nMixing Engineer: Chris Lord-Alge\nMixing Second Engineer: Dim -E\nMixing Second Engineer: Keith Armstrong\nStringarranger: Deborah Lurie\nEngineer: Casey Stone\nComposer Lyricist, Vocalist: Douglas Robb\nComposer Lyricist: Daniel Estrin\nComposer Lyricist: Markku Lappalainen\nComposer Lyricist: Chris Hesse\nDigital Editing Engineer: Vince Jones\nDigital Editing Engineer: Dan Adam\n\nAuto-generated by YouTube.",
        "thumbnails": {
                       [...omitted for brevity...]
        },
        "channelTitle": "Hoobastank - Topic",
        "tags": [
          "Hoobastank",
          "フーバスタンク",
          "The Reason",
          "ザリーズン",
          "ザ・リーズン"
        ],
        "categoryId": "10",
        "liveBroadcastContent": "none",
        "defaultLanguage": "en",
        "localized": {
          "title": "The Reason",
          "description": [...omitted for brevity...]
        },
        "defaultAudioLanguage": "en"
      },
      "contentDetails": {
        "duration": "PT3M53S",
        "dimension": "2d",
        "definition": "hd",
        "caption": "false",
        "licensedContent": true,
        "regionRestriction": {
          "allowed": [
            [...omitted for brevity...]
          ]
        },
        "contentRating": {},
        "projection": "rectangular"
      },
      "player": {
        "embedHtml": "\u003ciframe width=\"480\" height=\"360\" src=\"//www.youtube.com/embed/qQ0zxuWFxrY\" frameborder=\"0\" allow=\"accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share\" referrerpolicy=\"strict-origin-when-cross-origin\" allowfullscreen\u003e\u003c/iframe\u003e"
      },
      "recordingDetails": {}
    }
  ],
  "pageInfo": {
    "totalResults": 1,
    "resultsPerPage": 1
  }
}

```

listenbrainz/background/listens_importer/youtubemusic.py
shirsakm: In that case, I see no reason either to not include it. I didn't find it in the [JSON docs](https://listenbrainz.readthedocs.io/en/latest/users/json.html#id1) hence I figured it wasn't a preexisting field. My bad on that.

MonkeyDo (commented):
We're steadily moving forward :)

I procured a YT history export file and did some local testing, and successfully imported some listens, so it's already working nicely, good job!

listenbrainz/tests/integration/test_listens_importer.py
nthsneha: Okay got it!
Made the changes.

listenbrainz/background/listens_importer/youtubemusic.py
nthsneha: I say we get all such video IDs (ones with  only the URL) from every unique watchHistory.json uploaded. We then club them together and send them as a single API call, even if there is just one such video upload in the json. Better user experience that way? 
Also, should we cache each new look up?

MonkeyDo: Yes I think that sounds right.
Some details that need figuring out:
- is there a limit of number of videoIds we can send to the YT API? If so we need to batch them.
- when in the process does this happen? it might make sense to import everything we can first, and only then do the further API calls
- caching the lookup responses does sound reasonable, and we do have a cache for some other services (see [for spotify for example](https://github.com/metabrainz/listenbrainz-server/blob/d9570d66cfae7f98852d16012660080b5cc8a2d8/listenbrainz/background/listens_importer/spotify.py#L136-L137)). I don't know enough on the topic and would need to rope in @amCap1712

nthsneha: Right, so 
-  I tested it out and the limit seems to be 50. Any number above that, we'll batch.
- Yes, I agree. We should make the call after all possible imports.

amCap1712: I think a cache for youtube api responses makes sense, however youtube library is so wide that we cannot "crawl" it effectively and most of the content in there is not music either. So I think the Youtube cache would look very different from the existing caches. I think its fine to open a ticket to create a Youtube cache in the future for now. 

FWIW, the way I see the youtube cache working would be just caching the responses we lookup during imports and not actively search for tracks to cache metadata for in the background like we do for other caches.

admin/sql/updates/2026-01-16-add-youtubemusic-data-import-type.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE user_data_import_service_type ADD VALUE 'youtubemusic';
```
amCap1712: @MonkeyDo We'll need to coordinate the deployment of this PR due to the changes in the schema. Last time we forgot to update the background tasks container and that blocked user data exports too.

MonkeyDo: Yes, that was on my mind too. 
I'll add a note in the description PR in case I forget.

listenbrainz/background/listens_importer/youtubemusic.py
MonkeyDo: @nthsneha OK, so let's proceed like this:
- we store a list of entries that don't have a `channel_name` but otherwise pass validation/filtering
- after we finished processing the "normal" entries, we enter a post-processing step
- in that post processing, we batch entries in groups of 50 and do a YT API call (you'll find the token at `current_app.config["YOUTUBE_API_KEY"]`; you'll need to create a token for yourself for local dev)
- from the API response we still need to match to the history entries and get the missing info (`title` replaces the original one, `channelTitle` added), storing the result in a list, to be validated again
- finally return both the regular entries and our newly fixed entries


This will probably require some refactoring, considering:
1. you will need to do the same validation/string replacement to both regular and irregular entries, at different moments
2. parse_listen_batch is you single entry point, for clarity it would be best to separate regular processing and irregular entries processing

MonkeyDo:
I took the liberty to fix the merge conflicts that were created after merging #3452

nthsneha:
@MonkeyDo Okay! I've started working on the API integration, will push it tomorrow.

brend-an:
Unsure if this is the right forum, I understand it isn't a listenbrainz specific issue. I've been trying to find a way to import my YouTube watch history. However, when I use Google Takeout to request the `watch-history.html` or `watch-history.json` files, I get differing file sizes and incomplete watch history records when using the JSON option. It only shows about a year of records, whereas the HTML file shows back to 2017 (when I began using YouTube). Not having a way to import the HTML file means my listens will be incomplete.

shirsakm:
@brend-an You have just made me realize even my own `watch-history.json` is only from September, whereas the whole history goes as far back as January. I don't know what this means for the importer, but I think a standalone script for parsing the HTML and submitting listens might be better.

Either way, thank you for reporting this!

brend-an:
To anyone trying to import YouTube music listens in the meantime with their `watch-history.html` from Google Takeout, I had luck with [ytm-exporter](https://github.com/defvs/ytm-extractor) on the [ListenBrainz forum here](https://community.metabrainz.org/t/ytm-extractor-import-youtube-music-listens-to-listenbrainz/707619). Except, that the script can't differentiate between songs listened most of the way through and skipped tracks. It will submit absolutely everything creating tons of listens for songs you had skipped or only listened to a few seconds before switching to the next track.

brend-an:
I've used AI to help me create some python scripts to help detect skipped tracks and duplicate submitted tracks in both the YouTube Google Takeout `watch-history.html` HTML file for submission to ListenBrainz using ytm-exporter and Spotify's Extended Listening history JSON files for submission to ListenBrainz using the import tool.

Repo here: https://github.com/brend-an/prune-scrobbles/

yateendogra2109:
Hi @nthsneha  and @MonkeyDo  

I've been following this PR and I'd like to  work on the YouTube metadata cache [(LB-1929)](https://tickets.metabrainz.org/projects/LB/issues/LB-1929?filter=allopenissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC) feature 

As discussed by @amCap1712 earlier , the yt cache would save the api calls to youtube for the same video accessed multiple time.I would like to work on top of this branch , and then merge on this branch . 
Please let me know if that works.

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
+import json
+import re
+import requests
+from datetime import datetime, timezone
```
anshg1214: nit: timezone is imported but never used.

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+                            if snippet.get("title"):
+                                orig["title"] = snippet.get("title")
+                            if snippet.get("channelTitle"):
+                                orig["subtitles"] = [{"name": snippet.get("channelTitle")}]
+                            converted2 = self._convert_item_to_listen(orig)
+                            if converted2:
+                                listens.append(converted2)
+
+        return listens
+    def _convert_item_to_listen(self, item: dict[str, Any]) -> dict[str, Any] | None:
```
anshg1214: nit: add blank line before this function

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+        """Attempt to convert a single history item into a ListenBrainz listen.
+
+        Returns the listen dict on success or None if required metadata is
+        missing (title or channel).
+        """
+        try:
+            title = item.get("title", "")
+            if title.startswith("Watched "):
+                title = title[8:]
```
anshg1214:
```suggestion
                title = title.removeprefix("Watched ")
```

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+
+            if from_date <= timestamp <= to_date:
+                youtube_music_items.append(item)
+
+        if not youtube_music_items:
+            return iter(())
+
+        return chunked(youtube_music_items, self.batch_size)
+
+    def parse_listen_batch(self, batch: list[dict[str, Any]]) -> list[dict[str, Any]]:
```
anshg1214, line 62: Let's define proper types instead of using any.

```
class YouTubeSubtitle(TypedDict, total=False):
    name: str
    url: str

class YouTubeHistoryItem(TypedDict, total=False):
    header: str
    title: str
    titleUrl: str
    subtitles: list[YouTubeSubtitle]
    time: str
    products: list[str]
    activityControls: list[str]
````

File: listenbrainz/background/listens_importer/youtubemusic.py [hunk 19]
```diff
@@ -0,0 +1,189 @@
...
+            subtitles = item.get("subtitles", [])
+            channel_name = ""
+            if subtitles and isinstance(subtitles, list):
+                channel_name = subtitles[0].get("name", "")
+
+            if not channel_name:
+                return None
+
+            if channel_name.endswith(" - Topic"):
+                channel_name = channel_name[:-8]
```
anshg1214:
```suggestion
                channel_name = channel_name.removesuffix(" - Topic")
```

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+            listened_at = int(timestamp.timestamp())
+
+            return {"listened_at": listened_at, "track_metadata": track_metadata}
+
+        except (KeyError, TypeError, ValueError):
+            current_app.logger.error("Error parsing YouTube item: %s", item, exc_info=True)
+            return None
+
+    def _extract_video_id(self, video_url: str) -> str | None:
+        patterns = [
```
anshg1214: This can be a static method.

```suggestion
    @staticmethod
    def _extract_video_id(video_url: str) -> str | None:
        patterns = [
```

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+        api_key = current_app.config.get("YOUTUBE_API_KEY")
+        if missing and api_key:
+            id_map: dict[str, list[dict[str, Any]]] = {}
+            for orig, vid in missing:
+                if not vid:
+                    continue
+                id_map.setdefault(vid, []).append(orig)
+
+            if id_map:
+                for ids_chunk in chunked(list(id_map.keys()), 50):
```
anshg1214: chunked accepts any iterable

```suggestion
                for ids_chunk in chunked(id_map.keys(), 50):
```

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+            "id": ",".join(video_ids),
+            "key": api_key,
+        }
+
+        resp = requests.get(url, params=params, timeout=10)
+        resp.raise_for_status()
+        data = resp.json()
+        items = data.get("items", [])
+
+        return {item.get("id"): item for item in items}
```
anshg1214: If "id" is absent the dict gets a None key silently. 

```suggestion
        return {
            item["id"]: item
            for item in items
            if item.get("id") is not None
        }
```

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Returns a dict mapping video_id -> metadata (the API item object).
+        """
+        api_key = current_app.config.get("YOUTUBE_API_KEY")
+        if not api_key:
+            raise RuntimeError("YOUTUBE_API_KEY not configured")
+
+        url = "https://www.googleapis.com/youtube/v3/videos"
+        params = {
+            "part": "snippet,contentDetails,player",
```
anshg1214: Why are we requesting contentDetails and player if we're not using it's data?

```suggestion
            "part": "snippet",
```

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+        return listens
+    def _convert_item_to_listen(self, item: dict[str, Any]) -> dict[str, Any] | None:
+        """Attempt to convert a single history item into a ListenBrainz listen.
+
+        Returns the listen dict on success or None if required metadata is
+        missing (title or channel).
+        """
+        try:
+            title = item.get("title", "")
+            if title.startswith("Watched "):
```
anshg1214, line 126: Just a question, is this same for non-english users as well?

listenbrainz/background/listens_importer/youtubemusic.py
```diff
@@ -0,0 +1,189 @@
...
+                        item_meta = metas.get(vid)
+                        if not item_meta:
+                            continue
+                        snippet = item_meta.get("snippet", {})
+                        for orig in id_map.get(vid, []):
+                            if snippet.get("title"):
+                                orig["title"] = snippet.get("title")
+                            if snippet.get("channelTitle"):
+                                orig["subtitles"] = [{"name": snippet.get("channelTitle")}]
+                            converted2 = self._convert_item_to_listen(orig)
```
anshg1214: Let's name this variable as conveted_listen.

```suggestion
                            converted_listen = self._convert_item_to_listen(orig)
```

anshg1214 (changes requested):
Great work on the overall structure!

Let's focus on improving the code readability

yateendogra2109:
> Great work on the overall structure!
> 
> Let's focus on improving the code readability

Thanks for the feedback!
I will go through your suggestions and make changes soon.

MonkeyDo:
Found a few more details to adjust, and added instructions on how users can get their export.
I asked for help with testing on the LB chat, so that we can put it on test.LB and have users test it out live , but not on their main account.
After some manual confirmation (I couldn't find a sample history file online and don't have on myself) I'm happy to merge this, it's going to make some users happy :)

MonkeyDo:
For the record, there are still some annoying aspects to this, mainly that the prefixes ("Watched") and suffixes (" - Topic"), that we use for track title and artist name detection, are localized in the YT export format.
I've added a warning in the instructions for users to change their language to English before requesting the export;
It's far from ideal but that seems to be the best we can do for now as we are unlikely to implement language detection for all the languages YT supports...
One idea would be to request users to go watch a specific YT Music track before the export, so that we have a baseline from which to calculate the prefix and suffix, but that sounds mad.

MonkeyDo:
This was tested on the test server by a couple of people, and I think it is ready to merge

MonkeyDo:
Merging into feature branch, will look at the youtube cache PR next to round the feature before merging to master

PR #3548, opened 2026-02-01
LB-1848: Add support for using MB URL relationships in BrainzPlayer

shirsakm:
# Problem

This PR aims to add support for BrainzPlayer to check URL relationships already present in MB, and use them wherever possible. This should increase the quality of the music selected by BP for playing quite significantly.

This includes two features. Firstly, surfacing the data from the MB database to the `mapping.mb_metadata_cache` table in LB. Secondly, updating BP and the individual player implementations to use this new information. The most difficult part of this ticket was neither of these features, but instead setting up the MB database and getting the cache to build. :laughing: 

# Solution

Currently, the feature is only fully functional for YouTube, though I can't imagine it will take too much more work to get it up and running for the other data sources. I will go over the solution in two parts to hopefully make it a little easier to review the PR.

Firstly, the updates to the metadata cache, and how the data is being surfaced,
1. I have tacked on the URL relationships field to the already existing `recording_data` field similar to how the artist relationships are currently surfaced. `recording_data` was already of type JSONB, so no schema changes were required for the `mapping.mb_metadata_cache` table.
2. I haven't made any changes to the API endpoint (which was `/1/metadata/recording`) which already queries this JSON data from the Timescale database and surfaces it as a part of the `"recording"` field. If necessary, we could decouple the URL relationships data and make it a new endpoint, but this works for now.

On to the BrainzPlayer updates,
1. I have updated BP to go through the data sources the user has enabled and then find matches if present from the filtered URLs we find from the metadata. We then give priority to the URLs, and find the first data source which has an URL available for playing. I have also noticed a bug with the logic while going through the changes to create the PR, so that's nice.
2. I have updated the `playListen(...)` method in the `DataSoureType` type with a new argument for sending one particular streaming URL to the data source. This will need to be individually implemented for each player. Currently, everything except YouTube just ignores the argument. I am planning on implementing this for SoundCloud, Spotify and AppleMusic, exactly in that order. I don't know if this can be implemented for Funkwhale and Navidrome because they ...complicate things. Also, I figure no one is adding relationships to a Funkhwale / Navidrome instance anyways.
3. Small note on that I have added a set of 3 1000ms retries when checking if the YouTube player instance is available because somehow my changes makes the YouTube player initialize slower, and that breaks everything.
4. I have added an error toast for when we catch errors when attempting this MB URL fetch and more. I would have just let it fail silently, but based on recent talks with Monkey about giving the end user enough information, I think showing the error makes more sense. I mention this solely because I want to show how it looks, what it says, and if I should update it.
image

# Action

I primarily need help with the SQL side of things primarily by being told if I have either missed something, or added something that I shouldn't have. My approach to the SQL was mainly just search for the already existing `RECORDING_LINK_GIDS` and then update it to it's new name `RECORDING_ARTIST_LINK_GIDS` and add similar queries for the new `RECORDING_URL_LINK_GIDS` with minor changes based on the structure of the existing data, of course. This has been more successful than I would have imagined reasonable, but I have plausibly missed places where this data is used.

I also have always used the `build-mb-metadata-cache` command for my testing, and I feel `update-mb-metadata-cache` would be the command used in production, and maybe that requires some other changes, as well.


MonkeyDo:
Haven't tested any of this yet, but ran through the code quickly.
I wonder if it would make sense to attach the metadata directly in the listen metadata, in the mbid_mapping that we [already attach to listens](https://github.com/metabrainz/listenbrainz-server/blob/ce11bfe7124b2d723d74bb18d5960b3270a00ea4/listenbrainz/db/msid_mbid_mapping.py#L42C7-L66C27).

This would save a query for each time we want to listen to a track, which is good, and would add it everywhere at once, allowing for example to easily show links to externals services in track pages without fetching more information from MB's API.

On the other hand, it would attach that data **everywhere**, including in API responses where it might not be needed or wanted.

Perhaps this could be added in certain circumstances and not in others?

shirsakm:
> I wonder if it would make sense to attach the metadata directly in the listen metadata, in the mbid_mapping that we [already attach to listens](https://github.com/metabrainz/listenbrainz-server/blob/ce11bfe7124b2d723d74bb18d5960b3270a00ea4/listenbrainz/db/msid_mbid_mapping.py#L42-L71).

What I gather from this is that the `url_rels` values would be stored in the listen itself, similar to how the `origin_url` is currently stored and used to play if possible, correct? I am confirming this because admittedly, I can't make heads or tails of the file you attached as a reference.

If so, I think that's a sensible option. I already planned on having a chat about how we could cache this data, because it feels like something that should be cached. While this won't be an explicit cache, it is the same thing I think.

MonkeyDo:
Yeah, sorry, lots of shortcuts there...

Basically, as you may know we automatically do a best effort matching of incoming listens that don't have a MusicBrainz identifier, matching it to an MB recording.
This information is then attached to the listen before we send it down the wire (API or web views). It's not stored in the listen in the DB, just attached to the listen at fetching time.

So this whole block `mbid_mapping` you will find if you inspect a listen is not part of the listen sent to LB, but is added when the listen is fetched:
image

In comparison, the `origin_url` is part of the data sent to LB by the listen submitter.; it's user-sent data.

I am proposing to attach the URL relationships along with the rest of the mbid_mapping metadata, and like you do in the code already, prefer the user-sent `origin_url` and fallback to our best guess.

>I can't make heads or tails of the file you attached as a reference.

This is where the best match metadata is attached to the listens, and might be (?) the right place to attach the URL relationships as well.

> While this won't be an explicit cache, it is the same thing I think.

Yes, more or less, at least it serves the same purpose of limiting extra API requests and serving data from cached tables instead (this metadata is fetched from the `mapping.mb_metadata_cache` table).

shirsakm:
> [...] this whole block mbid_mapping you will find if you inspect a listen is not part of the listen sent to LB, but is added when the listen is fetched [...]

That makes a lot of sense, that's on me for assuming it was stored alongside the listen metadata.

> This is where the best match metadata is attached to the listens, and might be (?) the right place to attach the URL relationships as well.

That also makes a lot of sense, I will look into it, and try to figure out how that would work. Thanks!

MonkeyDo:
Don't hesitate if you have questions. We know this is a pretty involved feature :rocket:

MonkeyDo:
There are a couple of linting errors, please have a look at 
https://github.com/metabrainz/listenbrainz-server/actions/runs/22966778463/job/66672375834?pr=3548#step:4:2415

shirsakm:
@MonkeyDo Yep, fixed the linting errors. The other failing tests seem unrelated, though I am not entirely sure.

You should however, be able to test it for Spotify now, as well as review the code when you get the time. I am sure there's lots of changes that yet need to be made. :)

As a side, will you also be able to test Apple Music, or should we send a call for volunteers that would be willing to test the feature with Apple Music, perhaps? Or any other option really.

Or we could leave it out of this PR, though that doesn't seem particularly prudent, and isn't exactly what I wanna do if I can avoid it.

Anyways, looking forward to a whole lot of requested changes! Especially I would like to draw your attention to the used regexes (regexi?), because I learnt it to the best of my abilities for this PR itself, though most of it was a bunch of copying other people's homework. XD

docker/docker-compose.yml
MonkeyDo, line 1: Would be good to remove the unrelated changes here

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -519,14 +526,68 @@ export default function BrainzPlayer() {
...
+      if (listenedFromIndex !== -1) {
+        selectedDatasourceIndex = listenedFromIndex;
+      } else {
+
+        if (!urlsToUse) {
+          const urlRels = listen?.track_metadata?.mbid_mapping?.url_rels;
+
+          if (urlRels?.length) {
+            const filteredUrls = urlRels.filter(
+              (rel) => rel.type === "free streaming" || rel.type === "streaming"
```
MonkeyDo, line 546: We'll need to check if there are other relationship types that could contain links, I suspect there are some like "streaming page" for apple music that might contain the useful links.

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -519,14 +526,68 @@ export default function BrainzPlayer() {
...
+            }
+          );
+          if (streamingDatasourceIndex !== -1) {
+            selectedDatasourceIndex = streamingDatasourceIndex;
+          } else {
+            selectedDatasourceIndex = 0;
+          }
+        } else {
+          selectedDatasourceIndex = 0;
+        }
```
MonkeyDo, line 588: I think this needs to be its own utility function like we do for isListenFromDatasource, to keep this playListen function somewhat usable (it's already spaghetti...).

I think this would probably be the right point to separate into another PR, leaving the player's and datasources implementation details separate.

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -73,11 +73,17 @@ import useWindowTitle from "./hooks/useWindowTitle";
 import useCrossTabSync from "./hooks/useCrossTabSync";
 import useListenSubmission from "./hooks/useListenSubmission";
 
+const DATASOURCE_URL_MAP: Record<string, RegExp> = {
+  youtube: /(?:https?:\/\/(?:www\.|m\.)?)?(?:youtu\.be|youtube\.com)\//, // modified from https://gist.github.com/rodrigoborgesdeoliveira/987683cfbfcc8d800192da1e73adc486?permalink_comment_id=5567955#gistcomment-5567955
+  soundcloud: /soundcloud\.com/,
+  spotify: /spotify\.com/,
+};
```
MonkeyDo, line 80: I think these should live with the datasource component itself, like we do for its name and icon.
You'll be able to simply do something like `const urlPattern = datasource.urlPattern` or maybe even a static helper function like `isListenFromThisService`.

You'll find some likely [reusability](https://github.com/metabrainz/listenbrainz-server/blob/1e1889fefb445d08421bcddf920a0c19a7ebe2d5/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx#L82) in the components themselves, too.

I think this should be in a new PR though

MonkeyDo (changes requested):
Very promising!

What I think we should do is split the playback changes in one or more separate PRs, and keep in this one only the backend +  setting up music sources to receive the URLs, and make it easier to review the nitty gritty of the third-party implementations.

I'll be able to test apple music, I'm going to activate a one month trial to test this and a couple of other PRs

MonkeyDo:
I just thought of something that would be good to add along with the URL rels while you're in there: ISRCs
It could potentially be for a separate PR, considering it's not quite the same mechanism, but I know some services allow searching by ISRC so that could be another great way to avoid inaccurate text searches (looking at you, [Apple Music](https://developer.apple.com/documentation/applemusicapi/get-multiple-catalog-songs-by-isrc))

docker/docker-compose.yml
shirsakm: Hi, I was just about to revert it, but then I saw the changes and remembered why I kept it in to begin with. The indentations are kind of messed-up, and just uncommenting the lines leads to syntax errors. It's not a big deal to fix the indentation at all, but I figured it was fine to fix it permanently anyways. I probably should have noted that somewhere.

But yeah, if you think I should revert it anyway, that's also fine.

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
shirsakm: I did do a check before implementation, and at a cursory glance, those two seem to be the only ones that are relevant.

```
musicbrainz_db=# SELECT lt.name, COUNT(*)
FROM link_type lt
JOIN link l ON l.link_type = lt.id
JOIN l_recording_url lru ON lru.link = l.id
GROUP BY lt.name
ORDER BY count DESC;
         name          | count 
-----------------------+-------
 free streaming        | 58672
 streaming             | 35993
 purchase for download | 15019
 secondhandsongs       |  1030
 other databases       |   758
 download for free     |   212
 IMDB samples          |   139
 allmusic              |    61
 license               |    42
 crowdfunding          |     1
(10 rows)
```
Above is the snippet I ran, and none of the rest really seem to be things we could use.

If possible, running this on the actual instance might be helpful, as this is only the sample data. It might include the `streaming page` relationship or other such instances, as you mentioned. I am not very knowledgeable on the common link types we can have, I am afraid.

shirsakm:
> What I think we should do is split the playback changes in one or more separate PRs, and keep in this one only the backend + setting up music sources to receive the URLs, and make it easier to review the nitty gritty of the third-party implementations.

Right, that makes sense. I had misunderstood what you meant last time I read it, was quite confused. Well, I will make a copy of this branch, and just revert the player implementations and such.

> I just thought of something that would be good to add along with the URL rels while you're in there: ISRCs
> It could potentially be for a separate PR

I think I will leave that to a separate PR, mainly because I will have to take a look at if that is surfaced, and if not, where it is stored. Not difficult, just I think doing it separately makes more sense to me.

shirsakm:
@MonkeyDo Hi, I just took a look at the cache. The good news is, the ISRC data does seem to be there already, at least for the massive two entries in my database. We can just add it to the `mbid_mapping` block in the retrieval process then, if that suits the cause (I think it does). The bad news is, ...actually, I don't think there's any this time.

Other than that, the rest seems to be working just fine after reverting the changes. Probably good to go to have another look at. Thanks! :)

MonkeyDo (approved):
Looks good, thanks for this part 1 !

PR #3563, opened 2026-02-08
Fix LB-1920: Display only integer values on stats graph axes

Swaim-Sahay:
## Problem
Stats graphs were showing fractional values (like 0.5, 1.5) on their Y-axes which doesn't make sense for listen counts.

**Ticket:** LB-1920

## Solution
Updated graph components to display only whole numbers on axes. For small values (≤10 listens) graphs now show explicit integer ticks. For larger values non-integer ticks are hidden.

## Changes
- Modified 6 graph components to enforce integer only Y-axis values
- Added safe data handling to prevent errors with empty data
- All existing tests passing


Swaim-Sahay:
Hey @MonkeyDo  I've updated the PR according to your suggestions. Could you please review it?

frontend/js/src/user/stats/components/BarDualTone.tsx
```diff
@@ -195,7 +209,8 @@ export default function BarDualTone(props: BarDualToneProps) {
               : undefined,
         }}
         axisLeft={{
-          format: ".2~s",
```
MonkeyDo, line 198: I haven't tested this in situ on the graph, but part of me wonders if the whole tickValues thing is necessary, or if changing the format here to `".0~s"` would be enough.
This change of the format loses some crucial feature; it does solve for integers but if you have 30000 listens it will show 30000 instead of 30K, so I don't think it's a satisfactory solution. You could test this on the global stats for example:

Image


Testing with d3.format, which is used under the hood, directly in browser console on https://d3js.org/d3-format#locale_format :

Image

You might also try setting [tickFormat](https://d3js.org/d3-axis#axis_tickFormat)
Worth a try?

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -452,9 +457,8 @@ export function UserArtistEvolutionActivityStats(
   return (
     <Card className="user-stats-card" data-testid="artist-evolution">
       <div
-        className={`d-flex align-items-center justify-content-between ${
-          isMobile ? "mb-3" : ""
-        } flex-wrap mt-3`}
+        className={`d-flex align-items-center justify-content-between ${isMobile ? "mb-3" : ""
```
MonkeyDo: General comment: for the future, please do not commit these unrelated formatting changes.
I think you probably don't have the same automatic formatter installed in your IDE, and it makes the PRs harder to review.

MonkeyDo (changes requested):
I think this is a good start, but needs more testing of a variety of use cases, not just sub-integer scale issues.

MonkeyDo (changes requested):
I think this is a good start, but needs more testing of a variety of use cases, not just sub-integer scale issues.

MonkeyDo (commented):
I think this needs much more thorough local testing using data from the website.
Look for example at global stats y axis:

Image
In prod:

Image

So those changes that remove the `format` value and replace it with an empty string are wrong, at the very least.

I'd like to see more a more thought through plan showing that you understand what you are changing and why, because at the moment it looks like you're shooting in the dark and hoping to hit the target.
Identify the issue with precision and then evaluate your options to resolve it. I think you'll find you probably don't need stuff like `< 1000` to change the mechanism.
You can inspect API requests in the network tab of your browser devtools to get the data.
For example for the sitewide listening stats I show above:
https://api.listenbrainz.org/1/stats/sitewide/listening-activity?range=this_year

MonkeyDo (approved):
Thanks, this is looking good now!

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
```diff
@@ -1,4 +1,5 @@
 import { ResponsiveStream, TooltipProps } from "@nivo/stream";
+import { format as d3Format } from "d3-format";
```
MonkeyDo, line 2: Unused import

MonkeyDo (changes requested):
Forgot one bit :)

frontend/js/src/user/stats/components/UserArtistEvolutionActivity.tsx
MonkeyDo:
```suggestion
```

Swaim-Sahay:
hey @MonkeyDo  do i need to change/update anything now or is it looking ok?

MonkeyDo (changes requested):
It all looks good now, thank you!

MonkeyDo (approved):

PR #3575, opened 2026-02-12
feat: Add Genre Page

anshg1214:
Add a genre page: `/genre/rock`. 

Redirect `/genre/0e3fc579-2d24-4f20-9dae-736e1ec78798/` to `/genre/rock`.


frontend/js/src/genre/Genre.tsx
```diff
@@ -0,0 +1,396 @@
...
+import queryClient from "../utils/QueryClient";
+
+const genreQuery = (identifier: string) => ({
+  queryKey: ["genre", identifier],
+  queryFn: async () => {
+    const res = await fetch(`/genre/${identifier}/`, {
+      method: "POST",
+      headers: { Accept: "application/json" },
+    });
+    if (!res.ok) throw new Error("Failed");
```
MonkeyDo:
```suggestion
    if (!res.ok) throw new Error(`Failed to load genre ${identifier}`);
```

frontend/js/src/genre/Genre.tsx
```diff
@@ -0,0 +1,396 @@
...
+              aria-label="Toggle dropdown"
+            />
+            <div className="dropdown-menu">
+              <Link
+                to={`/explore/lb-radio/?prompt=tag:(${encodeURIComponent(
+                  name
+                )})&mode=easy`}
+                className="dropdown-item"
+              >
+                Genre radio
```
MonkeyDo: I think we could do with a small addition here to understand the difference between the two modes.

Maybe something like "This genre (and similar)" ?
That is, if that's actually what is happening in the background (judging from "nosim" on the other option)

frontend/js/src/genre/Genre.tsx
```diff
@@ -0,0 +1,396 @@
...
+                )})::nosim&mode=easy`}
+                className="dropdown-item"
+              >
+                This genre only
+              </Link>
+            </div>
+          </div>
+        </div>
+      </div>
+      <div className="entity-page-content">
```
MonkeyDo, line 282: I think above all of this content (or maybe below? 🤔) it would be nice to have a search bar that searches only for genres.
I believe we store a list of genres for the front-end, and we could use a react-select to search through that like we do in the TagsComponent, and save a trip to the search page.

listenbrainz/db/genre.py
```diff
@@ -43,3 +77,117 @@ def get_tag_hierarchy_data():
...
+            entity_type: {"count": 0, "entities": []}
+            for entity_type, _tag_table, _id_col, _order_col, _limit in TAGGED_ENTITY_TYPES
+        }
+
+    override_limits = limits or {}
+    result: dict[str, Any] = {}
+    for entity_type, tag_table, id_column, order_column, default_limit in TAGGED_ENTITY_TYPES:
+        limit = override_limits.get(entity_type, default_limit)
+        # Table names in MusicBrainz: artist, release_group, recording
+        entity_table = "release_group" if entity_type == "release_group" else entity_type
```
MonkeyDo: I might be thick right now, but I don't understand this.
The way I read this, entity_table should always equal the entity_type, and special casing "release_group" does nothing

MonkeyDo (commented):
Looks great!

A few minor comments.
I think it would be nice to link to LB pages from the TagComponent, but IIUC we can't do that for all the tags shown in that component (as LB does an auto-conversion from tag name to genre UUID).
However we seem to have a `genre_mbid` available for genres, so we can either link to the LB page by UUID or detect items with a `genre_mbid` available but still link to the LB page using the tag name.
For other tags that don`t have a genre_mbid I guess we continue linking to MB?

Might be good to get a review of the back-end bits from @amCap1712 and especially the SQL queries, I only skimmed

frontend/js/src/genre/Genre.tsx
anshg1214: Instead, how about we have the search bar on the top right of the screen on the top navbar?

image

listenbrainz/db/genre.py
```diff
@@ -8,11 +8,11 @@
 
 # Entity types that have tags in MusicBrainz: (entity_type, tag_table, id_column, order_column, limit).
 # order_column: artist uses sort_name, others use name (per MusicBrainz EntityTag role).
-TAGGED_ENTITY_TYPES = [
+TAGGED_ENTITY_TYPES = (
     ("artist", "artist_tag", "artist", "sort_name", 12),
     ("release_group", "release_group_tag", "release_group", "name", 100),
```
MonkeyDo, line 13: Is there a specific reason the limit is 12 above and below, but 100 for RGs?

frontend/js/src/genre/Genre.tsx
```diff
@@ -0,0 +1,396 @@
...
+                    "artist",
+                    artist.name,
+                    artist.mbid
+                  )}
+                  showTimestamp={false}
+                  showUsername={false}
+                  additionalActions={
+                    artist.total_listen_count != null ? (
+                      <span className="badge bg-info">
+                        {formatListenCount(artist.total_listen_count)}
```
MonkeyDo, line 310: We should add the headphones icon here that we use on the artist page for example, to denote we're talking about number of listens (and not for example number of tags, in this context):

Image

listenbrainz/db/genre.py
```diff
@@ -43,3 +75,108 @@ def get_tag_hierarchy_data():
...
+            ORDER BY tt.count DESC, e.{order_column} COLLATE musicbrainz, e.id
+            LIMIT %s OFFSET 0
+        """
+        mb_curs.execute(query, (tag_id, limit))
+        rows = mb_curs.fetchall()
+        count_query = f"""
+            SELECT COUNT(*) AS total
+            FROM musicbrainz.{tag_table} tt
+            WHERE tt.tag = %s AND tt.count > 0
+        """
```
MonkeyDo, line 167: Is there a way to do the count in the same query as above? Is there a specific reason why it is done in a separate query, to improve performance or something?

listenbrainz/webserver/views/entity_pages.py
```diff
@@ -444,3 +446,128 @@ def recording_entity(recording_mbid: str):
...
+
+@genre_bp.get('/<genre>/')
+def genre_page(genre: str):
+    return render_template("index.html")
+
+
+# This endpoint supports both genre name and genre mbid in the URL.
+@genre_bp.post("/<genre_slug>/")
+@web_listenstore_needed
+def genre_entity(genre_slug: str):
```
MonkeyDo, line 459: Just want to make sure this is cached very aggressively, as it does all kinds of queries and seems like it would be a potential new popular but resource-intensive endpoint to add to our list of DDOSable ones.
Probably also a good idea to rate-limit it more for the same reasons

listenbrainz/db/genre.py
```diff
@@ -43,3 +75,108 @@ def get_tag_hierarchy_data():
...
+        mb_curs.execute(count_query, (tag_id,))
+        total = mb_curs.fetchone()["total"]
+
+        result[entity_type] = {
```
MonkeyDo, line 171: I triedf running this locally, and despite the MB DB having tags for RGs and recordings, I'm only seeing results for artist_tag somehow:

Image

Not sure what is going on there, but can confirm I have data in the DB:

Image

MonkeyDo (commented):
Some comments as I test this locally.
Couldn't get it to fully run, but what I've seen is a great start :)

frontend/js/src/genre/Genre.tsx
MonkeyDo: I suppose we could, but it seems to me more of a "there's space there' decisions rather than "this is where users will expect it" decision.
It is quite separated from the main content so it's not clear that it relates to genre, and I think that will make it hard for users to find.

Another option that would not be for this PR would be to somehow have a select associated with the primary search bar on the top left, and make it default to the entity type being viewed (so in this case, it would have "genre" selected). But that's a bigger refactor and I don't think we want to end up with a musicbrainz-style search...

I will rope in @Aerozol for anopinion of where you think we could put a genre search bar in this page

Aerozol: Hmm, tricky. I'm trying to think of other websites where there is some kind of "search the entity type you're already viewing", but I can't think of one.

Perhaps just above the 'Open in MusicBrainz' button? And re-use the style we have in the art creator:

image

It's going to be a bit of an ugly combo, but tbh that right-hand column area (with links etc) could do with a rework across all entity pages eventually, anyway. And leaves the header tidy.

There could also be a way to tap the entity name and then search or type something in? I just can't quiiiite picture how to make it work without it being confusing. Food for thought, in case one of you has ideas:

image

P.S. I guess this page won't work for all tags, just genres? That's a shame, but I'm guessing you had your reasons!

PR #3580, opened 2026-02-13
LB-1926: Enhance error handling for JSON response parsing in Navidrome server

Swaim-Sahay:
# Problem

Navidrome server connections were failing with the error:
> **"Unable to connect to server: Expecting value: line 1 column 1 (char 0)"**

This occurred even when the Navidrome server was responding with valid JSON, preventing users from connecting their Navidrome instances to ListenBrainz.

**Ticket:** LB-1926

# Solution

Fixed exception handling in the Navidrome authentication flow to properly distinguish between network errors and JSON parsing errors, **eliminating false "connection failed" messages** for valid JSON responses.

# Changes

- **Reordered exception handlers** in `navidrome.py` to catch `ExternalServiceError` before `RequestException`
- **Added explicit handling** for `ValueError` and `requests.exceptions.JSONDecodeError`
- **Added error logging** with response text for debugging JSON parsing failures.

### DIFF
--- a/listenbrainz/domain/navidrome.py
+++ b/listenbrainz/domain/navidrome.py
@@ -64,10 +64,16 @@ class NavidromeService:
 
             response = requests.get(ping_url, params=params, timeout=10)
             response.raise_for_status()
-            data = response.json()
+
+            # Try to parse JSON response
+            try:
+                data = response.json()
+            except (ValueError, requests.exceptions.JSONDecodeError) as e:
+                current_app.logger.error(f"Failed to parse Navidrome JSON response. Response text: {response.text[:1000]}")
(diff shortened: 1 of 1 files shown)


mAmanullah7:
Hey @Swaim-Sahay thank you for opening this pr! Just wanna ask know whether you were able to re-produce the error on your local and have you tested this fix locally?

mAmanullah7 (commented):
So the PR improves error handling but may not fix the root cause, i would suggest either try to find fix the root cause why it failed to connect to navidrome server else make this pr independent of LB-1926.

But let's wait for @amCap1712 review!

anshg1214 (approved):
The PR looks good. But this does not fix LB-1926 completely.

PR #3591, opened 2026-02-19
fixes[LB-1752]: Improve playlist export error handling 

Swaim-Sahay:
# Problem
 Exporting a playlist to Apple Music was failing with an unhelpful "data" error message in the UI.

This occurred because the API endpoint only caught HTTP errors, meaning any other exceptions (like a missing key in the metadata response from Apple Music) caused an unhandled 500 server error, masking the true reason for the failure.

Ticket: LB-1752

# Solution 
Improved exception handling in the playlist export API views and the Apple Music export utility to catch generic exceptions and KeyErrors. This ensures users receive specific, actionable error messages instead of an unhandled server failure.

# Changes
1) listenbrainz/troi/export.py  Added explicit handling for missing metadata and a try/except KeyError block to provide clearer error descriptions if Apple Music's response is missing expected keys.
2) listenbrainz/webserver/views/playlist_api.py : Updated export_playlist and export_playlist_jspf to catch generic Exceptions and properly wrap them into a 500 APIError containing the underlying error string.


shirsakm:
@Swaim-Sahay Not qualified enough to comment on the changes, but just wanted to let you know that your PR summary seems unrelated to the changes you have made.

Swaim-Sahay:
Hi @shirsakm appreciate you catching that! That was my mistake the PR summary was mismatched with the actual changes. I've updated it now. Thanks for flagging it!

MonkeyDo:
I think the changes in the PR are beneficial, but they seem to be only very superficial:
image

Sure, now it tells me more than just "data" but that does not do anything to solve the root issue. 

I would appreciate it if rather than rushing to create new PRs you took the time to properly test stuff to ensure it is actually all working, as this sort of superficial PR doesn't really do anything.

MonkeyDo:
Hi @Swaim-Sahay 
Just checking if you tested this manually with an actual import from Apple Music?

PR #3592, opened 2026-02-19
Fix [LB-1701]: Spotify playlist import by paginating through all available user

Swaim-Sahay:
# Problem 
The Spotify playlist import feature was only showing a maximum of 50 playlists.
This happened because the backend only requested the first page of results from Spotify and didn't fetch the rest.

Ticket: LB-1701

# Solution
 Updated the backend to check for additional pages of playlists and fetch them all using pagination. This ensures every saved playlist is shown.

# Change
listenbrainz/webserver/views/playlist_api.py : Modified the Spotify playlist import logic to loop through all pages of Spotify results using spotify.next() instead of stopping after the first page.

### DIFF
--- a/listenbrainz/domain/musicbrainz.py
+++ b/listenbrainz/domain/musicbrainz.py
@@ -29,11 +29,15 @@ class MusicBrainzService(BaseBrainzService):
         )
 
     def get_user_info(self, token: str):
-        response = requests.get(
+        response = requests.post(
             current_app.config["OAUTH_INTROSPECTION_URL"],
-            headers={
-                "Authorization": f"Bearer {token}",
+            data={
+                "client_id": current_app.config["OAUTH_CLIENT_ID"],
+                "client_secret": current_app.config["OAUTH_CLIENT_SECRET"],
+                "token": token,
+                "token_type_hint": "access_token",
             },
+            headers={"Content-Type": "application/x-www-form-urlencoded"},
         )
         response.raise_for_status()
         return response.json()
--- a/listenbrainz/domain/musicbrainz.py
+++ b/listenbrainz/domain/musicbrainz.py
@@ -29,15 +29,11 @@ class MusicBrainzService(BaseBrainzService):
         )
 
     def get_user_info(self, token: str):
-        response = requests.post(
+        response = requests.get(
             current_app.config["OAUTH_INTROSPECTION_URL"],
-            data={
-                "client_id": current_app.config["OAUTH_CLIENT_ID"],
-                "client_secret": current_app.config["OAUTH_CLIENT_SECRET"],
-                "token": token,
-                "token_type_hint": "access_token",
+            headers={
+                "Authorization": f"Bearer {token}",
             },
-            headers={"Content-Type": "application/x-www-form-urlencoded"},
         )
         response.raise_for_status()
         return response.json()
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -28,6 +28,8 @@ SPOTIFY_LISTEN_PERMISSIONS = {
     'streaming',
     'user-read-email',
     'user-read-private',
+    'playlist-read-private',
+    'playlist-read-collaborative',
     'playlist-modify-public',
     'playlist-modify-private',
 }
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -36,7 +36,9 @@ SPOTIFY_LISTEN_PERMISSIONS = {
 
 SPOTIFY_PLAYLIST_PERMISSIONS = {
     'playlist-modify-public',
-    'playlist-modify-private'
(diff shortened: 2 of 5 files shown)


MonkeyDo:
How did you test this?

I deployed it to our test server and am only getting shown 27 our of 67 playlists (I checked the total count with the Spotify API), the exact same ones as on the production website.

Swaim-Sahay:
The 27 playlists you're seeing are just your public ones. I verified that Spotify silently filters out private/collaborative playlists unless explicitly asked for them, which is why the previous 50-item pagination loop finished immediately for you. I've updated the requested scopes to include playlist-read-private and playlist-read-collaborative if you completely relink your Spotify account on the test server with these new scopes, the importer will now fetch all 67 playlists!

listenbrainz/domain/spotify.py
```diff
@@ -28,6 +28,8 @@
     'streaming',
     'user-read-email',
     'user-read-private',
+    'playlist-read-private',
+    'playlist-read-collaborative',
```
MonkeyDo, line 32: You're going to want those in the `SPOTIFY_PLAYLIST_PERMISSIONS` object just below instead.
There will also be some related changes to make in a couple of error messages where `SPOTIFY_PLAYLIST_PERMISSIONS` is used

MonkeyDo: Also. were you able to test this manually?
Disconnecting and reconnecting to spotify and testing import?

listenbrainz/domain/musicbrainz.py
```diff
@@ -29,15 +29,11 @@ def __init__(self):
...
             current_app.config["OAUTH_INTROSPECTION_URL"],
-            data={
-                "client_id": current_app.config["OAUTH_CLIENT_ID"],
-                "client_secret": current_app.config["OAUTH_CLIENT_SECRET"],
-                "token": token,
-                "token_type_hint": "access_token",
+            headers={
+                "Authorization": f"Bearer {token}",
             },
-            headers={"Content-Type": "application/x-www-form-urlencoded"},
```
MonkeyDo, line 40: What is this change?
It seems both like a really important change and one that should not at all be part of this particular PR.
Can you please explain why this was changed and why it was commited?

Please be careful about what changes you commit.

MonkeyDo: I would like an answer to this, I think it's important to recognize the issue and not sweep it under the rug.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -892,7 +892,8 @@ def export_playlist(playlist_mbid, service):
         raise APIBadRequest(f"Service {service} is not linked. Please link your {service} account first.")
 
     if service == 'spotify' and not SPOTIFY_PLAYLIST_PERMISSIONS.issubset(set(token["scopes"])):
-        raise APIBadRequest(f"Missing scopes playlist-modify-public and playlist-modify-private to export playlists."
+        missing_scopes = ", ".join(SPOTIFY_PLAYLIST_PERMISSIONS)
```
MonkeyDo: If we're here improving the error message, let's compared the sets and print out the difference, meaning the actual missing scopes and not all the required scopes.

https://www.w3schools.com/python/ref_set_difference.asp

listenbrainz/webserver/views/settings.py
MonkeyDo, line 1: I don't think these changes are required, as SPOTIFY_LISTEN_PERMISSIONS already contains SPOTIFY_PLAYLIST_PERMISSIONS

MonkeyDo (changes requested):
Can confirm this is working as expected now, that's good!

The oauth changes that were reverted are giving me pause.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -891,10 +891,15 @@ def export_playlist(playlist_mbid, service):
...
         raise APIBadRequest(f"Service {service} is not linked. Please link your {service} account first.")
 
-    if service == 'spotify' and not SPOTIFY_PLAYLIST_PERMISSIONS.issubset(set(token["scopes"])):
-        raise APIBadRequest(f"Missing scopes playlist-modify-public and playlist-modify-private to export playlists."
-                            f" Please relink your {service} account from ListenBrainz settings with appropriate scopes"
-                            f" to use this feature.")
+    if service == "spotify":
+        export_permissions = {"playlist-modify-public", "playlist-modify-private"}
+        if not export_permissions.issubset(set(token["scopes"])):
+            missing_scopes = " and ".join(sorted(export_permissions - set(token["scopes"]), reverse=True))
```
MonkeyDo: Let's implement this in the same simpler way as the others to keep it consistent:

```suggestion
            missing_scopes = ", ".join(export_permissions - set(token["scopes"]))
```

MonkeyDo (changes requested):
Final little nitpick, thanks for the PR!

PR #3604, opened 2026-02-28
Fix LB-1942: Change grid-auto-flow to row in cover-art-container

Gopal7387:
### **Problem**
The problem is mentioned in this ticket: LB-1942
Even though the albums were sorted by release date, the grid layout was making them appear top-to-bottom in columns instead of left-to-right across rows when multiple rows were shown.

### **Solution**
Updated the grid to use grid-auto-flow: row and removed the unnecessary fixed row constraints, so albums now display left-to-right across rows as intended, also fixing the same issue on the track page.

Screenshot 2026-02-28 at 10 55 17 PM

### DIFF
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -200,7 +200,7 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
         .cover-art-container {
           display: grid;
           grid-gap: 1.5em;
-          grid-auto-flow: column;
+          grid-auto-flow: row;
           justify-items: center;
           grid-template-columns: repeat(auto-fill, minmax(190px, max-content));
           grid-template-rows: repeat(2, 1fr);
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -201,11 +201,12 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
         .album-grid-outer {
           // Outer wrapper for overflow detection and height constraint
           position: relative;
-          // Measured in browser at 14px root font:
-          // Card height ≈ 215px (~15.35em), gap = 21px (1.5em)
-          // 2 rows + gap + partial 3rd-row hint ≈ 33.5em
-          // Set to 25.75em to show ~1.5 rows; increase to 35em for 2 full rows + hint
-          max-height: 25.75em; // 2 rows visible — tune to 35em if fuller rows needed
+          // Height tuned based on real browser measurements at 14px root font:
+          // - Card height ≈ 215px (~15.35em)
+          // - Grid gap = 1.5em (~21px)
+          // - 2 rows + gap + partial third-row hint ≈ 33.5em
+          // Rounded to 35em for stable visual spacing across breakpoints
+          max-height: 35em;
           overflow: hidden; // vertical overflow clipped
           
           // Bottom gradient overlay to hint at more content
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -201,14 +201,20 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
         .album-grid-outer {
           // Outer wrapper for overflow detection and height constraint
           position: relative;
-          // Height tuned based on real browser measurements at 14px root font:
-          // - Card height ≈ 215px (~15.35em)
-          // - Grid gap = 1.5em (~21px)
-          // - 2 rows + gap + partial third-row hint ≈ 33.5em
-          // Rounded to 35em for stable visual spacing across breakpoints
-          max-height: 35em;
-          overflow: hidden; // vertical overflow clipped
-          
+          // - Card height ≈ 21.556rem (215.56px at root font-size 10px)
+          // - Grid gap = 1.5rem = 15px
+          // - 2 full rows + 2 gaps + 0.5 row = 56.889rem ≈ 56.9rem
+          // Using rem (not em) to anchor to the 10px root font-size,
+          // since this element inherits a 14px parent font-size.
+          max-height: 56.9rem;
+          overflow: hidden; // clips content; ::after gradient stays fixed at bottom
+
+          .album-grid-inner {
+            max-height: inherit;
+            overflow-y: auto;
+            overflow-x: hidden;
+          }
+
           // Bottom gradient overlay to hint at more content
           &::after {
             content: "";
@@ -223,7 +229,7 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
             // Hide by default, show only when has-overflow class is present
             display: none;
           }
-          
+
           // Show gradient only when container has vertical overflow
           &.has-overflow::after {
             display: block;
@@ -233,11 +239,11 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
         .cover-art-container {
           display: grid;
           grid-gap: 1.5em;
-          grid-auto-flow: column;
+          grid-auto-flow: row;
           justify-items: center;
           grid-template-columns: repeat(auto-fill, minmax(190px, max-content));
           // Grid flows naturally without row constraints
-          
+
           .cover-art {
             aspect-ratio: 1;
             background: lightgrey;
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -198,44 +198,6 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
       .albums {
         width: 100%;
 
-        .album-grid-outer {
-          // Outer wrapper for overflow detection and height constraint
-          position: relative;
-          // - Card height ≈ 21.556rem (215.56px at root font-size 10px)
-          // - Grid gap = 1.5rem = 15px
-          // - 2 full rows + 2 gaps + 0.5 row = 56.889rem ≈ 56.9rem
-          // Using rem (not em) to anchor to the 10px root font-size,
-          // since this element inherits a 14px parent font-size.
-          max-height: 56.9rem;
-          overflow: hidden; // clips content; ::after gradient stays fixed at bottom
-
-          .album-grid-inner {
-            max-height: inherit;
-            overflow-y: auto;
-            overflow-x: hidden;
-          }
-
-          // Bottom gradient overlay to hint at more content
-          &::after {
-            content: "";
-            position: absolute;
-            bottom: 0;
-            left: 0;
-            width: 100%;
-            height: 4em;
-            pointer-events: none;
-            background-image: $white-gradient;
-            z-index: 5;
-            // Hide by default, show only when has-overflow class is present
-            display: none;
-          }
-
-          // Show gradient only when container has vertical overflow
-          &.has-overflow::after {
-            display: block;
-          }
-        }
-
         .cover-art-container {
           display: grid;
           grid-gap: 1.5em;
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -201,10 +201,11 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
         .album-grid-outer {
           // Outer wrapper for overflow detection and height constraint
           position: relative;
-          // Height = (190px card height × 2) + (1.5em gap) + (0.5em hint)
-          // 190px ≈ 11.875em at 16px base font
-          // Total = (11.875em × 2) + 1.5em + 0.5em = 25.75em
-          max-height: 25.75em; // Exactly 2 rows + gap + hint of third row
+          // Measured in browser at 14px root font:
+          // Card height ≈ 215px (~15.35em), gap = 21px (1.5em)
+          // 2 rows + gap + partial 3rd-row hint ≈ 33.5em
+          // Set to 25.75em to show ~1.5 rows; increase to 35em for 2 full rows + hint
+          max-height: 25.75em; // 2 rows visible — tune to 35em if fuller rows needed
           overflow: hidden; // vertical overflow clipped
           
           // Bottom gradient overlay to hint at more content
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -197,16 +197,45 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
       }
       .albums {
         width: 100%;
+
+        .album-grid-outer {
+          // Outer wrapper for overflow detection and height constraint
+          position: relative;
+          // Height = (190px card height × 2) + (1.5em gap) + (0.5em hint)
+          // 190px ≈ 11.875em at 16px base font
+          // Total = (11.875em × 2) + 1.5em + 0.5em = 25.75em
+          max-height: 25.75em; // Exactly 2 rows + gap + hint of third row
+          overflow: hidden; // vertical overflow clipped
+          
+          // Bottom gradient overlay to hint at more content
+          &::after {
+            content: "";
+            position: absolute;
+            bottom: 0;
+            left: 0;
+            width: 100%;
+            height: 4em;
+            pointer-events: none;
+            background-image: $white-gradient;
+            z-index: 5;
+            // Hide by default, show only when has-overflow class is present
+            display: none;
+          }
+          
+          // Show gradient only when container has vertical overflow
+          &.has-overflow::after {
+            display: block;
+          }
+        }
+
         .cover-art-container {
           display: grid;
           grid-gap: 1.5em;
-          grid-auto-flow: row;
+          grid-auto-flow: column;
           justify-items: center;
           grid-template-columns: repeat(auto-fill, minmax(190px, max-content));
-          grid-template-rows: repeat(2, 1fr);
-          &.single-row {
-            grid-template-rows: repeat(1, 1fr);
-          }
+          // Grid flows naturally without row constraints
+          
           .cover-art {
             aspect-ratio: 1;
             background: lightgrey;
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -203,7 +203,7 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
           grid-gap: 1.5em;
           grid-auto-flow: row;
           justify-items: center;
-          grid-template-columns: repeat(auto-fill, minmax(190px, max-content));
+          grid-template-columns: repeat(auto-fill, minmax(12em, 1fr));
           // Grid flows naturally without row constraints
 
           .cover-art {
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -197,15 +197,16 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
       }
       .albums {
         width: 100%;
-
         .cover-art-container {
           display: grid;
           grid-gap: 1.5em;
           grid-auto-flow: row;
           justify-items: center;
-          grid-template-columns: repeat(auto-fill, minmax(12em, 1fr));
-          // Grid flows naturally without row constraints
-
+          grid-template-columns: repeat(auto-fill, minmax(190px, max-content));
+          grid-template-rows: repeat(2, 1fr);
+          &.single-row {
+            grid-template-rows: repeat(1, 1fr);
+          }
           .cover-art {
             aspect-ratio: 1;
             background: lightgrey;
--- a/frontend/css/sass/main.scss
+++ b/frontend/css/sass/main.scss
@@ -33,10 +33,10 @@
 @import "tags";
 @import "music-neighborhood";
 @import "stats-art-creator";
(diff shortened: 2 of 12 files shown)


MonkeyDo:
Hi, the ticket was lacking some extra information: We want to avoid having an endless grid of album that would make the page hard to use.
For a lot of artists it wouldn't be a problem, but for those with lots of albums it will make the page much less usable.

So I think you will need to constrain the container element to a specific max size  so only two rows (or two and a half?) show. We will also likely want some indicator that there are more items below, a small transparent-to-white gradient at the bottom that hints at more content.
We do this for other elements, see for example the "Popular tracks" section on the artist page

This is where my suggestion of showing two and a half rows would some in, showing two full rows and just a little bit of the third row, with the gradient over it. We would only want to show the gradient when the container element is scrollable, which might require some JS to calculate number of items to display vs. available container size (unfortunately this cannot be done dynamically with CSS).
(All these complications are the reason why this wasn't implemented yet)

Gopal7387:
Hi @MonkeyDo, thanks for taking the time and for having a good look. That makes complete sense from a usability standpoint.

I’ll take a look at the “Popular tracks” section and replicate a similar approach by constraining the container’s height, showing about two rows, and having a gradient indicator at the bottom. 
The gradient indicator will only show if the container is scrollable, which can be determined via JS.

I’ll work on this implementation and update the PR as well.

Thanks!

MonkeyDo:
Please don't forget to assign tickets to yourself on Jira before you start working on them, to avoid someone else picking the same ticket.
I've assigned the ticket to you this time

Gopal7387:
Hi @MonkeyDo ,
I have implemented the height constraint and conditional overflow gradient as discussed.
Please let me know if you have any adjustments to the max-height value.
Thanks.


Screenshot 2026-03-03 at 7 30 52 PM

MonkeyDo:
Sorry if I was not clear, this isn't exactly what I had in mind.
Rather than have a single row, we want to show two and a half rows maximum, and have **vertical** scroll rather than horizontal scroll.

If there are two full rows or less, no need for scroll or for the transparent overlay:
image

If there are more here's a mockup of what I had in mind:
image


You could adapt the `horizontal-scroll-container` and `horizontal-scroll` classes to expand it and create a similar `vertical-scroll-container` and `vertical-scroll` using the same styles.

Gopal7387:
Hi @MonkeyDo,  
Updated the PR with the changes you suggested.
And an apology for little misunderstanding by me.

Key changes:
- Swapped HorizontalScrollContainer for a two-div structure
  (.album-grid-outer > .album-grid-inner) ,this keeps the gradient
  pinned at the bottom while the inner div handles the actual scroll
- Used rem instead of em for max-height (56.9rem) since this element
  inherits a 14px parent font-size , em was resolving to ~797px (3.5 rows)
  instead of the intended 569px (2.5 rows)
- Gradient only shows when the content actually overflows, detected via
  ResizeObserver checking scrollHeight > clientHeight
- No gradient or scrollbar for artists with 2 rows or fewer

One note on the gradient color and  it uses the same $white-gradient 
already used for the "See full discography" fade on the same page, 
So it should be consistent. 

Let me know if there are any adjustments, if you need.

Screenshot 2026-03-06 at 8 25 54 PM

MonkeyDo:
I think ideally some of this would be abstracted so it can be reused more easily.

I suggest refactoring the existing [HorizontalScrollContainer component](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/components/HorizontalScrollContainer.tsx) that already does a lot of work and styling, including dragging support, arrows, setting beginning and end css classes to show/hide the gradient.
It needs to be modified to accept a `direction` prop of either "horizontal" (the default, minimizes changes required around the codebase) and "vertical".

If the direction is vertical you will need to change some values conditionally, for example `scrollWidth` > `scrollHeight` to adapt it to the direction,as well as change the css classes `horizontal-scroll-*` to `vertical-scroll-*`

Ideally you would also modify the SCSS styles ([this file](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/css/sass/scroll-container.scss)) so that there is a common `.scroll-container` with properties common to both directions, and adding new `&.horizontal` and `&.vertical` classes to the same element to define direction-specific styles.
If that sounded like gibberish and you don't know SCSS, let me know and I can give you more information.

Gopal7387:
Hi @MonkeyDo  I am done. I have made changes to the code as you suggested.

I have added a direction prop to the HorizontalScrollContainer component. The direction prop is set to horizontal by default. I did not have to make any other changes to the code.

When the HorizontalScrollContainer component is in mode I have made some changes. I have changed the scroll logic to use scrollTop and clientHeight. I have also replaced the right arrows with up and down arrows. Additionally I have drag scroll because this feature only works with horizontal scrolling.

I have also made some changes to the SCSS code. I have refactored the scroll-container.scss file into a shared scroll-container class. This class has two subclasses: vertical. The vertical class has a height of 56.9rem and a bottom gradient. The horizontal class is the same as it was before.

The ArtistPage component now uses the HorizontalScrollContainer component with the direction set to vertical. This component also uses a custom overflow state, refs and ResizeObserver. However this is no longer necessary because the HorizontalScrollContainer component takes care of these things internally.

While I was working on this I also fixed the order of the main.scss files. The scroll-container file now loads before the entity-pages file. This is important because the entity-pages file would have caused a mixin compile error if it had loaded first.

Let me know if you have any suggestions, for the HorizontalScrollContainer component.

Thanks.

Gopal7387:
Hey @MonkeyDo  🙂
Just wanted to circle back on this work updated by me, and it's been a while.
So, plz have a look at this when u got time.

Gopal7387:
Hi @MonkeyDo , just checking in on this. I implemented all the changes you suggested in your last review (direction prop, SCSS refactor with shared .scroll-container + .horizontal/.vertical subclasses, vertical scroll logic, up/down arrows). The branch is clean and conflict-free with master.

No rush, just want to make sure it's on your radar when you get a moment. 🙂

frontend/js/src/components/HorizontalScrollContainer.tsx
```diff
@@ -19,11 +22,14 @@ const MANUAL_SCROLL_AMOUNT = 500;
 export default function HorizontalScrollContainer({
```
MonkeyDo: This should be renamed wherever it is used:

```suggestion
export default function ScrollContainer({
```

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -528,23 +529,31 @@ export default function ArtistPage(): JSX.Element {
...
+                <HorizontalScrollContainer
+                  direction="vertical"
+                  showScrollbar={false}
+                  enableDragScroll={false}
+                >
+                  <div
+                    className={`cover-art-container ${
+                      rgGroup.length <= COVER_ART_SINGLE_ROW_COUNT
+                        ? "single-row"
+                        : ""
```
MonkeyDo: This is now unused as far as I understand, and should be removed.
I think you might be able to remove the wrapping div? Do check before removing though.

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -528,23 +529,31 @@ export default function ArtistPage(): JSX.Element {
...
+            return (
+              <div className="albums" key={type}>
+                <div className="listen-header">
+                  <h3 className="header-with-line">{type}</h3>
+                  <SortingButtons sort={sort} setSort={setSort} />
+                </div>
+                <HorizontalScrollContainer
+                  direction="vertical"
+                  showScrollbar={false}
+                  enableDragScroll={false}
```
MonkeyDo, line 542: You should apply the same changes in the frontend/js/src/track/Track.tsx file

frontend/css/sass/entity-pages.scss
```diff
@@ -197,16 +197,15 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
...
       .albums {
         width: 100%;
+
         .cover-art-container {
           display: grid;
           grid-gap: 1.5em;
-          grid-auto-flow: column;
+          grid-auto-flow: row;
           justify-items: center;
           grid-template-columns: repeat(auto-fill, minmax(190px, max-content));
```
MonkeyDo: Did some testing, I think this value will give better results on all sizes I tested with apart from largest desktops:

```suggestion
          grid-template-columns: repeat(auto-fill, minmax(12em, 1fr));
```

On larger desktops, it might be  better with the original 190px instead of the  12em but on mobile sizes it is for sure an improvement:

Now:

Image

With my changes applied:

Image

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -528,23 +529,31 @@ export default function ArtistPage(): JSX.Element {
...
           }`}
         >
-          {releaseGroupTypesNames.map(([type, rgGroup]) => (
-            <div className="albums">
-              <div className="listen-header">
-                <h3 className="header-with-line">{type}</h3>
-                <SortingButtons sort={sort} setSort={setSort} />
+          {releaseGroupTypesNames.map(([type, rgGroup]) => {
+            return (
+              <div className="albums" key={type}>
```
MonkeyDo, line 534: This `key` prop is new. Was it added on purpose for a reason, or just an LLM doing its thing?

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -528,23 +529,31 @@ export default function ArtistPage(): JSX.Element {
...
+                <div className="listen-header">
+                  <h3 className="header-with-line">{type}</h3>
+                  <SortingButtons sort={sort} setSort={setSort} />
+                </div>
+                <HorizontalScrollContainer
+                  direction="vertical"
+                  showScrollbar={false}
```
MonkeyDo, line 541: What is the rationale for hiding the scrollbar and disabling drag-scroll?

frontend/js/src/components/HorizontalScrollContainer.tsx
```diff
@@ -116,22 +133,28 @@ export default function HorizontalScrollContainer({
   }, []);
 
   return (
-    <div className="horizontal-scroll-container">
+    <div
+      className={
+        isVertical
+          ? "scroll-container vertical"
+          : "scroll-container horizontal horizontal-scroll-container"
```
MonkeyDo: Why does the horizontal, mode have the extra `horizontal-scroll-container` class?

frontend/css/sass/scroll-container.scss
```diff
@@ -32,60 +32,101 @@ $scrollbar-thumb-color-dark: color.adjust(
...
+      @include small-scrollbar;
+      &.no-scrollbar {
+        /* Hide the scrollbar: */
+        -ms-overflow-style: none; /* Internet Explorer and Edge */
+        scrollbar-width: none; /* Firefox */
+        &::-webkit-scrollbar {
+          /* Chrome, Safari and Opera */
+          display: none;
+        }
+      }
```
MonkeyDo: These are not specific to the horizontal or vertical scroll, so should be moved up a level and deduplicated.

MonkeyDo (changes requested):
Thanks for continuing work on this.
We're definitely in the right ballpark now, and I've switched to a fine-toothed comb for reviewing:

frontend/js/src/artist/ArtistPage.tsx
Gopal7387: The key prop was missing on the .albums div in the original code. React requires it on elements returned from .map(),  so I added it to fix that.

frontend/js/src/artist/ArtistPage.tsx
Gopal7387: Hidden the scrollbar since the gradient arrows already indicate overflow. 
Disabled drag-scroll as it conflicts with vertical scrolling, users scroll naturally with the mouse wheel or touch.

frontend/js/src/components/HorizontalScrollContainer.tsx
Gopal7387: It was a leftover from before the SCSS refactor. Removed in this update.

frontend/css/sass/scroll-container.scss
Gopal7387: Moved to the root .scroll-container level in this update.

frontend/js/src/components/HorizontalScrollContainer.tsx
Gopal7387: Done.

frontend/js/src/artist/ArtistPage.tsx
Gopal7387: Removed. Kept the wrapping div as it's needed for the cover-art-container grid styles in entity-pages.scss.

frontend/js/src/artist/ArtistPage.tsx
Gopal7387: Applied the same changes in Track.tsx.

frontend/css/sass/entity-pages.scss
```diff
@@ -197,16 +197,15 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
...
-          grid-auto-flow: column;
+          grid-auto-flow: row;
           justify-items: center;
-          grid-template-columns: repeat(auto-fill, minmax(190px, max-content));
-          grid-template-rows: repeat(2, 1fr);
-          &.single-row {
-            grid-template-rows: repeat(1, 1fr);
-          }
+          grid-template-columns: repeat(auto-fill, minmax(12em, 1fr));
+          // Grid flows naturally without row constraints
```
MonkeyDo, line 207:
```suggestion
```

frontend/css/sass/main.scss
```diff
@@ -33,10 +33,10 @@
 @import "tags";
 @import "music-neighborhood";
 @import "stats-art-creator";
+@import "scroll-container";
```
MonkeyDo, line 36: What was the reason for moving this?

frontend/js/src/artist/ArtistPage.tsx
MonkeyDo: Not sure I agree with the choices, but I'll ask around for a second opinion/

frontend/css/sass/scroll-container.scss
```diff
@@ -32,60 +32,92 @@ $scrollbar-thumb-color-dark: color.adjust(
...
-  }
-  // hide overflow behind button with gradient to indicate there's more
   .nav-button {
     position: absolute;
-    top: 0;
-    bottom: 0;
     width: 55px;
     border: none;
     font-size: 2em;
     color: $gray-500;
```
MonkeyDo, line 50: While we are here making modifications, I think we can improve the styling a bit. On some backgrounds the icon is nearly invisible:

Image

Maybe a CSS drop-shadow filter on the path inside the SVGs?
`filter: drop-shadow(0px 50px 10px black)`:

Image

MonkeyDo (commented):
Thanks for continuing work on this!

I'm not completely satisfied with the design as it stands, although that is not your fault, you're following my suggestions...
Somehow the new version looks much busier and the separation between sections seems more blurred, where they were well defined before.

So, sorry for the long text:

What do you think about another option:
Instead of having all the separate sections (Albums, EP, Single, etc.) visible by default, we could have accordions for each section, with the first one being opened by default.

The result would be something like this on first page load:

Image
And with the entire discography shown:

Image

For comparison, it currently looks like this:

Image

And as an example of what it would look like for a small band:

Image

Compared to the current version on prod:

Image


You should be able to reuse/adapt the react-bootstrap accordion component and corresponding SASS styles that we use on [the explore page](https://beta.listenbrainz.org/explore/), and have it look just like the existing headers, with an extra little arrow on the left to show open/close state.
See usage [here](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/explore/Explore.tsx#L97-L135)

Aerozol:
Overall I agree with this being less confusing than that double sideways scroll. Though I wonder if changing to a single-row sideways scroll would be better?

This is going to give users a hell of a time when scrolling on mobile (it's going to be hard to scroll to the bottom of the page without accidentally scrolling through albums). It is already causing me pain when scrolling with my mouse, especially with everything being slow atm so that it keeps pausing to load albums/artwork.

The key question is if seeing heaps of albums at a glance (rather than after a scroll) is something that is important to our LB users, which I don't know. We already offer some sorting options. It may be that it's more important to have a compact overview, where users can side-scroll if they want to explore more of something. Probably a ticket where it would have been good to do some polling of users etc before going to the code stage.

If we do stick with this version perhaps slightly more generous spacing before/after each section would help, since it starts to get a bit busy at the bottom.

frontend/css/sass/main.scss
Gopal7387: Moved it while refactoring ScrollContainer to keep utility styles 
before page-specific ones. I can revert if preferred.

Gopal7387:
Hey @MonkeyDo, sorry for the gap. I have not had time for this lately. Before I put work in did you get a second opinion, on the scrollbar/drag-scroll thing? Also, is the accordion idea still the direction you want? Should I hold off on the accordion idea for now? Please let me know either way.

MonkeyDo:
Hi @Gopal7387 I'll have a chat with aerozol and let you know once we have a decision

MonkeyDo:
Hi @Gopal7387 !

Don't hate me... :see_no_evil: We had a more thorough chat with aerozol, and we were not happy with any of the solutions we had proposed so far.
We would like to refactor this page more rather than patching issues here and there.
Here's what we have in mind:

Instead of showing all the releases separated by categories, each in a scroll container, we want to have a single regular container, with plain old native browser scroll, and some filters at the top to allow users to show/hide the categories they are interested in.
That may mean that for some artists, activating all filters will result in a crazy long page, but hey, if that's the user's choice...
And by default we will have a pretty contained view without need of a "see more" button or any weird scroll.

Here's a mockup I'll refer to below
image


In practice, that means:
At the top, where we currently have the "Albums" header, we want to show pills that act as filters, visually similar to the one on the right of the page for sorting behavior. These new pills will act as checkboxes rather than radio buttons, so users can click on them to activate/deactivate that filter.
By default, "Album" and "EP" should be active, and any others inactive. In my mockup, imagine the user has loaded the page and clicked on "compilation" to activate that filter as well.
The list of filters should be computed based on the available types (we already have that code, the one we use for separating the categories currently).
We have some logic to separate primary types first ("album", "ep", "single", etc.) then for releases that also have a secondary type ("live", "remix", "demo", etc.), we will currently show only the secondary type.
For context, here are [some docs](https://musicbrainz.org/doc/Release_Group/Type) regarding types in MusicBrainz.

Within the container, we will show all releases for the selected filters, sorted as requested (by date or by listen count). No more separation between categories.
Notice in my mockup that we see album compilation and ep all together, sorted by date in this case.
We show the primary type for releases with only a primary type ("album" & "ep" in my mockup), and show only the secondary type for releases that have one (the "compilation" entries in my mockup are all "album"+"compilation")
We might have to revisit how we display this at a later date but for now we're going with that.

I think your improvement of using `grid-auto-flow: row` will be useful here.

Let me know if you have any questions or if anything I said seems unclear to you

MonkeyDo:
Also, thanks for hanging in there for 6 MONTHS (!!) and through so many changes...

Gopal7387:
Hi @MonkeyDo, no hate here at all not a chance 😅. I'm honestly kinda relieved we are going for the better version.

Before I work further, a few quick questions:

Multi-secondary-type releases --> the MusicBrainz docs mention a release can have more than one secondary type at once (e.g. Soundtrack + Compilation). Your rule was "show only the secondary type when one exists", but if there are two, which one wins the display? Or should such a release just show up under both pills?

Old work -->  Should I revert the ScrollContainer/vertical-scroll stuff from earlier, or just leave it sitting unused in case it's going to be in use of somewhere else down the line? 

For everything else (Other/Broadcast pills, default-filter edge cases, sort behavior) I'll just make a reasonable call and flag it clearly when I push progress.

 I genuinely appreciate you for the reviews.🙂

MonkeyDo:
> Hi @MonkeyDo, no hate here at all not a chance 😅. I'm honestly kinda relieved we are going for the better version.

Thanks, that's nice to read :)

> Multi-secondary-type releases --> the MusicBrainz docs mention a release can have more than one secondary type at once (e.g. Soundtrack + Compilation). Your rule was "show only the secondary type when one exists", but if there are two, which one wins the display? Or should such a release just show up under both pills?

Good catch, I forgot about multiple secondary types!
Tricky one. I think the current logic flatly ignores this complication, and I would suggest doing the same for the moment, showing only the first secondary type. This will be the base for the filtering logic too. Imperfect, but...
In practice those cases will be relatively rare

https://github.com/metabrainz/listenbrainz-server/blob/e4ebb512fa79bac86ec7af94f836fbba8733308a/frontend/js/src/artist/ArtistPage.tsx#L191

The longer-term solution might be to show secondary types separately, something along those lines:
But that's for another PR and we need to think more about the implications.
9d8d6f81-651a-4dc2-adc7-71772f735a4f

> Old work --> Should I revert the ScrollContainer/vertical-scroll stuff from earlier, or just leave it sitting unused in case it's going to be in use of somewhere else down the line?

Yes I think we should revert it, probably even start a new branch and PR afresh so this one stays here unchanged.
Not lost work though, if we ever need a vertical scroll container I know where to find one :)

> I genuinely appreciate you for the reviews.🙂

I appreciate you sticking with this !

PR #3609, opened 2026-03-02
LB-1933: Art Creator: Add styles for artwork container and text elements

uzman2406:
Added styles for the artwork container to manage overflow and word wrapping.
# Problem 
[LB-1933](https://tickets.metabrainz.org/browse/LB-1933) Albums in Art Creator creates overflow on mobile devices.
The artwork container section on mobile devices causes horizontal page overflow, making the page scrollable horizontally and creating a poor user experience on smaller screens.

# Solution
Added CSS styles to the artwork container to properly handle text wrapping and prevent horizontal overflow on mobile devices. The solution includes:

- Setting proper width constraints with `width: 100%` and `max-width: 100%`
- Adding `box-sizing: border-box` to ensure padding doesn't add to width
- Implementing `overflow-wrap: break-word` and `word-break: break-word` to ensure long text strings wrap to the next line naturally
- Ensuring all text elements inside the container inherit proper wrapping behavior with `white-space: normal`
- Removing `overflow-x: hidden` from the container level and applying it only at the body/html level if needed
- Adding responsive adjustments through media queries for optimal mobile viewing

I am attaching screenshots as well .

image

Screenshot 2026-03-03 015703
image

# AI usage

# Action
 Review the CSS changes to ensure they don't affect desktop layouts negatively.


uzman2406:
Hey @MonkeyDo 
I've fixed the CSS linting errors in my PR for the Art Creator section overflow issue. However, the CI is failing with an unrelated test error in LinkListens.test.tsx.
This test failure doesn't seem related to my changes as I only modified stats-art-creator.scss and its running fine in my local setup. All the linting checks are now passing. Should I try to fix this test as part of my PR?

MonkeyDo:
> Hey @MonkeyDo I've fixed the CSS linting errors in my PR for the Art Creator section overflow issue. However, the CI is failing with an unrelated test error in LinkListens.test.tsx. This test failure doesn't seem related to my changes as I only modified stats-art-creator.scss and its running fine in my local setup. All the linting checks are now passing. Should I try to fix this test as part of my PR?

Failing test is unrelated and should be a transient issue.
Re-ran the suite, it should pass.

uzman2406:
@MonkeyDo Thanks for re-running the tests😊. Can you please review it now.

uzman2406:
Hey @MonkeyDo Could you please have a look to this PR if you got a minute .

frontend/css/sass/stats-art-creator.scss
MonkeyDo, line 1: This should be refactored to take advantage of sass syntax, where you can nest specifiers.
So for example instead of defining `.artwork-container`  and `.artwork-container *` you should have the `*` styles inside the `.artwork-container` block. Same goes for media queries:
```sass
.artwork-container {
  [styles for `.artwork-container` here]

  * {
    [styles for `.artwork-container *` here]
  }

  @media screen and (max-width: 768px) {
    [styles for `.artwork-container` with media query here]
  }
}
```

If you are making CSS changes directly int eh browser console and haven't set your local dev up to compile sass files and check yous changes locally, now is the time.

frontend/css/sass/stats-art-creator.scss
```diff
@@ -353,3 +353,42 @@ $color-selector-size: 24px;
...
+}
+
+.artwork-container a {
+  display: inline-block;
+  max-width: 100%;
+  overflow: visible;
+}
+
+/* Media queries */
+@media screen and (max-width: 768px) {
```
MonkeyDo: Instead of using pixel values here, use the `$grid-breakpoints` sass variable like so (with the right size):
https://github.com/metabrainz/listenbrainz-server/blob/c4d9d7d77d626335f0a1072c066408805c8326e5/frontend/css/sass/homepage.scss#L23


You will need to add this line at the top of the file:
`@use "sass:map";`


See more info here:
https://getbootstrap.com/docs/5.0/layout/breakpoints/#available-breakpoints

frontend/css/sass/stats-art-creator.scss
```diff
@@ -353,3 +353,42 @@ $color-selector-size: 24px;
...
+  hyphens: auto;
+}
+
+.artwork-container * {
+  max-width: 100%;
+  box-sizing: border-box;
+  white-space: normal;
+  overflow-wrap: break-word;
+  word-break: normal;
+}
```
MonkeyDo: I think this is applied much too widely (every child element), and not what we want for all items (such as buttons).

It seems that much of the overflowing issues are caused by the button bar, in which case the style for the bar itself should be reviewed, for example enabling wrapping for the flex element to allow the button bar to be spread on a second line.
This will likely need more detail work to ensure for example the input foes no spread too wide.

frontend/css/sass/stats-art-creator.scss
```diff
@@ -353,3 +353,42 @@ $color-selector-size: 24px;
...
 }
+/*Fixd artwork container*/
+.artwork-container {
+  width: 100%;
+  max-width: 100%;
+  box-sizing: border-box;
+  overflow-wrap: break-word;
+  word-wrap: break-word;
+  word-break: normal;
+  hyphens: auto;
```
MonkeyDo: What does this apply to?
Currently everything, which is weird for buttons for example.
You need to review what are the exact specific changes that you need (hints below) rather than apply a bunch of styles and seeing what sticks.
Your final patch should contain only the changes that are required to fix the page.

MonkeyDo (changes requested):
Hi,

I think you're on the right track generally, but the button changes are not what we want.
The buttons and their icons have become too small to be usable on mobile, and the alt-text one for example is wrapping to a new line which is unexpected for a button (even though it looks like a link)

Image

uzman2406:
Hey @MonkeyDo! I understood whatever changes you want me to do and I'll try to fix it as soon as possible but right now I need to work on my GSoC proposal....So I apologies to pause it here for a while ...I will get back to it in 2 to 3 days after posting my proposal on community forum...Hope you understand ......Thanks for the review😊.

frontend/css/sass/stats-art-creator.scss
uzman2406: Actually ,I was initially taking a bit of a hit-and-try approach by applying different styles to address the overflow issue:) But now I’ve identified the actual cause of the overflow (mainly around the button bar layout) and have started working on a more targeted fix. will update the PR soon.. Thanks for the review:)

uzman2406:
Hi @MonkeyDo ! I've made the following changes to fix the horizontal scroll issue on mobile:

Added @use "sass:map" at the top of the file to enable.
Added display: flex, `align-items: center and justify-content: center to .share-buttons so the icon tray is centered.
Scoped flex-wrap: wrap and row-gap: 0.5em on .share-buttons inside a @media query, so wrapping only happens on small phones and tablets/desktops retain the original single-row layout.
Added a matching media query inside .link-container to set min-width ,max-width on small screens so the URL input doesn't cause overflow.
Here are some screenshots:

image

image



Please let me know if anything needs to be adjusted!

frontend/css/sass/stats-art-creator.scss
```diff
@@ -318,6 +329,12 @@ $color-selector-size: 24px;
       margin-left: 0.6em;
       margin-right: auto;
 
+      @media (max-width: map.get($grid-breakpoints, "sm")) {
+        min-width: 0;
```
MonkeyDo, line 333: Why a min-width of 0?
Wouldn't that potentially collapse it to an invisible size?

Actually, I tried removing all 3 declarations in this block and it looked the same to me. Are they necessary?

frontend/css/sass/stats-art-creator.scss
```diff
@@ -300,13 +301,23 @@ $color-selector-size: 24px;
...
+    .share-buttons {
+      display: flex;
+      align-items: center;
+      justify-content: center;
+
+      @media (max-width: map.get($grid-breakpoints, "sm")) {
+        flex-wrap: wrap;
+        row-gap: 0.5em;
+      }
+    }
```
MonkeyDo, line 320: Unless I missed something, you are redefining a separate `.share-buttons` block here instead of putting the styles in the existing one, and applying `justify-content: center;` twice:

```suggestion
      justify-content: center;
      display: flex;
      align-items: center;

      > .btn-link > span {
        margin-right: 0.2em;
        font-size: 14px;
      }

      @media (max-width: map.get($grid-breakpoints, "sm")) {
        flex-wrap: wrap;
        row-gap: 0.5em;
      }
    }
```

frontend/css/sass/stats-art-creator.scss
```diff
@@ -1,3 +1,4 @@
+@use "sass:map";
 $color-selector-size: 24px;
 
 #stats-art-creator {
```
MonkeyDo, line 4: On mobil;e sizes it might be good to use display:block and set  a max-width of 100% so that the margin on the right side applies.
Currently it's lopsided with more margin on the left:

Image

MonkeyDo (changes requested):
Looking a lot better, thank you for continuing work on this.
Some more changes below:

uzman2406:
Hey @MonkeyDo, I will continue working on this PR after May 28th once my end-sem exams are over.

PR #3611, opened 2026-03-04
LB-1941 : Show playlist track collage as opengraph image

Shreshtthh:
Problem- 
LB-1941
When playlist links are shared on social media (Twitter, Discord, Slack, etc.), the link preview shows a generic/blank image. Playlists should have a composed OG image that combines track cover art with ListenBrainz branding for better social sharing.

Solution
- Added a new public GET endpoint /1/art/playlist/<mbid>/og/ that generates a composed 1280×640 PNG on the fly:
- Downloads up to 4 track cover art images from the Cover Art Archive
- Arranges them in a 2×2 grid (left ~72% of the canvas), or a single image if < 4 available
- Composites the ListenBrainz logo overlay on the right side
- Falls back to share-header.png via redirect if no cover art is available or generation fails
- Also set og:image meta tags on playlist pages pointing to this endpoint.

Files changed:
- listenbrainz/art/og_image.py- new image composition module (uses Pillow)
- listenbrainz/webserver/views/art_api.py - new endpoint
- listenbrainz/webserver/views/playlist.py - og:image meta tag
- requirements.txt - added Pillow
- frontend/img/og-overlay.png - overlay asset
- listenbrainz/webserver/views/test/test_playlist_og_image.py - 19 tests

AI Usage
Used AI for navigating unfamiliar parts of the codebase. All the code changes were written, reviewed and tested by me.

Action
- New dependency `Pillow==11.2.1` will be installed on deployment (already in `requirements.txt`)
- Manual verification needed: My local dev DB didn't have MusicBrainz metadata populated, so the endpoint always fell back to `share-header.png`. I verified the image composition by temporarily adding a test endpoint with hardcoded cover art URLs from archive.org. Could a reviewer please manually test the real endpoint with a public playlist that has tracks with cover art on staging? - `GET /1/art/playlist/<playlist_mbid>/og/` should return a composed PNG.

image


MonkeyDo:
Can confirm it works:

image

listenbrainz/webserver/views/art_api.py
```diff
@@ -821,3 +822,72 @@ def playlist_cover_art_generate(playlist_mbid, dimension, layout):
                            show_caption=False), 200, {
                                'Content-Type': 'image/svg+xml'
                            }
+
+
+@art_api_bp.get("/playlist/<uuid:playlist_mbid>/og/")
+@crossdomain
```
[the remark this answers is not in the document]
MonkeyDo: What he said.

listenbrainz/webserver/views/playlist.py
```diff
@@ -32,7 +32,9 @@ def playlist_page(playlist_mbid: str):
                 "type": "music:playlist",
                 "url": f'{current_app.config["SERVER_ROOT_URL"]}/playlist/{playlist_mbid}',
                 "music:creator": f'{current_app.config["SERVER_ROOT_URL"]}/user/{playlist.creator}',
-                # "image": Once we have playlist images we can try adding it here
+                "image": f'{current_app.config["SERVER_ROOT_URL"]}/1/art/playlist/{playlist_mbid}/og/',
+                "image:width": "1280",
+                "image:height": "640",
```
MonkeyDo: What are those based on? Do you have a reference I can look at?

First two search results I found suggest 1200*630 instead, but TBH I know nothing on that topic:
https://ogimage.io/resources/og-image-size
https://myogimage.com/blog/og-image-size-meta-tags-complete-guide

They should use the OG_WIDTH and height variables defined in og_image.py to avoid duplication

listenbrainz/webserver/views/test/test_playlist_og_image.py
```diff
@@ -0,0 +1,310 @@
...
+        # Mock the image generation to return a valid PNG
+        dummy_png_buf = io.BytesIO(_create_dummy_png(OG_WIDTH, OG_HEIGHT))
+        mock_generate_og.return_value = dummy_png_buf
+
+        resp = self.client.get(
+            self.custom_url_for('art_api_v1.playlist_og_image',
+                                playlist_mbid="b757afbf-1b6a-4bd1-9d3f-2ad9cac9c3d6"))
+        self.assert200(resp)
+        self.assertEqual(resp.content_type, "image/png")
+        self.assertTrue(resp.data[:4] == b'\x89PNG')
```
MonkeyDo, line 338: Would be nice to document this, I know there is a comment `PNG magic bytes` somewhere above but it's not particularly clear, and not here where this assertion is made.

Generally the comments shouldn't describe what is happening, which is usually fairly obvious (such as "Verify cache header"), but rather explain *why* we are doing it.

The more esoteric parts like this `resp.data[:4] == b'\x89PNG'` could really do with some explanation, but are left to the reader to figure out.

Which brings me to another question:
>All the code changes were written, reviewed and tested by me.

Is that true also for these 300 lines of test code?
Because they look like LLM comments to me.

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,153 @@
...
+            return None
+        canvas = _compose_single(img)
+
+    # slap the overlay on top
+    canvas = Image.alpha_composite(canvas, overlay)
+    canvas = canvas.convert("RGB")
+
+    buf = io.BytesIO()
+    canvas.save(buf, format="PNG", optimize=True)
+    buf.seek(0)
```
MonkeyDo, line 201: Some comments here for some steps would be useful.

Why the conversion to RGB?
Why the buf.seek(0)?

MonkeyDo (changes requested):
Looking pretty good, thanks for the feature!

Shreshtthh:
Made the changes. The image dimensions were based on [GitHub's Documentation](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/customizing-your-repositorys-social-media-preview) but I've changed the size to 1200x630 now to align with the OpenGraph docs.

Also improved the comments mentioning why instead of what. The `magic bytes` thing was that the RFC states that the first bytes of a PNG must be 0x89 followed by "PNG" (The exact wording of the spec is that the first bytes of a PNG should be decimal values 137, 80, 78, and 71 which translate to \x89PNG). Also added this in the comments now.

About the test file comments - I actually did use an AI coding assistant to help with scaffolding and boilerplate (including the tests), then reviewed, iterated on, and tested everything myself. The overly descriptive comment style was a leftover from that process. I've cleaned them up now and improved them so they are much standard now!

MonkeyDo:
Thanks for the clarifications.

Regarding AI usage, I figured.
I want to point out that your description says 
>Used AI for navigating unfamiliar parts of the codebase. All the code changes were written, reviewed and tested by me.

which is not what you are saying now, after being caught.
It's not the most egregious of lies, but when we ask you to disclose AI usage it is so that we know how much confidence to place in the generated code and how thoroughly to review. We would trust **you** not to hallucinate random code, less so LLM output.

On a personal level though, I will say I really don't like being lied to, and it is disappointing to have to catch you in the act.
If you use AI, own it and be upfront about it for everyone's sake, especially if disclosure is requested. It is way worse to be caught in a lie than to admit to using LLM to work faster.

Shreshtthh:
Hi @MonkeyDo
I didn't intend to be deceptive, but I realize now that saying "all code was written by me" was factually incorrect since I used AI for the boilerplate and scaffolding. Because I spent a lot of time manually reviewing, iterating, and cross-referencing the tests, I didn't think through how the initial generation affects the review process.
I didn't mean any harm, but I do take full responsibility for the misrepresentation. I will be 100% upfront about any AI usage in all future contributions.

Thank you for your time reviewing this feature

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,157 @@
...
+from concurrent.futures import ThreadPoolExecutor
+from pathlib import Path
+
+import requests as req
+from PIL import Image
+
+logger = logging.getLogger(__name__)
+
+OG_WIDTH = 1200
+OG_HEIGHT = 630
```
MonkeyDo: For a bit more clarity, especially if imported in another file:

```suggestion
OPENGRAPH_IMAGE_WIDTH = 1200
OPENGRAPH_IMAGE_HEIGHT = 630
```

You will obviously need to change this wherever those constants are used.

listenbrainz/webserver/views/playlist.py
```diff
@@ -8,6 +8,7 @@
 import listenbrainz.db.playlist as db_playlist
 from listenbrainz.db.model import playlist as model_playlist
 from listenbrainz.art.cover_art_generator import CoverArtGenerator
+from listenbrainz.art.og_image import OG_WIDTH, OG_HEIGHT
```
MonkeyDo:
```suggestion
from listenbrainz.art.og_image import OPENGRAPH_IMAGE_WIDTH, OPENGRAPH_IMAGE_HEIGHT
```

listenbrainz/webserver/views/playlist.py
```diff
@@ -32,7 +33,9 @@ def playlist_page(playlist_mbid: str):
                 "type": "music:playlist",
                 "url": f'{current_app.config["SERVER_ROOT_URL"]}/playlist/{playlist_mbid}',
                 "music:creator": f'{current_app.config["SERVER_ROOT_URL"]}/user/{playlist.creator}',
-                # "image": Once we have playlist images we can try adding it here
+                "image": f'{current_app.config["SERVER_ROOT_URL"]}/1/art/playlist/{playlist_mbid}/og/',
+                "image:width": str(OG_WIDTH),
+                "image:height": str(OG_HEIGHT),
```
MonkeyDo:
```suggestion
                "image:width": str(OPENGRAPH_IMAGE_WIDTH),
                "image:height": str(OPENGRAPH_IMAGE_HEIGHT),
```

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,157 @@
...
+
+OVERLAY_PATH = Path(__file__).resolve().parent.parent.parent / "frontend" / "img" / "og-overlay.png"
+OVERLAY_PATH_PROD = Path("/static/img/og-overlay.png")
+
+
+def _get_overlay_path():
+    """Check /static/img first (production), fall back to frontend/ (local dev)."""
+    if OVERLAY_PATH_PROD.exists():
+        return OVERLAY_PATH_PROD
+    return OVERLAY_PATH
```
MonkeyDo: Could this use the existing tool we use elsewhere `url_for('static', ....`?

You use it in art_api.py at line 852.

listenbrainz/webserver/views/art_api.py
```diff
@@ -821,3 +822,71 @@ def playlist_cover_art_generate(playlist_mbid, dimension, layout):
                            show_caption=False), 200, {
                                'Content-Type': 'image/svg+xml'
                            }
+
+
+@art_api_bp.get("/playlist/<uuid:playlist_mbid>/og/")
```
MonkeyDo: After some more thought, I figure we should spell this out:

```suggestion
@art_api_bp.get("/playlist/<uuid:playlist_mbid>/opengraph/")
```

listenbrainz/webserver/views/playlist.py
```diff
@@ -32,7 +33,9 @@ def playlist_page(playlist_mbid: str):
                 "type": "music:playlist",
                 "url": f'{current_app.config["SERVER_ROOT_URL"]}/playlist/{playlist_mbid}',
                 "music:creator": f'{current_app.config["SERVER_ROOT_URL"]}/user/{playlist.creator}',
-                # "image": Once we have playlist images we can try adding it here
+                "image": f'{current_app.config["SERVER_ROOT_URL"]}/1/art/playlist/{playlist_mbid}/og/',
```
MonkeyDo: This will need to be updated too after the above comment

MonkeyDo (changes requested):
Some final comments I think.

I also wanted to say I was this close to closing the PR after the last comments.
After a blatantly incorrect disclosure I can't take you at your word when you say "I didn't intend to be deceptive".

That being said, you seem to reason with the code, so after another round of thorough review I will merge this.
Please apply your own skills and your own brain going forward, that is what we are here to evaluate and to mentor, not an LLM.

Shreshtthh:
Not sure what to say except thank you for choosing to mentor me regardless
I've made all the changes addressing all the review feedback. Regarding the overlay path, I couldn't use `url_for('static', ...)` directly because the overlay is opened by Pillow via `Image.open()`, which requires a filesystem path, not a URL. `url_for('static', filename='img/og-overlay.png')` returns a URL string like `/static/img/og-overlay.png` or `https://listenbrainz.org/static/img/og-overlay.png`, which Pillow can't open.
Instead, I used `current_app.static_folder` which gives us the actual filesystem path to Flask's static directory (e.g., `/static` in production). This is essentially the same mechanism Flask uses internally to resolve `url_for('static', ...)`, just the file path version instead of the URL version. It also handles the dev/production difference cleanly: in production `static_folder` points to `/static,` and outside a Flask context (e.g., tests) it falls back to the local `frontend/img/ path`.

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,163 @@
...
+
+OPENGRAPH_IMAGE_WIDTH = 1200
+OPENGRAPH_IMAGE_HEIGHT = 630
+
+# cover art takes up the left ~72% of the canvas,
+# the rest is for the LB overlay branding
+GRID_WIDTH = int(OPENGRAPH_IMAGE_WIDTH * 0.72)  # 921px
+
+# Local dev fallback: frontend/img lives alongside the source tree
+_OVERLAY_PATH_DEV = Path(__file__).resolve().parent.parent.parent / "frontend" / "img" / "og-overlay.png"
```
MonkeyDo, line 19: Thanks for the explanation regarding getting the actual file path, I missed that.

I think this code with the parent.parent.parent is brittle and could break if we move files around.

There has to be a better way to access a specific file that is more declarative than this.
Like we do here for example:
https://github.com/metabrainz/listenbrainz-server/blob/1e1889fefb445d08421bcddf920a0c19a7ebe2d5/listenbrainz/webserver/static_manager.py#L6

listenbrainz/webserver/views/test/test_playlist_og_image.py
```diff
@@ -0,0 +1,296 @@
...
+        result = generate_playlist_og_image(
+            ["https://example.com/art1.jpg"],
+            overlay_path=overlay_path,
+        )
+        assert result is not None
+
+        result_img = Image.open(result)
+        assert result_img.size == (OPENGRAPH_IMAGE_WIDTH, OPENGRAPH_IMAGE_HEIGHT)
+        assert result_img.mode == "RGB"
+
```
MonkeyDo, line 120: Missing an assertion to justify the test here:
 assert mock_download.call_count == 1

listenbrainz/webserver/views/test/test_playlist_og_image.py
```diff
@@ -0,0 +1,296 @@
...
+        assert mock_download.call_count == 1
+
+    @patch("listenbrainz.art.og_image._download_image")
+    def test_download_failure_with_four_urls_falls_back_to_single(self, mock_download, tmp_path):
+        """If some of the 4 downloads fail, fall back to single image."""
+        overlay_path = self._create_overlay_file(tmp_path)
+        good_img = Image.new("RGBA", (500, 500), (255, 0, 0, 255))
+        # First one succeeds, rest fail
+        mock_download.side_effect = [good_img, None, None, None]
```
MonkeyDo, line 154: Would add another good_image  so it's two good ones, two fails, for a better test.
You're also missing an assertion for the number of download calls at the end.

listenbrainz/webserver/views/test/test_playlist_og_image.py
```diff
@@ -0,0 +1,296 @@
...
+    @patch("listenbrainz.art.og_image._download_image")
+    def test_all_downloads_fail_returns_none(self, mock_download, tmp_path):
+        overlay_path = self._create_overlay_file(tmp_path)
+        mock_download.return_value = None
+
+        result = generate_playlist_og_image(
+            ["https://example.com/art1.jpg"],
+            overlay_path=overlay_path,
+        )
+        assert result is None
```
MonkeyDo, line 172: Add an assertion for call_count == 0

listenbrainz/webserver/views/test/test_playlist_og_image.py
```diff
@@ -0,0 +1,296 @@
...
+
+    def test_missing_overlay_returns_none(self):
+        result = generate_playlist_og_image(
+            ["https://example.com/art1.jpg"],
+            overlay_path="/nonexistent/path/overlay.png",
+        )
+        assert result is None
```
MonkeyDo, line 179: Same here

MonkeyDo (changes requested):

listenbrainz/webserver/views/test/test_playlist_og_image.py
Shreshtthh: hello, I set the assertion to mock_download.call_count == 1 instead of 0 here. This is since this test verifies the fallback behavior after a download attempt fails, it executes exactly 1 time before returning None.

Shreshtthh:
Thanks for the feedback. I've made all the improvements

listenbrainz/webserver/views/test/test_playlist_og_image.py
```diff
@@ -0,0 +1,303 @@
...
+        mock_download.side_effect = [good_img, good_img, None, None]
+
+        urls = [f"https://example.com/art{i}.jpg" for i in range(4)]
+        result = generate_playlist_og_image(urls, overlay_path=overlay_path)
+        assert result is not None
+
+        result_img = Image.open(result)
+        assert result_img.size == (OPENGRAPH_IMAGE_WIDTH, OPENGRAPH_IMAGE_HEIGHT)
+
+        assert mock_download.call_count == 4
```
MonkeyDo, line 254: This does not actually test that it "fall back to single image"
How could you test that `_compose_single` is called rather than `_compose_grid_2x2` ?

listenbrainz/webserver/views/test/test_playlist_og_image.py
MonkeyDo: Makes sense, thanks for the explanation.

MonkeyDo (commented):
Just some small comment regarding tests

listenbrainz/webserver/views/test/test_playlist_og_image.py
Shreshtthh: Sorry, did not end up getting the notification for this somehow and missed it. I’ve updated the fallback tests by patching `_compose_single` and `_compose_grid_2x2` with wraps, so the real functions still run while tracking calls. The tests now verify that `_compose_single` is used and `_compose_grid_2x2` is not in the fallback case.

Shreshtthh:
All fixed!

listenbrainz/webserver/views/playlist.py
```diff
@@ -32,7 +33,9 @@ def playlist_page(playlist_mbid: str):
                 "type": "music:playlist",
                 "url": f'{current_app.config["SERVER_ROOT_URL"]}/playlist/{playlist_mbid}',
                 "music:creator": f'{current_app.config["SERVER_ROOT_URL"]}/user/{playlist.creator}',
-                # "image": Once we have playlist images we can try adding it here
+                "image": f'{current_app.config["SERVER_ROOT_URL"]}/1/art/playlist/{playlist_mbid}/opengraph/',
```
MonkeyDo: This is't working I realized after testing it deployed on the test website.
This points to `https://listenbrainz.org/1/art/playlist/<UUID>/opengraph/` (or in this case `test.listenbrainz.org/1/art/...`), but this is not a reachable page as the API is on a different subdomain (`api.listenbrainz.org/1/...`  and for test: `test-api.listenbrainz.org/1/...`)


```suggestion
                "image": f'{current_app.config["API_URL"]}/1/art/playlist/{playlist_mbid}/opengraph/',
```
Or maybe even using `from listenbrainz.webserver import API_PREFIX`

```suggestion
                "image": f'{current_app.config["API_URL"]}{API_PREFIX}/art/playlist/{playlist_mbid}/opengraph/',
```

We ran into that issue before:
https://github.com/metabrainz/listenbrainz-server/blob/9dc4434a594cb743e5c304c750c82316d36e6472/listenbrainz/webserver/views/atom.py#L879-L882

MonkeyDo (changes requested):
I did some manual testing and it looks to be working fine for the most part, good work!
We're almost there, just a couple of things to iron out.

I don't know why, but this playlist with many tracks returns an image with just one album, something to investigate. The playlist page shows a collage with many covers:
https://test.listenbrainz.org/playlist/3924b7dc-9fac-4b8b-8916-51647cfa3ae3/

Image

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,150 @@
...
+from PIL import Image
+
+logger = logging.getLogger(__name__)
+
+OPENGRAPH_IMAGE_WIDTH = 1200
+OPENGRAPH_IMAGE_HEIGHT = 630
+
+# cover art takes up the left ~72% of the canvas,
+# the rest is for the LB overlay branding
+GRID_WIDTH = int(OPENGRAPH_IMAGE_WIDTH * 0.72)  # 921px
```
anshg1214: 1200 * 0.72 = 864. You should update the comment here.

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,150 @@
...
+        overlay = overlay.resize((OPENGRAPH_IMAGE_WIDTH, OPENGRAPH_IMAGE_HEIGHT), Image.LANCZOS)
+    except Exception:
+        logger.error("Failed to load OG overlay from %s", overlay_path, exc_info=True)
+        return None
+
+    if len(cover_urls) >= 4:
+        # download all 4 concurrently to keep latency reasonable
+        with ThreadPoolExecutor(max_workers=4) as pool:
+            downloaded = list(pool.map(_download_image, cover_urls[:4]))
+        downloaded = [img for img in downloaded if img is not None]
```
anshg1214: This is risky. With many concurrent requests, you can end up with `N requests × 4` threads in addition to your WSGI worker threads. If you have 200 concurrent requests, that can mean ~800 extra threads, plus sockets, plus memory pressure. 

Instead use a single shares executor. Something like 
```
_OG_EXECUTOR = ThreadPoolExecutor(max_workers=… )
```
This caps total download worker threads per process, regardless of request concurrency.
We can switch to a process-wide shared executor and add a global semaphore / bounded concurrency for cover downloads (plus a per-request deadline), so total in-flight fetches are capped and the endpoint degrades gracefully under load?

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,150 @@
...
+OVERLAY_PATH = os.path.join("/", "static", "img", "og-overlay.png")
+
+# Timeout for downloading cover art images (seconds)
+DOWNLOAD_TIMEOUT = 5
+
+
+def _download_image(url):
+    """Grab an image from a URL, return as PIL Image or None if anything goes wrong."""
+    try:
+        resp = req.get(url, timeout=DOWNLOAD_TIMEOUT)
```
anshg1214: Can we allow streaming?

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,150 @@
...
+    # slap the overlay on top
+    canvas = Image.alpha_composite(canvas, overlay)
+    # drop the alpha channel. The overlay is already composited,
+    # and PNG-without-alpha is smaller / friendlier to social-media crawlers
+    # plus transparent pngs aren't supported by all social media platforms but RGB is
+    canvas = canvas.convert("RGB")
+
+    buf = io.BytesIO()
+    canvas.save(buf, format="PNG", optimize=True)
+   # Reset the file pointer to the beginning of the stream so the image data can be read
```
anshg1214: Indentation
```suggestion
    # Reset the file pointer to the beginning of the stream so the image data can be read
```

listenbrainz/art/og_image.py
anshg1214: Consider basic content-type allowlist (e.g. image/jpeg, image/png)

anshg1214 (changes requested):
Since it's an expensive endpoint, we can consider using ETag here for client side caching.

listenbrainz/art/og_image.py
Shreshtthh: I remember manually getting the width to around 921 pixels, and then I ended up changing the dimensions to 1200 (from 1280). I've changed the multiplier to .77 to make the width around 924, let me know if that looks okay

listenbrainz/art/og_image.py
Shreshtthh: I've removed the per-request executor and replaced it with a shared, module-level `_OG_EXECUTOR = ThreadPoolExecutor(max_workers=8)`

listenbrainz/art/og_image.py
Shreshtthh: Both done

listenbrainz/webserver/views/playlist.py
Shreshtthh: All done

Shreshtthh: About the playlist with many tracks returning an image with just one album, I think it might've been that the image failed to load due to nginx 502 or 504. I tried this manually and once in a while that happened.  I've hence added retry logic and we now submit all candidate URLs to the shared pool concurrently. Since we only need 4 to succeed to build the 2x2 grid, processing extra candidates gives us a much better buffer against flaky archive.org responses.

MonkeyDo: > processing extra candidates gives us a much better buffer

While it does do that, it also means we are consuming a lot more resources, loading a bunch of images we likely will not use. I'd rather serve an image with one cover image rather than load 8 to use 4.

listenbrainz/webserver/views/art_api.py
```diff
@@ -821,3 +823,79 @@ def playlist_cover_art_generate(playlist_mbid, dimension, layout):
                            show_caption=False), 200, {
                                'Content-Type': 'image/svg+xml'
                            }
+
+
+@art_api_bp.get("/playlist/<uuid:playlist_mbid>/opengraph/")
```
anshg1214, line 828: Let's update the PR Description which says `/og/` route

listenbrainz/webserver/views/art_api.py
```diff
@@ -821,3 +823,79 @@ def playlist_cover_art_generate(playlist_mbid, dimension, layout):
...
+            break
+
+    if not cover_art_urls:
+        return redirect(fallback_url)
+
+    # ETag based on the set of cover art URLs. If the playlist's tracks
+    # haven't changed, clients can skip re-downloading the composed image
+    etag = hashlib.md5("|".join(sorted(cover_art_urls)).encode()).hexdigest()
+    if request.headers.get("If-None-Match") == etag:
+        return "", 304
```
anshg1214: ETag check runs after full metadata fetch -  `get_by_mbid(load_recordings=True)` and `fetch_playlist_recording_metadata` run before the If-None-Match check. Crawlers revalidating with ETag still hit DB on every request.

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,198 @@
...
+    if len(cover_urls) >= 4:
+        # download all candidates concurrently using the shared executor;
+        # we only need 4 to succeed for the grid, but archive.org can be
+        # flaky so extra candidates give us a better chance of filling it
+        futures = [_OG_EXECUTOR.submit(_download_image, url) for url in cover_urls]
+        downloaded = []
+        for future in futures:
+            try:
+                img = future.result(timeout=DOWNLOAD_TIMEOUT * 2)
+            except Exception:
```
anshg1214, line 168: `_download_image` can take up to ~11s (5s + 1s sleep + 5s retry) but future.result(timeout=10) treats in-flight retries as failure.

listenbrainz/webserver/views/art_api.py
```diff
@@ -821,3 +823,79 @@ def playlist_cover_art_generate(playlist_mbid, dimension, layout):
...
+
+    # Resolve cover art URLs — collect extras so the grid can still be
+    # composed even if some archive.org downloads fail intermittently
+    cover_art_urls = []
+    for img in images:
+        caa_id = img.get("caa_id")
+        caa_release_mbid = img.get("caa_release_mbid")
+        if caa_id and caa_release_mbid:
+            url = f"https://archive.org/download/mbid-{caa_release_mbid}/mbid-{caa_release_mbid}-{caa_id}_thumb500.jpg"
+            cover_art_urls.append(url)
```
anshg1214: Let's reuse `CoverArtGenerator.resolve_cover_art(caa_id, caa_release_mbid, 500)`

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,198 @@
...
+        else:
+            return None
+    else:
+        # < 4 urls, just use the first
+        img = _download_image(cover_urls[0])
+        if img is None:
+            return None
+        canvas = _compose_single(img)
+
+    # slap the overlay on top
```
anshg1214: When `len(cover_urls) < 4`, we only try `cover_urls[0]`. 

Let's iterate until one download succeeds

listenbrainz/art/og_image.py
```diff
@@ -0,0 +1,198 @@
...
+            try:
+                img = future.result(timeout=DOWNLOAD_TIMEOUT * 2)
+            except Exception:
+                img = None
+            if img is not None:
+                downloaded.append(img)
+
+        if len(downloaded) >= 4:
+            canvas = _compose_grid_2x2(downloaded[:4])
+        elif downloaded:
```
anshg1214, line 175: Once `len(downloaded) >= 4`, we don't need to wait on the remaining futures. But the loop still blocks on all 8 in submission order. That holds the worker longer than necessary.

Let's break early once we have 4 successful downloads.

anshg1214 (changes requested):

Shreshtthh:
hi @anshg1214 @MonkeyDo, just realized I didn't drop a comment after making the changes. Please check them and let me know if there's something more to be done. Thanks!

MonkeyDo (approved):
Thanks for this cool feature!

PR #3619, opened 2026-03-06
LB-1141: Indicate which listens are marked for deletion

SomishaAdak:
**Problem**
The staged_for_deletion  listens on fetch_listens() don't actually show any indication of their state, so frontend cannot distinguish between regular and pending deletion listens.

**Solution**
Extend fetch_listens() to join against listen_delete_metadata with status = 'pending' and expose a boolean "staged_for_deletion" flag, passed via Listen.from_timescale() added to Listen object.

**Action**
For timescale_listenstore.py :
- Join listen_delete_metadata with status = 'pending in fetch_listens()
- Add boolean (ldm.id IS NOT  NULL) as staged_for_deletion in query
- Add "staged_for_deletion" in select and group by clauses
- pass flag to Listen.from_timescale()

For listen.py :
- Add "staged_for_deletion" attribute to Listen object

Ref: LB-1141

### DIFF
--- a/listenbrainz/webserver/views/import_listens.py
+++ b/listenbrainz/webserver/views/import_listens.py
@@ -58,7 +58,7 @@ def create_import_task():
         raise APIBadRequest("No service selected!")
     service = service.lower()
 
-    allowed_services = ["spotify", "listenbrainz", "librefm", "maloja", "panoscrobbler", "audioscrobbler"]
+    allowed_services = ["spotify", "listenbrainz", "librefm", "maloja", "panoscrobbler", "audioscrobbler", "spinitron"]
     if service not in allowed_services:
         raise APIBadRequest("This service is not supported!")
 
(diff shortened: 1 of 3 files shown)


SomishaAdak:
I tried running tests locally but there was some issues installing uWSGI on windows while installing dependencies,wanted to confirm if there's a another way to run tests in this repo that is recommended.

MonkeyDo:
Are you following this guide? https://listenbrainz.readthedocs.io/en/latest/developers/devel-env.html
Are you using docker?
We don't necessarily recommend installing and running everything directly on your own machine.

SomishaAdak:
Thanks for the guidance ill make sure to run tests fully on docker now to make lesser unnecessary commits. Also, i noticed LB-1255 is related to 1141, and aerozol mentioned about combining the two so just wanted to confirm about the current implementation and if there is anything added to the scope.

listenbrainz/webserver/views/import_listens.py
MonkeyDo, line 1: This does not seem related to listen deletion. Should this be a separate PR?

SomishaAdak: Right, i ended up doing that as i was tracing how listens move through the imports when adding the new flag...I'll remove it from this PR

MonkeyDo: Please be careful about the code you commit, and review it before pushing your commits

MonkeyDo: There is still more unrelated code commited in this file.

PR #3627, opened 2026-03-10
LB-1878: Standardize recording_msid in JSONL listen export

Harizz076:
# Problem
The JSONL export format was inconsistent with the `/1/user/<name>/listens` API response and the Inspect page format because the `recording_msid` attribute was structured differently (nested inside `track_metadata.additional_info`) and `inserted_at` was exported as a float epoch rather than an integer.
Fixes LB-1878
# Solution
- Promoted `recording_msid` to a top-level field in the exported JSONL to match API and Inspect response formats.
- Used `json_build_object` in [export_listens_for_time_range](cci:1://file:///home/haris/listenbrainz-server/listenbrainz/background/export.py:80:0-165:118) for alphabetical key ordering: `inserted_at`, `listened_at`, `recording_msid`, `track_metadata`.
- Cast `inserted_at` to an integer (was float).
- Removed `recording_msid` from `track_metadata.additional_info` inside `Listen.from_timescale` to avoid duplication in API responses now that it's correctly selected as a top-level field.
- ~Fixed [export_query_to_jsonl](cci:1://file:///home/haris/listenbrainz-server/listenbrainz/background/export.py:61:0-77:19) to correctly handle dict rows returned from `json_build_object` using `orjson.dumps` instead of string writes.~
- Added a volume mount to the `timescale_writer` in [docker-compose.test.yml](cci:7://file:///home/haris/listenbrainz-server/docker/docker-compose.test.yml:0:0-0:0) so it correctly picks up live source code (and [config.py](cci:7://file:///home/haris/listenbrainz-server/listenbrainz/config.py:0:0-0:0) fixes), consistent with other test services.
- Updated [test_export.py](cci:7://file:///home/haris/listenbrainz-server/listenbrainz/tests/integration/test_export.py:0:0-0:0) to wait for 3 listens to appear in timescale and added explicit assertions to verify the top-level `recording_msid` and integer `inserted_at`.
# AI usage
* [ ] I did not use any AI
#### If you did use AI:
> Used Claude Sonnet 4.6 to help write the integration tests and corresponding assertions in [test_export.py](cci:7://file:///home/haris/listenbrainz-server/listenbrainz/tests/integration/test_export.py:0:0-0:0) to verify the JSON data output formatting.
# Action
No special post-merge deployment actions are required.


Harizz076:
Remember this being  a pretty small change, maybe it flew under the radar cuz it was during the GSOC proposal period. 

Anyways, @MonkeyDo tagging you here for a review.

Harizz076:
@MonkeyDo : Sorry for tagging you again but can you please look at this PR and LMK if this can be merged? thanks.

Harizz076:
@MonkeyDo : Thanks for running the tests, I got a notif that they failed so I checked and realised that the test_api.py file needed a small fix. Ive ran the entire suite this time and im sure that they wont fail the workflow. Can you please run the test suite again and review? sorry for the inconvenience.

MonkeyDo:
I will start by asking if you generated a dump in your local environment to ensure all the modified code paths were hit and that they run with no issues

Harizz076:
Hi monkey, no i had not done that earlier, i thought one of the tests did cover this. I did now, it seems to be working perfectly.

listenbrainz/background/export.py
```diff
@@ -65,11 +65,15 @@ def export_query_to_jsonl(conn, file_path, query, **kwargs):
...
         for partition in result.partitions():
             for row in partition:
-                file.write(row.line)
-                file.write("\n")
+                line = row.line
+                if isinstance(line, dict):
+                    file.write(orjson.dumps(line))
+                else:
+                    file.write(line.encode("utf-8"))
+                file.write(b"\n")
```
MonkeyDo, line 90: Can you explain the need for writing binary instead of text?

listenbrainz/background/export.py
```diff
@@ -95,14 +99,16 @@ def export_listens_for_time_range(ts_conn, file_path, user_id: int, start_time:
...
-                            'listened_at'
+                SELECT json_build_object(
+                            'inserted_at'
+                          , (extract(epoch from inserted_at))::integer
+                          , 'listened_at'
                           ,  extract(epoch from listened_at)
-                          , 'inserted_at'
-                          ,  extract(epoch from inserted_at)
+                          , 'recording_msid'
+                          , recording_msid
```
MonkeyDo: Does this need to be cast to text with `::text` ?
It is of type `UUID` in the database

MonkeyDo (commented):
Nice work so far :)

You will also need to update the front-end code to expect the recording_msid at the top level, and deprecated at the additional_info level.

For starters, search for `track_metadata.additional_info.recording_msid` in the frontend.
There is a utility getRecordingMSID that should support both (first top level, then nested in track metadata for compatibiliy).
Then the other instances should really be using this getRecordingMSID helper instead of whatever they were doing.
Finally, broaden the search to `recording_msid` in the frontend folder (ignore .json files and .test.* files) and check that ther'es no other place in the code that will break with those changes. You will find for example that the typescript types need to be changed as well.

listenbrainz/background/export.py
Harizz076: It isnt "needed" per se, but i noticed that other UUID types for eg recording_mbid were casted to string so i thought it should be consistent with them, should i remove it..?

listenbrainz/background/export.py
Harizz076: I used orjson here due to its speed and lower memory usage for large exports. Because orjson returns bytes (not a string), the file must be opened in binary mode so we can write those bytes directly.

Harizz076:
@MonkeyDo :Hi! can you please run the gh workflows and review the changes. Thanks :)

listenbrainz/background/export.py
Harizz076: I switched to orjson since it's a lot faster. It returns bytes instead of a string, so we have to save the file in binary.

MonkeyDo: I'm not sure about this.
If I follow, listens rows are now materialized as python dicts then re-serialized with orjson afterwards.
For large exports that might slow things down with the extra JSON decode/encode work per row.
I'm not sure whether that will be significative, but I'd like to see some benchmarking before jumping to conclusions.
Can you find a good way to benchmark this?

Harizz076: Codex helped generate and run this benchmark.

I compared the full listens-export path using 100,000 synthetic listens in the local Timescale test database. Each variant had two warm-up runs followed by seven alternating measured runs. The benchmark included query execution, PostgreSQL/psycopg conversion, JSON serialization where applicable, and writing the JSONL file.

| Export path | Median time | Median throughput | Median Python CPU | Output size |
| --- | ---: | ---: | ---: | ---: |
| `json_build_object` -> Python `dict` -> `orjson.dumps` | 6.84 s | 14,615 rows/s | 2.42 s | 45.5 MB |
| `json_build_object(...)::text` -> direct write | 5.91 s | 16,909 rows/s | 0.43 s | 49.1 MB |

Returning JSON text directly was about 16% higher-throughput and used about 82% less Python CPU. PostgreSQL's text JSON was about 8% larger due to whitespace. I canonicalized every emitted JSONL record and compared SHA-256 hashes: all runs from both paths were semantically identical.


Based on this, using `json_build_object(...)::text` for listens exports and cast `inserted_at` to `integer` in SQL will be better.... Thank you mr. monkey for pointing it out

MonkeyDo: Great, thanks for benchmarking!
I've learned a long time ago that I can't predict how fast a complex task will be compared to another method, no matter my intuition.
Thanks for taking the time to verify this, glad that it wasn't in vain.
I'll do a final review pass but I think that was my last holdout on this PR, the rest looked good to me.

listenbrainz/background/export.py
```diff
@@ -114,14 +121,16 @@ def export_listens_for_time_range(ts_conn, file_path, user_id: int, start_time:
                    AND listened_at <= :end_time
                    AND l.user_id = :user_id
           )
-                SELECT jsonb_build_object(
-                            'listened_at'
+                SELECT json_build_object(
+                            'inserted_at'
+                          , extract(epoch from inserted_at)::integer
+                          , 'listened_at'
                           ,  extract(epoch from listened_at)
```
MonkeyDo, line 128: Just checking that the `listened_at` does not need the same treatment of being cast as an integer

MonkeyDo: Never mind, I requested an export of my data to check this myself, which I could have done before posting here :P
For posterity, current exports look like:
```json
{
    "inserted_at": 1751461730.501592,
    "listened_at": 1751461699,
    "track_metadata": {
        "track_name": "Houdini",
        "artist_name": "Fuckin Whatever",
        "mbid_mapping": {
            "caa_id": 32304004882,
            "artists": [
                {
                    "artist_mbid": "a707e757-128f-4c6a-898f-bdab422c3ab5",
                    "join_phrase": "",
                    "artist_credit_name": "Fuckin Whatever"
                }
            ],
            "artist_mbids": [
                "a707e757-128f-4c6a-898f-bdab422c3ab5"
            ],
            "release_mbid": "9bb0c12b-6ac0-4df1-bcc2-16f9c8e2ea05",
            "recording_mbid": "a6c673b7-5754-497a-8e05-3373785782fd",
            "recording_name": "Houdini",
            "caa_release_mbid": "9bb0c12b-6ac0-4df1-bcc2-16f9c8e2ea05"
        },
        "release_name": "Houdini",
        "recording_msid": "829cdba7-bc58-4c98-81dc-3c9e77f9ef53",
        "additional_info": {
            "duration_ms": 222356,
            "submission_client": "Pano Scrobbler",
            "submission_client_version": "3.45 - 2025, Mar 16"
        }
    }
}
```

MonkeyDo (approved):
Thanks for fixing this!

PR #3634, opened 2026-03-11
LB-1651: Add granular caption options and color customization to Art Creator grid stats

Gopal7387:
## Summary

This update adds a feature to customize captions on the Album Grid stats template. Users can now turn on or off each caption field. Choose their own colors.

### Backend

- The `listen_count` info from user release stats is now sent to the cover art image pipeline in `cover_art_generator.py`.

- The `/1/art/grid-stats/` endpoint in `art_api.py` now accepts two query parameters: `caption-text-color` and `caption-bg-color`.

- These color values are passed on to the SVG template.

### SVG Templates

- Each caption element now has its own CSS class: `.caption-rank` `.caption-release` `.caption-artist` `.caption-listen-count` in `macros.j2`.

- The `loop.index` is sent as the rank to the `render_image_link` macro in `simple-grid.svg`.

- The inline `fill` color from query parameters is applied to each `tspan` and background `rect`.

### Frontend

- The single "Show caption"  toggle is replaced with four toggles:  Show Rank, Show Release Title, Show Artist, Show Listen Count.

- Users can now pick their own caption text color and background color.

- The `showCaption` option is now based on the options: `showRank || showRelease || showArtist || showListenCount`.

- The color options are sent to the grid-stats API URL as query parameters.

- The `ToggleOption` component is refactored into a controlled component with `checked` and `onChange` props.

- `Preview.tsx` now accepts visibility for each field via CSS class targeting.

## Defaults

- Show Rank: off

- Show Release on

- Show Artist: on

- Show Listen Count: off

- Caption text color: #ffffff

- Caption background color: #0000007a

## Not included

- A font selection dropdown is not added yet. This needs to confirm which fonts are available in the SVG renderer.

## Testing

- Each field can be toggled on or off independently. Shows or hides the correct caption element.

- The color pickers update the preview in time.

- The color values are passed as query parameters to the API and rendered inline in the SVG.

## AI Usage

I got some help from AI during this implementation. It helped with understanding the data flow across Python, Jinja2 and React and, with debugging some SVG rendering issues. I understood every change before committing it. The architecture decisions are entirely based on my thinking, and also file choices and PR structure are my own.

Screenshot 2026-03-12 at 2 30 24 AM


Gopal7387:
I made some changes to the caption layout so it looks like the original plan. Now the strip is set up like a 2×2 grid(small 4 squares):

* Top left square is for the Rank

* Bottom left square is for the number of people who have Listened

* Top right square is for the Release * Bottom right square is, for the Artist

Each of these squares can be turned on or off by itself. The layout still looks neat. Tidy no matter which ones are turned on.

Aerozol:
Looks good (just looking at the screenshot, I haven't tested any behaviour). Some notes:

- I would change "Caption Options" to the same heading style as 'Advanced' and just call it "Captions"
- The spacing above the captions options section looks very big
- You can use the same toggles as in fresh releases, here: https://listenbrainz.org/explore/fresh-releases/ (the spacing and everything should match, and reuse as much CSS/styles as possible)
- I would change the checkbox next to 'skip missing covers' to a toggle as well, to keep it consistent.

Gopal7387:
Hey @Aerozol 
I really appreciate you taking the time to give me feedback. I have made some changes to the commits.

Here is what I did:

- I changed the name of "Caption Options" to "Captions". I used h4 styling so it looks like "Advanced".

- I took out the line that separated the captions section from the rest.

- I changed the "Skip missing covers" checkbox to a toggle button so it looks the same as the buttons.

- I replaced the ToggleOption component that we used in this part of the website with the Switch component that we use in the rest of the website like the fresh-releases toggles so they look exactly the same with the same classes and spacing.

Screenshot 2026-03-12 at 2 02 08 PM
Screenshot 2026-03-12 at 2 02 18 PM

Let me know if I need to make any changes, to the captions or anything else.

Aerozol:
Nice work, the UI looks good to me :)

Gopal7387:
@anshg1214 
Just wanted to circle back on this, as its been visually approved.
So, if you have some time, do check this out.
Thanks 🙂

anshg1214:
I've deployed your PR on [test.listenbrainz.org](https://test.listenbrainz.org/explore/art-creator/?range=this_year&style=grid-stats). 

image


Can we find a way to align the test for Release name and artist name better? Most of the titles are being cut off.

Gopal7387:
Hi @anshg1214 🙂
I made a change to stop the text from getting cut off. I used a filter called truncate, from Jinja on the title and artist name of the image in the caption. Now when the names are really long they will show with three dots at the end of going over the edge. The title and artist name of the image will be shorter now. I used truncate with 18 characters so the title and artist name of the image will have 18 characters before the three dots appear. Let me know if the title and artist name of the image are too short or just right.

Aerozol:
Thanks for deploying to test ansh!

A few notes from me:
- I think the background looks better slightly transparent (so you can see some of the cover art) (regression)
- When 'show rank' and 'show listen count' are disabled then the artist and release title should be left-aligned (regression)
- When rank and count are enabled, there is a huge gap between them and the titles

In general I think it needs to be smarter and move elements around based on what's been enabled, not just hide and show parts. If I've only selected 'release title' then it should be at the bottom left, with a smaller background. Currently there's no point in a user disabling anything, since it looks the same just with less info. Does that make sense?

No notes on the colour UI, nice one! But when I download the file the background is always black (text colours download fine).

Gopal7387:
Hi @anshg1214  can you redeploy on test.listenbrainz.org when you have a moment? I made changes based on Aerozol's feedback.
I cannot check rank. Listen count locally because our dev database has no stats.
So it would be great if you could redeploy the changes to see them visually.

anshg1214:
I've deployed the latest version to test.LB.

You can always use mock data in your backend to test things like these.

Gopal7387:
Hi @Aerozol 
Please have a look at the updated changes.
https://test.listenbrainz.org/explore/art-creator/?range=this_year&style=grid-stats
if there are any further issues or suggestions, just ping me 🙂

Aerozol:
Thanks! Looking better! A few things still:

1. There is still a massive gap between the numbers on the left and the title/artist.
2. When only title/artist are selected, the text gets truncated too early (e.g. we don't need to end the text with '...' as early)
3. Great that title/artist now move to the left if they are the only ones displayed, but there's more combos that should be taken into account and displayed in a tidy manner. For instance, if only one element is displayed (or two that are side by side), the caption bar should be half height.

Example of large gap + placement that should be tidier/half height:
image

Gopal7387:
Hi @Aerozol , pushed a new commit addressing  latest feedback:

Half-height bar (15px) when only a single row of caption info is visible
Wider truncation (~32 chars) when only title/artist are shown and left column is absent
Fixed the gap between left and right columns
show_release and show_artist now properly passed through art_api.py → simple-grid.svg → macro so layout responds correctly to all toggle combos
Please take a look ,when you get a chance.  Thanks 🙂

MonkeyDo:
There an issue with the placement of the titles, they start from the middle
That was point number 1 from aerozol's feedback
image

Not sure if that looked correct to you, but please be thorough when implementing feedback.
We don't have a lot of time to review these PRs and we're wasting some if we're just giving you the same easily visible feedback.

Aerozol:
Has some of this been pushed to beta? I was just making a cover art collage on beta (for myself, not testing or anything) and I saw that it had set the text colour to the colour scheme from the 'designer top 5' design. This is a problem because the 'designer top 5' default text colour is dark purple, which is pretty much invisible on the album art grid template.

I think:

**1. The colour settings here should always default to white on black and
2. the colour settings here should be decoupled from the other image templates - e.g. changing them here should not change them on the other templates, and vice versa.**

Gopal7387:
Hi @MonkeyDo @Aerozol 🙂
I have fixed the gap. The right column x was set to image.x + image.width // 2, 
placing it at the midpoint regardless of how wide the left column 
actually is. Changed it to an image.x + 28, which sits just after the 
rank/listen-count column.
Also, i have tested locally with mock data before pushing. Just let me know if this looks correct on your side as well.

Screenshot 2026-04-22 at 1 50 45 PM

Gopal7387:
Hey @Aerozol 
I have also fixed caption colours bleeding into other templates. 
captionTextColor and captionBgColor were always passed to the Preview component regardless of the template type. They are now only passed when style.type === "grid". So the caption CSS rules are never injected for Designer or LPs templates. 
I have verified via DevTools that .caption text tspan rule is absent on non-grid templates. Just let me know if this looks correct. 😊

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -414,21 +434,45 @@ export default function ArtCreator() {
...
     if (!previewSVGRef?.current) {
       return;
     }
     const { current: svgElement } = previewSVGRef;
-    const { outerHTML } = svgElement;
+    let svgString = svgElement.outerHTML;
+
+    // Inject @font-face into SVG for reliable canvg rendering
+    const fontFile = fontFamily.toLowerCase().replace(/\s+/g, "-");
+    const fontUrl = `${window.location.origin}/static/fonts/${fontFile}.woff2`;
```
MonkeyDo: Current font options are not in fact in this folder yet.
Did you test the fonts functionality at all?

I don't see the styleblock injected in the SVG either, so I'm wondering how this is supposed to work.

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -414,21 +434,45 @@ export default function ArtCreator() {
...
   const onClickDownload = useCallback(async () => {
     if (!previewSVGRef?.current) {
       return;
     }
     const { current: svgElement } = previewSVGRef;
-    const { outerHTML } = svgElement;
+    let svgString = svgElement.outerHTML;
+
+    // Inject @font-face into SVG for reliable canvg rendering
+    const fontFile = fontFamily.toLowerCase().replace(/\s+/g, "-");
```
MonkeyDo: It might make more sense to have `fontOptions` as a dictionnary rather than an array, to allow for adding a filename:
```ts
const fontOptions= {
   "Sintony": "syntony-font-filename.woff2"
   ...
```

Or maybe even the full path, would potentially allow us to load them from google fonts for example
```ts
const fontOptions= {
   "Sintony": "/static/fonts/syntony-font-filename.woff2"
   ...
```

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -437,28 +481,44 @@ export default function ArtCreator() {
...
+      <style>
+        @font-face {
+          font-family: "${fontFamily}";
+          src: url("${fontUrl}") format("woff2");
+          font-weight: normal;
+          font-style: normal;
+        }
+      </style>
+    `;
+    svgString = svgString.replace(/>/, `>${styleBlock}`);
```
MonkeyDo: All this duplicated code is a bad code smell. You should refactor this to avoid duplication

listenbrainz/webserver/templates/art/svg-templates/macros.j2
```diff
@@ -37,18 +37,68 @@
...
+        {% if left_rows == 2 %}
+          <tspan class="caption-rank" x="{{ left_x }}" y="{{ row1_y }}">{% if show_rank and rank is not none %}{{ rank }}{% endif %}</tspan>
+          <tspan class="caption-listen-count" x="{{ left_x }}" y="{{ row2_y }}">{% if show_listen_count and image.listen_count %}{{ image.listen_count }}{% endif %}</tspan>
+        {% else %}
+          {% set single_y = bar_y + bar_height // 2 + 4 %}
+          {% if show_rank and rank is not none %}
+            <tspan class="caption-rank" x="{{ left_x }}" y="{{ single_y }}">{{ rank }}</tspan>
+          {% endif %}
+          {% if show_listen_count and image.listen_count %}
+            <tspan class="caption-listen-count" x="{{ left_x }}" y="{{ single_y }}">{{ image.listen_count }}</tspan>
```
MonkeyDo: Would be nice to have the headphones icon we use in other parts of the  website, to differentiate rank and listen count.

Maybe also adding a "#" in front of the rank would help?

Image

Image

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -209,10 +222,17 @@ export default function ArtCreator() {
 
   const [gridSize, setGridSize] = useState(4);
   const [gridLayout, setGridLayout] = useState(0);
-  const [showCaption, setShowCaption] = useState(true);
```
MonkeyDo, line 212: I think it was useful to have a single toggle to turn captions off entirely.
At the moment I have to toggle 4 different toggles to turn it off.

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -645,9 +742,31 @@ export default function ArtCreator() {
...
+      fontFace
+        .load()
+        .then((loadedFace) => {
+          document.fonts.add(loadedFace);
+        })
+        .catch((e) => {
+          console.error("Failed to load font for preview", e);
+        });
+    }
+  }, [fontFamily]);
```
MonkeyDo, line 906: More partly duplicated code to refactor.

Why is there a need for a separate font loading mechanism?

frontend/js/src/explore/art-creator/components/ToggleOption.tsx
MonkeyDo, line 1: Is this component actually used anywhere?
Not sure what those modifications are, but it looks like we are using a component named Switch rather than this one.
frontend/js/src/components/Switch.tsx

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -798,6 +934,85 @@ export default function ArtCreator() {
...
+                          Caption background color:
+                        </label>
+                        <div className="input-group">
+                          <input
+                            id="caption-bg-color"
+                            type="color"
+                            className="form-control form-control-color"
+                            value={captionBgColor}
+                            onChange={(e) => setCaptionBgColor(e.target.value)}
+                          />
```
MonkeyDo, line 1131: This should probably allow for setting alpha as well.
Not sure we can do that with the native color input?

MonkeyDo (changes requested):
Hi there,

This needs more work, there are a few features that are just not working.
Please test everything thoroughly.

Downloading the image, it looks like the colors are not applied correctly:

Preview:

Image

Downloaded image:

Image

Gopal7387:
Hi @MonkeyDo 🙂

Following up on this one with a pass this time.

**What’s fixed from your May review:**

1. Added the `#` prefix on rank. Placed the headphones icon next to the listen count

2. Changed `fontOptions` to be a `Record<string,string>` dictionary

3. Merged the font-block code into a single `buildFontStyleBlock` helper function

4. Removed the ToggleOption.tsx` file

5. Added an opacity slider for the caption background

6. Fixed the issue where caption colors were leaking into the Designer and LPs templates

**Fonts:** The font files that were referenced aren’t available yet. Aerozol confirmed there’s no approved list at the moment. So fonts are out of scope for this PR. Will be moved to a follow-up ticket (I’ll link it once created). For now only the default font is used.

**Two bugs I found during re-testing (not in your original review but real):**

1. The rank and count text was overlapping with the release and artist text on the row when the listen count was long. Fixed by adjusting the column math so it scales properly

2. The "background black on download" bug was still present. `Canvg` doesn’t interpret the same CSS as the browser preview so it ignored the custom color and opacity. Fixed by writing the color onto the SVG elements before export

Tested everything end to end this time. Used mocked data tried every toggle combination and checked actual downloaded and copied files, not just the preview.
Ready, for another look whenever you have time 🙂

All working sidebar
image

Download bug fix 
(Preview)
image

(Downloaded)
image

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -166,19 +173,128 @@ const hardCodedPresets: ColorPreset[] = [
...
+  const hex = colorHex.replace("#", "");
+  const alphaByte = Math.min(
+    255,
+    Math.max(0, Math.round((opacityPercent / 100) * 255))
+  );
+  const alphaHex = alphaByte.toString(16).padStart(2, "0");
+  return `#${hex}${alphaHex}`;
+}
+
+function prepareSvgForExport(
```
MonkeyDo: Let's move these functions (buildFontStyleBlock, getCombinedCaptionBgColor, prepareSvgForExport) to utils.tsx in the same folder, to reduce bloating of this already very long file.
We might as well export fontOptions and DEFAULT_FONT from utils as well.

frontend/js/src/explore/art-creator/ArtCreator.tsx
MonkeyDo: This one was lost in the fray, let's reinstate this toggle
When showCaption = false, it should also disable the other 4 caption-related toggles

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -607,9 +805,36 @@ export default function ArtCreator() {
...
             queryParams.set("skip-missing", "false");
           }
+          if (captionTextColorArg && captionTextColorArg !== "#ffffff") {
+            queryParams.set("caption-text-color", captionTextColorArg);
+          }
+          const combinedBgColor = getCombinedCaptionBgColor(
+            captionBgColorArg,
+            captionBgOpacityArg
+          );
+          if (combinedBgColor && combinedBgColor !== "#0000007a") {
```
MonkeyDo, line 710: Let's add a comment about this "0000007a" magic color. for future developers.  It is the default value in the template, right, is this why there is this check?

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -749,24 +1021,13 @@ export default function ArtCreator() {
...
                   <>
-                    <label className="form-check-label">
-                      <input
-                        className="form-check-input me-2"
-                        type="checkbox"
-                        checked={showCaption}
-                        onChange={(evt) => setShowCaption(evt.target.checked)}
-                      />{" "}
-                      Show caption
-                    </label>
```
MonkeyDo, line 760: As mentioned above let's reinstate this one (but with the Switch component like you did for skip-missing, that's good :))

listenbrainz/webserver/templates/art/svg-templates/macros.j2
```diff
@@ -37,18 +37,74 @@
...
+
+    {% set left_rows = (1 if show_rank else 0) + (1 if show_listen_count else 0) %}
+    {% set right_rows = (1 if show_release else 0) + (1 if show_artist else 0) %}
+    {% set total_rows = [left_rows, right_rows] | max %}
+    {% set bar_height = 15 if total_rows <= 1 else 30 %}
+
+    {% set text_limit = 18 if show_left else 32 %}
+
+    {% if show_left and show_right %}
+      {% set right_x = (image.x + 48) if show_listen_count else (image.x + 30) %}
```
MonkeyDo, line 52: I think here we could potentially improve it further by setting the x position dynamically depending on the listen count (+ 10px for each order of magnitude, for example), but I wouldn't consider it a requirement for this PR.
A small improvement for the future :)

listenbrainz/webserver/views/art_api.py
```diff
@@ -248,7 +254,13 @@ def cover_art_grid_stats(user_name, time_range, dimension, layout, image_size):
                            entity="release",
                            width=image_size,
                            height=image_size,
-                           show_caption=show_caption), 200, {
+                           show_rank=show_rank,
+                           show_listen_count=show_listen_count,
+                           show_release=show_release,
+                           show_artist=show_artist,
+                           show_caption=show_caption,
```
MonkeyDo, line 261: I see that show_caption is still passed here, so once the switch is back there should be no other changes required to have it back

MonkeyDo (commented):
It's looking a whole lot better @Gopal7387 !
Thanks for the thorough testing, the result of that is that I just have to say the UI and design parts are done!
It looks great and it's a really good improvement for users, nice one!

Agree regarding fonts, it's going to be a chunk of work so let's leave the continuation of that part to another PR.

A couple of small comments/refactoring bits below, but nothing major

Gopal7387:
Hi @MonkeyDo 🙂

Really appreciate the reviews.

Moved the font/caption helpers + fontOptions/DEFAULT_FONT into utils.tsx
Brought back the master "Show Caption" toggle,  real state now, disables the 4 sub-toggles without wiping their values, so nothing's lost when you flip it back on
Left a comment on that #0000007a


Skipped the dynamic x-position idea for now, as you mentioned before. 
Thanks again for the thorough reviews on this whole PR; they genuinely made it a lot better. 🙂

Let me know if something's off.

MonkeyDo:
Happy with the state of things! Nice work :)

There is a pretty complicated rebase I need to do before I can merge the PR, and my brain is just too tired for me to trust myself right now, so I'm pushing this back to next week but the feature is overall ready to merge! :tada:

MonkeyDo (approved):
Very nice, thanks for working on this feature, users will love it !

MonkeyDo:
@Gopal7387 Can I ask you to update your master branch? Just in case it helps in the process

Gopal7387:
Hi @MonkeyDo 
Just synced my fork's master. All good on my end now 🙂
Again, thanks a lot for your kind reviews along the way, which made this feature a lot better than it was before.

MonkeyDo:
Looking :fire: @Gopal7387 !
image

PR #3639, opened 2026-03-13
Fixed: ERROR "500:Failed to fetch additional metadata of playlist"

uzman2406:
# Problem
additional_metadata not saved when creating new playlist. [LB-1843]
The issue occurs because `fetch_playlist_recording_metadata()` attempts to query the MusicBrainz database using "MB_DATABASE_URI".This causes the server to attempt a connection to an empty URI and result in a 500 error when viewing playlists.

# Solution
This PR adds a guard so that metadata lookup is only performed when "MB_DATABASE_URI" is configured. When it is not configured, the server skips the metadata lookup and logs a warning instead.

*✅I did not use any AI
* [  ] I have used AI in this PR (add more details below)

#### If you did use AI:
* [  ] I used AI tools for communication
* [  ] I used AI tools for coding
* [  ] I understand all the changes made in this PR

### DIFF
--- a/listenbrainz/webserver/views/playlist_api.py
+++ b/listenbrainz/webserver/views/playlist_api.py
@@ -568,7 +568,7 @@ def get_playlist(playlist_mbid):
     if not playlist.is_visible_by(user_id):
         raise APINotFound("Cannot find playlist: %s" % playlist_mbid)
 
-    if fetch_metadata and current_app.config.get("MB_DATABASE_URI"):
+    if fetch_metadata:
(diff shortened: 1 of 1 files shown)


uzman2406:
Hey @MonkeyDo , my PR is ready for review . Could you please have a look on  it when you have a moment?

anshg1214:
@uzman2406 This is not the appropriate fix for this issue. I suggest you should read the create playlist endpoint to debug this issue.

PR #3658, opened 2026-03-28
Added Beta Toggle button in user settings

uzman2406:
# Problem
Currently, there is no easy way for users to switch between the beta and production versions of ListenBrainz. This makes it inconvenient for users who want to test new features or move between environments.

[LB-1611](https://tickets.metabrainz.org/browse/LB-1611)

# Solution

I introduced a "Beta Toggle button" in the user settings that allows users to seamlessly switch between beta and production environments.
Screenshot:
Screenshot 2026-03-26 204007

* [✅] I have run the code and manually tested the changes

# AI usage

* [✅ ] I did not use any AI


anshg1214:
@uzman2406 Have you checked how the beta toggle button on MusicBrainz behaves?

uzman2406:
Thanks for the review @anshg1214  . I saw the UI of musicbrainz not the codebase . However I just took a look in the codebase of musicbrainz , I see that it uses backend endpoint  to persists the user preference and then redirects accordingly and I’ve currently implemented this as a frontend-only toggle that redirects based on the current URL , since I can't find an equivalent backend endpoint.
Would you prefer that I introduce a similar backend  approach for ListenBrainz as well, or keep this as a simpler client-side switch for now?

anshg1214:
Yes, it makes sense to persist this information in the backend.

frontend/js/src/settings/Settings.tsx
```diff
@@ -18,6 +19,13 @@ export default function Settings() {
 
   const [showToken, setShowToken] = React.useState(false);
   const [copied, setCopied] = React.useState(false);
+  const [isBetaState, setIsBetaState] = React.useState(
```
MonkeyDo: setIsBetaState does not seem to be used. That suggests it does not need to be stored in state and should be a regular const instead.

frontend/js/src/settings/Settings.tsx
```diff
@@ -87,6 +95,21 @@ export default function Settings() {
         </div>
 
         <FlairsSettings />
+        <h3>Beta Preferences</h3>
```
MonkeyDo: I would put this section at the bottom of this page.

frontend/js/src/settings/Settings.tsx
uzman2406: Yup, That makes sense, thanks! I will remove the unnecessary state for isBetaState and convert it to a regular const.

frontend/js/src/settings/Settings.tsx
uzman2406: Ofcourse I will put this section at the botton of the page and then inform you ..
Thanks for the review:)

uzman2406:
Here are the suggested changes :

image

frontend/js/src/settings/Settings.tsx
```diff
@@ -130,6 +136,21 @@ export default function Settings() {
...
           </span>
         </p>
+        <h3>Beta Preference</h3>
+
+        <Switch
+          id="enable-beta-site"
+          value="beta-site"
+          checked={isBetaState}
+          onChange={handleBetaToggle}
+          switchLabel={
```
anshg1214, line 146: I don't think we need all the <span> tags here. We can simply use 

```suggestion
          switchLabel={isBetaState ? "Stop using Beta Site" : "Use Beta Site"}
```

listenbrainz/webserver/views/settings.py
```diff
@@ -67,6 +67,17 @@ def validate_funkwhale_url(url: str) -> str:
 
 settings_bp = Blueprint("settings", __name__)
 
+@settings_bp.get("/set-beta-preference/")
+@login_required
+def set_beta_preference():
+    use_beta = request.cookies.get("use_beta", "false")
+    nv = "false" if use_beta == "true" else "true"
+    return_to = request.args.get("returnto", "/")
+    response = redirect(return_to)
```
anshg1214, line 76: If someone passes ?returnto=https://evil.com, you’ll redirect users there? You need to add a domain level validation here.

listenbrainz/webserver/views/settings.py
```diff
@@ -67,6 +67,17 @@ def validate_funkwhale_url(url: str) -> str:
...
+    response.set_cookie("use_beta", nv)
```
anshg1214, line 77: `set_cookie("use_beta", nv)` does not automatically match SESSION_COOKIE_SECURE, SESSION_COOKIE_SAMESITE, SESSION_COOKIE_HTTPONLY.

So. you need to pass these explicitly. 

something like:
```
response.set_cookie(
    "use_beta",
    nv,
    max_age=60 * 60 * 24 * 365,
    secure=current_app.config.get("SESSION_COOKIE_SECURE", True),
    httponly=current_app.config.get("SESSION_COOKIE_HTTPONLY", True),
    samesite=current_app.config.get("SESSION_COOKIE_SAMESITE", "Lax"),
)
```

listenbrainz/webserver/views/settings.py
```diff
@@ -67,6 +67,17 @@ def validate_funkwhale_url(url: str) -> str:
 
 settings_bp = Blueprint("settings", __name__)
 
+@settings_bp.get("/set-beta-preference/")
```
anshg1214, line 70: Do we have a reason for keeping this endpoint as a GET endpoint?

anshg1214 (changes requested):

frontend/js/src/settings/Settings.tsx
uzman2406: I used the extra <span> wrappers initially to match the styling pattern used in other switches. However , I agree this is simpler and cleaner.

listenbrainz/webserver/views/settings.py
uzman2406: I initially passed returnto directly to preserve the current page after toggling like how MusicBrainz uses returnToCurrentPage(). I didn’t consider the possibility of external URLs being passed here so thanks for pointing that out. I will update it soon.

listenbrainz/webserver/views/settings.py
uzman2406: I initially used a simple set_cookie call since this is a lightweight preference and not sensitive session data. But it makes sense to align it with the app’s cookie security configuration. So I will update this as well.

listenbrainz/webserver/views/settings.py
uzman2406: I used a GET endpoint initially to mirror the MusicBrainz implementation, where the beta toggle is triggered via a link in the footer. But this endpoint modifies user state , using POST is more appropriate. I will update it to post .

uzman2406:
Hey @MonkeyDo  and @anshg1214 , I will continue working on this PR after May 28th once my end-sem exams are over.

PR #3659, opened 2026-03-29
LB-1973: Fix the scroll-bar on "Fresh releases" page

Harizz076:
# Problem

The Fresh Releases timeline scrollbar (originally using `rc-slider`) suffered from several interaction and layout bugs, looked kinda ugly too.

Ticket: [LB-1973](https://tickets.metabrainz.org/browse/LB-1973)

## Solution

I ended up rewriting the `ReleaseTimeline` because the old version just didn’t feel right.

The idea came from aerozol’s mockup, the scrollbar style timeline felt way more natural than the slider we had before so i tried to follow that (available on the same ticket) .

First thing i did was remove `rc-slider` and just build it manually. that made it way easier to control how it behaves and got rid of the weird dragging issues.

Scroll tracking is now properly tied to the page position, so it doesn’t go out of sync anymore. Before it would sometimes be off, especially with sticky headers. I also removed the horizontal mobile version and kept everything vertical. it’s fixed on the right now so it stays consistent everywhere. Added some padding on mobile so the timeline doesn’t sit on top of the cards. It just feels a lot nicer to use now and lines up with the mockup much better

bubble
youpage
allpage

# AI usage

* [ ] I did not use any AI

#### If you did use AI:
* [x] I used AI tools for coding (Planning the rewrite, understanding the current code for the slider, suffering cuz its ass and then having to do most of it myself anyway)

# Action

Review changes and suggest more additions. This page was recommended to use for the upcoming events page (one of the GSoC 2026 projects), so I would request that we make this scrollbar properly.


Harizz076:
@MonkeyDo: Im not good at fixing build fails but I think it should be fine now, can you run it on gh again and lmk on the matrix channel if it works now? TYSM.

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
```diff
@@ -185,7 +185,6 @@ export default function FreshReleases() {
         } else {
           const userFreshReleases = await APIService.fetchUserFreshReleases(
             currentUser.name,
-            filterRangeOptions[range].value,
             showPastReleases,
```
MonkeyDo, line 189: One of the errors preventing the image build was this one.
What was the reason for removing `filterRangeOptions[range].value`?

MonkeyDo (commented):
I fixed the linting errors and put the days value back in the `fetchUserFreshReleases` API call, should build fine now so I can deploy to test

frontend/js/src/explore/fresh-releases/FreshReleases.tsx
Harizz076: Oh, mb I thought it was related to rc-slider so took it out. Thanks for fixing it :)

MonkeyDo (commented):
Looking good! As a matter of fact, looking a lot better than the existing component!

I have some final comments, in no particular order, some are fixable without too much trouble, some might be more complex;

1. the ticks (numbers)on the side of the bar should be clickable to directly go to a specific date, especially now that the scrollbar itself is so thin as it makes it harder to click it if you want to go to the middle of the page. This is a regression.

2. We lost the icon that showed today's date (calendar icon) which was quite useful in some views (e.g. 3 months range past + future).  This is a regression.

3. There should be fewer ticks (dates) in some views, as it gets quite busy in the 3 months range past + future for example, and especially with my next suggestion:

Image

4. The font size should be increase a bit, it's very small at the moment and would be an accessibility issue.

5. Sometimes there is a mismatch sometimes between the position of the thumb and the placement on the page (see screenshot below). I think it might be less of an issue after  point 3 is tackled as it might just be lack of space for the dates.

Image

frontend/css/sass/fresh-releases.scss
MonkeyDo, line 1: Also you can run `npm run format` command to fix the formatting issues that are causing the tests to fail.

MonkeyDo:
So I think this both looks a whole lot nicer now and also works well enough. There were a couple of regressions I took the liberty of fixing, in particular clicking on a date to jump to it, and the horizontal view.
Still have some improvements to make when there are a lot of dates on the horizontal view, but in the meantime this should improve the situation.

Harizz076:
Thanks Mr.Monkey for the work :). Do you think this PR is merge-worthy for now?

MonkeyDo:
I worked on it some more, while we're here we might as well have somethingh that works better than the previous version, as we're likely not gonna touch it for another few years after that...
I improved the layout detection depending on screen size but also orientation and width/height ratio to show the most appropriate layout, and also limited the number of ticks shown on the timeline when there is limited space.
image

Still need to do a bit more testing though.

MonkeyDo:
A few more hours down the line, I've fixed all the issues that were annoying me in that page.
Hope you don't mind the takeover, it was easier to investigate myself than to explain the weird UI interactions I was seeing.
In particular there was a mismatch between the expected scroll position and the actual position on the page, and another one caused by layout shifts from the album card changing size after lazy-loading images.

MonkeyDo (approved):
Happy with this now, thanks for all the work, this looks so  much better than it did before!

PR #3661, opened 2026-03-30
LB-1964: Fix youtube player cross tab sync issue

meekhumor:
# Problem

Ticket: LB-1964

When a user starts playback in a new tab, `stopOtherBrainzPlayers()` triggers a localstorage event that calls `pauseCurrentPlayback()` with `dataSource.stop()` on the previous tab.

Issue arises due to the `stop()` method in [YoutubePlayer.tsx](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/common/brainzplayer/YoutubePlayer.tsx)

Also I found this same issue with spotify player as well, soundcloud appears to sync correctly. I haven't tested other music services. 

# Solution

image

After going through [this](https://developers.google.com/youtube/iframe_api_reference), fix was simply replacing `stopVideo()` with `pauseVideo()` in the `stop()` method. 
Also removed the `cueVideoById("")` call since we no longer need to hard reset the player instead the video should stay loaded for potential resume.

# AI usage

* [x] I did not use any AI

# Action

No further actions needed beyond testing.

### DIFF
--- a/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
@@ -80,6 +80,7 @@ export type DataSourceType = {
   playListen: (listen: Listen | JSPFTrack, streamingUrl?: string) => void;
   togglePlay: () => void;
   stop: () => void;
+  pause: () => void;
   seekToPositionMs: (msTimecode: number) => void;
   canSearchAndPlayTracks: () => boolean;
   datasourceRecordsListens: () => boolean;
@@ -461,7 +462,7 @@ export default function BrainzPlayer() {
   const pauseCurrentPlayback = async (): Promise<void> => {
     try {
       const dataSource = dataSourceRefs[getCurrentDataSourceIndex()]?.current;
-      dataSource?.stop();
+      dataSource?.pause();
     } catch (error) {
       handleError(error, "Could not pause playback");
     }
--- a/frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
+++ b/frontend/js/src/common/brainzplayer/AppleMusicPlayer.tsx
@@ -159,16 +159,12 @@ export default class AppleMusicPlayer
     this.disconnectAppleMusicPlayer();
   }
 
+  pause = () => {
+    this.appleMusicPlayer?.pause();
+  };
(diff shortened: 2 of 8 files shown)


MonkeyDo:
I don't think this is the right solution. There are other uses of this `stop` function that legetimately require the stopping and unloading of resources from music datasources.

So perhaps datasources need a new `pause` method that pauses without unloading. In some music sources they will be the same (see [Funkwhale](https://github.com/metabrainz/listenbrainz-server/blob/master/frontend/js/src/common/brainzplayer/FunkwhalePlayer.tsx#L125-L127)), in others like YT it will be a different mechanism (pause vs. stop).

Then the `pauseCurrentPlayback` function can use `dataSource?.pause();` instead of `.stop()` .

frontend/js/src/common/brainzplayer/BrainzPlayer.tsx
```diff
@@ -460,8 +461,13 @@ export default function BrainzPlayer() {
...
     try {
-      const dataSource = dataSourceRefs[getCurrentDataSourceIndex()]?.current;
-      dataSource?.stop();
+      const dataSource = dataSourceRefs[getCurrentDataSourceIndex()]
+        ?.current as (DataSourceTypes & { pause?: () => void }) | null;
+      if (dataSource?.pause) {
+        dataSource.pause();
+      } else {
+        dataSource?.stop();
+      }
```
MonkeyDo: Instead of implementing this as an optional method on the interface, let's make it mandatory and implement it for each datasource. That will be cleaner and we know what to expect.

Each datasource should have an easy way to call `pause` for it, such as `this.spotifyPlayer?.pause();`.

For some sources like Soundcloud or Navidrome, they are the same as the `stop` call, in which case you can simply refactor it to rename the function to `pause`, and make a new `stop` function that just calls `this.pause()`.
https://github.com/metabrainz/listenbrainz-server/blob/9dc4434a594cb743e5c304c750c82316d36e6472/frontend/js/src/common/brainzplayer/SoundcloudPlayer.tsx#L180-L184

meekhumor:
@MonkeyDo Done with changes, please have a look

MonkeyDo (approved):
Looking good, thanks for the changes!

PR #3662, opened 2026-03-30
LB-1969: Add message when no Similar Artists on artist page

meekhumor:
# Problem

Ticket: [LB-1969](https://tickets.metabrainz.org/browse/LB-1969)

Currently, the `Similar Artists` graph on artist pages only shows when there is enough data to generate it. Right now when there is no data, the entire section is hidden. This can be confusing for users browsing lesser known artists, they may not realize the feature exists. 

# Solution

- Always show the `Similar Artists` header on artist pages regardless of data availability
- Added a tooltip icon (?) next to the title explaining: *"This graph will be displayed when more users have listened to this artist."*
- When data is available, the graph renders as before with no tooltip icon

Screenshot from 2026-03-30 22-32-06

# AI usage

* [x] I did not use any AI

# Action

No further action needed.

### DIFF
--- a/frontend/js/src/artist/ArtistPage.tsx
+++ b/frontend/js/src/artist/ArtistPage.tsx
@@ -304,6 +304,8 @@ export default function ArtistPage(): JSX.Element {
       0
     ) > 4;
 
+  const hasSimilarArtists = similarArtists && similarArtists.artists.length > 0;
+
   return (
     <div id="entity-page" className="artist-page" role="main">
       <Helmet>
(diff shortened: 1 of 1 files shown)


Aerozol:
Thank you for picking this up!

Looks like a big gap under 'Similar Artists'? Can we make this smaller, when nothing is displayed? (not tested, based off screenshot only)

meekhumor:
@Aerozol 
image

image

Does this look fine?

Aerozol:
If it looks like this then that's still a lot of empty:

image

meekhumor:
1. 
image
image


2. 
image
image

3.
image
image

4.
image
image

@Aerozol among these which one will u prefer

Aerozol:
Thank you @meekhumor! Option 1 please.
(Since the tooltip is only displayed on hover/click we don't have to worry about it covering elements)

meekhumor:
Alright @Aerozol , I have updated the code accordingly, Thanks!

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -559,22 +561,40 @@ export default function ArtistPage(): JSX.Element {
         </div>
       </div>
 
-      {similarArtists && similarArtists.artists.length > 0 ? (
+      <h3 className="header-with-line">
+        Similar Artists
+        {!(similarArtists && similarArtists.artists.length > 0) && (
```
anshg1214, line 566:
```
(similarArtists && similarArtists.artists.length > 0)
```

We have the same check in 3 different places. I think it makes sense to save the value in a variable and then use it everywhere.

anshg1214 (commented):
I believe if we don't have the similar artist's data for a particular artist, instead of hiding that information in a tooltip, we should explicitly make it clear to the users that this data is not there. We can just have a line below Similar Artists saying that this data is unavailable. 

From an end user's perspective, on first sight, it looks like something is missing on the page, and then they'll have to drag to the tooltip and then find out that the data is not available, and it's not a bug.

frontend/js/src/artist/ArtistPage.tsx
meekhumor: image


is it fine?

MonkeyDo: I think that does match ansh's feedback.

I would go one step further and ask if you think this is definitely better user experience than simply hiding the section when no data is available.

Aerozol: @MonkeyDo I made this ticket because I got confused when I visited an artist that isn't that unpopular (well, by my standards) and couldn't find the graph... I knew what the situation would be, but for other users I think it is likely they will think it's broken or not even know that we have a 'Similar Artists' feature.

anshg1214: @meekhumor Yes, that covers my feedback. Please go ahead and make the changes.

meekhumor: @anshg1214 Done with the changes!

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -559,22 +561,26 @@ export default function ArtistPage(): JSX.Element {
...
-              similarArtistsLimit={18}
-              graphParentElementRef={graphParentElementRef}
-            />
-          </div>
-        </>
-      ) : null}
+      <h3 className="header-with-line">Similar Artists</h3>
+      {!hasSimilarArtists && (
+        <p>
+          This graph will be displayed when more users have listened to this
```
anshg1214:
```suggestion
          Similar artists will appear here once enough people have listened to this artist on ListenBrainz.
```

MonkeyDo (approved):
Thank you!
Good to go 🚀

PR #3663, opened 2026-03-31
Options menu layout breaks/wraps incorrectly in grid view

uzman2406:
# Problem

The options dropdown menu in grid view was being clipped and rendered inconsistently compared to list view.
[LB-1976](https://tickets.metabrainz.org/browse/LB-1976)
Caused by "overflow:hidden" in .playlist{} card container.

# Solution

I removed "overflow:hidden" from .playlist{} card container.
I also applied targeted overflow handling to inner elements and adjust dropdown alignment from overflowing.
It ensures no clipping/wrappings issues and clean , stable UI layout

* [✅] I have run the code and manually tested the changes
 Screenshots:
 Before:
 
Screenshot 2026-03-31 213457

After:

Screenshot 2026-03-31 221230

* [ ✅] I did not use any AI


uzman2406:
Hey @anshg1214 Can you please take a look in this PR as well if you got a moment :)

uzman2406:
Hey @MonkeyDo  , could you please take a look at this PR when you get a moment? It's been two weeks.

frontend/css/sass/playlists.scss
```diff
@@ -27,17 +28,25 @@ $long-description-lines: 8;
     width: 15em;
     height: 15em !important;
     position: relative;
-    overflow: hidden;
+    border-radius: 12px;
```
MonkeyDo, line 31: Border radius does not match what's on prod, let's use this SASS variable to have the same value:

```suggestion
    border-radius: $card-border-radius;
```

frontend/css/sass/playlists.scss
```diff
@@ -27,17 +28,25 @@ $long-description-lines: 8;
...
+    background: white;
+    box-shadow: 0 4px 12px rgba(0, 0, 0, 0.1);
```
MonkeyDo, line 33: Where did these box shadow values come from?
It changes the style of the component, but to what end?

frontend/css/sass/playlists.scss
```diff
@@ -27,17 +28,25 @@ $long-description-lines: 8;
     width: 15em;
     height: 15em !important;
     position: relative;
-    overflow: hidden;
+    border-radius: 12px;
+    background: white;
```
MonkeyDo, line 32: Same question here, why overwrite the background color?

frontend/css/sass/playlists.scss
```diff
@@ -27,17 +28,25 @@ $long-description-lines: 8;
...
+        right: 0;
+        left: auto;
+        width: 100%;
+      }
     }
 
     .playlist-card-action-button {
       position: absolute;
       bottom: 0;
+      border-radius: 0 0 12px 12px;
```
MonkeyDo, line 49: Let's use `$card-border-radius` here to to conserve the same look.

frontend/css/sass/playlists.scss
```diff
@@ -56,6 +65,8 @@ $long-description-lines: 8;
       height: 100%;
       width: 100%;
       background-color: $playlist-background-color;
+      border-radius: 12px;
```
MonkeyDo, line 68: Same remark here

frontend/css/sass/playlists.scss
```diff
@@ -27,17 +28,25 @@ $long-description-lines: 8;
...
+    box-shadow: 0 4px 12px rgba(0, 0, 0, 0.1);
 
     .playlist-card-action-dropdown {
       position: absolute;
       bottom: 0;
       width: 100%;
+      .dropdown-menu {
+        right: 0;
+        left: auto;
+        width: 100%;
```
MonkeyDo, line 42: Why are those needed?
The menu seems to look and behave the same without them.

MonkeyDo (approved):
Thanks for opening a PR for this, it's an annoying bug!

Generally you should review your work more thoroughly, there are a few changes in the PR that don't seem necessary and/or change the existing look of the page without justification.

Not sure if they were on purpose or generated and commited without review, but it needs more review time on your end, please.

MonkeyDo:
I also wonder if the solution might be to more the menu component, instead of modifying the CSS.

Looking at [the current code](https://github.com/metabrainz/listenbrainz-server/blob/1e1889fefb445d08421bcddf920a0c19a7ebe2d5/frontend/js/src/user/playlists/components/PlaylistCard.tsx#L212-L231), we instantiate a menu for each playlist card component, wasting resources.

The PlaylistMenu component could probably be refactored to use react-bootstrap which offers a [`renderMenuOnMount`](https://react-bootstrap.github.io/docs/components/dropdowns#dropdownmenu) option we could set to false.
I think this might solve the issue of rendering the menu component inside the card, or should [likely be configurable](https://popper.js.org/react-popper/v2/react-portals/) with popperjs (library used under the hood by react-bootstrap).

Care to try out a demo version of that to see if it fixes the overflow issue?

frontend/css/sass/playlists.scss
uzman2406: Thanks for pointing that out ,  I should avoid hardcoding values 🙂

frontend/css/sass/playlists.scss
uzman2406: The box-shadow and white background was added while trying to visually compensate for removing overflow: hidden .It’s not directly related to fixing the dropdown issue.

frontend/css/sass/playlists.scss
uzman2406: I’ll replace the hardcoded border-radius with $card-border-radius here as well and avoid try to avoid hardcoding :)

frontend/css/sass/playlists.scss
uzman2406: These were added to prevent the dropdown from overflowing outside the viewport when the card is near the left edge. I’ll re-test without them and remove them if they don’t provide any real benefit.

uzman2406:
Yeah I hadn’t thought of  refactoring the menu rendering itself. My current fix focuses on resolving the overflow issue at the CSS level, but I agree that rendering the dropdown outside the card or controlling its mount behavior could be a cleaner solution.
I’ll explore a small demo using react-bootstrap’s dropdown:)

MonkeyDo:
> Yeah I hadn’t thought of refactoring the menu rendering itself. My current fix focuses on resolving the overflow issue at the CSS level, but I agree that rendering the dropdown outside the card or controlling its mount behavior could be a cleaner solution. I’ll explore a small demo using react-bootstrap’s dropdown:)

Great.

I would recommend generally spending more time reproducing and analyzing the issue you are trying to solve, rather than charging ahead with a fix, any fix.

uzman2406:
Yeah, you’re right 😅
I tried out the demo you suggested for React-Bootstrap's Dropdown with renderOnMount={false}. To fix the overflow issue, I had to move overflow:hidden from the .playlist-card down to the .info element in the SCSS. I tried popperConfig as well, but it didn't resolve the issue, so removing overflow:hidden from the parent seems like the only way for now. I’m updating the PR; if you have any suggestions or alternatives for handling the overflow, please let me know! :)

uzman2406:
Hey @MonkeyDo, Can you please check the latest changes and let me know if I missed anything!

MonkeyDo (changes requested):
The PR currently breaks the use of the PlaylistMenu when used on the playlist page (not on the playlists overview page but the one that displays one playlist).
Please test thoroughly, there might be other pages too that use this functionality, they should all work.

PR #3664, opened 2026-04-01
LB-1965 : smooth progress bar animation with drag-to-seek

Gopal7387:
## Problem

The BrainzPlayer progress bar had some problems. LB-1965

It had two issues.

The first issue with the BrainzPlayer progress bar was that the playback animation did not move smoothly. Also BrainzPlayer progress bar animation was controlled by React state updates that happened every 300 milliseconds, which made the BrainzPlayer progress bar jump in steps of moving smoothly.
The second issue with the BrainzPlayer progress bar was that you could not drag the BrainzPlayer progress bar to change the position of the playback.
You could only click on a position, on the BrainzPlayer progress bar to change the playback position.

## Solution

Replaced the React state-driven `width%` approach with a
`requestAnimationFrame` loop that writes `transform: scaleX()` directly
to the DOM via a ref. This gives true 60fps animation with zero React
re-renders per frame.

Added a circular scrubber handle that appears on hover and follows the
playhead. Users can now click and drag anywhere on the bar to seek.

## What changed

**ProgressBar.tsx**
- rAF loop reads `performance.now() - updateTime` to interpolate live
  playhead position between atom updates
- `progressBarInnerRef` receives direct `scaleX` writes — no setState
  in the animation path
- Added `isDraggingRef`, `rectCacheRef`, `pendingSeekMsRef` to coordinate
  drag state without touching React
- Document-level `pointermove` / `pointerup` listeners so drag works
  even if mouse leaves the bar
- `pendingSeekMsRef` holds the optimistic position after seek fires,
  preventing the bar from snapping back while the player catches up
- Keyboard ArrowLeft/Right seek also goes through `pendingSeekMsRef`
  for the same snap-free behaviour
- `cursor: grabbing` on `document.body` during active drag

**brainzplayer.scss**
- Added `width: 100%` to `.progress-bar` (required for scaleX base)
- Added `overflow: visible` to `.progress` to unclip the handle
- Added `.dragging` rule to pin bar height during drag
- Added `.progress-handle` styles (14px circle, `$primary-color`,
  hover-only via `scaleX` transition)
- Removed legacy `border-right` nubbin — the handle replaces it

## Testing

Tested locally with a YouTube-backed track:
- Bar animates smoothly at 60fps during playback
- Handle appears on hover, disappears on mouse leave
- Drag works across the full bar width, including when mouse
  leaves the element
- No snap-back after releasing drag
- Keyboard seek (ArrowLeft/Right) works correctly
- `.music-player` fullscreen view unaffected

## Notes

Handle color and hover-only behaviour confirmed with @Aerozol . No new colours introduced.

AI usage: 
If you did use AI:
[☑️] I did not use AI

https://github.com/user-attachments/assets/66c95acd-fc2c-4def-9b2c-6a9f2d4c672f

### DIFF
--- a/frontend/css/sass/brainzplayer.scss
+++ b/frontend/css/sass/brainzplayer.scss
@@ -408,6 +408,7 @@ $youtube-expanded-width: 1280px;
   transition: height, 0.2s;
 
   .progress-bar {
+    width: 100%;
     height: 100%;
     background-color: $primary-color;
   }
--- a/frontend/css/sass/brainzplayer.scss
+++ b/frontend/css/sass/brainzplayer.scss
@@ -405,7 +405,9 @@ $youtube-expanded-width: 1280px;
   width: 100%;
   cursor: pointer;
   z-index: 5;
+  overflow: visible;
   transition: height, 0.2s;
+
 
   .progress-bar {
     width: 100%;
--- a/frontend/css/sass/brainzplayer.scss
+++ b/frontend/css/sass/brainzplayer.scss
@@ -429,7 +429,7 @@ $youtube-expanded-width: 1280px;
     width: 10px;
     height: 10px;
     border-radius: 50%;
-    background-color: white;
+    background-color: $primary-color;
     transform: translate(-50%, -50%) scaleX(0);
     transition: transform 0.15s ease;
     will-change: transform;
--- a/frontend/css/sass/brainzplayer.scss
+++ b/frontend/css/sass/brainzplayer.scss
@@ -408,7 +408,6 @@ $youtube-expanded-width: 1280px;
   overflow: visible;
   transition: height, 0.2s;
 
-
   .progress-bar {
     width: 100%;
     height: 100%;
--- a/frontend/css/sass/brainzplayer.scss
+++ b/frontend/css/sass/brainzplayer.scss
@@ -426,8 +426,8 @@ $youtube-expanded-width: 1280px;
   .progress-handle {
     position: absolute;
     top: 50%;
-    width: 10px;
-    height: 10px;
+    width: 14px;
+    height: 14px;
     border-radius: 50%;
     background-color: $primary-color;
     transform: translate(-50%, -50%) scaleX(0);
--- a/frontend/css/sass/brainzplayer.scss
+++ b/frontend/css/sass/brainzplayer.scss
@@ -410,7 +410,6 @@ $youtube-expanded-width: 1280px;
   .progress-bar {
     height: 100%;
     background-color: $primary-color;
-    border-right: 2px solid $dark-turquoise;
   }
   &:hover {
     height: $progress-bar-hover-height;
--- a/frontend/css/sass/brainzplayer.scss
+++ b/frontend/css/sass/brainzplayer.scss
@@ -417,6 +417,26 @@ $youtube-expanded-width: 1280px;
     top: -$progress-bar-hover-height;
   }
 
+  &.dragging {
+    height: $progress-bar-hover-height;
+    top: -$progress-bar-hover-height;
+    cursor: grabbing;
+  }
+
+  .progress-handle {
+    position: absolute;
+    top: 50%;
+    width: 10px;
+    height: 10px;
+    border-radius: 50%;
+    background-color: white;
+    transform: translate(-50%, -50%) scaleX(0);
+    transition: transform 0.15s ease;
+    will-change: transform;
+    pointer-events: none;
+    left: var(--handle-x, 0px);
+  }
+
   .progress-tooltip {
     background: $primary-color;
     margin-top: -4px;
--- a/frontend/js/src/common/brainzplayer/ProgressBar.tsx
+++ b/frontend/js/src/common/brainzplayer/ProgressBar.tsx
@@ -27,6 +27,7 @@ const MOUSE_THROTTLE_DELAY: number = 300;
 
 const TOOLTIP_INITIAL_CONTENT: string = "0:00";
 const TOOLTIP_TOP_OFFSET: number = 39;
+const HANDLE_RADIUS = 5; // px — handle is 10×10px
 
 // Originally by ford04 - https://stackoverflow.com/a/62017005
 const useThrottle = (callback: any, delay: number | undefined) => {
@@ -51,6 +52,7 @@ function ProgressBar(props: ProgressBarProps) {
   const [tipContent, setTipContent] = React.useState(TOOLTIP_INITIAL_CONTENT);
   const progressBarRef = React.useRef<HTMLDivElement>(null);
   const progressBarInnerRef = React.useRef<HTMLDivElement>(null);
+  const handleRef = React.useRef<HTMLDivElement>(null);
   const rafRef = React.useRef<number>(0);
 
   React.useEffect(() => {
@@ -60,6 +62,13 @@ function ProgressBar(props: ProgressBarProps) {
       const ratio = Math.min(liveProgressMs / durationMs, 1) || 0;
       if (progressBarInnerRef.current) {
         progressBarInnerRef.current.style.transform = `scaleX(${ratio})`;
+      }
+      if (handleRef.current) {
+        const barWidth =
+          progressBarInnerRef.current?.parentElement?.getBoundingClientRect()
+            .width ?? 0;
+        const handleX = ratio * barWidth;
+        handleRef.current.style.setProperty("--handle-x", `${handleX}px`);
       }
       rafRef.current = requestAnimationFrame(tick);
     };
@@ -133,6 +142,18 @@ function ProgressBar(props: ProgressBarProps) {
         onClick={mouseEventHandler}
         onMouseMove={mouseEventHandler}
         onKeyDown={onKeyPressHandler}
+        onMouseEnter={() => {
+          if (handleRef.current) {
+            handleRef.current.style.transform =
+              "translate(-50%, -50%) scaleX(1)";
+          }
+        }}
+        onMouseLeave={() => {
+          if (handleRef.current) {
+            handleRef.current.style.transform =
+              "translate(-50%, -50%) scaleX(0)";
+          }
+        }}
         aria-label="Audio progress control"
         role="progressbar"
         aria-valuemin={0}
@@ -146,6 +167,24 @@ function ProgressBar(props: ProgressBarProps) {
           className="progress-bar bg-info"
           ref={progressBarInnerRef}
           style={{ transform: "scaleX(0)", transformOrigin: "left" }}
+        />
+        <div
+          ref={handleRef}
+          className="progress-handle"
+          style={{
+            position: "absolute",
+            top: "50%",
+            left: "var(--handle-x, 0px)",
+            width: `${HANDLE_RADIUS * 2}px`,
+            height: `${HANDLE_RADIUS * 2}px`,
+            borderRadius: "50%",
+            backgroundColor: "white",
+            transform: "translate(-50%, -50%) scaleX(0)",
+            transformOrigin: "center",
+            transition: "transform 0.15s ease",
+            willChange: "transform",
+            pointerEvents: "none",
+          }}
         />
         <ReactTooltip
           className="progress-tooltip"
--- a/frontend/js/src/common/brainzplayer/ProgressBar.tsx
+++ b/frontend/js/src/common/brainzplayer/ProgressBar.tsx
@@ -3,7 +3,12 @@ import * as React from "react";
 import ReactTooltip from "react-tooltip";
 import { useAtomValue } from "jotai";
 import { millisecondsToStr } from "../../playlists/utils";
-import { durationMsAtom, progressMsAtom } from "./BrainzPlayerAtoms";
+import {
+  durationMsAtom,
+  progressMsAtom,
+  playerPausedAtom,
+  updateTimeAtom,
+} from "./BrainzPlayerAtoms";
 
 type ProgressBarProps = {
   seekToPositionMs: (msTimeCode: number) => void;
@@ -39,13 +44,28 @@ const useThrottle = (callback: any, delay: number | undefined) => {
 function ProgressBar(props: ProgressBarProps) {
   const progressMs = useAtomValue(progressMsAtom);
   const durationMs = useAtomValue(durationMsAtom);
+  const playerPaused = useAtomValue(playerPausedAtom);
+  const updateTime = useAtomValue(updateTimeAtom);
 
   const { seekToPositionMs, showNumbers } = props;
   const [tipContent, setTipContent] = React.useState(TOOLTIP_INITIAL_CONTENT);
   const progressBarRef = React.useRef<HTMLDivElement>(null);
-  const progressPercentage = Number(
-    ((progressMs * 100) / durationMs).toFixed()
-  );
+  const progressBarInnerRef = React.useRef<HTMLDivElement>(null);
+  const rafRef = React.useRef<number>(0);
+
+  React.useEffect(() => {
+    const tick = () => {
+      const elapsed = performance.now() - updateTime;
+      const liveProgressMs = playerPaused ? progressMs : progressMs + elapsed;
+      const ratio = Math.min(liveProgressMs / durationMs, 1) || 0;
+      if (progressBarInnerRef.current) {
+        progressBarInnerRef.current.style.transform = `scaleX(${ratio})`;
+      }
+      rafRef.current = requestAnimationFrame(tick);
+    };
+    rafRef.current = requestAnimationFrame(tick);
+    return () => cancelAnimationFrame(rafRef.current);
+  }, [progressMs, durationMs, playerPaused, updateTime]);
 
   const mouseEventHandler = useThrottle(
     (event: React.MouseEvent<HTMLInputElement>): void => {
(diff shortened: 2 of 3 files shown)


Gopal7387:
Hi @anshg1214 
Please take a look at this, when you get a moment.
🙂

anshg1214:
@Gopal7387 Please fill out the AI Declaration in the PR Description

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -22,6 +27,7 @@ const MOUSE_THROTTLE_DELAY: number = 300;
 
 const TOOLTIP_INITIAL_CONTENT: string = "0:00";
 const TOOLTIP_TOP_OFFSET: number = 39;
+const HANDLE_RADIUS = 7; // px — handle is 14×14px
```
anshg1214: This variale is not being used.

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -100,36 +203,67 @@ function ProgressBar(props: ProgressBarProps) {
...
+            handleRef.current.style.transform =
+              "translate(-50%, -50%) scaleX(0)";
+          }
+        }}
         aria-label="Audio progress control"
         role="progressbar"
         aria-valuemin={0}
         aria-valuemax={100}
-        aria-valuenow={progressPercentage || 0}
+        aria-valuenow={0}
```
anshg1214: aria-valuenow is fixed at 0. Before it reflected progressPercentage. For a role="progressbar" with aria-valuemax={100}, aria-valuenow should track the same 0–100 semantics as the visual bar

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +45,108 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
-  const progressPercentage = Number(
-    ((progressMs * 100) / durationMs).toFixed()
-  );
+  const progressBarInnerRef = React.useRef<HTMLDivElement>(null);
+  const handleRef = React.useRef<HTMLDivElement>(null);
+  const rafRef = React.useRef<number>(0);
+  const isDraggingRef = React.useRef<boolean>(false);
+  const rectCacheRef = React.useRef<DOMRect | null>(null);
+  const pendingSeekMsRef = React.useRef<number>(-1);
+  const seekVisualMsRef = React.useRef<number>(-1);
```
anshg1214: This is also not being used. We can safely remove this. - seekVisualMsRef

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +45,108 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
+    const onPointerUp = (e: PointerEvent) => {
+      if (!isDraggingRef.current) return;
+      isDraggingRef.current = false;
+      document.body.style.cursor = "";
+      const msPos = getMsFromClientX(e.clientX);
+      pendingSeekMsRef.current = msPos;
+      seekToPositionMs(msPos);
+      document
+        .querySelectorAll(".progress.dragging")
+        .forEach((el) => el.classList.remove("dragging"));
```
anshg1214: Storing the dragging element on a ref is clearer

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +45,108 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
 
   const { seekToPositionMs, showNumbers } = props;
   const [tipContent, setTipContent] = React.useState(TOOLTIP_INITIAL_CONTENT);
   const progressBarRef = React.useRef<HTMLDivElement>(null);
-  const progressPercentage = Number(
-    ((progressMs * 100) / durationMs).toFixed()
-  );
+  const progressBarInnerRef = React.useRef<HTMLDivElement>(null);
+  const handleRef = React.useRef<HTMLDivElement>(null);
+  const rafRef = React.useRef<number>(0);
```
anshg1214: Do we actually need this ref?  a `let rafId` inside the same useEffect is enough for cancelAnimationFrame on cleanup

anshg1214 (changes requested):
Can you add some more comments to the code? It'll make it easier for us to review the PR

frontend/js/src/common/brainzplayer/ProgressBar.tsx
Gopal7387: Done, removed.

frontend/js/src/common/brainzplayer/ProgressBar.tsx
Gopal7387: seekVisualMsRef is actually written to in two places, on mousedown and on pointermove.
so I've kept it for now. 
Could you clarify if you'd like a different approach here, or if those two write sites should also be removed?

frontend/js/src/common/brainzplayer/ProgressBar.tsx
anshg1214: `seekVisualMsRef` is written but never read. You're assigning it at lines 110 and 230, but this is never consumed anywhere

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -100,36 +202,68 @@ function ProgressBar(props: ProgressBarProps) {
...
+          flushVisuals(msPos);
+        }}
+        onMouseEnter={() => {
+          if (handleRef.current) {
+            handleRef.current.style.transform =
+              "translate(-50%, -50%) scaleX(1)";
+          }
+        }}
+        onMouseLeave={() => {
+          if (handleRef.current) {
```
anshg1214, line 247:
```suggestion
          if (!isDraggingRef.current && handleRef.current) {
```

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -100,36 +202,68 @@ function ProgressBar(props: ProgressBarProps) {
...
+          document.body.style.cursor = "grabbing";
+          rectCacheRef.current = (e.currentTarget as HTMLDivElement).getBoundingClientRect();
+          if (handleRef.current) {
+            handleRef.current.style.transform =
+              "translate(-50%, -50%) scaleX(1)";
+          }
+          (e.currentTarget as HTMLDivElement).classList.add("dragging");
+          draggingElementRef.current = e.currentTarget as HTMLDivElement;
+          const msPos = getMsFromClientX(e.clientX);
+          seekVisualMsRef.current = msPos;
```
anshg1214:
```suggestion
```

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +44,108 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
+    if (handleRef.current) {
+      handleRef.current.style.setProperty("--handle-x", `${handleX}px`);
+    }
+  };
+
+  React.useEffect(() => {
+    const onPointerMove = (e: PointerEvent) => {
+      if (!isDraggingRef.current) return;
+      const msPos = getMsFromClientX(e.clientX);
+      seekVisualMsRef.current = msPos;
```
anshg1214:
```suggestion
```

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +44,108 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
+      draggingElementRef.current?.classList.remove("dragging");
+      draggingElementRef.current = null;
+    };
+
+    document.addEventListener("pointermove", onPointerMove);
+    document.addEventListener("pointerup", onPointerUp);
+    return () => {
+      document.removeEventListener("pointermove", onPointerMove);
+      document.removeEventListener("pointerup", onPointerUp);
+    };
```
anshg1214, line 133: `pointercancel` not handled - drag state can get permanently stuck. 


```suggestion
    document.addEventListener("pointermove", onPointerMove);
    document.addEventListener("pointerup", onPointerUp);
    document.addEventListener("pointercancel", onPointerUp);
    return () => {
      document.removeEventListener("pointermove", onPointerMove);
      document.removeEventListener("pointerup", onPointerUp);
      document.removeEventListener("pointercancel", onPointerUp);
    };
```

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +44,108 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
+
+  React.useEffect(() => {
+    const onPointerMove = (e: PointerEvent) => {
+      if (!isDraggingRef.current) return;
+      const msPos = getMsFromClientX(e.clientX);
+      seekVisualMsRef.current = msPos;
+      flushVisuals(msPos);
+    };
+
+    const onPointerUp = (e: PointerEvent) => {
```
anshg1214: Let's rename this function. 

```suggestion
    const endDrag = (e: PointerEvent) => {
```

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +44,108 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
+          progressBarInnerRef.current?.parentElement?.getBoundingClientRect()
+            .width ?? 0;
+        const handleX = ratio * barWidth;
+        handleRef.current.style.setProperty("--handle-x", `${handleX}px`);
+      }
+      rafId = requestAnimationFrame(tick);
+    };
+    rafId = requestAnimationFrame(tick);
+    return () => cancelAnimationFrame(rafId);
+  }, [progressMs, durationMs, playerPaused, updateTime]);
```
anshg1214, line 82: The animation loop was set up to stop completely when dragging started.

The problem is that the loop also restarts every ~400ms when the player reports its position.

Each restart during a drag would stop immediately and never resume. So when you released the drag, there was no loop running - so the bar would freeze for up to 400ms.


```suggestion
  React.useEffect(() => {
    let rafId: number;
    const tick = () => {
      if (!isDraggingRef.current) {
        const base =
          pendingSeekMsRef.current >= 0 ? pendingSeekMsRef.current : progressMs;
        const elapsed = playerPaused ? 0 : performance.now() - updateTime;
        const liveProgressMs = base + elapsed;
        const ratio = Math.min(liveProgressMs / durationMs, 1) || 0;
        if (progressBarInnerRef.current) {
          progressBarInnerRef.current.style.transform = `scaleX(${ratio})`;
        }
        if (handleRef.current) {
          const barWidth = rectCacheRef.current?.width ?? 0;
          const handleX = ratio * barWidth;
          handleRef.current.style.setProperty("--handle-x", `${handleX}px`);
        }
      }
      rafId = requestAnimationFrame(tick);
    };
    rafId = requestAnimationFrame(tick);
    return () => cancelAnimationFrame(rafId);
  }, [progressMs, durationMs, playerPaused, updateTime]);

```

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +44,108 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
+
+  React.useEffect(() => {
+    const onResize = () => {
+      if (isDraggingRef.current && progressBarRef.current) {
+        rectCacheRef.current = progressBarRef.current.getBoundingClientRect();
+      }
+    };
+    window.addEventListener("resize", onResize);
+    return () => window.removeEventListener("resize", onResize);
+  }, []);
```
anshg1214, line 156:
```suggestion
  React.useEffect(() => {
    if (progressBarRef.current) {
      rectCacheRef.current = progressBarRef.current.getBoundingClientRect();
    }
    const onResize = () => {
      if (progressBarRef.current) {
        rectCacheRef.current = progressBarRef.current.getBoundingClientRect();
      }
    };
    window.addEventListener("resize", onResize);
    return () => window.removeEventListener("resize", onResize);
  }, []);
```

anshg1214 (changes requested):

frontend/js/src/common/brainzplayer/ProgressBar.tsx
Gopal7387: Removed all three sites. The ref declaration, the mousedown write, and the pointermove write

frontend/js/src/common/brainzplayer/ProgressBar.tsx
Gopal7387: Applied.

frontend/js/src/common/brainzplayer/ProgressBar.tsx
Gopal7387: Applied the suggestion, rect is now cached on mount as well.

frontend/js/src/common/brainzplayer/ProgressBar.tsx
Gopal7387: Applied. pointercancel now maps to endDrag alongside pointerup

frontend/js/src/common/brainzplayer/ProgressBar.tsx
Gopal7387: Renamed to endDrag.

frontend/js/src/common/brainzplayer/ProgressBar.tsx
Gopal7387: That was a catch. I restructured the tick. Now the visuals are, inside an if statement that checks if we are not currently dragging. This if statement is !isDraggingRef.current. I also moved the requestAnimationFrame part with the tick function outside of this if statement. So now the loop will keep going even when we are dragging.. It will start again right away when we let go.

Gopal7387:
Hi @anshg1214 
I have taken care of everything we talked about yesterday. The latest commit is also updated to match the master version. Please take a look, when you have some time 🙂

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -100,36 +210,69 @@ function ProgressBar(props: ProgressBarProps) {
...
+          className="progress-bar bg-info"
+          ref={progressBarInnerRef}
+          style={{ transform: "scaleX(0)", transformOrigin: "left" }}
+        />
+        <div
+          ref={handleRef}
+          className="progress-handle"
+          style={{
+            transform: "translate(-50%, -50%) scaleX(0)",
+            left: "var(--handle-x, 0px)",
```
MonkeyDo: These are already in frontend/css/sass/brainzplayer.scss now and should be removed from here

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -100,36 +210,69 @@ function ProgressBar(props: ProgressBarProps) {
...
+        }
         tabIndex={0}
         data-tip={tipContent}
         ref={progressBarRef}
       >
-        <div className="progress-bar bg-info" style={progressBarStyle} />
+        <div
+          className="progress-bar bg-info"
+          ref={progressBarInnerRef}
+          style={{ transform: "scaleX(0)", transformOrigin: "left" }}
```
MonkeyDo: Similarly this style should make its way to the brainzplayer.scss file instead of being set here.

frontend/js/src/common/brainzplayer/ProgressBar.tsx
```diff
@@ -39,13 +44,116 @@ const useThrottle = (callback: any, delay: number | undefined) => {
...
+  React.useEffect(() => {
+    if (progressBarRef.current) {
+      rectCacheRef.current = progressBarRef.current.getBoundingClientRect();
+    }
+    const onResize = () => {
+      if (progressBarRef.current) {
+        rectCacheRef.current = progressBarRef.current.getBoundingClientRect();
+      }
+    };
+    window.addEventListener("resize", onResize);
```
MonkeyDo, line 158: Should probably debounce this handler.
Example here: 
https://github.com/metabrainz/listenbrainz-server/blob/4584b89888937756b6a0c5dff66ef8b37e09a4a6/frontend/js/src/user/year-in-music/components/YIMYearSelection.tsx#L94-L97

frontend/css/sass/brainzplayer.scss
```diff
@@ -405,18 +405,39 @@ $youtube-expanded-width: 1280px;
...
   &:hover {
     height: $progress-bar-hover-height;
     top: -$progress-bar-hover-height;
   }
 
+  &.dragging {
+    height: $progress-bar-hover-height;
+    top: -$progress-bar-hover-height;
+    cursor: grabbing;
+  }
```
MonkeyDo: This could combine as 

```suggestion
  &:hover, &.dragging {
    height: $progress-bar-hover-height;
    top: -$progress-bar-hover-height;
  }

  &.dragging {
    cursor: grabbing;
  }
```

MonkeyDo (changes requested):
Good improvement, thank you!

Some small improvements below, but it is working fine apart from that.

Gopal7387:
Hi @MonkeyDo 
I debounced the resize handler, removed the duplicate inline styles (moved one into the SCSS because it was not there yet) and merged the hover/dragging selectors. I tested everything locally. It is all working fine.
Let me know if anything else needs a look 🙂

MonkeyDo (approved):
Looks good, thanks for the final touches!

PR #3667, opened 2026-04-03
Security: Stored/Reflected XSS via unescaped username in HTML Markup

tuanaiseo:
## Problem

`generate_username_link` builds HTML with an f-string and wraps it in `Markup`, which disables escaping. If `user_name` contains attacker-controlled HTML/JS, it will be rendered directly in the browser.

**Severity**: `high`
**File**: `listenbrainz/model/utils.py`

## Solution

Avoid `Markup` with raw interpolation. Use `markupsafe.escape(user_name)` for both link text and URL segment (plus URL encoding for path components), or render links in Jinja templates where autoescaping applies.

## Changes

- `listenbrainz/model/utils.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/listenbrainz/model/utils.py
+++ b/listenbrainz/model/utils.py
@@ -1,5 +1,9 @@
-from markupsafe import Markup
+from urllib.parse import quote
+
+from markupsafe import Markup, escape
 
 
 def generate_username_link(user_name):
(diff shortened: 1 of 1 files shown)


MonkeyDo:
Nowhere in the codebase can this method be called with user-controlled data.

Please only open PRs for issues that are actually reproducible.

PR #3668, opened 2026-04-03
Security: SQL injection risk from dynamic SQL identifier concatenation

tuanaiseo:
## Problem

`insert_rows` concatenates `table` (and likely `cols`) directly into SQL strings. If these values can be influenced by untrusted input, an attacker can inject SQL through identifiers.

**Severity**: `high`
**File**: `mbid_mapping/mapping/utils.py`

## Solution

Treat table/column names as trusted constants only. Enforce allowlists for identifiers and construct SQL with `psycopg2.sql.SQL` + `Identifier` instead of string concatenation.

## Changes

- `mbid_mapping/mapping/utils.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/mbid_mapping/mapping/utils.py
+++ b/mbid_mapping/mapping/utils.py
@@ -2,6 +2,7 @@ import sys
 from time import asctime
 
 import psycopg2
+from psycopg2 import sql
 from psycopg2.extras import execute_values
 from psycopg2.errors import OperationalError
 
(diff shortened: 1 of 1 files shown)


MonkeyDo:
Hello, this looks like automated AI work, and we are not fond of merging code changes that was not reasoned by a human.

Did you investigate the code in question to ascertain whether there is an actual risk?

MonkeyDo:
With no response, closing this PR as it is not clear whether this is a legitimate risk.
Please reopen if you can replicate an actual issue

PR #3669, opened 2026-04-03
Fix[LB-1947]: Track listen count stuck

Swaim-Sahay:
# Description
This PR fixes an issue where the global track and album listen counts were actively getting stuck and not updating and also corrects a bug computing release group popularity.

Ticket: LB-1947

# Problem
Our background system  caches a large chunk of listen data so it doesn't have to recalculate it from scratch every day. The problem was that when fresh listen data was imported, the system didn't realize it had been updated so it kept using the old frozen cache forever. Because of this the final listen numbers shown to users never went up. There was also a small bug where the wrong math was being used for release group listen counts.

# Solution
Taught the caching system to remember exactly where it got its data from. Now, if it notices that the main data location has changed it knows to throw away the old cache and build a new one. We also pointed the release group counts to the correct calculation.

# Changes Made
1) `listenbrainz_spark/schema.py`: Added a `dump_location` field to keep track of where the data came from.
2) `listenbrainz_spark/stats/incremental/incremental_stats_engine.py`: Updated the system to check and save the  `dump_location` so it can detect when new data arrives.
3) `listenbrainz_spark/popularity/listens.py`: Corrected the query used for calculating release group popularity.
4) `listenbrainz_spark/popularity/mlhd.py`: Minor update to keep the old code compatible with the new `dump_location` tracking.

# AI Usage
- Received assistance with pinpointing the logical bug and writing out the fix.
- All code has been thoroughly reviewed, manually tested and fully understood prior to submission.

### DIFF
--- a/listenbrainz_spark/popularity/listens.py
+++ b/listenbrainz_spark/popularity/listens.py
@@ -3,7 +3,7 @@ from typing import List, Optional
 
 from listenbrainz_spark.path import LISTENBRAINZ_POPULARITY_DIRECTORY
 from listenbrainz_spark.popularity.common import get_popularity_per_artist_query, \
-    get_release_group_popularity_per_artist_query, get_popularity_query
+    get_release_group_popularity_per_artist_query, get_release_group_popularity_query, get_popularity_query
 from listenbrainz_spark.postgres.release import get_release_metadata_cache
 from listenbrainz_spark.stats.incremental.query_provider import QueryProvider
 from listenbrainz_spark.stats.incremental.range_selector import ListenRangeSelector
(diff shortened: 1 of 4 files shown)


MonkeyDo:
Hello, I would caution against touching important parts of the code without knowing exactly what you are doing.
The fact that you had to use an LLM to understand and analyse the issue as well as write "a fix" (to what issue in particular remains a question) does not give me a whole lot of confidence.
For example

> All code has been thoroughly reviewed, manually tested and fully understood prior to submission.

How did you test the code changes exactly?
Did you run the incremental database dump scripts (if so, which?), and how did you compare the resulting dumps?
How does your solution compare to the existing mechanism using from_date and to_date to decide on recalculating aggregates?

MonkeyDo:
Closing for lack of activity and unclear methodology for testing.
Not keen on merging this if it wasn't tested, and neither is there very clear explanations of what the issue was with release group listen counts.

PR #3674, opened 2026-04-13
LB-1971: lps floor options

Gopal7387:
### Summary
Implements the LB-1971 overhaul for the "LPs on the Floor" Art Creator template, adding scene selection, wear & tear options, and a record player colour overlay.
Root Cause:
The existing template used a single hardcoded background image with no user-configurable options, limiting visual variety.
Fix

Added three new photo scenes (close, vinyl, many) alongside the existing wood scene, each with its own bg.png, worn 1.png, worn 2.png, and player color.svg/.png assets
Extended art_api.py to accept scene, wear_tear, and player_color query params with validation and fallback defaults
Updated lps-on-the-floor.svg to use bg_url, worn_url, player_color_url, and feColorMatrix RGB values passed from the backend
Added four scene-picker icon SVGs (cover-art_photo-1..4.svg) for the frontend selector
Updated ArtCreator.tsx with scene radio selector, wear & tear dropdown, and colour picker; preview URL refreshes reactively via existing debounced mechanism
Added test_cover_art_lps_on_the_floor_params unit test covering scene, wear & tear, and player colour assertions

### Files Changed

listenbrainz/webserver/views/art_api.py
listenbrainz/webserver/templates/art/svg-templates/lps-on-the-floor.svg
listenbrainz/webserver/views/test/test_art.py
frontend/js/src/explore/art-creator/ArtCreator.tsx
frontend/img/art/lps-on-the-floor/ (new scene assets + icons)

### Testing

Unit test passes: test_cover_art_lps_on_the_floor_params verifies correct bg, worn, and player color URLs plus feColorMatrix values for scene=vinyl, wear_tear=loved, player_color=ff0000
Manually verified preview updates reactively in Art Creator for all scene/wear/colour combinations

AI Usage
If you did use AI:
[☑️] I did not use AI

Screenshot 2026-04-13 at 4 02 48 PM


Aerozol:
Thanks, the screenshot looks good Gopal! I won't be able to feedback more until this is on test or based on more pics.
Some v. early feedback:

- The 'Mama...' album does not line up with the overlay. This is the part that I said I don't want to micromanage - if you need the layer to be changed because it's not possible to fit in a cover, please let me know.
- Wear & Tear is 'new' but I can see some wear on the records. As per the ticket, there should be no 'wear' layer on that option. I didn't put it in the ticket, but the default should be 'used'.

Gopal7387:
Fixed both points. wear_tear default changed to "used", and wood scene covers now rotate around their own center (rotate(deg, cx, cy)) to fix the Mama...fication misalignment. 
Let me know if it looks good on the test server.

anshg1214:
@Gopal7387 Please fill out the AI Declaration in the PR Description

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -800,6 +823,72 @@ export default function ArtCreator() {
                     </div>
                   </>
                 )}
+                {style.name === TemplateNameEnum.lPsOnTheFloor && (
+                  <>
+                    <small>Choose a grid layout:</small>
```
MonkeyDo:
```suggestion
                    <small>Choose a layout:</small>
```

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -223,6 +223,9 @@ export default function ArtCreator() {
     defaultStyleOnLoad.defaultColors[2]
   );
   const [customPresets, setCustomPresets] = useState<ColorPreset[]>([]);
+  const [scene, setScene] = useState<string>("wood");
+  const [wearTear, setWearTear] = useState<string>("used");
+  const [playerColor, setPlayerColor] = useState<string>("#ffffff");
```
MonkeyDo: I feel this could be named a bit clearer so we know it relates to the vinyl stuff.
Maybe prefixing all three with "vinyl"? `vinylScene`, `vinylWearTear` and `vinylPlayerColor`

Actually I would probably also rename "scene" as "layout", so `vinylLayout`

frontend/js/src/explore/art-creator/ArtCreator.tsx
```diff
@@ -223,6 +223,9 @@ export default function ArtCreator() {
     defaultStyleOnLoad.defaultColors[2]
   );
   const [customPresets, setCustomPresets] = useState<ColorPreset[]>([]);
+  const [scene, setScene] = useState<string>("wood");
+  const [wearTear, setWearTear] = useState<string>("used");
```
MonkeyDo: Both of those could do with stronger types to validate the strings used. See examples at the top of this same file

MonkeyDo (changes requested):
Hi,

Not sure how you tested this, but it is not working for any of the layouts for me, the transforms look completely off

Pretty funky, but not exactly the brief:

Image

Gopal7387:
Pushed the alignment fixes for close/vinyl/many, matrix transforms tuned 
to match each background, same as wood.
Also found and fixed a real bug along the way: art_api.py was hardcoded to load 5 images no matter the 
scene, which would've crashed close (8) and many (12) in prod. 

Just let me know if this looks well aligned.🙂

Screenshot 2026-08-03 at 5 21 30 PM
Screenshot 2026-08-03 at 5 21 47 PM
Screenshot 2026-08-03 at 5 22 06 PM

Aerozol:
Thanks Gopal! Those need more work with the alignment - all the gaps along the edges are going to be super obvious when there is dark cover art.

I would err towards slightly too big covers, rather than too small, that's less obvious. The only place you have to be careful with bigger art is if it's overlapping another piece of art (in which case it will be obvious!)

PR #3681, opened 2026-04-16
LB-1804: Integrate tidal music 

mAmanullah7:
Implement Tidal Music as a music source


frontend/js/src/utils/types.d.ts
```diff
@@ -1,6 +1,44 @@
 /* eslint-disable camelcase */
 
 declare module "react-responsive";
+declare module "@tidal-music/player" {
```
MonkeyDo, line 4: Just checking that the package doesn't export types in that package?

package.json
```diff
@@ -34,6 +34,7 @@
   "homepage": "https://listenbrainz.org",
   "dependencies": {
     "@babel/runtime": "^7.20.6",
+    "@tidal-music/player": "^0.2.0",
```
MonkeyDo, line 37: Should be updated to latest: [0.20.1](https://github.com/tidal-music/tidal-sdk-web/tree/player-0.20.1)
```suggestion
    "@tidal-music/player": "^0.20.1",
```

MonkeyDo (commented):
Thanks for adding this, and sorry to only have another look at it now, it's been open for quite a while!
Overall looks good, I'll do some local testing but in the meantime, a couple of comments:

PR #3683, opened 2026-04-16
Playlist metadata hidden when description is too long

uzman2406:
# Problem
[LB-1980](https://tickets.metabrainz.org/browse/LB-1980)
Playlist metadata becomes hidden in grid views when playlist descriptions are too long. 
This happens because descriptions are rendered without proper truncation, causing layout overflow and pushing important information out of view. 

# Solution

- Applied the existing `.text-summary` class to playlist descriptions in both list and grid views to enable controlled multi-line truncation.
- Reduced description length to 2 lines for better layout balance.

* ✅ I did not use any AI

# Screenshot:

Before:

Screenshot 2026-04-05 212237

Screenshot 2026-04-16 220502

After:

Screenshot 2026-04-16 205812

Screenshot 2026-04-16 220127

### DIFF
--- a/frontend/css/sass/playlists.scss
+++ b/frontend/css/sass/playlists.scss
@@ -256,5 +256,3 @@ $long-description-lines: 8;
   align-items: center;
   gap: 1rem;
 }
-
-
--- a/frontend/css/sass/playlists.scss
+++ b/frontend/css/sass/playlists.scss
@@ -256,3 +256,5 @@ $long-description-lines: 8;
   align-items: center;
   gap: 1rem;
 }
+
+
--- a/frontend/css/sass/playlists.scss
+++ b/frontend/css/sass/playlists.scss
@@ -260,6 +260,6 @@ $long-description-lines: 8;
 }
 .playlist-card-list-view .text-summary,
 .playlist .text-summary {
-mask-image: none;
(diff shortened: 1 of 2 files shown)


frontend/css/sass/playlists.scss
```diff
@@ -231,8 +231,6 @@ $long-description-lines: 8;
   -webkit-box-orient: vertical;
   padding-right: 2em;
   margin-top: 0.5rem;
-  mask-image: linear-gradient(180deg, #000 60%, transparent 98%);
-  -webkit-mask-image: linear-gradient(180deg, #000 60%, transparent 98%);
```
anshg1214, line 235: Why have you removed these css properties? We're using this class on the playlist page as well.

uzman2406:
Thank you so much for the review:)
These lines were creating a fading effect like below and I thought it's not looking good that's why I removed them but I did not realize that the same class is used in playlist page as well .


image


 Instead of removing those lines I will just add this  
".playlist-card-list-view .text-summary, 
.playlist .text-summary {
 mask-image: none;
 -webkit-mask-image: none;
}"
this to remove fading effect .

MonkeyDo (approved):
Thank you!

PR #3684, opened 2026-04-18
LB-1983 Showing Artist Name Relationship 

yateendogra2109:
# Problem
[LB-1983](https://tickets.metabrainz.org/browse/LB-1983?jql=project%20%3D%20%2210101%22)

Musicbrainz shows the different stage names of a particular artist as `Performance name of ` or `Also performs as`,
But this section is missing in Listenbrainz . Sometimes we remember the Artist name but not the `stage/performance` name or we forget the real name.
So we wont be able to see the full work of the artist.

# Solution

We now fetch and display these "performance name" relationships directly on the artist page.
If artist has an aliases or a real name behind a stage name, we will see a new "Artist relationships" section showing who they're connected to. On clicking the artist name , we will be directed to the artist page of that connected artist.If an artist has more than 3 relationships, only the first 3 are shown by default — we can click "Show N more" to see the rest, and "Show less" to collapse again.

Stored the relationship data in the artist metadata cache so that the  new MusicBrainz API is not hit for every repeated search.

# Testing

Since full  MusicBrainz database is not available  in local dev environment, so I manually inserted test data directly into the cache table — two artists, one with 4 stage name links, the other with a single link back to the real person

Also The artist page was crashing entirely when the MusicBrainz database wasn't available, so added proper error handling so that sections (like popular tracks) just show up empty instead of taking down the entire page - changes were made in 
`entity_pages.py ` file.

Screenshot 2026-04-17 045008

  Screenshot 2026-04-17 045018

# AI usage
 I took help of AI to write the  SQL query for fetching performance name relationships in `mb_artist_metadata_cache.py`

### DIFF
--- a/admin/timescale/create_tables.sql
+++ b/admin/timescale/create_tables.sql
@@ -142,6 +142,7 @@ CREATE TABLE mapping.mb_artist_metadata_cache (
     artist_data             JSONB NOT NULL,
     tag_data                JSONB NOT NULL,
     release_group_data      JSONB NOT NULL
+    artist_relationship_data JSONB NOT NULL
 );
 
 -- the various mapping columns should only be null if the match_type is no_match, otherwise the columns should be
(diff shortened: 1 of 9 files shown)


MonkeyDo:
Hello, did you discuss this strategy with anyone before implementing it?

yateendogra2109:
> Hello, did you discuss this strategy with anyone before implementing it?

No I didn't discussed the strategy with anyone , I analyzed how other Artist information is loaded from backend to frontend and implemented similarly based on my understanding.Though I used AI for writing the SQL querry .

PR #3685, opened 2026-04-19
LB-1977: Fix Art Creator black text by using CSS custom properties

Shreshtthh:
# Problem
  
  Fixes LB-1977: Art creator: Some rows/text is displayed in black
  
  This appears to be a Chromium GPU compositing issue. When colors are updated via string interpolation inside the `<style>` block, the browser correctly recomputes the CSSOM, but since the SVG DOM structure remains unchanged, the GPU compositor reuses a stale cached texture. This explains why DevTools shows the correct computed styles while the rendered output intermittently shows black text.
  
  # Solution
  
  Replaced interpolated color values in the `<style>` block with CSS custom properties:
  
  - `Preview.tsx` : Color values are now set as `--lb-*` custom properties on the root `<svg>` via inline style. The `<style>` block uses `var()` references instead of interpolated values, making it effectively static. Color changes only change the style attribute, which triggers the browser's CSS variable invalidation path and forces a proper repaint.
  
  - `utils.tsx`: Added `resolveCustomProperties()` to handle Canvg export compatibility. Canvg doesn't support CSS `var()`, so this function resolves `var(--lb-*)` references to their computed values before passing the SVG to Canvg. It also strips `!important` since the Canvas API's `addColorStop` cannot parse values like `#ff0000 !important`.
  
  - `resolveCustomProperties.test.ts` (new): Unit tests for the resolve function.
  
  # AI usage
  
  #### If you did use AI:
  Used Claude and Gemini to understand the Chromium GPU compositing behavior causing the rendering glitch, navigate the art-creator codebase, and plan the CSS custom properties refactor. All changes were reviewed and understood before committing.
  
  
  # Action
  
  Manual verification has been done- 
  - Rapidly changed text and background colors via the color picker. No black flicker was observed
  - Download as PNG works correctly with user-selected colors
  - Copy SVG produces valid output with var(--lb-*) references and matching inline style values on the root `<svg>` : 
  
  ```html
  <svg class="preview" style="--lb-text-color: #321529; --lb-bg-color-1: #9b3361; --lb-bg-color-2: #a76798;">
    <style>
      text > tspan, .accent-color { fill: var(--lb-text-color) !important; }
      stop:first-child { stop-color: var(--lb-bg-color-1) !important; }
      ...
    </style>
  </svg>
  ```
  
Since this bug is a non-deterministic GPU compositing glitch, manual verification on a setup that can reproduce it consistently would be valuable. 
  
Note on the fix- It is possible that this issue might turn out to have a different root cause. Even then, this refactor seems like a net improvement as it replaces string interpolation inside `<style>` with CSS custom properties, which is the standard approach for dynamic styling in SVG. The `<style>` block is now effectively static, making the rendering pipeline more predictable and easier to maintain.


frontend/js/tests/explore/art-creator/resolveCustomProperties.test.ts
```diff
@@ -0,0 +1,53 @@
+import { resolveCustomProperties } from "../../../src/explore/art-creator/utils";
+
+describe("resolveCustomProperties", () => {
+  it("resolves --lb-* custom properties in SVG string", () => {
+    const input = `<svg style="--lb-text-color: #ff0000; --lb-bg-color-1: #00ff00; --lb-bg-color-2: #0000ff"><style>text { fill: var(--lb-text-color) !important; } stop:first-child { stop-color: var(--lb-bg-color-1) !important; } stop:nth-child(2) { stop-color: var(--lb-bg-color-2) !important; }</style></svg>`;
```
MonkeyDo, line 5: Let's give this a real world example, the output of your (or my) art creator SVG.
Currently you only test a contrived ideal scenario, which does not give us full confidence in the test (including for future changes in implementation).

MonkeyDo (changes requested):
Hello!

I deployed this to the test website, and it does not seem like the PR fixes the issue.
I don't know how you tested it but it does not seem sufficient...

Image

Generally, if you are just following an LLM's conclusions and instructions I would advise against it, they are good for prototyping new things but bad at understanding and replicating bugs in existing code.

I suggest going back to square one and trying to understand where the issue comes from.
One way or another it is related to the styling of visited links, but that's all the info I can gather without looking into it deeper. Issue seems to happen to link I have opened in a new tab, sometime only on hovering those links.

Shreshtthh:
Hello @MonkeyDo . Thank you for the feedback as always and sorry I got a little late. 
I really did end up overcomplicating the issue by a long shot. My original research led me to believe this was a GPU compositing issue since I found some cases where Chromium's hardware acceleration causes SVG fills to render incorrectly specifically because the compositor can fail to properly redraw SVG elements oninvalidation, reusing stale GPU textures instead. Since DevTools showed the correct computed styles while the rendered output was black, it seemed to fit perfectly. Once I committed to that hypothesis, I kept going deeper into it (CSS custom properties, variable invalidation paths) instead of stepping back to look for other root causes. LLMs definitely had a part to play but all in all, I should've been more thorough.

Went back to square one and found this -  [Chromium bug #40356084](https://issues.chromium.org/issues/40356084) - "SVG a:visited links erroneously get fill:black style".

The SVG templates wrap artist/release names in `<a>` tags. When a user visits one of those links, Chromium incorrectly applies `fill: black` to the `<a>` element in its `:visited` state. Since the [SVG spec](https://www.w3.org/TR/SVG11/painting.html#FillProperty) defines black as the default fill value, and fill is in the [allowed :visited property list](https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector#limits_to_visited_link_styles) per the privacy model, the browser's `:visited` styling overrides the intended color. (Funny enough, I actually considered the `:visited` state early on. But I remembered reading about how browsers lock down `:visited` styles to prevent history sniffing and spoofing, so I prematurely crossed it off my list. I completely forgot that color and fill are some of the few exceptions browsers actually do allow you to style.)

I've now reverted the CSS custom properties refactor and applied the documented workaround from the Chromium bug forcing all SVG `<a>` elements to inherit fill from their parent regardless of visited state:

```css
a, a:visited, a:link, a:hover {
  fill: inherit !important;
  stroke: inherit !important;
} 
```

Also added `!important` stripping before passing SVGs to Canvg for PNG export, since the Canvas API's `addColorStop `cannot parse values like `#ff0000 !important`.

Testing
I was able to reproduce the bug intermittently on my end, and the fix appears to resolve it. Could you re-test on the test site when you get a chance? If there's still something missing, happy to dig deeper.

MonkeyDo:
Thanks for looking into it again, this time around you did find the issue, and the appropriate fix, nice work.
The fix works on the preview component, but it looks like the custom text color is not applied in the generated bitmap image (download or copy buttons).
I think it is probably because of applying the removal of `!important` in the entire SVG, as the only other significant change, but I have not done more testing than confirming that the prod website generates images with the text color applied.
Maybe the stripping will need to be applied  more selectively, but it's worth checking first if we need that `!important` there in the first place; preliminary testing i did suggests maybe not...

Shreshtthh:
hello! I checked and `!important` was indeed not required. i kept it earlier because removing it threw `'CanvasGradient': The value provided ('#9b3361 !important') could not be parsed as a color.` but i figured that it was since I was stripping it in `utils.tsx` with a regex that wasn't matching correctly, so the old `!important` values were still reaching Canvg. This time I removed `!important` directly from the CSS declarations in `Preview.tsx` and everything seems to work fine.

MonkeyDo (approved):
Looks good, thank you for fixing this annoying bug!

PR #3688, opened 2026-04-24
Allow caching immutable frontend assets

MonkeyDo:
Compiled frontend resources like JS and CSS files have a cache-busting hash in their name, so there is no reason why it should not be cached "indefinitely".
This was a cache misconfiguration hotfix in #3131, applied globally as what looks like the nuclear option, in retrospect.
I only modified cache control for compiled assets and images in this PR, but it would be good to investigate whether we could apply the cache control restrictions to specific paths instead to fix the issue describes in 3131.

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -192,8 +192,20 @@ def create_app(debug=None):
 
     @app.after_request
     def add_cache_header(response):
-        response.cache_control.private = True
-        response.cache_control.public = False
+        if request.path.startswith('/static'):
+            # cache hashed assets (JS + CSS) for 1 year
+            hashed_assets_max_age = 31536000
+            # cache image assets for 1 week
+            image_assets_max_age = 604800
+            response.cache_control.public = True
+            response.cache_control.immutable = True
+            if request.path.startswith('/static/dist'):
+                response.cache_control.max_age = hashed_assets_max_age
+            elif request.path.startswith('/static/img'):
+                response.cache_control.max_age = image_assets_max_age
+        else:
+            response.cache_control.private = True
+            response.cache_control.public = False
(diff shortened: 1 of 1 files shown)


listenbrainz/webserver/__init__.py
```diff
@@ -192,8 +192,20 @@ def after_request_callbacks(response):
...
-        response.cache_control.public = False
+        if request.path.startswith('/static'):
+            # cache hashed assets (JS + CSS) for 1 year
+            hashed_assets_max_age = 31536000
+            # cache image assets for 1 week
+            image_assets_max_age = 604800
+            response.cache_control.public = True
+            response.cache_control.immutable = True
+            if request.path.startswith('/static/dist'):
+                response.cache_control.max_age = hashed_assets_max_age
```
rinsuki: I think `/static/img` are not immutable

```suggestion
            if request.path.startswith('/static/dist'):
                response.cache_control.immutable = True
                response.cache_control.max_age = hashed_assets_max_age
```

MonkeyDo: Yes good catch,. thank you!

MonkeyDo:
Adding a note to self that we might want to process our images folder with webpack too, to add a content hash so we can cache those more aggressively as well.

anshg1214 (commented):
Do we want dev and prod to share the same static caching policy?

MonkeyDo:
> Do we want dev and prod to share the same static caching policy?

I checked again locally, there are other cache-busting mechanism in place (etag, last-modified timestamp) that prevent stale assets in local dev, so these settings work fine for dev too.
(and I also ensured it wasn't because I had "disable cache when devtools are open" enabled)

amCap1712 (approved):

PR #3689, opened 2026-04-26
Enable webpack default minimizer (terser)

rinsuki:
# Problem

`/static/dist/indexPage.*.js` is not minified and pretty big (8MB uncompressed, 2MB gzipped).

# Solution

Just (re-)enable JavaScript file minify. Webpack does by default, but override at https://github.com/metabrainz/listenbrainz-server/commit/d032dd8163c3f4ee9bc052822afd6357d5589a39#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166fL118 . according to commit message, I think they didn't mean to disable JS minify, they just want to replace CSS minifier.

`"..."` will be replaced with default config, see https://webpack.js.org/configuration/optimization/#optimizationminimizer for more information.

  * I ran `docker build` and check output is minified now, but didn't test about actual application works (or not). probably does (since it just inherits webpack default configuration, I assume they have a relatively safe configuration), but you might be want to test it if you cared about that

# AI usage
* [x] I did not use any AI

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -108,6 +108,8 @@ module.exports = function (env, argv) {
     },
     optimization: {
       minimize: isProd,
+      // "..." is a shortcut to access the default Webpack minimizer config.
+      // see https://webpack.js.org/configuration/optimization/#optimizationminimizer
       minimizer: [new CssMinimizerPlugin(), "..."],
     },
     resolve: {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -108,7 +108,7 @@ module.exports = function (env, argv) {
     },
     optimization: {
       minimize: isProd,
-      minimizer: [new CssMinimizerPlugin()],
+      minimizer: [new CssMinimizerPlugin(), "..."],
     },
     resolve: {
       modules: [


webpack.config.js
```diff
@@ -108,7 +108,7 @@ module.exports = function (env, argv) {
     },
     optimization: {
       minimize: isProd,
-      minimizer: [new CssMinimizerPlugin()],
+      minimizer: [new CssMinimizerPlugin(), "..."],
```
MonkeyDo, line 113: Someone (likely me) is bound to be confused by this in the future... Adding a comment to save some time.
```suggestion
      // "..." is a shortcut to access the default Webpack minimizer config.
      // see https://webpack.js.org/configuration/optimization/#optimizationminimizer
      minimizer: [new CssMinimizerPlugin(), "..."],
```

MonkeyDo (approved):
Thank you so much, I appreciate you fixing my past mistakes :)

Can confirm output assets are much smaller when deployed.
Obviously it was not the intent to disable minification in the first place...

PR #3691, opened 2026-04-28
Fix and enhance service icon handling for multiple platforms

Szpieg0:
# Refactor: Centralized brand icons (Simple Icons) and array-based relationship handling

## Problem
The previous implementation of artist/album relationship links and brand icons had several major flaws:

1. **Icon Inconsistency:** Many music and social platforms (like TikTok, Pandora, or Qobuz) were missing proper logos because they aren't available in the free version of FontAwesome. 
2. **Data Loss (Object Collision):** Relationship data (`rels`) was stored as an object `{ [type]: url }`. This meant if an artist had multiple links of the same type (e.g., two different "streaming" links for Spotify and Deezer), the last one would overwrite the others, leading to missing data.
3. **Maintenance Overhead:** Brand icon imports were scattered across the codebase, making it difficult to maintain a consistent look or provide fallbacks for niche services.

### 1. Centralized Brand Icons with Simple Icons Integration
* Created a new hub: `brandIcons.ts` (`frontend/js/src/common/icons/`).
* **Adapter Pattern:** This module wraps **Simple Icons** (which has 3000+ brand logos) and converts them into **FontAwesome-compatible** `IconDefinition` objects. This allows us to use `<FontAwesomeIcon>` with virtually any brand logo (TikTok, Pandora, Amazon Music, etc.) without changing our component logic.
* Migrated all platforms (Spotify, YouTube, Bandcamp, etc.) to use these high-quality, brand-specific icons.

### 2. Migration to Array-Based Relationships
* Refactored the data structure of `rels` from a **Dictionary/Object** to an **Array of Objects** (`ArtistRel[]`).
* **Deduplication:** The backend now stores all unique URLs, preventing the "key collision" problem where one link would hide another.
* **Logic Simplification:** Updated `AlbumPage.tsx`, `ArtistPage.tsx`, and `MetadataViewer.tsx` to iterate over arrays using `.map()`, which is more "idiomatic" (natural) for React.

### 3. Utility & Rendering Improvements
* Enhanced `getRelIconLink` in `album/utils.tsx` to detect services based on URL patterns, providing correct icons and brand colors for a wider range of streaming services.
* **React Key Stability:** Used relationship URLs as unique `key` props, ensuring the React **Reconciliation** engine can track elements efficiently during updates.

### 4. Code Cleanup
* Removed unused utility imports (`isEmpty`, `isPlainObject`) that were previously needed to handle object-based checks.
* Deleted obsolete custom icon files.

- [ ] I have run the code and manually tested the changes

# AI usage
* [ ] I did not use any AI
* [x] I have used AI in this PR

#### If you did use AI:
* [x] I used AI tools for communication (Structuring and refining the PR description)


frontend/js/src/album/AlbumPage.tsx
```diff
@@ -341,10 +341,10 @@ export default function AlbumPage(): JSX.Element {
...
         <div className="right-side gap-1">
           <div className="entity-rels">
-            {!isEmpty(artist?.artists?.[0]?.rels) &&
-              Object.entries(
-                artist?.artists?.[0]?.rels ?? {}
-              ).map(([relName, relValue]) => getRelIconLink(relName, relValue))}
+            {Boolean(artist?.artists?.[0]?.rels?.length) &&
+              artist?.artists?.[0]?.rels?.map((rel) =>
+                getRelIconLink(rel.type, rel.url)
+              )}
```
MonkeyDo, line 347: Not sure why this needed refactoring?

frontend/js/src/album/utils.tsx
```diff
@@ -141,7 +201,7 @@ export function getRelIconLink(relName: string, relValue: string) {
   }
   return (
     <a
-      key={relName}
+      key={relValue}
```
MonkeyDo, line 204: Why this change?

frontend/js/src/artist/ArtistPage.tsx
```diff
@@ -354,9 +354,9 @@ export default function ArtistPage(): JSX.Element {
         <div className="right-side gap-1">
           <div className="entity-rels">
             {artist &&
-              !isEmpty(artist?.rels) &&
-              Object.entries(artist.rels).map(([relName, relValue]) =>
-                getRelIconLink(relName, relValue)
+              Boolean(artist?.rels?.length) &&
+              artist.rels.map((rel) =>
+                getRelIconLink(rel.type, rel.url)
```
MonkeyDo, line 359: Same comment as above, why this refactor?

frontend/js/src/common/icons/brandIcons.ts
MonkeyDo, line 1: This seems to me a bit convoluted to end up with fontawesome compatible exports, when we were already using the FA icons directly

frontend/js/src/common/icons/faInternetArchive.tsx
MonkeyDo, line 1: If this one custom icon was the sole reason for refactoring everything and adding a new library, when this was working completely fine, I don't think it was necessary at all.

frontend/js/src/metadata-viewer/components/MetadataViewer.tsx
MonkeyDo, line 1: To me this reads as an LLM refactoring things without a specific reason, so I am questioning whether these changes improve anything (including readability).
If it makes the code less readable is it really an improvement?

frontend/js/src/metadata-viewer/types.d.ts
MonkeyDo, line 1: Yes, this additional type is useful 👍

mbid_mapping/mapping/mb_artist_metadata_cache.py
MonkeyDo, line 1: Changes to this file do not seems related to the PR and should be reverted

MonkeyDo (changes requested):
Hello, thanks for opening a PR!

I am not convinced there is a need to add a new library for most of these changes. Which brand icons were missing from our current setup with Fontawesome brands?
Similarly I'm not sure what the brandIcons.ts brings us that wasn't already in place.

Szpieg0:
In the library that is currently used there were no icons for Pandora, musicmatch, quobuz etc., so I had to combine both libraries to ensure the best possible "coverage".
Thanks to the fact that the libraries are now connected, there should be no problem with icons😊

frontend/js/src/common/icons/brandIcons.ts
Szpieg0: Hmmm after a bit of thinking you are right - I think that I have to do this PR once more and I will add only this icons that aren't available in FA. This will be better solution...

Szpieg0:
After thinking I will to this PR once more and I will change only these parts of code that will improve "coverage" of icons. Sorry for problem🙂

MonkeyDo:
> After thinking I will to this PR once more and I will change only these parts of code that will improve "coverage" of icons. Sorry for problem🙂

All good, thank you for looking into it closer :)

For what it's worth, there might be some new brand icons that have landed in fontawesome since whatever version we have installed, so it could be a good idea to start by checking that and updating the library version if there are indeed new useful brand icons.

I am personally not adverse to our current solution of making custom FA icons for the few that are missing; if you want I can help with that process if you can make a list of missing icons (possibly after the FA upgrade)

PR #3697, opened 2026-05-06
Feed: Load cover art  from release-group

MonkeyDo:
# Problem

I am seeing cache key collisions on the feed page when release MBIDs are not available.
This cover art is not the cover for any of those three entries, but instead is the one loaded first with missing information:
image

image

# Solution

Two issues here:
1. cache collision when missing some data: the cacke key end up being ` "ca:::"` for all of these entries, so they show data for the first loaded cover with that cache key
2. cover art is not loaded from release group even when an MBID is available (only loading from release MBID) 

* [ x ] I have run the code and manually tested the changes

Shown here: first cover loaded from RG data, others don't have the data to load cover art but don't reuse another cover
image

# AI usage

* [ x ] I did not use any AI


anshg1214 (approved):
LG

PR #3698, opened 2026-05-07
Add detailed validation message for history imports 

MonkeyDo:
# Problem

Spotify users have been left confused by the imported vs. attempted count we show on the importer page details.

The only listens we filter out are less-than-30-seconds plays and incognito mode listens, so we should clarify how many of either are the cause for the non-imported listens.

# Solution
This PR adds an optional detailed_message in import details that we can surface to the user.
The implementation details of what goes into that message are left to each importer and optional.
This allows for implementing something for Spotify now (keeping count of skips and incognito), and incrementally improve other services with their own logic.

# AI usage

#### If you did use AI:

Used GH Copilot to make some of the changes, based on a detailed implementation plan I wrote first.
Reviewed each modified line manually (because I don't trust it) and made my own changes on top.

image
image

### DIFF
--- a/frontend/js/src/settings/import/ImportListens.tsx
+++ b/frontend/js/src/settings/import/ImportListens.tsx
@@ -57,6 +57,7 @@ type ImportMetadata = {
   status: ImportStatus;
   attempted_count?: number;
   success_count?: number;
+  detailed_message?: string;
 };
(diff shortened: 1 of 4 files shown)


anshg1214 (approved):
LG!

PR #3699, opened 2026-05-07
feat: upgrade sentry to work with react router v7

anshg1214:
Sentry's doc used for reference: https://docs.sentry.io/platforms/javascript/guides/react/features/react-router/v7/

### DIFF
--- a/frontend/js/src/index.tsx
+++ b/frontend/js/src/index.tsx
@@ -25,7 +25,7 @@ document.addEventListener("DOMContentLoaded", async () => {
     Sentry.init({
       dsn: sentry_dsn,
       integrations: [
-        Sentry.reactRouterV6BrowserTracingIntegration({
+        Sentry.reactRouterV7BrowserTracingIntegration({
           useEffect: React.useEffect,
           useLocation,
           useNavigationType,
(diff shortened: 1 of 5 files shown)


frontend/js/src/index.tsx
```diff
@@ -35,16 +36,18 @@ document.addEventListener("DOMContentLoaded", async () => {
       ],
       tracesSampleRate: sentry_traces_sample_rate,
     });
-  }
 
-  const { currentUser } = globalAppContext;
+    if (currentUser?.name) {
+      Sentry.setUser({ username: currentUser.name });
+    }
```
MonkeyDo, line 42: Not sure about tracking username, we should probably discuss the implications.
What use-case did you have in mind for it?

MonkeyDo (commented):
Apart from the username, the PR looks good. I deployed on test and have seen an issue come in

frontend/js/src/index.tsx
anshg1214: Having a username will make it easier to triage an error. In cases where an issue is caused by stale/corrupted user data or something user-specific, it becomes much easier to identify patterns and debug the issue quickly.

MonkeyDo: I suppose it could, but I am wary about tracking more identifiable user information, especially if it could potentially make users re-identifiable. I'm going to ask on the security channel to get other opinions.

MonkeyDo: Looks like this is in accordance with what MB does on sentry as well, green light :green_circle:

MonkeyDo (approved):

PR #3700, opened 2026-05-07
Fix validation logic for listens endpoint

anshg1214:
We don't need to log this error on sentry: https://metabrainz-foundation-inc.sentry.io/issues/95469296/


MonkeyDo (approved):
Yep, that makes sense

PR #3704, opened 2026-05-11
Use retry/backoff mechanism for all API calls

MonkeyDo:
Calls to MB and LB APIs are not respecting rate limits.
We do have a mechanism in place, implemented so that we could switch to it easily.
Implements basic retry strategy that takes 429 response + retry-after headers into account

### DIFF
--- a/frontend/js/src/utils/APIService.ts
+++ b/frontend/js/src/utils/APIService.ts
@@ -1,13 +1,10 @@
 import { isNil, isUndefined, kebabCase, lowerCase, omit } from "lodash";
+import fetchBuilder from "fetch-retry";
 import { TagActionType } from "../tags/TagComponent";
 import type { SortOption } from "../explore/fresh-releases/FreshReleases";
 import APIError from "./APIError";
 import type { Flair } from "./constants";
 import { Modes } from "../explore/lb-radio/components/Prompt";
-
-const fetchWithRetry = require("fetch-retry")(
-  (...args: Parameters<typeof fetch>) => window.fetch(...args)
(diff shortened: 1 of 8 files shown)


frontend/js/src/utils/APIService.ts
```diff
@@ -22,7 +19,7 @@ export default class APIService {
   MAX_LISTEN_SIZE: number = 10000; // Maximum size of listens that can be sent
   private fetchWithRetry: any;
```
anshg1214: Let's define the types for fetchWithRetry here.
```suggestion
  private fetchWithRetry: (input: RequestInfo, init?: RequestInitWithRetry) => Promise<Response>;

```

frontend/js/src/utils/APIService.ts
```diff
@@ -534,7 +531,7 @@ export default class APIService {
       // Now submitListens focused on payload handling
 
       return this.withRetry(async () => {
-        const response = await fetch(url, {
+        const response = await this.fetchWithRetry(url, {
```
anshg1214, line 545: Once we add the types - This throws an error because `fetchWithRetry` does not support the structure `URL`.

frontend/js/tests/user/follow/UserSocialNetwork.test.tsx
MonkeyDo, line 1: These tests were broken...
The APiService fixture uses bthe `/foo/1/` prefix and the wrong similar-users endpoint.

anshg1214 (approved):
LG

PR #3705, opened 2026-05-12
Fix Music Neighborhood page

MonkeyDo:
Currently fetching wikipedia-extract from MB is failing, which seems to break the page because of our use of Promise.all (one failing promise = all fail).
This PR separates promises with [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) which is a bit cumbersome to use, but maintains parallel requests.

Obviously this does not resolve the root issue, but the page should not break just because we can't load the wikipedia extract.
In the meantime, the gateway configuration was changed to allow incoming trafic from LB to MB without marking it as potential bot (and serving a proof-of-work page), solving the root issue.
We might still want to implement a proxy endpoint on LB server that calls MB internally before returning the response to the client.

Tested on test.LB, page loads even with the failing WP call
Before:
image

After:
image

### DIFF
--- a/frontend/js/src/explore/music-neighborhood/MusicNeighborhood.tsx
+++ b/frontend/js/src/explore/music-neighborhood/MusicNeighborhood.tsx
@@ -145,21 +145,31 @@ export default function MusicNeighborhood() {
   const fetchArtistInfo = React.useCallback(
     async (artistMBID: string): Promise<ArtistInfoType> => {
       const [
-        artistInformation,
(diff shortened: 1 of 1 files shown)


PaulKiddle:
I see this is merged now but if you wanted to avoid `allSettled` it might be more readable to add `.catch(()=>[])` to the API calls before passing them into `Promise.all`.

PR #3707, opened 2026-05-18
Connect to timescale through pgbouncer

amCap1712:
Configure all services to connect to timescale db through pgbouncer except dumps and timescale writer.


MonkeyDo (commented):
I think we probably need to define `SQLALCHEMY_TIMESCALE_PGBOUNCER_URI` in the dev config.py.sample file as well, to be the same value as `SQLALCHEMY_TIMESCALE_URI`.

I still see `SQLALCHEMY_TIMESCALE_URI` used directly in a bunch of files.
Some make sense to automatically bypass pgbouncer (mb_cache, mb_metadata_cache, typesense_index), while other probably not (soundcloud/spotify/apple metadata cache, tracks_of_the_year, tag_similarity, rtd_config, etc.)

Are they were omitted on purpose?

MonkeyDo:
Fixed the tests that were missing the new connection key, but still unsure about the hardcoded uses of the original `SQLALCHEMY_TIMESCALE_URI` in the mbid_mapping folder

amCap1712:
@MonkeyDo I left mbid_mapping completely alone. All of them are long running cron jobs anyway. We can do a second pass later imo.

PR #3709, opened 2026-05-18
Fix misc bugs

amCap1712:
1. Add missing 'status' column to 'get_user'.
2. Fix passing mbid list to bulk tag lookup query.
3. Run _process_listens in mapper under flask context.

### DIFF
--- a/listenbrainz/labs_api/labs/api/bulk_tag_lookup.py
+++ b/listenbrainz/labs_api/labs/api/bulk_tag_lookup.py
@@ -40,9 +40,18 @@ class BulkTagLookup(Query):
 
     def fetch(self, params, source, offset=-1, count=-1):
 
-        mbids = tuple([psycopg2.extensions.adapt(p.recording_mbid) for p in params])
+        mbids = []
+        for p in params:
+            for mbid in p.recording_mbid.split(","):
+                mbid = mbid.strip()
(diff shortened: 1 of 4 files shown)


listenbrainz/labs_api/labs/api/bulk_tag_lookup.py
```diff
@@ -54,7 +63,7 @@ def fetch(self, params, source, offset=-1, count=-1):
                              FROM tags.lb_tag_radio
                             WHERE recording_mbid IN %s'''
 
-                curs.execute(query, tuple([mbids]))
+                curs.execute(query, (tuple(mbids),))
```
MonkeyDo, line 66: Did not test this, and can't really attest to the necessity of the comma there at the end.
I guess re-read this change carefully 'cause I can't 🙈

MonkeyDo (commented):
Looks sane, with some doubt about this one:

listenbrainz/labs_api/labs/api/bulk_tag_lookup.py
amCap1712: This is a nested tuple and that comma is needed otherwise python will not treat it as one.

MonkeyDo (approved):

PR #3710, opened 2026-05-19
Use SQLAlchemy connection pooling

amCap1712:
We recently set up pgbouncer for connection pooling across containers.

The application level code currently doesn't do client side connection pooling. In the NullPool mode, ListenBrainz clients are opening a fresh TCP connection to pgbouncer for every single transaction. This caused host-level ephemeral port exhaustion on gaga, with some services seeing "Cannot assign requested address when connecting to other pgbouncer instances" error.

To address this issue, use queue pools for web services to reuse connections to pgbouncer and reduce dramatic churns.


MonkeyDo (approved):
Looks good from here, but not sure how to test this properly other than deploying and checking every container.

A reminder to set *_pool_size and *_max_overflow values in docker-server-configs

PR #3711, opened 2026-05-20
Use SQLAlchemy connection pooling in labs api

amCap1712:
Similar to #3710, use connection pooling mbid mapping writer and labs api.


MonkeyDo (approved):
Looks sane, same comment as on 3710, not sure how to test this other than in production (beta) and carefully.

MonkeyDo:
Seems like there is something that is crashing the test suite before it can output results.
Tried re-running the suite but same result

PR #3714, opened 2026-05-23
LB-1944: Implement SoundCloud OAuth 2.1 PKCE support

anshg1214:
[LB-1944](https://tickets.metabrainz.org/browse/LB-1944)

We're migrating to Soundcloud OAuth 2.1. We've added PKCE (Proof Key for Code Exchange) support for secure authorization.

Ref: https://developers.soundcloud.com/blog/oauth-migration

### DIFF
--- a/frontend/js/src/settings/music-services/details/MusicServices.tsx
+++ b/frontend/js/src/settings/music-services/details/MusicServices.tsx
@@ -410,6 +410,7 @@ export default function MusicServices() {
     const funkwhaleSuccess = params.get("success");
     const navidromeError = params.get("navidrome_error");
     const navidromeSuccess = params.get("navidrome_success");
+    const soundcloudError = params.get("soundcloud_error");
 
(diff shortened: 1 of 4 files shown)


MonkeyDo:
Couldn't test this after deploying on test.LB because of a mismatch in the callback URL.
How did you test this? local dev with your own soundcloud oauth app?

anshg1214:
I used the Production callback URL while sending a request to soundcloud.

MonkeyDo (approved):
Tested locally (with some manual manipulation required) and logged in fine. Good to go!

Do we need to do anything on the database side like signing out users that were previously signed in with old oauth?
Do we need to send those users a message to ask them to log in again?

anshg1214:
> Do we need to do anything on the database side like signing out users that were previously signed in with old oauth?
Do we need to send those users a message to ask them to log in again?

No, the functionality should work out of the box.

PR #3715, opened 2026-05-31
GSoc-26  Feature-1  Add Playlist Search (Part 1)

yateendogra2109:
# PR Description

This is the first PR for the Feature 1 ( Add search functionality on playlists ) . I had first implemented 
the basic search with a valid input query of greater than or equal to 3 chars with the help of existing user playlist 
search API `( /1/user/<username>/playlists/search API via APIService.searchPlaylistsForUser)`
This makes searching playlists easy, without relying on the global search.Existing sorting options are preserved when search results are loaded.
Then I properly implemented the pagination for the searched playlists shown on the screen .

Currently the scope of this pr is basic search with proper pagination.

# Testing 

For this PR I tested manually by creating playlists with different names/descriptions , directly from the frontend screen.
And the results are as expected.

## Testing Plan : 

-  Open your own playlists page and confirm the search bar appears next to Sort by .

-  Submit a query with fewer than 3 characters — original playlist list should remain unchanged

-  Search with a valid term (3+ chars) — only matching playlists should appear

-  If results exceed 25, verify pagination works (next/previous) and page count is correct

-  Start a new search while on page 2+ — URL should reset to page 1 and show results for the new query

# Upcoming Features

 In upcoming PR's for this feature . I will be implementing : 
- Spinner loader 
- Showing proper EmptyMessage for no search match or invalid query . 
- When user clear input after search  then restores original playlists automatically without hitting ENTER

### DIFF
--- a/frontend/js/src/user/playlists/Playlists.tsx
+++ b/frontend/js/src/user/playlists/Playlists.tsx
@@ -282,7 +282,8 @@ export default class UserPlaylists extends React.Component<
         searchPageCount: pageCount,
       },
       () => {
-        this.setSortOption(sortBy || initialSort);
+        const activeSort = sortBy || initialSort;
+        this.setSortOption(activeSort);
       }
     );
   };
@@ -308,7 +309,8 @@ export default class UserPlaylists extends React.Component<
       },
       () => {
         // Use the current sort if it exists otherwise use default
-        this.setSortOption(sortBy || initialSort);
+        const activeSort = sortBy || initialSort;
+        this.setSortOption(activeSort);
       }
     );
   };
--- a/frontend/js/src/user/playlists/Playlists.tsx
+++ b/frontend/js/src/user/playlists/Playlists.tsx
@@ -51,6 +51,7 @@ export type UserPlaylistsState = {
   sortBy: SortOption;
   view: PlaylistView;
   searchTerm: string;
+  activeSearchQuery: string;
   isSearching: boolean;
   searchPageCount: number;
 };
@@ -70,6 +71,7 @@ type UserPlaylistsClassProps = UserPlaylistsProps & {
   playlistType: PlaylistType;
   handleClickPrevious: () => void;
   handleClickNext: () => void;
+  resetToPageOne: () => void;
   handleSetPlaylistType: (newType: PlaylistType) => void;
   initialView: PlaylistView;
   setPersistentView: (view: PlaylistView) => void;
@@ -99,26 +101,51 @@ export default class UserPlaylists extends React.Component<
 
   constructor(props: UserPlaylistsClassProps) {
     super(props);
-    const { playlists, playlistCount, initialView, initialSort } = props;
+    const { playlists, pageCount, initialView, initialSort } = props;
     this.state = {
       playlists: playlists?.map((pl) => pl.playlist) ?? [],
       sortBy: initialSort,
       view: initialView,
       searchTerm: "",
+      activeSearchQuery: "",
       isSearching: false,
-      searchPageCount: playlistCount,
+      searchPageCount: pageCount,
     };
   }
 
   componentDidUpdate(prevProps: Readonly<UserPlaylistsClassProps>): void {
-    const { playlists, initialView, initialSort, playlistType } = this.props;
-    const { sortBy } = this.state;
+    const {
+      playlists,
+      initialView,
+      initialSort,
+      playlistType,
+      page,
+    } = this.props;
+    const { sortBy, activeSearchQuery } = this.state;
+    const pageChanged = prevProps.page !== page;
+    const isSearchActive = activeSearchQuery.length >= MIN_SEARCH_LENGTH;
+
+    if (pageChanged && isSearchActive) {
+      void this.performPlaylistSearch(activeSearchQuery, page).catch(() => {
+        this.setState({ isSearching: false, activeSearchQuery: "" });
+        toast.error(
+          <ToastMsg
+            title="Search failed"
+            message="Unable to search playlists. Please try again."
+          />
+        );
+      });
+      return;
+    }
+
     if (prevProps.playlistType !== playlistType) {
       this.setState(
         {
           playlists: playlists.map((pl) => pl.playlist),
           sortBy: initialSort,
           view: initialView,
+          activeSearchQuery: "",
+          searchTerm: "",
         },
         () => {
           this.setSortOption(initialSort);
@@ -126,7 +153,7 @@ export default class UserPlaylists extends React.Component<
       );
       return;
     }
-    if (prevProps.playlists !== playlists) {
+    if (prevProps.playlists !== playlists && !activeSearchQuery) {
       this.setState(
         {
           playlists: playlists.map((pl) => pl.playlist),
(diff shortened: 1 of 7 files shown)


anshg1214 (changes requested):
The functionality looks good. A couple of improvements:
 
1. Pagination needs a fix: While navigating search results, I see two API calls. Clicking page 2 updates `?page=` in the URL, which triggers the normal playlists page load. The search API also runs. Only the search call should run during search.

2. Search doesn’t match the active tab - on Playlists you normally only see playlists you created. On Collaborative you only see ones you collaborate on. Search returns all of them together (including created-for-you). Results should match whichever tab is open.

3. Sorting overrides search order:  Search returns results by best match, then the sort dropdown re-orders them. The closest match does not stay at the top. We need to find a way to improve the experience here.

4. I believe we need to find a better way to improve how the search bar looks like. It looks too crowded.

Image

frontend/js/src/user/playlists/Playlists.tsx
```diff
@@ -261,6 +289,129 @@ export default class UserPlaylists extends React.Component<
...
+    const offset = (page - 1) * PLAYLISTS_PAGE_SIZE;
+    const result = await APIService.searchPlaylistsForUser(
+      query,
+      user.name,
+      PLAYLISTS_PAGE_SIZE,
+      offset
+    );
+
+    const playlistsFromApi = result.playlists ?? [];
+    const playlists = playlistsFromApi.map((pl: any) => pl.playlist);
```
anshg1214: Let's have proper types here.

yateendogra2109:
> The functionality looks good. A couple of improvements:
> 
> 1. Pagination needs a fix: While navigating search results, I see two API calls. Clicking page 2 updates `?page=` in the URL, which triggers the normal playlists page load. The search API also runs. Only the search call should run during search.
> 2. Search doesn’t match the active tab - on Playlists you normally only see playlists you created. On Collaborative you only see ones you collaborate on. Search returns all of them together (including created-for-you). Results should match whichever tab is open.
> 3. Sorting overrides search order:  Search returns results by best match, then the sort dropdown re-orders them. The closest match does not stay at the top. We need to find a way to improve the experience here.
> 4. I believe we need to find a better way to improve how the search bar looks like. It looks too crowded.
> 
> Image

Hello @anshg1214 !
Thanks for the review .

The first two improvements you mentioned were real bugs . I had worked on them and corrected them locally .

### **For the third point** : 
Yes the current behaviour is little distorted as when the search query returns a set of playlists , they are overwritten by the current active sort option automatically , So yes the user is not able to see the best match .

**What I would propose to deal with sort auto applied on search : **

Browse mode (no search) — keep the current sort dropdown (Date created, Date updated, Title, etc.) and persist the user’s preference as we do today.

Search mode (active query) — when user makes a search query . He should have all the  pre existing sort options including a new option "`BEST MATCH"` which will give the raw search playlists from the backend without any sort applied . And it will  be the default. 

Optional re-sort among results — if the user explicitly chooses e.g. Date updated while searching, we allow it, but only with clear UI context (e.g. “Sorted by Date updated among search results”). But currently the sort will only be applied on the 25 playlists on that page . Longer term, paginated search + sort should be handled on the backend so ordering is consistent across pages .

I dont want to completely disable the sort functionality while search is active as it would limit the ability to find the particular playlist.

To fully implement this feature we would require proper backend with proper sort+search+pagination. Currently this sounds heavy work. What would you suggest now ?


### **Also for the 4th point mentioned by you .** 

Yes I have been working on this for weeks So i was used to this layout . But a new user will definately find it little messy .

So we can use two rows rather than one 

Like : 

image


**I will deliver once we finalyze on point 3 and 4**

anshg1214:
@yateendogra2109 Let's have the improvements in this PR itself?

yateendogra2109:
> @yateendogra2109 Let's have the improvements in this PR itself?

@anshg1214 I think For the first two points , I will improve here and re-push . For the 3rd and 4th point I will do in a later PR . Maybe once the GSoC project is completed.

yateendogra2109:
Hello @anshg1214  , I had fixed the first two points mentioned by you 

The only failing test in the CI pipeline is  LinkListens.test.tsx . My PR only changes playlist search in Playlists.tsx, APIService.ts, and backend playlist search. No overlap with Link Listens. So this one is un-related to this pr . You can go on with review .

anshg1214 (changes requested):
I can see you’ve added a lot of special handling to incorporate search. A simpler way would be to leverage the React Query Loader we’ve set-up. As the user types the search query, you can update the url to `/user/<username>/playlists/?limit=50&offset=0&search=query`

Once the search query get’s added to the URL, the React Router loaders will kick in. They’ll call the Loader Props endpoint (the one we’re using to load the playlists).  You can add a handling on that route saying that if you have the search parameter, you can call  `db_playlist.search_playlist`.

With this approach, you don’t have to maintain states, and all the special handling you’ve added.

What do you think?

yateendogra2109:
@anshg1214 
Thanks for the feedback 

I had thought about this and 
Your approach seems simple and fits how playlists are loaded already. So instead of separate search state, API calls, and pagination logic . I will use ( URL → RouteLoader → POST handler)

**I’ll refactor this PR accordingly:**

**Data loading (URL + loader)**
I’ll move playlist search onto the same path as normal browsing: search will be a URL parameter (e.g. ?search=...&page=1&type=...), and the existing React Router loader / POST handler for /user/<username>/playlists/ will fetch the data. 

**Search + sort**
To solve the sort vs search issue, when a search is active I’ll pass a sort parameter through to `search_playlists_for_user` so ordering happens on the server before pagination. The default during search will be relevance (“Best match”); if the user picks another sort (e.g. date updated, title), that will update the URL and refetch via the loader so sort and pagination stay consistent across pages.  Currently we are using  client-side re-sort search results after fetch. So we will get rid of that and this will properly solve the issue .

**Loading & empty states (Part 2)**
I’ll use useNavigation for loading indicators instead of custom search loading state, and show distinct empty messages based on URL + loader data (e.g. no matches for a search query vs no playlists yet when not searching).


I’ll start working on this and need some timme to properly implement search, pagination, tab switching, sort during search, and loading/empty states. I’ll push updates on this PR once the refactor is done.

Let me know  should we finally sit with this plan .

anshg1214:
Sounds good

yateendogra2109:
HI @anshg1214 
I had implemented the changes accordingly : 
Search, sort, page, and tab are now driven by URL params and the existing React Router loader (POST /user/<name>/playlists/), instead of a separate client-side API call. Removed extra states.
Sort during search is handled on the backend with pagination -> sort is passed in` search_playlist_for_user().` 
Tab switches reset search/sort and page.

**Loader / useNavigation and dealing with empty states will be implemented in part 2 PR , keeping this as base pr , once we finalize and merge this pr .** 

Also regarding the UI part , as you mentioned it feels congested . So there will be many UI changes in this page in follow up PR , during tags integration. 
So I will fix the search bar in correct position at that time.

yateendogra2109:
Also this failing test is not related to this pr .

frontend/js/src/utils/APIService.ts
```diff
@@ -2293,14 +2293,21 @@ export default class APIService {
     searchQuery: string,
     musicbrainzID: string,
     count: number = 25,
-    offset: number = 0
+    offset: number = 0,
+    type?: "owned" | "collaborative"
```
anshg1214, line 2297: Do we still need these changes?

listenbrainz/webserver/views/user.py
```diff
@@ -158,11 +166,32 @@ def playlists(user_name: str):
     }
 
     include_private = current_user.is_authenticated and current_user.id == user.id
-
-    playlists = []
     offset = (page - 1) * DEFAULT_NUMBER_OF_PLAYLISTS_PER_CALL
 
-    if type == "collaborative":
+    if search_query and len(search_query) >= 3:
+        if sort not in SEARCH_PLAYLIST_SORTS:
```
anshg1214, line 172: In case of incorrect payload, we should throw an error explicitly stating that the sort should be one of the supported types

anshg1214 (approved):
Looks good. We can improve how we’re handling pagination and search UX.

Approving and merging for now. Will revisit when we have all the improvements (which you’ve already started working on in your follow up PRs) merged.

listenbrainz/webserver/views/user.py
yateendogra2109: Sure ,  instead of silently falling back to "relevance", an error should be thrown . I will fix this

frontend/js/src/utils/APIService.ts
yateendogra2109: We dont need type here since APIService.searchPlaylistsForUser is used by global search which doesnt need type , however offset is needed . I will fix this

PR #3717, opened 2026-06-02
Support RabbitMQ service cluster config

amCap1712:
Use a rabbitmq cluster for listenbrainz that is separate from MusicBrainz to avoid overloading it and also to avoid single point of failures.

### DIFF
--- a/listenbrainz/rabbitmq.py
+++ b/listenbrainz/rabbitmq.py
@@ -1,0 +1,37 @@
+from urllib.parse import quote
+
+from kombu import Connection
+
+from listenbrainz.utils import get_fallback_connection_name
+
+
+def _get_config_value(config, key, default=None):
+    return config.get(key, default)
+
+
+def _rabbitmq_url(host, port, config):
+    username = quote(str(_get_config_value(config, "RABBITMQ_USERNAME", "")), safe="")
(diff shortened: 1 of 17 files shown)


listenbrainz/rabbitmq.py
```diff
@@ -0,0 +1,37 @@
...
+
+def _get_config_value(config, key, default=None):
+    return config.get(key, default)
+
+
+def _rabbitmq_url(host, port, config):
+    username = quote(str(_get_config_value(config, "RABBITMQ_USERNAME", "")), safe="")
+    password = quote(str(_get_config_value(config, "RABBITMQ_PASSWORD", "")), safe="")
+    vhost = quote(str(_get_config_value(config, "RABBITMQ_VHOST", "/")), safe="")
+    return f"amqp://{username}:{password}@{host}:{port}/{vhost}"
```
MonkeyDo, line 16: ~~Seeing as the default value for vhost is "/", does this connection string end up with "//"?~~


Never mind, I see the examples in kombu do show 'amqp://guest:guest@localhost:5672//'
https://docs.celeryq.dev/projects/kombu/en/stable/reference/kombu.connection.html

MonkeyDo (approved):
Looks good!

PR #3719, opened 2026-06-02
RabbitMQ Updates

amCap1712:
1. Make RabbitMQ exchanges and queues durable
2. Move incoming listens to a quorum queue
3. Ensure incoming queue is declared the web container to avoid lost listens


MonkeyDo (approved):
Looks good without being able to test it

PR #3720, opened 2026-06-02
Use virtual- services to connect to main cluster db

amCap1712:
lips and some other nodes need to connect to db on 10.10.10.x so use the virtual service for now.


MonkeyDo (approved):
Small changes, look good

PR #3721, opened 2026-06-02
Declare RMQ queues on publish and enable publisher confirms

amCap1712:
Listens were published into the `incoming` fanout exchange relying on the queue already existing. Only the main web app declared the incoming queue at startup (via the declare_incoming_queue flag); api_compat and the spotify/lastfm importers create their app without it, so they could silently drop listens into an unrouted exchange. The timescale writer also crashed publishing unique listens with "NOT_FOUND - no exchange 'unique'" because it never declared the unique exchange.

Move queue declaration into the publish path instead of per-entry-point startup. publish_data_to_queue now declares the incoming queue (which also declares its exchange and binding) so every producer guarantees a routable queue before publishing. kombu's maybe_declare caches this per connection, so it's a one-time declare per connection, not per publish.

Enable confirm_publish on the producer connection and publish incoming listens with mandatory=True so unroutable/lost listens raise loudly (surfaced as APIServiceUnavailable) instead of vanishing. Declare the unique exchange/queue in the timescale writer to fix the NOT_FOUND error. Remove the now-unused declare_incoming_queue startup flag.

### DIFF
--- a/listenbrainz/timescale_writer/timescale_writer.py
+++ b/listenbrainz/timescale_writer/timescale_writer.py
@@ -181,7 +181,8 @@ class TimescaleWriterSubscriber(ConsumerProducerMixin):
             exchange=self.unique_exchange,
             routing_key="",
             body=orjson.dumps([listen.to_json() for listen in unique]).decode("utf-8"),
-            delivery_mode=PERSISTENT_DELIVERY_MODE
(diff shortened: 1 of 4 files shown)


MonkeyDo (approved):
That all makes sense

PR #3722, opened 2026-06-03
Update stats calculation status mechanism

MonkeyDo:
# Problem

Our Grafana alert for stats age has been broken since March.
The reason is that it is based on Rob's listening history, and subsequent stats calculation. 
We were aware of the brittleness of it before his passing: Rob not listening to music for a week would result in an undue alert even if stats calculated fine for other users.
Since his passing though the alert has been useless. We had planned to replace the mechanism with this much more reasonable strategy.

# Solution

Get Spark to generate a message when it successfully completes a stats calculation, which writes a completion timestamp in the database that can then be read by our status API.
As before that data is cached, but I have a dded a cache invalidation strategy to avoid stale info

# AI usage

* [ ] I did not use any AI

Used Codex to generate most of this PR, based on a strategy first discussed and decided with lucifer

#### If you did use AI:

# Action

- Run the DB migration
- No update necessary for stats age grafana alert, but we could add more precise monitoring if needed


MonkeyDo:
@anshg1214 do you have a moment to review this PR? 
Wanna get those telegram alerts under control so we can actually keep an eye on services instead of ignoring incorrect alerts

anshg1214 (approved):
LG

PR #3725, opened 2026-06-04
Add schema and builder for events cache

shirsakm:
# Problem

:beginner: This covers most of Week 1 and 2 of my GSoC proposal.

The goal was to build the events cache, this includes creating the schema, and writing the builder that will pull the data from MusicBrainz, and populate the ListenBrainz tables.

# Solution

The schema changes are in `admin/timescale/` which includes updating the `create_tables.sql` script as well as adding an update patch script. The schema is almost exactly the same as in my proposal, with the changes I remember making being as follows:
- Changing `area_mbid` from `UUID[]` to `UUID` as we are not doing the unpacking anymore.
- Adding `rating` and `rating_count` as I realized that they would be pretty handy for a feature to sort events by rating. I don't know if it definitely _will_ be used, but it seems useful to me.

The main changes are in `mbid_mapping/mapping/mb_event_metadata_cache.py` which handles both `mapping.mb_event_cache` and `mapping.mb_event_artist_cache`. The `MusicBrainzEventMetadataCache` is partially based on [the artist cache](https://github.com/metabrainz/listenbrainz-server/blob/master/mbid_mapping/mapping/mb_artist_metadata_cache.py).

* [X] I have run the code and manually tested the changes

The outputs of my Timescale database are available [as a gist](https://gist.github.com/shirsakm/9803a8cdaeb1d300dda082d8d1421829). Things seem to be working fine.

# AI usage

* [ ] I did not use any AI
* [X] I have used AI in this PR (add more details below)

#### If you did use AI:
* [X] I understand all the changes made in this PR

I used AI to find the event MBIDs to use for testing in the minimal dataset. Mostly because it was mind numbing to go through the awful formatting of the entries. Good thing AI's don't have eyes.

# Action

1. Specifically for Monkey, since I recall you mentioning checking the performer / artist types like "main performer", "support act" et al. this would be a great time to check the existing types in the whole MB database, to see if any others would make sense to include.

2. If any of the tables [here](https://github.com/metabrainz/musicbrainz-server/blob/90ba9db3919fea57621b99e146f3b844011b57e1/admin/sql/CreateTables.sql#L1514-L1656) seem to contain things that should be included, please let me know.

3. Please double check if the schema is fine, rather, if anything that is currently stored in the JSON blob should be stored directly as a column instead. None of the tables I have not included should be too difficult to add to the JSON blob later, if we need that data, so it seems fine to me.

### DIFF
--- a/admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
+++ b/admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
@@ -1,0 +1,61 @@
+BEGIN;
+
+CREATE TABLE mapping.mb_event_cache (
+    dirty                BOOLEAN DEFAULT FALSE,
+    last_updated         TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+
+    event_mbid           UUID NOT NULL,
+    event_id             INTEGER NOT NULL,
+    event_name           TEXT NOT NULL,
+
+    begin_date_year      SMALLINT,
+    begin_date_month     SMALLINT,
+    begin_date_day       SMALLINT,
+    end_date_year        SMALLINT,
+    end_date_month       SMALLINT,
+    end_date_day         SMALLINT,
+    event_time           TIMESTAMPTZ,
+    cancelled            BOOLEAN NOT NULL DEFAULT FALSE,
+    ended                BOOLEAN NOT NULL DEFAULT FALSE,
+
+    event_type_gid       UUID,
+    event_art_presence   TEXT NOT NULL DEFAULT 'absent',
+
+    place_mbid           UUID,
+    place_name           TEXT,
+    area_mbid            UUID,
+
+    event_data           JSONB NOT NULL DEFAULT '{}'::jsonb
+);
+
+ALTER TABLE mapping.mb_event_cache
+    ADD CONSTRAINT mb_event_cache_pkey PRIMARY KEY (event_mbid);
+
+CREATE UNIQUE INDEX mb_event_cache_idx_event_id ON mapping.mb_event_cache (event_id);
+CREATE INDEX mb_event_cache_idx_date ON mapping.mb_event_cache (begin_date_year, begin_date_month, begin_date_day);
+CREATE INDEX mb_event_cache_idx_dirty ON mapping.mb_event_cache (dirty);
+
+CREATE TABLE mapping.mb_event_artist_cache (
+    event_mbid           UUID NOT NULL,
+    event_id             INTEGER NOT NULL,
+    artist_mbid          UUID NOT NULL,
+    artist_id            INTEGER NOT NULL,
+
+    link_id              INTEGER NOT NULL,
+    link_type_gid        UUID NOT NULL,  
+    link_type_name       TEXT NOT NULL,
+
+    relationship_data    JSONB NOT NULL DEFAULT '{}'::jsonb,
+    last_updated         TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+ALTER TABLE mapping.mb_event_artist_cache
+    ADD CONSTRAINT mb_event_artist_cache_pkey
+    PRIMARY KEY (event_id, artist_id, link_id);
+
+CREATE INDEX mb_event_artist_cache_idx_artist_id_event_id ON mapping.mb_event_artist_cache (artist_id, event_id);
+CREATE INDEX mb_event_artist_cache_idx_artist_mbid ON mapping.mb_event_artist_cache (artist_mbid);
+CREATE INDEX mb_event_artist_cache_idx_event_id ON mapping.mb_event_artist_cache (event_id);
+CREATE INDEX mb_event_artist_cache_idx_link_type_gid ON mapping.mb_event_artist_cache (link_type_gid);
+
+COMMIT;
--- a/admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
+++ b/admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
@@ -28,7 +28,7 @@ CREATE TABLE mapping.mb_event_cache (
     rating               SMALLINT,
     rating_count         INTEGER,
 
-    event_data           JSONB NOT NULL DEFAULT '{}'::jsonb
+    event_data           JSONB NOT NULL
 );
 
 ALTER TABLE mapping.mb_event_cache
@@ -48,7 +48,7 @@ CREATE TABLE mapping.mb_event_artist_cache (
     link_type_gid        UUID NOT NULL,  
     link_type_name       TEXT NOT NULL,
 
-    relationship_data    JSONB NOT NULL DEFAULT '{}'::jsonb,
+    relationship_data    JSONB NOT NULL,
     last_updated         TIMESTAMPTZ NOT NULL DEFAULT NOW()
 );
 
--- a/admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
+++ b/admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
@@ -45,6 +45,7 @@ CREATE TABLE mapping.mb_event_artist_cache (
     artist_id            INTEGER NOT NULL,
 
     link_id              INTEGER NOT NULL,
+    link_order           INTEGER NOT NULL DEFAULT 0,
     link_type_gid        UUID NOT NULL,  
     link_type_name       TEXT NOT NULL,
 
@@ -54,7 +55,7 @@ CREATE TABLE mapping.mb_event_artist_cache (
 
 ALTER TABLE mapping.mb_event_artist_cache
     ADD CONSTRAINT mb_event_artist_cache_pkey
-    PRIMARY KEY (event_id, artist_id, link_id);
+    PRIMARY KEY (event_id, artist_id, link_id, link_order);
 
 CREATE INDEX mb_event_artist_cache_idx_artist_id_event_id ON mapping.mb_event_artist_cache (artist_id, event_id);
 CREATE INDEX mb_event_artist_cache_idx_artist_mbid ON mapping.mb_event_artist_cache (artist_mbid);
--- a/admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
+++ b/admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
@@ -25,6 +25,9 @@ CREATE TABLE mapping.mb_event_cache (
     place_name           TEXT,
     area_mbid            UUID,
 
+    rating               SMALLINT,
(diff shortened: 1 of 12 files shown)


mbid_mapping/mapping/mb_cache_base.py
MonkeyDo, line 1: Are these changes for debugging? Should they be commited?

MonkeyDo:
Paging @amCap1712 for reviewing this, my knowledge of the cache and these giant queries is inadequate fro a proper review

mbid_mapping/mapping/mb_event_metadata_cache.py
```diff
@@ -0,0 +1,722 @@
...
+)
+ARTIST_EVENT_LINK_GIDS_SQL = ", ".join([f"'{x}'" for x in ARTIST_EVENT_LINK_GIDS])
+
+EVENT_URL_LINK_GIDS = (
+    "c26808b0-4e67-31a7-a587-913720dfb3f3",  # official homepage
+    "68f5fcaa-b58c-3bfe-9b7c-75c2b56e839a",  # social network
+    "125afc57-4d33-3b63-ab41-848a3a18d3a6",  # songkick
+    "b022d060-e6a8-340f-8c73-6b21b1d090b9",  # wikidata
+    "08a982f7-d754-39b2-8315-d7cae474c641",  # wikipedia
+    "fea46163-dc45-3af9-917e-1798f325d21a",  # youtube
```
MonkeyDo: And the more generic "Video channel" perhaps?
https://beta.musicbrainz.org/relationship/1f3df2eb-3d0b-44f1-9599-1309c692bc7c

MonkeyDo:
I think we should add more URL link types, basicalyl all the ones that have ID 803 ("other databases") as a parent would be useful to show on the page, as well as "ticketing" and "schedule"

For reference, a list of all the link types for events:
image

mbid_mapping/mapping/mb_event_metadata_cache.py
```diff
@@ -0,0 +1,722 @@
...
+)
+EVENT_URL_LINK_GIDS_SQL = ", ".join([f"'{x}'" for x in EVENT_URL_LINK_GIDS])
+
+EVENT_PLACE_LINK_GID = "e2c6f697-07dc-38b1-be0b-83d740165532"
+
+# falls back to area if no place
+AREA_EVENT_LINK_GID = "542f8484-8bc7-3ce5-a022-747850b2b928"
+
```
MonkeyDo, line 46: Probably also useful to support multi-part events, for example for events with multiple days where each day is a separate MB event entry.

GID = 65742183-b25c-469e-b094-ff6739e6699c
name = parts
This indicates that an event is made up of multiple parts (e.g. a festival happening on multiple venues over the course of a few days).

mbid_mapping/mapping/mb_event_metadata_cache.py
```diff
@@ -0,0 +1,722 @@
...
+    create_metadata_cache,
+    incremental_update_metadata_cache,
+    MusicBrainzEntityMetadataCache,
+)
+from mapping.bulk_table import BulkInsertTable
+from mapping.utils import insert_rows, log
+
+MB_EVENT_METADATA_CACHE_TIMESTAMP_KEY = "mb_event_metadata_cache_last_update_timestamp"
+
+ARTIST_EVENT_LINK_GIDS = (
```
MonkeyDo: I wonder instead which ones we should explicitly exclude.
It's nice to surface a lot of useful information if we find space for it on the page. Some links should have more prominence obviously, but there could be a place lower on the page with additional relationships.

Some relationships are only relevant in the context of classical music for example ("conductor", "orchestra"), and while they won't make sense for a rock festival they would be crucial to an opera.

Reading the list of link types, these are the ones I see as not-so-useful probably:
- rescheduled as
- artwork
- design
- graphic design
- illustration
- jury member

Image

mbid_mapping/mapping/mb_cache_base.py
shirsakm: It's a bugfix.

[This is the function](https://docs.python.org/3/library/traceback.html#traceback.format_exc) I believe it's meant to be calling. It throws a `NameError` otherwise, since it wasn't imported before. I mean, that's the function that's called elsewhere in the codebase:

```
> rg "traceback.format_exc()" .
./mbid_mapping/mapping/mb_cache_base.py
232:                log(traceback.format_exc())

./listenbrainz/dumps/check.py
92:        msg = "Cannot fetch full dump age: %s\n\n%s" % (str(err), traceback.format_exc())
103:        msg = "Cannot fetch incremental dump age: %s\n\n%s" % (str(err), traceback.format_exc())
114:        msg = "Cannot fetch feedback dump age: %s\n\n%s" % (str(err), traceback.format_exc())

./listenbrainz/metadata_cache/crawler.py
64:                    self.app.logger.info(traceback.format_exc())

./listenbrainz/metadata_cache/spotify/handler.py
133:            self.app.logger.info(traceback.format_exc())

./listenbrainz/metadata_cache/apple/handler.py
158:            self.app.logger.info(traceback.format_exc())

./listenbrainz/mbid_mapping_writer/job_queue.py
364:            self.app.logger.info(traceback.format_exc())
```

shirsakm:
> I wonder instead which ones we should explicitly exclude.

Ah, you are suggesting using a blocklist instead of an allowlist approach?

> Reading the list of link types, these are the ones I see as not-so-useful probably:
> -    rescheduled as
> -    artwork
> -    design
> -    graphic design
> -    illustration
> -    jury member

There's also `teacher`, `engineer`, `design` maybe? It's quite a bit both ways.

image

This one (event to URL link types), I don't really see anything objectionable and it does seem it would be better as a blocklist.

In that case, I think I should update both to have a blocklist approach. That seems about right to me. Please do let me know if that doesn't seem right.

mbid_mapping/mapping/mb_event_metadata_cache.py
shirsakm: As for this one specifically, we can just include it in the `event_data` field which is `JSONB`.

I don't see it affecting anything specifically in my proposal, but that wouldn't be too great an approach if we plan on ever adding anything that might need to query a lot of event to event relationships.

Do you see that ever being an issue? If not, I will just implement it that way. The alternative is creating yet another intersection table to keep queries efficient.

mbid_mapping/mapping/mb_event_metadata_cache.py
```diff
@@ -0,0 +1,884 @@
...
+            event_data["area_name"] = row["area_name"]
+
+        if row["event_links"]:
+            filtered = {}
+            for name, url in row["event_links"]:
+                if name is None or url is None:
+                    continue
+                filtered[name] = url
+            if filtered:
+                event_data["rels"] = filtered
```
anshg1214, line 251: event_data.rels loses links - keys a dict by link-type name. create_json_data does filtered[name] = url, so any event with multiple URLs of the same type keeps only the last. The SQL aggregates them all correctly; the loss is purely in the Python fold.


Verified against the MB DB: 6,879 events lose 15,403 URLs total. e.g. 12d4fd6d-fecf-4049-8323-6e8c5e4718d5 (Peace and Love 2007, Day 5) has 36 setlistfm links collapsed to 1. 

Fix: make rels a list of {type, url}, or {type: [urls]}.

admin/timescale/updates/2026-05-28-add-mb-events-cache.sql
```diff
@@ -0,0 +1,64 @@
...
+    link_type_gid        UUID NOT NULL,  
+    link_type_name       TEXT NOT NULL,
+
+    relationship_data    JSONB NOT NULL,
+    last_updated         TIMESTAMPTZ NOT NULL DEFAULT NOW()
+);
+
+ALTER TABLE mapping.mb_event_artist_cache
+    ADD CONSTRAINT mb_event_artist_cache_pkey
+    PRIMARY KEY (event_id, artist_id, link_id);
```
anshg1214: Artist-edge PK assumes `(event_id, artist_id, link_id)` is unique; MB doesn't guarantee that. MB's constraint is `UNIQUE (entity0, entity1, link, link_order)` the PK drops link_order, so two ordered relationships on a shared link would collide and fail the unique index (breaking the rebuild) or the incremental insert. 

Checked the DB: 0 rows today, so it's a latent landmine, not a live bug.

anshg1214 (changes requested):
Are we planning to add `l_event_recording`, `l_event_release`, `l_event_release_group` links?

shirsakm:
> Are we planning to add `l_event_recording`, `l_event_release`, `l_event_release_group` links?

I didn't really imagine they would be needed, no. But I can just add it to `event_data` if the goal is to just show the songs / albums in a "Related songs" section or something. I had thought the `setlist` field would have the data on what songs were played, and since it's supposed to contain the relevant UUID we could just use that.

But if we want to include that, please do let me know.

mbid_mapping/mapping/mb_event_metadata_cache.py
```diff
@@ -0,0 +1,894 @@
+import uuid
+from datetime import datetime, timezone
+from typing import List, Set
+
+import psycopg2
+import psycopg2.extras
+import ujson
+
+import config
```
MonkeyDo, line 9: This is throwing an [error in CI](https://github.com/metabrainz/listenbrainz-server/actions/runs/29003449067/job/86069273652?pr=3725#step:5:2728).
Did you have a separate uncommitted file for the config? If so, we should find the right place(s) for the config values to go.

shirsakm: Yep, that is what the `pytest.ini` file change was meant to fix, but apparently not. It's working locally, I am looking into why it doesn't wanna work in the workflow.

MonkeyDo: Wanna try this? `from listenbrainz import config` as seen [in other places](https://github.com/search?q=repo%3Ametabrainz%2Flistenbrainz-server+%22from+listenbrainz+import+config%22&type=code)
Doesn't give a pylint warning for me in my IDE, where the simple `import config` does

MonkeyDo: image

VS

image

shirsakm: The issue seems to have been that the config file inside the `mbid_mapping` folder was not being copied over in the workflow file, because that folder isn't really tested. It works now.

Edit: I can't believe I pressed enter on this comment and the workflow failed, though it's a different error now, so ...progress?

shirsakm: Yep, the new failing test doesn't seem related to any changes I made.

PR #3727, opened 2026-06-08
Fix duration_ms fallback bug in timescale_writer

srajang1805:
LB-2004
The code incorrectly checks duration instead of duration_ms when using the fallback field from additional_info, preventing duration_ms values from being stored. This change updates the condition to use duration_ms correctly.


anshg1214 (approved):
LG!

srajang1805:
@anshg1214 would appreciate a merge if it is lgtm. Thanks

srajang1805:
@anshg1214 Thanks 😊

PR #3728, opened 2026-06-09
Serve static assets from Staticbrainz

MonkeyDo:
# Problem

We have been seeing some issues with LB-Radio's long-running queries that are filling up the uWSGI queue, and causing the website to become unavailable for hours at a time.
In an effort to reduce the amount of queries served through uwsgi, this PR implements the same mechanism MusicBrainz uses: copy the built assets to StaticBrainz at the end of the automated image building, then serve those same assets directly from staticbrainz/gateways to reduce traffic to the web containers.

This will compress and copy file sin the formats

# AI usage

#### If you did use AI:

I used Codex to help with planning based on the musicbrainz rsync setup and some staticbrainz documentation files, and to implement most of the code changes.

# Action

- There is a companion PR in docker-server-configs that adds new SSH keys (private and public) and sets them up in the SSHD container.
- The secret key will need to be [added to this repo's secrets](https://github.com/metabrainz/listenbrainz-server/settings/secrets/actions/new) as `STATICBRAINZ_SSH_PRIVATE_KEY`, to be written to the container after successfully pushing the build docker image.
- Similarly the `STATICBRAINZ_SERVERS` needs to be added to github actions secrets. Those could potentially be added as org-wide secrets
- I am not able to test this locally for obvious reasons, so this should first be tested with a a test image and the push-dev-image workflow (both the dev and prod images workflows are set up in this pr)


docker/prod/nginx/nginx.conf
```diff
@@ -2,6 +2,20 @@ server {
     listen 80;
     server_name proxy.listenbrainz.org;
     root /usr/share/nginx/html;
+    location /static/dist/ {
```
amCap1712, line 5: This is the dockerfile used by Last.fm api compat nginx image. These changes would need to go in the gateways config instead. I think that's where they are for MB too.

admin/rsync-staticbrainz-files.sh
```diff
@@ -0,0 +1,105 @@
...
+        -o -name '*.js' \
+        -o -name '*.json' \
+        -o -name '*.map' \
+        -o -name '*.svg' \
+        -o -name '*.txt' \
+        -o -name '*.webmanifest' \) \
+        -exec gzip -9 --keep --force {} \;
+fi
+
+for server in $STATICBRAINZ_SERVERS; do
```
amCap1712, line 90: Does these remove files with old hashes from old releases? That would break production until the new image is updated on containers and ability to rollback releases. If not, then there needs to be a way to periodically clean these files on staticbrainz. Especially because the JS files for LB are much larger than MB. Again probably makes sense to check what MB does.

docker/prod/nginx/nginx.conf
MonkeyDo: Ah, that makes sense. I thought it might have been use for some part of  local dev somehow, thanks for the clarification

admin/rsync-staticbrainz-files.sh
MonkeyDo: MB does do --delete which removes the files.
However I'm not sure how that fits with their workflow or how they prevent early removal of assets before deploying the image, and I think they build assets for prod+beta+test always, possibly to prevent the removal of those assets when you build for one env.
Needs some more research.

MonkeyDo:
Current status:

Staticbrainz has been set up on the gateways to receive listenbrainz assets, in separate folders per environment (/prod/, /beta/ & /test/). SSH keys are set up for rsync.

The last hurdle is that the gateways ssh ports are not exposed publicly, only on private network which github actions do not have access to.
This means we will have to run the rsync script from inside a container deployed on our network.
So likely running rsync as the first thing when starting the LB containers, before the webapp starts. It introduces a delay when first deploying a new image.

PR #3729, opened 2026-06-11
Separate test environment from beta/prod

amCap1712:
Make database user, password and name configurable. 
Add configurable registration URL usage for account creation links. 
Make Redis namespace configurable and add CouchDB database prefix support.

### DIFF
--- a/consul_config.py.ctmpl
+++ b/consul_config.py.ctmpl
@@ -12,6 +12,10 @@
     json.loads('''{{template "KEY" .}}''')
 {{- end -}}
 
+{{- define "KEY_OPTIONAL" -}}
+  {{ keyOrDefault (printf "docker-server-configs/LB/config.%s.json/%s" (env "DEPLOY_ENV") .) "" }}
+{{- end -}}
+
 # Put everything after the defines otherwise the rendered config might eat up some newlines
 import os
 import json
@@ -30,7 +34,7 @@ REMEMBER_COOKIE_HTTPONLY = True
 
 {{if service "virtual.virtual-pgbouncer-master"}}
 {{with index (service "virtual.virtual-pgbouncer-master") 0}}
-SQLALCHEMY_DATABASE_URI = "postgresql://listenbrainz:listenbrainz@{{.Address}}:{{.Port}}/listenbrainz"
+SQLALCHEMY_DATABASE_URI = "postgresql://{{template "KEY" "listenbrainz_db_user"}}:{{template "KEY" "listenbrainz_db_password"}}@{{.Address}}:{{.Port}}/{{template "KEY" "listenbrainz_db"}}"
 POSTGRES_ADMIN_URI="postgresql://postgres@{{.Address}}:{{.Port}}/postgres"
 SQLALCHEMY_METABRAINZ_URI = "postgresql://metabrainz:metabrainz@{{.Address}}:{{.Port}}/metabrainz_db"
 {{end}}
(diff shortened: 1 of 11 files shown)


consul_config.py.ctmpl
```diff
@@ -12,6 +12,10 @@
     json.loads('''{{template "KEY" .}}''')
 {{- end -}}
 
+{{- define "KEY_OPTIONAL" -}}
```
amCap1712, line 15: This is needed because consul template treats empty values as absent.

listenbrainz/db/couchdb.py
```diff
@@ -70,7 +87,12 @@ def list_databases(prefix: str) -> list[str]:
     response.raise_for_status()
     all_databases = response.json()
 
-    databases = [database for database in all_databases if database.startswith(prefix)]
+    prefixed_prefix = _add_database_prefix(prefix)
```
MonkeyDo: "prefixed_prefix" seems a little goofy as a name.
I would suggest renaming the `prefix` arg to `stat_prefix` or the like.
Then maybe sometihng like `env_prefixed_stat_prefix` or  `database_prefixed_stat_prefix` or something like that?
It's not clear by any means, but it's clearer than prefixed_prefix I think.

consul_config.py.ctmpl
```diff
@@ -30,7 +34,7 @@ REMEMBER_COOKIE_HTTPONLY = True
 
 {{if service "virtual.virtual-pgbouncer-master"}}
 {{with index (service "virtual.virtual-pgbouncer-master") 0}}
-SQLALCHEMY_DATABASE_URI = "postgresql://listenbrainz:listenbrainz@{{.Address}}:{{.Port}}/listenbrainz"
+SQLALCHEMY_DATABASE_URI = "postgresql://{{template "KEY" "listenbrainz_db_user"}}:{{template "KEY" "listenbrainz_db_password"}}@{{.Address}}:{{.Port}}/{{template "KEY" "listenbrainz_db"}}"
```
MonkeyDo: Wondering if `listenbrainz_db_name` and `timescale_db_name` would be clearer, just in case it gets more confusing in the future

frontend/js/src/utils/utils.tsx
```diff
@@ -918,6 +920,7 @@ const getPageProps = async (): Promise<{
     globalAppContext = {
       APIService: apiService,
       websocketsUrl: websockets_url,
+      registrationUrl: registration_url,
```
MonkeyDo, line 923: It seems like we could just set the computed string returned by `getRegistrationUrl(registrationUrl)` directly in the global context, instead of making the other pages call a utility function that doesn't really do anything dynamic.

```suggestion
      registrationUrl: getRegistrationUrl(registration_url),
```

MonkeyDo (commented):
Good rpeparations. Some comments for clarifying a couple of names as I see a potenatial for confusion in the future.

MonkeyDo:
Also I guess the tests need to be amended to add the prefix arg to the couchdb init function
https://github.com/metabrainz/listenbrainz-server/actions/runs/27340467642/job/80775689757?pr=3729#step:5:2831

frontend/js/src/utils/utils.tsx
amCap1712: `getRegistrationUrl` reads current window url so doing this would fix the `next` page that the user is sent back to after registration as the first page and it wouldn't change dynamically with user navigation,

MonkeyDo (approved):
Ready!

PR #3731, opened 2026-06-13
GSoC-26 Feature-2 In Playlist Track sorting

yateendogra2109:
# PR Description

[LB-1374](https://tickets.metabrainz.org/browse/LB-1374)

Added a Sort By dropdown on the individual playlist page so users can reorder how tracks are displayed without changing the playlist's saved order.

Sorting is frontend-only. Drag-to-reorder is available in Default Mode only; all other sort modes are temporary views.

### UI

Added a Sort By `<select>` in the playlist tracks header, next to **Play all**.

**Sort options:**

- Default — original playlist order (same as stored on the server)
- Recently Added — newest `added_at` first (from JSPF track extension)
- Title (A–Z) — case-insensitive
- Artist (A–Z) — case-insensitive
- Shuffle — random permutation (same approach used on the main playlists page)

A `useEffect` re-applies the active sort when tracks change, so the view stays consistent.

When `dragEnabled` is disabled, the drag handle is dimmed and the cursor becomes `not-allowed`.

### Manual testing

For local verification, I used one cached test playlist in my dev environment instead of creating new playlists each time.

That playlist had multiple tracks with distinct titles, artists, and `added_at` values in the JSPF track extension.

### Tests (`Playlist.test.tsx`)

Added a new test, **"sorts playlist tracks in the UI"**, which verifies Default, Title, Artist, and Recently Added ordering using mocked JSPF tracks.

Screenshot 2026-06-13 054114

### DIFF
--- a/frontend/js/src/playlists/Playlist.tsx
+++ b/frontend/js/src/playlists/Playlist.tsx
@@ -145,75 +145,72 @@ export default function PlaylistPage() {
       const base = tracks ?? [];
       setSortKey(option);
 
-      if (option === "shuffle") {
-        // Keep the same shuffle logic as the main playlists page
-        setDisplayedTracks([...base].sort(() => Math.random() - 0.5));
-        return;
+      switch (option) {
+        case "shuffle": {
+          setDisplayedTracks([...base].sort(() => Math.random() - 0.5));
+          return;
+        }
+        case "default": {
+          setDisplayedTracks(base);
+          return;
+        }
+        case "title": {
+          setDisplayedTracks(
+            [...base].sort((a, b) => {
+              const titleA = a.title ?? "";
+              const titleB = b.title ?? "";
+              const cmp = titleA
+                .toLowerCase()
+                .localeCompare(titleB.toLowerCase());
+              return cmp !== 0 ? cmp : getTieBreaker(a, b);
+            })
+          );
+          return;
+        }
+        case "artist": {
+          setDisplayedTracks(
+            [...base].sort((a, b) => {
+              const artistA = a.creator ?? "";
+              const artistB = b.creator ?? "";
+              const cmp = artistA
+                .toLowerCase()
+                .localeCompare(artistB.toLowerCase());
+              return cmp !== 0 ? cmp : getTieBreaker(a, b);
+            })
+          );
+          return;
+        }
+        case "recently_added": {
+          setDisplayedTracks(
+            [...base].sort((a, b) => {
+              const aTs = Date.parse(getTrackExtension(a)?.added_at || "");
+              const bTs = Date.parse(getTrackExtension(b)?.added_at || "");
+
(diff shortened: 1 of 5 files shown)


frontend/js/src/playlists/Playlist.tsx
```diff
@@ -117,11 +125,97 @@ export default function PlaylistPage() {
...
+  }, []);
+
+  const setTrackSortOption = React.useCallback(
+    (option: TrackSortKey) => {
+      const base = tracks ?? [];
+      setSortKey(option);
+
+      switch (option) {
+        case "shuffle": {
+          setDisplayedTracks([...base].sort(() => Math.random() - 0.5));
```
anshg1214: Let’s use Fisher-Yates algorithm as it would be more robust.

frontend/js/src/playlists/Playlist.tsx
```diff
@@ -548,27 +643,49 @@ export default function PlaylistPage() {
...
-          </h3>
+          <div className="playlist-tracks-header">
+            <h3 className="header-with-line">Tracks</h3>
+            <div className="playlist-tracks-header-controls">
+              <div className="playlist-sort-controls">
+                <label htmlFor="playlistTrackSort" className="text-muted">
+                  Sort By:
+                </label>
+                <select
+                  id="playlistTrackSort"
```
anshg1214, line 681: Let’s add aria-label for screen readers.

anshg1214 (changes requested):
Looks good overall.

frontend/js/src/playlists/Playlist.tsx
yateendogra2109: Yes, I thought of using this , but i went with the approach already used in main playlists page for shuffling playlists.
So , let's go with fisher-yates here  .

yateendogra2109:
Hi @anshg1214 
Please have a look!

anshg1214 (changes requested):
A few observations:

"Play all" always plays original order, never the sorted/shuffled view .

`displayedTracks` doesn't need to be a separate `useState`. For title/artist/recently_added, the displayed order is just a sorted version of tracks — that's a `useMemo([tracks, sortKey])`, not state.

The current approach uses a `useEffect` to re-sync `displayedTracks` whenever tracks changes, which causes a bug: if the user has Shuffle active and adds/removes/drags a track, the effect re-runs the shuffle and silently randomizes the order again.

Shuffle is the only case that genuinely needs stored state (because it's random). Everything else should be computed inline.

yateendogra2109:
Hello @anshg1214 .
I fixed the play-all issue , Also applied useMemo for other sort options excluding shuffle.
For shuffle , I stored it as state  and applied logic to handle insertion/deletion of track during active shuffle , now the order of other tracks remain same without re-shuffle.

anshg1214 (approved):

PR #3738, opened 2026-06-17
fix: LinkListens Test

anshg1214:
Since pagination is client-side, changing `?page=` shouldn’t need route loader changes. The shouldRevalidate hook added in #3729 was masking a flaky test, not fixing a real bug.

The test is updated to render LinkListensPage with MemoryRouter and wait properly for both Next and Previous navigation.


MonkeyDo (approved):
Thanks for fixing this one!

PR #3740, opened 2026-06-18
docs: clarify when imported/backdated listens appear in statistics

Holek:
## Problem

The [Data Update Intervals](https://listenbrainz.readthedocs.io/en/latest/general/data-update-intervals.html) table lists **"Updating statistics for new listens → Daily"**. A user who imports a large historical archive (e.g. a Spotify *extended streaming history* spanning years) reasonably reads this as "my stats will update tomorrow" — but they don't, and the page never explains why. The only relevant line is **"Full dumps → 1st and 15th"**, which reads like it's about the downloadable database dumps rather than a stats rebuild.

This came up in the community Discord: a new user imported ~10 years of Spotify history, saw listen counts update but stats stay flat, and dug through the wiki without finding an answer. A maintainer confirmed the relevant behaviour is the full dump on the 1st/15th, and that the wording could be clearer.

## Why it happens

"New listens" silently means *recently played* listens. The daily incremental stats job filters listens by `listened_at` against the window each statistic covers (this week/month/year), so imported listens — which carry old `listened_at` timestamps even though they were just submitted — are skipped daily. They only enter statistics when the full-dump rebuild on the 1st/15th recalculates everything from the complete history.

## Changes (docs + one UI note)

- **`docs/general/data-update-intervals.rst`**
  - Expand footnote `[#f2]` to spell out that "new listens" = recently played, and that imported/backdated listens land at the next full dump.
  - Add a note to the *User Statistics* section covering import behaviour, the ~2-week expectation, and clarifying that "full dump" is the same process behind the downloadable data dumps.
- **`frontend/js/src/settings/import/ImportListens.tsx`**
  - Add a short note to the *Import completed!* message linking to the intervals docs, so the expectation is set right where people import.

No behavioural/code-logic changes — wording only.

Opening as a draft for maintainer review of the wording; happy to adjust tone, trim, or split the UI note into a separate PR if preferred.

### DIFF
--- a/docs/general/data-update-intervals.rst
+++ b/docs/general/data-update-intervals.rst
@@ -9,6 +9,7 @@ System                                          Update schedule
 Receiving listens, updating listen counts		Immediate [#f1]_
 Deleting listens					                  Removed at the top of the next hour (UTC)
 Updating statistics for new listens			    Daily [#f2]_
+Updating statistics for older listens/imports   1st and 15th of each month
 Artist/album stats				                  Daily
 Removing deleted listens from stats			    On the 2nd and 16th of each month
 Full dumps						                      1st and 15th of each month
@@ -28,7 +29,7 @@ This complicated schedule is caused by ListenBrainz having a lot of interconnect
 different scales. For more details, read on!
 
 .. [#f1] Listens via a connected Spotify account may not be submitted immediately, causing a short delay
-.. [#f2] "New listens" means listens played recently. Imported or backdated listens only appear in statistics after the next full dump (1st or 15th); see `User Statistics`_.
+.. [#f2] Statistics may take longer on the 1st and 15th of each month
 .. [#f3] To generate daily playlists, follow `troi-bot <https://listenbrainz.org/user/troi-bot/>`_.
 
 Listens and Listen Counts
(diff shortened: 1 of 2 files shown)


Aerozol:
Thanks for the PR! imo this needs to be cut way down - the page is already too wordy. Most users will just look at the table and the footnotes - if the footnotes get longer they will stop looking at those too.

Aerozol:
Looking much better, thanks!

It is currently removing/replacing the intent of the second footnote - we want to let people know that updates may take longer to process on the 1st and the 15th.

Perhaps it would be simpler to add another row in the table, under 'Updating statistics for new listens':
[ Updating statistics for older listens/imports | 1st and 15th of each month ]

And leave the footnote as-is.

Holek:
Good call, done in fb6e920:

- Restored `[#f2]` to its original wording (stats may take longer on the 1st/15th).
- Added a new row directly under *Updating statistics for new listens*:

  | Updating statistics for older listens/imports | 1st and 15th of each month |

Verified the table still parses cleanly with docutils (the new row aligns to the same column boundary as the header). The short *Imported and backdated listens* note in the User Statistics section is still there for anyone who clicks through for the "why" + rough timeframe — happy to drop it if you'd rather keep the page table-only.

Aerozol:
Looks good to me! Thank you very much, this is a small but very helpful addition.

One of the developers will (@MonkeyDo or @amCap1712) will now have to approve for merge.

MonkeyDo (approved):
Thank you!

PR #3743, opened 2026-06-23
Entity updates/tidying

Aerozol:
# Solution

Minor edits, giving the entity pages a tidy and getting (kinda) familiar with the LB code again:

- Added rounded border to background gradient.
- Fixed rounded corner on album artwork.
- Added backup margin to the top (so there is a gap even if there is no breadcrumb nav).
- Made radio button evenly rounded.
- Changed Artist Radio button to read "Artist Radio", even when on artist pages (keep it consistent across entity pages).

# Action

Don't have a local setup to test, just in the web editor, sorry!

### DIFF
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -5,9 +5,11 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
 .entity-page-header {
   padding: 2em;
   gap: 2em;
+  border-radius: 8px;
   background: linear-gradient(45deg, var(--bg-color), transparent 80%);
   flex-wrap: wrap;
   justify-content: center;
+  margin-top: 1.5em;
   .cover-art {
     aspect-ratio: 1;
     height: 100%;
@@ -16,6 +18,7 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
     flex-grow: 1;
     flex-shrink: 0;
     border-radius: 5px;
+    overflow: hidden
     text-align: center;
     > * {
       max-width: inherit;
@@ -57,6 +60,8 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
   .lb-radio-button {
     align-self: flex-end;
     gap: 0.1em;
+    border-top-left-radius: 0.5rem;
+    border-bottom-left-radius: 0.5rem;
     .tags-list {
       overflow: hidden;
       text-overflow: ellipsis;
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -33,7 +33,7 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
   }
   .artist-info,
   .playlist-info {
-    flex-grow: 2;
+    flex-grow: 1.5;
     display: flex;
     flex-direction: column;
     min-width: 15em;
--- a/frontend/css/sass/entity-pages.scss
+++ b/frontend/css/sass/entity-pages.scss
@@ -18,7 +18,7 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
     flex-grow: 1;
     flex-shrink: 0;
     border-radius: 5px;
-    overflow: hidden
+    overflow: hidden;
     text-align: center;
     > * {
       max-width: inherit;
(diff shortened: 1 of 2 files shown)


frontend/css/sass/entity-pages.scss
```diff
@@ -5,9 +5,11 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
 .entity-page-header {
   padding: 2em;
   gap: 2em;
+  border-radius: 8px;
```
MonkeyDo, line 8: Ideally we would have a variable defined for a standard border-radius value to be used in most scenarios, but that doesn't seem to be the case in the codebase, leading to different values being [used willy-nilly](https://github.com/search?q=repo%3Ametabrainz%2Flistenbrainz-server+border-radius+language%3ASCSS&type=code&l=SCSS).

Not suggesting we change anything now, just reinforcing what I was saying in my other comment about `$btn-border-radius-lg`.

frontend/css/sass/entity-pages.scss
```diff
@@ -5,9 +5,11 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
...
   background: linear-gradient(45deg, var(--bg-color), transparent 80%);
   flex-wrap: wrap;
   justify-content: center;
+  margin-top: 1.5em;
```
MonkeyDo, line 12: ~~This change didn't seem to do anything when I added it, because the secondary nav above it already has `margin-bottom: 1.5em;` and because of the nature of CSS the two margins overlap.~~

Never mid, just re-read the description, this is intended when no secondary nav is present on the page

frontend/css/sass/entity-pages.scss
```diff
@@ -16,6 +18,7 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
     flex-grow: 1;
     flex-shrink: 0;
     border-radius: 5px;
+    overflow: hidden
```
MonkeyDo: Wow, that's a good fix and a stark difference!
The cover art becomes smaller because without the hidden overflow one of the flexbox rules was not being respected.
In order to keep the intended 400x400 px max size, we need one more change below to the artist-header class, specifically the flex-grow property.

The linter is unhappy about the missing semicolon at the end:

```suggestion
    overflow: hidden;
```

frontend/css/sass/entity-pages.scss
MonkeyDo: Here is the change required to give more space to the cover art collage:
```suggestion
    flex-grow: 1.5;
```

frontend/css/sass/entity-pages.scss
```diff
@@ -57,6 +60,8 @@ $white-gradient: linear-gradient(to bottom, transparent, #fff 65%);
   .lb-radio-button {
     align-self: flex-end;
     gap: 0.1em;
+    border-top-left-radius: 0.5rem;
+    border-bottom-left-radius: 0.5rem;
```
MonkeyDo, line 64: Would suggest this instead:
the overflow is required so that the border-radius actually hides the corners.
The border-radius declaration just ensures we don't accidentally end up with different values for the left-hand vs right-hand side.
The `$btn-border-radius-lg` value is [defined globally](https://github.com/metabrainz/listenbrainz-server/blob/50bbb3e1ac44cc7a0b18de1d7076c5e22d74f6ee/frontend/css/sass/variables.scss#L102), using it here instead of its computed value ensures everything stays consistent if it ever changes

```suggestion
    overflow: hidden;
    border-radius: $btn-border-radius-lg;
```

MonkeyDo (changes requested):
Good improvements, thanks for the cleanup!

frontend/css/sass/entity-pages.scss
Aerozol: Thanks monkey - did you check that this doesn't round the right-side corners as well? Since this butts up against the dropdown button on the right.

frontend/css/sass/entity-pages.scss
Aerozol: I am also unhappy about the missing semicolon!

frontend/css/sass/entity-pages.scss
Aerozol: I'm not 100% sure what you mean, but let's do it!

frontend/css/sass/entity-pages.scss
MonkeyDo: Yep!

MonkeyDo (approved):
Thanks for the improvements!

PR #3744, opened 2026-06-24
Add event interaction and artist follow tables

shirsakm:
> [!NOTE]
> Since my changes are stacked on top of my first PR, this also contains all of my first PR. I don't know how to clean it up (yet) and I will try to soon. Though hopefully it should fix itself, once the first PR gets merged.

# Problem

:beginner: This PR covers Week 3 of my GSoC project.

Adds the table to enable users to follow artists, which will allow us to display relevant upcoming events to them. Can be extended for other fun stuff later, as well.

Also, adds the table to allow users to start watching an event.

# Solution

I have added the `user_artist_relationship` table and the `event_interaction` table in the Postgres database. Both are the same as was in my proposal. `user_artist_relationship` is almost exactly a copy of the `user_relationship` table.

Added functions for both tables to the data access layer in `db/`, and tests to check the functionality of the functions.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:
* [x] understand all the changes made in this PR

I used AI in creating the initial `test_user_artist_relationship.py` file before updating the tests that didn't make sense. Tests are the worst, I don't like tests. They are important, but I don't like tests.

# Action

Nothing.

### DIFF
--- a/admin/sql/create_indexes.sql
+++ b/admin/sql/create_indexes.sql
@@ -35,6 +35,12 @@ CREATE UNIQUE INDEX user_id_entity_ndx_do_not_recommend ON recommendation.do_not
 CREATE INDEX user_0_user_relationship_ndx ON user_relationship (user_0);
 CREATE INDEX user_1_user_relationship_ndx ON user_relationship (user_1);
 
+CREATE INDEX user_id_user_artist_relationship_ndx ON user_artist_relationship (user_id);
+CREATE INDEX artist_mbid_user_artist_relationship_ndx ON user_artist_relationship (artist_mbid);
+
+CREATE INDEX user_id_event_interaction_ndx ON event_interaction (user_id);
+CREATE INDEX event_mbid_event_interaction_ndx ON event_interaction (event_mbid);
+
 CREATE UNIQUE INDEX user_id_rec_mbid_ndx_feedback ON recommendation_feedback (user_id, recording_mbid);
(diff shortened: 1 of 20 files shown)


admin/sql/create_indexes.sql
```diff
@@ -35,6 +35,12 @@ CREATE UNIQUE INDEX user_id_entity_ndx_do_not_recommend ON recommendation.do_not
 CREATE INDEX user_0_user_relationship_ndx ON user_relationship (user_0);
 CREATE INDEX user_1_user_relationship_ndx ON user_relationship (user_1);
 
+CREATE INDEX user_id_user_artist_relationship_ndx ON user_artist_relationship (user_id);
```
anshg1214, line 38: Let’s review if we need composite indexes here.

listenbrainz/db/event_interaction.py
```diff
@@ -0,0 +1,121 @@
+from typing import List
+
+import sqlalchemy
+
+
+def watch_event(db_conn, user_id: int, event_mbid: str) -> None:
```
anshg1214, line 6: Let’s add RETURNING 1 in the SQL query - that helps callers know if the row was created or already existed

Same for Delete operation as well

anshg1214 (changes requested):
I reviewed the changes and they look good. A few improvements:

PR #3747, opened 2026-06-29
Allow pausing/unpausing users from reported_users admin view

MonkeyDo:
Copying code from listenbrainz/model/users.py to allow easily pausing reported users (previously only available from the users admin view, which does not have the reported status).

Used Codex to refactor the code in commit 471fd23

### DIFF
--- a/listenbrainz/db/user.py
+++ b/listenbrainz/db/user.py
@@ -593,55 +593,80 @@ def get_all_usernames():
     return user_id_map
 
 
-def pause(db_conn,id):
-    """ Sets the user's is_paused flag to true
-    with specified row ID from the database.
-    
+def pause(db_conn, user_ids):
+    """ Sets the user's is_paused flag to true with specified row ID(s) from the database.
+
     Args:
         db_conn: database connection
-        id (int): the row ID of the listenbrainz user
+        user_ids (int or list): row ID(s) of the listenbrainz user(s)
     """
+    return set_users_paused(db_conn, _normalize_user_ids(user_ids), True)
+
+
+def unpause(db_conn, user_ids):
+    """ Sets the user's is_paused flag to false with specified row ID(s) from the database.
+
+    Args:
+        db_conn: database connection
+        user_ids (int or list): row ID(s) of the listenbrainz user(s)
+    """
+    return set_users_paused(db_conn, _normalize_user_ids(user_ids), False)
+
+
+def _normalize_user_ids(user_ids):
+    if isinstance(user_ids, (str, int)):
+        return [user_ids]
+    return user_ids
+
+
+def set_users_paused(db_conn, user_ids, is_paused):
+    user_ids = [int(user_id) for user_id in user_ids]
+    return _set_users_paused(db_conn, ":user_ids", {"user_ids": tuple(user_ids)}, is_paused)
+
+
+def set_reported_users_paused(db_conn, report_ids, is_paused):
+    report_ids = [int(report_id) for report_id in report_ids]
+    return _set_users_paused(
+        db_conn,
+        """
+            (SELECT DISTINCT reported_user_id
+               FROM reported_users
+              WHERE id IN :report_ids)
+        """,
+        {"report_ids": tuple(report_ids)},
+        is_paused,
+    )
+
+
+def _set_users_paused(db_conn, user_ids_expression, params, is_paused):
+    if not any(params.values()):
+        return []
+
     try:
-        db_conn.execute(sqlalchemy.text("""
+        result = db_conn.execute(sqlalchemy.text(f"""
             UPDATE "user"
-               SET is_paused = true
-             WHERE id = :id
-            """), {
-            'id': id,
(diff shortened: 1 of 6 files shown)


listenbrainz/model/reported_users.py
```diff
@@ -13,6 +17,7 @@ class ReportedUsers(db.Model):
     reason = db.Column(db.String)
     reporter = db.relationship('User', foreign_keys=[reporter_user_id])
     reported = db.relationship('User', foreign_keys=[reported_user_id])
+    is_paused = db.Column(db.Boolean, nullable=False)
```
amCap1712: I don't think this makes sense. Whether a user is paused or not is is not the property of a report.

MonkeyDo: Ah right, and it's in another table too

listenbrainz/model/reported_users.py
```diff
@@ -21,10 +25,53 @@ class ReportedUserAdminView(AdminModelView):
...
+        'reported.is_paused',
     ]
 
     column_formatters = {
         "reporter.musicbrainz_id": lambda view, context, model, name: generate_username_link(model.reporter.musicbrainz_id),
         "reported.musicbrainz_id": lambda view, context, model, name: generate_username_link(model.reported.musicbrainz_id)
     }
+
+    # With select action to pause users.
+    @action(
```
amCap1712, line 38: Since its a bulk action, query all users in 1 go both pause and unpause?

MonkeyDo: Done!
Also did the same for the user view that was doing the calls sequentially

MonkeyDo:
Also reworded the email that is sent to paused users, to change "we will reach out to you" to "please reach out to us" as we are unlikely to try to contact spam accounts that we block.

listenbrainz/model/reported_users.py
MonkeyDo: Actually I rolled back those changes, we also want to send notifications to each user, and the refactor removed that. I think it might make more sense to send the emails one by one after each user pause.

I can still refactor it as a bulk SQL query, but I will also need to reimplement the notification part.

listenbrainz/db/user.py
```diff
@@ -593,55 +593,80 @@ def get_all_usernames():
...
+            "is_paused": is_paused,
         })
+        users = result.mappings().all()
         db_conn.commit()
-        _notify_user_paused(db_conn,id,False)
+
+        for user in users:
+            _notify_user_paused(db_conn, user["id"], is_paused)
+
+        return [user["musicbrainz_id"] for user in users]
```
anshg1214: The except only catches `sqlalchemy.exc.ProgrammingError`. Let's say `send_email` raises for user 1, it propagates past that except, out to the view's except Exception, which flashes `Failed to pause users`. But every selected user was alreasy paused and committed, and users 2...N never get their email.

listenbrainz/model/reported_users.py
```diff
@@ -1,6 +1,10 @@
 from listenbrainz.model import db
 from listenbrainz.model.utils import generate_username_link
+from listenbrainz.db import user as db_user
+from listenbrainz.webserver import db_conn
 from listenbrainz.webserver.admin import AdminModelView
+from flask_admin.model import action
+from flask import  flash, redirect
```
anshg1214:
```suggestion
from flask import flash, redirect
```

listenbrainz/db/user.py
```diff
@@ -593,55 +593,80 @@ def get_all_usernames():
...
 
     except sqlalchemy.exc.ProgrammingError as err:
         logger.error(err)
-        raise DatabaseException("Couldn't unpause user: %s" % str(err))
+        action = "pause" if is_paused else "unpause"
+        raise DatabaseException("Couldn't %s user: %s" % (action, str(err)))
```
anshg1214, line 661: The except only catches `sqlalchemy.exc.ProgrammingError`. If `send_mail` raises for user 1 - it propagates past that except, out to the view's except Exception, which flashes "Failed to pause users."

But every selected user was already paused and committed, and users 2..N never get their email.

listenbrainz/db/user.py
```diff
@@ -593,55 +593,80 @@ def get_all_usernames():
...
+        db_conn,
+        """
+            (SELECT DISTINCT reported_user_id
+               FROM reported_users
+              WHERE id IN :report_ids)
+        """,
+        {"report_ids": tuple(report_ids)},
+        is_paused,
+    )
+
```
anshg1214, line 639: Let's add a test for this path as well.

anshg1214 (changes requested):

listenbrainz/db/user.py
MonkeyDo: I also added reporting the user IDs for failed notifications (if any) and flashing them on the admin page

anshg1214 (approved):
LG

PR #3748, opened 2026-06-29
Fix incorrect parameter documentation in db helpers

eeshsaxena:
### Problem

Four docstrings in the `listenbrainz` package document parameters that do not match their function signatures:

- `db.user.get_by_login_id` documents `id` instead of its actual `login_id` parameter.
- `domain.apple.AppleService.add_new_user` documents a `token` parameter it does not take (the token is fetched internally via `fetch_access_token`).
- `listenstore.timescale_listenstore.TimescaleListenStore.delete` documents a `musicbrainz_id` parameter it does not take (it takes `user_id`).
- `db.pinned_recording.update_comment` documents `user_id` instead of its actual `row_id` parameter.

### Solution

Corrects the documented parameter names/entries to match each signature. Documentation-only change; no behaviour is affected.

### DIFF
--- a/listenbrainz/db/pinned_recording.py
+++ b/listenbrainz/db/pinned_recording.py
@@ -274,7 +274,7 @@ def update_comment(db_conn, row_id: int, blurb_content: str) -> bool:
 
         Args:
             db_conn: Database connection
-            user_id: The user for which the comment of pinned record has to be updated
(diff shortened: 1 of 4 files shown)


amCap1712 (approved):
LGTM, thanks!

PR #3751, opened 2026-07-02
Optimize MessyBrainz lookup during listen ingestion

amCap1712:
Bulk-resolve MessyBrainz submissions in the timescale writer path instead of doing one lookup and optional insert per listen. Deduplicate equivalent submissions within a batch, preserve the existing earliest-submitted MSID behavior for historical duplicates, and bulk-insert only missing rows.

Increase the MsB lookup chunk size to match the RMQ listen batch size, and fix duration_ms handling so it is sent to MsB when duration is absent.

### DIFF
--- a/listenbrainz/messybrainz/__init__.py
+++ b/listenbrainz/messybrainz/__init__.py
@@ -1,6 +1,7 @@
 import uuid
 from typing import Iterable
 
+import psycopg2
 import sqlalchemy
 import sqlalchemy.exc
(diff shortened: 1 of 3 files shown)


MonkeyDo (approved):
Looks good, thanks for opening that PR

MonkeyDo:
Failing tests are unrelated, see #3749

anshg1214 (approved):

PR #3752, opened 2026-07-03
Fix incorrect parameter documentation in db helpers

eeshsaxena:
A few more `Args:` entries in the db helpers documented parameter names that don't match the actual function signatures:

- `insert_user_missing_musicbrainz_data`: `data` -> `missing_musicbrainz_data`
- `search_playlist`: removed the phantom `include_private` argument (not in the signature)
- `insert_sitewide_stats`: replaced the phantom `database` argument with the actual `stats_type` and `stats_range` parameters

Documentation-only change; no behavior change.

### DIFF
--- a/listenbrainz/db/missing_musicbrainz_data.py
+++ b/listenbrainz/db/missing_musicbrainz_data.py
@@ -37,7 +37,7 @@ def insert_user_missing_musicbrainz_data(db_conn, user_id: int,
         Args:
             db_conn: database connection
             user_id : row id of the user.
-            data : Data that is submitted to ListenBrainz by the users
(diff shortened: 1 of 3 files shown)


eeshsaxena:
Just a note on the CI failure: the failing test is test_listens_importer.py::ImportTestCase::test_import_spotify (AssertionError: 4 != 6), which is an integration test about Spotify listen counting - unrelated to my changes which only touch docstrings in db/missing_musicbrainz_data.py, db/playlist.py, and db/stats.py. Looks like a pre-existing flaky integration test.

anshg1214 (approved):

PR #3753, opened 2026-07-03
LB-1584: Limit time and number of passes for timescale queries

MonkeyDo:
# Problem
We have had this issue for a very long time. Because of the way we partition timescale listens table by time buckets (per-month), if there is a large gap between a user's listens, the query can take a lot of time (I've seen >690 seconds for API calls!) and causes a timeout on the user side or a completely blank dashboard page while the query runs.
With our current setup the query will run, expanding the scope (adding 3 months each pass) until it has the requested number of listens or it reaches 10 passes (as a safety valve).
See LB-1584

# Solution

Instead of blindly continuing until reaching the requested number of listens, stop early if a certain time threshold is reached, or 3 passes (= 9 months). *We can tweak the max passes, it could be higher I think*
This requires some changes in the API response and in the user dashboard UI.
We now return some additional information in the API response in this format:
```json
"searchStatus": {
  "continueMaxTs": 1780420144,
  "partial": true
},
```
Where `partial: true` indicates we returned early and that the count will be less than requested, and `continueMaxTs`/`continueMinTs` (depending on original search params) gives the user a timestamp from where to continue requesting data.
This is because instead of the usual method of using the first/last listen's timestamp, these partial requests can sometimes return 0 results (say if you did not listen to any listen in over 9 months). The continuation timestamp represents the boundary of the last fetch window (so in drastic cases, requested timestamp minus 9 months).

Not certain about the phrasing of "searchStatus" since this isn't really search? Suggestions welcome.

### Limit query time
The new FETCH_LISTENS_SOFT_TIME_LIMIT_MS config variable is set to None by default (for local dev) but can be tweaked in production to limit maximum query times.
If by the end of the current pass the elapsed time is greater than that value, the query will stop there and return early.

The combination of the max passes and time limit allow for multiple passes to happen (if they are fast enough), while not going over a time budget.

### UI changes
Added some text and a button for easier navigation, when the page is not showing the usual 25 listens:
image

And on the last page:
image

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

Used Codex for most of the backend changes and tests (commits df36fe6ac and e4b0e6268), based on an implementation plan I wrote from scratch because I didn't trust codex to generate code if I didn't fully understand the issue and solution.

# Action

Returning fewer than the requested number of listens is an API breaking change. We will need to announce it, perhaps even wait before merging and deploying it.
We could potentially enable this mechanism using query params in such a way that regular API calls are not affected, and only the calls from our front-end are limited (`@user_bp.post("/<mb_username:user_name>/")`). But that kind of defeats the purpose...

Thoroughly test this on test, then beta, before merging.
Only tested in local dev so far.

### DIFF
--- a/listenbrainz/listenstore/timescale_listenstore.py
+++ b/listenbrainz/listenstore/timescale_listenstore.py
@@ -37,6 +37,7 @@ DEFAULT_FETCH_WINDOW = timedelta(days=30)  # 30 days
 
 # When expanding the search, how fast should the bounds be moved out
 WINDOW_SIZE_MULTIPLIER = 3
+MAX_FETCH_PASSES = 3  # maximum number of times to expand the search window when fetching listens
 
 LISTEN_COUNT_BUCKET_WIDTH = 2592000
 
@@ -201,7 +202,15 @@ class TimescaleListenStore:
 
         return inserted_rows
 
-    def fetch_listens(self, user: Dict, from_ts: datetime = None, to_ts: datetime = None, limit: int = DEFAULT_LISTENS_PER_FETCH):
+    def fetch_listens(
+        self,
+        user: Dict,
+        from_ts: datetime = None,
+        to_ts: datetime = None,
+        limit: int = DEFAULT_LISTENS_PER_FETCH,
+        max_passes: int = 1,
+        return_search_status: bool = False,
+    ):
         """ The timestamps are stored as UTC in the postgres datebase while on retrieving
             the value they are converted to the local server's timezone. So to compare
             datetime object we need to create a object in the same timezone as the server.
@@ -213,7 +222,16 @@ class TimescaleListenStore:
                 listens will be returned in descending order
             to_ts: seconds since epoch, in float
             limit: the maximum number of items to return
+            max_passes: the maximum number of time windows to search
+            return_search_status: if True, return metadata describing a partial search
         """
+        if max_passes < 1:
+            raise ValueError("max_passes should be at least 1")
+        if max_passes > MAX_FETCH_PASSES:
+            raise ValueError("max_passes should be at most %d" % MAX_FETCH_PASSES)
+
+        search_status = {"partial": False}
+
         if from_ts and to_ts and from_ts >= to_ts:
             raise ValueError("from_ts should be less than to_ts")
         if from_ts:
@@ -224,6 +242,8 @@ class TimescaleListenStore:
         min_user_ts, max_user_ts = self.get_timestamps_for_user(user["id"])
 
         if min_user_ts == EPOCH and max_user_ts == EPOCH:
+            if return_search_status:
+                return [], min_user_ts, max_user_ts, search_status
             return [], min_user_ts, max_user_ts
 
         if to_ts is None and from_ts is None:
@@ -305,13 +325,9 @@ class TimescaleListenStore:
         t0 = time.monotonic()
 
         passes = 0
+        stopped_early = False
         while True:
             passes += 1
-
-            # Oh shit valve. I'm keeping it here for the time being. :)
-            if passes == 10:
-                done = True
-                break
 
             curs = ts_conn.execute(
                 sqlalchemy.text(query),
@@ -329,6 +345,11 @@ class TimescaleListenStore:
                         break
 
                     if to_ts > datetime.now(tz=timezone.utc) + MAX_FUTURE_SECONDS:
+                        done = True
+                        break
+
+                    if passes >= max_passes:
+                        stopped_early = True
                         done = True
                         break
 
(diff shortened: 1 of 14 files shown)


MonkeyDo:
Sorry about requesting a review before the tests passed!
Now all ready for review.

MonkeyDo:
I've deployed limit-timescale-passes to the test website to ensure it is working fine.
On test, since the DB is basically empty, we can't really see any improvement time-wise, but at least everything seems to be working as expected:
Before:
```log
2026-07-08 18:09:34,771 listenbrainz.webserver INFO     fetch listens mr_monkey 0.02s (1 passes)
2026-07-08 18:18:47,351 listenbrainz.webserver INFO     fetch listens mr_monkey 0.06s (4 passes)
2026-07-08 21:18:04,667 listenbrainz.webserver INFO     fetch listens mr_monkey 0.04s (1 passes)
2026-07-09 11:49:17,651 listenbrainz.webserver INFO     fetch listens mr_monkey 0.05s (4 passes)
2026-07-09 11:50:29,356 listenbrainz.webserver INFO     fetch listens mr_monkey 0.02s (4 passes)
2026-07-09 11:50:33,933 listenbrainz.webserver INFO     fetch listens mr_monkey 0.00s (1 passes)
2026-07-09 11:51:06,800 listenbrainz.webserver INFO     fetch listens mr_monkey 0.01s (1 passes)
2026-07-09 11:51:10,074 listenbrainz.webserver INFO     fetch listens mr_monkey 0.02s (4 passes)
```
After:
```log
2026-07-09 11:53:52,726 listenbrainz.webserver INFO     fetch listens mr_monkey 0.02s (1 passes)
2026-07-09 11:55:19,749 listenbrainz.webserver INFO     fetch listens mr_monkey 0.02s (3 passes, partial)
2026-07-09 11:56:01,167 listenbrainz.webserver INFO     fetch listens mr_monkey 0.02s (3 passes)
```

Obviously not much to go off of, other than no more than 3 passes and the "partial" indicator in the log.

listenbrainz/listenstore/timescale_listenstore.py
```diff
@@ -374,8 +408,24 @@ def fetch_listens(self, user: Dict, from_ts: datetime = None, to_ts: datetime =
...
             listens.reverse()
 
-        self.log.info("fetch listens %s %.2fs (%d passes)" % (user["musicbrainz_id"], fetch_listens_time, passes))
+        if stopped_early:
+            search_status["partial"] = True
+            if from_dynamic:
+                search_status["continue_max_ts"] = int(from_ts.timestamp())
+            elif to_dynamic:
+                search_status["continue_min_ts"] = int(to_ts.timestamp())
+
```
anshg1214, line 417:
```suggestion
        if stopped_early:
            search_status["partial"] = True
            if from_dynamic:
                search_status["continue_max_ts"] = int(from_ts.timestamp()) + 1
            elif to_dynamic:
                search_status["continue_min_ts"] = int(to_ts.timestamp()) - 1

```


The fetch query uses exclusive bounds (listened_at > from_ts AND listened_at < to_ts). When a search stops early, continue_max_ts/continue_min_ts are set to the raw from_ts/to_ts with no ±1 adjustment.

A follow-up request using that value as its own max_ts/min_ts builds a window that's again exclusive at that same second — so any listen landing exactly on that boundary is excluded from both windows and effectively vanishes from the user's history permanently.

listenbrainz/webserver/listens_cache.py
```diff
@@ -9,7 +9,9 @@ def _get_listens_cache_key(
     user_id: int,
     min_ts: Optional[int] = None,
     max_ts: Optional[int] = None,
-    count: Optional[int] = None
+    count: Optional[int] = None,
+    max_passes: Optional[int] = None,
```
anshg1214, line 13: A small observation here: Let's say the DB is slow / degraded. Each session takes longer than usual. By two passes, elapsed time crosses the soft limit, so stopped_early fires and it returns only 8 out of the requested 25 listens.

This data gets cached for 3600 seconds.

Assuming the DB load is back to normal after a minute. If a fresh (uncached) request came right now, it should have returned 25 listens with `partial: false`.

But, since it has been cached, we'll be returning the 8 listens for the next 3600 seconds. 

One possible solution I can thnk of is to cache the results for a shorter duration, if we're returning partial results.

anshg1214 (changes requested):

listenbrainz/listenstore/timescale_listenstore.py
MonkeyDo: Thanks for catching that!

listenbrainz/webserver/listens_cache.py
MonkeyDo: Good thinking. I'll have to review this caching situation with fresh eyes

MonkeyDo:
Update: This was a patch rather than a fix.
We are looking at moving away from timescale (which is ideal for time series, but not for our use case) and instead use regular PG and partition the listens table by user, so these lookups get much,. much faster.

MonkeyDo:
The aforementioned migration to a user-partitioned Postgres database is in progress. Closing this PR which is not needed anymore.

PR #3746, opened 2026-06-28
GSoC-26 Add Tags To Playlists - Backend

yateendogra2109:
Ticket : [LB-1302](https://tickets.metabrainz.org/browse/LB-1302)

# Description

Adds backend support for playlist tags , so users can label and filter  playlists by tag.

Tags are stored in TimescaleDB, returned in playlist JSPF metadata, and can be used to filter playlist lists and search results. A new sidebar endpoint returns tags with playlist counts, with separate views for owned vs collaborated playlists.

Only the playlist owner may add or remove tags; collaborators can view them. This PR is backend only.

# Implementation

## Database
New playlist.playlist_tag table 
Schema in admin/timescale/create_tables.sql
Migration: admin/timescale/updates/2026-05-04-add-playlist-tags.sql

## DB layer 

- Tag normalization: lowercase, whitespace collapse, max 40 chars, deduplication via _normalize_playlist_tags
- CRUD: add_tags_to_playlist, remove_tag_from_playlist, get_tags_for_playlists, get_tags_for_user
- GET .../playlists/tags?collaborator=true — tags from collaborated playlists only; default is owned

# Tests 

Three integration tests in `listenbrainz/tests/integration/test_playlist_api.py:`

`test_playlist_tags_owner_only` -  Owner can add/remove; collaborator gets 403; tags visible in playlist fetch

`test_playlist_tags_sidebar_endpoint_and_filtering `- Sidebar endpoint, single/multi-tag AND filter, tag + search

`test_playlist_tags_sidebar_by_section` - Owned vs collaborative tag lists via ?collaborator=true

# AI Usage 

I had used AI for Debugging and understanding existing code.

PR #3755, opened 2026-07-06
GSOC-26 Add Tags To Playlists - Frontend

yateendogra2109:
# Problem

Currently there is no UI for playlist tags on top of gsoc-organize-playlists, which already includes the backend work ([#3746](https://github.com/metabrainz/listenbrainz-server/pull/3746)).

The problem to solve : Users can add tags when creating or editing a playlist, see tags on playlist cards and the playlist page, and filter their playlists using a tags sidebar. Tag filters stored in the URL (?tag=) and work together with the existing playlist search and sort from the gsoc branch.

# Implementation

## User playlists page 
Tags sidebar — lists available tags with counts , supports multi-select filtering
URL-driven filtering — active tags stored as ?tag=rock,jazz , switching playlist type clears the tag filter
Reload on change — sidebar tag list refreshes after create, edit, or delete
Integrated with existing search, sort.
## Tag editing 
Multi-select tag input  with suggestions from the user’s existing tags
Tags saved via tag API endpoints after playlist create/edit, then playlist is re-fetched so returned JSPF includes tags
## Tag display
Playlist cards — show up to 3 tags with “+N more” overflow
Single playlist page — show all tags in the header

# AI usage

* [ ] I did not use any AI
* [ Y ] I have used AI in this PR (add more details below)

I used claude to code some part especially the responsiveness .

#### If you did use AI:
* [ Y ] I used AI tools for coding
* [ Y ] I understand all the changes made in this PR

Screenshot 2026-07-14 121229
image
Screenshot 2026-07-14 121323
Screenshot 2026-07-14 121405

image

## New UI

image

image


anshg1214 (changes requested):
The PR looks good on a high level. After that, we have revamped the frontend.
1. The search input looks a bit bloated. We need to reduce the text size here.
2. As I attached a screen recording, whenever I'm filtering by a tag, you can observe that the background color of the sort by input box changes. This is not an expected behavior.
3. I randomly added 25 long tags to a particular playlist. On the side panel, we are technically allowing users to have a selector for all the tags, each with 15 different tags. That is then equivalent to 300 different tags, right? For power users, this won't make much sense, as we are technically showing all 300 tags on the side panel. It will give a very bad scrolling experience. Let's try to optimize that if we can have a better experience here. 

https://github.com/user-attachments/assets/90c05439-0e23-4d33-9e0c-6c762cb40be9

yateendogra2109:
Hi @anshg1214 

I fixed the first two issues.
However I didn't understand the tags math you quoted in third point , You  are pointing out to displaying every tag in sidebar which will make the list grow enormous ?
If we just show the first N tags and add autocomplete in the search bar , this couls be one solution , but lets say a user wants to see his full list or he want to find a tag where a user forgets the starting initial of the tag , in that case he would not able to find the tag through search also. 
What else could be done is showing first N tags and then a liitle button to see more tags , but i didn't find it any better because we currently had a scroll , so user can willingly scroll/not scroll depending on his need .

So , I would like to keep the current behaviour as it is , However if you have any suggestion for this , then please do share .

anshg1214:
> However I didn't understand the tags math you quoted in third point , You are pointing out to displaying every tag in sidebar which will make the list grow enormous ?

Yes

> If we just show the first N tags and add autocomplete in the search bar , this couls be one solution , but lets say a user wants to see his full list or he want to find a tag where a user forgets the starting initial of the tag , in that case he would not able to find the tag through search also.

I believe we should show the first 10 Tags. That's how Search should work. Let's begin with that.

yateendogra2109:
> > However I didn't understand the tags math you quoted in third point , You are pointing out to displaying every tag in sidebar which will make the list grow enormous ?
> 
> Yes
> 
> > If we just show the first N tags and add autocomplete in the search bar , this couls be one solution , but lets say a user wants to see his full list or he want to find a tag where a user forgets the starting initial of the tag , in that case he would not able to find the tag through search also.
> 
> I believe we should show the first 10 Tags. That's how Search should work. Let's begin with that.

Should we add autocomplete in search ?

anshg1214:
> Should we add autocomplete in search ?

Let's do a basic search for now.

frontend/js/src/user/playlists/Playlists.tsx
MonkeyDo, line 800: Unrelated to this PR, but I just noticed that this would overwrite any other search params there might have been. Would be better to use the existing searchParams as a base, making a copy of it first as I think it's mutable, and use `newParams.set("type", "collaborative")` then `setSearchParams(newParams);`

Might apply to other hooks too but I haven't checked.

frontend/js/src/utils/APIService.ts
```diff
@@ -1091,6 +1136,32 @@ export default class APIService {
     return response.json();
   };
 
+  getUserPlaylistTags = async (
+    userName: string,
+    userToken?: string,
+    collaborated: boolean = false
+  ): Promise<{ tags: Array<{ tag: string; count: number }> }> => {
```
MonkeyDo: This should use the existing type

```suggestion
  ): Promise<{ tags: PlaylistTag[] }> => {
```

MonkeyDo (changes requested):
Looking good :)

MonkeyDo (approved):
Good for me 👍

PR #3756, opened 2026-07-07
Add website alert messages and corresponding admin interface

MonkeyDo:
This is to allow us to flash messages on the website to let users know when the server is overloaded, for example, or other announcements.
The messages are stored in redis to survive restarts.
Tried using flask.flash but they are better for transient messages rather than this use case. Implemented it with our Toast library instead in the front end.

These new alerts are shown at the top of the page in a separate container from our other toast messages:
image

The admin interface:
image

# AI usage

* [ ] I did not use any AI

 I used Claude to set up the admin interface/back-end side of this PR, based on my own implementation plan.


listenbrainz/webserver/templates/index.html
```diff
@@ -94,6 +94,7 @@
 <script src="{{ url_for('static', filename='js/lib/dragscroll.js') }}"></script>
 <script id="global-react-props" type="application/json">{{ global_props|safe }}</script>
 <script id="page-react-props" type="application/json">{{ props|safe }}</script>
+<script id="initial-alerts" type="application/json">{{ initial_alerts|safe }}</script>
```
anshg1214: Let's change the ID here to `initial-alerts-data`. The same ID is being used by react-toastify.

anshg1214 (approved):
One change, otherwise LG

listenbrainz/webserver/templates/index.html
anshg1214: `getPageProps` does `getElementById("initial-alerts")` and only gets the right (script) element because it runs on DOMContentLoaded before React mounts the container. It works today purely by load-order luck.

MonkeyDo: Nice catch
I used initial-alerts-props to mimic the page-react-props and global-react-props next to it.

PR #3759, opened 2026-07-08
Fix test database connection string

MonkeyDo:
timescale writer for the test environment is currently failing to insert listens with a connection error, trying (and failing) to connect as user listenbrainz_ts. Looks like some parts of the connection string in consul template were not changed accordingly.
```log
psycopg2.OperationalError connection to server at "10.10.10.31", port 13046 failed:
FATAL:  password authentication failed for user "listenbrainz_ts"
```

# AI usage

* [x] I did not use any AI


anshg1214 (approved):

MonkeyDo:
Confirmed working once deployed in test environment, listen was ingested and written to the test DB:
image

PR #3763, opened 2026-07-13
feat: Add claim-based work distribution for live polling importers

anshg1214:
# Problem

The live importers (LastFM, Spotify, LibreFM) previously fetched ALL users for a service with no coordination. Running multiple workers meant duplicate imports and wasted API calls.

# Solution

This PR adds a claim/release pattern using FOR UPDATE SKIP LOCKED so workers safely distribute users:

- Phase 1 (claim): quick transaction grabs N users, sets claimed_at, commits
- Phase 2 (process): imports run with no lock held
- Phase 3 (release): clears claimed_at on finish or error
- Auto-reclaim: users claimed >24h ago are re-eligible (crashed worker recovery)

Also moved `is_paused` filtering from Python to SQL to avoid wasting batch slots.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

I've used cursor for writing the code based on the plan I created manually.

# Action

   1. Add tests
   2. Run an SQL update script


listenbrainz/listens_importer/base.py
```diff
@@ -139,9 +142,6 @@ def process_all_users(self):
         failure = 0
         for user in users:
             try:
-                if user["is_paused"]:
-                    continue
```
MonkeyDo, line 143: Why remove this condition?
We still want to avoid importing listens for paused users.

Is it because `release_user_claim` doesn't get called if we `continue`?
If so we should also have a `release_user_claim` call in this condition.

listenbrainz/listens_importer/base.py
```diff
@@ -160,6 +160,8 @@ def process_all_users(self):
                 current_app.logger.error(f'{self.name} could not import listens for user %s:',
                                          user['musicbrainz_id'], exc_info=True)
                 failure += 1
+            finally:
+                self.service.release_user_claim(user['user_id'])
```
MonkeyDo: Would it make sense to release all users in that batch in one go lik ewe do for the claiming, instead of doing it user-per-user?
We don't have to worry about the unreleased users needing to be re-processed anytime soon, since they were just part of a batch.

MonkeyDo (commented):
Looks good, thanks for implementing that.
Have some questions/doubts below:

MonkeyDo:
Also, AI disclaimer please

listenbrainz/listens_importer/base.py
anshg1214: I've added this condition in the SQL Itself. Paused users are now excluded at claim time so they don't waste batch slots.

listenbrainz/listens_importer/base.py
anshg1214: Fair point. Releasing one-by-one means N small transactions instead of 1. 

But the trade-off: if a worker crashes mid-batch, batch-release means all claimed users stay locked until the 24-hour auto-reclaim. Per-user release means only the current user is still claimed - everyone already processed is free.

For a batch of 50 users where each takes seconds (Spotify polling), it doesn't matter much. For LastFM where one user can have 7000 pages, a crash 40 users in could leave 10 users locked for 24 hours unnecessarily.

listenbrainz/listens_importer/base.py
MonkeyDo: Perfect, that's much better

listenbrainz/listens_importer/base.py
MonkeyDo: In that case, I wonder if the 24h delay is too long.
If we half that value, we ensure users get processed at least once a day even in case of a crash (and I even think 6h would be reasonable).

Is there any way to perform a cleanup operation in case of crash? I mean best-effort of course, like some global error catching that will release the current batch. (and if the application crashes in a way that doesn't trigger it, we still have the fallback auto-reclaim)

anshg1214: Let's start with 12 hours I would say. We'll observe the behaviour and probably then move it to 6/8 hours?

> Is there any way to perform a cleanup operation in case of crash?

Good Point. I've now handled the case where we need to release the current user's claim when the container stops / restarts.

anshg1214: Do we have an idea what's the max amount of time can a task take?
If it takes more than 12 hours, we probably would need to implement heartbeats - worker periodically refreshes claimed_at while still working

MonkeyDo: To process a batch of 50 users, I expect it would always be under a few hours, I wouldn't worry about it too much.

MonkeyDo: So what do you reckon about the batch release in the end?
I think even in the SIGTERM path, it should release all users and not just the one in progress.

listenbrainz/db/listens_importer.py
```diff
@@ -105,6 +105,92 @@ def get_active_users_to_process(db_conn, service, exclude_error=False) -> list[d
...
+    result = db_conn.execute(text(f"""
+        WITH claimable AS (
+            SELECT listens_importer.id
+              FROM external_service_oauth
+              JOIN "user"
+                ON "user".id = external_service_oauth.user_id
+              JOIN listens_importer
+                ON listens_importer.external_service_oauth_id = external_service_oauth.id
+             WHERE {filter_str}
+          ORDER BY listens_importer.latest_listened_at ASC NULLS FIRST
```
MonkeyDo: I think this could result in processing the same users over and over, if `latest_listened_at` does not change.

Say for example the firtst batch of 50 users are all not listening to anything currently, and we are only running one container. Whyen the batch is processed (nothing imported) and the users released, won't the next batch have the same users ordered the same way, and reprocess the same 50 users over and over until a`latest_listened_at` changes?

Not sure how to deal with that though. A dynamic OFFSET of some sort? Or some other mechanism to avoid reprocessing the same users straight away?

MonkeyDo: Actually, I do see that we set `last_updated = now()` when we release the claim, so probably just using that as well in this query

listenbrainz/listens_importer/base.py
anshg1214: Agreed on batch release for shutdown. The worker now tracks pending claims for the current batch and calls release_user_claims() on SIGTERM for all unprocessed users, not just the one in progress.

Per-user release in finally is still there for normal completion and mid-batch crashes (so only the in-progress user stays claimed until timeout). On graceful shutdown, the whole pending batch is released in one go.

Also switched from SystemExit(0) in the signal handler to a shutdown flag so the main loop can exit cleanly after releasing claims.

listenbrainz/db/listens_importer.py
anshg1214: Good catch. You're right that with only latest_listened_at ordering, a single worker could keep reclaiming the same inactive users.

Updated the claim query to order by last_updated ASC NULLS FIRST, then latest_listened_at ASC NULLS FIRST. Since release_user_claim() already sets last_updated = now(), recently processed users rotate to the back of the queue.

listenbrainz/db/listens_importer.py
```diff
@@ -105,6 +105,101 @@ def get_active_users_to_process(db_conn, service, exclude_error=False) -> list[d
     return users
 
 
+CLAIM_TIMEOUT_HOURS = 12
```
MonkeyDo, line 108: Leaving a note here that we want to review this value after deployment.
Waiting 12 hours is not a great user experience.

listenbrainz/domain/importer_service.py
```diff
@@ -16,6 +16,18 @@ def get_active_users_to_process(self, exclude_error=True) -> list[dict]:
         """ Return list of active users for importing listens. """
         return listens_importer.get_active_users_to_process(db_conn, self.service, exclude_error)
 
+    def claim_users_to_process(self, batch_size=50, exclude_error=True) -> list[dict]:
+        """ Claim a batch of users for import processing using SKIP LOCKED. """
+        return listens_importer.claim_users_to_process(db_conn, self.service, batch_size, exclude_error)
+
+    def release_user_claim(self, user_id: int):
+        """ Release a claimed user after processing. """
+        listens_importer.release_user_claim(db_conn, user_id, self.service)
```
MonkeyDo, line 25: Do we need `listens_importer.release_user_claim`?
Can we not do `listens_importer.release_user_claims(db_conn, [user_id], self.service)` passing an array containing the single id instead?

MonkeyDo (commented):
Looks good, but it's not easy to test properly in local dev.

Is there a way you could stress-test this locally so that we're not testing in prod? Might involve having to replace the importers with fake ones that do a sleep call instead of calling spotify api etc.

PR #3765, opened 2026-07-14
GSOC-26 Musicbrainz collections import for Recording Collections

yateendogra2109:
# Problem

Ticket : [LB-961](https://tickets.metabrainz.org/browse/LB-961)

This pr currently provides Backend for recording collections.

Currently we can import playlists from different external services like Spotify , Apple Music , Soundcloud . But there is no feature to import/view musicbrainz collections as playlists. Users cannot modify collections in Listenbrainz, but can  fetch their tracks for preview / save-as-playlist.

# Solution

Currently added the Backend Support for the Recording Collections . This is the beginning portion of this feature. 
Data is read live from the MusicBrainz Postgres DB via `MB_DATABASE_URI` , not the MusicBrainz web API.  “Save as playlist” stays a later frontend step .
### API endpoints 
   **`GET /1/playlist/import/musicbrainz/collections`**
  - Authorization is required 
  - Lists the authenticated user’s **recording** collections 
  
 **`GET /1/playlist/import/musicbrainz/collections/<collection_mbid>`**
  - Authorization is  optional
  - Returns collection metadata + a paginated track list
  - **Access control:**
    - Public collections → viewable by anyone
    - Private collections → owner only
   
# Tests 

Integration tests are added to check for different scenarios like : 

- list endpoint response shape
- public collection detail without login
- private collection requires login / blocks other users
  

# AI usage

* [ ] I did not use any AI
* [YES ] I have used AI in this PR (add more details below)

Used AI a little for mainly debugging.

#### If you did use AI:
* [ YES ] I understand all the changes made in this PR

### DIFF
--- a/frontend/js/src/user/playlists/components/ImportMusicBrainzCollectionModal.tsx
+++ b/frontend/js/src/user/playlists/components/ImportMusicBrainzCollectionModal.tsx
@@ -1,0 +1,107 @@
+import React from "react";
+import NiceModal, { bootstrapDialog, useModal } from "@ebay/nice-modal-react";
+import { Modal } from "react-bootstrap";
+import { toast } from "react-toastify";
+import { useNavigate } from "react-router";
+
+import GlobalAppContext from "../../../utils/GlobalAppContext";
+import { ToastMsg } from "../../../notifications/Notifications";
+import Loader from "../../../components/Loader";
+import type { MusicBrainzCollectionSummary } from "../../../utils/APIService";
+
+export default NiceModal.create(() => {
+  const modal = useModal();
+  const navigate = useNavigate();
+  const { APIService, currentUser } = React.useContext(GlobalAppContext);
+
+  const [collections, setCollections] = React.useState<
+    MusicBrainzCollectionSummary[]
+  >([]);
+  const [isLoading, setIsLoading] = React.useState(false);
+
+  React.useEffect(() => {
+    async function loadCollections() {
+      if (!currentUser?.auth_token) {
+        toast.error(
+          <ToastMsg
+            title="Error"
+            message="You must be logged in for this operation"
+          />,
+          { toastId: "auth-error" }
+        );
+        return;
+      }
+
+      setIsLoading(true);
+      try {
+        const result = await APIService.getMusicBrainzCollections(
+          currentUser.auth_token
+        );
+        setCollections(result);
+      } catch (error) {
+        toast.error(
+          <ToastMsg
+            title="Error loading collections"
+            message={error?.message ?? error}
+          />,
+          { toastId: "load-collections-error" }
+        );
+      } finally {
+        setIsLoading(false);
+      }
+    }
+
+    loadCollections();
+  }, [APIService, currentUser]);
+
+  const close = () => {
+    modal.hide();
+  };
+
+  const openCollection = (collectionMbid: string) => {
+    close();
+    navigate(`/collection/${collectionMbid}`);
+  };
+
+  return (
+    <Modal
+      {...bootstrapDialog(modal)}
+      title="Import collections from MusicBrainz"
+      aria-labelledby="ImportMBCollectionsModalLabel"
+      id="ImportMBCollectionsModal"
+    >
+      <Modal.Header closeButton>
+        <Modal.Title id="ImportMBCollectionsModalLabel">
+          Import collections from MusicBrainz
+        </Modal.Title>
+      </Modal.Header>
+      <Modal.Body>
+        <div
+          className="list-group"
+          style={{ maxHeight: "50vh", overflow: "auto" }}
+        >
+          {collections.map((collection) => (
+            <button
+              type="button"
+              key={collection.mbid}
+              className="list-group-item list-group-item-action"
+              disabled={isLoading}
+              onClick={() => openCollection(collection.mbid)}
+            >
+              <div style={{ fontWeight: 600 }}>{collection.name}</div>
+              <div className="text-muted">
+                {collection.item_count} recordings
+              </div>
+            </button>
+          ))}
+        </div>
+        <Loader isLoading={isLoading} />
+      </Modal.Body>
+      <Modal.Footer>
+        <button type="button" className="btn btn-secondary" onClick={close}>
+          Close
+        </button>
+      </Modal.Footer>
+    </Modal>
+  );
+});
--- a/frontend/js/src/user/playlists/components/ImportMusicBrainzCollectionModal.tsx
+++ b/frontend/js/src/user/playlists/components/ImportMusicBrainzCollectionModal.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import NiceModal, { bootstrapDialog, useModal } from "@ebay/nice-modal-react";
 import { Modal } from "react-bootstrap";
 import { toast } from "react-toastify";
-import { useNavigate } from "react-router";
 
 import GlobalAppContext from "../../../utils/GlobalAppContext";
 import { ToastMsg } from "../../../notifications/Notifications";
@@ -11,7 +10,6 @@ import type { MusicBrainzCollectionSummary } from "../../../utils/APIService";
 
 export default NiceModal.create(() => {
   const modal = useModal();
-  const navigate = useNavigate();
   const { APIService, currentUser } = React.useContext(GlobalAppContext);
 
   const [collections, setCollections] = React.useState<
@@ -58,11 +56,6 @@ export default NiceModal.create(() => {
     modal.hide();
   };
 
-  const openCollection = (collectionMbid: string) => {
-    close();
-    navigate(`/collection/${collectionMbid}`);
-  };
-
   return (
     <Modal
       {...bootstrapDialog(modal)}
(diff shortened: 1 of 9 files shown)


yateendogra2109:
Also added the MB Collections import modal on top of backend , where one can view list of their Recording collections with number of tracks.

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -1079,6 +1080,86 @@ def import_playlist_from_music_service(service):
...
+        body, code = error
+        if code == 503:
+            raise APIInternalServerError(body.get("error"))
+        if code == 404:
+            raise APINotFound(body.get("error"))
+        if code == 401:
+            raise APIUnauthorized(body.get("error"))
+        if code == 403:
+            raise APIForbidden(body.get("error"))
+        raise APIInternalServerError(body.get("error"))
```
anshg1214: This hand-written code→exception mapping is wrong :

- **`400` has no branch** — invalid UUID, `count <= 0`, `count > 500`, and `offset < 0` all return code `400` from `fetch_collection_payload`, fall through the `if`-chain, and hit the final `raise APIInternalServerError` → HTTP **500**. A client sending a malformed mbid or `?count=1000` gets a 500 instead of a 400.

`APIError` carries its status code as data and the global `@app.errorhandler(APIError)` renders any code, so you don't need the chain at all. 


```suggestion
    if error:
        body, code = error
        raise APIError(body.get("error"), code)
```

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -1079,6 +1080,86 @@ def import_playlist_from_music_service(service):
         raise APIError(error.get("error") or exc.response.reason, exc.response.status_code)
 
 
+@playlist_api_bp.get("/import/musicbrainz/collections")
+@crossdomain
+@ratelimit()
+@api_listenstore_needed
+def import_musicbrainz_collections():
+    """List the authenticated user's MusicBrainz recording collections."""
```
anshg1214, line 1088: The docstring says "recording collections," but the query filters only on `ec.editor` with no `editor_collection_type` join.

A user's release/artist/event/etc. collections all come back too, each with `item_count = 0` (the `LEFT JOIN` on `editor_collection_recording` matches nothing), and clicking one yields an empty detail view.

frontend/js/src/user/playlists/components/ImportMusicBrainzCollectionModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+    loadCollections();
+  }, [APIService, currentUser]);
+
+  const close = () => {
+    modal.hide();
+  };
+
+  const openCollection = (collectionMbid: string) => {
+    close();
+    navigate(`/collection/${collectionMbid}`);
```
anshg1214: This route does not exist right now.

frontend/js/src/user/playlists/components/ImportMusicBrainzCollectionModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+        );
+        return;
+      }
+
+      setIsLoading(true);
+      try {
+        const result = await APIService.getMusicBrainzCollections(
+          currentUser.auth_token
+        );
+        setCollections(result);
```
anshg1214, line 38: When `collections` is empty (loaded, none found) the modal body is blank with just the closed loader. Let's add an empty-state message ("No recording collections found")

listenbrainz/tests/integration/test_musicbrainz_collections_import.py
```diff
@@ -0,0 +1,162 @@
+from unittest import mock
```
anshg1214, line 1: Let's add a path that covers the 400 Status Code path as well

anshg1214 (changes requested):

listenbrainz/webserver/views/collection.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+@collection_bp.get("/", defaults={"collection_mbid": ""})
+@collection_bp.get("/<collection_mbid>/")
+def collection_page(collection_mbid: str):
+    return render_template("index.html")
+
+
+@collection_bp.post("/<collection_mbid>/")
+@web_listenstore_needed
```
anshg1214: We should use `@web_musicbrainz_needed` instead of `@web_listenstore_needed` here as the endpoint serves MB data with no dependency on the TimeScale DB.

listenbrainz/webserver/views/collection.py
```diff
@@ -0,0 +1,153 @@
...
+def load_collection(collection_mbid: str):
```
anshg1214: The decorator stack is missing `@crossdomain` and `@ratelimit()`

listenbrainz/webserver/views/collection.py
anshg1214: Who's calling this route?

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -1079,6 +1080,81 @@ def import_playlist_from_music_service(service):
...
+    user = validate_auth_header()
+
+    if not current_app.config.get("MB_DATABASE_URI"):
+        raise APIInternalServerError("MusicBrainz database is not configured on this server.")
+
+    if not user.get("musicbrainz_row_id"):
+        raise APIInternalServerError("Cannot determine MusicBrainz editor id for this user.")
```
anshg1214: Should be a 400 or 403  with the error message clearly saying `Your ListenBrainz account is not linked to a MusicBrainz account.`

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -1079,6 +1080,81 @@ def import_playlist_from_music_service(service):
...
+@playlist_api_bp.get("/import/musicbrainz/collections/<collection_mbid>")
+@crossdomain
+@ratelimit()
+@api_listenstore_needed
+def import_musicbrainz_collection_detail(collection_mbid):
+    """Fetch a MusicBrainz collection as a read-only track list for preview"""
+    user = validate_auth_header(optional=True)
+    viewer_editor_id = user.get("musicbrainz_row_id") if user else None
+
+    count = get_non_negative_param("count", 100)
```
anshg1214: get_non_negative_param allows count=0 (the name implies ≥0 is valid).

However, fetch_collection_payload immediately rejects it:
     if count <= 0: return None, ({"error": "count must be a positive integer"}, 400)


Instead we can use a get_positive_param helper (or add count > 0 check here before calling fetch_collection_payload)

anshg1214: This is to make the spec visible at the contract level.

listenbrainz/webserver/views/collection.py
```diff
@@ -0,0 +1,153 @@
...
+               , r.name AS title
+               , r.length AS length
+               , ac.name AS artist_credit_name
+            FROM musicbrainz.editor_collection_recording ecr
+            JOIN musicbrainz.recording r
+              ON r.id = ecr.recording
+            JOIN musicbrainz.artist_credit ac
+              ON ac.id = r.artist_credit
+           WHERE ecr.collection = %s
+           ORDER BY ecr.position NULLS LAST, ecr.id
```
anshg1214: `editor_collection_recording` table has no id column, its PK is the composite (collection, recording)

anshg1214 (changes requested):

listenbrainz/webserver/views/collection.py
yateendogra2109: It is called by the React RouteLoader and return the collection payload JSON.

anshg1214: Can you point to the frontend code calling this endpoint?

listenbrainz/webserver/views/playlist_api.py
```diff
@@ -1079,6 +1080,83 @@ def import_playlist_from_music_service(service):
...
+            "item_count": int(row["item_count"] or 0),
+        }
+        for row in collections
+    ])
+
+
+@playlist_api_bp.get("/import/musicbrainz/collections/<collection_mbid>")
+@crossdomain
+@ratelimit()
+@api_listenstore_needed
```
anshg1214: We should use @web_musicbrainz_needed instead of @web_listenstore_needed here as the endpoint serves MB data with no dependency on the TimeScale DB.

listenbrainz/webserver/views/collection.py
anshg1214: @yateendogra2109  How was this code tested if the column did not exist?

yateendogra2109: > @yateendogra2109 How was this code tested if the column did not exist?

@anshg1214 
For local testing I used mocked version of mini MB Postgres with schema written and seeded data , so in schema I had incorrectly written id as column so it didnt showed local error.Also in tests the fetch_collection_payload supplied data without actually running the real querry using mock_fetch

listenbrainz/webserver/views/collection.py
yateendogra2109: > Can you point to the frontend code calling this endpoint?

@anshg1214 
It is not there in this pr , but included in the upcoming pr which shows collection page preview . The pattern is similar to how current playlist page gets data that is from post method in `views/playlist.py` file  and then through Loader getting rendered in `playlist.tsx` file .

However now i will be using the direct` API GET `approach , the route defined in already existing playlist_api.py file and through APIService.ts and finally rendering in` /collection mbid`   route , because virtualization is used so I  think supplying direct chunk through already existing manner will not be beneficial, and direct GET request will be more suitable.

I will remove this post method in the following pr .

listenbrainz/webserver/views/playlist_api.py
yateendogra2109: Since this is a API route so instead of putting @web_musicbrainz_needed , i will be adding @api_musicbrainz_needed in decorators.py that checks IS_MUSICBRAINZ_UP and raises APIServiceUnavailable.

anshg1214 (approved):

PR #3766, opened 2026-07-15
Spotify importer: adapt to refresh tokens expiring after 6 months

MonkeyDo:
Starting next week (July 20th), Spotify will start expiring refresh token after 6 months.
See announcement here: https://developer.spotify.com/blog/2026-06-18-refresh-token-expiration

We were previously keeping the refresh tokens after an invalid_grant response because in the past an incident on their side meant we discarded all user's tokens despite them being valid (we jsut received invalid_grant responses for no reason).

**Every Spotify user's current refresh tokens are going to expire on July 20th.**

With this expiry change, discard the auth row when receiving an invalid_grant, as we expect that to happen for each spotify user every 6 months. Users will have to sign in again manually.
We send an email in such cases, but I've also added an alert on the front-end for visibility and to take the user to the relevant page:
 
Screenshot from 2026-07-14 15-02-34

# AI usage

* [ ] I did not use any AI

Used Codex to implement changes based on my own plan (itself based on Spotify's documentation) and to write tests (i am still not comfortable with the python testing harness).
Lots of hand-holding and supervision and changes and review were involved to keep the code clean and easy to read.

#### If you did use AI:

# Action

We need to write a blog post to announce this change, basically users will have to come back every 6 months to reconnect their spotify account.

We might also want to store a timestamp or a TTL for the refresh tokens. When user successfully logs into Spotify, calculate now + 6 months and store that timestamp in the database (`refresh_token_expires` to mimic `token_expires`?).
Spotify does not provided a timestamp for that, jus the certainty that it will expire in 6 months.

This could allow us to prompt users to reconnect ahead of time to avoid losing any history, and to allow third parties to do the same.

### DIFF
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -59,9 +59,9 @@ def get_refresh_token_expires():
 
 
 def notify_refresh_token_expiry(db_conn):
-    """Send one pre-expiry notification for Spotify refresh tokens expiring within one month."""
-    users = external_service_oauth.get_users_with_expiring_refresh_tokens(db_conn, ExternalServiceType.SPOTIFY)
+    """Send pre-expiry notifications for Spotify refresh tokens expiring within one month and one week."""
     link = current_app.config['SERVER_ROOT_URL'] + '/settings/music-services/details/'
+    users = external_service_oauth.get_users_with_expiring_refresh_tokens(db_conn, ExternalServiceType.SPOTIFY)
     stats = {
         "sent": 0,
         "skipped": 0,
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -50,6 +50,7 @@ SPOTIFY_INVALID_GRANT_ERROR_MESSAGE = (
     "Please reconnect Spotify to resume imports."
 )
 SPOTIFY_REFRESH_TOKEN_TTL = relativedelta(months=6)
+SPOTIFY_REFRESH_TOKEN_EXPIRY_EMAIL_SUBJECT = "[Action required] Reconnect Spotify to keep importing your history"
 
 
 def get_refresh_token_expires():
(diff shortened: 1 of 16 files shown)


MonkeyDo:
The more I think about it, the more it makes sense to me to store a refresh_token_expires timestamp or TTL in order to prompt users to reconnect well in advance to avoid interrumptions, say 1 month in advance.
Then send another email if the token did expire (i.e. no action was taken after first email)

EDIT: Created a branch with this mechanism implemented:
https://github.com/metabrainz/listenbrainz-server/tree/spotify-expiry-advance-notice

EDIT 2: And a PR: #3767

listenbrainz/domain/spotify.py
```diff
@@ -162,6 +166,7 @@ def refresh_access_token(self, user_id: int, refresh_token: str):
             elif response.status_code == 400:
                 error_body = response.json()
                 if "error" in error_body and error_body["error"] == "invalid_grant":
+                    self.revoke_user(user_id)
```
amCap1712, line 221: This method is also called by the web player, playlist APIs, Daily Jams, and not only the importer. If one of those callers encounters the expired token first, only the OAuth row is deleted; the invalid_grant reason and email are added solely by the importer’s handler so the user is never notified or processed again in future.

We need to centralize the email bookkeeping/notification as well in revoke_user.

amCap1712: This needs to be fixed in https://github.com/metabrainz/listenbrainz-server/pull/3767.

MonkeyDo: I thought about this quite a bit, tried a few things to keep better state/lifestyle, but it will require a bunch more changes that may impact other oauth services (changes to the db oauth tables, etc.)

We don't have time to review that properly, so I would suggest this plan:
Merge this PR and 3767, publish a new release, then deploy the release only on the spotify reader container.
Considering it should process all active spotify users in about an hour, after a full cycle we should have set all users in error state. Other existing containers will not revoke the spotify users int he DB.
We then deploy the new image to the prod containers, web and cron specifically are required, which will then send the emails to users to reconnect their account.

What do you think?

PR #3767, opened 2026-07-15
Give Spotify users advance notice of refresh token expiry 

MonkeyDo:
# Problem

Following up from #3766 , we want to notify users ahead of time that they need to reconnect their Spotify account, to avoid disruptions and lost listening history.
This PR adds email notice one month before expiry, and again one week before expiry, if they haven't reconnected yet. This is in addition to existing emails sent when the token has expired.

# AI usage

Used Codex to write most of this PR, with steering, supervision and review, as well as to review both PRs
 
#### If you did use AI:

# Action
I am not sure what action to take for this coming Monday, but in case we want to send everyone an email ahead of time, there is an SQL script (admin/sql/updates/2026-07-15-backfill-spotify-refresh-token-expiry.sql) to set all spotify user's new `refresh_token_expires` column (will be NULL by default) to a timestamp of this Monday July 20th (can be run independently of deployment of the first PR).

If we then run the `notify_spotify_refresh_token_expiry` command, emails should be sent to all spotify users to prompt them to disconnect/reconnect

### DIFF
--- a/listenbrainz/domain/spotify.py
+++ b/listenbrainz/domain/spotify.py
@@ -59,9 +59,9 @@ def get_refresh_token_expires():
 
 
 def notify_refresh_token_expiry(db_conn):
-    """Send one pre-expiry notification for Spotify refresh tokens expiring within one month."""
-    users = external_service_oauth.get_users_with_expiring_refresh_tokens(db_conn, ExternalServiceType.SPOTIFY)
+    """Send pre-expiry notifications for Spotify refresh tokens expiring within one month and one week."""
     link = current_app.config['SERVER_ROOT_URL'] + '/settings/music-services/details/'
+    users = external_service_oauth.get_users_with_expiring_refresh_tokens(db_conn, ExternalServiceType.SPOTIFY)
     stats = {
         "sent": 0,
         "skipped": 0,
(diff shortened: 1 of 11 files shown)


listenbrainz/domain/spotify.py
```diff
@@ -109,7 +160,8 @@ def add_new_user(self, user_id: int, token: dict) -> bool:
 
         external_service_oauth.save_token(db_conn, user_id=user_id, service=self.service, access_token=access_token,
                                           refresh_token=refresh_token, token_expires_ts=expires_at,
-                                          record_listens=active, scopes=scopes, external_user_id=external_user_id)
+                                          record_listens=active, scopes=scopes, external_user_id=external_user_id,
+                                          refresh_token_expires=get_refresh_token_expires())
```
amCap1712, line 164: Hmm, Does automatic refresh extend the expiry by 6 months?

amCap1712: "Refreshing an access token does not reset or extend the refresh token's 6-month lifetime. The lifetime is measured from the moment of the user's original authorization. If the user re-authorizes your app, a fresh 6-month lifetime begins." so nope.

MonkeyDo: Indeed, unfortunately no.

PR #3754, opened 2026-07-04
LB-1008 docs: Document mbid_mapping field in listen JSON API

NAME-ASHWANIYADAV:
# Problem

The `mbid_mapping` property that appears inside `track_metadata` when fetching listens was completely undocumented in the API JSON documentation. This lack of documentation caused confusion regarding what fields are available, how they differ from the user-submitted MBIDs in `additional_info`, and the precedence rules used by the server to resolve these MBIDs. 

Addresses JIRA ticket: [LB-1008](https://tickets.metabrainz.org/browse/LB-1008)

# Solution

Added comprehensive documentation for the `mbid_mapping` element in `docs/users/json.rst`.

The changes include:
- A clear explanation of what `mbid_mapping` is and how it differs from `additional_info`.
- Documentation of the MBID resolution precedence order (User-submitted > User manual > Auto Mapper > Other users' manual mappings).
- A detailed table describing all 8 fields present in the mapping (`recording_mbid`, `recording_name`, `release_mbid`, `artist_mbids`, `artists`, `caa_id`, `caa_release_mbid`, `url_rels`).
- A complete, real-world JSON example demonstrating how both `additional_info` and `mbid_mapping` appear in the API response.

* [x] I have run the code and manually tested the changes (Built the Sphinx RST docs locally to verify formatting and valid JSON blocks).

# AI usage

* [ ] I did not use any AI
* [x] I have used AI in this PR (add more details below)

> I used an AI assistant to help me navigate the codebase, understand the database queries in timescale_listenstore.py, trace the API response flow, and draft the RST documentation/formatting. I made sure to manually verify all the documented fields and the JSON example against the live API (/1/user/rob/listens) and the source code to ensure everything is accurate.

#### If you did use AI:
* [x] I used AI tools for communication
* [x] I used AI tools for coding
* [x] I understand all the changes made in this PR

# Action

1. No specific action needed on merge.

PR #3768, opened 2026-07-19
LB-1217: Add range parameter to listen-count endpoint

NAME-ASHWANIYADAV:
# Problem
This addresses **LB-1217** (Extend `/(user_name)/listen-count` endpoint with range parameter).

Currently the API only exposes a user's overall (all-time) listen count via `GET /1/user/(user_name)/listen-count`. There is no way to get the listen count for a single statistics period (week, month, year, …).

The website already shows this per-period total on the reports page (e.g. `/user/<name>/reports/?range=week`), but it is not directly available via the API. It also cannot be trivially reconstructed from the `listening-activity` endpoint, because that response spans **both** the previous and the current period with no clear boundary between them.

# Solution

**Endpoint (`api.py`):**
* Added an optional `range` query parameter (any of `ALLOWED_STATISTICS_RANGE`, defaults to `all_time`).
* `all_time` / no `range` keeps the existing behaviour exactly — the precise listen count from TimescaleDB — so the change is fully backward compatible.
* For any other valid range, the count is derived from the user's precomputed `listening_activity` statistic and matches the per-period total shown on the website.
* Invalid range returns `400`; a range whose stats haven't been computed yet returns `204` (consistent with the other stats endpoints).
* The range response additionally includes `range`, `from_ts`, `to_ts` and `last_updated`.

**Current-period logic (`stats_utils.py`, new):**
* Added a pure helper `current_period_listen_count()`. The `listening_activity` stat spans two consecutive periods (previous + current) so the site can draw a comparison; the helper sums only the current period's segments.

**Tests:**
* Unit tests (`test_stats_utils.py`) for the helper — week, month, year, quarter, half_yearly, all_time, plus unordered and empty input.
* Integration tests (`test_stats_api.py`) for the endpoint — valid range count, invalid range (`400`) and missing stats (`204`).

**Documentation:**
* Documented the new `range` parameter, the `204`/`400` status codes and an example response in the endpoint docstring, which renders on the API docs page via the existing `autoflask` directive.

# AI usage
* [ ] I did not use any AI

#### If you did use AI:

> The code in this PR (the endpoint changes, the current_period_listen_count helper, and the tests) was written with the help of an AI coding assistant working under my direction. I decided the approach (deriving the per-period count from the existing listening_activity statistic), reviewed every change, and ran the unit tests, integration tests and manual endpoint checks myself to verify the returned output.

# Action
None.

### DIFF
--- a/listenbrainz/tests/integration/test_stats_api.py
+++ b/listenbrainz/tests/integration/test_stats_api.py
@@ -404,6 +404,45 @@ class StatsAPITestCase(IntegrationTestCase):
                                            query_string={'range': range_})
                 self.assertListeningActivityEqual(payload, response)
 
+    def test_listen_count_with_range(self):
+        """ The core listen-count endpoint returns the current-period listen
+        total, derived from the listening_activity stat, when a range is given. """
+        endpoint = "api_v1.get_listen_count"
+
+        # The week fixture (user 1) has 14 daily segments; the current week is
+        # the last 7 days, which sum to 137 (matches the website's total).
+        with open(self.path_to_data_file('user_listening_activity_db_data_for_api_test_week.json')) as f:
+            payload = json.load(f)
+            payload[0]["user_id"] = self.user["id"]
+        db_stats.insert("listening_activity_week_20220718", 0, 5, payload)
+
+        response = self.client.get(
+            self.custom_url_for(endpoint, user_name=self.user['musicbrainz_id']),
+            query_string={'range': 'week'}
+        )
+        self.assert200(response)
+        data = json.loads(response.data)['payload']
+        self.assertEqual(data['count'], 137)
+        self.assertEqual(data['range'], 'week')
+        self.assertEqual(data['from_ts'], 1588550400)  # Monday of the current week
+        self.assertEqual(data['to_ts'], 5)  # stat's to_ts, passed through (test fixtures use 0/5)
+        self.assertIn('last_updated', data)
+
+    def test_listen_count_with_invalid_range(self):
+        response = self.client.get(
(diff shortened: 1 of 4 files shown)


NAME-ASHWANIYADAV:
@MonkeyDo @anshg1214 PTAL !!

MonkeyDo:
After our interraction yesterday on #3754 I'd like to know:
- if you used AI to generate (parts of) the code, please update the description
- did you run the code to manually check for errors and check the returned ourput?

NAME-ASHWANIYADAV:
Hi @MonkeyDo, thanks for the follow-up - both fair questions.

Yes - I've updated the PR description. 
The code in this PR was written with the help of an AI coding assistant working under my direction, so I've ticked both AI checkboxes and described the split accurately. 
Apologies for the earlier wording, which understated the AI's role in the coding.
Yes, I've run the code and checked the returned output myself:
Unit tests: 13 passed - including new tests for this_week/this_month/this_year (with partial current periods) and a check that every StatisticsRange member has a period offset.
Integration tests: test_stats_api.py - 15 passed, covering count=137 for the week fixture, 400 for an invalid range, and 204 when stats haven't been computed.
I also queried the endpoint manually for all 8 ranges against realistic fixture data - every returned count matched the independently computed expected total, and all_time (with or without an explicit range param) returns exactly the same payload shape as before.

NAME-ASHWANIYADAV:
While re-testing I pushed a small follow-up commit : the listenstore availability check now applies only to the all_time path (ranged counts come from precomputed stats, so they stay available if timescale is down), corrected the to_ts value in the docstring example (the stat stores an exclusive midnight boundary), and added the missing unit tests mentioned above.

NAME-ASHWANIYADAV:
@MonkeyDo ptal !!

PR #3770, opened 2026-07-20
Require authentication for popularity API

MonkeyDo:
Due to heavy API scraping from unscrupulous third-parties, we are force to put computationally expensive endpoints behind key.

You need a user account in order to use the popularity endpoint.

See our blog post for more context: https://blog.metabrainz.org/2025/12/11/we-cant-have-nice-things-because-of-ai-scrapers/


anshg1214 (commented):
Heads up - Music Neighbourhood also breaks because of this change.

listenbrainz/webserver/views/popularity_api.py
```diff
@@ -217,6 +241,14 @@ def popularity_artist():
     :statuscode 200: you have data!
     :statuscode 400: invalid artist_mbid(s)
     """
+    # Ensure that the user is passing an auth header
+    try:
```
amCap1712: Put this in the helper method?

amCap1712 (approved):

PR #3771, opened 2026-07-20
feat: Add claim-based work distribution for background task workers

anshg1214:
Replace the plain SELECT with a claim/release pattern using FOR UPDATE SKIP LOCKED so multiple background task workers can run without processing the same task. The claim survives intermediate commits during long-running tasks (e.g. multi-hour file imports).

- claim_task(): CTE locks and marks claimed_at in one quick transaction
- release_task(): clears claimed_at on failure so task is retryable
- remove_task(): deletes on success (unchanged)
- Auto-reclaim after 6 hours for crashed worker recovery
- Exceptions in process_task() now propagate instead of being swallowed

# AI usage

* [ ] I did not use any AI

I've used Cursor while writing code based on the plan I've created.

#### If you did use AI:

# Action

   1. Add tests
   2. Run an SQL update script

### DIFF
--- a/listenbrainz/background/background_tasks.py
+++ b/listenbrainz/background/background_tasks.py
@@ -1,3 +1,4 @@
+import signal
 import time
 
 from flask import current_app
@@ -61,6 +62,9 @@ def remove_task(task):
 
 class BackgroundTasks:
 
+    def __init__(self):
+        self._current_task = None
+
     def process_task(self, task):
         """ Perform the task """
         current_app.logger.info(f"Processing task: {task.id}")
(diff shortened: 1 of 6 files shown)


listenbrainz/background/background_tasks.py
```diff
@@ -8,65 +9,110 @@
...
+            current_app.logger.error(f"Unknown task type: {task}")
+
+    def _release_on_shutdown(self, signum, frame):
+        """ Best-effort release of the current task on SIGTERM (docker stop, deploys). """
+        if self._current_task:
+            try:
+                release_task(self._current_task)
+                current_app.logger.info("Released task %s on shutdown.", self._current_task.id)
+            except Exception:
+                current_app.logger.error("Failed to release task on shutdown:", exc_info=True)
```
MonkeyDo: Would be useful to log the task ID here as well so we can reset the task in question if needs be after reading the logs.

MonkeyDo (commented):
Just one small comment, other than that it looks fine but I haven't tried runniing it locally.

PR #3773, opened 2026-07-22
Spotify imports improvements

MonkeyDo:
# Problem

After the recent changes to the Spotify refresh tokens and in #3766 , some transient Spotify errors (returning 403 "Insufficient scopes" despite actually sufficient scopes) result in non-retriable error state

This is partly because the Spotify importer was not set up to retry on errors (even those marked with retry=true).

Also see #3769 

# Solution
Made the following changes:
- reworded some error messages to make things clearer for users (or align with what is actually happening)
- retry spotify imports with retriable errors: modified ListensImporter base class to allow setting `exclude_error` for the subclassed service
- catch 403 + "insufficient client scope" errors and handle those separately, adding retries and a more specific error message 
- added importer status to the Spotify card in MusicServices page to show errors and status
image

# AI usage

* [x] I did not use any AI

# What happened

After some Spotify API issues, supposedly, we were receiving 403 errors with  but no reason.
AFAICT the scope were fine (I know for myself at the very least, having reconnected just yesterday) hence sending emails to everybody ands stopping processing

One big issue is that SpotifyImporter does not retry accounts in error status despite the error object containing `retry: true`.
This morning (July 22nd) we were processing only 50-60 users, supposedly only the ones that reconnected their account after the spotify incident.

At 13:31 CET I set status and error columns to NULL for all users with this error "Insufficient client scope" with:
```sql
UPDATE public.listens_importer AS li
  SET status = NULL,
      error = NULL
  WHERE error ->> 'message' LIKE '%Insufficient client scope%';
```

This kickstarted the process and all 5902 active Spotify users (did not check count, this is from logs) started being processed again.


listenbrainz/listens_importer/base.py
```diff
@@ -48,8 +48,9 @@ def notify_error(self, musicbrainz_id: str, error: str):
         link = current_app.config['SERVER_ROOT_URL'] + '/settings/music-services/details/'
         text = render_template('emails/listens_importer_error.txt', error=error, link=link)
         send_mail(
-            subject=f'ListenBrainz {self.user_friendly_name} Importer Error',
+            subject=f'[Action required] ListenBrainz {self.user_friendly_name} Importer Error',
             text=text,
+            service=self.service,
```
MonkeyDo: Reaslized this was missing in the email, the template expects a "service" prop

anshg1214 (approved):

PR #3778, opened 2026-07-26
ci: version bumps and zizmor-powered hardening

julian45:
TL;DR: Replicates metabrainz/musicbrainz-server#3777 and metabrainz/musicbrainz-server#3782, but for LB

# Problem

GitHub Actions and their misconfigurations are often a target for threat actors due to a number of security pitfalls, some of which GitHub is aware of and provides best practice guidance for, others are inherent to the architecture of GitHub Actions.

In addition, most if not all of our action `uses` pointed at outdated versions. (A significant portion of the annotations on any given workflow run were due to Node runtime deprecations, for example.)

# Solution

In each commit, this PR introduces a new set of improvements to our GitHub Actions configurations:

- Bump major versions of actions where available (among other things, this fixes the deprecation warnings described above)
- Apply auto-fixes identified by [zizmor](https://zizmor.sh/), a powerful static analysis tool for GitHub Actions (most of the fixes are action version pins, but one disables credential persistence; the latter can be explicitly added back if we figure out we really need it)
- Remediates another set of zizmor findings by disabling all permissions by default at the workflow level (if we find we need them for specific jobs, we can add them back at the per-job level)
- Adds a new workflow designed to run zizmor and report back findings
- Adds a Dependabot config for GitHub Actions only (since versions are now pinned, it's good to keep them up to date, [albeit with a cooldown period](https://docs.zizmor.sh/audits/#dependabot-cooldown))

# AI usage

* [x] I did not use any AI
* [ ] I have used AI in this PR ~~(add more details below)~~

# Further action

It's possible that zeroing out most permissions by default may be too restrictive. I found one instance where a workflow had certain required permissions (namely, in the release drafter) and added those, but that may not be all that's needed to keep some workflows functioning.

The zizmor-based fixes here are based on the default ["persona"](https://docs.zizmor.sh/usage/#using-personas). There are more findings in this repo emitted under stricter personas that are probably worth at least glancing at if not also remediating, but I wanted to at least give default-persona changes a go to see if, e.g., we need to add back credential persistence or job-level permissions before going stricter.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
--- a/.github/workflows/build-prod-image.yml
+++ b/.github/workflows/build-prod-image.yml
@@ -1,4 +1,5 @@
 name: ListenBrainz Build Production Image
+permissions: {}
 
 on:
   push:
--- a/.github/workflows/build-prod-image.yml
+++ b/.github/workflows/build-prod-image.yml
@@ -28,7 +28,7 @@ jobs:
       if: github.event.pull_request.draft == false
 
       steps:
-      - uses: actions/checkout@v4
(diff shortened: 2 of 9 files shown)


amCap1712 (approved):

amCap1712:
LGTM, Thanks!

PR #3784, opened 2026-07-29
Split similar users data into multiple messages

amCap1712:
The spark job works but the message produced by it greater than the configured rabbitmq limit, hence chunk the similar users data into multiple messages in a similar way as stats.


amCap1712:
Test failure unrelated.

PR #3761, opened 2026-07-10
LB-580: Document and test /player POST endpoints

NAME-ASHWANIYADAV:
# Problem

This addresses [LB-580](https://tickets.metabrainz.org/browse/LB-580) (Document and test `/player` POST endpoint).

The `/player` blueprint endpoints (`load_instant` and `load_release`) had missing and inaccurate tests. 
The existing tests in `test_player.py` used `GET` requests instead of `POST` on these routes. Because there is a catch-all `GET /<path:path>/` route that serves the `index.html` React SPA, the tests were succeeding by rendering the index page instead of actually testing the `POST` endpoints.
Furthermore, the endpoints lacked proper documentation for Sphinx `autoflask`.

# Solution

1. **Test Rewrite (`test_player.py`)**:
   - Switched all relevant endpoints to use `client.post()` instead of `client.get()`.
   - Added comprehensive tests for `load_instant` and `load_release` to verify correct JSPF payload generation under different conditions.
   - Added tests for various error cases, including missing parameters and invalid UUIDs (verifying `400` and `404` responses).
   - Mocked MusicBrainz `mb_engine` and DB queries to properly test the `load_release` endpoint behavior.

2. **Documentation (`player.py` and `docs/`)**:
   - Improved docstrings for `load_instant` and `load_release` with standard Sphinx directives (`:param:`, `:statuscode:`, `:resheader:`).
   - Added a new Sphinx docs page `docs/users/api/player.rst` utilizing `autoflask` to auto-generate the API docs.
   - Added `player` to the Sphinx documentation toctree in `docs/users/api/index.rst`.

* [x] I have run the code and manually tested the changes

# AI usage

* [ ] I did not use any AI
* [x] I have used AI in this PR (add more details below)

#### If you did use AI:
* [x] I used AI tools for communication
* [ ] I used AI tools for coding
* [x] I understand all the changes made in this PR

> Used an AI coding assistant to understand the existing codebase, identify the GET/POST routing issue, and structure the approach. However, all the actual coding and testing was done manually by me.

# Action
None.

PR #3786, opened 2026-07-29
LB-1450:Add API for querying top composers and writers per user

NAME-ASHWANIYADAV:
## Problem

There is no way to see statistics about the songwriters, composers, and lyricists behind the music a user listens to.

Fixes [LB-1450](https://tickets.metabrainz.org/browse/LB-1450)

## Solution

Adds a new `writers` stats entity end to end, following the exact pattern of the existing entity stats (artists, recordings, releases, release_groups):

* **Data model**: new `WriterRecord` Pydantic model (`writer_name`, `writer_mbid`, `listen_count`), added to the `EntityRecord` union.
* **MusicBrainz cache query**: traverses `recording → l_recording_work (performance) → work → l_artist_work (composer/lyricist/writer) → artist`. Uses `SELECT DISTINCT` so an artist credited under multiple roles on the same work (e.g. both composer and lyricist) is only counted once per listen.
* **Spark pipeline**: `WriterUserEntity` incremental stats provider plus the HDFS cache importer, registered in `entity_model_map` and `incremental_entity_map`.
* **Scheduling**: `writers` added to `cron_request_all_stats` and the `request_user_stats` CLI, so the scheduled stats job computes writers stats for every range.
* **API**: `GET /1/stats/user/<user_name>/writers` with `count`, `offset` and `range` params — 204 when stats haven't been computed yet, 400 for an invalid range, 404 for unknown users. Docs render via the existing autoflask directive.
* **Tests**: JSON fixtures for all ranges plus a `too_many` cap case; `writers` is registered in the shared `entity_endpoints` matrix, so every existing entity subtest also runs against it (default response, count, offset, cap, per-range, 204, 400, 404).

Only listens with a mapped `recording_mbid` whose recording has work→artist relationships contribute, and each listen counts once for every writer of the recording's linked work(s) — this is documented in the endpoint docstring.

### Testing

`test_stats_api.py` + `test_request_manage.py` run locally in the Docker test environment — **15 passed, 117 subtests passed** (all writers subtests included). The endpoint was also exercised manually: 404 for an unknown user, 204 before stats are computed.

Screenshots of the local test run:

Screenshot 2026-08-03 030728
Screenshot 2026-08-03 030750

- [x] I have run the code and manually tested the changes

## AI usage

- [x] I have used AI in this PR (add more details below)
- [x] I used AI tools for communication

The code in this PR was written with AI coding assistants working under my direction — Google Gemini (Antigravity) for the initial implementation (as disclosed in the first commit message) and Claude Code for the follow-up review and fixes commit. The AI helped analyse the existing entity-stats pipeline, implement the writers entity following that pattern, draft the MusicBrainz SQL, build test fixtures, and review the changes. I decided the approach, reviewed every change, and ran the test suites and manual endpoint checks myself to verify the returned output.

## Action

None on merge. `request_import_pg_tables` (already invoked by `cron_request_all_stats`) imports the new recording-writer cache, and the next scheduled stats run will compute and populate writers stats automatically.


NAME-ASHWANIYADAV:
While re-testing this PR locally, I found and fixed two issues in a follow-up commit (59fd65e):

cron_request_all_stats and the request_user_stats CLI didn't include the new writers entity, so the scheduled stats job would never have computed writers stats and the endpoint would always have returned 204 in production.
Added DISTINCT to the recording-writer cache query - an artist credited as both composer and lyricist of the same work produced duplicate cache rows, which would have double-counted that writer's listens.
Verified locally: the stats API integration tests (including all writers subtests) and request_manage tests pass. The earlier CI failure was the known-flaky test_import_spotify (AssertionError: 4 != 6, same as on [#3761](https://github.com/metabrainz/listenbrainz-server/pull/3761)), unrelated to this PR.

NAME-ASHWANIYADAV:
@MonkeyDo @anshg1214 have a look !!

NAME-ASHWANIYADAV:
@MonkeyDo @anshg1214 PTAL !!

PR #3789, opened 2026-07-30
feat(radio): replace LB Radio HTTP loopbacks with direct DB service calls

anshg1214:
# Problem

LB Radio generation fires HTTP loopback calls back to the same Flask server it is running inside. Each loopback occupies an extra uWSGI thread for the duration of the DB query it triggers. Under load, this causes thread starvation — radio requests hold threads waiting for their own loopbacks, which can't be served because all threads are busy. Disabling the LB Radio endpoint reduces site load noticeably, confirming this is the root cause.

For a typical 3-element radio prompt (`artist:X tag:Y stats:user`), up to 5 uWSGI threads are held simultaneously for a single browser request.

# Solution

Introduce `listenbrainz/radio/` — a module of Troi service implementations that replace HTTP loopbacks with direct Python function calls.

**Three services added:**

- `RecordingSearchByArtistService` — replaces `GET /1/lb-radio/artist/<mbid>` with a direct call to `db.lb_radio_artist()`. Registered via Troi's existing `patch.register_service()` hook.

- `RecordingSearchByTagService` — replaces `GET /1/lb-radio/tags` with a direct call to `db.tags.get_and()` / `db.tags.get_or()`.

- `RecordingLookupService` — replaces `POST /1/metadata/recording` with a direct call to `db.metadata.get_metadata_for_recording()`. Requires a one-line change in Troi (see Troi PR) that checks for this service before making the HTTP call.

All three services return the exact same Python objects as the HTTP path — same function, same data, same Troi object construction. 1:1 parity verified by dedicated parity tests in `listenbrainz/radio/tests/test_parity.py`.

**After this change:** a 3-element prompt uses 1 uWSGI thread instead of up to 5.

**What is unchanged:** All public endpoints (`/1/lb-radio/artist/<mbid>`, `/1/lb-radio/tags`, `/1/metadata/recording`) are untouched and continue to work for external callers. The Troi package fallback is fully backwards compatible — if our services are not registered, Troi uses HTTP as before.

**Troi dependency:** `requirements.txt` is pinned to commit `3483f8e` of `troi-recommendation-playground` which includes the `RecordingLookupElement` service hook. See Troi PR: metabrainz/troi-recommendation-playground#189.

# AI usage

Used Claude Code to assist with implementation, test writing, and SQL analysis throughout this PR. All code was reviewed, understood, and validated against the existing codebase before committing.

# Action

The Troi PR (metabrainz/troi-recommendation-playground#189) must be merged and released before `requirements.txt` can be updated to a stable version pin. For now it is pinned to commit `3483f8e`.

### DIFF
--- a/listenbrainz/radio/artist.py
+++ b/listenbrainz/radio/artist.py
@@ -1,0 +1,47 @@
+from troi import Recording, Artist, ArtistCredit
+from troi.plist import plist
+from troi.service import Service
+
+from listenbrainz.db.lb_radio_artist import lb_radio_artist
+
+
+class RecordingSearchByArtistService(Service):
+    """Replaces troi.recording_search_service.RecordingSearchByArtistService.
+
+    Calls lb_radio_artist() directly instead of the HTTP loopback to
+    /1/lb-radio/artist/<mbid>. Object construction and random_item call are
+    copied verbatim from the troi original so return shape is identical.
+    """
+
+    SLUG = "recording-search-by-artist"
+
+    def __init__(self):
+        super().__init__(self.SLUG)
+
+    def search(self, mode, artist_mbid, pop_begin, pop_end, max_recordings_per_artist, max_similar_artists):
+        artists = lb_radio_artist(
+            mode,
+            artist_mbid,
+            max_similar_artists,
+            max_recordings_per_artist,
+            pop_begin / 100.0,
+            pop_end / 100.0,
+        )
+
+        artist_recordings = {}
+        msgs = []
+        for artist_mbid in artists:
+            recordings = plist()
+            for recording in artists[artist_mbid]:
+                artist_credit = ArtistCredit(artists=[Artist(mbid=recording["similar_artist_mbid"])],
+                                             name=recording["similar_artist_name"])
+                recordings.append(Recording(mbid=recording["recording_mbid"],
+                                            artist_credit=artist_credit,
+                                            musicbrainz={"total_listen_count": recording["total_listen_count"]}))
+
+            if 0 < len(recordings) < max_recordings_per_artist - 1:
+                msgs.append("Artist %s has only few top recordings in %s mode" % (recordings[0].artist_credit.name, mode))
+
+            artist_recordings[artist_mbid] = recordings.random_item(pop_begin, pop_end, max_recordings_per_artist)
+
+        return artist_recordings, msgs
--- a/listenbrainz/radio/artist.py
+++ b/listenbrainz/radio/artist.py
@@ -39,6 +39,7 @@ class RecordingSearchByArtistService(Service):
                                             artist_credit=artist_credit,
                                             musicbrainz={"total_listen_count": recording["total_listen_count"]}))
 
+            # Below is a hack, since the endpoint seems to return one track too few
             if 0 < len(recordings) < max_recordings_per_artist - 1:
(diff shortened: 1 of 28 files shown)


listenbrainz/radio/artist.py
```diff
@@ -0,0 +1,47 @@
...
+        for artist_mbid in artists:
+            recordings = plist()
+            for recording in artists[artist_mbid]:
+                artist_credit = ArtistCredit(artists=[Artist(mbid=recording["similar_artist_mbid"])],
+                                             name=recording["similar_artist_name"])
+                recordings.append(Recording(mbid=recording["recording_mbid"],
+                                            artist_credit=artist_credit,
+                                            musicbrainz={"total_listen_count": recording["total_listen_count"]}))
+
+            if 0 < len(recordings) < max_recordings_per_artist - 1:
```
MonkeyDo, line 43: Might be good to keep the comment that was there for context on what is happening here

```suggestion
            # Below is a hack, since the endpoint seems to return one track too few
            if 0 < len(recordings) < max_recordings_per_artist - 1:
```

listenbrainz/webserver/views/explore_api.py
```diff
@@ -195,6 +198,9 @@ def lb_radio():
             "min_recordings": 1,
             "auth_token": auth_token
         })
+        patch.register_service(RecordingSearchByArtistService())
```
MonkeyDo, line 202: This might get confusing with the same names as the ones in troi, suggesting a comment:
```suggestion
        # Register local patches pulling directly from the DB instead of using HTTP API calls
        patch.register_service(RecordingSearchByArtistService())
```

listenbrainz/radio/recording_lookup.py
```diff
@@ -0,0 +1,36 @@
...
+            data = {"recording": entry.recording_data}
+            if "artist" in incs:
+                data["artist"] = entry.artist_data
+            if "tag" in incs:
+                data["tag"] = entry.tag_data
+            if "release" in incs:
+                data["release"] = entry.release_data
+            result[str(entry.recording_mbid)] = data
+
+        return result
```
MonkeyDo: Can this `return fetch_metadata(recording_mbids, inc)` instead?
https://github.com/metabrainz/listenbrainz-server/blob/5da800ce76dad39982c303a9b07a987f04cd4208/listenbrainz/webserver/views/metadata_api.py#L44-L60

MonkeyDo:
P.S: Great :) This is going to improve the situation a lot, let's see if it will be enough for us to reopen the LBRadio endpoints :crossed_fingers:

listenbrainz/radio/artist.py
anshg1214: Fixed — comment restored.

listenbrainz/webserver/views/explore_api.py
anshg1214: Fixed — comment added.

listenbrainz/radio/recording_lookup.py
anshg1214: Fixed — `lookup()` now delegates to `fetch_metadata(recording_mbids, inc.split())`. Dropped the local imports and the duplicated body.

MonkeyDo (approved):
Great, thank you!
Looking forward to opening this feature back up.

Deployed to beta.LB for manual testing in situ, working fine.

PR #3792, opened 2026-08-01
perf: parallelize feed event queries and fix entity page connection pooling

anshg1214:
# Problem

Two unrelated issues were both wasting WSGI threads and DB connections:

1. The `/user/<user>/feed/events` endpoint runs 7 independent DB queries back to back inside `get_feed_events_for_user`. Each query waits for the previous one to finish even though none of them depend on each other. The thread is held for the sum of all 7 instead of the slowest one.

2. The recording entity page (`/track/<mbid>/`) was opening 5 raw `psycopg2.connect()` calls per request -- 3 separate connections to the MusicBrainz DB and 2 to TimescaleDB -- all bypassing the SQLAlchemy connection pool. Each one is a fresh TCP handshake that gets discarded after a single use.

# Solution

**Feed endpoint:** Replaced the sequential calls in `get_feed_events_for_user` with a `ThreadPoolExecutor`. All 7 event type queries now run concurrently. Each thread gets its own Flask app context so `db_conn`/`ts_conn` resolve correctly against the pool. Failures in individual sub-queries are caught per thread and log an error rather than crashing the whole feed response.

**Entity page connections:**
- `recording_entity`: consolidated from 5 raw connections to 1 shared MB connection + the pooled `ts_conn` for all TimescaleDB work. The popularity counts call now uses `ts_conn` directly, matching the pattern already used in `album_entity`.
- `artist_entity`: removed a redundant `get_top_recordings_for_artist(count=1)` call that was only used to get the top recording's color. That value is already in `popular_recordings[0]` which was fetched earlier in the same request with `count=10`.

# AI usage

Used Claude Code to assist with investigation, implementation, and analysis throughout this PR. All changes were reviewed and validated against the codebase.

### DIFF
--- a/listenbrainz/webserver/views/entity_pages.py
+++ b/listenbrainz/webserver/views/entity_pages.py
@@ -191,7 +191,7 @@ def artist_entity(artist_mbid: str):
         top_release_group_color = None
 
     try:
-        top_recording_color = popularity.get_top_recordings_for_artist(db_conn, ts_conn, artist_mbid, 1)[0]["release_color"]
+        top_recording_color = popular_recordings[0]["release_color"]
     except IndexError:
         top_recording_color = None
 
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/entity_pages.py
```diff
@@ -383,48 +383,45 @@ def recording_entity(recording_mbid: str):
...
-            psycopg2.connect(current_app.config["SQLALCHEMY_TIMESCALE_PGBOUNCER_URI"]) as ts_conn, \
-            mb_conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as mb_curs, \
-            ts_conn.cursor(cursor_factory=psycopg2.extras.DictCursor) as ts_curs:
-        recording_data = load_recordings_from_mbids_with_redirects(mb_curs, ts_curs, [recording_mbid])
-    if recording_data is None or len(recording_data) == 0 or recording_data[0].get("recording_mbid") is None:
-        return jsonify({"error": f"Recording {recording_mbid} not found in the metadata cache"}), 404
-
-    recording_data = recording_data[0]
-
+    mb_conn = psycopg2.connect(current_app.config["MB_DATABASE_URI"])
```
MonkeyDo, line 386: /me wonders if we should have a `_get_mb_conn` helper like we have for others, in listenbrainz/webserver/init

listenbrainz/webserver/views/user_timeline_event_api.py
```diff
@@ -794,69 +795,42 @@ def get_feed_events_for_user(
...
+        "pin": in_context(get_recording_pin_events, users_for_events=users_for_events, min_ts=min_ts, max_ts=max_ts, count=count),
+    }
+
+    all_events = []
+    with ThreadPoolExecutor(max_workers=len(tasks)) as executor:
+        futures = {executor.submit(fn): name for name, fn in tasks.items()}
+        for future in as_completed(futures):
+            name = futures[future]
+            try:
+                all_events.extend(future.result())
```
MonkeyDo, line 828: I know this is not your choice, but this is hard to read.

Reminds me of JS promises before ES6 (and the syntax sucks even now with concurrent promises...)

I see utilities like [this](https://github.com/python-useful-helpers/threaded) but I'm not even sure it would help much in this situation.

Anyway, no suggestion, just a rant.

MonkeyDo (approved):
Looks good!

I deployed to beta.LB for manual testing, loaded feed events just fine.
Didn't notice much of a difference in terms of page load times, maybe a tad faster, but it certainly will make a difference to use the connection pool instead of 5 connections.

PR #3794, opened 2026-08-03
GSOC-26 Musicbrainz Collections Import for Recording Collections - Frontend

yateendogra2109:
# Problem

We had already wired the Backend portion for the Musicbrainz Recording collections. We can see the list of the user owned recording collections. This pr is built on top of that and allows users to preview the collection and save it as LB style playlist.
# Solution

The preview page is created in the `frontend\js\src\collections\Collection.tsx` file , it gets the Collection data from the getMusicBrainzCollectionDetail() which calls the fetch_collection_payload() and runs the query to the MB DB directly . Virtualization is used to display the tracks with a request of 100 tracks per call . To save as playlist , 500 tracks are called in one request.

* [ YES ] I have run the code and manually tested the changes

# AI usage

* [ ] I did not use any AI
* [ YES ] I have used AI in this PR (add more details below)

Used claude to add packages , integrate the virtualization in collection.tsx file and for debugging.

#### If you did use AI:
* [ YES ] I used AI tools for coding
* [ YES ] I understand all the changes made in this PR

# Action

   1. Add tests
   2. Run an SQL update script

### DIFF
--- a/frontend/js/src/collections/Collection.tsx
+++ b/frontend/js/src/collections/Collection.tsx
@@ -1,0 +1,289 @@
+import * as React from "react";
+
+import { Helmet } from "react-helmet";
+import { useParams } from "react-router";
+import { toast } from "react-toastify";
+import { useWindowVirtualizer } from "@tanstack/react-virtual";
+import type { VirtualItem } from "@tanstack/react-virtual";
+
+import GlobalAppContext from "../utils/GlobalAppContext";
+import { ToastMsg } from "../notifications/Notifications";
+import Loader from "../components/Loader";
+import PlaylistItemCard from "../playlists/components/PlaylistItemCard";
+import { PLAYLIST_TRACK_URI_PREFIX } from "../playlists/utils";
+
+import type {
+  MusicBrainzCollectionDetailResponse,
+  MusicBrainzCollectionTrack,
+} from "../utils/APIService";
+
+const DEFAULT_PAGE_SIZE = 100;
+const DEFAULT_ESTIMATED_ROW_HEIGHT_PX = 110;
+const LOAD_MORE_THRESHOLD_ROWS = 20;
+
+function asJSPFTrack(track: MusicBrainzCollectionTrack): JSPFTrack {
+  const recordingMBID = track.recording_mbid;
+  return {
+    identifier: `${PLAYLIST_TRACK_URI_PREFIX}${recordingMBID}`,
+    id: recordingMBID,
+    title: track.title ?? "",
+    creator: track.artist_credit_name ?? "",
+    duration: track.length ?? undefined,
+  };
+}
+
+export default function CollectionPage() {
+  const { currentUser, APIService } = React.useContext(GlobalAppContext);
+  const { collectionMBID } = useParams();
+
+  const [collection, setCollection] = React.useState<
+    MusicBrainzCollectionDetailResponse["collection"] | null
+  >(null);
+  const [tracks, setTracks] = React.useState<JSPFTrack[]>([]);
+  const [trackCount, setTrackCount] = React.useState<number>(0);
+  const [isLoadingInitial, setIsLoadingInitial] = React.useState(false);
+  const [isLoadingMore, setIsLoadingMore] = React.useState(false);
+  const [hasMore, setHasMore] = React.useState(true);
+
+  const totalRows = tracks.length + (hasMore ? 1 : 0); // +1 row for loader
+  const rowVirtualizer = useWindowVirtualizer({
+    count: totalRows,
+    estimateSize: () => DEFAULT_ESTIMATED_ROW_HEIGHT_PX,
+    // PlaylistItemCard height varies depending on metadata/menu.
+    // Measure real heights to avoid visual gaps between rows since virtualization is used.
+    measureElement: (el) => el.getBoundingClientRect().height,
+    overscan: 6,
+  });
+  const virtualItems = rowVirtualizer.getVirtualItems();
+
+  const loadPage = React.useCallback(
+    async (offset: number) => {
+      if (!collectionMBID) {
+        return undefined;
+      }
+      const userToken = currentUser?.auth_token;
+      const result = await APIService.getMusicBrainzCollectionDetail(
+        collectionMBID,
+        userToken,
+        DEFAULT_PAGE_SIZE,
+        offset
+      );
+      return result;
+    },
+    [APIService, collectionMBID, currentUser?.auth_token]
+  );
+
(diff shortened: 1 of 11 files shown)


yateendogra2109:
Hi @ansh 
In this pr I had built the basic collection preview page with save as playlist option.
I will add the additional metadata and caa to the tracks in the upcomming commits .
Also the server error is unrelated to this pr.

yateendogra2109:
Added the cover art and additional data to tracks . I had tested the cover art feature by seeding the caa ids to respective tracks.Attached is the display screen of the preview page .


image

frontend/js/src/collections/Collection.tsx
```diff
@@ -0,0 +1,551 @@
...
+    MusicBrainzCollectionDetailResponse["collection"] | null
+  >(null);
+  const [tracks, setTracks] = React.useState<JSPFTrack[]>([]);
+  const [trackCount, setTrackCount] = React.useState<number>(0);
+  const [isLoadingInitial, setIsLoadingInitial] = React.useState(false);
+  const [isLoadingMore, setIsLoadingMore] = React.useState(false);
+  const [hasMore, setHasMore] = React.useState(true);
+  const [isSaving, setIsSaving] = React.useState(false);
+  const [isPlayingAll, setIsPlayingAll] = React.useState(false);
+  const [coverArt, setCoverArt] = React.useState<string | null>(null);
```
anshg1214: Few questions: Do we even need 9 States to manage the data?

frontend/js/src/collections/Collection.tsx
```diff
@@ -0,0 +1,551 @@
...
+      const result = await APIService.getMusicBrainzCollectionDetail(
+        collectionMBID,
+        userToken,
+        DEFAULT_PAGE_SIZE,
+        offset
+      );
+      return result;
+    },
+    [APIService, collectionMBID, currentUser?.auth_token]
+  );
```
anshg1214: Can we use props and loader instead?

frontend/js/src/collections/Collection.tsx
```diff
@@ -0,0 +1,551 @@
...
+    tracks.length,
+    virtualItems,
+  ]);
+
+  const title = collection?.name ?? "MusicBrainz collection";
+  const mbUrl = collectionMBID
+    ? `https://musicbrainz.org/collection/${collectionMBID}`
+    : undefined;
+
+  const playAllTracks = React.useCallback(async () => {
```
anshg1214, line 173: I believe we can reuse `fetchAllCollectionTracks` function here.

anshg1214 (changes requested):
Looks good overall. 

QQ: What happens if a playlist owner calls POST /1/playlist/edit/<mbid> with 
```
{
  "playlist": {
    "extension": {
      "https://...": {
        "additional_metadata": {
          "tags": ["this can be anything", "no length limit", "no count limit"]
        }
      }
    }
  }
}
```

Let's evaluate the behaviour end to end.

yateendogra2109:
> Looks good overall.
> 
> QQ: What happens if a playlist owner calls POST /1/playlist/edit/ with
> 
> ```
> {
>   "playlist": {
>     "extension": {
>       "https://...": {
>         "additional_metadata": {
>           "tags": ["this can be anything", "no length limit", "no count limit"]
>         }
>       }
>     }
>   }
> }
> ```
> 
> Let's evaluate the behaviour end to end.

I tested it and you are right to point out this bug. If user directly add tags through this endpoint then they will be written in the playlist.playlist.additional_metadata. without validating/filtering  tags .If the playlist already has no tags in the playlist.playlist_tag table then these additional_metadata tags will be visible in the frontend also . However if playlist.playlist_tag table  already has tags then they will not be visible in frontend. So to stop this from happening , if user tries to put tags in additional_metadata then i will restrict that or strip only the tags section , So now when user edits tags from UI , the same endpoints  POST /1/playlist/<mbid>/tags     , DELETE /1/playlist/<mbid>/tags/<tag> will be called for add and delete

frontend/js/src/collections/Collection.tsx
yateendogra2109: Yes previously I had thought of using the loader as used in playlists , but collections can be huge and instead of posting the whole page data at once it is better to use API GET with offset for proper pagination.So I think the current choice will be better.

anshg1214: We can use the props in a similar fashion and have pagination implemented there as well. Router Loaders provide a cleaner way to fetch data and with those you don't need to use callbacks and useeffect to load the data and save them in states.

frontend/js/src/collections/Collection.tsx
```diff
@@ -132,65 +162,17 @@ export default function CollectionPage() {
...
-    setTrackCount(0);
-    loadInitial();
-  }, [collectionMBID, loadPage]);
+    setCollection({
+      ...loaderData.collection,
+      cover_art: loaderData.cover_art ?? null,
+    });
+    setTrackCount(loaderData.track_count);
+    setTracks((loaderData.tracks ?? []).map(asJSPFTrack));
+  }, [loaderData]);
```
anshg1214: Since these are props now, do we even need states?

anshg1214 (changes requested):

frontend/js/src/collections/Collection.tsx
```diff
@@ -0,0 +1,497 @@
...
+        Accept: "application/json",
+      },
+    }
+  );
+  const data = await response.json();
+  if (!response.ok) {
+    throw new Error(data?.error ?? response.statusText);
+  }
+  return data;
+}
```
anshg1214: How about we use infiniteQuery here?


```
const { data, fetchNextPage, hasNextPage, isFetchingNextPage } = useInfiniteQuery({
  queryKey: ["collection", collectionMBID],
  queryFn: ({ pageParam = 0 }) =>
    queryClient.fetchQuery(
      RouteQuery(
        ["collection", collectionMBID, pageParam],
        `/collection/${collectionMBID}/?count=100&offset=${pageParam}`
      )
    ),
  getNextPageParam: (lastPage, pages) =>
    pages.flatMap(p => p.tracks).length < lastPage.track_count
      ? pages.flatMap(p => p.tracks).length
      : undefined,
  initialPageParam: 0,
});
```

frontend/js/src/collections/Collection.tsx
yateendogra2109: Removed these as states , kept only tracks as a state , since tracks get updated on scrolling and on new collection so running .

anshg1214 (approved):

PR #3797, opened 2026-08-04
LB-2023: Allow users to disable display/animation of other users' flair

ghost:
# Problem

[LB-2023](https://tickets.metabrainz.org/browse/LB-2023)

There is currently no way for users to control how other users' flair badges and
username animations are displayed across the site. Some users find the animated
flair distracting, and others may prefer not to see it at all for accessibility
or personal preference reasons. The only existing workaround is a uBlock Origin
custom filter, which is not practical for most users.

# Solution

I added a display preference dropdown under Settings > Flair Settings with three
options:

- Show all flairs and animations (default, existing behavior unchanged)
- Disable flair animations — flair badges still render but all CSS animations and
  transitions are suppressed
- Hide all flairs entirely — usernames render as plain text with no badge or styling

The preference is stored in localStorage so it persists across sessions without
requiring any server-side changes or new API endpoints. Changing the setting
updates the context immediately and all username components across the page reflect
it without a reload.

Files changed:

- `frontend/js/src/utils/constants.ts` — added `FlairDisplayPreferenceEnum` and the `FlairDisplayPreference` type
- `frontend/js/src/utils/GlobalAppContext.tsx` — added `flairDisplayPreference` and `setFlairDisplayPreference` to the context
- `frontend/js/src/index.tsx` — wired up proper React state for the preference at the app root so changes propagate to all consumers
- `frontend/js/src/common/Username.tsx` — reads the preference from context and conditionally skips the flair class or applies a no-animation class
- `frontend/js/src/common/flairs/FlairsExplanationButton.tsx` — hides itself when flairs are not fully shown
- `frontend/js/src/settings/flairs/FlairsSettings.tsx` — added the dropdown UI for the preference
- `frontend/css/sass/flairs.scss` — added `.flair-no-animation` to suppress all animation and transition properties

# Action

No deployment or migration steps needed. All changes are client-side and use
localStorage for persistence.

### DIFF
--- a/frontend/js/src/index.tsx
+++ b/frontend/js/src/index.tsx
@@ -11,11 +11,52 @@ import {
 } from "react-router";
 import { Helmet } from "react-helmet";
 import ErrorBoundary from "./utils/ErrorBoundary";
-import GlobalAppContext from "./utils/GlobalAppContext";
+import GlobalAppContext, { GlobalAppContextT } from "./utils/GlobalAppContext";
 import { getPageProps } from "./utils/utils";
 import getRoutes from "./routes/routes";
 import queryClient from "./utils/QueryClient";
 import ReactQueryDevtool from "./utils/ReactQueryDevTools";
+import {
+  FlairDisplayPreference,
+  FlairDisplayPreferenceEnum,
+} from "./utils/constants";
+
+function AppRoot({
+  globalAppContext,
+  router,
+}: {
+  globalAppContext: GlobalAppContextT;
+  router: ReturnType<typeof createBrowserRouter>;
+}) {
+  const [flairDisplayPreference, setFlairDisplayPreference] = React.useState<
+    FlairDisplayPreference
+  >(
+    () =>
+      (localStorage.getItem(
+        "lb_flair_display_preference"
+      ) as FlairDisplayPreference) || FlairDisplayPreferenceEnum.All
+  );
+
+  const contextValue: GlobalAppContextT = React.useMemo(
+    () => ({
+      ...globalAppContext,
+      flairDisplayPreference,
+      setFlairDisplayPreference,
+    }),
+    [globalAppContext, flairDisplayPreference]
+  );
+
+  return (
+    <ErrorBoundary>
+      <GlobalAppContext.Provider value={contextValue}>
+        <Helmet defaultTitle="ListenBrainz" titleTemplate="%s - ListenBrainz" />
+        <ReactQueryDevtool client={queryClient}>
+          <RouterProvider router={router} />
+        </ReactQueryDevtool>
+      </GlobalAppContext.Provider>
+    </ErrorBoundary>
+  );
+}
 
 document.addEventListener("DOMContentLoaded", async () => {
   const {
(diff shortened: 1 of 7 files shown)


kepstin:
> The preference is stored in localStorage so it persists across sessions without requiring any server-side changes or new API endpoints.

Ideally the preference should be saved server-side, I would like to keep it in sync across multiple devices and be restored if I log in on a new device (or on the same device if cookies/local storage get cleared)

ghost:
Thanks for the feedback @kepstin! That makes complete sense for cross-device persistence.

My goal with this PR was to lay the client-side implementation and CSS groundwork to handle disabling/freezing flairs and respecting prefers-reduced-motion.

Would it be best to merge this as Phase 1 (client-side/localStorage foundation) and open a follow-up issue/PR to persist this setting in the user account backend, or would you prefer I add the server-side preference field directly into this PR?

frontend/js/src/settings/flairs/FlairsSettings.tsx
```diff
@@ -53,6 +58,26 @@ export default function FlairsSettings() {
...
+  );
+
+  const handleDisplayPrefChange = (newPref: FlairDisplayPreference) => {
+    setDisplayPref(newPref);
+    localStorage.setItem("lb_flair_display_preference", newPref);
+    if (globalContext.setFlairDisplayPreference) {
+      globalContext.setFlairDisplayPreference(newPref);
+    }
+  };
+
```
MonkeyDo, line 80: You can use atomWithStorage for this, see for example https://github.com/metabrainz/listenbrainz-server/blob/e632a96114488c602cae2e22b7068bd042ba1ed4/frontend/js/src/user/playlists/Playlists.tsx#L75C12-L78C3

Although this will likely not be needed when saving the preference server-side

frontend/js/src/index.tsx
MonkeyDo, line 1: Once the server-side changes have been  impleemented,. this big refactor and all changes on this file will likely need to be reverted, as the preference will be part of globalAppContext.settings or globalAppContext.flair or something

frontend/js/src/utils/GlobalAppContext.tsx
MonkeyDo, line 1: Here, some changes will be required when the flair display preference is part of the props sent by the server (see the [type here](https://github.com/metabrainz/listenbrainz-server/blob/e632a96114488c602cae2e22b7068bd042ba1ed4/frontend/js/src/utils/utils.tsx#L834-L848) that will need to be modified too)

MonkeyDo (commented):
Hi @AviralKaushal thank you for opening this PR.

At a quick glance, it looks pretty good.
I agree with kepstin, this should be saved server-side instead.
You can likely re-use the existing [/settings/flair endpoint](https://github.com/metabrainz/listenbrainz-server/blob/e632a96114488c602cae2e22b7068bd042ba1ed4/listenbrainz/webserver/views/user_settings_api.py#L34-L56)
I will wait until that is implemented for a full review.

Please fill up the "AI usage" from the [PR template](https://github.com/metabrainz/listenbrainz-server/blob/master/.github/PULL_REQUEST_TEMPLATE.md) that you removed.
I can see you used AI at the very least for communications, and likely for the code generation as well.
Please fill it up honestly, it is important for us to track this.

ghost:
Hi @MonkeyDo ,

Thanks for the quick review and detailed feedback!

1. Server-side State & Refactoring
I’ll refactor the implementation to persist the preference server-side using the /settings/flair endpoint and update GlobalAppContext accordingly. I’ll also revert the client-side localStorage state and the extra refactoring in index.tsx.

2. AI Usage Section
Apologies for removing that from the template! I’ll restore and fill out the AI Usage section in the PR description along with the server-side updates.

I’ll mark this PR as a draft (or update it here) once the server-side changes and disclosures are pushed and ready for full review.

PR #3799, opened 2026-08-05
Fetch and backfill MetaBrainz user emails

amCap1712:
Request the email scope for MetaBrainz OAuth and use the UserInfo endpoint when creating new ListenBrainz users. Store confirmed addresses locally while treating UserInfo request failures as login failures for new users.

Replace the legacy MusicBrainz editor database email copier with a batched backfill from the MetaBrainz user database. Only fill missing ListenBrainz emails so webhook-synced and existing values are preserved.

### DIFF
--- a/listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
+++ b/listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
@@ -1,34 +1,88 @@
 import psycopg2
 from psycopg2.extras import execute_values
-from brainzutils.musicbrainz_db.editor import fetch_multiple_editors
 from flask import current_app
 
 from listenbrainz import db as lb_db
+from listenbrainz.db import donation
+
+EMAIL_BATCH_SIZE = 10_000
 
 
-def copy_emails():
+def copy_emails(batch_size: int = EMAIL_BATCH_SIZE) -> int:
+    """Backfill missing ListenBrainz emails from confirmed MetaBrainz accounts."""
+    if donation.engine is None:
+        raise RuntimeError("MetaBrainz database connection is not configured")
+    if batch_size <= 0:
+        raise ValueError("Email batch size must be positive")
+
     current_app.logger.info("Beginning to update emails for users...")
-    connection = lb_db.engine.raw_connection()
+    lb_connection = None
+    meb_connection = None
+    updated = 0
+    last_lb_user_id = 0
+
     try:
-        with connection.cursor() as cursor:
-            cursor.execute('''SELECT musicbrainz_row_id FROM "user"''')
-            current_app.logger.info("Fetched editor ids from ListenBrainz.")
-            editor_ids = [row[0] for row in cursor.fetchall()]
-            editors = fetch_multiple_editors(editor_ids)
-            current_app.logger.info("Fetched editor emails from MusicBrainz.")
-            emails = [(editor_id, editor['email']) for editor_id, editor in editors.items()]
-            query = """
-                UPDATE "user"
-                SET email = editor_email_temp.email
-                FROM (VALUES %s)
-                AS editor_email_temp(id, email)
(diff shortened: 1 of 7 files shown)


listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
```diff
@@ -1,34 +1,88 @@
...
 
 from listenbrainz import db as lb_db
+from listenbrainz.db import donation
 
+EMAIL_BATCH_SIZE = 10_000
+
+
+def copy_emails(batch_size: int = EMAIL_BATCH_SIZE) -> int:
+    """Backfill missing ListenBrainz emails from confirmed MetaBrainz accounts."""
+    if donation.engine is None:
```
MonkeyDo, line 13: The naming is confusing here, we should consider renaming this "donation" file -> "metabrainz"

listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
```diff
@@ -1,34 +1,88 @@
...
+                if not lb_users:
+                    break
+
+                last_lb_user_id = lb_users[-1][0]
+                meb_user_ids = [row[1] for row in lb_users]
+                meb_cursor.execute(
+                    """
+                    SELECT id, email
+                      FROM "user"
+                     WHERE id = ANY(%s)
```
MonkeyDo, line 51: For my own education, is there a difference between `WHERE id = ANY(%s)` and `WHERE id IN %s` ?

listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
```diff
@@ -1,34 +1,88 @@
...
+                       AND email IS NOT NULL
```
MonkeyDo, line 52: ~Does this take into account email validation?~

EDIT: I checked the MeB.org codfe, there is a separate `unconfirmed_email` column so anything in this `email` column should be validated.

listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
```diff
@@ -1,34 +1,88 @@
...
+                    """
+                    UPDATE "user" AS lb_user
+                       SET email = meb_user.email
+                      FROM (VALUES %s) AS meb_user(id, email)
+                     WHERE lb_user.musicbrainz_row_id = meb_user.id
+                       AND lb_user.email IS NULL
+                 RETURNING lb_user.id
+                    """,
+                    emails,
+                    template=None,
```
MonkeyDo, line 71: What does this line do?

listenbrainz/webserver/login/provider.py
MonkeyDo, line 21: Same comment as above for the naming, we should consider renaming this to MetaBrainzService to avoid confusion in the future.

MonkeyDo (commented):
Some details, no blockers.

listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
amCap1712: Can do but its currently used for donations / flair so let's do separately.

listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
amCap1712: Its mostly similar, slightly different syntax. https://stackoverflow.com/questions/34627026/in-vs-any-operator-in-postgresql

listenbrainz/webserver/login/copy_files_from_mb_to_lb.py
amCap1712: Use the default template for interpolation.

listenbrainz/webserver/login/provider.py
amCap1712: Yes makes sense.

PR #3804, opened 2026-08-08
perf(lb-radio): fix auth guard, add rate limit, cache services, eliminate remaining HTTP loopbacks

anshg1214:
# Problem

Following #3789, four HTTP loopbacks remain in the LB Radio pipeline: `stats`, `playlist`, `recs`, and recording metadata lookup. Two additional issues existed independently:

1. `ensure_user_token_for_expensive_endpoint()` returned a 401 response tuple that callers silently dropped -- unauthenticated requests reached the radio pipeline unblocked.
2. Authenticated users inherited the global 100k/10s rate limit intended for Troi CLI usage, making it trivial to hammer the MusicBrainz API from the `country` entity type.

# Solution

**Auth guard fix** -- `ensure_user_token_for_expensive_endpoint()` now raises `APIUnauthorized` instead of returning a response tuple.

**Per-endpoint rate limit** -- lb-radio enforces a separate Redis counter (`lb_radio:<token>`, 10 req/10s) independent of the global limit.

**Three new DB-direct services** replacing HTTP loopbacks:

- `LBRadioStatsService` -- replaces `liblistenbrainz.get_user_recordings()` with a direct CouchDB call. Results cached in Redis at 1-hour TTL (stats update weekly/monthly).
- `LBRadioPlaylistService` -- replaces `GET api.listenbrainz.org/1/playlist/{mbid}` with `db.playlist.get_by_mbid()`. Access control uses `playlist.is_visible_by()` -- identical to the HTTP endpoint, covering creator and collaborators.
- `LBRadioRecsService` -- replaces the paginated `liblistenbrainz.get_user_recommendation_recordings()` loop with a single `db.recommendations_cf_recording.get_user_recommendation()` call.

**Recording metadata cache** -- `fetch_metadata()` in `metadata_api.py` now caches per-MBID results in Redis at 24-hour TTL. All callers benefit: lb-radio recording lookup, entity pages, and the `/1/metadata/recording` endpoint.

All services return the same data shape as their HTTP counterparts. 1:1 parity verified by integration tests that mock at the DB layer and test through the full service -> Troi element pipeline.

**Troi dependency:** `requirements.txt` needs updating to include commit `5da64e4` of `troi-recommendation-playground` which adds the stats, playlist and recs service hooks. See Troi PR: metabrainz/troi-recommendation-playground#190.

# AI usage

Used Claude Code to assist with implementation, test writing, and code review throughout this PR. All code was reviewed, understood, and validated against the existing codebase before committing.

# Action

metabrainz/troi-recommendation-playground#190 must be merged and a new release cut before `requirements.txt` can be updated to a stable version pin. For now it is pinned to commit `5da64e4`.

### DIFF
--- a/listenbrainz/radio/playlist.py
+++ b/listenbrainz/radio/playlist.py
@@ -26,6 +26,6 @@ class LBRadioPlaylistService(Service):
                 user_id = user["id"]
 
         if not playlist.is_visible_by(user_id):
-            raise RuntimeError(f"Playlist {playlist_mbid} is private.")
+            raise RuntimeError(f"Cannot find playlist {playlist_mbid}.")
 
         return [str(r.mbid) for r in playlist.recordings]
--- a/listenbrainz/radio/playlist.py
+++ b/listenbrainz/radio/playlist.py
@@ -1,0 +1,31 @@
+from troi.service import Service
+
+import listenbrainz.db.user as db_user
+import listenbrainz.db.playlist as db_playlist
+from listenbrainz.webserver import db_conn, ts_conn
+
+
+class LBRadioPlaylistService(Service):
+    """Replaces the GET https://api.listenbrainz.org/1/playlist/{mbid} HTTP call
+    in LBRadioPlaylistRecordingElement with a direct DB lookup."""
+
+    SLUG = "playlist"
+
+    def __init__(self):
+        super().__init__(self.SLUG)
+
+    def fetch(self, playlist_mbid: str, auth_token: str | None) -> list[str]:
+        playlist = db_playlist.get_by_mbid(db_conn, ts_conn, playlist_mbid, load_recordings=True)
+        if playlist is None:
+            raise RuntimeError(f"Cannot find playlist {playlist_mbid}.")
+
+        user_id = None
+        if auth_token is not None:
+            user = db_user.get_by_token(db_conn, auth_token)
+            if user is not None:
+                user_id = user["id"]
+
+        if not playlist.is_visible_by(user_id):
+            raise RuntimeError(f"Playlist {playlist_mbid} is private.")
+
+        return [str(r.mbid) for r in playlist.recordings]
--- a/listenbrainz/radio/tests/test_recs_service.py
+++ b/listenbrainz/radio/tests/test_recs_service.py
@@ -103,12 +103,13 @@ class TestRecsElementWithService:
     def test_medium_mode_skips_first_333_recommendations(self, _mock_user, mock_recs):
         # medium offset=333: only recs after index 333 should be considered
         padding = [{"recording_mbid": f"pad-{i:04d}", "score": 1.0, "latest_listened_at": None}
-                   for i in range(333)]
+                   for i in range(400)]
         target_rec = {"recording_mbid": "target-0001", "score": 9.0, "latest_listened_at": None}
         mock_recs.return_value = _make_db_recommendations(padding + [target_rec])
         result = _element_with_service(listened="all", mode="medium").read([])
         mbids = {r.mbid for r in result}
         assert target_rec["recording_mbid"] in mbids
+        assert "pad-0334" in mbids
         assert "pad-0000" not in mbids
 
     @patch("listenbrainz.radio.recs.db_recommendations_cf_recording.get_user_recommendation")
--- a/listenbrainz/radio/tests/test_recs_service.py
+++ b/listenbrainz/radio/tests/test_recs_service.py
@@ -1,0 +1,127 @@
+from unittest.mock import patch, MagicMock
+
+from troi import Recording
+from troi.patches.lb_radio_classes.recs import LBRadioRecommendationRecordingElement
+
+from listenbrainz.radio.recs import LBRadioRecsService
+
+USER_NAME = "recsuser"
+USER_ID = 7
+FAKE_USER = {"id": USER_ID}
+
+REC_UNLISTENED_1 = {"recording_mbid": "aaaa-0001", "score": 9.5, "latest_listened_at": None}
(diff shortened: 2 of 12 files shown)


listenbrainz/radio/tests/test_recs_service.py
```diff
@@ -0,0 +1,127 @@
...
+    def test_medium_mode_skips_first_333_recommendations(self, _mock_user, mock_recs):
+        # medium offset=333: only recs after index 333 should be considered
+        padding = [{"recording_mbid": f"pad-{i:04d}", "score": 1.0, "latest_listened_at": None}
+                   for i in range(333)]
+        target_rec = {"recording_mbid": "target-0001", "score": 9.0, "latest_listened_at": None}
+        mock_recs.return_value = _make_db_recommendations(padding + [target_rec])
+        result = _element_with_service(listened="all", mode="medium").read([])
+        mbids = {r.mbid for r in result}
+        assert target_rec["recording_mbid"] in mbids
+        assert "pad-0000" not in mbids
```
MonkeyDo, line 113: Does it make sense to also test the positive assertion, not just the negative one ?
```suggestion
        assert "pad-0334" in mbids
        assert "pad-0000" not in mbids
```

listenbrainz/radio/playlist.py
```diff
@@ -0,0 +1,31 @@
...
+            raise RuntimeError(f"Cannot find playlist {playlist_mbid}.")
+
+        user_id = None
+        if auth_token is not None:
+            user = db_user.get_by_token(db_conn, auth_token)
+            if user is not None:
+                user_id = user["id"]
+
+        if not playlist.is_visible_by(user_id):
+            raise RuntimeError(f"Playlist {playlist_mbid} is private.")
```
MonkeyDo: This is a departure from the previous 404 we returned:
https://github.com/metabrainz/listenbrainz-server/blob/e632a96114488c602cae2e22b7068bd042ba1ed4/listenbrainz/webserver/views/playlist_api.py#L562-L563

I think we should keep the same mechanism and raise the same `Cannot find playlist {playlist_mbid}` as above, to keep private playlists truly private.
I know it's very very unlikely someone would try random UUIDs to find private playlists they can't access, but still, if the goal is to transfer functionality 1:1 we should maintain the logic.

listenbrainz/webserver/views/explore_api.py
```diff
@@ -14,6 +14,11 @@
 from listenbrainz.radio.artist import RecordingSearchByArtistService
 from listenbrainz.radio.tags import RecordingSearchByTagService
 from listenbrainz.radio.recording_lookup import RecordingLookupService
+from listenbrainz.radio.stats import LBRadioStatsService
+from listenbrainz.radio.playlist import LBRadioPlaylistService
+from listenbrainz.radio.recs import LBRadioRecsService
+
+LB_RADIO_PER_TOKEN_LIMIT = 10  # per rate-limit window (10s), separate from the global 100k/10s limit
```
MonkeyDo: I wonder if this is still too liberal for this endpoint,
I would half it and see if people complain, because I don't expect that regular users would do more than a request every 2 seconds in a window of 10s.
Did you have anything specific in mind with choosing 10/10sec?

listenbrainz/webserver/views/test/test_fetch_metadata_cache.py
```diff
@@ -0,0 +1,110 @@
...
+    "artist": {"name": "Radiohead"},
+    "tag": {"recording": []},
+    "release": {"name": "Pablo Honey"},
+}
+FULL_B = {
+    "recording": {"name": "Karma Police"},
+    "artist": {"name": "Radiohead"},
+    "tag": {"recording": []},
+    "release": {"name": "OK Computer"},
+}
```
MonkeyDo, line 26: Just wanted to point out this is a good occasion to put in your favorite artist of the moment :)

listenbrainz/webserver/views/test/test_fetch_metadata_cache.py
```diff
@@ -0,0 +1,110 @@
...
+        assert "release" not in result[MBID_A]
+
+    @patch("listenbrainz.webserver.views.metadata_api.cache")
+    @patch("listenbrainz.webserver.views.metadata_api.get_metadata_for_recording")
+    def test_incs_filter_applied_on_cache_miss(self, mock_db, mock_cache):
+        mock_cache.get.return_value = None
+        mock_db.return_value = [_make_entry(MBID_A, FULL_A)]
+        result = fetch_metadata([MBID_A], ["release"])
+        assert "release" in result[MBID_A]
+        assert "artist" not in result[MBID_A]
```
MonkeyDo, line 85: To stay consistent with the previous test
```suggestion
        assert "artist" not in result[MBID_A]
        assert "tag" not in result[MBID_A]
```

MonkeyDo (changes requested):
Thanks for these changes!
This should continue to improve LBRadio so we can put it back online with a saner setup.

listenbrainz/webserver/views/explore_api.py
anshg1214: Nothing specific, 10 was just a first guess. Dropped it to 5 per window.

listenbrainz/webserver/views/test/test_fetch_metadata_cache.py
anshg1214: Ha, leaving Radiohead for now.

listenbrainz/radio/playlist.py
anshg1214: Done. Hidden playlists now raise the same Cannot find playlist {mbid} as a missing playlist, so the two cases are indistinguishable.

MonkeyDo (approved):
Happy with the changes, feel free to merge when ready

PR #3808, opened 2026-08-18
Add script to migrate listens to user_id partitioned postgres database

amCap1712:
Adds manage.py migrate_listens with create-schema, create-indexes, full, incremental and replay-deletes commands to copy listens from timescale to a new plain postgres database hash-partitioned by user_id, and the SQLALCHEMY_LISTENS_URI config for it. Commands run inside a flask app context and log through the app logger.

Disable the listen_delete_metadata cleanup after spark dumps until the migration is done, the rows are needed to replay deletes on the target.

ACTION:
There is some spark dump cleanup code to revert once the migration is finished.
See https://github.com/metabrainz/listenbrainz-server/pull/3808/changes#diff-712749a7b0f54bf3d194147fa0a73599fe6ec3c928d015eb77935ddcb162a596R184-R188

AI disclosure:                                                                                                                                                                                                         
OpenAI Codex was used to help develop and review this migration tooling, including the follow-up safeguards and tests. All changes remain subject to human review before deployment.

### DIFF
--- a/listenbrainz/listenstore/migrate_listens.py
+++ b/listenbrainz/listenstore/migrate_listens.py
@@ -1,0 +1,404 @@
+""" Migrate listens from timescale (SQLALCHEMY_TIMESCALE_URI) to a plain postgres database
+(SQLALCHEMY_LISTENS_URI) where listens are hash-partitioned by user_id.
+
+Once the unique index exists on the target, copies use ON CONFLICT DO UPDATE (only when the source
+row is newer) so re-runs are idempotent and a listen re-submitted after being deleted picks up its
+new `created`. Before the index exists (initial bulk load) plain inserts are used and each
+listened_at window is committed as a whole: if a full run dies, resume it with
+--start <start of the window that was running>. Do not re-run a full copy over windows that
+completed while the unique index did not exist yet.
+
+`created` defaults to the transaction start time, so subtract a small overlap from the logged
+checkpoint when passing it as --since. Deletes are not propagated by the copies, replay-deletes
+applies listen_delete_metadata (status = 'complete') and deleted_user_listen_history to the target.
+Its logged --since-id / --since-history-id already include an overlap of DELETE_ID_OVERLAP rows
+because rows whose id was allocated but not committed yet when the checkpoint was read would
+otherwise be skipped forever. Replays are idempotent so re-applying them is harmless.
+
+Each cycle should run replay-deletes before incremental. Deletes committed in timescale after the
+last replay-deletes are never applied to the target, so before the final cycle stop writes to
+timescale (and let the delete_listens cron drain) and then run replay-deletes + incremental one
+last time.
+"""
+import os
+import re
+import time
+from datetime import datetime, timedelta, timezone
+
+import click
+import psycopg2
+import psycopg2.extras
+from flask import current_app
+from psycopg2.extras import execute_values
+
+from listenbrainz.webserver import create_app
+
+psycopg2.extras.register_uuid()
+
+ADMIN_SQL_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "..", "admin", "listens")
+CREATE_TABLES_SQL_FILE = os.path.join(ADMIN_SQL_DIR, "create_tables.sql")
+CREATE_INDEXES_SQL_FILE = os.path.join(ADMIN_SQL_DIR, "create_indexes.sql")
+
+PARTITION_COUNT = 256
+DEFAULT_BATCH_SIZE = 10000
+DEFAULT_WINDOW = timedelta(days=30)
+# number of ids subtracted from the logged delete checkpoints, see the module docstring
+DELETE_ID_OVERLAP = 1000
+UNIQUE_INDEX_NAME = "user_id_listened_at_recording_msid_ndx_listen"
+
+CREATE_PARTITION_SQL = """
+    CREATE TABLE IF NOT EXISTS listen_p{index:03d}
+    PARTITION OF listen FOR VALUES WITH (MODULUS {modulus}, REMAINDER {index})
+"""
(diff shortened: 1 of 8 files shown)


listenbrainz/listenstore/migrate_listens.py
```diff
@@ -0,0 +1,404 @@
...
+        target_conn.close()
+
+
+def migrate_incremental(since, until, batch_size):
+    """ Copy all listens created at or after `since` (and before `until` if given). """
+    source_conn = connect_source()
+    target_conn = connect_target()
+    try:
+        max_created = _log_checkpoint(source_conn)
+        insert_sql = _insert_sql(target_conn)
```
MonkeyDo: Codex points out that this call should fail if the unique iundex does not already exist (for the incremental path only) and would otherwise duplicate listens if running an incremental migration without the index being created first.
That may be an issue in some unintended scenarios,for example
1.creating the schema
2. runinng the full migration
3. creating the indexes fails without us noticing it
4. then running an incremental migration.
I guess a bit unlikely but should be guarded against.

The original output:
```
  - [P2] Require the unique index for incremental copies — /home/monkey/Work/ListenBrainz/listenbrainz-server/listenbrainz/listenstore/migrate_listens.py:285-285
    When incremental runs before create-indexes has successfully created the target unique index, _insert_sql() returns plain INSERT here, so the overlapped --since workflow described in
    this file will insert a second copy of any already-migrated listen instead of resolving via ON CONFLICT. Because the CLI advertises incremental copies as idempotent and this would
    silently corrupt the new listen table, make this path fail unless UNIQUE_INDEX_NAME exists, keeping plain INSERT only for the initial full load.
```

listenbrainz/listenstore/migrate_listens.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+def connect_target():
+    return psycopg2.connect(_uri("SQLALCHEMY_LISTENS_URI"))
+
+
+def _parse_ts(value):
+    if value is None:
+        return None
+    dt = value if isinstance(value, datetime) else datetime.fromisoformat(value)
```
MonkeyDo, line 126: This will raise if we pass a date in the wrong format, right?
My understanding is that in that case it would raise at the very beginning when invoking the command, before actually running anything.
If so that should be fine.

listenbrainz/listenstore/migrate_listens.py
```diff
@@ -0,0 +1,404 @@
...
+        params = [since]
+        if until is not None:
+            query += " AND created < %s"
+            params.append(until)
+
+        t0 = time.monotonic()
+        read, inserted = copy_batches(source_conn, target_conn, query, params, insert_sql, INSERT_TEMPLATE,
+                                      batch_size, "migrate_listens_incremental")
+        current_app.logger.info(
+            "incremental migration complete (created >= %s%s): read %d, inserted/updated %d in %.1fs, next --since: %s",
```
MonkeyDo: Here we are printing `max_created`, but I guess the next run should use a modified value as per the log_checkpoint print statement
> max(created) in source at start of run: %s, use it (minus a small overlap) as --since 

Would make sense to give the same warning in this print statement if it applies (I'm not sure whether it does or not)

listenbrainz/listenstore/migrate_listens.py
```diff
@@ -0,0 +1,404 @@
...
+completed while the unique index did not exist yet.
+
+`created` defaults to the transaction start time, so subtract a small overlap from the logged
+checkpoint when passing it as --since. Deletes are not propagated by the copies, replay-deletes
+applies listen_delete_metadata (status = 'complete') and deleted_user_listen_history to the target.
+Its logged --since-id / --since-history-id already include an overlap of DELETE_ID_OVERLAP rows
+because rows whose id was allocated but not committed yet when the checkpoint was read would
+otherwise be skipped forever. Replays are idempotent so re-applying them is harmless.
+
+Each cycle should run replay-deletes before incremental. Deletes committed in timescale after the
```
MonkeyDo, line 18: This is easy to miss if we run command manually, no? Especially if cycles take hours and we do other things in-between.
Might be good to have a step-by-step plan for each cycle, a checklist similar to [pointing and calling](https://en.wikipedia.org/wiki/Pointing_and_calling)

EDIT: I created one, please review https://docs.google.com/document/d/1YVzdB8f1-TLxzj8VI8R3XHeYM7lTseLWXcbiFBSOSv8/edit?usp=sharing

MonkeyDo (commented):
AI disclaimer please!

To be honest a lot of the SQL in there is beyond my knowledge, so my review is not as throught as one might like for this important job.
I did review everything (learning things along the way) and it does look fine to my untrained eyes; I also separately ran a review with codex, with one suggestion in comments below.

We discussed some of these points, I would suggest the following plan before final production deployment:
1. test the script on the sample database in local dev
2. migrate test.LB so we see how it behaves in real-life infra
3. run the migration on the prod DB and point only beta.LB to the new DB, so that we can test on a "production" setup without potentially breaking the prod website.
4. At that point we can ask our community to thoroughly test beta.LB for a week or whatever time, to catch anything else we might have missed.


Additionally, regarding some comments below it looks possible to create duplicates unintentionally.
Would it make sense to have one more command to check the integrity fo the table and that there are no duplicates?
Something we can run at the end of runs to ensure everything is fine before running the next one, or at the very end.

listenbrainz/listenstore/migrate_listens.py
amCap1712: yup.

listenbrainz/listenstore/migrate_listens.py
amCap1712: LGTM, thanks!

amCap1712:
Updated PR description, ran against test and checklist looks good.

PR #3812, opened 2026-08-19
fix: address critical security vulnerabilities (XSS, Secret Leakage)

MIHIRCHAUDHARI390:
# Problem

There are critical security vulnerabilities in the ListenBrainz server:
1. Potential stored XSS via crafted playlist descriptions due to unsanitized playlist annotations.
2. Potential credential leakage when logging configuration values in `prod`/`beta`/`test` environments.

# Solution

This PR addresses the security vulnerabilities mentioned above:
1. **XSS Prevention in React**: Sanitizes playlist annotations with DOMPurify in `PlaylistCard.tsx` and `PlaylistSearch.tsx` to prevent stored XSS via crafted playlist descriptions. (DOMPurify was already used elsewhere but missed in two `dangerouslySetInnerHTML` usages).
2. **Prevent Credential Leakage**: Redacts sensitive configuration keys (SECRET_KEY, database URIs, API secrets, passwords) when logging config values in `prod`/`beta`/`test` environments to prevent credential leakage to logs in `listenbrainz/webserver/__init__.py`.

# AI usage
* [ ] I did not use any AI
* [x] I have used AI in this PR (Used AI for coding and drafting the pull request description)

#### If you did use AI:

# Action
None

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -73,7 +73,23 @@ def load_config(app):
     # Output config values and some other info
     if deploy_env in ['prod', 'beta', 'test']:
         print('Configuration values are as follows: ')
-        print(pprint.pformat(app.config, indent=4))
+        sensitive_keys = {
+            'SECRET_KEY', 'SQLALCHEMY_DATABASE_URI', 'POSTGRES_ADMIN_URI',
+            'SQLALCHEMY_TIMESCALE_URI', 'TIMESCALE_ADMIN_URI',
+            'TIMESCALE_ADMIN_LB_URI', 'RABBITMQ_PASSWORD', 'RABBITMQ_USERNAME',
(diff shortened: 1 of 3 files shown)


MonkeyDo:
Please fill in the original PR template including the AI disclosure.

MIHIRCHAUDHARI390:
yeah i updated it now.

listenbrainz/webserver/__init__.py
MonkeyDo, line 1: This we certainly do not want to do.
Logs are not public, and this specific log line allows us to see what configuration values are in use in our production environments, so it would be fully counterproductive to redact them.

MonkeyDo (commented):
Thanks for looking into these.
Were you able to replicate the potential issue described in local dev?
We sanitize playlist descriptions server-side, so we should never get in a position where the data sent from the server would be problematic.
That being said, we could have missed something, but I would like to see a reproducible vector of attack before merging this.

More generally, please open separate PRs for separate issues rather than clumping unrelated fixes together.

PR #3814, opened 2026-08-19
fix: Strip empty optional fields from listen metadata instead of rejecting

anshg1214:
# Problem

## What is the issue?
Listens submitted with `release_name: ""` (empty string) are rejected with a validation error, even though `release_name` is an optional field. This causes valid listens to be silently dropped during background imports.

Fixes [LISTENBRAINZ-KP](https://metabrainz-foundation-inc.sentry.io/issues/LISTENBRAINZ-KP).

## Why does it happen?
The `validate_basic_metadata` function in `api_tools.py` treats all fields identically: if the value is present but empty after stripping whitespace, it raises `ListenValidationError` regardless of whether the field is required or optional.

For optional fields like `release_name`, an empty value should simply be removed from the metadata rather than rejecting the entire listen. The function already has a `required` parameter but the empty-check path didn't use it.

Sentry shows 138K events with the exact pattern: `"field track_metadata.release_name is empty."` with payloads like `{'release_name': '', 'artist_name': 'Fall Out Boy', 'track_name': 'Love From The Other Side'}`. The artist and track are valid, only the empty release_name triggers rejection.

# Solution

When `required=False` and the value is empty after strip, delete the key from the metadata instead of raising an error. Required fields still raise as before.

Also downgraded the importer's validation failure log from `error` to `warning` since these are expected data quality issues, not system errors.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

### DIFF
--- a/listenbrainz/background/listens_importer/base.py
+++ b/listenbrainz/background/listens_importer/base.py
@@ -164,7 +164,7 @@ class BaseListensImporter(ABC):
                 validate_listen(listen, LISTEN_TYPE_IMPORT)
                 validated_listens.append(listen)
             except ListenValidationError as e:
-                current_app.logger.error("Invalid listen: %s", e)
+                current_app.logger.warning("Invalid listen: %s", e)
         
         validation_stats["success_count"] += len(validated_listens)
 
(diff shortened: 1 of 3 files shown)


amCap1712 (approved):
LGTM, but let's add a test.

PR #3819, opened 2026-08-19
GSOC-26 Musicbrainz Release Collections Import 

yateendogra2109:
# Problem

The import for recording MB collection in Listenbrainz is already done . 

Users could not open a release collection page to inspect its contents, and they could not save a release collection as a ListenBrainz playlist.

# Solution

This pr focuses on release collections . Clicking on import collection now shows both release and recording collections . When clicked on release collection , a new page opens showing list of releases .Each release row now links to the ListenBrainz release page at `/release/<release_mbid>/ `when clicked. . One can click on save as playlist button where releases will be flattened to tracks and then saved as usual LB playlist.
A` flatten=tracks` mode is used  for converting releases into playlist tracks

Attached is the example of a preview page , listing the releases.

image

# AI usage

* [ ] I did not use any AI
* [ Yes ] I have used AI in this PR (add more details below)

I used claude for generating some pieces of code .

#### If you did use AI:
* [Yes] I used AI tools for coding
* [Yes] I understand all the changes made in this PR

# Tests

Tests covering invalid MBIDs, public collections, and private collection access.
Integration tests for release collection payloads.
Added test for flattening release collections into recordings when saving as a playlist.

### DIFF
--- a/frontend/js/src/collections/Collection.tsx
+++ b/frontend/js/src/collections/Collection.tsx
@@ -3,7 +3,9 @@ import * as React from "react";
 import { faPlayCircle } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import {
+  FetchNextPageOptions,
   InfiniteData,
+  InfiniteQueryObserverResult,
   QueryClient,
   useInfiniteQuery,
   useQueryClient,
@@ -25,6 +27,7 @@ import {
   PLAYLIST_ARTIST_URI_PREFIX,
   PLAYLIST_TRACK_URI_PREFIX,
 } from "../playlists/utils";
+import ListenCard from "../common/listens/ListenCard";
 import { RouteQuery } from "../utils/Loader";
 import { generateAlbumArtThumbnailLink } from "../utils/utils";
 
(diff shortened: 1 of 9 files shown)


frontend/js/src/collections/Collection.tsx
```diff
@@ -78,6 +84,83 @@ function asJSPFTrack(track: MusicBrainzCollectionTrack): JSPFTrack {
   return jspfTrack;
 }
 
+function ReleaseCollectionItemRow({
```
anshg1214: Let's see if we can reuse `ListenCard` or `PlaylistItemCard`

frontend/js/src/collections/Collection.tsx
```diff
@@ -78,6 +84,83 @@ function asJSPFTrack(track: MusicBrainzCollectionTrack): JSPFTrack {
...
+}
+
+async function fetchAllFlattenedCollectionTracks(
+  queryClient: QueryClient,
+  collectionMBID: string
+): Promise<JSPFTrack[]> {
+  const allTracks: JSPFTrack[] = [];
+  let offset = 0;
+
+  while (true) {
```
anshg1214: `while(true)` loop fetches all pages sequentially with no upper bound. A release collection with hundreds of releases could flatten to tens of thousands of tracks, locking the UI thread and flooding the API. Let's add an upper limit here.

frontend/js/src/collections/Collection.tsx
```diff
@@ -242,26 +352,50 @@ export default function CollectionPage() {
     if (isSaving) {
       return;
     }
+    if (isRecordingCollection && tracks.length === 0) {
+      return;
+    }
+    if (isReleaseCollection && releases.length === 0) {
+      return;
+    }
```
anshg1214: The component now has `isRecordingCollection` / `isReleaseCollection` flags scattered across rendering, callbacks, disabled states, labels, and virtualization config. It's becoming two pages sharing one component body. Let's split into focused sub-components or a render strategy keyed on entity type.

anshg1214 (changes requested):

PR #3821, opened 2026-08-21
Use AUTOCOMMIT isolation to eliminate idle-in-transaction connections

anshg1214:
# Problem

## What is the issue?
Per-request database connections sit idle-in-transaction for the entire request lifetime, causing connection pool exhaustion and a 100x p95 latency cliff (p50=1.3ms vs p95=382ms on user table PK lookups).

## Why does it happen?
SQLAlchemy's `Connection.execute()` implicitly opens a PostgreSQL transaction (autobegin). The per-request connection proxies (`db_conn`, `ts_conn`, `meb_conn`) hold this transaction open until `teardown_appcontext` calls `.close()`. During non-DB work (Redis, template rendering, external calls), the connection sits idle-in-transaction, blocking autovacuum and consuming pool slots.

# Solution

Set `AUTOCOMMIT` isolation level on all three per-request connection proxies. In AUTOCOMMIT mode, each `execute()` auto-commits immediately at the database level, eliminating idle-in-transaction connections.

Multi-statement write functions that need atomicity (DELETE+INSERT, UPDATE+INSERT, etc.) are wrapped in `with conn.engine.begin() as txn:` blocks. This checks out a separate short-lived connection with a real PostgreSQL transaction -- commits on block exit, rolls back on exception. The per-request connection stays in AUTOCOMMIT mode throughout.

Functions wrapped with `engine.begin()`:
- `feedback.insert()` -- DELETE + INSERT
- `pinned_recording.pin()` -- UPDATE + INSERT
- `external_service_oauth.save_token()` -- INSERT + conditional INSERT
- `external_service_oauth.delete_token()` -- DELETE + conditional DELETE
- `playlist.create()`, `update_playlist()`, `delete_recordings_from_playlist()`, `add_recordings_to_playlist()`, `delete_playlists_by_user_id()`
- `export.create_export_task()`, `delete_export_archive()`
- `import_listens.create_import_task()`, `delete_import_task()`

Single-statement writes keep their existing `commit()` calls, which are harmless no-ops in AUTOCOMMIT mode.

Test connections are also set to AUTOCOMMIT to match production behavior.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

### DIFF
--- a/listenbrainz/background/export.py
+++ b/listenbrainz/background/export.py
@@ -175,7 +175,12 @@ def export_listens_for_user(export_id, db_conn, ts_conn, tmp_dir: str, user_id: 
             update_export_progress(db_conn, export_id, f"Exporting listens for the period {period_str}")
             file_path = os.path.join(year_dir, f"{period['month']}.jsonl")
 
-            rowcount = export_listens_for_time_range(ts_conn, file_path, user_id, period["start"], period["end"])
+            # ``yield_per`` below uses a PostgreSQL server-side cursor, which
+            # requires a real transaction. The request-scoped Timescale
+            # connection is AUTOCOMMIT, so check out a short-lived transactional
+            # connection for each streamed export file.
+            with ts_conn.engine.begin() as txn:
+                rowcount = export_listens_for_time_range(txn, file_path, user_id, period["start"], period["end"])
             if rowcount > 0:
                 files.append(file_path)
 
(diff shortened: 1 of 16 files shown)


MonkeyDo:
There is a failing integration test (test_export). I rebased the PR but the same test failed with the same error so I think it is likely related to the PR changes

MonkeyDo:
Since this affects literally everything, how do we test this properly?
I guess we test every single feature on the site?

anshg1214:
This should not break anything. Let's not merge the PR and deploy it to test / beta for some time?

MonkeyDo:
> This should not break anything. Let's not merge the PR and deploy it to test / beta for some time?

If we coordinate this deployment on test I'm happy to go click on everything on the website and test every feature I can

PR #3823, opened 2026-08-21
Fix generic Exception for invalid user in playlist creation

MIHIRCHAUDHARI390:
# Problem

The `create_playlist` function in `listenbrainz/db/playlist.py` was raising a generic `Exception("TODO: Custom exception")` when an invalid user ID was provided. While the docstring documented that it should raise `InvalidUser`, this exception was never actually implemented or used.

# Solution

- Created the `InvalidUser` exception in `listenbrainz/db/exceptions.py` inheriting from `DatabaseException`.
- Imported `InvalidUser` into `listenbrainz/db/playlist.py`.
- Replaced the generic `Exception` with `InvalidUser` to improve error handling and fulfill the existing TODO.

# AI usage
* [ ] I did not use any AI

#### If you did use AI:

# Action
None required.

### DIFF
--- a/listenbrainz/db/playlist.py
+++ b/listenbrainz/db/playlist.py
@@ -9,6 +9,7 @@ from sqlalchemy import text
 
 from listenbrainz.db.model import playlist as model_playlist
 from listenbrainz.db import user as db_user
+from listenbrainz.db.exceptions import InvalidUser
 from listenbrainz.db.model.playlist import Playlist
 from listenbrainz.db.recording import load_recordings_from_mbids_with_redirects
 from listenbrainz.db.year_in_music import LAST_FM_FOUNDING_YEAR, MAX_YEAR_IN_MUSIC_YEAR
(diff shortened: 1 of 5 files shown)


anshg1214 (approved):
LG

listenbrainz/db/playlist.py
```diff
@@ -724,15 +725,15 @@ def create(db_conn, ts_conn, playlist: model_playlist.WritablePlaylist) -> model
...
     # TODO: In a way this is less than ideal -- the caller must take the string name and find the ID,
     # and then the name is fetched for verification again. Should we accept created_for here and do
     # lookup only here and not he in the API call validation?
     if playlist.created_for_id:
         created_for = db_user.get(db_conn, playlist.created_for_id)
         if created_for is None:
-            raise Exception("TODO: Custom exception")
+            raise InvalidUser("Invalid created_for user ID")
 
     query = text("""
```
anshg1214, line 738: Let's add try/except InvalidUser in the API layer to return 400 instead of 500

MIHIRCHAUDHARI390: Thanks for the feedback! I've updated the PR to handle the InvalidUser exception in the API layer.

1. Imported InvalidUser and wrapped the db_playlist.create() and db_playlist.copy_playlist() calls in a try/except block.
2. We now use log_raise_400() to properly return a 400 Bad Request to the user instead of letting it fall through to a 500 Internal Server Error.
3. I also added unit tests for both the DB and API layers to make sure this behavior is covered.

Let me know if everything looks good!

anshg1214 (approved):
LG

PR #3825, opened 2026-08-22
fix: redact sensitive config values from startup logs

MIHIRCHAUDHARI390:
# Problem

This PR addresses a critical configuration leak vulnerability. Currently, when the server starts in `prod`, `beta`, or `test` environments, it unconditionally prints the entirety of `app.config` to standard output. This inadvertently leaks highly sensitive information (e.g. `SECRET_KEY`, `SQLALCHEMY_DATABASE_URI`, API credentials, passwords) into the application logs.

# Solution

I've implemented a dictionary comprehension that filters and redacts known sensitive keys (replacing their values with `***REDACTED***`) before the configuration is printed to the logs.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

# Action

None.

### DIFF
--- a/listenbrainz/webserver/__init__.py
+++ b/listenbrainz/webserver/__init__.py
@@ -73,7 +73,23 @@ def load_config(app):
     # Output config values and some other info
     if deploy_env in ['prod', 'beta', 'test']:
         print('Configuration values are as follows: ')
-        print(pprint.pformat(app.config, indent=4))
+        sensitive_keys = {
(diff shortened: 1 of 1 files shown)


MonkeyDo:
As commented in #3812 printing these values in our (non-public) logs is intended for debugging purposes.
The logs are not public and hosted on an isolated private network.

I would suggest it is better to come talk to us about potential issues than to blingly trust an AI's assessment.
When in doubt, we have lot of open tickets that are requested by users, at [our ticket tracker](https://tickets.metabrainz.org/projects/LB)

PR #3826, opened 2026-08-23
fix: Add timeouts to external API requests

MIHIRCHAUDHARI390:
# Problem

Several external API integrations in the `listenbrainz/domain` directory were making outgoing HTTP requests using `requests.get` and `requests.post` without specifying a `timeout` parameter. In Python's `requests` library, if the remote server hangs and doesn't close the connection, the request will block indefinitely. This is a notorious cause of backend worker thread exhaustion and cascading failures.

# Solution

Enforced a strict `timeout=10` on the external API calls across these domain integrations to ensure ListenBrainz processes will gracefully error out (or move on) rather than locking up forever if external services (like MusicBrainz, Spotify, etc.) become unresponsive.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:


anshg1214 (approved):
LG

PR #3828, opened 2026-08-26
Fix: Prevent XSS by removing safe filter on user_name

MIHIRCHAUDHARI390:
# Problem

The Jinja templates `pin_card.html`, `playing_now.html`, and `playing_now_card.html` were using the `| safe` filter on the `user_name` variable. This introduces a Cross-Site Scripting (XSS) vulnerability, as malicious or unescaped HTML characters in a username could be injected directly into the DOM (such as within an `href` or `title` attribute, or as inner HTML).

# Solution

Removed the `| safe` filter from all occurrences of `user_name` rendering in the templates. Jinja will now automatically HTML-escape the usernames, protecting against XSS injections while properly rendering the username string.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

### DIFF
--- a/listenbrainz/domain/critiquebrainz.py
+++ b/listenbrainz/domain/critiquebrainz.py
@@ -37,7 +37,7 @@ class CritiqueBrainzService(BaseBrainzService):
         payload = review.dict(exclude_none=True)
         payload["is_draft"] = False
         payload["license_choice"] = CRITIQUEBRAINZ_REVIEW_LICENSE
-        return requests.post(CRITIQUEBRAINZ_REVIEW_SUBMIT_URL, json=payload, headers=headers)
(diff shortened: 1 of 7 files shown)


MonkeyDo:
Hello @MIHIRCHAUDHARI390, thanks for the PR.

There are a lot of unrelated changes in this PR, looks like a comment from another PR is also part of this one.
When you work on a new fix/feature/whatever, make sure to create a new branch for it to avoid these issues.

Regarding the `| safe` removal, I have two concerns:
1. this was there for a reason. some user names can contain characters that will be escaped, which is unwanted.
2. The usernames are sent directly from the server. I would like to see a replication of the suggested attack vector to confirm this is actually an issue rather than a blanket LLM finding

PR #3281, opened 2025-05-19
LB-1761 : Save filter state in local

Mshahnawaz1:
- 
- # Problem

Currently we don't store the state of filter in fresh-releases so when user refreshes the page all the choices of user gets reset. 

# Solution

So to solve this issue we are storing the filter state in local storage using localForage library. 
In this we are storing these states:

-   checkedList,
-   releaseTagsCheckList,
-   releaseTagsExcludeCheckList,
-   includeVariousArtists,
-   coverartOnly,

We are using a custom hook `userfilterpersistence.ts`  in `listenbrainz/hooks/userfilterpersistence.ts` to define the operations for filter data.

# Action
This needs to be checked if all the parameters that needs to be stored are correct. Also filters gets reset when we toggle from `All` to `For you` and also when refreshed.

PR #3830, opened 2026-08-27
LB-1761: Persist Fresh Releases filters across page reloads

Akansh475:
## Problem
Filters in the Fresh Releases explore section (release types, tags, various artists, cover-art-only) reset on every page load/reload, forcing users to re-apply them every time.

## Solution
Added a `useFreshReleasesFilterPersistence` hook that persists filter state using `localForage` (matching the existing pattern used in `listenStorage.ts`). Filters are stored separately per `pageType` (e.g. `release-filters-user` vs `release-filters-sitewide`), so switching between "For You" and sitewide releases no longer resets or conflicts with each other's saved filters.

This follows the approach discussed by @MonkeyDo in the previous attempt at this ticket (#3281):
- Single `freshReleasesFilters` object exposed by the hook, with a `setFreshReleasesFilters` function that accepts a partial update
- A separate storage key per `pageType` to fix the reset-on-toggle bug
- Uses `localForage`, consistent with existing storage utilities in the codebase

## Testing
Verified with `npx tsc --noEmit` (no errors) and `npx eslint` (no errors, only a pre-existing unrelated warning in `utils.tsx`). Was not able to run the full local dev environment (Docker + OAuth registration) for this session, so manual browser testing is still pending — flagging this openly for review.

### DIFF
--- a/frontend/js/src/explore/fresh-releases/components/ReleaseFilters.tsx
+++ b/frontend/js/src/explore/fresh-releases/components/ReleaseFilters.tsx
@@ -87,9 +87,7 @@ export default function ReleaseFilters(props: ReleaseFiltersProps) {
     if (isChecked) {
       setFreshReleasesFilters({ checkedList: [...checkedList, value] });
     } else {
-      const filtersList = checkedList.filter(
-        (item: string | undefined) => item !== value
-      );
+      const filtersList = checkedList.filter((item) => item !== value);
       setFreshReleasesFilters({ checkedList: filtersList });
     }
   };
@@ -102,7 +100,7 @@ export default function ReleaseFilters(props: ReleaseFiltersProps) {
 
     // remove from exclude list if it's there
     const excludeFiltersList = releaseTagsExcludeCheckList.filter(
-      (item: string | undefined) => item !== value
+      (item) => item !== value
     );
 
     setFreshReleasesFilters({
@@ -112,9 +110,7 @@ export default function ReleaseFilters(props: ReleaseFiltersProps) {
   };
 
   const removeFilterTag = (tag: string) => {
-    const filtersList = releaseTagsCheckList.filter(
-      (item: string | undefined) => item !== tag
-    );
+    const filtersList = releaseTagsCheckList.filter((item) => item !== tag);
     setFreshReleasesFilters({ releaseTagsCheckList: filtersList });
   };
 
@@ -126,7 +122,7 @@ export default function ReleaseFilters(props: ReleaseFiltersProps) {
 
     // remove from include list if it's there
     const includeFiltersList = releaseTagsCheckList.filter(
-      (item: string | undefined) => item !== value
+      (item) => item !== value
     );
 
     setFreshReleasesFilters({
@@ -137,7 +133,7 @@ export default function ReleaseFilters(props: ReleaseFiltersProps) {
 
   const removeExcludeTag = (tag: string) => {
     const filtersList = releaseTagsExcludeCheckList.filter(
-      (item: string | undefined) => item !== tag
+      (item) => item !== tag
     );
     setFreshReleasesFilters({ releaseTagsExcludeCheckList: filtersList });
   };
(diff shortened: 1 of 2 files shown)


frontend/js/src/explore/fresh-releases/hooks/useFreshReleasesFilterPersistence.ts
```diff
@@ -0,0 +1,69 @@
...
+    },
+    [pageType]
+  );
+
+  const clearSavedFilters = React.useCallback(() => {
+    const storageKey = getStorageKey(pageType);
+    filterStore.removeItem(storageKey);
+    setFreshReleasesFiltersState(DEFAULT_FILTERS);
+  }, [pageType]);
+
```
anshg1214, line 75: clearSavedFilters is exported but never called anywhere in this PR. Either wire it up to a "Reset filters" button (which would be nice UX for a persistence feature) or remove it

frontend/js/src/explore/fresh-releases/hooks/useFreshReleasesFilterPersistence.ts
```diff
@@ -0,0 +1,69 @@
...
+    (newFilters: Partial<FreshReleasesFilters>) => {
+      setFreshReleasesFiltersState((prevFilters: FreshReleasesFilters) => {
+        const updatedFilters = { ...prevFilters, ...newFilters };
+        const storageKey = getStorageKey(pageType);
+        filterStore.setItem(storageKey, updatedFilters);
+        return updatedFilters;
+      });
+    },
+    [pageType]
+  );
```
anshg1214, line 68: Calling filterStore.setItem() (which returns a Promise) inside a React setState updater is an anti-pattern. The updater should be a pure function.

frontend/js/src/explore/fresh-releases/components/ReleaseFilters.tsx
```diff
@@ -150,15 +162,18 @@ export default function ReleaseFilters(props: ReleaseFiltersProps) {
...
+      (checkedList?.length ?? 0) > 0 ||
+      includeVariousArtists === true
+    ) {
+      setFreshReleasesFilters({
+        coverartOnly: false,
+        checkedList: [],
+        includeVariousArtists: false,
+      });
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
anshg1214, line 178: This effect reads coverartOnly, checkedList, includeVariousArtists, and setFreshReleasesFilters but doesn't list them as deps. Is this intended?

frontend/js/src/explore/fresh-releases/components/ReleaseFilters.tsx
```diff
@@ -97,37 +99,47 @@ export default function ReleaseFilters(props: ReleaseFiltersProps) {
   ) => {
     event.preventDefault();
     const { value } = event.target;
-    setReleaseTagsCheckList([...releaseTagsCheckList, value]);
 
     // remove from exclude list if it's there
-    const filtersList = releaseTagsExcludeCheckList.filter(
-      (item) => item !== value
+    const excludeFiltersList = releaseTagsExcludeCheckList.filter(
+      (item: string | undefined) => item !== value
```
anshg1214: All the explicit type annotations in .filter() and .map() callbacks are redundant:
```
 // This:
checkedList.filter((item: string | undefined) => item !== value)
// Is identical to:
checkedList.filter((item) => item !== value)
```

anshg1214 (changes requested):

Akansh475:
Thanks for the thorough review — addressed all four points:

1. **clearSavedFilters unused**: wired it up to a new "Reset filters" button in the filter panel.
2. **setItem inside setState updater**: fixed — `setFreshReleasesFilters` is now a pure updater, persistence now happens in a separate `useEffect` that watches `freshReleasesFilters` and writes to storage as a side effect.
3. **Missing effect deps**: this was intentional (the effect resets filter values, so including those values as deps would make it re-fire on every reset). Added a clearer comment explaining why they're omitted rather than just the eslint-disable.
4. **Redundant type annotations**: removed all of them, you're right that TS already infers the types correctly.

Let me know if any of this needs further changes.

MonkeyDo:
Hi @Akansh475, just a quick unrelated note to say that we prefer talking to a human than reading LLM output.
Let's keep our corner of the internet more human :)

PR #3831, opened 2026-08-27
Split Flask app into separate API and web workers

anshg1214:
# Problem

## What is the issue?
Both `listenbrainz.org` (website) and `api.listenbrainz.org` (API) route to the same container (`listenbrainz-web-prod`) running a single uWSGI worker pool (5 processes x 20 threads). When bots or scrapers hammer the API, they fill the uWSGI queue and the website becomes unresponsive.

## Why does it happen?
The Aug 24 2026 scraper incident demonstrated this: three botnets overwhelmed the shared worker pool, the uWSGI queue hit 2048/2048, and both the API and website went down together because they share the same workers.

# Solution

Split the Flask application and uWSGI workers so the API and website run independently.

## Infrastructure (commit 1)

Add a second uWSGI runit service (`uwsgi_api`) to the Docker image so the API can run in its own container with an isolated worker pool. This follows the existing `api_compat` service pattern exactly:

- New `docker/services/uwsgi_api/` directory with ini template, consul-template config, runit service, and finish scripts
- uWSGI reads process/thread config from Consul at `LB/uwsgi.api.{env}.json` (separate from the website's `LB/uwsgi.{env}.json`)
- Container name `listenbrainz-api-{env}` triggers `rc.local` to enable the `uwsgi_api` service

## Flask app split (commit 2)

Split the monolithic Flask app so each container only loads the routes it needs:

- Split `_register_blueprints` into `_register_api_blueprints` (all `/1/` prefixed routes) and `_register_web_blueprints` (UI, entity pages, webhooks, etc.)
- Add `create_api_app()` factory that only registers API blueprints, without static files, admin, GDPR checks, or Jinja context processors
- Add `listenbrainz/api_server.py` as the WSGI entry point for API workers
- Point `uwsgi-api.ini` at `listenbrainz.api_server` instead of `listenbrainz.server`
- Tag Sentry events with `service:api` or `service:web` so errors are filterable by app
- Remove the `/1` prefix skip from the GDPR before_request hook (API routes no longer live in the web app)
- Add `api` service to `docker-compose.yml` for local development (port 8103)
- Test harness registers both blueprint sets so existing integration tests keep working

**The corresponding `docker-server-configs` PR will add the deployment config (service function, port assignment, Consul config) and the upstream routing change to actually switch traffic.**

Once deployed, API scrapers can fill the API queue without affecting the website.

* [x] I have reviewed the code and tested it locally

# AI usage

* [ ] I did not use any AI

#### If you did use AI:


amCap1712 (commented):
Its probably best to also separate the flask app creation then so that the api endpoints are registered only on the api app and the website ones on the website app.

anshg1214:
Yes, that's the plan. I plan to do it in a follow up PR so that we can get things moving. I believe we can start with a small change first and then separate app - With the number of blueprints we have - we'll not be saving a huge amount of ram if we split it right now.

listenbrainz/webserver/__init__.py
MonkeyDo, line 386: Since I see it here, we should move this to the gateway-level blocking and remove this bit of code (in a separate PR). Would you mind coordinating that with zas when he's got a moment?

MonkeyDo (commented):
Looking good from here.

Would like to see this run on test first and ensuring both web and API are working fine, before deploying.

docker/docker-compose.yml
```diff
@@ -68,6 +68,21 @@ services:
       - rabbitmq
       - couchdb
```
MonkeyDo, line 69: web container will need the api running (for example if running only some containers lik `docker compose up web`)

```suggestion
      - couchdb
      - api
```

MonkeyDo (commented):
Made a couple of changes to reflect the new 8103 default API port.
Tested locally with those changes, working fine.

listenbrainz/webserver/__init__.py
anshg1214: Sure! I'll take it forward.

MonkeyDo:
Failing test is unrelated, just saw the same one on a PR that only updates one of the CI actions (zizmore):
https://github.com/metabrainz/listenbrainz-server/actions/runs/31436869096/job/93612812073?pr=3807


But that does mean we have a flaky test

anshg1214:
@MonkeyDo Increasing the timeout fixed the tests

MonkeyDo:
This looks ready to me, but there are some merge conflicts before we can merge it

PR #3833, opened 2026-08-28
fix: Handle Typesense errors in recording search endpoint

anshg1214:
# Problem

The recording search endpoint (`RecordingSearchQuery.fetch()`) crashes when the Typesense search service is unreachable or times out. The unhandled `TypesenseClientError` propagates up to datasethoster, which then tries to serialize the raw exception object via `jsonify({"error": err})`, causing a secondary `TypeError` (the exception is not JSON-serializable).

This is one of the highest-volume Sentry errors:
- [LISTENBRAINZ-BP](https://metabrainz-foundation-inc.sentry.io/issues/LISTENBRAINZ-BP/) (TypeError from serialization, ~41K total events)
- [LISTENBRAINZ-BN](https://metabrainz-foundation-inc.sentry.io/issues/LISTENBRAINZ-BN/) (ReadTimeout from Typesense)

# Solution

Wrap the Typesense search call in a try/except for `TypesenseClientError` (the base class for all Typesense exceptions including `ReadTimeout`). On failure, return an empty list, which is a graceful degradation: the caller gets no search results rather than a 500 error.

This fixes both the primary timeout error and the secondary serialization crash.

# AI usage

#### If you did use AI:

### DIFF
--- a/listenbrainz/labs_api/labs/api/recording_search.py
+++ b/listenbrainz/labs_api/labs/api/recording_search.py
@@ -69,7 +69,10 @@ class RecordingSearchQuery(Query):
             'num_typos': NUM_TYPOS
         }
 
-        hits = self.client.collections[COLLECTION_NAME_WITHOUT_RELEASE].documents.search(search_parameters)
+        try:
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Yes, simple one, thank you!

PR #3835, opened 2026-08-28
fix: Validate recording MBIDs before database query

anshg1214:
# Problem

`RecordingFromRecordingMBIDQuery.fetch()` passes user-supplied recording MBIDs directly into a database query without validating they are valid UUIDs. Invalid MBIDs (malformed strings, empty values, None) cause `psycopg2.DataError` or `InvalidTextRepresentation` errors at the database level.

Sentry issue:
- [LISTENBRAINZ-2VQ](https://metabrainz-foundation-inc.sentry.io/issues/LISTENBRAINZ-2VQ/)

# Solution

Validate each MBID with `UUID()` before including it in the query. Invalid MBIDs are silently skipped. If no valid MBIDs remain after filtering, return an empty list early (avoiding the database roundtrip entirely).

# AI usage

#### If you did use AI:

### DIFF
--- a/listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
+++ b/listenbrainz/labs_api/labs/api/recording_from_recording_mbid.py
@@ -62,7 +62,15 @@ class RecordingFromRecordingMBIDQuery(Query):
         if not current_app.config["MB_DATABASE_URI"]:
             return []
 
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Another easy one with good results.
Good to go, thank you!

PR #3836, opened 2026-08-28
Fix leading and trailing whitespace issue in Add Listen search modal

deepanshu-singh22:
# Problem

Leading and trailing whitespaces in user search inputs and pasted URLs caused unnecessary whitespace-only queries and improper search triggers when auto-switching modes in the "Add Listen" modal.

# Solution

- **`AddListenModal.tsx`**: Trimmed `pastedURL` inside `switchMode` before updating search text state to ignore extra spaces.
- **`AddAlbumListens.tsx`**: Added a `.trim()` check to `initialText` inside the `useEffect` hook to prevent unwanted triggers during mode auto-switching.
- **`AddSingleListen.tsx`**: Added a `.trim()` check to `initialText` inside the `useEffect` hook to ensure clean search terms are processed.

# AI usage

* [x] I did not use any AI

#### If you did use AI:

# Action

No special post-merge action required.

### DIFF
--- a/frontend/js/src/user/components/AddAlbumListens.tsx
+++ b/frontend/js/src/user/components/AddAlbumListens.tsx
@@ -119,9 +119,10 @@ const AddAlbumListens = forwardRef(function AddAlbumListens(
   const initialTextRef = useRef(initialText);
   React.useEffect(() => {
     // Trigger search manually if auto-switching from album to recording search
-    if (initialText && initialTextRef.current !== initialText) {
-      searchInputRef.current?.triggerSearch(initialText);
-      initialTextRef.current = initialText;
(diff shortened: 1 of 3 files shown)


anshg1214 (changes requested):
The parent-level trim in `switchMode` is the right place for this - good call. 

The trims you added in AddSingleListen and AddAlbumListens are redundant since the parent already cleans the value before passing it down.

Please add trailing newlines back to all three files — they got stripped.

Fwiw the actual root cause is in SearchTrackOrMBID.tsx — inputValue stores the raw event.target.value, so the URL regex matching runs against untrimmed input too.

Trimming there would be a one-line fix that covers everything

PR #3837, opened 2026-08-29
LB-1989: Fix tag dropdown selecting wrong tag on Enter key

Akansh475:
## Problem
When typing a tag in the tag input (e.g. on artist pages), the tag dropdown filters by simple alphabetical order rather than relevance. If a user types a tag and presses Enter expecting to apply what they typed, the first item in the (alphabetically sorted) list gets applied instead — which is often the wrong tag. This was reported on the community forum after a user accidentally mistagged an artist with an incorrect sub-genre and had no way to undo it.

## Solution
Added a custom `filterOption` function using `match-sorter` (as suggested by @MonkeyDo in the ticket comments), so dropdown options are ranked by relevance to the typed input instead of plain alphabetical order. This means the tag the user actually typed is much more likely to appear first, reducing the chance of applying the wrong tag on Enter.

## Testing
Verified with `npx tsc --noEmit` (no errors) and `npx eslint` (no errors/warnings). Have not been able to test this visually in the running app this session (local Docker/OAuth environment not set up) — flagging this for reviewer awareness, happy to follow up with a screen recording if useful.

### DIFF
--- a/frontend/js/src/tags/TagsComponent.tsx
+++ b/frontend/js/src/tags/TagsComponent.tsx
@@ -1,6 +1,6 @@
 import { isFunction, isUndefined, noop, set, sortBy } from "lodash";
 import * as React from "react";
-import { useCallback, useState, useEffect } from "react";
+import { useCallback, useMemo, useState, useEffect } from "react";
 import {
   ActionMeta,
   DropdownIndicatorProps,
@@ -11,7 +11,6 @@ import {
 import CreatableSelect from "react-select/creatable";
 import { toast } from "react-toastify";
 import { faPlus } from "@fortawesome/free-solid-svg-icons";
-import { matchSorter } from "match-sorter";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import TagComponent, { TagActionType } from "./TagComponent";
 import GlobalAppContext from "../utils/GlobalAppContext";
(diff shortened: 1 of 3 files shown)


anshg1214 (changes requested):
AI Disclaimer please.

This PR needs a rethink. The current code adds a dependency which can be avoided by solving this natively.

Akansh475:
AI disclosure: Claude was used to help draft this fix and implement the match-sorter approach originally suggested by @MonkeyDo in the ticket comments. I'll rework this to avoid the added dependency per your feedback below.

Akansh475:
Reworked this to avoid the match-sorter dependency — removed it and replaced the filtering/sorting with a native implementation:

- `getMatchRank()` ranks each genre option against the typed input: exact match (0), starts-with (1), contains (2), or no match (-1)
- Options are filtered (dropping non-matches) and sorted by rank, then alphabetically within each rank, using `useMemo`
- Wired this into react-select via controlled `inputValue`/`onInputChange`, passing the pre-filtered/sorted list as `options` and setting `filterOption={null}` so react-select doesn't re-filter on top of it

This should give the same practical benefit (typed input ranks to the top instead of getting lost in plain alphabetical order) without adding a new dependency. Let me know if you'd prefer a different ranking heuristic or approach.

PR #3838, opened 2026-08-29
LB-1991: Add standalone Spotify import breakdown tool

NAME-ASHWANIYADAV:
## Problem

When users upload their Spotify Extended Streaming History, we report how many listens were imported but not *why* the rest were skipped. The skips are legitimate (under-30s interrupted plays, incognito sessions, podcast episodes), but users have no way to see this for themselves.

Fixes [LB-1991](https://tickets.metabrainz.org/browse/LB-1991)

## Solution

Adds `listenbrainz/tools/spotify_import_breakdown.py` - a standalone, stdlib-only script users can run against their Spotify export (zip, extracted folder, or individual JSON files) to preview exactly what an import would do:

* Classifies every entry with the same rules as the real importer (`background/listens_importer/spotify.py`): would-import, short/interrupted play, incognito, no track link (podcasts/videos/audiobooks), needs-Spotify-lookup, invalid - with per-reason detail counts and optional `--show-skipped N` examples.
* Ignores the same files the importer ignores (video history), and detects the simplified "Account data" export (`StreamingHistory_music_*.json`) with a warning pointing users to the Extended Streaming History - a common source of confusion.
* `--json` flag for machine-readable output.
* Zero dependencies beyond the Python standard library, so the single file can be downloaded and run on its own - per the ticket's suggestion of a standalone tool rather than running this analysis on every import.

The module docstring notes it must be kept in sync with the importer's skip rules.

### Testing

* 19 unit tests covering every classification path, the file filter, and zip/folder/single-file analysis.
* A parity check: 378 combinations of `ms_played`/`skipped`/`reason_end`/`incognito_mode` fed to both the real `SpotifyListensImporter._skip_item` and this tool's classifier - 0 mismatches.
* Manual run on a 1000-entry sample export (screenshots below).

Screenshot 2026-08-29 135622
Screenshot 2026-08-29 135650

- [x] I have run the code and manually tested the changes

## AI usage

- [x] I have used AI in this PR (add more details below)
- [x] I used AI tools for communication

The code was written with an AI coding assistant (Claude Code) working under my direction. The AI analysed the existing zip importer's skip rules and mirrored them in the standalone tool, including a parity test against the real importer. I reviewed every change and ran the tests and the CLI myself to verify the output.

## Action

None on merge - the tool is standalone and changes no server behaviour. Happy to add a link to it from the import page or docs if you'd like (follow-up).

### DIFF
--- a/listenbrainz/tests/unit/test_spotify_import_breakdown.py
+++ b/listenbrainz/tests/unit/test_spotify_import_breakdown.py
@@ -1,0 +1,164 @@
+import io
+import json
+import unittest
+import zipfile
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+from listenbrainz.tools.spotify_import_breakdown import (
+    Breakdown, IMPORTED, INCOGNITO, INVALID, NEEDS_LOOKUP, NO_TRACK_LINK,
+    SHORT_PLAY, analyze_path, classify_entry, is_account_data_history,
+    matches_import_filter,
+)
+
+
+def _entry(**overrides):
+    """ A well-formed extended-streaming-history entry that would be imported. """
+    entry = {
+        "ts": "2023-05-04T10:23:45Z",
+        "ms_played": 215000,
+        "master_metadata_track_name": "Immigrant Song",
+        "master_metadata_album_artist_name": "Led Zeppelin",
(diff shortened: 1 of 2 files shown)


NAME-ASHWANIYADAV:
@anshg1214 @MonkeyDo PTAL !!

PR #3839, opened 2026-08-30
feat: store user data exports in garage

amCap1712:
User data exports were written to a directory shared between the webserver and the background tasks container through the USER_DATA_EXPORT_BASE_DIR volume, which tied the two to the same host. They are uploaded to the garage bucket named by GARAGE_USER_DATA_EXPORT_BUCKET now and streamed back to the user from there, so the containers no longer need shared storage.

* [X] I have run the code and manually tested the changes on test.listenbrainz.org.
* [X] I have used AI in this PR for tests, reviewing the code changes and applying certain fixes based on the automated review.

### DIFF
--- a/docker/docker-compose.test.yml
+++ b/docker/docker-compose.test.yml
@@ -24,6 +24,21 @@ services:
   rabbitmq:
     image: rabbitmq:3.8.16
 
+  garage:
+    build:
+      context: ./garage
+    environment:
+      GARAGE_ACCESS_KEY: GK1e5b4c3a2d0f9e8b7a6c5d40
+      GARAGE_SECRET_KEY: 0f1e2d3c4b5a69788796a5b4c3d2e1f00f1e2d3c4b5a69788796a5b4c3d2e1f0
+      GARAGE_BUCKETS: listenbrainz-user-data-exports
+    healthcheck:
+      # the S3 api port is open long before the layout is applied and the buckets exist,
+      # the entrypoint only writes this file once provisioning is complete
+      test: ["CMD", "test", "-f", "/tmp/garage-provisioned"]
+      interval: 2s
+      timeout: 2s
+      retries: 30
+
   listenbrainz:
     build:
       context: ..
@@ -38,9 +53,14 @@ services:
       PYTHON_TESTS_RUNNING: 1
     user: "${LB_DOCKER_USER:-root}:${LB_DOCKER_GROUP:-root}"
     depends_on:
-      - redis
-      - lb_db
-      - rabbitmq
+      redis:
+        condition: service_started
+      lb_db:
+        condition: service_started
+      rabbitmq:
+        condition: service_started
+      garage:
+        condition: service_healthy
 
   timescale_writer:
     image: listenbrainz
@@ -57,11 +77,14 @@ services:
     image: listenbrainz
     command: python3 -m "listenbrainz.background.background_tasks"
     depends_on:
-      - redis
-      - lb_db
+      redis:
+        condition: service_started
+      lb_db:
+        condition: service_started
+      garage:
+        condition: service_healthy
     environment:
       PYTHON_TESTS_RUNNING: 1
-    restart: unless-stopped
     user: "${LB_DOCKER_USER:-root}:${LB_DOCKER_GROUP:-root}"
     volumes:
       - ..:/code/listenbrainz:z
--- a/docker/garage/Dockerfile
+++ b/docker/garage/Dockerfile
@@ -1,10 +1,10 @@
-ARG GARAGE_VERSION=v1.0.1
+ARG GARAGE_VERSION=v2.3.0
 
 FROM dxflrs/garage:${GARAGE_VERSION} AS upstream
 
 # The upstream image is built FROM scratch and has no shell, so the (statically
 # linked) binary is copied into an image that can run the provisioning script.
-FROM alpine:3.21
+FROM alpine:3.24
 
 COPY --from=upstream /garage /usr/local/bin/garage
 COPY garage.toml /etc/garage.toml
--- a/docker/garage/Dockerfile
+++ b/docker/garage/Dockerfile
@@ -1,0 +1,15 @@
+ARG GARAGE_VERSION=v1.0.1
+
+FROM dxflrs/garage:${GARAGE_VERSION} AS upstream
+
+# The upstream image is built FROM scratch and has no shell, so the (statically
+# linked) binary is copied into an image that can run the provisioning script.
+FROM alpine:3.21
+
+COPY --from=upstream /garage /usr/local/bin/garage
+COPY garage.toml /etc/garage.toml
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+
+EXPOSE 3900 3901 3903
+
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
--- a/docker/garage/garage.toml
+++ b/docker/garage/garage.toml
@@ -1,0 +1,19 @@
+# Single node Garage setup for local development and tests. Do not use in production.
+metadata_dir = "/var/lib/garage/meta"
+data_dir = "/var/lib/garage/data"
+db_engine = "sqlite"
+
+replication_factor = 1
+
+rpc_bind_addr = "[::]:3901"
+rpc_public_addr = "127.0.0.1:3901"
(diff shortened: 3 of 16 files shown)


julian45:
Glad to see this!

FWIW, I have a minor suggestion here: use AWS's SDK `boto3` instead of MinIO's SDK for S3-compatible operations against Garage. Although from a birds-eye PoV the two SDKs might be seen as covering the same ground:
- IMO any output of the MinIO project is risky to rely on due to their closed-source turn a while back.
- On that note, MinIO's SDK hasn't seen a release in 9 months and lacks support for e.g. Python 3.14.
- On the other hand, `boto3` is quite actively maintained and includes support not only for 3.14, but also free-threading should we wish to try it. As far as FOSS concerns go, this SDK *is* licensed under Apache-2.0.

Thoughts?

docker/garage/Dockerfile
```diff
@@ -0,0 +1,15 @@
+ARG GARAGE_VERSION=v1.0.1
```
julian45: This version of Garage is *very* stale, at almost two years old. The latest stable release is [v2.3.0](https://git.deuxfleurs.fr/Deuxfleurs/garage/releases/tag/v2.3.0), and I've confirmed that it's [available on Docker Hub as well](https://hub.docker.com/r/dxflrs/garage/tags?name=v2.3.0).
```suggestion
ARG GARAGE_VERSION=v2.3.0
```

docker/garage/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+FROM dxflrs/garage:${GARAGE_VERSION} AS upstream
+
+# The upstream image is built FROM scratch and has no shell, so the (statically
+# linked) binary is copied into an image that can run the provisioning script.
+FROM alpine:3.21
```
julian45: Also stale, and support for it [ends November 1st](https://alpinelinux.org/releases/).
```suggestion
FROM alpine:3.24
```

julian45 (changes requested):
Just a couple of small things for making sure deps are up-to-date.

listenbrainz/config.py.sample
```diff
@@ -239,8 +239,14 @@ REJECT_LISTENS_WITHOUT_USER_EMAIL = False
 # If set to True, do not allow new users without email to register
 REJECT_NEW_USERS_WITHOUT_EMAIL = False
 
-# base directory for user data exports
-USER_DATA_EXPORT_BASE_DIR = "/code/listenbrainz/exports/"
+# Garage object storage, user data exports are stored here.
+# The endpoint must not include a URL scheme.
+GARAGE_ENDPOINT = "garage:3900"
+GARAGE_ACCESS_KEY = "GK1e5b4c3a2d0f9e8b7a6c5d40"
+GARAGE_SECRET_KEY = "0f1e2d3c4b5a69788796a5b4c3d2e1f00f1e2d3c4b5a69788796a5b4c3d2e1f0"
```
MonkeyDo, line 247: Just checking these are strictly for local setup and we are not leaking anything we are also using in prod

listenbrainz/manage.py
```diff
@@ -392,3 +393,20 @@ def delete_old_user_data_exports():
...
+@click.option("--dry-run", is_flag=True, help="Log what would be migrated without uploading or updating anything.")
+@click.option("--mark-missing-failed", is_flag=True,
+              help="Mark completed exports whose archive cannot be found as failed even if the directory"
+                   " contained no archive of a completed export at all.")
+def migrate_user_data_exports(export_dir, delete_source, dry_run, mark_missing_failed):
+    """ Migrate user data exports from the given directory to garage """
+    app = create_app()
+    with app.app_context():
+        app.logger.info("Migrating user data exports from %s to garage", export_dir)
+        migrate_exports(webserver.db_conn, export_dir, delete_source=delete_source, dry_run=dry_run,
```
MonkeyDo: Just some style I guess
```suggestion
        migrate_exports(webserver.db_conn, export_dir=export_dir, delete_source=delete_source, dry_run=dry_run,
```

docker/garage/garage.toml
```diff
@@ -0,0 +1,19 @@
+# Single node Garage setup for local development and tests. Do not use in production.
+metadata_dir = "/var/lib/garage/meta"
+data_dir = "/var/lib/garage/data"
+db_engine = "sqlite"
+
+replication_factor = 1
+
+rpc_bind_addr = "[::]:3901"
+rpc_public_addr = "127.0.0.1:3901"
+rpc_secret = "b8f1a1a4c8d94e4db0f8b1b0a0d8f5e2a7c3b9d1e4f60718293a4b5c6d7e8f90"
```
MonkeyDo, line 10: Same check here that thisis not used anywhere in prod, just in case.

docker/docker-compose.test.yml
```diff
@@ -24,6 +24,21 @@ services:
   rabbitmq:
     image: rabbitmq:3.8.16
 
+  garage:
+    build:
+      context: ./garage
+    environment:
+      GARAGE_ACCESS_KEY: GK1e5b4c3a2d0f9e8b7a6c5d40
+      GARAGE_SECRET_KEY: 0f1e2d3c4b5a69788796a5b4c3d2e1f00f1e2d3c4b5a69788796a5b4c3d2e1f0
```
MonkeyDo, line 32: Are those not read from config.py? Or are they necessary for initial setup first in the entrypoint script, then later by the app to auth with the bucket?

listenbrainz/background/export.py
```diff
@@ -281,43 +298,50 @@ def export_user(db_conn, ts_conn, user_id: int, metadata):
...
+                    all_files.append(pinned_recording_file)
+
+                update_export_progress(db_conn, export_id, "Writing export files")
+                for file in all_files:
+                    archive.write(file, arcname=os.path.relpath(file, tmp_dir))
+
+            update_export_progress(db_conn, export_id, "Finalizing user data export")
+            client = get_garage_client()
+            bucket = get_user_data_export_bucket()
+            ensure_bucket(client, bucket)
```
MonkeyDo: I wonder if those should be invoke at the very top of this try block, so that we ensure the destination bucket is available (or raises) before we do any of the other work

listenbrainz/webserver/views/export.py
```diff
@@ -121,8 +144,30 @@ def download_export_archive(export_id):
...
+        finally:
+            body.close()
+
+    headers = Headers()
+    headers.set("Content-Disposition", "attachment", **content_disposition_names(filename))
+    content_length = archive.get("ContentLength")
+    if content_length is not None:
+        headers.set("Content-Length", str(content_length))
+
+    return Response(stream_archive(), mimetype="application/zip", headers=headers)
```
MonkeyDo: Painful to have to do all this, but after a quick search it looks like there aren't really much better options.

Just checking that there is no way to pass the body to send_file directly, as it may be a "file-like object" (https://flask.palletsprojects.com/en/stable/api/#flask.send_file)

MonkeyDo (commented):
Thanks for working on this!

I would request the addition of some documentation in the maintainer docs.
Happy to do a video call so you can talk me through everything as I ask questions, and can transcribe as docs (if that is helpful)

listenbrainz/config.py.sample
amCap1712: yes this is for local only, the alternative would be requiring contributors to create a .env file.

docker/garage/garage.toml
amCap1712: not used in prod. prod garage is setup in https://github.com/metabrainz/metabrainz-ansible/pull/106

docker/docker-compose.test.yml
amCap1712: This is for the garage entrypoint yes. Alternative is parsing config.py in shell script and then executing commands in garage container.

listenbrainz/background/export.py
amCap1712: Fixed.

listenbrainz/webserver/views/export.py
amCap1712: With boto3, it should be possible. refactored.

julian45 (approved):
Thanks! Approved w.r.t. the specific things I touched on; up to @MonkeyDo to approve re: his portions.

MonkeyDo (commented):
I pushed a small change (allowing users to generate a new export if they have a failed one).

PR looks good, but there are some export-related failing tests to look at. If I had to guess I would imagine it is because garage might not be ready on time when the test suite expects it, but that is just a theory

MonkeyDo (approved):

PR #3841, opened 2026-09-01
Add documentation about building the documentation

MonkeyDo:
# Problem
There are a couple of lines in the readme but there should be some in the developer docs for those who look for it

# Solution
Add some docas about building docs...
I added a note about python version because the slightly older one I had installed caused me some issues with dependencies.

# AI usage

* [x] I did not use any AI

image

### DIFF
--- a/docs/developers/docs.rst
+++ b/docs/developers/docs.rst
@@ -1,0 +1,27 @@
+This Documentation
+==================
+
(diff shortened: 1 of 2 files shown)


MonkeyDo:
@RustyNova016 Could you please review this one and tell me if they are clear enough or missing something?

PR #3842, opened 2026-09-01
docs: Add user-agent section, review rate limiting, add a warning about both

MonkeyDo:
Like it says on the tin, added a section about proper user-agent strings, reviewed the rate-limiting section, and reorganized the page a bit so that they are more prominent (but also put the reference ToC higher up).

I also added a warning at the top so people don't skip it:
image

image

* [x] I did not use any AI

### DIFF
--- a/docs/users/api/index.rst
+++ b/docs/users/api/index.rst
@@ -5,10 +5,33 @@ All endpoints have this root URL for our current production site.
 
 - **API Root URL**: ``https://api.listenbrainz.org``
 
+.. warning::
+    To avoid being blocked or throttled, read carefully the `Rate limiting <rate-limiting_>`_
+    and `User-agent <user-agent_>`_ sections below to be a good citizen.
+
 .. note::
-    All ListenBrainz services are only available on **HTTPS**!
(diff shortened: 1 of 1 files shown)


docs/users/api/index.rst
```diff
@@ -5,10 +5,33 @@ All endpoints have this root URL for our current production site.
 
 - **API Root URL**: ``https://api.listenbrainz.org``
 
+.. warning::
+    To avoid being blocked or throttled, read carefully the `Rate limiting <rate-limiting_>`_
+    and `User-agent <user-agent_>`_ sections below to be a good citizen.
+
 .. note::
-    All ListenBrainz services are only available on **HTTPS**!
+    All ListenBrainz services are only available on **HTTPS** only!
```
anshg1214: Let's revert these changes. We have two `only` in this line.

anshg1214 (approved):
Minor comment, otherwise looks good.

docs/users/api/index.rst
MonkeyDo: Woops! Thanks for catching that only

PR #3847, opened 2026-09-03
fix: update pinned recording MBID when user submits manual mapping (LB-1985)

anshg1214:
# Problem

LB-1985

When MusicBrainz merges two recordings, any pinned recording that referenced the old recording MBID becomes stale. The pin still displays but shows an unlinked state because the MBID it stores no longer exists in MusicBrainz.

When a user tries to fix this by clicking "Link with MusicBrainz" and selecting the correct recording, the `submit_manual_mapping` endpoint updates the `mbid_manual_mapping` table, but the `pinned_recording` table has its own `recording_mbid` column that is not updated. On page refresh, the pin still shows the old invalid MBID.

# Solution

- Add `update_recording_mbid_for_msid()` to `listenbrainz/db/pinned_recording.py` that updates `recording_mbid` on all pins for a user matching a given `recording_msid`
- Call it from `submit_manual_mapping()` in `metadata_api.py` after creating the manual mapping, so relinking a recording also fixes any matching pins

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

# Action

None needed.

### DIFF
--- a/listenbrainz/webserver/views/metadata_api.py
+++ b/listenbrainz/webserver/views/metadata_api.py
@@ -4,6 +4,7 @@ from datasethoster import RequestSource
 from flask import Blueprint, request, jsonify, current_app
 
 from listenbrainz.db.mbid_manual_mapping import create_mbid_manual_mapping, get_mbid_manual_mapping
+from listenbrainz.db.pinned_recording import update_recording_mbid_for_msid
 from listenbrainz.db.metadata import get_metadata_for_recording, get_metadata_for_artist, get_metadata_for_release_group
 from listenbrainz.db.model.mbid_manual_mapping import MbidManualMapping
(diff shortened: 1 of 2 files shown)


listenbrainz/webserver/views/metadata_api.py
```diff
@@ -524,6 +525,7 @@ def submit_manual_mapping():
     )
 
     create_mbid_manual_mapping(ts_conn, mapping)
+    update_recording_mbid_for_msid(db_conn, user["id"], recording_msid, recording_mbid)
```
MonkeyDo, line 528: Out of the context of the pinned_recording file, it's not clear that it applies to pins. Would suggest something like:

```suggestion
    update_pinned_recording_mbid_for_msid(db_conn, user["id"], recording_msid, recording_mbid)
```

MonkeyDo (commented):
Yep, seems sensible.

PR #3849, opened 2026-09-05
LB-1992: Name the streaming service in artist and album link tooltips

kunalKumar-13:
# Problem

LB-1992.

On artist and album pages, every link MusicBrainz classes as `streaming` or
`free streaming` renders as the same generic music note, and its tooltip is the
relationship type — the word "streaming" — rather than the service. As the
reporter put it, you cannot tell where the icon leads until you follow it and
find out it was Qobuz.

# Solution

Identify the service from the URL's **hostname** and use its name as the tooltip,
falling back to the relationship name when the host is not recognised. Where Font
Awesome provides a brand icon, use it and its colour as well.

Worth stating explicitly: Font Awesome has no Qobuz or Tidal icon, so a
fix based on icons alone cannot solve the reported case. Naming the service in
the tooltip is what answers "where does this go?"; the brand icons are a bonus
where they exist.

Matching on the hostname rather than the whole URL keeps a service name in a path
from being mistaken for the host — `https://example.com/artist/tidal-tribute` is
not Tidal. There is a test for that case. This follows the existing
`social network` branch, which already chooses its icon by inspecting the URL.

Tests: 8 added — 5 for the lookup, 3 rendering the link and asserting the tooltip.
Reverting only `title={title}` back to `title={relName}` fails exactly the one
test that asserts the visible behaviour.

Full suite is unchanged against master: 6 suites / 13 tests fail on both, all
pre-existing and unrelated (HueSound, Playlist, Recommendations, NetworkFeed,
UserFeed, UserReports); this branch adds 8 passing tests on top.

* [ ] I have run the code and manually tested the changes

# AI usage

* [ ] I did not use any AI
* [x] I have used AI in this PR (add more details below)

* [x] I used AI tools for communication
* [x] I used AI tools for coding
* [ ] I understand all the changes made in this PR

Claude Code was used to investigate the ticket, write the change and its tests,
and draft this description. Every number quoted above came from running the code.

Two boxes are left unticked deliberately, because ticking them would not be true
yet, and your policy is the reason I would rather say so plainly than have you
find out in review:

- **"manually tested"** — I ran the test suite, including a test that renders the
  link and asserts the tooltip, but I have not run a local ListenBrainz instance
  and clicked an artist page.
- **"I understand all the changes"** — I have not yet read through the diff line
  by line myself.

I would rather submit this honestly and let you decide what it is worth than
tick boxes to get it reviewed. If the honest answer is that you would rather not
take AI-assisted contributions in this state, say so and I will close it without
[...]

PR #3853, opened 2026-09-07
chore: upgrade FontAwesome to v7

kunalKumar-13:
# Problem

FontAwesome is pinned to v6, so icons added in v7 aren't available — Tidal being the one that prompted this, in #3849.

Split out of #3849 at @MonkeyDo's request, since it needs changes unrelated to that PR.

# Solution

Bumps the five `@fortawesome` packages from `^6.4.0` to `^7.3.1`.

`react-fontawesome` had to move too: `0.2.0` peers on `fontawesome-svg-core: ~1 || ~6`, so npm rejects it against v7. `3.5.0` peers `~6 || ~7` and React `^18 || ^19`, which matches this project.

**No source changes.** I checked all 160 `fa*` identifiers imported across `frontend/js/src` against v7 — every one still exists. 11 look missing at first, but none of them were FontAwesome exports in v6 either: they're our own custom icons (`faQobuz`, `faNavidrome`, `faFunkwhale`) and local aliases like `faGrid`, `faStacked`, `faThumbsUpRegular`.

**One test needed updating.** v7 no longer renders a `<title>` element inside the icon SVG, so `MusicPlayer.test.tsx`'s `getByText("Love")` / `getByText("Hate")` stopped matching — that text only ever came from the SVG title. Those buttons already set `aria-label={title}`, so the queries move to `getByLabelText`.

Worth flagging: that change only works on v7. On v6 it fails with *"Found multiple elements with the text of: Love"*, because there both the `<title>` and the `aria-label` match. So the test change and the bump can't be split.

# Testing

`TZ='Europe/London' jest`, which is the project's `npm run test`:

```
master (v6):        6 failing suites, 13 tests
this branch (v7):   6 failing suites, 13 tests
```

Same six either way — HueSound, Playlist, Recommendations, NetworkFeed, UserFeed, UserReports — all pre-existing on master and unrelated to FontAwesome. Without the `MusicPlayer` fix it's 7, which is how I found it.

`tsc --noEmit` passes.

One thing I left alone: `MusicPlayer.test.tsx` fails `prettier --check` and has 6 eslint warnings, but it does so on master too, so I haven't reformatted it here.

# AI usage

* [ ] I did not use any AI

Claude Code was used for the investigation and to draft this description. The icon audit, the peer-dependency check and the before/after test runs were all executed.


MonkeyDo (approved):
Looking good, thank you!

PR #3855, opened 2026-09-08
Store user data imports in garage

amCap1712:
Store uploaded user data import files in a dedicated Garage bucket instead of a shared local directory. Download files to temporary storage while an import runs and remove them after completion, cancellation, or failure. Add commands to migrate existing uploads and clean up stale objects.

* [X] I used AI tools for coding
* [X] I understand all the changes made in this PR

### DIFF
--- a/listenbrainz/background/export.py
+++ b/listenbrainz/background/export.py
@@ -280,10 +280,6 @@ def export_user(db_conn, ts_conn, user_id: int, metadata):
         current_app.logger.error("No export with export_id: %s, skipping.", metadata["export_id"])
         return
 
-    client = get_garage_client()
-    bucket = get_user_data_export_bucket()
-    ensure_bucket(client, bucket)
-
     export_id = export.id
     try:
         client = get_garage_client()
--- a/listenbrainz/background/export.py
+++ b/listenbrainz/background/export.py
@@ -280,6 +280,10 @@ def export_user(db_conn, ts_conn, user_id: int, metadata):
         current_app.logger.error("No export with export_id: %s, skipping.", metadata["export_id"])
         return
 
+    client = get_garage_client()
+    bucket = get_user_data_export_bucket()
+    ensure_bucket(client, bucket)
+
     export_id = export.id
     try:
         client = get_garage_client()
--- a/listenbrainz/background/listens_importer/base.py
+++ b/listenbrainz/background/listens_importer/base.py
@@ -11,7 +11,8 @@ from flask import current_app
 from sqlalchemy.sql.expression import text
 from werkzeug.exceptions import InternalServerError, ServiceUnavailable
 
-from listenbrainz.background.listens_importer.storage import delete_import_file, download_import_file
+from listenbrainz.background.listens_importer.storage import delete_import_file, download_import_file, \
+    FINISHED_STATUSES
 from listenbrainz.db import user as db_user
 from listenbrainz.domain.external_service import ExternalServiceError
 from listenbrainz.webserver.errors import ImportFailedError, ListenValidationError
@@ -64,7 +65,7 @@ class BaseListensImporter(ABC):
 
         import_id = import_task["id"]
         metadata = import_task["metadata"]
-        if metadata["status"] in {"cancelled", "failed"}:
+        if metadata["status"] in FINISHED_STATUSES:
             # the import is not going to run, its file is of no use to anyone any more
             delete_import_file(import_task["file_path"])
             return
--- a/listenbrainz/background/listens_importer/base.py
+++ b/listenbrainz/background/listens_importer/base.py
@@ -1,9 +1,9 @@
 import json
+import tempfile
 import zipfile
 from abc import ABC, abstractmethod
 from datetime import datetime
 from io import TextIOWrapper
-from pathlib import Path
 from typing import Any, Iterator
 from zipfile import ZipFile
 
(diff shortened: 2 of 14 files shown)


listenbrainz/background/listens_importer/base.py
```diff
@@ -64,27 +65,35 @@ def import_listens(self, user_id: int, import_task: dict[str, Any]) -> None:
...
-            validation_stats = self._initialize_validation_stats()
-            self.persist_validation_stats(import_id, validation_stats)
-            for batch in self.process_import_file(import_task):
-                validated_listens, validation_stats = self.parse_and_validate_listen_items(
-                    batch,
-                    validation_stats,
-                )
-
-                self.submit_listens(validated_listens, user_id, user["musicbrainz_id"], import_id)
+            with tempfile.TemporaryDirectory() as tmp_dir:
```
MonkeyDo, line 77: Just checking: this temp directory and the file within don't need any specific calls to be removed, and will automatically be cleaned up at the end of this context window?

listenbrainz/background/listens_importer/storage.py
```diff
@@ -0,0 +1,102 @@
...
+    The importer deletes the file of an import as soon as it is done with it, but a file is
+    left behind whenever it does not get that far: the import was cancelled or had already
+    failed when the task ran, or the row the upload belongs to was never committed.
+    """
+    result = db_conn.execute(text("""
+        SELECT file_path
+          FROM user_data_import
+         WHERE metadata->>'status' = ANY(:statuses)
+           AND file_path IS NOT NULL
+    """), {"statuses": list(PENDING_STATUSES)})
```
MonkeyDo: This could potentially be brittle, although I don't think we update the list of statuses often...
Would it make sense to be a blacklist instead of a whitelist, i.e. something along the lines of `NOT IN ["cancelled", "failed"]` like we use [in listens_importer/base.py](https://github.com/metabrainz/listenbrainz-server/pull/3855/changes#diff-64aa9dc73c155bf399153771697f1adb58b5c8d7e5f1c692e9a7680178aa7334R68-R69)?

listenbrainz/background/export.py
```diff
@@ -280,6 +280,10 @@ def export_user(db_conn, ts_conn, user_id: int, metadata):
         current_app.logger.error("No export with export_id: %s, skipping.", metadata["export_id"])
         return
 
+    client = get_garage_client()
+    bucket = get_user_data_export_bucket()
+    ensure_bucket(client, bucket)
+
```
MonkeyDo: Duplicates lines just below in the try block

listenbrainz/background/migrate_imports.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def get_pending_imports(db_conn) -> dict[str, list[int]]:
+    """ Get the import ids of all imports that have not run yet, keyed by the file's object name. """
+    result = db_conn.execute(text("""
+        SELECT id, file_path
+          FROM user_data_import
+         WHERE metadata->>'status' = ANY(:statuses)
+           AND file_path IS NOT NULL
+    """), {"statuses": list(PENDING_STATUSES)})
```
MonkeyDo: Depending on the resolution of the comment above in cleanup_import_files, this should probably follow the same pattern

listenbrainz/background/migrate_imports.py
```diff
@@ -0,0 +1,174 @@
...
+        # that looked like an orphan into the file of a pending import that was never uploaded
+        known_now = get_all_import_filenames(db_conn)
+        for path in orphans:
+            if path.name in known_now:
+                current_app.logger.warning(
+                    "An import was created for %s while the migration ran, keeping the file."
+                    " Run the migration again to upload it.", path.name
+                )
+            else:
+                path.unlink(missing_ok=True)
```
MonkeyDo: I guess this could instead be since we unlink all deletables just underneath, but it's really all the same 
```suggestion
                deletable.append(path)
```

MonkeyDo (commented):
Look good!
As with the exports PR, this should be tested thoroughly on the test website/background-tasks containers.

A general observation: we will probably want to do a cleanup PR once everything is migrated, to remove the  migration scripts from the codebase and any mention of it/tests/code that is related to it, just to keep things clean.
Git history will still be there in any case.

MonkeyDo:
There is a fix for the unrelated flaky failing test in an open PR: https://github.com/metabrainz/listenbrainz-server/pull/3831/changes/2b2d1c3b7e7c6b3eb5c9913a809108c2e793c02e

listenbrainz/background/listens_importer/base.py
amCap1712: Yes, `TemporaryDirectory` removes the directory and everything in it when the `with` block exits, on the exception path too.

listenbrainz/background/listens_importer/storage.py
amCap1712: Good point, switched it around.

listenbrainz/background/export.py
amCap1712: Removed, that was left over from moving the client setup around.

listenbrainz/background/migrate_imports.py
amCap1712: Done.

listenbrainz/background/migrate_imports.py
amCap1712: Applied.

amCap1712:
Thanks for the review! Pushed 927420fdb addressing the inline comments.

Agreed on the follow-up cleanup PR — once the migration has run on production, `migrate_imports.py`, the `migrate_user_data_imports` command, the `object_name_for` basename fallback and the related tests can all go.

amCap1712:
tested on test.lb.

PR #3857, opened 2026-09-09
fix: Resolve crashes and warnings in api_compat session_info

MIHIRCHAUDHARI390:
# Problem

Multiple bugs existed in `api_compat` that could lead to 500 crashes instead of clean error returns:

1. **UnboundLocalError**: In `session_info`, `output_format` was assigned *after* parsing parameters that could raise a `KeyError`. Thus, if `data['sk']` was missing, the `except KeyError:` block tried to use `output_format` before it was assigned, causing an `UnboundLocalError`.
2. **AttributeError**: `User.load_by_name(db_conn, username)` could return `None` for a nonexistent user, but its `.id` was accessed without checking.
3. **Stray `print` statement**: `print("SESSION INFO ...")` was left in production code instead of using `current_app.logger.debug()`.
4. **SyntaxWarning**: `re.match('(.*)\[(\d+)\]', key)` triggered a Python invalid escape sequence warning.

# Solution

1. Moved the assignment of `output_format` *before* the `try` block in `session_info` (consistent with how it's done in `record_listens`).
2. Separated the `User.load_by_name` call and added a `not user` check before checking `user.id`.
3. Converted the stray `print` to `current_app.logger.debug`.
4. Changed the regex pattern string to a raw string (`r'(.*)\[(\d+)\]'`).

### Files changed

- `listenbrainz/webserver/views/api_compat.py`

# AI usage

* [ ] I did not use any AI

#### If you did use AI:


MonkeyDo (commented):
Changes look sensible.

When you ticked the box saying "I have run the code and manually tested the changes", how exactly did you test the changes in this PR?

PR #3858, opened 2026-09-09
test(frontend): cover coverArtCache

kunalKumar-13:
# Problem

`frontend/js/src/utils/coverArtCache.ts` has no tests.

It keeps two localforage stores in step — the artwork itself and a parallel
expiry timestamp — and that pairing is the whole point of the module. Several
of its branches only matter when the two stores disagree (an entry whose expiry
half never landed, an expiry that cannot be read), and none of that was
exercised.

# Solution

Ten tests that drive both stores and assert they stay consistent:

- a write stores the artwork and an expiry of `now + 7 days` together
- an empty key or value writes nothing to either store
- a live entry returns its artwork and removes nothing
- an expired entry is dropped from **both** stores, returns null, and the
  artwork is never read once the entry is known to be stale
- a missing expiry counts as expired rather than as live forever, so an entry
  whose expiry half never landed still gets evicted
- an empty key returns null without touching either store
- a store failure returns null rather than throwing
- the background sweep removes only the expired keys, from both stores
- an expiry that cannot be read is treated as expired

| | before | after |
|---|---|---|
| Statements | 0% | **95%** (38/40) |
| Branches | 0% | **100%** (11/11) |
| Functions | 0% | **100%** (7/7) |

Checked against mutations rather than only run — each fails **only** the tests
that assert that behaviour:

| mutation | test that catches it |
|---|---|
| treat a missing expiry as live | treats a missing expiry as expired… |
| drop the empty key/value guard | writes nothing for an empty key / value |
| make the sweep mark nothing expired | removes only the expired keys… |

One note on the setup: the suite mocks localforage locally rather than using
the global mock in `jest-setup.ts`, because that one returns a fresh object per
`createInstance` with no handle to assert against, and does not stub `keys()`,
which the sweep calls.

Ran locally: `npx jest frontend/js/tests/utils/coverArtCache.test.ts` — 10
passed. ESLint and Prettier clean, `tsc --noEmit` reports nothing for these
files.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

Claude Code was used to read `coverArtCache.ts`, write the tests, run them and
the mutation checks, and draft this description. The module is 105 lines and
the tests assert only its documented behaviour; every result quoted above was
produced by running the suite locally, not asserted from reading.

### DIFF
--- a/frontend/js/tests/utils/coverArtCache.test.ts
+++ b/frontend/js/tests/utils/coverArtCache.test.ts
@@ -1,0 +1,157 @@
+/*
+ * coverArtCache keeps two localforage stores in step: the artwork itself and a
+ * parallel expiry timestamp. The pairing is the whole point of the module, so
+ * these tests drive both stores and assert they stay consistent.
+ */
+
+// Two instances are created at module load — artwork first, expiry second —
+// so they are captured in that order and handed back to the tests.
+const instances: Array<Record<string, jest.Mock>> = [];
+const makeInstance = () => {
+  const inst = {
+    setItem: jest.fn().mockResolvedValue(undefined),
+    getItem: jest.fn().mockResolvedValue(null),
+    removeItem: jest.fn().mockResolvedValue(undefined),
+    keys: jest.fn().mockResolvedValue([]),
+  };
+  instances.push(inst);
+  return inst;
+};
+
+jest.mock("localforage", () => ({
+  __esModule: true,
+  default: {
+    createInstance: jest.fn(() => makeInstance()),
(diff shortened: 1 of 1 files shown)


kunalKumar-13:
Three test-only PRs from me are sitting here — this one, #3859 and #3860. All three are one new file each, no production code touched, green, and GitHub reports all three `MERGEABLE` / `CLEAN`.

Three separate PRs is more review overhead than the change deserves, so if it is easier I am happy to close #3859 and #3860 and roll all three files into this one. Say the word and I will; otherwise they can be merged in any order, there is no dependency between them.

PR #3650, opened 2026-03-21
LB-1532: Add API endpoints for user data export

pranavsource1:
Superseded by #3864 because I didn't manage to push to the fork this PR is based off of. - Monkey

# Problem

Users need a way to export their ListenBrainz data (listens, feedback, etc.) programmatically or via the web interface. Currently, there is no dedicated API to trigger and retrieve these exports asynchronously.

# Solution

Implemented a new User Data Export API to handle asynchronous data exports.

*   **New API Endpoints** (`/1/export/`):
    *   `POST /`: Triggers a new export task.
    *   `GET /`: Lists historical exports for the authenticated user.
    *   `GET /<export_id>`: Retrieves status and details of a specific export.
    *   `GET /<export_id>/download`: Redirects to the download URL for the exported file.
*   **Backend Changes**:
    *   Added `listenbrainz/webserver/views/export_api.py` containing the `ExportAPI` view class.
    *   Registered the new blueprint in `listenbrainz/webserver/__init__.py`.
*   **Testing**:
    *   Added `listenbrainz/webserver/views/test/test_export_api.py`.
    *   Verified functionality with integration tests covering creation, listing, retrieval, and authorization.

* [x] I have run the code and manually tested the changes

# AI usage

* [x] I did not use any AI
* [ ] I have used AI in this PR

#### If you did use AI:
* [ ] I used AI tools for communication
* [ ] I used AI tools for coding
* [ ] I understand all the changes made in this PR

# Action 
None

PR #3864, opened 2026-09-11
LB-1532: Add API endpoints for user data export

MonkeyDo:
This is built on top of PR #3650 and replaces it.
For some reason I ended up borking my local git copy and not being able to push to @pranavsource1 's fork. So I am opening this new PR, but his commits (as well as Ansh's) can be found below (with myself as co-commiter).

My thanks to @pranavsource1 for the good work accomplished here. I took it over following the changes in #3839, after which the PR required a bunch of changes to adapt to the new storage system.
I added a couple of improvements of my own while I was there, but mostly just adapting the existing code from both PRs.

# Problem

We have an increasing need of offering a programmatic way for users to request and download the entirety of their data.
With MCPs and third-party projects being created every month, this will offer an easier way to fetch a user's history than hammering our API and paginating through the entire history of the user one by one.
 
# Solution

Offer an API endpoint allowing users to:
- request an export
- list exports
- get details of a specific export
- delete an ongoing export
- download a completed export file

Tested each one locally with [Bruno](https://www.usebruno.com/) and using the website page

# AI usage

* [x] I used AI for generating the documentation and for review

# Action

We should also add some comprehensive tests and make an announcement that any third-party wanting extensive history for a user should prefer this.

I am not sure if it makes sense to do so, but we could also limit the number of sequential calls to the /users/%USERNAME/listens API endpoint after some number of calls (say 50 or whatever), and return an error that requests the caller use the export API endpoint instead to get it all in one go.
Something to ponder but out of scope for this PR

### DIFF
--- a/listenbrainz/db/user_data_export.py
+++ b/listenbrainz/db/user_data_export.py
@@ -1,0 +1,108 @@
+import json
+from typing import Optional, Dict, Any, List
+from sqlalchemy import text
+
+
+def request_user_data_export(db_conn, user_id: int) -> Optional[Dict[str, Any]]:
+    """ Add a request to export the user data to an archive in background. """
+    query = """
+        INSERT INTO user_data_export (user_id, type, status, progress)
+             VALUES (:user_id, :type, 'waiting', :progress)
+        ON CONFLICT (user_id, type)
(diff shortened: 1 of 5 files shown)


amCap1712 (approved):
LGTM.

PR #3865, opened 2026-09-11
Use fully qualified image specifications for containerfiles

mibmo:
# Problem
The Containerfiles assume what registry is used. This means container building tooling has to guess at which to use, so e.g. buildah fails.

This shouldn't really change anything, except make more tools work. Would personally like them fully qualified as I'm working on something to make deploying LBS locally simpler.

# Solution
Seemed to me like the assumed registry is `docker.io`, so I changed all `FROM` instructions **and `org.metabrainz.based-on-image` labels** to use the fully qualified path.

# Action
- [ ] Seems to build fine for me, but could someone do a basic sanity check?
- [ ] This also changes the `org.metabrainz.based-on-image` label. I'm unsure where this fits into the infrastructure (monitoring?), so would like to know whether it's should be included or not.
- [ ] Any other containerfiles in the repo? I tried to find them using `rg --files-with-matches '^FROM'`.

# AI usage
no.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ARG PYTHON_BASE_IMAGE_VERSION=3.13-20250313
 ARG NODE_VERSION=20-alpine
-FROM metabrainz/python:$PYTHON_BASE_IMAGE_VERSION AS listenbrainz-base
+FROM docker.io/metabrainz/python:$PYTHON_BASE_IMAGE_VERSION AS listenbrainz-base
 
 ARG PYTHON_BASE_IMAGE_VERSION
 
@@ -9,7 +9,7 @@ LABEL org.label-schema.vcs-url="https://github.com/metabrainz/listenbrainz-serve
       org.label-schema.schema-version="1.0.0-rc1" \
       org.label-schema.vendor="MetaBrainz Foundation" \
       org.label-schema.name="ListenBrainz" \
-      org.metabrainz.based-on-image="metabrainz/python:$PYTHON_BASE_IMAGE_VERSION"
+      org.metabrainz.based-on-image="docker.io/metabrainz/python:$PYTHON_BASE_IMAGE_VERSION"
 
 ENV DOCKERIZE_VERSION=v0.6.1
 RUN wget https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -9,7 +9,7 @@ LABEL org.label-schema.vcs-url="https://github.com/metabrainz/listenbrainz-serve
       org.label-schema.schema-version="1.0.0-rc1" \
       org.label-schema.vendor="MetaBrainz Foundation" \
       org.label-schema.name="ListenBrainz" \
-      org.metabrainz.based-on-image="metabrainz/python:$PYTHON_BASE_IMAGE_VERSION"
+      org.metabrainz.based-on-image="docker.io/metabrainz/python:$PYTHON_BASE_IMAGE_VERSION"
```
MonkeyDo: Paging @zas :
Does this make sense to fully qualify the docker image name for use in this based-on-image tag?
I don't know how they are used.

mbid_mapping/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM metabrainz/python:3.13-20250313 as mbid-mapping-base
+FROM docker.io/metabrainz/python:3.13-20250313 as mbid-mapping-base
```
MonkeyDo: I'm seeing a linting warning about mismatched casing, so we might as well fix it while we're here
```suggestion
FROM docker.io/metabrainz/python:3.13-20250313 AS mbid-mapping-base
```

mbid_mapping/Dockerfile
MonkeyDo: Also here please

```suggestion
FROM mbid-mapping-base AS mbid-mapping-prod
```

MonkeyDo (commented):
Built all the modified images with no issues.
I do have one doubt about the tag, as you do, paged our sysadmin for opinion.

Thanks for looking into alternative deployments!

mbid_mapping/Dockerfile
mibmo: I can just run [a formatter](https://github.com/reteps/dockerfmt) on them, if you'd like? does change a few more things (mostly indentation)

EDIT: have pushed a change doing just that. should revert and just do `AS` or is this fine?

Dockerfile
zas: The FROM changes make sense and are worth doing, fully qualifying the registry is good for tools that don't assume docker.io as Docker CLI does.

For the `org.metabrainz.based-on-image` label, though, I'd leave it unqualified. That label isn't resolved against a registry by any build tool, it's just metadata recording which upstream base image we built on, used for provenance and base-image update tracking.

MonkeyDo: @mibmo If you wouldn't mind reverting the labels :pray:

mbid_mapping/Dockerfile
MonkeyDo: Thanks, formatting good :)
That'll do nicely, it'll be cleaner now.

MonkeyDo (approved):
Looks good, thanks again!

PR #3867, opened 2026-09-12
Disable to disabled

Aerozol:
Changed the text on the Connect Services page from "Disable" to "Disabled" (https://listenbrainz.org/settings/music-services/details/)

After feedback in this reddit post: https://www.reddit.com/r/listenbrainz/comments/1wedjdp/
Although it's a very minor change, it probably does make sense to have it as the present tense, since it's the default/people don't click to enable it (in other words, they may think there are still steps required to actively disable it).

# Solution

Someone will have to test, sorry!

# AI usage

* [X] I did not use any AI
* [ ] I have used AI in this PR

### DIFF
--- a/frontend/js/src/settings/music-services/details/MusicServices.tsx
+++ b/frontend/js/src/settings/music-services/details/MusicServices.tsx
@@ -586,7 +586,7 @@ export default function MusicServices() {
                   service="spotify"
                   current={permissions.spotify}
                   value="disable"
-                  title="Disable"
+                  title="Disabled"
                   details="You won't be able to listen to music on ListenBrainz or import listens using Spotify."
                   handlePermissionChange={handlePermissionChange}
                 />
@@ -652,7 +652,7 @@ export default function MusicServices() {
                   service="critiquebrainz"
                   current={permissions.critiquebrainz}
                   value="disable"
-                  title="Disable"
+                  title="Disabled"
                   details="You will not be able to publish reviews from ListenBrainz."
                   handlePermissionChange={handlePermissionChange}
                 />
@@ -709,7 +709,7 @@ export default function MusicServices() {
                   service="soundcloud"
                   current={permissions.soundcloud}
                   value="disable"
-                  title="Disable"
+                  title="Disabled"
                   details="You will not be able to listen to music on ListenBrainz using SoundCloud."
                   handlePermissionChange={handlePermissionChange}
                 />
@@ -748,7 +748,7 @@ export default function MusicServices() {
                   service="appleMusic"
                   current={permissions.appleMusic}
                   value="disable"
-                  title="Disable"
+                  title="Disabled"
                   details="You won't be able to listen to music on ListenBrainz using Apple Music."
                   handlePermissionChange={handleAppleMusicPermissionChange}
                 />
@@ -836,7 +836,7 @@ export default function MusicServices() {
                   service="funkwhale"
                   current={permissions.funkwhale}
                   value="disable"
-                  title="Disable"
+                  title="Disabled"
                   details="You will not be able to listen to music on ListenBrainz using Funkwhale."
                   handlePermissionChange={handlePermissionChange}
                 />
@@ -993,7 +993,7 @@ export default function MusicServices() {
                   service="navidrome"
                   current={permissions.navidrome}
                   value="disable"
-                  title="Disable"
+                  title="Disabled"
                   details="You will not be able to listen to music on ListenBrainz using Navidrome."
                   handlePermissionChange={handlePermissionChange}
                 />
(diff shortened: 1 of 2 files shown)


frontend/js/src/settings/music-services/details/components/LFMMusicServicePermissions.tsx
```diff
@@ -224,7 +224,7 @@ export default function LFMMusicServicePermissions({
               className="alert alert-warning alert-dismissible fade show"
               role="alert"
             >
-              Before connecting, you must disable the &#34;Hide recent listening
+              Before connecting, you must  the &#34;Hide recent listening
```
MonkeyDo, line 227: This does not look like it's on purpose

frontend/js/src/settings/music-services/details/components/LFMMusicServicePermissions.tsx
```diff
@@ -139,11 +139,11 @@ export default function LFMMusicServicePermissions({
       toast.success(
         <ToastMsg
           title="Success"
-          message={`${serviceDisplayName} integration has been disabled.`}
+          message={`${serviceDisplayName} integration has been d.`}
         />
       );
 
-      setPermissions("disable");
+      setPermissions("");
```
MonkeyDo, line 146: This shouldn't be changed, it's not a user-facing word but rather an internal coding thing.

frontend/js/src/settings/music-services/details/components/LFMMusicServicePermissions.tsx
```diff
@@ -139,11 +139,11 @@ export default function LFMMusicServicePermissions({
       toast.success(
         <ToastMsg
           title="Success"
-          message={`${serviceDisplayName} integration has been disabled.`}
+          message={`${serviceDisplayName} integration has been d.`}
```
MonkeyDo, line 142: "integration has been d"?
Looks like there is something missing, or that this change is not on purpose

frontend/js/src/settings/music-services/details/MusicServices.tsx
MonkeyDo, line 1: So I'm a bit confused. As it is with those changes, if you are connected to say Spotify, the button that you should click on to disable it will say "Disabled" when it is actually the one you should click to disable integration.

So from what I understand, what we want might be to change from "Disabled" (when not connected) to "Disable" (when connected), is that right?
If so, maybe I can take this one over, it will need more complicated changes, basically different for each "disable" button,  something like:
`title={permissions.spotify === "disable" ? "Disabled" : "Disable"}`

Untested, but should be something along those lines.

MonkeyDo (changes requested):
This one doesn't look as simple as just a find and replace, there is  a bit more to it.
Also looks like some changes were committed by mistake, and a couple ones shouldn't be changed.

frontend/js/src/settings/music-services/details/MusicServices.tsx
Aerozol: Ah sorry, I'm not sure where those removals came from... potentially I accidentally did some kind of GitHub find and replace function, because I definitely didn't mean to delete anything.

The good news is that what I was trying to do is simpler than what you're suggesting - just change "**disable**" (what it says now) to "**disabled**".

Currently it always uses active language (disable), which is confusing to a new user who has not enabled anything. "Disabled" will be a bit less confusing, and still works ok when something has been enabled.

Having the language change depending on if something has already been enabled makes sense, but I'm not sure it's worth spending a lot of time on!

Let me know if that makes sense, and if my PR would work if I redid it without the accidental deletions?

MonkeyDo: I don't think it would be a lot of time, and seems to be what we already do for the connect/connected buttons:
image


Did you see [this PR](https://github.com/metabrainz/listenbrainz-server/pull/3641https://github.com/metabrainz/listenbrainz-server/pull/3641)?
I think that is really the way to go to make things crystal clear, it improves the layout and, more pointedly, shows a status indicator for each service:

image

I think those changes would be sufficient for clarity.
One way or another, we are ready to merge that one, so it would be best to wait and do the changes over there to avoid conflicts (code changed on both sides) which are a pain to resolve.
So let me know if you still want the disable->disabled change I proposed (active language when already connected, passive when not connected), and I will add that to the other PR which already touches all of that code.

MonkeyDo: P.S.: Put that one on test for funsies: https://test.listenbrainz.org/settings/music-services/details/

Aerozol: Ooh, excited for the new update! I think I missed it on test - it looks like this will be less of an issue there based on the screenshot alone. If you would still like to change active to passive language when appropriate, that would be nice, but I don't want to pull you away from more pressing issues. I will close this in the meantime <3

PR #3874, opened 2026-09-16
Write to new listens DB before Timescale and remove migration tooling

amCap1712:
Reverse the order writing and deleting listens by treating the new listens DB as primary and timescale as secondary.

[x] I have used AI in this PR.


listenbrainz/db/listens_schema.py
MonkeyDo, line 1: Note: same content as the deleted file listenbrainz/listenstore/migrate_listens.py except some deletions of migration code, no other modifications

MonkeyDo (commented):
Looks good, nothing much to report.

Are there any changes needed in the sample config.py file to point at the listens DB vs Timescale?

There was some commented out code in dumps manager to revert after the migration, is it for this PR or for the next one?
https://github.com/metabrainz/listenbrainz-server/blob/1123c11416d357670fc9971b3f2cd99015e865b2/listenbrainz/dumps/manager.py#L184-L188

MonkeyDo:
Also PR description/disclaimer please

listenbrainz/db/listens_schema.py
amCap1712: That is for a separate PR.

PR #3876, opened 2026-09-17
LB-1997: Hide volume slider when music player is minimised

PDGamerSG:
# Problem

:beginner: Mention the JIRA ticket (LB-XXX) if there is a relevant one

:beginner: First PR? Have a look at this [beginner's guide](https://github.blog/developer-skills/github/beginners-guide-to-github-creating-a-pull-request/)

LB-1997: https://tickets.metabrainz.org/browse/LB-1997

On mobile, opening the volume slider from the full-screen music player and then minimising the player leaves the slider on screen. The `.volume` panel is fixed-positioned independently of the music player, and `showVolumeSliderAtom` was never reset when the player closed.

# Solution

Added a `useEffect` in `BrainzPlayerUI` that resets `showVolumeSliderAtom` whenever `showMusicPlayer` becomes `false`. An effect is used instead of calling the setter inside the `setShowMusicPlayer` updater, because state updater functions should stay pure (React may call them twice in StrictMode).

The full music player can only be opened on mobile user agents (`isMobile`), so on desktop `showMusicPlayer` stays `false` and the effect does not re-run when the volume icon in the bottom bar is clicked. The desktop volume slider keeps working.

Added a test to `BrainzPlayerUI.test.tsx`: open the music player, open the volume slider, minimise the player, and check that the slider is hidden. The test fails without the fix.

The iOS part of the ticket (volume changes having no effect in iOS browsers) is not addressed here. iOS ignores programmatic volume changes on media elements, so I'd like to confirm the preferred approach (for example hiding the volume control on iOS) before doing that in a follow-up.

| Before (mobile) | After (mobile) |
|---|---|
| before-mobile | after-mobile |

# AI usage

Used Claude Code to help find the relevant code, suggest the fix, write the test, write this PR description, and set up the local development environment.

* [ ] I did not use any AI

#### If you did use AI:

# Action

   1. Add tests
   2. Run an SQL update script

None.


MonkeyDo (approved):
Thank you, simple enough fix!

